PR #7118, opened 2026-09-01
test(app-shell): pin the metadata advisory sink seam — an emitted event reaches the toast, both doors

os-warren:
Fixes #6969

One pin over the advisory chain's **middle link**, asserting the CONNECTION rather than either end of it: a server-sent advisory travels the real hook, the real client factory and the real `MetadataClient`, and arrives at the toast — for the save door and for the publish door.

## Leg 0 — the card's precondition, re-measured on current `main`

The card required the claimant to re-measure that the gap is still real. It is.

**Seam location — assumption falsified.** The card and the dispatch both name `packages/app-shell/src/hooks/useMetadata.ts`. **That file does not exist.** The real seam is `packages/app-shell/src/views/metadata-admin/useMetadata.ts` — the *line numbers* 130-135 were right, the directory was not. `packages/app-shell/src/hooks/useMetadataService.ts` is a different thing (it mints a `MetadataService` from the adapter) and carries no advisory wiring.

**Ablation** — predicted direction stated before running: every named suite stays GREEN, and that green IS the finding.

Cut, at the seam the card names:

```
-    () => createConsoleMetadataClient({ previewDrafts, environmentId, onSaveAdvisory }),
+    () => createConsoleMetadataClient({ previewDrafts, environmentId }),
```

| | value |
|---|---|
| predicted | all named suites stay green |
| observed | all named suites stayed green — **226 files / 2369 tests, zero reds** |
| blob before | `4bbebbf78439ca088ba313a7a708d8a007535a32` (identical to the HEAD blob) |
| blob after | `7d2ade859f42f27f76723fea95c3d4810243b7f1` |
| mutation proven | anchor count for the deleted text 1 to 0, for the injected text 0 to 1, **and** the blob hash moved |
| restore proven | by STATE: blob back to `4bbebbf78...` **and** `git diff HEAD` 0 bytes |

Green suites through the cut (reported because a suite that survives an ablation tells you which layer it does NOT cover):

- `metadata-client.saveAdvisories.test.ts`, `metadata-client.publishAdvisories.test.ts`, `onSaveAdvisory.test.ts` — producer end, green
- `saveAdvisoryToast.test.ts` — renderer end, green
- `MetadataService.saveAdvisories.test.ts`, `metadata-client-auth.ratchet.test.ts` — green
- all 220 suites in `views/metadata-admin/` — green, 2299 passed / 1 skipped, byte-identical to the baseline

Red suites through the cut: **none.**

[...]

PR #7131, opened 2026-09-01
test(app-shell): pin the adapter advisory sink seam — an emitted advisory reaches the toast

### LINKED ISSUE #7116: finding: the SIBLING advisory sink wiring in `AdapterProvider.tsx` is pinned by nothing — the same gap as #6969, on the adapter channel
Found while implementing #6969 (the `useMetadata.ts` seam). Recording, not claiming — no assignee, `priority:*` is triage's.

## The gap

There are **two** parallel advisory wirings in app-shell, not one. #6969 covers the first; this is the second.

| channel | middle link | pinned? |
|---|---|---|
| `MetadataClient` (console metadata-admin clients) | `useMetadata.ts` hands the toast sink to `createConsoleMetadataClient` | ✅ now pinned by #6969 |
| `ObjectStackAdapter` (the adapter every data surface uses) | `AdapterProvider.tsx` subscribes `a.onSaveAdvisory(...)` and calls `emitSaveAdvisories` | ⛔ **nothing** |

The second one is the wiring #4237 built. Both ends of ITS chain are pinned too — the adapter's emitter in `packages/data-objectstack/src/onSaveAdvisory.test.ts`, the renderer in `saveAdvisoryToast.test.ts` — and, exactly as in #6969, nothing asserts the two are connected.

## Measurement (grep, with a control — not an ablation)

Stated plainly so the next reader knows what was and was not measured. This is a grep result, not the green-through-ablation that #6969 carried:

- suites naming `AdapterProvider`: **40**
- of those, naming `dvisor` (matches advisory/advisories/Advisory): **1**
- CONTROL, of those, naming `expect`: **40/40** — so the query is live, and the 1 above is a reading rather than a dead zero

The single hit is `packages/app-shell/src/services/MetadataService.saveAdvisories.test.ts`, and its own docstring says it does not close this gap:

> `AdapterProvider` is what wires the last two links in the real app; the sink is handed over here instead of mounting a toaster, exactly as `saveAdvisoryToast.test.ts` does, so nothing depends on module mocking.

⇒ the subscription in `AdapterProvider.tsx` (the `a.onSaveAdvisory` call and the `emitSaveAdvisories` inside it, plus its unsubscribe) is traversed by no test.

## What a fix looks like

The same shape #6969 landed, one channel over: mount the real provider, have the real adapter emit, assert the sink receives it. ⛔ Not a bigger renderer test and ⛔ not a bigger adapter test — both ends are already covered.

⚠️ Worth re-measuring before claiming, same as #6969: this was measured by grep, so an ablation on `AdapterProvider.tsx` is the leg that actually settles it.

## Related

- **#6969** — the same gap on the `MetadataClient` channel; its pin is the template
- **#4237** — the card that built this adapter wiring
- **#4133** / **#4236** — the save door the shape is inherited from

os-warren:
Fixes #7116

The `ObjectStackAdapter` advisory chain had both ends pinned and its middle link pinned by nothing. This adds one pin that asserts the CONNECTION.

## Precondition — the card was filed as grep-measured, so it was re-measured by ablation

The card said so itself, and it was right to. Its grep reading reproduced, with the denominator moved by one since filing:

| reading | card | re-derived here |
|---|---|---|
| suites naming `AdapterProvider` | 40 | **41** (one landed since) |
| of those, naming `dvisor` | 1 | **1** (same file) |
| CONTROL: of those, naming `expect` | 40/40 | **41/41** |

A sharper census explains the blind spot, and it is starker than the sibling card's: **38 of the 41 mock the module away** with `vi.mock`, and **exactly 0 imported the real `AdapterProvider`**. Control for that zero, same query shape against `saveAdvisoryToast`: **2 hits**, so the zero is a reading and not a dead query.

### The ablation that settles it

Baseline: the 41 suites naming `AdapterProvider` plus the producer and renderer suites — **45 files / 633 tests, green**. Then, three separate cuts, each proven on disk by blob-hash change plus a marker count, each restored by state (`git diff HEAD` empty, `git diff --cached` empty, blob hash back):

| cut | what was removed | result |
|---|---|---|
| H1 | the `a.onSaveAdvisory` subscription, outright | **45/45 files green, 0 red** |
| H2 | the `emitSaveAdvisories` call inside it | **45/45 files green, 0 red** |
| H3 | the unsubscribe in the effect cleanup | **45/45 files green, 0 red** |

Predicted green before running, on the strength of the 0-real-imports census; observed green. The gap is real, and it has three independently cuttable points rather than the one the card assumed.

### It is NOT covered transitively by the sibling pin

Measured, not assumed. With this channel's seam cut (H1), the pin that landed as `7e2e6f42c` stays **green, 3/3**. They are sibling channels: that one is `MetadataClient`, minted per component, sink riding the factory config; this one is `ObjectStackAdapter`, the long-lived shared instance whose sink is a subscribe/unsubscribe registration wired once by the provider (built by #4237).

## The pin

`packages/app-shell/src/providers/AdapterProvider.advisorySink.test.tsx` — three cases: the save renders the finding, a clean save says nothing (with a control that the write really travelled), and the subscription is released on unmount.

The load-bearing design point: **the provider must build its own adapter.** `AdapterProvider` takes an optional `adapter` prop, and passing it makes the effect return early — before the subscription is installed. A test that hands in a ready-made adapter is green against all three ablations above and sees nothing. So nothing is passed; the provider runs its real `init()`.

Real: the provider and its effect, the `ObjectStackAdapter` it constructs, that adapter's save interceptor, the real SDK client and its response parsing, `createAuthenticatedFetch`, `withSettleSignal`, the registration and its unsubscribe, the i18n `t` read through the provider's ref, `readSaveAdvisories`, and `emitSaveAdvisories`. Stubbed, and only these two: `sonner` (the terminal sink, imported as a module binding so it cannot be handed over) and `globalThis.fetch` (the server).

### Red-first evidence

The pin goes red at each cut point, each for the right reason — and H3 discriminates, failing only its unmount case:

| cut | pin result | assertion that fired |
|---|---|---|
| H1 | 2 failed, 1 passed | `expected [] to have a length of 1` |
| H2 | 2 failed, 1 passed | `expected [] to have a length of 1` |
| H3 | **1 failed**, 2 passed | `expected "vi.fn()" to not be called at all, but actually been called 1 times` |

Ablation validity: this lane is **source-resolved** — the root vitest config aliases every `@object-ui/*` to that package's `src`, so no `dist` rebuild stands between the mutation and the run, and a stale build cannot manufacture a false green here.

## Scope

Deliberately not a bigger test at either end — both ends are already covered, and adding to either would grow the suite without closing the hole. The tier, per-finding formatting, door verbs and empty-list drop stay the renderer suite's; response-shape filtering, `mode` derivation and listener isolation stay the producer suite's. No production code changed.

## Verification, at `ad32812de`

- 46 files / 636 tests green (the 45 named suites plus the new pin)
- `type-check` for app-shell: exit 0, 0 errors — re-run after building the dependency closure, since the first attempt failed only on unresolved workspace modules and was NOT MEASURED rather than red
- `check-changeset-presence`: exit 0. Its verdict was quoted rather than predicted — it demanded a declaration, and names an empty frontmatter as "a pass, not a workaround" for a change that releases nothing, which is what the changeset carries
- `changeset:check` (fixed + no-major), `check-changeset-overwrite`, `check-control-bytes`, `lint:coverage`, `type-check:coverage`: all exit 0
- eslint on the new file: exit 0

Declared narrowing: the repo-wide lint scan is CI's and was not run here. Evidence the narrowing excludes nothing — population read from eslint's own config rather than guessed (1038 files under `packages/app-shell`, counted from `--format json`, 0 errors), the new file confirmed present in that population, and no type-aware linting configured (no `projectService`, no `parserOptions.project`; control: the config file reads back `rules` 10 times, `export default` once), so this diff cannot move the verdict of any untouched file.

---

### DIFF
--- a/.changeset/adapter-advisory-sink-pin.md
+++ b/.changeset/adapter-advisory-sink-pin.md
@@ -1,0 +1,4 @@
+---
+---
+
+Test-only change: pins that a metadata-save advisory emitted by the `ObjectStackAdapter` actually reaches the toast sink `AdapterProvider` wires to it. No published behaviour changes.
--- a/packages/app-shell/src/providers/AdapterProvider.advisorySink.test.tsx
+++ b/packages/app-shell/src/providers/AdapterProvider.advisorySink.test.tsx
@@ -1,0 +1,249 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The ADAPTER channel's advisory chain: an emitted event actually REACHES the
+ * sink (objectui#7116).
+ *
+ * Sibling of `views/metadata-admin/useMetadataClient.advisorySink.test.tsx`
+ * (objectui#6969), which pins the same middle link on the OTHER client class.
+ * These are two channels, not one — #6969 covers `MetadataClient`, minted per
+ * component by `useMetadataClient`, whose sink rides the factory config; this
+ * one covers `ObjectStackAdapter`, the long-lived instance every data surface
+ * shares, whose sink is a `subscribe → unsubscribe` registration that
+ * `AdapterProvider` wires once (#4237). Cutting either leaves the other green.
+ *
+ * ## What was missing, and why nothing red said so
+ *
+ * The chain has three links, and before this file only its two ends were
+ * pinned:
+ *
+ *   producer  ObjectStackClient.meta.saveItem, wrapped by the adapter's
+ *             installSaveAdvisoryInterceptor, emits the event
+ *             -> pinned by data-objectstack/src/onSaveAdvisory.test.ts
+ *   MIDDLE    AdapterProvider subscribes with a.onSaveAdvisory and renders
+ *             through emitSaveAdvisories into sonner
+ *             -> pinned by NOTHING
+ *   renderer  emitSaveAdvisories turns an event into the warning
+ *             -> pinned by saveAdvisoryToast.test.ts
+ *
+ * The producer suite asserts the event is EMITTED, into a sink it constructs
+ * itself. The renderer suite asserts the message is BUILT, over an event it
+ * writes by hand. Neither ever asserts the thing emitted and the thing rendered
+ * are connected. Measured before writing this file, on the 45 suites that name
+ * `AdapterProvider` plus the producer and renderer suites (633 tests): with the
+ * `a.onSaveAdvisory` subscription deleted outright, all 45 stayed GREEN; with
+ * the subscription kept but its `emitSaveAdvisories` call cut, all 45 stayed
+ * GREEN; with the unsubscribe cut, all 45 stayed GREEN.
+ *
+ * The mechanism of the blind spot is the same one #6969 named, and starker
+ * here: 38 of those 41 `AdapterProvider` suites `vi.mock` the module away —
+ * legitimately, they are testing pages, not plumbing — and, before this file,
+ * exactly ZERO imported the real one. (Control for that zero: the same query
+ * shape against `saveAdvisoryToast` returns 2.) The seam was mocked away
+ * everywhere it appeared.
+ *
+ * ## Why the provider must build its own adapter here
+ *
+ * `AdapterProvider` takes an optional `adapter` prop, and passing it makes the
+ * effect return EARLY — before the subscription is installed. So a test that
+ * hands in a ready-made adapter cannot see this seam at all: it would be green
+ * against every ablation above. Nothing is passed here; the provider runs its
+ * real `init()`, constructs the real `ObjectStackAdapter`, and the child reads
+ * that instance back out of the context the provider publishes.
+ *
+ * ## What is real here, and what is not
+ *
+ * Real: the provider, its effect, the `ObjectStackAdapter` it constructs, that
+ * adapter's `installSaveAdvisoryInterceptor`, the real SDK `ObjectStackClient`
+ * and its response parsing, `createAuthenticatedFetch`, `withSettleSignal`, the
+ * `onSaveAdvisory` registration and its unsubscribe, the i18n `t` read through
+ * the provider's ref, `readSaveAdvisories`, and `emitSaveAdvisories`.
+ *
+ * Stubbed, and only these two:
+ *   - `sonner` — the terminal sink. `AdapterProvider` imports `toast` as a
+ *     module binding rather than taking it as a parameter, so intercepting the
+ *     module is the only way to observe what arrives. Everything BETWEEN the
+ *     producer and it is the real thing, which is the whole point.
+ *   - `globalThis.fetch` — the server. It answers discovery, and answers the
+ *     metadata PUT with the body the framework's runtime authoring gate
+ *     actually sends.
+ *
+ * ## Scope — deliberately NOT a bigger test at either end
+ *
+ * This asserts the CONNECTION and nothing else. The tier, the per-finding
+ * formatting, the door verbs and the empty-list drop are the renderer suite's;
+ * the response-shape filtering, the `mode` derivation and listener isolation
+ * are the producer suite's. Duplicating any of them here would grow the suite
+ * without closing the hole this file exists to close.
+ */
+
+import { useEffect } from 'react';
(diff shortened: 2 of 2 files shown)


os-warren:

⛔ Landing not armed — 7 checks `in_progress`. 23 completed, 0 failure. I arm on green.

### ⭐⭐ The discovery that saved this card from shipping a decoration

> *`AdapterProvider`'s optional `adapter` prop makes the effect return EARLY, before the subscription is installed. The obvious way to write this pin — hand in a ready-made adapter — yields a test that is green against all three ablations and measures nothing.*

That is the single most valuable line in the report. The natural way to write this test — inject a prepared adapter, drive a save, assert the toast — would have produced a **pin that passes whether or not the seam exists**. It would have closed the card, looked rigorous, and left the hole exactly where it was.

⇒ Passing **no** adapter and letting the provider run its real `init()` is what makes the pin load-bearing. And this is not asserted, it is **proven**: the pin goes RED at all three cuts. A vacuous pin cannot do that.

⚠️ Worth naming the general shape, because it is subtler than the instrument failures this lane has been cataloguing: **an escape hatch in the code under test can silently route your test around the very seam it is pinning.** The test isn't broken and the tooling isn't broken — the fixture is simply too convenient.

### ⭐ THREE hand-offs, not one — my assumption falsified, and H3 is the one that matters

I assumed one. Measured: three independently cuttable points, all inside `AdapterProvider.tsx` — the `a.onSaveAdvisory` subscription, the `emitSaveAdvisories` call nested in its callback, and the **unsubscribe** in the effect cleanup.

Each was ablated separately; each left all 45 suites green. And the pin's ablation **discriminates**: H1 and H2 fail the delivery cases (2 failed / 1 passed, *"expected [] to have a length of 1"*), while H3 fails **only** its unmount case (1 failed / 2 passed, *"expected vi.fn() to not be called at all, but actually been called 1 times"*) — exactly as predicted.

⭐ **A delivery-only pin would have missed H3 entirely.** A leaked subscription after unmount is a real defect with no visible symptom, and the card's own wording (*"plus its unsubscribe"*) is what flagged it. This is the same lesson #6969 produced two hours ago — its "middle link" was two hand-offs, not one — now generalised: **count the hand-offs, do not assume one.**

### The transitivity question was measured, not reasoned

I told you to check whether the just-landed #7118 pin covers this channel and ⛔ not to assume the boundary. Measured directly: `useMetadataClient.advisorySink.test.tsx` stays **GREEN 3/3 under H1**. ⇒ Genuinely separate channels — `MetadataClient` → toast via the hook, versus the `ObjectStackAdapter` channel via the provider. `covered_transitively_by_7118: false` is a reading, not an inference.

### ⭐ You discarded an uncontrolled zero from a GATE, which is the harder call

An earlier `check-changeset-presence` run printed *"0 file(s) changed … no changeset is owed"* — a clean, quotable, green-looking verdict. You discarded it, because the gate reads `git diff` and the new file was **still untracked**. After `git add`, the same gate said exit 1, a changeset **is** owed.

⚠️ That is a trap I could easily have accepted on review: a gate's own verdict line, quoted verbatim, that is nonetheless a measurement of an empty population. ⇒ **A gate that reads the index measures what you have staged, not what you have written.** Going into the seat's standing lessons.

### eslint found a real bug and you fixed it rather than suppressing it

Reassigning a module-scope binding during render (`react-hooks/globals`), repaired by moving the capture into `useEffect`. ⭐ And the three ablation legs were **re-run against the final version** after that change — a fix that lands after the ablation invalidates the ablation, and re-running rather than assuming is the correct instinct.

### Census, controls, narrowing

The `1 of 40` denominator has moved to **1 of 41** (a suite landed since filing) — re-derived rather than quoted, exactly as asked. The zero-import census carries a control chosen deliberately as *a term the target genuinely contains* (`saveAdvisoryToast`, 2 hits), which is the refinement from today's earlier false-control incident. The repo-wide lint narrowing is a three-part **measurement** — population read from eslint's own config, error count from `--format json`, and no type-aware linting configured (with a control proving the config file is readable) — not a shrug.

### Scope

No production code touched ✅. Empty-frontmatter changeset with the gate's own exemption language quoted ✅. Restores proven by state including `git diff --cached` ✅. Every ablation edit script refused to write unless its anchor matched exactly once, so a zero-hit edit could not masquerade as a successful mutation ✅.

---

PR #7120, opened 2026-09-01
fix(app-shell): say what is true when the studio-design designer registries are empty (#6795 part C)

os-warren:
Part of #6795 — part **C** only, exactly as ruled (维护者, 2026-08-30, 第 5 场总监席决裁批 #4, verbatim「同意」): repair the misleading/silent consumer states now, defer part A (registry observability → lazy conversion) to the maintainer's sequencing surface.

⛔ Part A is untouched. No lazy conversion, no subscribe/notify, no `useSyncExternalStore`, no change to `MAX_EAGER_CLOSURE_GZIP_BYTES`, the #6683 `sideEffects` array, or `preview-gallery.tsx`. No registry module is in this diff.

## The mechanism that constrains what these messages may say

The three registries are plain `Map`s filled by a module-scope side effect, and every studio-design consumer reads them **during render with no subscription**. The card measured that a consumer which reads an empty registry **never recovers** when registration lands later:

```
fallback before registration: true | still fallback after registration: true | late inspector rendered: false
```

So every message here states a fact and **never promises recovery** — no "loading…", no "try again", no spinner. That would swap one false statement for another. Making recovery real is part A.

`Suspense` cannot repair any of this: these are *synchronous* reads returning `undefined`, and Suspense catches a thrown promise. Triage fenced this explicitly; it is mechanical, not stylistic.

## Sites — actual locations on this base (`03380aa14`)

The card's line numbers came from `26896c689` and were stale. Re-located:

| site | card said | actual location now | status |
|:--|:--|:--|:--|
| ⭐ Data pillar field rail | `StudioDesignSurface.tsx:3076` | **`:3076`** (unchanged), read at `:2564` | fixed |
| Interfaces canvas + rail | not line-numbered | canvas `:1646`, rail `:1786`, reads at `:1391`/`:1397`/`:1402` | fixed |
| `ObjectActionsPanel` | `:101` | **`:101`** (unchanged), render at `:266` | fixed |
| ⭐ **Automations pillar** | **not on the card at all** | reads at `:3245`/`:3246`, chip `:3520`, rail `:3574` | fixed (fourth site) |
| `ObjectSettingsPanel` | `:75` "silently empty" | `:75` | ⛔ untouched — already correct |
| `ObjectHooksPanel` | `:88` "silently empty" | `:88` | ⛔ untouched — already correct |

## Fourth-site sweep — result, with its control

[...]

PR #7138, opened 2026-09-01
fix(app-shell): no Design mode or block-canvas rail on studio-canvas leaves

### LINKED ISSUE #7121: finding(app-shell): the Interfaces pillar offers Design mode and a "click a block" rail on leaves that have no block canvas — measured with a POPULATED registry
Found while implementing part C of #6795 (PR #7120), swept up but deliberately **not** fixed there: it is a different cause from that card's, so repairing it in that PR would have widened the blast radius past the ruling's scope.

## The observation

`InterfacesPillar` derives `isEditable = !!Preview && !StudioCanvas` — false whenever the leaf resolves to a **studio-canvas surface** rather than a block-tree designer. Today exactly one type opts in: `object`, which renders the runtime records grid (`registerStudioCanvasPreview('object', StudioObjectRecordsCanvas)`). Schema editing for those lives in the Data pillar, by design.

But two affordances beside that canvas are not gated on `isEditable`:

1. **The right rail** falls through to its empty state and says *"Click a block on the canvas, and edit its properties right here."* — on a canvas that is a records grid with no blocks in it. Following the instruction is impossible.
2. **The Design/Run segmented control** (`data-testid="canvas-mode-toggle"`, added by #5800) is still offered. Pressing **Design** flips `canvasMode`, but `StudioObjectRecordsCanvas` takes no `editing` prop, so nothing changes — an inert control that looks live.

## Measured, not inferred — and note the precondition

This is **not** the empty-registry class #6795 part C repaired. It reproduces with the designer registry **populated**:

```
[probe] preview types now: ["dashboard"]
[probe] object-leaf text: Tasks | object · showcase_task | Acme · Navigation | Edit | Tasks | object |
        Design | Run | object · showcase_task | New | ... | Runtime list preview · edit fields /
        structure in the Data pillar | Properties | Click a block on the canvas, and edit its
        properties right here.
```

The registry is demonstrably non-empty (`listMetadataPreviewTypes()` returns a designer), the leaf renders its studio canvas correctly, and the rail *still* instructs the author to click blocks that do not exist. So the cause is the ungated affordance, not a missing registration.

⚠️ Worth stating because it is the reason this is a separate card: the canvas itself is honest here. It even carries a correct hint — *"Runtime list preview · edit fields / structure in the Data pillar"* — right next to a rail contradicting it.

## Why it was not fixed in PR #7120

That PR's ruling (维护者 2026-08-30, 决裁批 #4) scopes part C to the states caused by an **unpopulated** designer registry. Gating the Design toggle on `isEditable` reaches every studio-canvas leaf, where the registry is populated and the cause is different — a behaviour change on a surface #5800 owns, outside that fence. Filed here instead so the decision is taken deliberately.

## Not scoped here

Which of the two repairs is right. Options a taker should weigh rather than assume:

- **Rail:** give the studio-canvas branch its own empty state (something like "this surface is edited in the Data pillar", echoing the hint already on the canvas), instead of the block-tree invitation.
- **Toggle:** hide it, disable Design, or let `StudioCanvas` implementations opt into an `editing` prop so the control becomes live rather than removed. #5800's acceptance ("设计⇄运行 is a round trip, same renderer") is the constraint to respect — check `StudioDesignSurface.canvasMode.test.tsx` before changing its shape.

Both land in `packages/app-shell/src/views/studio-design/StudioDesignSurface.tsx` (`InterfacesPillar`).

## Related

#6795 (part C, where this was found) · PR #7120 · #5800 (added the Design/Run switch) · the `registerStudioCanvasPreview` extension point in `views/studio-design/studio-canvas-preview.tsx`

os-warren:
Fixes #7121

The Interfaces pillar offered **Design mode** and a *"Click a block on the canvas, and edit its properties right here."* rail beside studio-canvas leaves that have no block canvas at all. Both instructions are impossible to follow there.

## Leg 0 — reproduced on current `origin/main`, NOT already repaired by #7120

Measured on `220c18d05` (which contains #7120 as `994b73696`), with the designer registry **populated** — the card's precondition:

```
[A] preview types now: ["dashboard"]
[A] studio-canvas types: ["object"]
[A] toggle present: true
[A] rail says click-a-block: true
[A] canvas hint present: true
```

The canvas beside that rail is honest — it carries *"Runtime list preview · edit fields / structure in the Data pillar"* — while the rail contradicts it. So this is the ungated affordance, not the empty-registry class #6795 part C repaired.

## Leaf population — enumerated before editing

`registerStudioCanvasPreview` call sites across the whole repo:

| site | type | has a block canvas? |
| --- | --- | --- |
| `studio-canvas-preview.tsx:97` (built-in default) | `object` | no — runtime records grid |
| `studio-canvas-preview.test.tsx:43,48` | `object` (override, restored) | test-only |

**Population: one** (`object`), confirmed at runtime by `listStudioCanvasPreviewTypes() === ["object"]`. The registry is a public extension point, so the gate is written against the registry, not against the string `object`.

## ⚠️ `isEditable` is a proxy, and gating on it would have broken a landed pin

The card sketched gating on `isEditable`. That is **not** the right discriminator, and this is measured rather than argued.

`isEditable = !!Preview && !StudioCanvas` — a conjunction of two independent causes. It is false both for a studio-canvas leaf *and* for a leaf whose own type simply has no designer. The second is the state #6795 part C **pinned as still deserving the ordinary rail**, in `StudioDesignSurface.designerRegistryPartial.test.tsx`.

Ablation, direction predicted before running — swap `StudioCanvas` for `isEditable` at both gates, then run that landed pin. Predicted RED, observed RED:

```
MUTATED: 5209c038d421b8410c2ed7049d70eaf610afcda3 -> a80d71a6dc06e2f3200dfab970ad47b496821e16
grep proof -> !StudioCanvas remaining: 0 | isEditable gate: 1
 × says no designer is registered for this type, not that none loaded at all
 × keeps the ordinary "click a block" empty state when designers ARE registered
TestingLibraryElementError: Unable to find an element by: [data-testid="canvas-mode-toggle"]
 Tests  2 failed (2)
```

Restore proven by state: blob hash back to `5209c038d…`, `git diff HEAD` empty, `git diff --cached` empty, `!StudioCanvas` gate count back to 1.

⇒ the gate is `StudioCanvas` — the same value that selects the canvas branch. Blast radius is exactly the studio-canvas leaves.

## The repair

Three surfaces in `StudioDesignSurface.tsx` (`InterfacesPillar`), all reached through the shared `canvasEl` / `inspectorBodyEl`, so all three layouts — classic, folded-wide, folded-tabs — are covered by one change.

1. **Design/Run switch** gated on `!StudioCanvas`. `editing` is handed to exactly one canvas branch (`Preview`); `StudioCanvasPreviewProps` carries no `editing` by contract, so the switch moved `canvasMode` and reached no renderer.
2. **The rail** gets a studio-canvas branch stating what the canvas is. It promises no recovery — and here the constraint is stricter than part C's: the statement is not about registration at all, so there is nothing to wait for.
3. **Ordering + the header button** — a bounded in-place addition, declared rather than slipped in. The new rail branch is ordered **ahead of** the `selection` branch, and the header's "clear selection" button is gated with it. Without this, my own repair would have left the rail saying "no blocks here" beside a button offering to clear a selection.

That third item is a fix for a defect I measured while here:

```
[D] scoped inspector rendered on studio-canvas leaf: true
[D] inspector was handed type/name: object:showcase_task:block:blk_1
```

`blk_1` is a block on a *different* leaf's canvas — `selection` outlives a leaf change because the load effect's `setSelection(null)` sits inside the `isEditable` early-return. The rail symptom is closed here; the **underlying state leak reaches every non-editable leaf**, so it is filed separately as #7137 rather than swept in.

## New string

`engine.studio.inspector.studioCanvasNoBlocks`, added to `en` and `zh`.

⚠️ Worth recording: `engine.*` strings do **not** live in `packages/i18n/src/locales/` and are not covered by `all-locales-key-parity`. They are a flat dotted-key table with `en` and `zh` only, in `views/metadata-admin/i18n.ts`, whose header states the carve-out and which `check-i18n-call-site-keys.mjs` skips by declaration. `check:i18n-drift` confirms the packs did not move: *"0 en value(s) changed"*.

## Pins

`StudioDesignSurface.studioCanvasLeaf.test.tsx` — 7 tests. Four pin the studio-canvas leaf; **three are the regression fence** on a leaf that does have a block canvas, which matters more: the switch is still offered and still round-trips (`editing` true → false → true, #5800's acceptance), the ordinary "click a block" rail is untouched, and the selection affordances still work.

Ablation, direction predicted before running — revert the source fix, keep the tests. Predicted RED on the four, green on the fence:

```
MUTATED: 5209c038d421b8410c2ed7049d70eaf610afcda3 -> d5abcd2d3ab77e08c91ba811283d1d28d46cd383  (markers now: 0)
 × offers no Design/Run switch on a leaf whose canvas cannot read the mode
 × replaces the impossible "click a block" invitation with what is true
 × promises no recovery — there is nothing to wait for
 × does not open a scoped inspector for a block selected on a DIFFERENT leaf
 Tests  4 failed | 3 passed (7)
```

Restore proven by state: blob back to `5209c038d…`, `git diff HEAD` and `git diff --cached` both empty, markers back to 3 / 1.

## Gates — all run on `e462194d7`, the final commit

| gate | exit | verdict |
| --- | --- | --- |
| `vitest packages/app-shell/src/views/studio-design/` + locale parity | 0 | `Test Files 43 passed (43) · Tests 257 passed (257)` |
| `pnpm --filter @object-ui/app-shell run type-check` | 0 | `tsc --noEmit && tsc -p tsconfig.test.json`, 0 `error TS` |
| `eslint` (plain form, 3 changed files) | 0 | 0 errors, 20 warnings — **identical to the 20 on the same file at `220c18d05`**, none inside my hunks; new test file and `i18n.ts` contribute 0 |
| `check-changeset-presence` | 0 | `✅ 3 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` |
| `check-changeset-fixed` | 0 | `✅ All workspace packages are in the changeset fixed group.` |
| `check-changeset-no-major` | 0 | `✅ No changeset declares a major bump.` |
| `check:i18n-keys` | 0 | `Every in-scope call-site key resolves against the en pack (2845 keys)` |
| `check:i18n-drift` | 0 | `No en value changed in this range.` |
| `check:control-bytes` | 0 | `✅ OK (scanned 5909 tracked text file(s))` |
| `check-vi-mock-specifiers` | 0 | `✅ OK (771 relative specifier(s) resolved)` |
| `vitest-invocation-guard` | 0 | clean |
| `check-designer-field-key-parity` | 0 | `designer-field-key-parity: OK` |
| `check-lucide-icon-record-names` | 0 | `OK lucide icon names: 182 …` |

Gate set derived from the repo, not guessed: only `check-i18n-call-site-keys.mjs` references `studio-design` / `StudioDesignSurface` / `metadata-admin/i18n` among `scripts/*.mjs` (control: 5 scripts do reference `packages/app-shell`, so that zero is a reading). The remaining rows are the app-shell-reading gates plausibly touched by a JSX + i18n diff.

`typecheck` genuinely covers the new test: `tsc -p tsconfig.test.json --listFiles` lists `studioCanvasLeaf.test.tsx` (1 hit; control `StudioDesignSurface.tsx` also 1; 4508 files total) — so the second `tsc` invocation is not excluding tests.

## Observation, not changed here

The canvas hint below the grid is gated `!isEditable && current?.type === 'object'` — hardcoded to the type rather than to the registry. A second `registerStudioCanvasPreview` leaf would get the gating in this PR but no hint, because the hint's wording ("edit fields / structure in the Data pillar") is object-specific and generalising it is a wording decision. Latent only: population is 1 today. Left for triage.

## Base

Branched from `220c18d05`; `origin/main` has since moved to `2b3964aff`. All measurements above are on `e462194d7`.

---

### DIFF
--- a/.changeset/7121-studio-canvas-leaf-affordances.md
+++ b/.changeset/7121-studio-canvas-leaf-affordances.md
@@ -1,0 +1,31 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Studio Interfaces: no Design mode and no "click a block" rail on leaves that have
+no block canvas (objectui#7121).
+
+`registerStudioCanvasPreview(type, …)` opts a type into a surface-specific canvas
+that renders the running app rather than an editable draft — a contract, not a
+habit: `StudioCanvasPreviewProps` carries no `selection`, `onSelectionChange`,
+`onPatch` or `editing`. Two affordances beside such a leaf ignored that.
+
+- The Design/Run switch (objectui#5800) was still offered, though `editing` is
+  handed to exactly one canvas branch (`Preview`). On a studio-canvas leaf the
+  switch moved `canvasMode` and reached no renderer — a live-looking control
+  wired to nothing. It is now gated.
+- The right rail fell through to "Click a block on the canvas, and edit its
+  properties right here." beside a canvas that has no blocks, so the instruction
+  could not be followed. It now states what the canvas is, and — because this
+  canvas has no blocks by contract — promises no recovery.
+- The rail's new branch is ordered ahead of the selection branch, so a block
+  selected on a *different* leaf no longer opens a scoped inspector for a block
+  this canvas does not contain; the header's "clear selection" button is gated
+  with it.
+
+The discriminator is `StudioCanvas`, not `isEditable`. `isEditable` is
+`!!Preview && !StudioCanvas` — a conjunction of two independent causes — so
+gating on it would also strip these affordances from leaves whose only fault is
+that their own type has no designer, the state objectui#6795 part C pinned as
+still deserving the ordinary rail. Behaviour on every leaf with a block canvas
+is unchanged.
--- a/packages/app-shell/src/views/metadata-admin/i18n.ts
+++ b/packages/app-shell/src/views/metadata-admin/i18n.ts
@@ -1777,6 +1777,12 @@ const ENGINE_STRINGS_EN: Record<string, string> = {
   'engine.studio.inspector.tabSource': 'Source',
   'engine.studio.inspector.emptyLine1': 'Click a block on the canvas,',
   'engine.studio.inspector.emptyLine2': 'and edit its properties right here.',
+  // objectui#7121 — the rail beside a studio-canvas leaf (a
+  // `registerStudioCanvasPreview` type: the running app, not a block tree).
+  // States what the canvas IS; promises no recovery, because there is nothing
+  // to wait for — `StudioCanvasPreviewProps` carries no selection by contract.
+  'engine.studio.inspector.studioCanvasNoBlocks':
+    'This canvas renders the running app, not a block tree — it has no blocks to select, and nothing here is edited from this panel.',
   'engine.studio.inspector.designersMissing':
     'No metadata designers are registered in this session, so there is nothing to edit here.',
   'engine.studio.inspector.noPageSchema': 'Page settings are unavailable — the page schema could not be loaded.',
@@ -3639,6 +3645,8 @@ const ENGINE_STRINGS_ZH: Record<string, string> = {
   'engine.studio.inspector.tabSource': '源码',
   'engine.studio.inspector.emptyLine1': '在画布里点选一个积木,',
   'engine.studio.inspector.emptyLine2': '它的属性会在这里直接编辑。',
+  'engine.studio.inspector.studioCanvasNoBlocks':
+    '此画布渲染的是运行态应用，而不是积木树 —— 这里没有可选中的积木，也没有可在本面板编辑的内容。',
   'engine.studio.inspector.designersMissing': '本次会话没有注册任何元数据设计器,这里没有可编辑的内容。',
   'engine.studio.inspector.noPageSchema': '页面设置不可用——无法加载页面 schema。',
   'engine.studio.inspector.sourcePageLine1': '这个页面是 {kind} 源码,不是积木树 ——',
--- a/packages/app-shell/src/views/studio-design/StudioDesignSurface.studioCanvasLeaf.test.tsx
+++ b/packages/app-shell/src/views/studio-design/StudioDesignSurface.studioCanvasLeaf.test.tsx
@@ -1,0 +1,300 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * objectui#7121 — what the Interfaces pillar may OFFER beside a studio-canvas
+ * leaf.
+ *
+ * ## The mechanism
+ *
+ * `registerStudioCanvasPreview(type, …)` opts a type into a surface-specific
+ * canvas: the running app, not an editable draft. That is a contract, not a
+ * habit — `StudioCanvasPreviewProps` is deliberately a small read-only subset
+ * of `MetadataPreviewProps` with **no `selection`, no `onSelectionChange`, no
+ * `onPatch`, no `editing`**. So a studio-canvas leaf:
+ *
+ *   - can never produce a block selection (there is no block tree), and
+ *   - can never read the design/run mode (nothing is handed `editing`).
+ *
+ * Two affordances beside it ignored both facts: the Design/Run switch (#5800)
+ * was still offered though `canvasMode` reached no renderer, and the rail fell
+ * through to *"Click a block on the canvas, and edit its properties right
+ * here."* — an instruction that cannot be followed.
+ *
+ * ## ⚠️ This is NOT #6795 part C's cause, and the precondition says so
+ *
+ * Part C repaired what the pillar says when the designer registries are
+ * **empty**. Every test here asserts a **POPULATED** registry first
+ * (`listMetadataPreviewTypes()` non-empty), because a zero-registry reading
+ * would measure that other card instead. The cause here is an ungated
+ * affordance, not a missing registration.
+ *
+ * ## ⛔ The message promises no recovery
+ *
+ * Part C established by measurement that these registries are plain `Map`s read
+ * during render with no subscription, so a consumer that read an empty one
+ * never recovers ("late inspector rendered: false") — no "loading…", no "try
+ * again". Here the constraint is even stricter: the statement is not about
+ * registration at all. This canvas has no blocks **by contract**, so there is
+ * nothing to wait for, and the last test pins the absence of recovery language.
+ *
+ * ## ⚠️ The discriminator is `StudioCanvas`, NOT `isEditable`
+ *
+ * `isEditable = !!Preview && !StudioCanvas` is a conjunction of two independent
+ * causes. Gating on it would also strip these affordances from leaves whose
+ * only fault is that **their own type** has no designer — the exact state
+ * `StudioDesignSurface.designerRegistryPartial.test.tsx` pins as still deserving
+ * the ordinary "click a block" rail. That file is the live fence: gate on
+ * `isEditable` and it goes red.
+ */
+
+import '@testing-library/jest-dom/vitest';
+import * as React from 'react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+const objectDef = {
+  name: 'showcase_task',
+  label: 'Task',
+  fields: [{ name: 'title', label: 'Title', type: 'text' }],
+};
+
+/**
+ * One leaf of each kind, in one app: `object` opts into the studio canvas,
+ * `dashboard` does not (it is an ordinary block-canvas designer). The contrast
+ * is the point — the second leaf is the regression fence.
+ */
+const NAV = [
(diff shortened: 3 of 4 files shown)


os-warren:

Landing armed once CI converges. ⚠️ `behind` is expected — you branched from `220c18d05` and `main` is now `2b3964aff`; the queue rebuilds each entry, and I re-run the pre-check before arming.

### ⭐⭐ My prescribed discriminator would have BROKEN a pin that landed 40 minutes earlier

My ZONE 1 said to gate on `isEditable`. You did not take it, and you did not merely argue against it — **you ablated it**: swapping `StudioCanvas` for `isEditable` at both gates turns the **landed** part-C pin `StudioDesignSurface.designerRegistryPartial.test.tsx` **red** (2 failed / 2, on the missing `canvas-mode-toggle`).

The reason is structural: `isEditable = !!Preview && !StudioCanvas` is a **conjunction of two independent causes**, so it is also false for a leaf whose own type merely has no designer — precisely the state #6795 part C pinned as still deserving the ordinary rail. ⇒ My gate would have silently widened the blast radius from "studio-canvas leaves" to "every leaf without a designer", regressing work that landed the same hour.

⭐ **This is the strongest form of falsification available**: not "I think your discriminator is a proxy" but "here is the landed test it turns red." Shipping on `StudioCanvas` — the same value that selects the canvas branch — keeps the change to exactly the leaves it is about.

⚠️ For my own record: my ZONE 2 asked you to check whether `isEditable` was a real discriminator or a correlated proxy, and *"a proxy that drifts is how this class of defect is born."* It was a proxy. I should not have named it in ZONE 1 as the route while asking in ZONE 2 whether it was sound.

### I verified the leaf population independently

Exactly **one** non-test `registerStudioCanvasPreview(...)` call site on `main` — `studio-canvas-preview.tsx:97`, registering `'object'`. The other matches are the function definition, two doc comments, one comment in `StudioDesignSurface`, and two test overrides. Control (`registerMetadataPreview`) hits multiple files, so the count is a reading. ⇒ Population 1, as reported, and the "reaches every leaf" worry I raised is bounded.

### Two surfaces I did not predict — and the split between fix and file is right

My order named two affected surfaces. You measured **four**: the toggle, the rail, a **stale-selection scoped inspector** (a block selected on a *different* leaf survives the leaf change and opens `ObjectFieldInspector` on it), and the header's leaked "Clear selection" button.

⭐ And the underlying cause was **filed, not fixed**: `setSelection(null)` sits inside the `isEditable` early-return, so selection outlives a leaf change on **every** non-editable leaf — a strictly larger population than this card's fence, also feeding `hasInspectorTarget` and the folded-layout `centerTab` auto-switch. Filed as **#7137** (triaged p2). Fixing the rail symptoms here while routing the state leak is the correct cut: the symptom is in scope, the cause is not.

The clear-selection button is declared as a **bounded in-place** addition with its reason — leaving it would have made the rail contradict the header after fix 2. Declared rather than absorbed.

### ⚠️ My i18n instruction was wrong, and you found the real home via a controlled zero

I told you: `en.ts` **and its nine sibling packs**, `all-locales-key-parity` green. **That does not apply here.** `engine.*` strings are a flat dotted-key table with **en and zh only**, in `views/metadata-admin/i18n.ts`, whose header documents the carve-out and which `check-i18n-call-site-keys.mjs` skips **by declaration**.

⭐ And the way you established it is the point: the key returned **0** in `packages/i18n/src/locales/en.ts`, and rather than reading that as "not added yet" you took controls that must hit (the literal rail string → 4 files; `studio` in `en.ts` → 3) — which proved the instrument worked and led you to the real table. Corroborated by `check:i18n-drift`: *"0 en value(s) changed."*

### ⭐ The recorder control — the ninth instrument failure this session

Your probe's first run wrote **eight empty lines**, because bash expanded `$1` inside a double-quoted perl replacement before perl saw it, so `rec()` was called with no arguments. **Exit 0. `Tests 2 passed`.** Nothing errored.

You caught it because the byte count was implausible, then rewrote it with a quoted heredoc **and gave it a must-hit control** (`[C] control-must-hit: 42`) present in every subsequent run. ⇒ That is the correct repair: not "be careful with quoting" but "make the recorder itself carry a control."

### ⭐ You caught your own false red before committing

The *"promises no recovery"* test initially scanned the **document** and read the real records canvas's `Error loading grid` (jsdom, no data source) as the rail promising recovery. Scoped to the rail element, with a control asserting the rail contains the message before asserting the absences. ⇒ A negative assertion scoped too widely is a false positive waiting to happen, and you found it in your own work.

### Scope

Population-1 gating on the real discriminator ✅. No recovery promised, and **pinned** rather than asserted ✅. Ablation predicted 4 red / 3 green and observed exactly that, with restores proven by state including `git diff --cached` ✅. eslint baseline compared against the same file at `220c18d05` to show the 20 warnings are pre-existing ✅. Changeset staged-then-run, both verdicts quoted ✅.

⚠️ The unfiled observation you reported — the canvas hint gated on `current?.type === 'object'` rather than on the registry — is correctly left as triage input: latent at population 1, and generalising object-specific wording is a decision, not a mechanical fix. Noted on the seat's list.

---

PR #7124, opened 2026-09-01
fix(plugin-dashboard,i18n): a self-explaining default empty state for dashboard widgets

os-warren:
Fixes #7063

Maintainer ruling 2026-08-31 (hotcrm#1212 adjudication, verbatim 「1212 也是」, following 「hotcrm#1203 …这种也是平台的问题啊」): this class is the platform's to fix uniformly; apps must not compensate widget by widget (objectstack#13848 — apps are simplified business implementations, uniform behaviour belongs to the platform).

All evidence below was measured at **79accd66f** (the final commit on this branch).

---

## The card's premise was half wrong, and that changed the shape of the fix

Two of the dispatch's stated assumptions did not survive measurement. Both are reported here rather than worked around.

### 1. `暂无数据行` is NOT one shared placeholder — there was no seam to fix

Three dashboard surfaces wrote their own empty state, in **two different strings**, with no common code:

| surface | key | en | zh |
|---|---|---|---|
| `DatasetWidget` (the measured one) | `dashboard.noRows` | `No rows` | `暂无数据行` |
| `ObjectDataTable` | `dashboard.noDataAvailable` | `No data available` | `暂无数据` |
| `PivotTable` | `dashboard.noDataAvailable` | `No data available` | `暂无数据` |

`dashboard.noRows` had exactly **one** call site repo-wide (`DatasetWidget.tsx:727`), so the measured `暂无数据行` came from the dataset-bound path only — which is the ADR-0021 governed path the hotcrm quota-attainment tile takes.

So "uniformly" could not mean "fix the shared placeholder". The new `WidgetEmptyState` **is** the seam: it is added, and the three surfaces are routed through it.

### 2. The authored `emptyState` override does not exist on this surface

Clause 3 assumes an existing authored `emptyState` contract that always overrides. Measured:

- `emptyState` occurs **0 times** in `packages/plugin-dashboard/src` and `packages/plugin-charts/src`. Control in the same sweep, same paths: `widget`, **1154** hits — so the zero is a reading, not a broken query.
- `emptyState` is a **list-view** contract (`ObjectGridSchema`, `NamedListView` in `packages/types/src/objectql.ts`), honoured in `packages/plugin-list/src/ListView.tsx`.
- `@objectstack/spec@17.2.0`'s `DashboardWidgetSchema` declares no such key. Its full key set, read from the built package at runtime rather than recalled:

[...]

PR #7139, opened 2026-09-01
fix(plugin-charts): a self-describing empty state for ObjectChart over an empty result

### LINKED ISSUE #7130: finding(plugin-charts): `ObjectChart` has NO empty branch at all — an empty result draws a bare chart frame, the fourth distinct answer on this surface to "is it broken or is it young"
Filed by the `domain:ui` seat (session `session_012wwHa4aaFybxXrfmfHioDM`) out of #7063 / PR #7124.

⚠️ **Provenance, stated plainly:** the implementer measured this while sweeping for #7063's placeholder sites and **deliberately did not file it**, on the grounds that a chart frame with axes may be self-describing and they would rather the seat decide than pre-file on their own reading. That restraint was correct as a default. **The seat is overriding it and filing** — for the three reasons below, not because the reading was wrong.

## The measurement

`packages/plugin-charts` `ObjectChart` has **no empty branch**. Over an empty result it draws an empty chart frame with no message — not a placeholder, not an explanation, nothing.

This sits alongside three *other* answers to the same question on the same surface, which #7063 has just been consolidating:

| surface | what an empty result rendered, before #7063 |
|---|---|
| `DatasetWidget` | `dashboard.noRows` — a terse fragment (the string hotcrm#1212 measured as indistinguishable from a load failure) |
| `ObjectDataTable` | `dashboard.noDataAvailable` — a *different* terse key |
| `PivotTable` | `dashboard.noDataAvailable` **with its title suppressed**, so only the description rendered |
| **`ObjectChart`** | ⛔ **nothing at all** |

⇒ Four surfaces, four answers. PR #7124 converges the first three onto one seam (`WidgetEmptyState`) and leaves this one untouched, correctly — the ruling's measured scenario is a table tile, and this is a different package.

## ⛔ This card does NOT assert the current behaviour is wrong

Carrying the implementer's reading forward verbatim, because it is the honest one and a future claimant should not inherit a conclusion the measurement does not support:

> *it is an observation about a state that may well be intentional (a chart frame with axes is arguably self-describing)*

⚠️ **A chart is genuinely not a table.** An empty table is a blank rectangle and says nothing; an empty chart still renders labelled axes, which may already tell the reader *what would have been plotted*. Whether that clears the bar the maintainer set in hotcrm#1212 — *distinguishable from a load failure at a glance* — is **unmeasured**, and it is the question this card exists to route, not to answer.

⇒ **First deliverable for whoever takes it: look at it.** Render `ObjectChart` over an empty result and judge whether the frame alone reads as a state or as a failure. If it reads as a state, close this card as measured-and-declined — that is a legitimate and valuable outcome, not a failed dispatch.

## Why it is filed anyway

1. It is a **measured absence** (no empty branch exists), not an aesthetic judgement — the kind of fact that should be in the tracker rather than in a transcript.
2. PR #7124 **builds the seam**, so if a fix is wanted it is now small. The cost of the card is low precisely because the hard part just landed.
3. ⭐ An observation that lives only in a subagent report is **lost**. This lane has already paid that cost: #7070 exists only because an earlier implementer's dedupe channel was rate-limited and they handed the finding back rather than dropping it.

## ⛔ Sequencing

⚠️ **Do not start before PR #7124 lands.** It creates the `WidgetEmptyState` seam this would route through; starting earlier means designing the same thing twice — the failure mode the #7063/#7064 dispatch pair was explicitly fenced against.

⚠️ Note the package boundary: the seam is in `@object-ui/plugin-dashboard` and `ObjectChart` is in `@object-ui/plugin-charts`. Whether the seam should move, be imported across, or be duplicated is a real design question the claimant must **measure and report**, ⛔ not assume — and if the answer is "promote it to a shared package", that is a cross-surface abstraction and belongs to the convergence card below, not to this one.

## Related

- **#7063** / PR #7124 — the ruling this came out of; converges the three dashboard surfaces.
- **#7064** / PR #7123 — the sibling ruling, same principle, record-detail sections.
- ⭐ **Convergence, not yet filed:** `DataEmptyState` (`packages/components/src/custom/view-states.tsx`) is already consumed by `plugin-list`, `plugin-kanban`, `plugin-dashboard` **and** `plugin-detail`. PR #7124's implementer identified changing its defaults as a smaller fix spanning both rulings' surfaces, and **declined it on purpose** because two agents designing one cross-surface abstraction in parallel is worse than two local fixes. The seat will sequence that as its own card once #7123 and #7124 have both landed.
- **#7125** — the dead-key residue from #7124.

os-warren:
Fixes #7130

## The card asked for a measurement first, so here is the measurement

`ObjectChart` was rendered in a real browser (Chromium, console dev server) over a real empty result — a stub dataSource whose `find` resolves to `[]` — beside the two states the hotcrm#1212 bar asks it to be distinguishable from, at `220c18d05`.

**The filing hypothesis is false.** The card carried the original implementer's reading forward verbatim: *"a chart frame with axes is arguably self-describing"* — an empty table is a blank rectangle, but an empty chart still draws labelled axes telling the reader what would have been plotted. It does not. Recharts derives its ticks **from the data**, so an empty domain labels nothing:

| tile | `text` nodes in the SVG | marks | what a reader sees |
|---|---|---|---|
| empty result, bar | **0** | 0 paths, 2 lines | two hairline rules on white |
| empty result, line | **0** | 0 paths, 2 lines | two hairline rules on white |
| empty result, pie | **0** | 0 paths, 0 lines | an empty box |
| populated **control**, bar | **8** (`Qualify`, `Propose`, `Won`, `0`…`24`) | 3 paths, 5 lines | a chart |

Zero characters render on an empty chart. So it fails both halves of the bar at once: it is not *self-describing without authored copy*, and beside this component's own red `chart-error` box — which at least says "Failed to load chart data" — a blank tile is **less** informative than the failure it must be told apart from. Verdict: it reads as a failure, not as a state.

⇒ measured-and-declined was a real possible outcome here and this is not it.

## Two PM assumptions were falsified, and the second changes the argument

**The KPI precedent does not apply — the landed code already says so.** `DatasetWidget` carves out metric families and names charts on the *other* side of the line, in the comment `220c18d05` shipped:

> `// A metric (single value) over an empty dataset is 0, not an empty state —`
> `// the latter reads as broken for KPIs like "Total Books" on a fresh app.`
> `// Charts and tables keep the empty state (there is genuinely nothing to plot).`

`METRIC_TYPES` is `metric | kpi | gauge | solid-gauge | bullet`; bar/line/pie are not in it. So a **dataset-bound chart has rendered the self-describing empty state since #7124, and the object-bound chart did not** — same family, same empty result, two different answers depending on which widget drew it. A KPI's `0` is a datum; a chart's blank frame is an absence. This is the surface that ruling did not reach, not a new judgement.

**The chart-side population is a single surface, swept with a control.** Of the 9 non-test modules in `packages/plugin-charts/src`, exactly one takes a `dataSource` and can hold an empty query result: `ObjectChart` (control: `grep -l import` hits 7 of the 9, so the channel reads). It is registered under **two** keys — `object-chart` and `view:chart`, both via `ObjectChartBlock` — so one branch covers both. The other two registrations (`plugin-charts:chart`, `bar-chart`) take authored inline data and are untouched. One sub-path noted for the record: `AdvancedChartImpl`'s sankey arm returns a bare `div` when `links.length === 0`, and `hasNoCategoryKey` / `hasNoPlottableSeries` are both gated `rows.length > 0`, so neither refusal ever fired on an empty result.

## Package boundary — measured, not assumed

| option | verdict |
|---|---|
| import `WidgetEmptyState` across from `plugin-dashboard` | **rejected.** It is absent from that package's `index.tsx` (control: 7 `export` hits in the same file), i.e. package-private, and `plugin-charts` has no dependency on `plugin-dashboard`. Reaching it means promoting it to public API for a foreign plugin — the cross-surface abstraction #7132 owns. No cycle would be created (`plugin-dashboard` does not depend on `plugin-charts`), so this was ruled out on ownership, not mechanics. |
| duplicate the composition | **rejected** — a third copy is what the seam exists to prevent. |
| promote to a shared package | **not taken.** That is #7132's job and it stays open for it. |
| **consume `DataEmptyState` from `@object-ui/components`** | **taken.** Already shared, already consumed by `plugin-list` / `plugin-kanban` / `plugin-detail` and by `WidgetEmptyState` itself; already a dependency of `plugin-charts`. |

Net: **no package promoted, no dependency edge added, no new i18n key, no new spec key.** The copy is the keys #7124 put in all ten packs (`dashboard.empty.title` / `.message` / `.sourceLabel`) — and `ObjectChart` already reads `dashboard.openInList`, so the namespace is precedented here. When #7132 converges the defaults, this becomes a fifth call site that collapses exactly like the other four.

No message promises recovery: it states that the load **succeeded**, which is the one fact the reader of a blank tile cannot otherwise get. `role="status"` against the `role="alert"` on `chart-error` is the machine-readable half of the distinction. Inline-data charts are byte-for-byte unchanged — they ran no query, so "its query returned no records" would be false of them.

## Evidence

**The pin can fail — ablation, direction predicted before running.** Disabling the branch should fail the two empty-asserting arms and leave the three negative ones green.

- mutation proven on disk before the run: marker `if (isQueryBacked && !boundData` 1 to 0, injected marker 0 to 1, blob `6d012917` to `ca37f4c2`
- result: **`Tests 2 failed | 3 passed (5)`** — as predicted, with the passing count asserted so a collapsed suite could not masquerade as the red
- restore proven by state: `git diff HEAD` and `git diff --cached` both empty, blob back to `6d012917`, markers back to 1 / 0

**Gates, at final commit `5e5d07ce1`** — each verdict quoted from the gate's own output, exit captured before any pipe:

| gate | exit | verdict |
|---|---|---|
| `vitest run packages/plugin-charts` | 0 | `Test Files 36 passed (36)` · `Tests 245 passed (245)` |
| `vitest run packages/plugin-dashboard packages/plugin-view apps/console` | 0 | `Test Files 201 passed (201)` · `Tests 2019 passed (2019)` |
| `vitest run .../all-locales-key-parity.test.ts` | 0 | `Tests 32 passed (32)` |
| `type-check` (plugin-charts) | 0 | clean; `--listFiles` confirms **both** edited files are in the checked set (1 hit each of 1362) |
| `eslint` (plain form, changed files) | 0 | `61 problems (0 errors, 61 warnings)` — 0 warnings inside the inserted lines; the test file's 3 match the sibling test's own profile |
| `check:i18n-keys` | 0 | `2605/2605 literal keys resolve` · every inline default matches the value the pack serves |
| `check:i18n-drift` | 0 | `No en value changed in this range.` |
| `check:control-bytes` | 0 | `OK (scanned 5909 tracked text file(s))` |
| `check:phantom-deps` | 0 | `Every in-scope import is declared by the package that publishes it.` |
| `check:element-data-source-declaration` | 0 | `OK — 13 gate-consuming file(s) checked` |
| `check-changeset-presence` | 0 | `2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` |
| `check-changeset-overwrite` | 0 | `No pre-existing changeset was modified or deleted.` |
| `check:sdui-registration-pins` | 2 | **NOT MEASURED**, not a pass — `No console build to weigh at apps/console/dist/assets`. Declared narrowing: this diff adds no `ComponentRegistry.register` call and moves no registration array; left to CI, which builds the console. |

Repo-wide `pnpm lint` was not run locally — a declared narrowing, left to CI, which runs the farm regardless.

---

### DIFF
--- a/.changeset/wild-charts-explain-themselves.md
+++ b/.changeset/wild-charts-explain-themselves.md
@@ -1,0 +1,19 @@
+---
+'@object-ui/plugin-charts': patch
+---
+
+`ObjectChart` now renders a self-describing empty state when its query succeeds
+and returns no rows, instead of falling through to a bare chart frame.
+
+The frame was measured in a browser rather than assumed: recharts derives its
+ticks from the data, so with an empty result the bar and line families emit two
+hairline axis rules and no `text` nodes at all, and pie/donut emit nothing —
+there are no labelled axes to tell the reader what would have been plotted.
+Beside the component's own red "Failed to load chart data" box, a blank tile
+gives the reader nothing to distinguish a young chart from a broken one.
+
+The copy is the one `plugin-dashboard` already shows on the dataset-bound path
+("No data yet" / the load succeeded / the source name), so the same chart over
+the same empty result no longer reads two different ways depending on which
+widget drew it. Charts with inline authored data are unchanged — they ran no
+query to report on.
--- a/packages/app-shell/src/__tests__/widget-dom-leak-sweep.test.tsx
+++ b/packages/app-shell/src/__tests__/widget-dom-leak-sweep.test.tsx
@@ -431,6 +431,14 @@ const AUTHORED_PROPS = {
  * #4428 shipped a six-key first pass because a schema-only measurement cannot
  * see it. Answering with empty results keeps every data-bound target on its
  * real render path rather than an error state.
+ *
+ * ⚠️ That last sentence has ONE measured exception, and it is not an error
+ * state: since objectui#7130 an empty result is a real render path for
+ * `ObjectChart` that is not its CHART markup — it renders a self-describing
+ * empty state. The two object-bound chart targets therefore author their own
+ * rows; see {@link OBJECT_CHART_EXTRAS}. Read that before widening this
+ * adapter: handing rows to all ~200 targets would move many of them off the
+ * branch they are pinned on.
  */
 const FAKE_ADAPTER = {
   find: async () => [],
@@ -473,11 +481,36 @@ const CHART_DATA = [
   { name: 'Feb', sales: 300, revenue: 139, value: 300 },
 ];
 const CHART_SERIES = [{ dataKey: 'sales' }, { dataKey: 'revenue' }];
+/**
+ * The two OBJECT-BOUND chart targets (`plugin-charts:object-chart`,
+ * `view:chart`). They stay object-bound — `objectName` is what makes them a
+ * different registry path from the six inline chart targets above, which reach
+ * `ObjectChart` through `ObjectChartBlock` and its `ElementDataSourceGate`.
+ *
+ * `data` / `series` are authored on TOP of that binding (`data` is a declared
+ * registry input on this component: "Optional static data") because
+ * {@link FAKE_ADAPTER} answers every query with no rows, and since objectui#7130
+ * `ObjectChart` renders a self-describing empty state on an empty result
+ * instead of an empty chart frame. Without rows these two targets stop at that
+ * empty state, `[data-slot="chart"]` never matches, and the readiness guard
+ * below refuses to scan — correctly, because an empty state is not the chart
+ * markup this sweep exists to scan.
+ *
+ * This is objectui#5630's lesson in a third dress: a data-bound target whose
+ * clean reading would otherwise cover only its empty-state placeholder. That
+ * card answered it with a populated host where a host was needed, and with a
+ * plain `schemaExtras` change where the populated branch was reachable from
+ * pure schema (`element:definition-list`'s `items`). This is the latter case,
+ * and it also makes these two scan STRICTLY MORE markup than before: the
+ * pre-#7130 reading swept a chart frame with no marks in it.
+ */
 const OBJECT_CHART_EXTRAS = {
   objectName: 'accounts',
   chartType: 'bar',
   categoryField: 'name',
   valueField: 'amount',
+  data: CHART_DATA,
+  series: CHART_SERIES,
 };
 const CALENDAR_OBJECT_EXTRAS = {
   objectName: 'accounts',
--- a/packages/plugin-charts/src/ObjectChart.emptyResult.test.tsx
+++ b/packages/plugin-charts/src/ObjectChart.emptyResult.test.tsx
@@ -1,0 +1,122 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7130 — `ObjectChart` over an empty result renders a self-describing
+ * empty state instead of a bare frame.
+ *
+ * ## What the bare frame was, measured
+ *
+ * Before this branch the component fell through to `ChartRenderer` with
+ * `data: []`. Rendered in a real browser at 220c18d05, recharts emitted an SVG
+ * with TWO hairline axis rules and ZERO `<text>` nodes for bar/line, and
+ * nothing at all for pie — ticks are derived from the data, so an empty domain
+ * labels nothing. The filing hypothesis ("a chart frame with axes is arguably
+ * self-describing") is false: there are no labels on an empty chart.
+ *
+ * ## Why the assertions are shaped this way
+ *
+ * The maintainer's bar (hotcrm#1212) is *distinguishable from a load failure at
+ * a glance*, so the pin is not "an empty state exists" — it is that the empty
+ * state and the failure state are DIFFERENT, checked on the ARIA roles that
+ * carry that difference (`status` vs `alert`), plus the three directions this
+ * branch could be wrong in: firing over real rows, firing before the fetch
+ * resolves, and swallowing an error. Each is a separate `it` so an ablation
+ * reports which arm moved.
+ */
+import React from 'react';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { render, screen, cleanup, waitFor } from '@testing-library/react';
+
+vi.mock('recharts', async () => {
+  const actual = await vi.importActual<any>('recharts');
+  return {
+    ...actual,
+    ResponsiveContainer: ({ children }: any) =>
+      React.cloneElement(children, { width: 480, height: 320 }),
+  };
+});
+
+import { ObjectChart } from './ObjectChart';
+
+beforeEach(() => {
+  vi.stubGlobal('fetch', vi.fn(async () => ({ ok: true, json: async () => ({}) })));
+});
+afterEach(() => {
+  vi.unstubAllGlobals();
+  cleanup();
+});
+
+const schema = {
+  type: 'object-chart',
+  chartType: 'bar' as const,
+  objectName: 'crm_opportunity',
+  xAxisKey: 'stage',
+  series: [{ dataKey: 'amount', label: 'Amount' }],
+  isAnimationActive: false,
+};
+
+const renderWith = (dataSource: any, overrides: Record<string, unknown> = {}) =>
+  render(<ObjectChart schema={{ ...schema, ...overrides }} dataSource={dataSource} />);
+
+describe('ObjectChart — empty result (objectui#7130)', () => {
+  it('renders a self-describing empty state when the query returns no rows', async () => {
+    renderWith({ find: vi.fn().mockResolvedValue([]) });
+
+    const box = await screen.findByTestId('chart-empty-state');
+    // The copy states that the load SUCCEEDED — the fact a blank tile cannot
+    // give the reader — and promises no recovery (no "loading", no "retry").
+    expect(box).toHaveTextContent('No data yet');
+    expect(box).toHaveTextContent('loaded successfully');
+    expect(box.textContent).not.toMatch(/try again|retry|loading/i);
+    // Names WHAT is empty, so the tile is self-describing without authored copy.
+    expect(screen.getByTestId('chart-empty-source')).toHaveTextContent('crm_opportunity');
+  });
+
+  it('marks the empty state `status`, distinct from the failure box `alert`', async () => {
+    // The whole point of the card: the reader must be able to tell an empty
+    // result from a load failure. Asserted on the two roles rather than on
+    // copy, because that is the machine-readable half of the distinction.
+    const { unmount } = renderWith({ find: vi.fn().mockResolvedValue([]) });
+    expect(await screen.findByTestId('chart-empty-state')).toHaveAttribute('role', 'status');
+    unmount();
+
+    renderWith({ find: vi.fn().mockRejectedValue(new Error('Network request failed')) });
+    const failure = await screen.findByTestId('chart-error');
+    expect(failure).toHaveAttribute('role', 'alert');
+    // …and the empty branch must not have swallowed the error.
+    expect(screen.queryByTestId('chart-empty-state')).toBeNull();
+  });
+
+  it('does NOT fire over a populated result', async () => {
+    renderWith({
+      find: vi.fn().mockResolvedValue([
+        { stage: 'Qualify', amount: 12 },
+        { stage: 'Won', amount: 9 },
+      ]),
+    });
+    // Wait for the fetch to settle before asserting the absence, so this cannot
+    // pass merely by running before the data arrives.
+    await waitFor(() => expect(screen.queryByTestId('chart-loading')).toBeNull());
+    expect(screen.queryByTestId('chart-empty-state')).toBeNull();
+  });
+
+  it('does NOT flash before the fetch resolves — loading still wins', async () => {
+    renderWith({ find: vi.fn(() => new Promise(() => {})) });
+    expect(await screen.findByTestId('chart-loading')).toBeTruthy();
+    expect(screen.queryByTestId('chart-empty-state')).toBeNull();
+  });
+
+  it('leaves an inline-data chart alone — it ran no query to report on', async () => {
+    // `data: []` is an authoring choice, not an empty query result, so the copy
+    // ("its query returned no records") would be false of it.
+    renderWith({ find: vi.fn().mockResolvedValue([]) }, { data: [] });
+    await waitFor(() => expect(screen.queryByTestId('chart-loading')).toBeNull());
+    expect(screen.queryByTestId('chart-empty-state')).toBeNull();
+  });
+});
--- a/packages/plugin-charts/src/ObjectChart.tsx
+++ b/packages/plugin-charts/src/ObjectChart.tsx
@@ -3,8 +3,8 @@ import React, { useState, useEffect, useContext, useCallback, useMemo } from 're
 import { useDataScope, SchemaRendererContext, SchemaRenderer, useDrillNavigation, useFilterScope, ElementDataSourceGate, type ElementDataSourceMapping } from '@object-ui/react';
 import { ChartRenderer } from './ChartRenderer';
 import { ComponentRegistry, humanizeLabel, extractRecords, computeDrillFilter, isDrillEnabled, resolveDrillTitle, resolveFilterPlaceholders, resolveContextTokens, shiftFilterByCompareTo, compareToTrendLabelKey, buildChartSeries, buildOptionColorMap, deriveDimensionLabelMaps, dimensionOptionTranslator, loadDimensionFieldMeta, relabelDimensions, localizeFieldOptions, elementDataSourceBlock, type DimensionFieldMeta, type CompareToConfig, type DrillEvent, type ChartResultField, type ChartSegmentClickEvent } from '@object-ui/core';
-import { Sheet, SheetContent, SheetHeader, SheetTitle, Dialog, DialogContent, DialogHeader, DialogTitle, RefreshIndicator, Button, ChartSkeleton } from '@object-ui/components';
-import { AlertCircle, ArrowUpRight } from 'lucide-react';
+import { Sheet, SheetContent, SheetHeader, SheetTitle, Dialog, DialogContent, DialogHeader, DialogTitle, RefreshIndicator, Button, ChartSkeleton, DataEmptyState } from '@object-ui/components';
+import { AlertCircle, ArrowUpRight, Inbox } from 'lucide-react';
 import { useSafeFieldLabel, useSafeTranslate } from '@object-ui/i18n';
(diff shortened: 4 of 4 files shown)


os-warren:

⛔ Landing not armed — 5 checks `in_progress` (Type Check, 4 shards), 25 completed, 0 failure. I arm on green.

### ⭐⭐ You falsified the filer's own hypothesis, which is exactly what the card was for

This card was dispatched **measurement-first**, with "close it measured-and-declined" named as an accepted win — because the argument for declining was specific and plausible: *a chart is not a table; an empty chart still draws labelled axes, which may already tell the reader what would have been plotted.*

You rendered it in Chromium and measured: **zero text nodes.** Bar and line — 2 hairline rules, 0 paths, **not one character**. Pie/donut — completely blank. Populated control **in the same render**: 8 text nodes, 3 paths, 5 lines.

⇒ recharts derives ticks from the data, so an empty domain labels nothing. **The entire case for declining rested on labels that do not exist.** That is the difference between an opinion and a reading, and it is why this card was worth a slot even though it might have closed with no change.

⭐ And the sharpest observation: the empty tile is **less informative than the failure it must be distinguished from**. The chart-error box at least says *"Failed to load chart data"*; the empty one says nothing. It fails both halves of the hotcrm#1212 bar simultaneously.

### ⭐⭐ My KPI-precedent assumption was wrong, and the refutation reframes the card

I asked whether charts fall in the metric/KPI family that #7063 deliberately left alone. You refuted it **with landed code rather than argument** — and I verified it on `main`:

- `METRIC_TYPES` = `metric | kpi | gauge | solid-gauge | bullet`. Chart families are not in it.
- `DatasetWidget.tsx:726-728`: *"A metric (single value) over an empty dataset is 0, not an empty state… **Charts and tables keep the empty state** (there is genuinely nothing to plot)."*

⇒ **The consequence changes what this card is.** A *dataset-bound* chart has rendered the self-describing empty state since #7124; the *object-bound* chart did not. **Same family, same empty result, two different answers depending on which widget drew it.** So this is not a new judgement at all — it is the surface hotcrm#1212's ruling did not reach. That is a materially stronger justification than the card carried, and it came from measurement.

### ⭐ The package boundary: four options, and the one taken adds nothing

I told you the boundary was a real design question to establish with evidence, and that "promote to a shared package" must stop and report to #7132. You measured all four and took **(d)**: consume `DataEmptyState` from `@object-ui/components` — already shared, already consumed by `plugin-list` / `plugin-kanban` / `plugin-detail` **and by `WidgetEmptyState` itself**, and already a declared dependency of `plugin-charts`.

Verified independently: `DataEmptyState` present in `components/src/custom/view-states.tsx` (8 hits); `plugin-charts` declares `@object-ui/components` in its manifest (control: `react`, 2 hits). ⇒ **No package promoted, no dependency edge added, no new i18n key, no new spec key. #7132 untouched.**

⭐ And option (a) was rejected on **ownership, not mechanics** — you noted plainly that no cycle would have been created and that six plugin-to-plugin edges already exist. Rejecting something for the true reason rather than the convenient one is what makes the rejection reusable.

⭐ Reusing the three keys #7124 landed in all ten packs is the right call: this becomes a **fifth call site that collapses with the other four** when #7132 converges the defaults, instead of a fifth thing to converge.

### The pins assert the ways the branch could be wrong

Five, and three of them are negative: it must **not** fire over a populated result, must **not** flash before the fetch resolves, and must **not** fire for an inline-data chart. ⭐ Those are the three ways a well-meaning empty branch actually breaks things, and they are worth more than a second copy of the positive case.

Asserting distinguishability on the **ARIA roles** (`status` on the empty box vs `alert` on chart-error) rather than on copy is right — that is the machine-readable half of the maintainer's bar, and it survives rewording.

Ablation predicted the direction before running, proved the mutation on disk by marker counts **and** blob hash, reported `Tests 2 failed | 3 passed (5)` **with the passing count**, and proved the restore by state including `git diff --cached`. Every discipline this lane has accumulated tonight, applied without being re-asked.

### #7140 — filed at the right boundary

The sankey all-zero arm returns a bare `div`: `AdvancedChartImpl` filters to strictly positive values, so `links` is empty, and both existing refusals are gated `rows.length > 0` so neither ever fires. ⭐ **Not reachable by this PR's branch** (`finalData.length > 0` on that path), and correctly identified as *the last silent blank on the chart surface after this lands*. Different defect class, different file, filed not fixed.

I will triage it.

---

os-warren:
## ⛔ NOT ARMED — CI is red, and it is this PR's

`Test (shard 3/4)` concluded **failure** at `04:48:09Z`. 29 of 30 checks completed with 0 other failures. ⛔ Landing is withheld; my ACCEPT above stands on the substance and does **not** carry to a red head.

Sent back to the implementer with the diagnosis. ⚠️ Recording it here so the thread shows live state rather than an ACCEPT sitting above an unexplained red.

### The failure, and why it is real

From the job log — `packages/app-shell/src/__tests__/widget-dom-leak-sweep.test.tsx:1190`:

```
Error: view:chart: readiness selector `[data-slot="chart"]` never matched.
The widget did not reach its real markup, so scanning it would measure nothing.
```

…and the body it dumped is **this PR's new empty state**: `data-slot="data-empty-state"`, `data-testid="chart-empty-state"`, *"No data yet"* / *"This widget loaded successfully and its query returned no records yet."* / *"Source: accounts"*.

⇒ The sweep mounts `view:chart` with a fixture whose query returns no rows. Before this PR that still produced chart markup, so the readiness selector matched and the sweep scanned it. Now the empty branch fires first, the selector never matches, and the sweep **refuses to scan**.

⭐ **That refusal is the sweep's own anti-vacuity guard working exactly as designed** — it is declining to report a green over markup that never rendered. It caught a real behaviour change on the first run. ⛔ Which is precisely why the fix must not be to relax it: a guard that just proved it can detect this must not be taught to accept the state that tripped it.

### The gap that let it reach CI

The report ran `packages/plugin-charts`, then `plugin-dashboard` / `plugin-view` / `apps/console` as "downstream consumers". **`packages/app-shell` was not in that set**, and that is where the sweep lives.

⚠️ Worth noting for the seat as much as for the implementer: #7063's implementer *did* include `widget-dom-leak-sweep.test.tsx` in their union, for exactly this reason — a widget-markup change reaches it. The consumer set for "I changed what a widget renders" is wider than the packages that import it.

⛔ **Not a flake, so no re-run.** It is deterministic, its message names the cause, and the cause is this diff.

### Instructions sent

Give the sweep's `view:chart` fixture a non-empty result so the chart reaches real markup — preserving both the sweep's purpose and this PR's behaviour. ⛔ Do not relax the readiness selector, ⛔ do not add the empty state to what counts as "ready", and ⚠️ stop-and-report if measurement shows the fixture cannot supply data. Plus a sweep for other app-shell tests mounting a chart over an empty result, since this failure is one instance of a class.

I re-arm on green.

---

os-warren:
## CI fix pushed as `d3e7373f1` — the sweep's readiness guard caught a real behaviour change

`Test (shard 3/4)` was red on `packages/app-shell/src/__tests__/widget-dom-leak-sweep.test.tsx`, and it was mine. Not a flake, and not a defect in the branch's behaviour.

### The fixture mechanism, measured

The sweep mounts every target against one shared adapter:

```
const FAKE_ADAPTER = {
  find: async () => [],
  findOne: async () => null,
  aggregate: async () => [],
  count: async () => 0,
  getObject: async () => null,
};
```

whose docstring states the invariant it relies on: *"Answering with empty results keeps every data-bound target on its real render path rather than an error state."* The two object-bound chart targets take `OBJECT_CHART_EXTRAS` = `objectName: 'accounts'`, `chartType: 'bar'`, `categoryField`, `valueField` — **no inline `data`**, so they are query-backed and that adapter hands them nothing.

That invariant was true for charts until this branch. An empty result now renders the self-describing empty state rather than chart markup, so `[data-slot="chart"]` never matches and the readiness guard refuses to scan. The guard is right: an empty state is not the chart markup this sweep exists to scan, and it was telling me it would have measured nothing.

### Reproduced locally, and it is TWO targets, not one

CI named `view:chart`; both targets share `OBJECT_CHART_EXTRAS`, so both fail:

```
× plugin-charts:object-chart 10024ms
× view:chart 10007ms
Tests  2 failed | 200 passed (202)
```

### The fix is on the fixture, and the guard is untouched

The two targets now author `data` / `series` while staying object-bound — the same two keys the six inline chart targets beside them already use, and `data` is a declared registry input on this component ("Optional static data"). This is objectui#5630's own lesson in a third dress: that card fixed data-bound targets whose clean reading covered only their empty-state placeholder, using a populated host where one was needed and a plain `schemaExtras` change where the populated branch was reachable from pure schema (`element:definition-list`'s `items`). This is the latter case. I also corrected the `FAKE_ADAPTER` docstring, whose universal claim this branch made false.

⚠️ Worth stating: these two targets now scan **strictly more** markup than before. The pre-existing reading swept a chart frame with no marks in it.

**Proof the guard was not weakened** — the whole diff to that file is `33 insertions(+), 0 deletions(-)`, and the only non-comment lines added are:

```
+  data: CHART_DATA,
+  series: CHART_SERIES,
```

No `ready:` selector changed, nothing was removed, and the empty state was **not** added to what counts as ready.

### Re-run at `d3e7373f1`

| run | exit | verdict |
|---|---|---|
| `vitest run packages/app-shell/src/__tests__/widget-dom-leak-sweep.test.tsx` | 0 | `Test Files 1 passed (1)` · `Tests 202 passed (202)` — same total as the red run's 2+200, so no test vanished |
| `vitest run packages/app-shell` (whole package) | 0 | `Test Files 591 passed (591)` · `Tests 5765 passed | 1 skipped (5766)` |
| `vitest run packages/plugin-charts packages/types packages/react packages/core packages/plugin-timeline packages/plugin-map packages/fields packages/data-objectstack` | 0 | `Test Files 508 passed (508)` · `Tests 7389 passed (7389)` |
| `app-shell type-check` | 0 | green **after** building the dependency closure; the first run's exit 2 was unbuilt-closure `TS2307`, i.e. NOT MEASURED, not a red. `--listFiles` on `tsconfig.test.json` confirms the edited sweep file is in the checked set (1 of 4507) |
| `eslint` (plain form) on the edited file | 0 | clean |
| `check:control-bytes` | 0 | `OK (scanned 5909 tracked text file(s))` |
| `check-changeset-presence` | 0 | `3 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)` — the app-shell edit is test-only, so no app-shell bump was added |

### The gap that let it through, named

My consumer union was `plugin-charts` + `plugin-dashboard` / `plugin-view` / `apps/console`. `packages/app-shell` was not in it, and that is where this sweep lives. I re-derived the union by measurement instead of by guess: 12 packages contain a test naming an object-bound chart type, and the eight I had never run are all green above. Within app-shell the sweep is the only test that mounts a chart against an empty-returning adapter.

---

os-warren:
## ✅ Green after the fix — reviewed and armed

Undrafted and auto-merge (squash) enabled at head `d3e7373f1`.

### CI, read in full and read twice

All **30** check runs in one call at `perPage=100`. **27 success, 3 skipped, 0 failed.** The check that was red — `Test (shard 3/4)`, `widget-dom-leak-sweep.test.tsx` — is **success**, completed 05:40:50Z. All four shards, Type Check and Lint green.

Two readings I want on the record because each nearly produced a wrong verdict:

**I did not arm on the first pass.** At 05:32Z six checks were still `in_progress` — all four shards, Type Check, Lint — against 21 success and 0 red. That is *not* green. `in_progress` is the absence of a reading, and the still-pending one was precisely the shard that had failed.

**`mergeable_state` read `unstable`, then `clean`.** Immediately after undrafting, this PR reported `unstable`. Check runs alone would not have explained it, so I read the combined commit status too — `state: success`, one Vercel context — and then re-read the PR: `clean`. The `unstable` was a **stale derived field captured seconds after a state-changing write**, not a signal. Standing note: mergeability read immediately after undraft/push is stale; re-read before believing it, and never treat a derived field as a signal source when the underlying instruments are readable.

### The fence held — verified from the diff, not from the report

I fenced this round with ⛔ *do not relax the readiness selector; give the fixture a non-empty result.* Checked independently against the file stats rather than taking the report's word:

- `widget-dom-leak-sweep.test.tsx` is **+33 / −0**. **Zero deletions** — nothing was relaxed, removed, or loosened.
- The only non-comment lines added are `data: CHART_DATA,` and `series: CHART_SERIES,`.
- The single `[data-slot="chart"]` occurrence in the diff sits **inside a new comment**, not in a `ready:` value.

And the direction is right: because the production branch is gated on `!schema.data`, these targets now reach real chart markup. The sweep scans **strictly more** than before — the pre-change reading swept a chart frame with no marks in it.

### The failure was real, and it was two targets

The sweep mounts every target against one shared adapter whose `find` returns `[]`, with a docstring asserting that empty results keep every data-bound target on its real render path. This PR's empty branch made that invariant false for charts — so the guard was reporting a genuine consequence of the change, not noise. Relaxing it would have destroyed a working control.

CI named only `view:chart`. It was **two**: `plugin-charts:object-chart` and `view:chart` share `OBJECT_CHART_EXTRAS`, and reproducing locally gave `2 failed | 200 passed (202)`, with the post-fix run at `202 passed (202)` — same total, so no test vanished into the green. Asserting the total across a red/green pair is what rules out a suite that quietly stopped collecting.

### The gap the dev named themselves — and why it matters beyond this PR

They chose the blast radius from the package's **declared consumers**. For a change to an *empty-result branch*, the correct criterion is **which test fixtures answer queries with no rows** — a different and wider set. Twelve packages contain a test naming an object-bound chart type; the original union covered four. The eight uncovered ones were then run and are green.

Naming that rather than letting the passing re-run stand in for it is the valuable part. It is the same shape as this seat's standing lesson — *the consumer set for "I changed what a widget renders" is wider than the packages that import it* — arriving from a third direction. It is now carried into objectui#7132's brief, since that card will meet this same sweep.

### Also worth noting

The filing hypothesis was **disproved by measurement**: an empty chart renders **zero** `<text>` nodes (ticks derive from data), against a populated control emitting 8. Measured-and-declined was a genuinely available outcome here; the measurement is what ruled it out, not a preference.

The package boundary was decided on **ownership, not mechanics** — `WidgetEmptyState` would have needed promoting to public API for a foreign plugin, so `DataEmptyState` from the already-shared package was taken instead. No package promoted, no dependency edge added, no new i18n key. objectui#7132 stays free to converge all five call sites.

`check:sdui-registration-pins` returned `PREREQUISITE NOT MET` locally — correctly reported as NOT MEASURED rather than a pass. Same unconfirmed-by-me status as on PR objectui#7141: the diff adds no registration, so the gate has no subject in it regardless.

Landing will be verified by content with a live control once `main` advances — a queue sha is not a landing.

---

PR #7128, opened 2026-09-01
fix(fields): FieldEditWidget delivers the declared NON-DOM host-plumbing block

os-warren:
Fixes #7008

`FieldEditWidget` now delivers the **NON-DOM** half of the contract it declares — the other half of objectui#6909 / #7009.

## Re-derived the declared block, and the seven-key list was wrong: it is **nine**

The dispatch and the card both named seven undelivered keys. Re-derived from `FieldWidgetComponentProps`' own blocks on `origin/main` at `71d83a6b1`, the set of DECLARED keys that neither `toDomProps` nor the factory itself delivered is **nine**:

| block | key | delivered before |
|---|---|---|
| controlled-input | `error` | no |
| controlled-input | `onUploadingChange` | no |
| Host plumbing | `dataSource` | no |
| Host plumbing | `dependentValues` | no |
| Host plumbing | `dependsOn` | no |
| Host plumbing | `dependsOnLabels` | no |
| Host plumbing | `emptyHint` | no |
| Host plumbing | `onSelectRecord` | **no — not in the card's list** |
| Host plumbing | `onCreateNew` | **no — not in the card's list** |
| Host plumbing | `compact` | yes, factory-owned |

The card's own TITLE names all nine; its re-measurement comment narrowed to seven. The two extra keys are forwarded here, on measurement rather than symmetry:

- `LookupField` reads `props.onSelectRecord` and **has no metadata fallback at all** — the prop is its ONLY carrier.
- `LookupField` reads `props.onCreateNew ?? lookupField?.onCreateNew` — the prop is the PREFERRED carrier.

So the card body's speculation that "field metadata may be the intended single carrier" for those two is falsified: withholding them leaves a hole the widget explicitly expects a host to fill. `compact` stays factory-owned (derived from the resolved field type) and is named in the exclusion list so the compiler treats it as a decision, not an omission.

## The card understated the live victim: there are **two** hosts, and one was already wired

The card and the dispatch both state that none of the three in-repo hosts passes any host-plumbing key. That is **false on current `main`**:

`packages/plugin-detail/src/InlineFieldInput.tsx:486` passes `error={error}` into this factory, and has since `03380aa14` (PR #7109, `git log origin/main`). The producer was already wired and the factory dropped it on the floor — so an inline-edit control that had failed server validation never reported `aria-invalid`. That is a live, already-shipping defect, not a latent one.

[...]

PR #7141, opened 2026-09-01
fix(fields): the last five inline widgets read the delivered `error` slot

### LINKED ISSUE #7126: finding(fields): five EDIT_WIDGETS widgets never read `error`, so inline `text` / `boolean` / `date` / `datetime` / `time` controls still report no `aria-invalid`
Found while implementing objectui#7008 (the factory now DELIVERS the declared `error` key). Filed rather than folded in, because #7008's ruling scopes that card to the factory destructure and its dispatch said this class should be **reported, not glossed**.

## What

`FieldEditWidget` resolves a field type to one of 27 distinct widget components (`EDIT_WIDGETS`). After objectui#7008 the factory forwards the declared `error` key to whichever one it renders. **Five of those 27 never read it**, so for their field types `aria-invalid` is still never set and delivering the key is inert:

| widget | inline field types it serves |
|---|---|
| `TextField` | `text` |
| `BooleanField` | `boolean`, `toggle` |
| `DateField` | `date` |
| `DateTimeField` | `datetime` |
| `TimeField` | `time` |

Each destructures `{ value, onChange, field, readonly, ...props }` and spreads `toDomProps(props)`; `error` is not in `DOM_PASS_THROUGH_KEYS`, so it reaches the component and is discarded there.

## Measured, with a control

Read from `origin/main` at `71d83a6b1` with `git grep` against an explicit ref (not the shared checkout):

- census of the 27 `EDIT_WIDGETS` components for a word-boundary `error` match: **21 read it, 6 return zero**;
- **CONTROL for the regex**: the same query returns 5 hits on `NumberField`, 11 on `LookupField`, so a zero is a reading and not a broken instrument;
- **CONTROL for the zeroes**: of the 6, `UserField` is a FALSE zero — it renders `LookupField` with `{...props}`, so it delivers `error` transitively and DOES mark. That leaves the five above. A naive census would have reported six and been wrong about one.

## Why it matters — there is a live host

`RequiredFieldsDialog` (`packages/plugin-kanban/src/RequiredFieldsDialog.tsx`) opens on a kanban drop and renders whatever field types the target column made required. Since #7008 it hands the control its computed required-validation state. A required **`text`** field — the most common type in any object — therefore still shows the red "Required" hint to a sighted user and tells a screen-reader user nothing. The same holds for the grid inline editor and the detail page's inline edit, whose `error` producer (`InlineFieldInput`, PR #7109) is already wired.

This is the same defect and the same remedy as objectui#6803, which closed it for `NumberField` with `aria-invalid={!!error || !!refusal}` one layer down, and objectui#4824 for the textarea dialog surface.

## Note for whoever takes it

- `SelectField` is the shape to copy: `aria-invalid={!!error}` placed AFTER the DOM spread so the widget's own computation wins, which makes a valid field say an explicit `"false"` rather than staying mute (the objectui#3222 discipline).
- `BooleanField` and the three date/time widgets each render a real single control, so there is no composite-target question to settle first.
- The message TEXT stays with the host — a widget reads `error` only to drive `aria-invalid`, per `FieldWidgetComponentProps`' own doc comment. Do not render it here.

Unassigned and ungraded on purpose.

os-warren:
Fixes #7126

The last five widgets in `EDIT_WIDGETS` now read the `error` slot that objectui#7008 delivers, so an inline `text` / `boolean` / `toggle` / `date` / `datetime` / `time` control that failed validation finally reports `aria-invalid`.

## Leg 0 — the population, re-derived on `origin/main` AFTER its blocker landed

The census in the card was taken before PR #7128 landed. Re-taken here at `f08bcd9af` (`fix(fields): FieldEditWidget delivers the declared NON-DOM block`, from `git log origin/main`), in a dedicated worktree, never the shared checkout.

Word-boundary `error` match over the **27 distinct components** in `EDIT_WIDGETS`:

- **21 read it**, **6 return zero** — unchanged from the pre-#7128 census.
- **CONTROL for the instrument**: `NumberField` returns 5, `LookupField` 11, so a zero is a reading and not a broken query.
- **CONTROL for the zeroes — each apparent zero was opened and read, not counted.** `UserField` is a **FALSE zero**: its whole body is `return LookupField {...(props as any)} field={fieldProp}`, so it delivers `error` transitively and has always marked. Independently corroborated in-repo by `data-table.tsx`, which already records "26 of the 27 components in `EDIT_WIDGETS` call `toDomProps` directly; `UserField` delegates its whole props object to `LookupField`, which does".

⇒ the population is **exactly the five the card names** — `TextField`, `BooleanField` (`boolean` + `toggle`), `DateField`, `DateTimeField`, `TimeField`. Nothing shrank, nothing grew.

## What changed

One idiom, five widgets — the objectui#3222 shape the other 21 already share (`SelectField`, `EmailField`, `NumberField`): destructure `error`, then write `aria-invalid={!!error}` **after** the DOM pass-through spread so the widget's own computation wins and a valid field says an explicit `"false"` rather than staying mute. Seven call sites, because two widgets have two render branches each.

**⛔ This buys the a11y MARKING only.** The objectui#3222 slot drives `aria-invalid` and renders no text; the visible message stays with the host (`FormMessage` in the form, the red hint in the kanban dialog). Nothing that was invisible becomes visible, and that is the defect being closed.

## The composite judgement, stated per widget

| widget | markable control? | what was marked | why |
|---|---|---|---|
| `TextField` | yes (x2) | the `input[type=text]` **and** the `textarea` of the `rows > 1` branch | two render branches, two real single controls; marking only one would leave authored `rows` configs silent |
| `BooleanField` | yes (x2) | the Radix `Checkbox` / `Switch`, i.e. `button[role=checkbox]` / `button[role=switch]` | **the one composite here.** The control sits inside a flex `div` with an `sr-only` label. The `div` is deliberately NOT the target: a wrapper mark satisfies a subtree query while telling a screen-reader user nothing (objectui#5223). `aria-invalid` is a global ARIA attribute, valid on both roles, and both Radix roots forward unknown props to the real button |
| `DateField` | yes | the `input[type=date]` | no composite question despite the name "picker": the browser picker is that same element's own UI, not a second element |
| `DateTimeField` | yes | the `input[type=datetime-local]` | as above |
| `TimeField` | yes | the `input[type=time]` | as above |

None of the five turned out to lack a markable control.

## The end-to-end case — the deliverable

`packages/plugin-kanban/src/__tests__/RequiredFieldsDialog.ariaInvalidText-7126.test.tsx`, following the shape of #7128's `RequiredFieldsDialog.ariaInvalid-7008.test.tsx` rather than inventing a harness:

1. a required **`text`** field reports `aria-invalid="false"` on a clean open and `"true"` after a submit attempt finds it empty — with the submit refusal and the visible `Required` hint asserted first, so a dialog that silently submitted could not pass;
2. **CONTROL, per-field**: with a `text` filled and a `select` empty, the select is `"true"` and the filled text is `"false"`, and there is exactly one `Required` hint — a dialog-wide flag would fail this;
3. the whole population through the same dialog: `text`, `boolean`, `toggle`, `date`, `datetime`, `time` in one dialog, each read off **its own** element selector, so a mark that landed on `BooleanField`'s wrapper could not satisfy it.

Plus a factory-level sweep, `packages/fields/src/__tests__/inline-widget-aria-invalid-7126.test.tsx`: 8 cases (6 types + the two branch variants) x 2 directions, each requiring the `aria-invalid="true"` carrier to be **focusable**, plus two controls — `select` (a widget that already read `error`, proving the harness measures a real delivery) and `user` (pinning the census's FALSE zero so nobody later "fixes" a widget that was never broken).

**Why the factory and not the form.** In the FORM these five were already announced correctly and always had been: `FormControl` is a Radix `Slot` whose `aria-invalid` reached each control through the props spread untouched, and `widget-aria-invalid-registry-e2e.test.tsx` sweeps that path with an **empty** `NOT_YET_DELIVERED` ledger. A form-based test would have been green before this change. `FieldEditWidget` renders no Slot, and it is the seam every non-form host composes — so that is the path measured.

## Ablation — the pins fail without the change

Direction predicted **before** the run and recorded in the script header. Mutation: delete all 7 `aria-invalid={!!error}` lines.

- **Mutation proven on disk**, not by an editor exit code: marker count 2/2/1/1/1 -> 0 for all five, and every blob hash moved off its `HEAD` blob (e.g. `TextField.tsx` `a9a17e42f` -> `b919aa64d`, `BooleanField.tsx` `c8325fff6` -> `d2adecc4c`, `TimeField.tsx` `239728047` -> `762385d77`). No rebuild leg is claimed and none is owed: `vitest.config.mts` aliases `@object-ui/fields` to `packages/fields/src`, so the mutated source is what both pin files resolve — there is no `dist` on this path.
- **PREDICTED**: `Tests 19 failed | 2 passed (21)`.
- **OBSERVED**: `Tests 19 failed | 2 passed (21)` — and the 2 that stayed green are exactly the two CONTROL cases (`select`, `user`), which do not depend on the five widgets. The passing count is asserted, not just the red: a suite that had collapsed to "no tests" would not have left 2 passers standing.
- **Restore proven by state**, with `git checkout HEAD -- ABSOLUTE_PATH` (never a bare `git checkout --`, which reads the index a path-scoped checkout also writes): all five blobs back to their `HEAD` values, marker counts back to 2/2/1/1/1, and `git diff HEAD`, `git diff --cached` and `git status --short` all empty.

## Measurements

All at `323f2c381`, the head of this branch.

| check | verdict |
|---|---|
| `pnpm exec vitest run packages/fields/ packages/plugin-kanban/ --maxWorkers=2` | `Test Files 151 passed (151)` / `Tests 2249 passed (2249)` — includes the form-path registry sweep and #7128's own dialog pin |
| `pnpm exec vitest run` (the 2 components-side negative-assertion files) | `Test Files 2 passed (2)` / `Tests 24 passed (24)` |
| `pnpm --filter @object-ui/fields --filter @object-ui/plugin-kanban run type-check` | exit 0, `Scope: 2 of 47 workspace projects`, both `Done`. Its `tsconfig.test.json` leg **provably covers the new tests** — `tsc --listFiles` lists both new files, so this is not the "typecheck excludes `*.test.tsx`" non-reading |
| `pnpm --filter @object-ui/fields --filter @object-ui/plugin-kanban run lint` | exit 0, `0 errors` (warnings are pre-existing repo-wide; no `--max-warnings` in the gate). Plain `eslint .` per package, no `--no-inline-config` |
| `pnpm check:control-bytes` | `check-control-bytes: OK (scanned 5914 tracked text file(s); skipped 85 binary)` — plus a self-scan of the 9 changed files with a probe control that hits |
| `pnpm check:vi-mock-specifiers` / `check:vi-mock-inherit` / `check:i18n-keys` | all exit 0 |
| `pnpm changeset:check` | `All workspace packages are in the changeset fixed group.` / `No changeset declares a major bump.` |
| `node scripts/check-changeset-presence.mjs` | `8 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)` |
| `node scripts/check-changeset-overwrite.mjs` | `No pre-existing changeset was modified or deleted.` |
| `pnpm check:sdui-registration-pins` | **NOT MEASURED**, not red — exit 2 with its own `PREREQUISITE NOT MET` text: "the registrations this gate pins are dropped by a WRONG ARRAY at BUNDLE time, so a run with nothing to read has measured nothing. Build the console first". This change registers nothing; CI builds the console and runs it |

`type-check` first came back with a wall of `TS2307: Cannot find module '@object-ui/components'` across files this branch never touched — an unbuilt closure, i.e. NOT MEASURED. `pnpm --workspace-concurrency=2 --filter '@object-ui/fields^...' --filter '@object-ui/plugin-kanban^...' build` (exit 0) fixed it, and the verdict above is from the re-run.

## PM assumptions, checked

- **"The five are exactly right post-#7128."** Confirmed, with the false-zero trap re-caught rather than re-introduced.
- **"Each of the five has a markable control."** Confirmed, per widget, in the table above.
- **"`EDIT_WIDGETS` is the right population boundary."** **Partially falsified in structure, not in outcome — reported, not silently widened.** `InlineFieldInput` (`@object-ui/plugin-detail`) renders four widgets that are NOT in `EDIT_WIDGETS` onto the same inline surface — `ImageField`, `AvatarField`, `SignatureField`, `FileField` — and already passes `error={error}` to each. Measured with the same instrument and the same `NumberField`-returns-5 control: `ImageField` 4, `AvatarField` 4, `FileField` 10 all read it and mark; `SignatureField` returns 0, and that zero is a **documented verdict** ("`aria-invalid` and `aria-required` are NOT here, and their absence is a verdict rather than an omission" — no keyboard path on the canvas), matching its `NOT_APPLICABLE` row in the registry sweep. ⇒ no gap outside the boundary, nothing to file, and this PR stays at five widgets.
- **"No test asserts these widgets do NOT mark."** No *assertion* does — the registry sweep's `NOT_YET_DELIVERED` ledger is empty, and it exercises the form path where these five always marked via the Slot. One *doc comment* did: #7128's dialog test explained its choice of `select` with "NOT every inline widget reads `error` — `TextField`, `BooleanField`, `DateField`, `DateTimeField` and `TimeField` do not". That sentence is now false, so it is **corrected in place with a note saying what changed and where the `text` case now lives** — not deleted, because it is the reason `select` was chosen.

## Fences honoured

`toDomProps.ts`, `toHostProps.ts` and `FieldEditWidget.tsx` are **untouched** (`git diff --stat` names none of them) — that is #7008's landed work and its three compile-time assertions partition the contract. No `content/docs/releases/` edit. Sibling lanes (#7121 `views/studio-design/`, #7130 `packages/plugin-charts`) were not entered; no collision seen.

---

### DIFF
--- a/.changeset/7126-inline-widgets-read-error.md
+++ b/.changeset/7126-inline-widgets-read-error.md
@@ -1,0 +1,41 @@
+---
+'@object-ui/fields': patch
+---
+
+The last five inline edit widgets read the delivered `error` slot, so a failed
+required `text` / `boolean` / `date` / `datetime` / `time` control finally
+reports `aria-invalid` (objectui#7126).
+
+objectui#7008 made `FieldEditWidget` DELIVER the declared `error` key to
+whichever widget it resolves. Of the 27 distinct components in `EDIT_WIDGETS`,
+21 read it; five did not — `TextField`, `BooleanField` (serving both `boolean`
+and `toggle`), `DateField`, `DateTimeField` and `TimeField` — so for their field
+types the delivery was inert and the attribute was still never set.
+
+`text` being in that set is what made this a live defect rather than tidiness.
+It is the most common field type in any object, so it is the likeliest thing a
+kanban column makes required: `RequiredFieldsDialog` computed the failure, drew
+the red "Required" hint, handed the state to the control, and the control said
+nothing to assistive tech. The grid's inline cell editor and the detail page's
+inline edit (`InlineFieldInput`) compose the same seam.
+
+Each of the five now computes `aria-invalid={!!error}` **after** its DOM
+pass-through spread — one existing idiom, the objectui#3222 discipline the other
+21 already share, so a valid field says an explicit `"false"` rather than staying
+mute. Two judgements worth stating:
+
+- **The FORM path was never broken and is unchanged.** `<FormControl>` is a
+  Radix `Slot` whose `aria-invalid` reached each control through the props
+  spread; the form also produces `error`, so the widget's own computation now
+  agrees with the value it replaces. The gap was every host WITHOUT that Slot.
+- **`BooleanField` is the one composite here, and the mark goes on the
+  control.** Its Radix `Checkbox` / `Switch` renders a real
+  `button[role=checkbox]` / `button[role=switch]`; the wrapping flex `div` is
+  deliberately not the target, because a wrapper mark satisfies a subtree query
+  while telling a screen-reader user nothing (objectui#5223). The three
+  date/time widgets each render one native input, so the browser's picker raises
+  no second-element question.
+
+This buys the MARKING only. The objectui#3222 slot drives `aria-invalid` and
+renders no text: the visible message stays with the host, and nothing that was
+invisible becomes visible.
--- a/packages/fields/src/__tests__/inline-widget-aria-invalid-7126.test.tsx
+++ b/packages/fields/src/__tests__/inline-widget-aria-invalid-7126.test.tsx
@@ -1,0 +1,193 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The five widgets that never read the delivered `error` now report
+ * `aria-invalid` on a FOCUSABLE control, through the real inline seam
+ * (objectui#7126).
+ *
+ * ## The defect
+ *
+ * objectui#7008 made `FieldEditWidget` DELIVER the declared `error` key
+ * (`toHostProps`, landed as f08bcd9af). Of the 27 distinct components in
+ * `EDIT_WIDGETS`, 21 read it; five did not — `TextField`, `BooleanField`
+ * (`boolean` + `toggle`), `DateField`, `DateTimeField`, `TimeField` — so for
+ * their field types the delivery was inert and `aria-invalid` was still never
+ * set. `text` being in that set is what made it live rather than tidy: it is
+ * the most common type in any object, and the kanban `RequiredFieldsDialog`
+ * renders whatever types the target column made required.
+ *
+ * ## Why the FACTORY and not the widgets directly
+ *
+ * In the FORM these five were already announced correctly and always had been:
+ * `<FormControl>` is a Radix `Slot`, its `aria-invalid` reached the control
+ * through each widget's props spread untouched, and
+ * `widget-aria-invalid-registry-e2e.test.tsx` sweeps exactly that path with an
+ * EMPTY `NOT_YET_DELIVERED` ledger. So a form-based test would have been green
+ * before this change and proves nothing about it.
+ *
+ * `FieldEditWidget` renders no Slot. It is the seam every NON-form host
+ * composes — the grid's inline cell editor, the detail page's inline edit
+ * (`InlineFieldInput`), the kanban required-fields dialog — and the only way
+ * the state reaches the control there is the declared `error` prop. That is
+ * the path that was broken, so that is the path measured here.
+ *
+ * ## What is NOT claimed
+ *
+ * The marking only. objectui#3222's slot drives `aria-invalid` and renders no
+ * text; the visible message stays with the host. Nothing here becomes visible
+ * that was not visible before.
+ */
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import React from 'react';
+
+import { FieldEditWidget } from '../FieldEditWidget';
+
+afterEach(() => cleanup());
+
+/**
+ * HTML's own focusability rules, as a selector — copied from
+ * `widget-aria-invalid-registry-e2e.test.tsx` on purpose, so both sweeps judge
+ * "the control a keyboard user can land on" by one definition.
+ *
+ * This is the objectui#5223 line: a mark on a non-focusable wrapper satisfies a
+ * subtree query while telling a screen-reader user nothing, and it is the
+ * cheapest way to make an assertion like the ones below go green without
+ * helping anyone.
+ */
+const FOCUSABLE = [
+  'a[href]',
+  'button:not([disabled])',
+  'input:not([disabled])',
+  'select:not([disabled])',
+  'textarea:not([disabled])',
+  '[tabindex]:not([tabindex="-1"])',
+  '[contenteditable="true"]',
+].join(',');
+
+function describeEl(el: Element): string {
+  const role = el.getAttribute('role');
+  const type = el.getAttribute('type');
+  return `${el.tagName.toLowerCase()}${type ? `[type=${type}]` : ''}${role ? `[role=${role}]` : ''}`;
+}
+
+function renderInline(field: Record<string, unknown>, error?: string) {
+  const { container } = render(
+    <FieldEditWidget
+      field={field as never}
+      value={undefined as never}
+      onChange={() => {}}
+      error={error}
+    />,
+  );
+  return container;
+}
+
+/**
+ * The population of objectui#7126, by the field TYPE each widget serves inline
+ * — six types, five widgets (`boolean` and `toggle` both resolve to
+ * `BooleanField`), plus the two branch variants that a type key alone does not
+ * reach: `TextField`'s textarea branch (`rows > 1`) and `BooleanField`'s
+ * checkbox branch (`widget: 'checkbox'`). Both are real authored configs, and
+ * each renders a DIFFERENT element, so a fix applied to only one branch of
+ * either widget still fails here.
+ */
+const CASES: ReadonlyArray<readonly [label: string, field: Record<string, unknown>]> = [
+  ['text', { name: 'f', type: 'text', label: 'F' }],
+  ['text (rows > 1 -> textarea branch)', { name: 'f', type: 'text', label: 'F', rows: 4 }],
+  ['boolean (switch branch)', { name: 'f', type: 'boolean', label: 'F' }],
+  ['boolean (widget: checkbox branch)', { name: 'f', type: 'boolean', label: 'F', widget: 'checkbox' }],
+  ['toggle', { name: 'f', type: 'toggle', label: 'F' }],
+  ['date', { name: 'f', type: 'date', label: 'F' }],
+  ['datetime', { name: 'f', type: 'datetime', label: 'F' }],
+  ['time', { name: 'f', type: 'time', label: 'F' }],
+];
+
+describe('inline field widgets announce a delivered `error` (objectui#7126)', () => {
+  it.each(CASES)(
+    '%s — carries aria-invalid="true" on a FOCUSABLE control when the host delivers `error`',
+    (_label, field) => {
+      const container = renderInline(field, 'Required');
+
+      const carriers = Array.from(container.querySelectorAll('[aria-invalid="true"]'));
+      expect(
+        carriers.map(describeEl),
+        'the host delivered `error` and nothing in the rendered widget says so — assistive tech is never told the field failed',
+      ).not.toEqual([]);
+
+      // THE WRAPPER-MARK HOLE (objectui#5223). `BooleanField` is the case this
+      // exists for: it renders its control inside a flex `div`, and marking
+      // that `div` would satisfy the query above while the switch the user
+      // actually operates announces nothing.
+      expect(
+        carriers.filter((el) => el.matches(FOCUSABLE)).map(describeEl),
+        `aria-invalid sits ONLY on non-focusable element(s) [${carriers.map(describeEl).join(', ')}] — that is a wrapper mark, not a control mark`,
+      ).not.toEqual([]);
+    },
+  );
+
+  it.each(CASES)(
+    '%s — says an explicit aria-invalid="false" when the host delivers no `error`',
+    (_label, field) => {
+      // The load-bearing half, and the reason this is a two-state reading
+      // rather than "the attribute exists": `!!undefined` must yield `"false"`,
+      // so a valid field SAYS it is valid instead of staying mute (the
+      // objectui#3222 discipline). Without this, an unconditional
+      // `aria-invalid="true"` would pass the case above.
+      const container = renderInline(field);
+
+      const control = container.querySelector(FOCUSABLE);
+      expect(control, 'no focusable control rendered at all').not.toBeNull();
+      expect(control).toHaveAttribute('aria-invalid', 'false');
+      expect(container.querySelector('[aria-invalid="true"]')).toBeNull();
+    },
+  );
+
+  it('CONTROL: a widget that ALREADY read `error` reports the same way through the same harness', () => {
+    // Without this, a green sweep above could not be distinguished from a
+    // harness that marks everything it renders. `select` -> `SelectField` was
+    // one of the 21 readers before this change (objectui#3306 / #7008's pin),
+    // so it must read `true`/`false` here for exactly the reasons the five now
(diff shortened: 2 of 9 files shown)


os-warren:

⛔ Landing not armed — 5 checks `in_progress` (4 shards, Type Check), 25 completed, 0 failure. I arm on green.

### ⭐ Every zero was OPENED AND READ, not counted

The census re-derived at `f08bcd9af` in a dedicated worktree — ⚠️ **never the shared checkout**, which is the trap that produced a false, card-shrinking reading on this card's own parent. Instrument controlled (`NumberField` 5, `LookupField` 11), and then **each of the six apparent zeroes was opened and read** rather than trusted as a count.

That is what caught **`UserField`** — the false zero the dispatch named. I verified it directly: `packages/fields/src/widgets/UserField.tsx:58` is `return <LookupField {...(props as any)} field={fieldProp} />;`, with its docblock stating it *"delegates to the shared LookupField"*. So it has always marked, transitively.

⭐ **And pinning it as a dedicated control case is the part that will still be paying off in a year** — it stops a future reader "fixing" a widget that was never broken, and it fails loudly if the delegation is ever flattened.

⚠️ One attribution I could **not** confirm: the corroborating quote is credited to `data-table.tsx`, and no such file exists under `packages/fields/src`. My probe for it returned zero — **and so did my control**, so I discarded that reading as void rather than reporting it as a refutation, and re-took it against the source directly. The substance holds; only the quote's file is unverified. ⛔ Not blocking, but do not carry that citation forward without re-locating it.

### ⭐ The composite was handled exactly as fenced, and verified rather than assumed

`BooleanField` was the one composite. You marked the **Radix Checkbox / Switch button** (`role="checkbox"` / `role="switch"`), ⛔ not the wrapper div — and gave the reason that matters: *a wrapper mark satisfies a subtree query while telling a screen-reader user nothing* (objectui#5223). Then you **read `checkbox.tsx` and `switch.tsx`** to confirm both Radix roots forward unknown props to the real button, instead of assuming a spread reaches it.

⭐ And `TextField` turned out to have **two** sites — the input and the `rows > 1` textarea branch — with the right reason: *"marking only one would leave authored `rows` configs silent."* Five widgets, **seven** call sites. The obvious count was wrong and the diff is right.

### The idiom is matched, not invented

Read `SelectField` / `EmailField` / `NumberField` / `RadioField` / `CheckboxesField` first; one idiom across all seven sites; `aria-invalid` written **after** the DOM pass-through spread so the widget's own computation wins; and `!!undefined` yielding an explicit `"false"` so every pin is a **two-state** reading. ⇒ An unconditional `true` could not have passed. That last detail is what makes the positive assertions worth anything.

### Ablation: predicted 19/2 in the script header before running, observed 19/2

The two survivors are exactly the two **control** cases (`select`, a pre-existing reader; `user`, the false zero) — neither depends on the five, so their survival is the predicted shape rather than luck. Passing count asserted, mutation proven by marker count **and** blob hash across five files, restore proven by state including `git diff --cached` and `git status --short`.

⭐ The focusable-carrier assertion on every positive case is the pin's real backbone: it forecloses exactly the wrapper-mark shortcut the `BooleanField` decision avoided, so the two halves of this PR check each other.

### ⭐ You corrected a doc comment your own PR made stale

#7128's `RequiredFieldsDialog.ariaInvalid-7008.test.tsx` header explained its choice of `select` with *"NOT every inline widget reads error — TextField, BooleanField, DateField, DateTimeField and TimeField do not."* True when written; **false as of this PR**.

Corrected **in place with a note**, ⛔ not deleted — because it is the stated reason `select` was chosen, and deleting it would remove the justification along with the staleness. ⇒ That is the seventh instance this session of *a comment certifying what the code no longer has*, and the first where the author noticed they were the cause and repaired it in the same change.

### My population boundary was porous — and you established there is no gap outside it

I assumed `EDIT_WIDGETS` was the right boundary. Measured: `InlineFieldInput` renders **four** widgets outside it onto the same inline surface (`ImageField`, `AvatarField`, `SignatureField`, `FileField`), already receiving `error`. Three read and mark it.

⭐ And `SignatureField`'s zero is **a documented verdict, not an omission** — its own header says *"aria-invalid and aria-required are NOT here, and their absence is a verdict rather than an omission"* (the canvas has no keyboard path), matching its `NOT_APPLICABLE` row in the registry sweep.

⇒ Boundary porous, **no gap**, nothing filed, nothing widened, PR stays at five. Checking outside the fence and reporting "clean" is worth as much as finding something there.

### Scope

`toDomProps` / `toHostProps` / `FieldEditWidget` untouched ✅. Changeset staged-then-run with a **non-empty population** quoted, explicitly distinguished from the `0 file(s) changed` empty-population verdict ✅. `check:sdui-registration-pins` reported **NOT MEASURED** with the gate's own prerequisite text, not as red ✅. The first `type-check` was an unbuilt closure, reported as NOT MEASURED and re-run after building ✅.

---

os-warren:
## ✅ Reviewed and armed — PM as reviewer-of-record

Undrafted and auto-merge enabled (squash) at head `323f2c381`.

### CI read in full, in one call

All **30** check runs fetched in a single request with `perPage=100`. This matters: earlier in this seat I mixed page sizes across two calls and silently skipped items 9–14, and nearly armed a PR off an incomplete reading. A partial check list looks exactly like a green one.

**27 success, 3 skipped, 0 failed, 0 pending.** The three skipped are `dependabot`, `Test (coverage)` and `Test (coverage shard ${{ matrix.shard }}/4)` — the unexpanded `${{ matrix.shard }}` in that last name is itself the evidence they were skipped *before* matrix expansion, i.e. conditional jobs that did not apply, not required checks that vanished.

Pre-check at arm time: `state: open`, `merged: false`, `mergeable_state: clean`, head sha unchanged from the sha the checks ran against.

### Clause ② — not engaged

This consumes the `error` slot that objectui#7008 / PR objectui#7128 already landed. It adds no export, changes no parser accept/reject behaviour, and touches neither `toDomProps.ts`, `toHostProps.ts` nor `FieldEditWidget.tsx`. It therefore does not need `CONTRACT_REVIEW_TIER` and is not held by the fable quota exhaustion.

### What earned the arm

Three things in this report are the reason it goes in without a round trip:

**The census was re-taken after its blocker landed, not reused.** The card's population was measured pre-#7128. Re-deriving it at `f08bcd9af` is the difference between a population and a memory of one.

**Every apparent zero was opened, not counted.** `UserField` returning 0 for a word-boundary `error` match is a **FALSE zero** — its body is a single delegating `return <LookupField {...(props as any)} … />`, so it marks transitively and always did. I verified that line directly myself before dispatch, and the dev caught it independently. A census that had trusted the six zeroes would have "fixed" a widget that was never broken, and shipped a change with no defect behind it. Pinning that false zero as a permanent CONTROL case in the sweep is the right durable answer.

**The ablation asserts the passing count, not just the red.** `19 failed | 2 passed (21)`, predicted before the run, with the 2 survivors being exactly the two control cases that do not depend on the five widgets. A suite that had collapsed to "no tests" would have shown red-free-of-passers and looked like a stronger result. Mutation proven by marker count **and** blob hash; restore proven by state (`git diff HEAD`, `git diff --cached`, `git status --short` all empty), not by an editor's exit code.

Also correct: `type-check` first returned a wall of `TS2307: Cannot find module '@object-ui/components'` and was reported as **NOT MEASURED** rather than red — an unbuilt dependency closure is the absence of a reading. Building the closure and re-running is the only thing that turns it into a verdict.

### The PM assumption that got falsified — good

I fenced the population at `EDIT_WIDGETS`. That boundary is **structurally wrong**: `InlineFieldInput` renders four widgets onto the same inline surface that are not in `EDIT_WIDGETS` at all. The dev measured them with the same instrument and the same live control rather than accepting my boundary, found `ImageField`/`AvatarField`/`FileField` already marking and `SignatureField`'s zero to be a *documented verdict* (no keyboard path on the canvas, matching its `NOT_APPLICABLE` registry row) — and then **did not widen the PR**. Reporting a falsified boundary that happens to have no gap behind it, instead of either hiding it or expanding scope on it, is exactly the handling I want.

Same for the stale doc comment in #7128's dialog test: corrected in place with a pointer to where the `text` case now lives, not deleted — it is the reason `select` was chosen and deleting it would have destroyed that reasoning.

### One sub-claim I am recording as unconfirmed, not confirmed

`pnpm check:sdui-registration-pins` came back `PREREQUISITE NOT MET` (exit 2) locally — correctly classified as NOT MEASURED rather than green. The report then says "CI builds the console and runs it." I do not see a check run under that name among the 30, so it may be a step inside `Build & E2E` or it may not run on PRs at all — **I have not confirmed it either way.**

Arming anyway, because that claim is not load-bearing: the independent argument stands on its own — this diff **registers nothing**, so a registration-pin gate has no subject in it regardless of whether CI executes the gate. Flagging it so the unconfirmed half is not later remembered as verified.

Will confirm the enqueue positively from the queue ref and post a landing note on objectui#7126 once `main` actually advances — a queue sha is not a landing, and even `main` landing on a queue sha is not proof the diff arrived, so the landing gets probed by content with a control.

---

PR #7145, opened 2026-09-01
fix(app-shell): scope the Interfaces block selection to its leaf

### LINKED ISSUE #7137: finding(app-shell): InterfacesPillar's block `selection` outlives a leaf change on every non-editable leaf — the clear sits inside the `isEditable` early-return
Found while implementing #7121 (studio-canvas leaves offering Design mode and a "click a block" rail). Swept up, **measured**, and deliberately **not** fixed there: the repair reaches every non-editable leaf, not just the studio-canvas leaves that card is fenced to.

## The mechanism

`InterfacesPillar`'s draft-load effect clears the canvas block selection — but the clear sits **after** the guard, so it never runs on the path that returns early:

```
React.useEffect(() => {
  if (!current || !isEditable) {
    setDraft({});
    setHasDraft(false);
    return;              // <-- returns BEFORE the clear below
  }
  let cancelled = false;
  setLoading(true);
  setError(null);
  setSelection(null);    // <-- only ever reached on the editable path
  ...
}, [client, current, isEditable, publishNonce]);
```

`packages/app-shell/src/views/studio-design/StudioDesignSurface.tsx`, the effect commented "Load the selected surface's draft (only for editable preview types)".

So walking from a leaf where a block is selected to any leaf with `isEditable === false` carries the selection across. The surviving `selection` describes a block on the **previous** leaf's canvas.

## Measured

With the designer registry populated, a stub preview for `dashboard` and a stub inspector registered for `object`, selecting a block on the dashboard leaf and then clicking the `object` leaf:

```
[B] selection live on dashboard leaf: true
[B] deselect affordance STILL present on object leaf: true
[D] inspector types now: ["object"]
[D] scoped inspector rendered on studio-canvas leaf: true
[D] inspector was handed type/name: object:showcase_task:block:blk_1
```

`blk_1` is a dashboard block. It does not exist on the records grid the object leaf renders. Production registers `ObjectFieldInspector` for `object` (`metadata-admin/inspectors/index.ts`), so the branch is reachable outside the harness.

## What #7121 already covers, and what it does not

#7121 gates the rail and the header's "clear selection" button on `StudioCanvas`, and orders the rail's studio-canvas branch ahead of the selection branch. That makes the two **rail** symptoms unreachable on studio-canvas leaves.

It does **not** clear the state. Still live afterwards:

- `hasInspectorTarget = Boolean((editNav && navSel) || selection)` feeds `nextCenterTab` (`centerTab.ts`). In the folded (narrow / chat-dock) layout, arriving at a leaf with a stale selection reads as "a target APPEARED" and auto-switches the center tab to **Properties** — which then states the canvas has no blocks. *(Read from the code path, not measured in a browser.)*
- The leak is not specific to studio-canvas leaves. It is every leaf where `isEditable` is false, which also includes leaves whose own type has no registered designer.

## Why it was not fixed in #7121

That card's ruling fences it to studio-canvas leaves — `registerStudioCanvasPreview` types, of which there is exactly one in-repo today (`object`). Moving `setSelection(null)` out of the guard changes behaviour on a strictly larger population, so it is a deliberate decision rather than a rider.

## Not scoped here

Which repair is right:

- hoist `setSelection(null)` above the guard so any leaf change clears it — simplest, and arguably what the effect already means to say;
- clear it in the leaf-change handler instead of the load effect, so the reset is not entangled with `isEditable` being in the dependency array;
- key the selection to the leaf (as `blockingReport` already is via `inspectorKey`) so it expires by construction rather than by an imperative clear. `inspectorKey` is the precedent in the same component.

## Related

#7121 (where this was found) · #6795 part C · `centerTab.ts` / ADR-0057 P3c (the folded-layout consumer of `hasInspectorTarget`)

os-warren:
Fixes #7137

The Interfaces pillar's block `selection` outlived a leaf change on every leaf
where `isEditable === false`, because the only clear sat *after* the draft-load
effect's early return:

```
React.useEffect(() => {
  if (!current || !isEditable) { setDraft({}); setHasDraft(false); return; }
  ...
  setSelection(null);   // never reached on the early-return path
}, [client, current, isEditable, publishNonce]);
```

`isEditable = !!Preview && !StudioCanvas` is a conjunction, so that early return
covers two disjoint populations, and the surviving selection then described a
block on the *previous* leaf's canvas.

## Premise re-derived on the current head

Verified in a fresh worktree at `origin/main` = `0d4c7890d` — the commit that
landed #7121, i.e. after the card was filed. The guard's shape and the stranded
clear are both still exactly as the card quotes them. Premise holds.

## Populations, measured

| population | in-repo count | live symptoms after #7121 |
| --- | --- | --- |
| (a) studio-canvas leaves | **1** (`object`, `studio-canvas-preview.tsx:97`) | folded-layout center tab only |
| (b) leaf whose own type has no designer | **0 under the shipped registration** | scoped inspector + clear button |

The pillar's leaves come from `resolveSurface`, which can produce exactly five
types (`page` / `object` / `dashboard` / `report` / `action`), and
`registerBuiltinPreviews()` registers a preview for all five. So population (b)
is **empty** when `register-builtins` has run. It is reachable only in a host
that registers a partial set — which is the state
`StudioDesignSurface.designerRegistryPartial.test.tsx` already treats as a
supported product fact, and which this PR's pin constructs.

On population (a) the rail and header symptoms are already unreachable, because
#7121 gates both on `StudioCanvas`. What that card did **not** gate is
`hasInspectorTarget`, which feeds `nextCenterTab` — so the folded center tab is
the one surviving observable there, and it is measured below.

## The repair, chosen by measurement

The selection is now stamped with the leaf it was made on and read back through
that key — the same shape `blockingReport` already uses against `inspectorKey`
in this component.

The card offered three candidates. Candidate 1, hoisting `setSelection(null)`
above the guard, was **implemented and measured** rather than reasoned about: it
closes the center-tab symptom, but the clear runs in an effect, i.e. one
committed render after the leaf change, so the foreign-block inspector still
mounts. Against the same pin:

- unfixed tree: ledger `[ 'page:home_page:block:blk_1', ...(2) ]` — 3 stale renders
- candidate 1 (hoist): ledger `[ 'page:home_page:block:blk_1' ]` — **1** stale render, `1 failed | 4 passed (5)`
- this PR (leaf-keyed): ledger `[]` — `5 passed (5)`

That single surviving render is what decided it. Keying also leaves no
imperative clear for a future guard to strand, which is how this defect arose.

Within a leaf nothing changes: the Design/Run round trip keeps its selection
(#5800's stated contract), and a same-leaf reload via `publishNonce` still
clears it — both pinned.

## Both symptoms are closed, and the card's NOT MEASURED sub-claim is now measured

The card flagged the `centerTab` symptom as *"read from the code path, not
measured in a browser."* It is measured here, in the folded layout with
`foldInspector`, reading the active tab from the DOM:

- control on the same instrument: picking a block **does** drive the tab to
  Properties, so the reading below is of a live auto-switch, not a static strip;
- unfixed: `expected 'Properties' to be 'Canvas'` — the stale target suppresses
  the return-to-Canvas edge;
- fixed: the tab returns to Canvas.

One refinement to the card's wording: the mechanism is not an auto-switch *to*
Properties on arrival. `hasInspectorTarget` never changes across the leaf walk,
so `nextCenterTab` sees no edge at all and the author is **stranded** on the
Properties tab they were already on. Same defect, same repair, more precise
cause.

## Not done here

- #7121's rail / Design-mode gating is untouched, and its discriminator remains
  `StudioCanvas`. It was never substituted with `isEditable`; its live fence
  `StudioDesignSurface.designerRegistryPartial.test.tsx` is green.
- `AutomationsPillar` in the same file was checked and does **not** share the
  defect — its guard is `if (!current) return`, with no `isEditable` term, so
  its clear runs on every leaf change.
- No public surface change: `InterfacesPillar` is a module export that
  `packages/app-shell/src/index.ts` does not re-export, and no exported type
  moved.

## Verification

All runs below are at `db22d18b0`, on a clean tree.

- `pnpm exec vitest run .../StudioDesignSurface.selectionLeafScope.test.tsx` —
  new pin. Before: `3 failed | 2 passed (5)`. After: `5 passed (5)`.
  The 2 that passed before are named and explained in the file: one is the
  over-fire fence (green on both sides by design), the other is the
  studio-canvas leaf, green because #7121 already blocks that path — the
  finding that made the center-tab test the real measurement for population (a).
- `pnpm exec vitest run packages/app-shell/src/views/studio-design/` —
  `43 passed (43)` files, `230 passed (230)` tests.
- `pnpm exec vitest run packages/app-shell/src/views/metadata-admin/` —
  `221 passed (221)` files, `2303 passed | 1 skipped (2304)`.
- `pnpm --filter '@object-ui/app-shell' type-check` — exit 0
  (`tsc --noEmit && tsc -p tsconfig.test.json`). Confirmed to actually cover the
  new file: `--listFiles` lists it, alongside a known-covered sibling as control.
- `pnpm --filter '@object-ui/app-shell' lint` — exit 0, `0 errors`
  (2853 pre-existing warnings package-wide, none in the changed files).
- `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`,
  `check:i18n-keys`, `check:shell-escape-residue` — all exit 0.
- `scripts/check-changeset-no-major.mjs` — `No changeset declares a major bump.`

The ablation restore was verified by state, not by the trap firing:
`git diff HEAD`, `git diff --cached` and `git status --short` all empty, and the
worktree blob back to HEAD's `f4c98092438320bba6469a3e886d21b2779cc2d8`.

Repo-wide `pnpm lint` and the full gate farm are left to CI, which runs them
once regardless.

---

### DIFF
--- a/.changeset/7137-selection-expires-with-leaf.md
+++ b/.changeset/7137-selection-expires-with-leaf.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Studio Interfaces pillar: a canvas block selection no longer outlives the leaf
+it was made on (objectui#7137).
+
+`InterfacesPillar`'s only clear of `selection` sat inside the draft-load effect,
+*after* its `if (!current || !isEditable) … return` guard, so it never ran on the
+early-return path. Since `isEditable = !!Preview && !StudioCanvas` is a
+conjunction, that is two populations of leaf: a studio-canvas leaf (`object`),
+and a leaf whose own type has no registered designer. Walking to either from a
+leaf with a block selected carried the selection across, still describing a block
+on the previous leaf's canvas.
+
+Two symptoms, both measured before and after:
+
+- the scoped inspector opened for a foreign block — recorded three renders as
+  `page:home_page:block:blk_1`, with `blk_1` a dashboard block — and the header
+  offered to clear a selection belonging to another leaf;
+- in the folded (chat-dock) layout, `hasInspectorTarget` stayed true across the
+  leaf change, so `nextCenterTab` saw no edge and stranded the author on the
+  Properties tab of a leaf with no properties to show.
+
+The selection is now stamped with its leaf and read back through that key — the
+same "expires by construction" shape `blockingReport` already uses against
+`inspectorKey` in this component. It goes null in the *same* render as the leaf
+change rather than one committed render later, and there is no imperative clear
+left for a future guard to strand. Within a leaf nothing changes: the Design/Run
+round trip still keeps its selection, and a same-leaf reload (`publishNonce`)
+still clears it.
+
+objectui#7121's gating of the rail and the Design/Run switch is untouched, and
+its discriminator remains `StudioCanvas` — not `isEditable`.
--- a/packages/app-shell/src/views/studio-design/StudioDesignSurface.selectionLeafScope.test.tsx
+++ b/packages/app-shell/src/views/studio-design/StudioDesignSurface.selectionLeafScope.test.tsx
@@ -1,0 +1,381 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * objectui#7137 — the Interfaces pillar's block `selection` must not outlive a
+ * leaf change.
+ *
+ * ## The mechanism it pins
+ *
+ * `InterfacesPillar`'s draft-load effect used to be the only thing that cleared
+ * the selection, and its clear sat AFTER the guard:
+ *
+ *     if (!current || !isEditable) { setDraft({}); setHasDraft(false); return; }
+ *     …
+ *     setSelection(null);   // never reached on the early-return path
+ *
+ * So walking from a leaf with a block selected to any leaf where
+ * `isEditable === false` carried the selection across, and it then described a
+ * block on the PREVIOUS leaf's canvas.
+ *
+ * The repair keys the selection to the leaf it was made on — the same
+ * "expires by construction" shape `blockingReport` already uses against
+ * `inspectorKey` in this component — so a leaf change nulls it in the SAME
+ * render rather than one committed render later.
+ *
+ * ## Two populations, two different observables — both are pinned here
+ *
+ * `isEditable = !!Preview && !StudioCanvas` is a conjunction, so it is false
+ * for two disjoint reasons, and #7121's landed gating covers only one of them:
+ *
+ *   (a) **studio-canvas leaves** (`object`). #7121 gates the rail and the
+ *       header's clear button on `StudioCanvas`, so both of those symptoms are
+ *       already unreachable here. What it did NOT gate is `hasInspectorTarget`,
+ *       which feeds `nextCenterTab` — so in the folded layout the ONLY
+ *       surviving observable of the leak is the center tab, and that is what
+ *       the folded-layout test below measures.
+ *
+ *   (b) **leaves whose own type has no registered designer**. Nothing gates
+ *       this population: the stale selection reaches the
+ *       `selection && Inspector && current` rail branch directly and opens the
+ *       scoped inspector for a foreign block. Pinned via a mount LEDGER rather
+ *       than a point-in-time query, because the defect being excluded is a
+ *       mount that is torn down again a render later.
+ *
+ * ## ⚠️ The registry is POPULATED on purpose
+ *
+ * Same precondition as `StudioDesignSurface.designerRegistryPartial.test.tsx`:
+ * a designer is registered for an unrelated type, so a `page` leaf lands in the
+ * no-designer branch for reason (1) "this type has none" — a product fact — and
+ * not for #6795 part C's reason (2) "the registries never loaded". A
+ * zero-registry reading would measure that other card instead.
+ *
+ * ## ⛔ Do NOT re-fence this on `isEditable`
+ *
+ * The discriminator for #7121's rail / Design-mode gating is `StudioCanvas`,
+ * never `isEditable` — `StudioDesignSurface.designerRegistryPartial.test.tsx`
+ * is the live fence for that and goes red on the substitution. This card is the
+ * opposite direction: the STATE lifecycle, which is correctly keyed to the leaf
+ * and not to either discriminator.
+ */
+
+import '@testing-library/jest-dom/vitest';
+import * as React from 'react';
+import { describe, it, expect, vi, afterEach, beforeEach } from 'vitest';
+import { render, screen, fireEvent, cleanup, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+// jsdom has no matchMedia — `useIsMobile` (rail overlay) and
+// `useIsWideViewport` (the folded layout's side-by-side threshold) both need a
+// stub. `innerWidth` stays at jsdom's 1024 default, i.e. BELOW the 1280 `xl`
+// breakpoint, so `foldInspector` really does produce the center-tabs layout
+// rather than the wide side-by-side one.
+window.matchMedia = ((query: string) => ({
+  matches: false,
+  media: query,
+  onchange: null,
+  addListener: () => {},
+  removeListener: () => {},
+  addEventListener: () => {},
+  removeEventListener: () => {},
+  dispatchEvent: () => false,
+})) as unknown as typeof window.matchMedia;
+
+/**
+ * Three leaves, one app — one of each population plus the editable leaf the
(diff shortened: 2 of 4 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs read in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `db22d18b0`.

## The reason this card went well: my suggested route was tested, not obeyed

I put `inspectorKey`-style leaf-keying forward as a **weak preference** in zone 3, and said to pick by measurement. That is exactly what happened — and it is worth being precise about, because "the PM's preference won" and "the PM's preference was measured to be better" look identical in a diff and are not the same thing.

Candidate 1 (hoist `setSelection(null)` above the guard) was **built from the base blob** and run against the same pin:

| repair | pin result | stale-selection ledger |
|---|---|---|
| unfixed | `3 failed \| 2 passed (5)` | 3 stale committed renders |
| candidate 1 — the hoist | `1 failed \| 4 passed (5)` | **1** |
| landed — leaf-keyed | `5 passed (5)` | **0** |

**One frame decided it.** Direction predicted before running. The hoist looked equivalent by reading and was not, and nothing but building both would have shown that.

Equally good: the two green arms in the red baseline were **named and explained**, not left as survivors. `does not carry the selection onto a studio-canvas leaf` passes on *both* sides because objectui#7121 already orders the studio-canvas rail branch ahead of the selection branch — which is *the finding*, and why the folded-tab arm is the real measurement for population (a).

## Three of my briefing assumptions were falsified, and I verified the load-bearing one myself

**A2.2(b) — sub-population (b) is EMPTY, not merely small.** I briefed "leaves whose type has no registered designer" as a second population. `registerBuiltinPreviews()` registers a preview for **all five** leaf types, so (b) is reachable only in a host registering a partial set. I checked this on `main` rather than accept it: `page`, `object`, `dashboard`, `report`, `action` all registered, plus nine more. The falsification holds. Population (a) is **1** (`object`) — thin, and reported as thin.

**A2.4 — falsified softly, and handled the right way.** No test *asserts* the selection survives, but `studioCanvasLeaf.test.tsx`'s fourth test carried it as a written **premise**, and two of its three assertions were measuring objectui#7121's gating *against a live stale selection*. After this fix they hold for a stronger reason. Nothing was weakened or deleted — the comment was corrected and the place objectui#7121 stays independently pinned was named. Deleting those assertions would have been the easy move and would have quietly reduced coverage.

**The `centerTab` mechanism was refined, not confirmed.** This was the sub-claim I required be measured or re-reported as NOT MEASURED. It was measured, and the card's description turned out wrong: it is **not** an auto-switch *to* Properties. `hasInspectorTarget` never changes across the leaf walk, so `nextCenterTab` sees no edge at all and the author is **stranded** on Properties. Same defect, more precise cause.

And the consequence matters for scope: **post-objectui#7121 the folded center tab is the only surviving observable of this leak on population (a)**, because the rail and clear-button symptoms are already gated off there. A fix validated on the rail symptoms alone would have measured nothing.

## Recorded: an observation deliberately not filed

You raised this for my call rather than filing it — `centerTab`'s effect early-returns on `!showFoldedTabs` **without updating `prevInspectorTargetRef`**, so the ref goes stale across a viewport resize. You traced both directions, found each reachable edge produced a defensible outcome, and **did not measure it**.

**I agree with not filing it.** An unmeasured "maybe" in the queue costs a future triage pass more than it saves, and this seat has spent real time on cards whose disqualifier only appeared on a full read. Recording it here instead so it is not lost — that is the right home for a traced-but-unmeasured observation.

## Ledger

- Clause ② **not engaged, verified rather than assumed**: `InterfacesPillar` is a module export that `packages/app-shell/src/index.ts` does not re-export, and no exported type moved.
- Regression surface: `views/studio-design/` 43 files / 230 tests, `views/metadata-admin/` 221 files / 2303 tests — includes objectui#7121's own pin, its live fence `designerRegistryPartial`, objectui#6795 part C's `designerRegistryMissing`, and `centerTab`'s unit tests.
- Mutation proven by injected-marker count and blob hash; restore proven by state (`git diff HEAD`, `git diff --cached`, `git status --short` all empty), not by the trap firing.
- Three comments this change falsified were corrected in place — two in `StudioDesignSurface.tsx` asserting the leak was still live, one premise note on the objectui#7121 pin.
- `mergeable_state` read `behind` while you worked; that is normal for parallel lanes and the queue rebuilds entries. `main` has since advanced twice more.

Landing will be verified by content with a live control once `main` advances — a queue sha is not a landing.

---

PR #7146, opened 2026-09-01
fix(plugin-charts): a sankey with no positive flow says so instead of rendering an empty div

### LINKED ISSUE #7140: finding(plugin-charts): a sankey whose rows are all zero-or-negative renders a literally empty div — the one silent blank left on the chart surface after #7130
Filed unassigned by the implementer of #7130 (PR #7139) while sweeping the chart-side empty-result population. ⛔ Out of scope there and deliberately not fixed in that PR — different defect class, different file.

## The measurement

`packages/plugin-charts/src/AdvancedChartImpl.tsx`, the sankey arm:

```
const rows = data.filter((r) => (Number(r?.[dataKey]) || 0) > 0);
...
const links = rows.map(...);
if (links.length === 0) {
  return <div className={className} />;
}
```

The filter keeps only strictly positive values. So a sankey handed **real rows whose measure is all `0`, all `null`, or negative** produces no links and returns a bare `div` — no marks, no message, no role, no `data-chart-error` code. Nothing at all renders.

This is **not** the empty-result case #7130 covers. `ObjectChart`'s new branch is gated on `finalData.length === 0`; here `finalData.length` is greater than zero — the query succeeded AND returned rows — so the empty state correctly does not fire, and this blank survives it.

## Why it is worth recording now rather than later

Before #7130 this arm was one silent blank among several, so it did not stand out. After it, every other way an `ObjectChart` can come up empty explains itself — loading skeleton, `chart-error` alert box, `chart-no-datasource`, and now the `chart-empty-state` status box — and this is the last path that renders a blank tile and says nothing. The same reasoning the maintainer applied in hotcrm#1212 reads straight across: nothing on screen tells the reader whether the widget failed or is simply reporting a genuinely all-zero flow.

Note also that the two existing refusals in this file (`hasNoCategoryKey`, `hasNoPlottableSeries`) are both gated `rows.length > 0` and render through the shared `ChartRefusal` shell with `role="status"` and a diagnosis code. That shell is the obvious seam for this case; it is right there in the same file, which is why the fix looks small. Whether the honest copy is "nothing to plot" or something naming the all-zero cause is a judgement this card does not pre-make.

## ⛔ What this card does NOT assert

It does not assert the blank is wrong. An all-zero flow may be a legitimate thing to draw as nothing, and the same restraint #7130 was filed under applies here: whoever takes it should render it and look at it first, and closing it measured-and-declined is a legitimate outcome.

## Related

- **#7130** / PR #7139 — the empty-result branch this sits beside; that fix does not reach this path.
- **#4695** — the adjacent open finding on the same file (a cartesian chart handed no series binding at all). Different trigger, same family of silence.
- **#7132** — the `DataEmptyState` convergence card, which may absorb the copy question.

os-warren:
Fixes #7140

## What was measured, before anything was written

The card asked for the blank to be **rendered and looked at first**, and said
closing it measured-and-declined would be a full success. So it was rendered
first, in a real Chromium, at `origin/main` **e8e4c4df5**, in a 520x240 tile
against a populated control — twelve tiles, one page.

| tile | descendants | svg | path | role | on screen |
|---|---|---|---|---|---|
| sankey, all values positive (CONTROL) | 26 | 1 | 7 | — | a sankey |
| sankey, every measure `0` | **1** | 0 | 0 | none | **nothing** |
| sankey, every measure `null` | **1** | 0 | 0 | none | **nothing** |
| sankey, every measure negative | **1** | 0 | 0 | none | **nothing** |
| sankey, every measure unparseable | **1** | 0 | 0 | none | **nothing** |
| sankey, one positive among zeros | 18 | 1 | 3 | — | a sankey |
| sankey, no rows at all | 1 | 0 | 0 | none | nothing |
| bar / treemap, all zero | 104 / 25 | 1 | — | — | axes, labels, legend |
| existing refusal (CONTROL) | 3 | 0 | 0 | `status` | its message |

The four blank sankey tiles hashed **byte-identical screenshots** to one
another (`b53cc8c4…`), and the control hashed differently — so the control
returns non-zero on the same instrument and the zeros are real.

**The decision this produced: fix, not decline.** Not because a blank is the
wrong picture of an empty flow, but because of what the DOM says: one element,
no marks, no text, **no `role`** — the only path in this file that puts nothing
at all on the page. A reader cannot tell a genuinely all-zero flow from a
widget that crashed, and a screen-reader user is handed literally nothing. That
is the exact distinction the file's two existing refusals exist to make, three
lines away.

## The change

`AdvancedChartImpl.tsx`, the sankey arm only. When the filter keeps no rows,
it now renders through the **`ChartRefusal` shell already in this file** — same
box, same `role="status"`, new `data-chart-error="no-positive-flow"`:

> This chart has no flow to draw: no row's `amount` is above zero.

Two boundaries are deliberate, and each is pinned by a test:

- **No rows at all still returns the bare div.** That is the empty-RESULT
  question, answered upstream in `ObjectChart` where the query outcome is
  known. "No row's measure is above zero" would be a false sentence about a
  dataset with no rows in it — measured: that tile hits the same branch today.
- **One positive row among zeros still draws.** The refusal fires on an empty
  link set, never on a thin one.

**One code, one sentence, three causes.** A genuinely all-zero flow, values a
flow cannot represent because they are negative, and measures `Number(...) || 0`
folds to zero all arrive at the same branch. The causes differ — an all-zero
pipeline is a young dataset, an all-negative one is a data-modelling mistake —
but naming any single one is a sentence that is false for the other two, so the
copy names the predicate the filter actually applies, which is true for all
three. This is the same reasoning `hasNoPlottableSeries`' own docstring gives
for its single code and two causes. No recovery is promised.

**No console warning**, unlike the two refusals below it: those carry a
diagnostic pair (`xAxisKey` plus the keys the rows carry) that does not fit on
screen. This message already names the key and the exact test it failed, and
the guard sits inside the arm — beside the filter it reads — where no hook can
be added and where it cannot drift from that filter.

## Blast radius, measured the same way

Re-rendered the same twelve tiles after the change and hashed every screenshot
against its before image:

- **CHANGED — exactly 4**: the all-zero, all-null, all-negative and
  unparseable sankey tiles.
- **UNCHANGED — the other 8, pixel-identical**: the populated control, the
  one-positive-among-zeros sankey, the no-rows sankey, bar / pie / funnel /
  treemap all-zero, and the existing `missing-category-key` refusal.

## Ablation

Direction predicted **before** the run: reverting the guard turns the four
no-positive-flow cases red and leaves the four boundary cases green.

Mutation proven on disk by **blob hash and marker count**, not by an editor's
exit code: `3105af33…` -> `baf47ce1…`, guard marker `1` -> `0`.

```
FAIL  ... when every measure is 0 / null / negative / unparseable
AssertionError: renders the explanatory placeholder: expected null not to be null
Test Files  1 failed (1)
     Tests  4 failed | 4 passed (8)
```

`4 failed | 4 passed (8)` — matching the prediction exactly. The 4 survivors
are the boundary tests, which assert behaviour the change does **not** touch;
a run that collapsed to "4 failed, 0 passed" would have meant the file stopped
loading rather than that the guard was doing the work.

Restore proven by **state**: `git diff HEAD`, `git diff --cached` and
`git status --short` all empty afterwards, worktree blob back to
`3105af33…`, marker back to `1`. No dist leg applies here — the test imports
`./AdvancedChartImpl` relatively inside its own package, so vitest reads the
mutated source directly, which the red/green flip itself demonstrates.

## Verification — union run at `c14d945b6`

| check | verdict |
|---|---|
| `pnpm exec vitest run packages/plugin-charts/` | `Test Files 36 passed (36)` / `Tests 248 passed (248)` |
| the new file alone | `Test Files 1 passed (1)` / `Tests 8 passed (8)` |
| `pnpm --filter '@object-ui/plugin-charts' run type-check` | exit 0 — and `tsc --listFiles` puts **both** changed files in the program, so this reading covers the new test file |
| `npx eslint .` in `packages/plugin-charts` | exit 0 — 0 errors, 273 warnings, all pre-existing `no-explicit-any` / `react-refresh`; 49 files linted, both changed files among them, 6 warnings on the new test (siblings carry 2-17 of the same) |
| `pnpm changeset:check` | `All workspace packages are in the changeset fixed group.` / `privatePackages declared: version=true, tag=false.` / `No changeset declares a 'major' bump.` |
| `check-control-bytes` | `OK (scanned 5922 tracked text file(s); skipped 85 binary)` |
| `check-vi-mock-specifiers` | `OK (4089 tracked source file(s), 2362 test-named; 521 carry a mock ...)` |
| `check-vi-mock-inherit` | `OK (... 117 inherit, 0 auto-mocked ...)` |
| `check-lint-coverage` | `lint coverage: 46/46 packages linted, 0 with outstanding errors` |
| `check-i18n-call-site-keys` | exit 0 |
| `check-i18n-en-drift` | `No en value changed in this range.` |
| `check-i18n-dead-keys` | exit 0 |
| `check-phantom-dependencies` | `Every in-scope import is declared by the package that publishes it.` |
| `check-package-self-import` | `No package names itself inside its own src/.` |
| `check-sdui-registration-pins` | **NOT MEASURED** — exit 2, `No console build to weigh at apps/console/dist/assets`; the gate itself says "This is exit 2, not a pass ... a run with nothing to read has measured nothing". CI builds the console and owns this reading. Declared narrowing: this diff adds no registration and touches no registration array. |

The whole-repo `pnpm lint` and the rest of the gate farm are CI's run, not
duplicated here.

## Scope

- **No i18n key added.** `AdvancedChartImpl.tsx` has no i18n call sites at all
  — both existing refusals are hardcoded English — and `check-i18n-drift`
  confirms no `en` value moved. Adding a translation hook here would be a new
  subsystem in a component whose hook count is deliberately fixed, not this
  card.
- **No shared abstraction built, promoted or extended.** The seam used is the
  `ChartRefusal` shell in this same file, which took this case **without
  modification** — it already accepts a `code`, a `className` and children.
- **#4695 is not addressed here** and its path is not touched: it is a
  cartesian chart handed no series binding, which never reaches the sankey arm.
- No public export widened, no parse accept/reject behaviour changed.

Files: `packages/plugin-charts/src/AdvancedChartImpl.tsx`,
`packages/plugin-charts/src/AdvancedChartImpl.sankeyNoPositiveFlow.test.tsx`
(new), `.changeset/7140-sankey-no-positive-flow.md` (new).

---

### DIFF
--- a/.changeset/7140-sankey-no-positive-flow.md
+++ b/.changeset/7140-sankey-no-positive-flow.md
@@ -1,0 +1,36 @@
+---
+'@object-ui/plugin-charts': minor
+---
+
+A sankey with no positive flow says so, instead of rendering an empty div
+(objectui#7140).
+
+`AdvancedChartImpl`'s sankey arm keeps only strictly positive measures, so a
+chart handed **real rows** whose measure is all `0`, all `null`, all negative,
+or unparseable built no links and returned a bare `<div>`. Measured in Chromium
+against a populated control: the control drew 1 `<svg>` / 7 `<path>` /
+26 descendants; each of those four tiles rendered `descendantCount: 1`,
+`svgCount: 0`, `textContent: ''`, and their screenshots hashed identical to one
+another. No marks, no text, no `role` — a tile indistinguishable from a widget
+that had crashed, which is the one distinction the file's other refusals exist
+to make.
+
+It now renders through the `ChartRefusal` shell those refusals already use —
+same box, same `role="status"`, and a new `data-chart-error="no-positive-flow"`
+— reading *"This chart has no flow to draw: no row's `<measure>` is above
+zero."*
+
+Two boundaries are deliberate and pinned:
+
+- **No rows at all is untouched.** That is the empty-result question, answered
+  upstream in `ObjectChart` where the query outcome is known; a sentence about
+  what the rows contain would be false about a dataset with no rows in it.
+- **One positive row among zeros still draws.** The refusal fires on an empty
+  link set, never on a thin one.
+
+One code and one sentence for three causes (a genuinely all-zero flow, values a
+flow cannot represent because they are negative, and measures `Number(…) || 0`
+folds to zero): naming any single cause would be false for the other two, so
+the copy names the predicate the filter actually applies, which is true for all
+three. No recovery is promised. Every other chart family is byte-identical —
+eight of the twelve tiles in the browser sweep hashed unchanged.
--- a/packages/plugin-charts/src/AdvancedChartImpl.sankeyNoPositiveFlow.test.tsx
+++ b/packages/plugin-charts/src/AdvancedChartImpl.sankeyNoPositiveFlow.test.tsx
@@ -1,0 +1,146 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * Sankey — rows arrived, none of them is a positive number (objectui#7140).
+ *
+ * The sankey arm keeps only strictly positive measures
+ * (`data.filter((r) => (Number(r?.[dataKey]) || 0) > 0)`), so a chart handed
+ * REAL rows whose measure is all `0`, all `null`, all negative, or unparseable
+ * built no links and returned a bare `<div className={className} />`.
+ *
+ * Measured in Chromium against a populated control before the fix, at
+ * `origin/main` e8e4c4df5: the control drew 1 `<svg>` / 7 `<path>` /
+ * 26 descendants; each of the four blank tiles rendered `descendantCount: 1`,
+ * `svgCount: 0`, `textContent: ''`, and their screenshots hashed identical to
+ * one another. Nothing was on the page — no marks, no text, no `role` — so the
+ * tile was indistinguishable from a widget that had crashed, which is the one
+ * distinction every other refusal in that file exists to make.
+ *
+ * The two boundaries this pins are the ones that make the message TRUE rather
+ * than merely present:
+ *
+ *   - **no rows at all still returns the bare div.** That is the empty-RESULT
+ *     question (objectui#7130), answered upstream in `ObjectChart` where the
+ *     query outcome is known. "No row's measure is above zero" would be a false
+ *     sentence about a dataset with no rows in it.
+ *   - **one positive row among zeros still DRAWS.** The refusal fires on an
+ *     empty link set, never on a thin one; a sankey that can draw anything is
+ *     never replaced by prose.
+ */
+import React from 'react';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+
+vi.mock('recharts', async () => {
+  const actual = await vi.importActual<any>('recharts');
+  return {
+    ...actual,
+    ResponsiveContainer: ({ children }: any) =>
+      React.cloneElement(children, { width: 480, height: 320 }),
+  };
+});
+
+import AdvancedChartImpl from './AdvancedChartImpl';
+
+afterEach(cleanup);
+
+const SERIES = [{ dataKey: 'amount', label: 'Amount' }];
+
+const renderSankey = (data: Array<Record<string, unknown>>) =>
+  render(
+    <AdvancedChartImpl
+      chartType="sankey"
+      xAxisKey="stage"
+      series={SERIES as any}
+      data={data as any}
+    />,
+  );
+
+const refusalOf = (container: HTMLElement) =>
+  container.querySelector('[data-chart-error="no-positive-flow"]');
+
+/**
+ * The four row shapes that reach the empty link set. They are NOT the same
+ * situation — a genuinely all-zero flow, values a flow cannot represent because
+ * they are negative, and measures `Number(…) || 0` folds to zero — but the
+ * predicate the filter applies is the one thing true of all of them, so one
+ * message serves all four without saying anything false about any of them.
+ */
+const NO_POSITIVE_ROWS: Array<[string, Array<Record<string, unknown>>]> = [
+  ['every measure is 0', [{ stage: 'Prospecting', amount: 0 }, { stage: 'Won', amount: 0 }]],
+  ['every measure is null', [{ stage: 'Prospecting', amount: null }, { stage: 'Won', amount: null }]],
+  ['every measure is negative', [{ stage: 'Refunds', amount: -40 }, { stage: 'Credits', amount: -12 }]],
+  ['every measure is unparseable', [{ stage: 'A', amount: 'n/a' }, { stage: 'B', amount: 'n/a' }]],
+];
+
+describe('AdvancedChartImpl — sankey with no positive flow (objectui#7140)', () => {
+  it.each(NO_POSITIVE_ROWS)('says so instead of rendering nothing when %s', (_label, rows) => {
+    const { container } = renderSankey(rows);
+
+    const refusal = refusalOf(container);
+    expect(refusal, 'renders the explanatory placeholder').not.toBeNull();
+    // A refusal is a STATE, not an alert — the shell the other two refusals in
+    // this file render through.
+    expect(refusal?.getAttribute('role')).toBe('status');
+    // Names the measure it tested, so the author knows WHICH column was all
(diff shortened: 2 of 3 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs read in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `c14d945b6`.

## ⚖️ Ruling on the open question — **Option A, as shipped. No change needed.**

You asked whether an all-zero flow and an all-**negative** flow deserve separate codes and copy. Keeping **A**: one code `no-positive-flow`, one sentence naming the predicate.

The reason is not that B is wrong in principle — you are right that the two situations are distinguishable at runtime, unlike the byte-identical upstream shapes that forced `hasNoPlottableSeries` into a single message. It is that **B branches copy on a distinction that does not change what the reader must do.** What they need to learn is that this is not a crash and which column failed which test; the predicate sentence delivers both and stays true in every reaching case.

And the situation B would gesture at — a dataset where negatives are *present and meaningful* — is **not reachable from this branch at all.** It is objectui#7148, where some rows survive the filter and the chart draws a confident partial flow. Spending this card's copy budget hinting at a defect that lives in the branch next door is worse than naming the predicate plainly. Filing it as its own card was the right instinct.

C is correctly rejected for the reason you gave: the file's doctrine is that the code names *which* refusal fired, so copy varying under one code makes the code stop meaning that.

### ⚠️ One imprecision I am accepting, recorded rather than hidden

`Number(r?.[dataKey]) || 0` folds an **unparseable** measure to `0`. So *"no row's amount is above zero"* is true of the **post-coercion** values, not literally of the raw data — `"abc"` is not-a-number rather than not-above-zero. Naming that would force the copy to explain coercion, which is worse than the imprecision. objectui#7148 covers the silent-drop family, unparseable measures included.

Stating it so nobody later reads that sentence as an exact claim about the dataset.

## What earned the arm

**You rendered it before writing anything.** Twelve tiles in a real Chromium at `e8e4c4df5`, against a populated control. That is the standard this card was dispatched under, and `measured-and-declined` was genuinely available — the finding is what ruled it out.

And the finding was stronger than the card claimed: not merely a blank, but `descendantCount: 1`, `svgCount: 0`, no `role`, **pixel-identical to a crashed render**. One element and nothing else — the only path in the file putting nothing at all on the page, three lines from two existing refusals that exist to make exactly this distinction.

**Blast radius by per-tile screenshot hash — exactly 4 changed, 8 pixel-identical.** That is a far better instrument than a test count for a change whose whole subject is what a reader sees, and the four unchanged boundary tiles (control, one-positive-among-zeros, no-rows, the existing refusal) carry the argument that nothing else moved.

**Ablation `4 failed | 4 passed (8)`**, predicted before running, with the four survivors being the boundary cases — so a collapse to `4 failed, 0 passed` would have been distinguishable from the real result. Mutation proven by blob hash **and** marker count; restore proven by state.

**Both boundaries pinned deliberately**: zero rows still returns the bare div (that is the empty-*result* question `ObjectChart` owns upstream, and a sentence about what the rows contain would be false of a dataset with no rows), and one positive row among zeros still draws.

### ⚠️ My i18n warning was well placed and the answer was neither system

I told you to verify which i18n system this file uses rather than take my word, because I got that wrong on objectui#7121. The answer turned out to be **neither**: `AdvancedChartImpl.tsx` has **no i18n call sites at all** and both existing refusals are hardcoded English. Matching the file and adding no key is correct — and "this file is simply untranslated" is a separate question from this card, correctly not folded in.

### A2.1 falsified, in the predicted direction and then some

PR objectui#7139 had not landed when you measured (it has since, as `dd35800af`). More importantly the card's framing — *"the last silent blank"* — is **false even now**: funnel and pie handed all-zero rows render a screenshot-identical blank from **rich** DOM (34 and 31 descendants, a real `svg`), failing through degenerate geometry rather than an early return. Filed as objectui#7147. Zero area is not zero elements, and no source-level sweep would have found it.

Both follow-ups (objectui#7147, objectui#7148) are `pm:blocked` on this PR landing, with the collision recorded. objectui#7148 has priority when it unblocks — it is the most dangerous defect surfaced this session, because it is invisibly wrong rather than visibly wrong.

Landing will be verified by content with a live control once `main` advances.

---

PR #7150, opened 2026-09-01
fix(plugin-detail): ActivityTimeline's empty state resolves from the locale packs

### LINKED ISSUE #7142: `ActivityTimeline`'s empty state is a hardcoded English literal, so a zh session reads "No activity recorded"
Found while measuring the empty-state surfaces for #7132. Out of scope there, filed rather than fixed.

`packages/plugin-detail/src/ActivityTimeline.tsx:138`:

```tsx
<DataEmptyState title="No activity recorded" className="py-6" />
```

The title is a raw English string literal. It is not a `t()` call and not an inline `defaultValue` — it never reaches the pack system at all, so it stays English in all ten locales.

## Why it is worth a card

Its sibling on the same surface does it correctly. `RecordActivityTimeline.tsx:415` renders the same state as:

```tsx
title={emptyLabel ?? t('detail.noActivity')}
```

So the key already exists and is already translated in the packs — this call site simply does not use it. That makes the fix small and the current state a straightforward inconsistency rather than a missing-translation project.

## Measured

Rendered `ActivityTimeline` with `activities={[]}`; the empty box reads `text="No activity recorded"` with the literal coming straight from the JSX, not from a pack lookup.

⚠️ Note for whoever takes this: confirm whether `detail.noActivity` is the right key for *this* component's phrasing before reusing it, rather than assuming the two surfaces want identical copy. If they do want different copy, that is a new key in `en.ts` plus its nine sibling packs, with `all-locales-key-parity` green — never an inline `defaultValue` (#3517).

Refs: #7132.

os-warren:
Fixes #7142

`ActivityTimeline`'s empty-state title was a raw English JSX literal — `title="No activity recorded"`, not a `t()` call and not an inline `defaultValue` — so it never reached the pack system and stayed English in all ten locales. It now reads `detail.noActivity` through `useDetailTranslation`, the hook its sibling already uses.

Three lines of production change: the hook import, the `const { t } = useDetailTranslation();` call, and the title expression. No pack was edited.

## The key-reuse decision was measured, not assumed

The card and the dispatch both flagged this as the thing to settle first: `RecordActivityTimeline.tsx:415` already renders `title={emptyLabel ?? t('detail.noActivity')}`, but that proves a key **exists**, not that the two surfaces want the same words. So both were rendered empty and read.

```
[en] ActivityTimeline       => "Activity(0)No activity recorded"
[en] RecordActivityTimeline => "Activity(0)All ActivityNo activity recorded"
[zh] ActivityTimeline       => "Activity(0)No activity recorded"      <- the defect
[zh] RecordActivityTimeline => "活动(0)全部动态暂无活动记录"
```

The `en` pack value for `detail.noActivity` is `'No activity recorded'` — **byte-identical** to the literal being replaced. The two surfaces were already saying exactly the same words in English, so reuse changes no copy in any locale, while a second key would fork one sentence across ten packs for zero copy difference. Reuse it is.

## The key is genuinely translated in all ten packs

Read out of the pack objects directly rather than grepped — the packs are nested (`detail: {` … `noActivity:`), so a dotted-key grep for `detail.noActivity` returns **zero** against a pack that defines it. Two controls ran on the same instrument: `detail.noRelatedRecords` and `detail.loading` returned present in all ten (positive), and `detail.zzzNotAKeyEver` returned absent in all ten (negative, proving the probe can report absence).

```
en  "No activity recorded"          zh  "暂无活动记录"
ja  "アクティビティの記録なし"          ko  "기록된 활동 없음"
de  "Keine Aktivitäten aufgezeichnet" fr  "Aucune activité enregistrée"
es  "Sin actividad registrada"      pt  "Nenhuma atividade registrada"
ru  "Нет зарегистрированной активности"  ar  "لا يوجد نشاط مسجل"
```

`DETAIL_DEFAULT_TRANSLATIONS` already carries the byte-identical English row, so the provider-less path is unchanged and `defaults-maps-mirror-en-pack` stays green. No inline `defaultValue` anywhere (#3517).

## Test

`packages/plugin-detail/src/ActivityTimeline.i18n.test.tsx`, six cases: zh, ar, en, no-provider, never-the-raw-key, and the second route into the box — a populated timeline filtered down to a type with no entries.

`zh` and `ar` are the load-bearing half. An `en`-only assertion would have been **green before the fix too**, since the literal and the `en` pack value are the same string; only a non-Latin locale discriminates a pack lookup from a hardcoded literal.

### Ablation

Predicted before running: restoring the literal turns exactly the three locale-discriminating cases red and leaves the other three green. Observed `Tests 3 failed | 3 passed (6)` — the passing count is quoted deliberately, since a suite collapsed to zero survivors also shows red and would look stronger.

Mutation proven on disk by marker count in both directions (`t('detail.noActivity')` 1 to 0, `title="No activity recorded"` 0 to 1) and by blob hash (`6bb6ce32` to `678d2e03`), never by an editor's exit code. Restore proven by state — `git diff HEAD`, `git diff --cached` and `git status --short` all empty, and the blob back to `6bb6ce32` — never by a trap firing. No rebuild leg applies: the test imports `./ActivityTimeline` relatively and the root vitest config aliases `@object-ui/i18n` to `packages/i18n/src`, so nothing in the closure resolves through `dist/`.

## Verification

All readings below are on `0e210de56`, the final commit.

Green:

- `pnpm exec vitest run packages/plugin-detail/ packages/i18n/src/__tests__/all-locales-key-parity.test.ts packages/app-shell/src/__tests__/defaults-maps-mirror-en-pack.test.tsx` — `Test Files 123 passed (123)`, `Tests 1157 passed (1157)`
- `check:i18n-keys` — "Every in-scope call-site key resolves against the en pack (2845 keys) … no call site carries a literal fallback beside itself" (2604/2604 literal keys resolve)
- `check:i18n-drift` — "No en value changed in this range."
- `check:i18n-dead-keys` — report, not a gate; `detail.noActivity` gains a call site here rather than losing one
- `check-changeset-presence` — "1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)"
- `check-changeset-no-major` — "No changeset declares a `major` bump." (quoted rather than predicted)
- `type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`) — exit 0. Confirmed it actually covers the diff: `--listFiles` puts `ActivityTimeline.tsx` in the main pass and `ActivityTimeline.i18n.test.tsx` in the `tsconfig.test.json` pass, with a present-file control and a nonexistent-path control both behaving.
- `eslint .` in `packages/plugin-detail` — 0 errors. The one warning on `ActivityTimeline.tsx` (`Filter` imported unused, line 11) is pre-existing on `origin/main` and untouched here.
- `check:control-bytes`, `check:self-import`, `check:esm-specifiers`, `check:phantom-deps`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:dist-completeness`, `check:side-effects-array`

NOT MEASURED — recorded as the absence of a reading, not as passes:

- `check:sdui-registration-pins` — exit 2, "a run with nothing to read has measured nothing. Build the console first". The console build needs 28 unbuilt packages; that full-monorepo build is CI's, and this diff registers no SDUI component.
- `check:eager-closure` — exit 2, "a broken gauge, not a sensitive gate": no `apps/console/dist/eager-closure.json` locally. Same cause.
- `check:readme-exports` — the gate declares its own run void ("the population COLLAPSED — this run proves nothing") while packages are unbuilt. Narrowed and declared: after building `@object-ui/plugin-detail`, its README dropped out of the unjudged list (1 hit to 0), so this package's slice is judged; the residual collapse is the other 24 unbuilt packages.

`type-check` first came back exit 2 with TS2307 across every `@object-ui/*` — an unbuilt dependency closure, which is a non-reading rather than a red. The closure was built (`pnpm --filter '@object-ui/plugin-detail^...' build`) and it was re-run to get the exit 0 above.

## Scope

Only the empty-state title, per the dispatch ruling. Sweeping the file found the component uses **no** translation hook at all on `origin/main` — 18 further hardcoded English literals (relative timestamps, the six filter chips, the `Activity` card title, the filter `aria-label`, the six activity-description sentences), 12 of which already have pack keys sitting unused. That inventory is filed as #7149, unassigned, rather than folded in here.

No public export added, no accept/reject behaviour changed — contract clause 2 stays disengaged. No `role` was added, removed or aligned on the `DataEmptyState` call site; that is #7132's decision and it is already made.

---

### DIFF
--- a/.changeset/7142-activity-timeline-empty-i18n.md
+++ b/.changeset/7142-activity-timeline-empty-i18n.md
@@ -1,0 +1,23 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+`ActivityTimeline`'s empty state speaks the session locale (objectui#7142).
+
+The title was a raw English JSX literal — `title="No activity recorded"`, not a
+`t()` call and not an inline `defaultValue` — so it never reached the pack
+system and stayed English in all ten locales. Measured before the fix by
+rendering `activities={[]}` under a zh `I18nProvider`: the card read
+`"Activity(0)No activity recorded"`, while its sibling `RecordActivityTimeline`
+rendered `"活动(0)全部动态暂无活动记录"` from the same packs.
+
+The call site now reads `detail.noActivity`, the key the sibling already uses.
+Reusing it rather than minting a second key is a measured decision: the `en`
+pack value for that key is `'No activity recorded'`, byte-identical to the
+literal it replaces, so both surfaces were already saying the same words in
+English and a new key would have forked one sentence across ten packs for no
+copy difference. No pack was edited — the key is already translated in all ten,
+verified by reading `detail.noActivity` out of each pack object.
+
+Both routes to the box are covered: an empty `activities` array, and a
+populated timeline filtered down to a type with no entries.
--- a/packages/plugin-detail/src/ActivityTimeline.i18n.test.tsx
+++ b/packages/plugin-detail/src/ActivityTimeline.i18n.test.tsx
@@ -1,0 +1,122 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `ActivityTimeline`'s empty state speaks the session locale — objectui#7142.
+ *
+ * The title was the raw English literal `"No activity recorded"`, written
+ * straight into the JSX. It was not a `t()` call and not an inline
+ * `defaultValue`, so it never reached the pack system at all and a zh session
+ * read English. Measured before the fix by rendering `activities={[]}` under a
+ * zh `I18nProvider`: the card came out `"Activity(0)No activity recorded"`,
+ * while its sibling `RecordActivityTimeline` rendered `"活动(0)全部动态暂无活动记录"`
+ * from the same packs.
+ *
+ * The call site now reads `detail.noActivity` — the key the sibling already
+ * uses (`RecordActivityTimeline.tsx`), reused rather than forked. That reuse is
+ * a measured decision, not an assumption: the `en` pack value for the key is
+ * `'No activity recorded'`, **byte-identical** to the literal being replaced,
+ * so both surfaces were already saying the same words in English and a second
+ * key would have forked one sentence across ten packs for no copy difference.
+ *
+ * `zh` and `ar` are the load-bearing assertions. An `en`-only test would have
+ * been green *before* the fix too — the literal and the `en` pack value are the
+ * same string — so English proves nothing here. Non-Latin locales are what
+ * discriminate the pack lookup from the hardcoded literal.
+ *
+ * The provider-less case is asserted alongside them because
+ * `useDetailTranslation` is a `createSafeTranslation` hook: a host with no
+ * `I18nProvider` must get the English default from
+ * `DETAIL_DEFAULT_TRANSLATIONS`, never the raw key `detail.noActivity` in the
+ * empty box. No inline `defaultValue` is used anywhere here (objectui#3517) —
+ * the key resolves from the packs or from that defaults map.
+ */
+
+import * as React from 'react';
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, screen, cleanup } from '@testing-library/react';
+import { I18nProvider } from '@object-ui/i18n';
+import type { ActivityEntry } from '@object-ui/types';
+import { ActivityTimeline } from './ActivityTimeline';
+
+afterEach(() => cleanup());
+
+/** The pack values for `detail.noActivity`, read from the packs themselves. */
+const EN = 'No activity recorded';
+const ZH = '暂无活动记录';
+const AR = 'لا يوجد نشاط مسجل';
+const RAW_KEY = 'detail.noActivity';
+
+function renderEmptyIn(language: string) {
+  return render(
+    <I18nProvider config={{ defaultLanguage: language, detectBrowserLanguage: false }}>
+      <ActivityTimeline activities={[]} />
+    </I18nProvider>,
+  );
+}
+
+describe('ActivityTimeline empty state — locale resolution (objectui#7142)', () => {
+  it('reads the zh pack value under a zh session', () => {
+    renderEmptyIn('zh');
+
+    expect(screen.getByText(ZH)).toBeTruthy();
+    // The regression itself: English must not survive into a zh session.
+    expect(screen.queryByText(EN)).toBeNull();
+  });
+
+  it('reads the ar pack value under an ar session', () => {
+    renderEmptyIn('ar');
+
+    expect(screen.getByText(AR)).toBeTruthy();
+    expect(screen.queryByText(EN)).toBeNull();
+  });
+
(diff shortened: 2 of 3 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `0e210de56`.

### You verified the thing I got wrong, instead of taking my word

I told you to determine which i18n system this file belongs to rather than trust me — because I gave the wrong answer on objectui#7121, where `engine.*` strings turned out to live in a two-locale en/zh table outside the pack system, and following me would have added nine dead entries.

You went to the code: `plugin-detail` uses the **ten-pack system** (`en/zh/ja/ko/de/fr/es/pt/ru/ar`) reached through `useDetailTranslation.ts:9` (`createSafeTranslation` over `@object-ui/i18n`), with `DETAIL_DEFAULT_TRANSLATIONS` as the provider-less mirror — **not** the metadata-admin table. Naming the file you verified it in is what makes that checkable rather than another assertion.

### The pack probe avoided the trap it needed to

⛔ A dotted-key grep for `detail.noActivity` returns **zero** against the `en` pack that defines it, because the packs are nested. You read the key out of each pack **object** via tsx instead, with **two** controls: positive (`detail.noRelatedRecords` + `detail.loading` present in all ten) and negative (`detail.zzzNotAKeyEver` absent from all ten). That pairing is what makes "present in all ten" a reading rather than a hope — and all ten carry real translations, not English fallbacks.

This seat has been bitten by exactly that false zero before. Good.

### Ruling 2 settled by rendering, not by assuming

The `en` value for `detail.noActivity` is **byte-identical** to the literal being replaced. So both surfaces already said the same words, and minting a second key would have forked one sentence across ten packs for **zero** copy difference. That is the right call, and it is only knowable by looking at both surfaces rather than reasoning that "two components might want different phrasing."

Ablation `3 failed | 3 passed (6)`, predicted before running — zh, ar and filtered-zh red, en / no-provider / raw-key green. The passing count is stated because a collapse to zero survivors also shows red. No rebuild leg claimed **and that was established, not assumed**: the test imports relatively and the root config aliases `@object-ui/i18n` to `packages/i18n/src`, so nothing resolves through `dist/`.

### Three NOT MEASURED, named rather than counted

`check:sdui-registration-pins` and `check:eager-closure` both exit 2 on an unbuilt console — the absence of a reading, not a pass. The third is the interesting one: `check:readme-exports` **declares its own run void** while packages are unbuilt, and rather than leave it there you built `@object-ui/plugin-detail` and showed its README left the unjudged list (1 → 0), so **this package's slice is judged** and the residual collapse belongs to the other 24. That is a partial reading reported honestly instead of being rounded to either green or red.

### ⚠️ What my scoping decision actually shipped

I fenced this card to the empty state and you held the fence, filing objectui#7149 rather than widening. Correct — and worth stating plainly: this ships **one** translated string into a component that is otherwise entirely untranslated. A zh card now reads `"Activity(0)暂无活动记录"` — empty state right, card title beside it still English. An improvement, and visibly half-done.

The deeper finding is not "18 more literals." It is that the component **used no translation hook at all**, while its sibling renders the identical surface fully in Chinese. objectui#7149 unblocks when this lands.

---

PR #7156, opened 2026-09-01
fix(plugin-grid,plugin-dashboard): derive the relational copy set from its consumers, so a lookup cell honours the spec-declared `displayField`

### LINKED ISSUE #6875: bug(plugin-grid): ObjectGrid's relational copy set is a strict subset of what its own lookup cell and inline picker read (`displayField`, `reference_field`, `descriptionField`, `lookup_columns` never arrive)
Provenance: measured while implementing objectui#6711 (retiring the dead
`reference_to_field` from the same list). Filed unassigned. Not folded into that card — it
names one key and removes it; this is the opposite direction (keys with readers that are
never copied) and is a behaviour change, not a retirement.

## What's measured

`ObjectGrid`'s `applyRelationalMeta` copies a fixed list off the object-schema field def onto
each relational column's `fieldMeta` (`packages/plugin-grid/src/ObjectGrid.tsx`,
`RELATIONAL_META_KEYS`), and that bag is what the cell renderer and the inline picker
receive. The list is a strict SUBSET of what those two consumers read off the same bag.

Consumer reads, verbatim from `packages/fields`:

- `src/index.tsx` (`LookupCellRenderer`):
  `display_field || displayField || reference_field`
- `src/widgets/LookupField.tsx:252` (the grid's inline picker):
  `fieldMeta?.display_field || fieldMeta?.displayField || fieldMeta?.reference_field || 'name'`
- `src/widgets/LookupField.tsx:253`:
  `fieldMeta?.description_field ?? fieldMeta?.descriptionField`
- `src/widgets/LookupField.tsx:268`:
  `fieldMeta?.lookup_columns ?? fieldMeta?.lookupColumns`
- `src/widgets/UserField.tsx:49`:
  `meta?.display_field || meta?.displayField || meta?.reference_field`

Copied by the grid: `display_field`, `description_field`, `lookup_filters`, `lookupFilters`,
`id_field`, `reference_to`, `reference` (+ `titleFormat`, see objectui#6874).

**Never copied, though read on the grid's own path:** `displayField`, `reference_field`,
`descriptionField`, `lookup_columns`, `lookupColumns`.

## Why this is not "the renderer just accepts extra spellings"

Two things make the omission specific rather than cosmetic:

1. **`reference_field` is a DECLARED objectui type member with readers all over the repo** —
   `packages/types/src/field-types.ts` (`LookupFieldMetadata.reference_field`),
   `packages/types/src/views.ts`, its zod mirror, and read by `plugin-detail`
   (`RelatedList`, `RecordMetaFooter`, `RecordDetailDrawer`, `fieldEnrichment`),
   `plugin-charts` (`ObjectChart`), `plugin-form` (`deriveMasterDetail`), `plugin-list`
   (`ListView`, `UserFilters`) and `app-shell` (`resolveActionParams`). Every one of those
   surfaces honours an author's `reference_field`; a lookup rendered by `ObjectGrid` does
   not, because the key never reaches the cell.

2. **The camelCase spellings are the SPEC-declared ones.** `@objectstack/spec@17.2.0`'s
   `FieldSchema` (64 props) declares `displayField`, `descriptionField`, `lookupColumns`,
   `lookupFilters`, `reference` — and none of `display_field`, `description_field`,
   `lookup_filters`, `reference_to`. The grid's list is mostly the objectql runtime
   spellings. It is internally inconsistent about this: `lookupFilters` (camel) IS in the
   copy set next to `lookup_filters`, while `displayField` and `descriptionField` are not
   next to their snake twins — so the list was assembled from a defect report rather than
   derived from the readers.

## What is NOT measured here

- Whether the object schema a live `getObjectSchema` returns ever carries the camelCase
  spellings (the server may normalise on the way out). That check decides how much of item 2
  is reachable today, and it is the first thing triage should run.
- No user report is attached. This is a static producer-vs-consumer measurement.
- `packages/plugin-dashboard`'s `CELL_RELATIONAL_META_KEYS` has the same gap for
  `displayField` / `reference_field`: its docblock enumerates what `LookupCellRenderer`
  reads as three keys, and that enumeration missed the two extra spellings in the same
  fallback chain.

## Shape of a fix

Contract-first says the answer may be at the producer (one spelling, normalised once) rather
than widening a hand-kept alias list in the renderer — `AGENTS.md` #0.1 is explicit that a
tolerant renderer fossilises a second dialect. Whatever the direction, the copy set should
be DERIVED from the consumers' read set rather than restated, so the two cannot drift again;
that is the same repair objectui#6711 and objectui#6874 approach from the dead-key side.

---

os-warren:
Fixes #6875

Verified on `dd35800af` (`origin/main` at branch point). Gate + suite numbers below are
from `a5a6df951`, the final commit.

---

## The card's own first step, run first — and it does not shrink the card

#6875 named one thing triage had to measure before anything else: *"Whether the object
schema a live `getObjectSchema` returns ever carries the camelCase spellings (the server
may normalise on the way out)."* If the producer normalises, widening the copy set fixes
nothing reachable.

It does not normalise, and the answer runs the other way — **the camelCase spellings are
the only ones a live path can carry.**

1. **`@objectstack/spec` 17.2.0's `FieldSchema` is strict, with 71 properties.** Enumerated
   off `FieldSchema.shape` on the installed package, `name`/`type`/`label` as the positive
   control:

   ```
   DECLARED : displayField, descriptionField, lookupColumns, lookupFilters, reference
   absent   : display_field, description_field, lookup_columns, lookup_filters,
              reference_to, reference_field, id_field, idField, titleFormat
   ```

   Strict means an absent key does not pass through — it is refused by name:

   ```
   safeParse({ name:'account', type:'lookup', reference_field:'name' })
     → success: false, code: 'unrecognized_keys', keys: ['reference_field']
   safeParse({ name:'account', type:'lookup', displayField:'name' })
     → success: true
   ```

2. **The producer repo emits no snake_case.** Whole-repo sweep of `objectstack`, each zero
   against a control on the same instrument: `display_field` 3 files / `displayField` 68;
   `description_field` 1 / `descriptionField` 17; `lookup_columns` 2 / `lookupColumns` 13;
   `lookup_filters` 4 / `lookupFilters` 50; `reference_field` **0**. Every snake hit is
   prose — release notes, a CHANGELOG line, liveness-ledger notes describing what objectui
   reads. The platform's published authoring rule states the split outright
   (`skills/objectstack-data/rules/naming.md`): field KEYS snake_case, schema PROPERTIES
   camelCase, with `lookup_filters: []` shown as ❌.

3. **Nothing renames anything on the way in.** `ObjectStackAdapter.getObjectSchema`
   (`@object-ui/data-objectstack`) is the choke point every schema read passes through. Its
   only key rewrites are `normalizeSchemaReferenceKeys` — the `reference` ⇄ `reference_to`
   pair, and nothing else — and `applyFieldWidgetOverrides`, which adds `widget`. Every
   other property arrives verbatim.

⇒ An author who declares `displayField` gets a def carrying `displayField`, and the grid
copied `display_field` instead. Item 2 of the card is not merely reachable; it is the only
reachable half.

## The bug reaching a user

The card is labelled `bug` with no user report attached, so this is shown rather than
asserted. `__tests__/lookupDisplayFieldSpelling-6875.test.tsx` renders two columns over one
data source, one referenced record and one cell renderer, differing only in the spelling on
the field def. Referenced record: `{ id:'p1', name:'Wrong Name', project_code:'ACME-42' }`.

| column | field def | before | after |
|---|---|---|---|
| `code_snake` (control) | `display_field: 'project_code'` | `ACME-42` | `ACME-42` |
| `code_camel` | `displayField: 'project_code'` | **`Wrong Name`** | `ACME-42` |

Measured on the unmodified tree: `{ acme: 1, wrongName: 1, rawId: 0 }` — the spec-spelled
column rendered the referenced record's generic `.name`, because with no display pointer
`getRecordDisplayName` falls through to the `.name`/`.title` heuristic. That is the bug: a
lookup cell showing the wrong display value because the key never arrived. The control
column is what makes it a reading rather than "the fixture never reached the lookup path".

## Derived, not restated

The ruling on this card was that the copy set must be **derived** from the consumers' read
set — a longer hand-kept list is the same defect with a later expiry date. It is also the
card's strongest internal evidence: `lookupFilters` (camel) was in the list while
`displayField` and `descriptionField` were not, though all three come off the same kind of
chain in the same files. **Confirmed** — that asymmetry is real, and it is the tell that
the list was assembled from defect reports.

- `packages/plugin-grid/src/relationalMetaKeys.ts` — a table classifying **every** key the
  three consumers read off this bag, each with a verdict and a reason.
  `RELATIONAL_META_KEYS` is `Object.entries(...)` filtered by verdict. There is no
  hand-kept list left to drift.
- `__tests__/relationalMetaCopySet.derivation.test.ts` — re-extracts the read set from the
  consumer sources on every run (`LookupCellRenderer`'s cast reads, bounded to its own
  function body; `LookupField`'s `fieldMeta?.k`; `UserField`'s `meta?.k`) and requires the
  table to match **exactly, in both directions**. An unclassified spelling is red; an
  orphan is red.
- The gate proves the verdicts against the installed spec rather than against this PR's
  prose: every `no-producer` key must be absent from `FieldSchema.shape`, every `spec` key
  present, and every `legacy-alias` key absent — so if a future spec version declares one,
  the gate says reclassify.

`ObjectGrid.tsx` loses 87 lines and gains 8: the literal, the helper and the retirement
docblock move to the new module, where the retirement history for `reference_to_field`
(#6711) and `titleFormat` (#6874) is preserved in full. Both are now enforced structurally
too — neither appears in the extracted read set, so re-adding either turns the gate red as
an orphan. Their behavioural pins are untouched and still green.

## Which keys were added, and which were deliberately left out

Of the five the card named as never copied, **three** are added:
`displayField`, `descriptionField`, `lookupColumns` — spec-declared, read on this path.

**Two are not**, and that is the card shrinking where it should: `reference_field` and
`lookup_columns` are read on this path but `FieldSchema` declares neither, so no producer
can fill them. Copying them would write a member from the def on every column build that
nothing can ever set — precisely what #6711 (`reference_to_field`), #6625 (`decimals`) and
#6597 (`referenceTo`) retired, and what #6531 removed from `getRecordDisplayName` on the
same reasoning.

## `plugin-dashboard` — the card's unmeasured mention, now measured

The card flagged `CELL_RELATIONAL_META_KEYS` as having the same gap and marked it explicitly
NOT measured. It does. Its docblock enumerated what `LookupCellRenderer` reads as three keys
because it was written from the FIRST leg of each chain; the renderer resolves
`display_field || displayField || reference_field`. `displayField` is added there too (same
spelling, same reason, same read-only path), `reference_field` stays out for the same
producer reason, `FieldMeta` gains the member, the docblock is corrected, and the #6694 pin
is extended in both directions. The four picker-only keys it deliberately does not copy are
untouched — that verdict does not move with this card.

Two of the four assumptions the dispatch asked me to falsify came back changed, both
recorded above: the card's line numbers had moved (#7141 landed in
`packages/fields/src/widgets/`), and the read set is **larger** than the five keys named.

## Out of scope, filed not fixed

- **#7154** — the re-sweep found four more spec-declared keys read on this path and never
  copied: `multiple`, `allowCreate`, `lookupPageSize`, `dependsOn`. Each changes picker
  behaviour rather than a display string, so each wants its own rendering test. They are
  carried in the new table with verdict `deferred`, so the gate keeps them visible.
- **#7155** — four keys that were already in the copy set (`display_field`,
  `description_field`, `lookup_filters`, `id_field`) are also refused by the strict
  `FieldSchema`. They are kept here — retiring a shipped key is its own adjudication, and
  the chains read snake FIRST so removing a leg changes which value wins. Recorded as
  `legacy-alias` so the asymmetry is visible rather than implied.
- #6874 and #6711 were not touched.

## Verification

Run from the repo root with root-relative paths (#3378).

| what | result |
|---|---|
| `vitest run packages/plugin-grid/` | **103 files / 940 tests passed** |
| `vitest run packages/plugin-dashboard/` | **85 files / 795 tests passed** |
| `type-check` (both packages) | exit 0 — `Scope: 2 of 47`, both echoed their `tsc --noEmit` plus `tsc -p tsconfig.test.json`; `--listFiles` confirms all 3 new files are inside the test program |
| `lint` (both packages, plain `eslint .`) | exit 0 — 0 errors; the only findings on new files are `no-explicit-any` on the string-to-any `Record` signature carried over unchanged |
| `changeset:check` | `✅ All workspace packages are in the changeset fixed group.` · `✅ privatePackages declared: version=true, tag=false.` · `✅ No changeset declares a 'major' bump.` |
| `check-changeset-presence` | `✅ 6 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)` |
| `check-changeset-overwrite` | `✅ No pre-existing changeset was modified or deleted.` |
| `check:control-bytes` | `✅ OK (scanned 5926 tracked text file(s))` |
| `check:phantom-deps` | `✅ Every in-scope import is declared by the package that publishes it.` (`@objectstack/spec` was already a `plugin-grid` dep) |
| `check:self-import` · `check:spec-symbols` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` · `check:side-effects-array` · `check:element-data-source-declaration` · `check:entry-guard` · `check:esm-specifiers` | all exit 0 |

**Two readings are absent, not green:**

- `check:sdui-registration-pins` → **exit 2, `PREREQUISITE NOT MET`**: *"No console build to
  weigh at apps/console/dist/assets."* Not run here (a full console build on a shared box,
  for a gate this diff has no exposure to: the new module declares types, a frozen array and
  a pure function, with zero load-time registration, and `check:side-effects-array` — the
  instrument that judges load-time registration — walked the tree green). CI builds console
  and runs it.
- `check:readme-exports` → **exit 1, entirely `dist/index.d.ts is not on disk — run
  pnpm build first`** across 22 packages. After building `plugin-grid` and
  `plugin-dashboard`, **both drop to 0 findings**; the remaining 240 belong to 20 packages
  this diff does not touch and are the unbuilt closure, which CI resolves by building first.

### Ablation — the gate can fail, in the way that matters

Predicted red before running; mutation proven on disk by marker count **and** blob hash,
restore proven by state (`git diff HEAD`, `git diff --cached`, `git status --short` all
empty). No rebuild needed: `vitest.config.mts` aliases `@object-ui/fields` to
`packages/fields/src`, so these runs read source — which leg 2 also demonstrates.

1. **Delete the `displayField` entry from the table.** Marker 1 → 0, blob
   `b9b046c2` → `2c7724dc`. Result: `Tests 3 failed | 7 passed` — the gate's unclassified
   check, its copy-set exactness check, and the rendering test all go red together.
2. **A new spelling enters a consumer chain** — `fieldMeta?.zzNewSpelling` added to
   `LookupField.tsx`. Marker 0 → 1, blob `8e75d7ce` → `920b18c1`. Result:
   `Tests 1 failed | 7 passed`, message `expected [ 'zzNewSpelling' ] to deeply equal []`.

Leg 2 is the one that matters: it is the drift that produced this bug, and the gate now
catches it from the consumer side without anyone editing a list.

## Clause ② — not engaged

Nothing published moves. `RELATIONAL_META_KEYS`, `applyRelationalMeta` and the new
`relationalMetaKeys.ts` are internal to `plugin-grid` and are not re-exported from
`src/index.tsx` (named exports only, verified). `FieldMeta` and `CELL_RELATIONAL_META_KEYS`
live in `plugin-dashboard/src/recordFields.tsx`, which `src/index.tsx` does not re-export —
so the `FieldMeta` member is internal too. No parse accept/reject behaviour changes; the
strict `FieldSchema` is read, never modified.

The producer-side alternative was considered and rejected on the merits, not on budget.
`normalizeFieldReferenceKeys` in `@object-ui/core` is the repo's blessed choke-point
normaliser and extending it would have been a one-line fix — but it would run backwards
here. For `reference` it stamps the objectui-side spelling from the served one; for these
keys the spec spelling *is* the served one, so stamping `display_field` from `displayField`
would manufacture a twin the contract refuses. That is AGENTS.md #0.1's anti-pattern in its
worst direction — the renderer inventing a dialect the spec rejects — and it is a public
export's behaviour change on top. The renderer-side fix forwards what the def already
carries and invents no spelling; the dialect tolerance stays where it already was, in
`@object-ui/fields`, unchanged, and #7155 records the case for retiring it.

Session: https://claude.ai/code/session_012wwHa4aaFybxXrfmfHioDM

### DIFF
--- a/.changeset/6875-grid-relational-meta-derive.md
+++ b/.changeset/6875-grid-relational-meta-derive.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/plugin-grid': patch
+'@object-ui/plugin-dashboard': patch
+---
+
+A lookup cell in `ObjectGrid` now honours the author's `displayField`
+(objectui#6875).
+
+`ObjectGrid` copies a set of relational keys off the object-schema field def
+onto each column's `fieldMeta`, and that bag is what the lookup cell renderer
+and the inline picker receive. The set was hand-kept and had become a strict
+SUBSET of what those two consumers read — `displayField`, `descriptionField`
+and `lookupColumns` were read on the grid's own path and never copied.
+
+They are the spellings that matter. `@objectstack/spec` 17.2.0's `FieldSchema`
+is strict and declares `displayField` / `descriptionField` / `lookupColumns` /
+`lookupFilters` / `reference`, and none of the snake_case twins the copy set
+mostly carried — those parse to `unrecognized_keys`, so a spec-compliant
+producer cannot emit them. Nothing renames anything on the way in either: the
+adapter's `getObjectSchema` choke point rewrites only the `reference` ⇄
+`reference_to` pair. So an author who declared `displayField: 'project_code'`
+got a grid cell showing the referenced record's generic `.name` instead.
+
+- The copy set is now DERIVED, in `plugin-grid/src/relationalMetaKeys.ts`, from
+  a table that classifies every key the consumers read off this bag. A gate
+  re-extracts that read set from the consumer sources on each run and fails on
+  any unclassified spelling or orphan, so the two cannot drift apart again.
+- `reference_field` and `lookup_columns` — the other two never-copied keys —
+  stay out on purpose: `FieldSchema` declares neither, so no producer can fill
+  them. The gate proves that against the installed spec rather than asserting it
+  in prose.
+- `plugin-dashboard`'s `CELL_RELATIONAL_META_KEYS` had the same omission in the
+  same fallback chain and gains `displayField` too.
--- a/packages/plugin-dashboard/src/__tests__/lookupRelationalMeta-6694.test.tsx
+++ b/packages/plugin-dashboard/src/__tests__/lookupRelationalMeta-6694.test.tsx
@@ -260,6 +260,14 @@ describe('objectui#6694 — buildFieldMeta copies the cell-read relational keys 
     reference_to: 'project',
     reference: 'project',
     display_field: 'project_code',
+    // The SPEC spelling of the same pointer (objectui#6875). `FieldSchema`
+    // declares `displayField` and none of the snake twins, so this is the leg a
+    // live `getObjectSchema` actually serves — it must be copied.
+    displayField: 'project_code',
+    // The chain's third leg. Read by `LookupCellRenderer`, but `FieldSchema`
+    // refuses it with `unrecognized_keys`, so no producer can emit it and
+    // copying it would reach nothing (objectui#6711's reasoning). NOT copied.
+    reference_field: 'x',
     // Six keys with no reader on this path. FOUR of them the grid still copies
     // (its picker-only keys); the other two it has since retired as well —
     // `reference_to_field` (objectui#6711) and `titleFormat` (objectui#6874).
@@ -273,11 +281,14 @@ describe('objectui#6694 — buildFieldMeta copies the cell-read relational keys 
     titleFormat: '{project_code}',
   };
 
-  it('copies reference_to / reference / display_field', () => {
+  it('copies reference_to / reference / display_field / displayField', () => {
     const meta = buildFieldMeta({ accessorKey: 'project', label: 'Project', def }) as any;
     expect(meta.reference_to).toBe('project');
     expect(meta.reference).toBe('project');
     expect(meta.display_field).toBe('project_code');
+    // objectui#6875 — the spec-declared spelling, previously dropped here and in
+    // `ObjectGrid` at the same time.
+    expect(meta.displayField).toBe('project_code');
   });
 
   it('does NOT copy the picker-only keys', () => {
@@ -285,6 +296,9 @@ describe('objectui#6694 — buildFieldMeta copies the cell-read relational keys 
     for (const k of [
       'reference_to_field', 'id_field', 'description_field',
       'lookup_filters', 'lookupFilters', 'titleFormat',
+      // Read by `LookupCellRenderer`, but unproducible under the strict
+      // `FieldSchema` — objectui#6875 measured it and left it out on purpose.
+      'reference_field',
     ]) {
       expect(meta).not.toHaveProperty(k);
     }
@@ -294,7 +308,7 @@ describe('objectui#6694 — buildFieldMeta copies the cell-read relational keys 
     const meta = buildFieldMeta({
       accessorKey: 'amount', label: 'Amount', def: { type: 'currency' },
     }) as any;
-    for (const k of ['reference_to', 'reference', 'display_field']) {
+    for (const k of ['reference_to', 'reference', 'display_field', 'displayField']) {
       expect(meta).not.toHaveProperty(k);
     }
   });
--- a/packages/plugin-dashboard/src/recordFields.tsx
+++ b/packages/plugin-dashboard/src/recordFields.tsx
@@ -112,8 +112,27 @@ export const NUMERIC_FIELD_TYPES = new Set([
  * the module `getCellRenderer` dispatches into — the complete set of relational
  * keys read off a cell's `field` prop is:
  *
- *  - `reference_to`, `reference`, `display_field` — read by
+ *  - `reference_to`, `reference`, `display_field`, `displayField` — read by
  *    `LookupCellRenderer` itself. ✅ COPIED.
+ *
+ *    ⭐ `displayField` ARRIVED with objectui#6875. The enumeration above used
+ *    to name three keys, because it was written from the FIRST leg of each
+ *    chain rather than from the whole chain: `LookupCellRenderer` resolves the
+ *    display pointer as `display_field || displayField || reference_field`, and
+ *    the two extra spellings in that one chain were missed here and in the
+ *    grid's own list at the same time. `displayField` is the spelling
+ *    `@objectstack/spec` 17.2.0's strict `FieldSchema` DECLARES — so on a live
+ *    path served through `getObjectSchema` it is the only one that can arrive,
+ *    and a lookup cell here rendered the referenced record's generic `.name`
+ *    instead of the author's pointer. The grid's twin of this defect is pinned
+ *    behaviourally in `plugin-grid/src/__tests__/lookupDisplayFieldSpelling-6875.test.tsx`.
+ *
+ *  - `reference_field` — the chain's third leg, and still ⛔ NOT copied.
+ *    `FieldSchema` does not declare it (it parses to `unrecognized_keys`) and
+ *    the producer repo has zero occurrences of the identifier, against a
+ *    `displayField` control that hits 68 files. Copying it would write a member
+ *    from the def on every call that no producer can fill — objectui#6711's
+ *    reasoning, unchanged.
  *  - `id_field`, `description_field`, `lookup_filters`, `lookupFilters` — ZERO
  *    mentions in that module; read only by `fields/src/widgets/LookupField.tsx`
  *    and `UserField.tsx`, both EDITORS. ⛔ NOT copied.
@@ -136,7 +155,7 @@ export const NUMERIC_FIELD_TYPES = new Set([
  * picker keys. The boundary is pinned in
  * `__tests__/lookupRelationalMeta-6694.test.tsx`.
  */
-const CELL_RELATIONAL_META_KEYS = ['reference_to', 'reference', 'display_field'] as const;
+const CELL_RELATIONAL_META_KEYS = ['reference_to', 'reference', 'display_field', 'displayField'] as const;
 
 /**
  * Copy {@link CELL_RELATIONAL_META_KEYS} off a schema field def, with
@@ -228,6 +247,13 @@ export interface FieldMeta {
   reference?: string;
   /** Author-declared display field on the lookup — beats every resolver in the cell. */
   display_field?: string;
+  /**
+   * Same pointer, SPEC spelling (`FieldSchema.displayField`) — the second leg of
+   * `LookupCellRenderer`'s `display_field || displayField || reference_field`
+   * chain, and the only leg a spec-compliant producer can actually emit
+   * (objectui#6875).
+   */
+  displayField?: string;
 }
 
 /**
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -46,6 +46,7 @@ import { useColumnSummary } from './useColumnSummary';
 import { resolveRowCrudAffordances, resolveRowRecordCrudAffordance } from './rowCrudAffordances';
 import { useRecordCrudVerdicts } from './hooks/useRecordCrudVerdicts';
 import { resolveLegacyRowActions } from './resolveLegacyRowActions';
+import { applyRelationalMeta } from './relationalMetaKeys';
 import { resolveBulkActions } from './resolveBulkActions';
 import { partitionBulkRows } from './bulkEligibility';
 import { resolvesToDataColumn, describeUnresolvedColumns } from './columnSpellingDiagnostics';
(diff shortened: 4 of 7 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `a5a6df951`.

## The gating measurement came back opposite to the hypothesis — and the card got *bigger*

I made the card's own first step a ruling: *does a live `getObjectSchema` ever emit the camelCase spellings?* The card framed a "no" as shrinking the work. The answer inverts it:

- `@objectstack/spec@17.2.0`'s `FieldSchema` is a **strict** object (71 props — the card said 64) declaring `displayField` / `descriptionField` / `lookupColumns` / `lookupFilters` / `reference` and **none** of the snake twins; `reference_field` parses to `unrecognized_keys`.
- objectstack emits **zero** snake_case (`display_field` 3 files against a `displayField` control of 68; `reference_field` 0).
- `ObjectStackAdapter.getObjectSchema` — the choke point — rewrites **only** the `reference`/`reference_to` pair.

⇒ **camelCase is the only dialect a live path can carry, and it is exactly the half the grid never copied.** The defect is more reachable than the card supposed, not less.

## "Derive, do not restate" — delivered as a gate, not a longer list

I ruled that a hand-maintained list is not the fix, because a hand-maintained list is what drifted. The copy set now lives in `relationalMetaKeys.ts` as a classification table, with a gate that **re-extracts the read set from the consumer sources** and fails on any unclassified spelling or orphan.

Ablation **leg 2 is the load-bearing one**: adding `fieldMeta?.zzNewSpelling` to `LookupField.tsx` turns the gate red with `expected [ 'zzNewSpelling' ] to deeply equal []`. That is *the exact drift that produced this bug*, now mechanically caught. Predicted before running; mutation proven by marker count **and** blob hash; restore proven by state.

**Three of the five keys added, two deliberately not** — `FieldSchema` declares neither `reference_field` nor `lookup_columns`, so no producer can fill them, and the gate proves that absence against the installed spec rather than asserting it. A measured narrowing with its evidence attached.

## The user-visible bug was SHOWN, not asserted

I said "the key isn't copied" is a code fact and "the cell shows the wrong thing" is the bug. You built two columns over one data source and one referenced record, differing **only in spelling**: on the unmodified tree the control column resolved `ACME-42` and the spec-spelled column resolved `Wrong Name` — `1 failed | 1 passed (2)`, with an instrumented dump. After the fix both resolve correctly and `Wrong Name` is absent.

## ⚖️ The producer-side direction was rejected on the merits

This is the part I most wanted to see. I fenced against picking the cheap direction and presenting it as the design conclusion. You did not take the internal fix because it was affordable — you rejected the producer fix because extending `normalizeFieldReferenceKeys` would stamp `display_field` **from** `displayField`, **manufacturing the dialect the contract refuses**. That is AGENTS.md §0.1's anti-pattern running backwards, and it is a substantive reason, not a budget one.

Clause ② verified **not** engaged rather than assumed — neither `RELATIONAL_META_KEYS` / `applyRelationalMeta` / `relationalMetaKeys.ts` nor `FieldMeta` / `CELL_RELATIONAL_META_KEYS` are re-exported from their packages' indexes.

## A3.2 falsified: the read set is not the five named

Four more spec-declared keys are read on this grid's path and never copied — `multiple`, `allowCreate`, `lookupPageSize`, `dependsOn` — found *by the derivation itself*. Filed as objectui#7154, which unblocks when this lands.

**⚖️ Ruling on your open question: option A, keep it filed.** Your reasoning is right and it is the same standard this card just met: `multiple` flips the inline picker's cardinality and `dependsOn` has a closed prior card (objectui#2215) reporting the table picker bypassing the dependent filter. Folding four behaviour-changing keys in **without the rendering proof `displayField` just got** would ship exactly the class of change this card exists to fix. And the usual cost of deferring — that it gets forgotten — **is already neutralised**: the derivation carries them with verdict `deferred`, so the gate keeps them visible and they cannot lapse. That is what makes A safe rather than merely cautious.

**⚖️ On objectui#7155: agreed, C then likely B, and it is not yours or mine to make.** Labelled `needs-user-decision`. Your decisive argument is the ordering one — the consumer chains read the snake spelling **first**, so removing a leg changes which value *wins*, not merely which keys exist. That is a back-compat change on published behaviour, and the measurement available here bounds two repos while a host application outside them is invisible to it. Correct to file rather than act.

Landing will be verified by content with a live control once `main` advances.

---

PR #7157, opened 2026-09-01
docs(plugin-timeline): scope `spellGanttDateValue`'s totality claim to what is exercised

### LINKED ISSUE #7036: finding(plugin-timeline): `spellGanttDateValue`'s `Array.isArray` is the last non-total operation on the gantt date path — a revoked `Proxy` crashes it
Found while implementing #7027 (the gantt date brand gate). Not repaired there: #7027's charter is the type gate, and this is a different reflective operation in a different function, whose current form was a **deliberate** choice by #6907 rather than an oversight. Fixing it would be second-guessing a landed ruling, so it is recorded instead.

## What is wrong

`spellGanttDateValue` chooses between `an array` / `a function` / `an object` with `Array.isArray` and `typeof`. Its docblock argues — correctly — that these "read no author-controlled property", and it explicitly rejected `Object.prototype.toString.call` because that consults a `Symbol.toStringTag` getter that can throw.

`Array.isArray` reads no property, but it is still not total. On a **revoked** `Proxy` it throws:

```
const { proxy, revoke } = Proxy.revocable({}, {}); revoke();
Array.isArray(proxy)   -> THREW TypeError:
    Cannot perform 'IsArray' on a proxy that has been revoked
```

`IsArray` recurses into the proxy's `[[ProxyTarget]]`, and a revoked proxy has none. `typeof proxy` is still `'object'`, so nothing upstream separates it out.

## What #7027 does and does not change about it

Measured with a node probe replicating both functions' branch bodies verbatim (not measured in-render):

| tree | where a revoked `Proxy` crashes |
|---|---|
| `7fc5c3c12` (before #7027) | `isGanttDateType` — `instanceof` walks `[[GetPrototypeOf]]`, which throws on a revoked proxy |
| after #7027 | `spellGanttDateValue` — the new brand test refuses it cleanly, then `Array.isArray` throws while naming it |

So this is **not a regression** and #7027 does not introduce it: the value crashed the render before and crashes it after, at a different line. What changed is that it is now the *only* remaining throw site on this path, which is why it is worth a card of its own rather than staying invisible inside a larger one.

## Why this is p3-shaped, and the honest argument against fixing it

- **Not reachable from an authored document.** Same analysis #7027 verified: ObjectUI metadata is JSON, JSON cannot spell a proxy, and the repo has no `Proxy.revocable`, no `setPrototypeOf`, and no prototype-reconstructing deserializer in `packages/*/src` (swept on `7fc5c3c12`).
- **A revoked proxy has no honest diagnostic anyway.** Every operation on it throws by design; the best available answer is `an object`, which says nothing the path did not already say.
- **#6907 chose `Array.isArray` deliberately**, with a measured argument. The correct replacement shape is not settled by existing evidence — a `try`/`catch` around it, an ordering change, or "leave it and say so" are all defensible, and picking one is an adjudication.

## Suggested scope if picked up

`packages/plugin-timeline/src/renderer.tsx`, `spellGanttDateValue`'s trailing three branches. Either make the array test total, or state in the docblock that the totality claim excludes a revoked proxy and why that is acceptable — the same either/or #7027 was filed under. If it is taken, the row belongs in the existing pinned adversarial set (`packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx`), not in a new file.

⚠️ Worth reading with #7027's header: this is the **fifth** totality claim on this one code path, and the fourth to be falsified by the next card's measurement (#6759 → #6905 → #6907 → #7027 → this). The lesson that pattern keeps teaching is that totality asserted in prose is a hypothesis; the only thing that has settled it here is an exercised input set.

Refs: #7027 (the brand gate) · #6907 / PR #7026 (the spelling rule, which chose `Array.isArray`) · #6781 / PR #6905 (the type rule)

os-warren:
Fixes #7036

Took the card's **second** branch: the docblock overclaimed, so the claim is corrected and the exclusion is exercised. `Array.isArray` is left exactly as #6907 chose it. Measured on `dd35800af`; verified on `e615be490`.

## Why not "make the array test total"

Three reasons, none of them cost. A `try`/`catch` around `Array.isArray` would be four lines.

1. **Unreachable from an authored document.** ObjectUI metadata is JSON, and JSON cannot spell a proxy. Re-swept on `dd35800af`, each zero beside a live control on the same instrument: zero `Proxy.revocable` in the repo (control: 52 `Proxy` mentions in package sources), zero `Object.setPrototypeOf` calls in package sources (control: 19 `Object.assign` calls). The three non-comment `__proto__` hits are denylists that block it.
2. **The catch would be substitution, not a read.** `isDate`'s catch is argued in its own docblock as *the read* — the language exposes the `[[DateValue]]` bit only by throwing, so there is no other way to ask. A revoked proxy has no array-ness to read: catching would discard a failure and substitute `an object`. That is the consumer-side tolerance #6750 and #6759 both refused, and a second `catch` sitting beside `isDate`'s would erase the distinction this file is built on.
3. **It would buy no invariant** — see the next section.

## The card's headline is false as measured

#7036's title says `Array.isArray` is *the last non-total operation on the gantt date path*. It is not. It is the last one **inside `spellGanttDateValue`**. Measured in-render through the real `TimelineRenderer`, five further crash sites exist and every one runs **before** that helper is entered — in the property reads that fetch the date out of the document (`findUnusableGanttDate`'s `items[i]?.items` and `rowItems[j]?.[key]`; `calculateDateRange` repeats both):

| input | outcome |
|---|---|
| `items[0].items[0]` with a throwing `endDate` getter | THREW `Error: endDate getter throws` |
| `items[0].items[0]` is a revoked `Proxy` | THREW `TypeError: Cannot perform 'get' on a proxy that has been revoked` |
| `items[0]` is a revoked `Proxy` | THREW same |
| `items[0].items` is a revoked `Proxy` | THREW same |
| `items[0]` with a throwing `items` getter | THREW `Error: items getter throws` |
| CONTROL — an ordinary row/item | drew 1 bar |

Enumerated in #7153 (p3, same reachability class), not repaired here — different function, outside this card's ruled scope.

## What was measured, and how

The card measured with *"a node probe replicating both functions' branch bodies verbatim (not measured in-render)"*. This PR did not use a replica: every reading below comes from `render(<TimelineRenderer …>)` driving the real code.

```
endDate: <a revoked Proxy over {} / [] / a function / a real Date>
    -> THREW TypeError: Cannot perform 'IsArray' on a proxy
                        that has been revoked        (at Array.isArray)
schema.minDate: <a revoked Proxy>   -> THREW, same site
CONTROL  endDate: new Proxy({}, {}) -> NAMED "is an object"
CONTROL  endDate: ['2024-01-01']    -> NAMED "is an array"
CONTROL  endDate: new Date(...)     -> drew 1 bar
```

Two facts the card did not have: a revoked proxy over a **real `Date`** also lands here (`isDate` refuses it first, so it falls through to `Array.isArray`), and the **pinned `minDate` / `maxDate`** limb reaches the same site.

**Not a regression, confirmed.** On `7fc5c3c12` the value crashed at `isGanttDateType`'s `value instanceof Date` (`instanceof` walks `[[GetPrototypeOf]]`, which throws on a revoked proxy — measured), and that tree's `spellGanttDateValue` carried a second `instanceof` plus the same `Array.isArray`. #7027 did not introduce anything: it cut this input's throw sites from three to one.

## The change

- `packages/plugin-timeline/src/renderer.tsx` — docblock only, no executable line moved. The section heading no longer says every branch is total; the "everything else" bullet carries the exclusion inline; a `⚠️ #7036` note follows #7027's, naming the sentence it falsifies, the measurement, the three reasons, and the five upstream sites. The falsified sentences above it are kept verbatim — the file's own idiom (#7027 did the same), so the chain stays readable.
- `packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx` — `pin 4`, six rows in the **existing** set (no new file), plus a header entry for the fifth claim.

**No new confident totality sentence.** Everything the docblock now claims is explicitly bounded by the exercised input set, because prose has been the failure mode on this path four times running. That is also why the exclusion is pinned rather than merely written: a sentence would be a fifth claim of the same species.

## Tests

`pnpm exec vitest run packages/plugin-timeline/` on `e615be490` — **19 files, 226 passed** (220 before; `pin 4` adds 6). The two source blobs are byte-identical between the tested commit and the head above.

Rows assert the throw's **message**, never a bare `toThrow()`: a bare one passes for any error from any line, so it would stay green if the crash moved to `instanceof`, to `Object.prototype.toString`, or upstream into the row walk — which is exactly what has happened here four times. A live-proxy CONTROL row keeps the four throw rows from passing on a wholly broken gantt branch.

**Ablation.** Predicted before running: applying the repair this card declined (wrapping `Array.isArray` in `try`/`catch`) turns the 5 throw rows red and leaves the CONTROL green — 10 passed / 5 failed of 15. Observed exactly that. Mutation proven on disk by injected-marker count (0 → 1) and blob hash (`5acd179b…` → `bcd53d6a…`), never by an editor's exit code. No rebuild was needed: the suite imports `../renderer` by relative path, so it resolves to source, not through the package's `exports` into `dist/`. Restore proven by state — `git diff HEAD`, `git diff --cached` and `git status --short` all empty, the on-disk blob back to `5acd179b…`, marker count 0.

## Gates run locally (union, on `e615be490`)

`check-changeset-presence` · `check-changeset-no-major` · `check-changeset-fixed` · `check-control-bytes` · `check-lint-coverage` · `check-type-check-coverage` · `check-i18n-call-site-keys` · `check-i18n-en-drift` — all exit 0.

Changeset gate verdict, verbatim and not predicted:

```
✅  2 source file(s) of 1 released package(s) changed, and this change declares 1
    changeset(s): .changeset/spellgantt-revoked-proxy-exclusion.md.
    Every one of them has an EMPTY frontmatter — declared as releasing nothing,
    which is the explicit exemption and a complete answer to this gate.
```

`eslint .` in `packages/plugin-timeline` — exit 0, 0 errors (97 warnings, all pre-existing and none on added lines). `pnpm --filter @object-ui/plugin-timeline run type-check` — exit 0, and `tsc --listFiles` confirms both edited files are actually in the program (1369 files), so that green is a reading rather than an exclusion.

Repo-wide `pnpm lint` was **not** run locally; CI owns that farm.

---

### DIFF
--- a/.changeset/spellgantt-revoked-proxy-exclusion.md
+++ b/.changeset/spellgantt-revoked-proxy-exclusion.md
@@ -1,0 +1,11 @@
+---
+---
+
+Docs and tests only for `@object-ui/plugin-timeline`. `spellGanttDateValue`'s
+docblock claimed every branch was total; measured in-render, `Array.isArray`
+throws on a revoked `Proxy`, so the claim is now scoped to the input set that
+is actually exercised and the exclusion is pinned as rows in the existing
+adversarial set. The same measurement falsified the wider claim that
+`Array.isArray` is the last non-total operation on the gantt date path — five
+reads upstream of the helper throw first, recorded in objectui#7153. No
+published behaviour changes.
--- a/packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx
+++ b/packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx
@@ -30,6 +30,17 @@
  * about being a Date, each one asserted to reach a DEFINED outcome. That is
  * what #7026 gave the speller and what the gate did not have.
  *
+ *   5. objectui#7036: the gate's repair held, and the SPELLER still was not
+ *      total — `Array.isArray` throws on a revoked `Proxy`. That card did not
+ *      repair it; it ruled the exclusion STATED AND EXERCISED, which is what
+ *      `pin 4` at the foot of this file is. It also measured that the card's
+ *      own headline was wrong: `Array.isArray` is not the last non-total
+ *      operation on the PATH, only inside that function (objectui#7153).
+ *
+ * ⚠️ So this file's input set is not a claim of totality either. It is the
+ * boundary that has actually been exercised, and the sixth card will be the
+ * one that says where it ends.
+ *
  * ## What "a defined outcome" means here — two kinds, and never a third
  *
  * ⚠️ Not every row below is a refusal, and forcing them all to be one would
@@ -180,6 +191,17 @@ class HijackedGetTime extends Date {
     throw new Error('getTime hijacked by the authored document');
   }
 }
+
+/**
+ * A REVOKED proxy — the one input `spellGanttDateValue` is documented NOT to
+ * survive (objectui#7036). `Proxy.revocable` is the only way to spell it, and
+ * nothing in any package `src/` calls it: this is a test-only value.
+ */
+const revokedProxy = (target: object = {}) => {
+  const { proxy, revoke } = Proxy.revocable(target, {});
+  revoke();
+  return proxy as unknown;
+};
 
 /** An object whose `Symbol.toStringTag` is a throwing getter. */
 const throwingToStringTag = () =>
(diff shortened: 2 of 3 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `e615be490`.

## You falsified the card's headline, and that is what chose the route

I ruled "leave it and say so in the docblock" a full pass, but the interesting part is *why* it turned out to be the right branch rather than the cheap one.

**A3.4 is FALSE.** `Array.isArray` is the last non-total operation only **inside** `spellGanttDateValue`. **Five more sit upstream** — `findUnusableGanttDate`'s `items[i]?.items` and `rowItems[j]?.[key]`, repeated in `calculateDateRange` — and every one runs *before* the speller is entered. Filed as objectui#7153.

That demolishes the card's value proposition. It argued: *this is the last one, so repairing it closes the path.* With five upstream sites, a `try`/`catch` here would close **1 of 6** while leaving the docblock free to keep claiming totality — the worst of both, a change that looks like a fix and makes the documentation *more* wrong.

Two further reasons, both substantive rather than economical: the input is unreachable from an authored document (re-swept with live controls), and a catch here would **substitute** `an object` for a failure rather than read anything — the opposite of `isDate`'s catch, whose docblock argues the catch *is* the read because the language exposes `[[DateValue]]` only by throwing. That is the consumer-side tolerance objectui#6750 and objectui#6759 both refused.

## The ablation is an inversion, and it is the right instrument

You ablated **the repair you declined**. Wrapping `Array.isArray` in `try`/`catch` turns the 5 throw rows red and leaves the control green — `Tests 5 failed | 10 passed (15)`, predicted exactly before running. Showing that the rejected alternative *changes behaviour in five measurable places* is a far stronger argument for declining it than any amount of prose.

Mutation proven by injected-marker count **and** blob hash; restore proven by state.

## ⚠️ Two self-corrections I want on the record

**You caught your own broken annotation.** The ablation script's "deleted-text expect 0" counter stayed at 1, because the mutation *wraps* the line rather than deleting it. You named that the annotation was wrong and pointed at the evidence that actually fired (injected marker + hash change) instead of quietly letting a stale counter sit in a green report. That is the same class as the control I broke on objectui#7130 — an instrument reporting something it was not measuring — and you found yours.

**And you recorded a self-inflicted red.** A first docblock draft spelled a glob containing the two characters that terminate a block comment, closing it early and taking 18 suites down with a `PARSE_ERROR`. Found, fixed in both files, re-scanned, and the 226-green is from after it. Reporting a mistake you made and repaired costs nothing and makes the rest of the report believable.

## A2.1 falsified in the strengthening direction

The card measured with a **node probe replicating** the branch bodies. You drove the **real function** through `render(TimelineRenderer)` and found two facts the replica could not have shown: a **real-Date** target also lands at this site (`isDate` refuses it first, so it falls through), and the pinned **`minDate`/`maxDate` limb** is affected. A replica is not the code.

A3.3 confirmed — not a regression; objectui#7027 cut this input's throw sites from three to one. A3.5 survives, each zero beside a live control on the same instrument (0 `Proxy.revocable` against 52 `Proxy` mentions; 0 `Object.setPrototypeOf` against 19 `Object.assign`; the three `__proto__` hits are denylists that *block* it).

## Zone 2 honoured

I asked you not to write the sixth confident totality sentence on a path that has produced five claims and four falsifications. Every claim in the shipped docblock is explicitly bounded by the exercised input set, the falsified sentences are kept **verbatim above** the correction (the file's own idiom), and the six new rows went into the **existing** adversarial set rather than a new file.

Changeset verdict quoted, not predicted: **empty frontmatter**, the explicit exemption for a test-only change.

objectui#7153 unblocks when this lands.

---

PR #7158, opened 2026-09-01
test(plugin-dashboard): the legacy-retired rows say what they really are, and the fixture claim becomes a read

### LINKED ISSUE #7151: finding(plugin-dashboard): DashboardGridLayout.legacyRetired's "Byte-for-byte filtered-dashboard.json → widgets[0]" comment names a shape that fixture has never carried
Filed unassigned by the `os-dev` seat while implementing objectui#7102 (session `session_012wwHa4aaFybxXrfmfHioDM`). Grading and `domain:*` are the triage seat's. Deliberately NOT folded into objectui#7102's PR — that card's surface is `packages/components/src/__tests__/`, this is a different package and a different repair shape.

Surfaced by the census objectui#7102's dispatch asked for: *which other tests under `packages/` transcribe catalog fixtures inline?* This is the one other file whose transcription carries a claim that does not hold.

## The defect

`packages/plugin-dashboard/src/__tests__/DashboardGridLayout.legacyRetired.test.tsx:52-56` introduces one of its `it.each` rows like this:

```
// Byte-for-byte `examples/schema-catalog/src/schemas/plugin-dashboard/
// filtered-dashboard.json` -> `widgets[0]`: the real stored shape, not a
// fixture invented to match the detector.
['catalog bar', { id: 'invoices_by_status', title: 'Invoices by Status', type: 'bar', object: 'invoices', categoryField: 'status', aggregate: 'count' }],
```

and a second row four lines below:

```
// `widgets[2]` of the same catalog entry.
['metric', { id: 'w1', type: 'metric', object: 'invoices', aggregate: 'count' }],
```

**Neither claim holds.** Measured on `origin/main` `e8e4c4df5f51fc3a1d5ffc678881d3480f97a70e`, that fixture's `widgets[0]` is:

```json
{
  "id": "invoices_by_status",
  "title": "Invoices by Status",
  "type": "bar",
  "options": { "xField": "status", "yField": "count", "data": [ ... ] }
}
```

No `object`, no `categoryField`, no `aggregate` — it is the ADR-0021 `options` shape. `widgets[2]` is `{ id: "total_invoices", title, type: "metric", options: { value: "1,284" }, filterBindings: { region: false } }`, which shares only `type` with the row that names it.

## It was never true

`git log` on both paths returns exactly one commit: `ad0f5f11f` (`docs(plugin-grid): make the two catalog entries real object-grid nodes`, objectui#6028) **added both files in the same commit** — the fixture already in the `options` shape, the test already claiming byte-identity with a legacy shape. So this is not drift that a later migration introduced; the provenance claim was wrong when it was written.

⚠️ Bounded: this repo checkout is shallow (531 commits), so "exactly one commit" is a statement about the visible history. The present-tense measurement above does not depend on that.

## Why it matters, and why it is not just a comment nit

The comment is not decoration — it is the **stated justification** for why this particular literal is trustworthy: *"the real stored shape, not a fixture invented to match the detector."* That sentence is the test's whole epistemic claim. If the shape was in fact invented, the row proves only that the detector fires on a shape the author wrote for it, which is precisely the failure mode the comment exists to rule out.

Same class as objectui#7127 (a comment asserting something the code does not do) and objectui#7102 (a transcription that no longer matches what it names).

## Not user-visible, and nothing is red

`DashboardGridLayout.legacyRetired.test.tsx` passes today and will keep passing — the literal is cast through `as unknown as DashboardComponentSchema`, so no type or parse checks it, and the retired-format placeholder genuinely renders for that shape. The test's *behaviour* is fine. Only its provenance claim is false. Grade accordingly.

## Suggested repair — needs a choice, which is why this is filed rather than fixed

Three routes, and picking between them is a judgement the triage seat should make:

1. **Correct the comment** — drop the byte-for-byte claim, say plainly that the literal is a hand-written specimen of the pre-ADR-0021 shape. Cheapest, keeps the coverage, and is honest. But then the row loses the independent-corpus property the comment was claiming for it.
2. **Find a real corpus specimen** and transcribe *that* — if any catalog fixture still authors the legacy shape. ⚠️ objectui#4600 suggests these were all migrated, so this may have no candidate; that needs measuring before it is chosen.
3. **Derive instead of transcribe** — read the fixture at test time, the way `packages/types/src/__tests__/timeline-catalog-fixture-migrated.test.ts` already does for `gantt-style-timeline.json`. Kills the drift class rather than repairing one instance, but that file reads a fixture from `packages/types` specifically because *"this package's test tsconfig carries no node types"*, so the same move in `plugin-dashboard` needs that checked first.

Refs: objectui#7102 (sibling, the `packages/components` half) · objectui#7127 (same class) · objectui#4612 (the card this test was written for) · objectui#4600 (the catalog's dashboard examples and the pre-ADR-0021 shape) · objectui#6028 (`ad0f5f11f`, added both files).

os-warren:
Fixes #7151

Measured on `origin/main` **`dd35800af`**; shipped at **`81b5da034`**.

## The routing tree landed on route 3, and both gates were measurements

**Route 2 gate — is there a live corpus specimen of the pre-ADR-0021 shape?** No. Two independent instruments, each with a live non-zero control on the same instrument:

| instrument | subject | live control |
|---|---|---|
| structural walk of every `widgets[]` element in **all 628 tracked JSON files** | **0** widgets carry `object` / `categoryField` / `valueField` / `aggregate` / `rowField` / `columnField` at top level; **0** the shipped detector would flag | population **28** widgets; **15** `options`-shaped, **2** `dataset`-shaped |
| `git grep -c` over tracked `*.json` | `categoryField` **0**, `rowField` **0**, `columnField` **0**, `aggregate` **0** | `xField` **11**, `dataset` **2** |

The only two structural hits anywhere were `globalFilters[N].optionsFrom` nodes — a filter's option source, not a widget, and a live surface.

**A3.4 checked directly rather than inherited:** objectui#4600's "all the dashboard catalog examples were migrated" holds against the fixtures, measured, not taken from another card's summary.

**Route 1-vs-3 gate — can this package's test tsconfig read a fixture at test time?** Yes. `packages/plugin-dashboard/tsconfig.test.json` declares `"types": ["node"]`, and its own comment records that as load-bearing and measured. Independently confirmed after the fact: the test program loads 82 `@types/node` files. This is exactly the precondition `packages/types/src/__tests__/timeline-catalog-fixture-migrated.test.ts` could not meet in `plugin-timeline` and relocated for. **So: route 3.**

## The card's premise is half falsified — the claim was TRUE when written

The card's title says the fixture "has never carried" the shape, and its "It was never true" section rests on `git log` returning exactly one commit for both paths. That reading is an artifact.

`ad0f5f11f` **is the shallow clone's graft boundary** — the oldest commit present. Every file alive at the boundary reports as added there. Control: of 40 randomly sampled tracked files, **35 return exactly that same one commit**, and the boundary commit also reports adding `legacyRetiredWidget.ts`, which this suite's own header attributes to objectui#4612 — a lower number than the boundary commit's own PR. The instrument can distinguish (`package.json` returns 18 commits, `packages/types/src/complex.ts` returns 3); it simply has nothing to say about paths untouched since the boundary.

Recovered the real history from the public commit log without deepening the shared `.git` (zero-quota, no shared-state mutation), and it inverts the finding:

| commit | when | what |
|---|---|---|
| `8640cec19` | **2026-08-14T01:05:19Z** | added this suite, with the annotated literal |
| `e028dfcd8` | **2026-08-14T02:11:37Z** | migrated the `filtered-*` catalog entries off the retired shape (objectui#4600, PR objectui#4615) |

**66 minutes apart.** The fixture as it stood at `8640cec19` has `widgets[0]` equal to the test literal **byte-for-byte — same keys, same order, same values** (checked mechanically, with a self-comparison control returning true on the same comparator). So the row was a genuine independent corpus specimen; it was not "a fixture invented to match the detector". This is drift, and the migration that caused it is named in the card's own Refs.

The second row's comment made the looser claim "`widgets[2]` of the same catalog entry" and was never byte-for-byte; its retired binding was verbatim, with `id` genericised to `w1` and `title` / `filterBindings` dropped. Substantially true then, false now.

Both are false against today's fixture — that part of the card stands, mechanically confirmed key-for-key.

## What changed

Both annotations now state the measured history and say **explicitly that the rows no longer carry the independent-corpus property** — they are historical specimens of stored metadata, which is still the right thing to pin, but is weaker evidence and now says so rather than leaving the next reader to re-derive the confidence.

Route 3's half: the relationship stops being prose. A derived block reads the same fixture at test time and

- pins that **no** shipped catalog widget authors the retired binding — the executable form of the sentence that used to be a false comment, and a catalog-regression guard in the other direction;
- runs the **three real shipped widgets** through the surface as corpus-derived negative controls. That restores the independent-corpus property on the side where the corpus still has specimens.

The property cannot be restored for the legacy rows themselves, and the block says why: the zero is by design. The retirement's whole content is that no authoring surface emits the shape — `WidgetConfigPanel` scrubs it on save via `LEGACY_ANALYTICS_KEYS` — and the catalog is an authoring corpus.

## A3.3 sweep — no third false claim

Swept every remaining provenance claim in the file and checked each mechanically. All hold:

- `object` / `categoryField` / `aggregate` are not on `DashboardWidgetSchema` — enumerated the spec Zod schema's 22 top-level keys plus objectui's own additions; controls `dataset` / `dimensions` / `values` / `options` all present.
- `DashboardGridLayout` is separately exported and registered as `dashboard-grid`.
- One detector module, two consumers; one shared placeholder constant, one wording.
- Pivot's legacy spelling really is `rowField` / `columnField` — both in `LEGACY_ANALYTICS_KEYS`.
- `DashboardRenderer` really does return the placeholder family-wide for pivot, and this surface really does have a pivot renderer.
- The `provider: 'object'` config really is read off `widgetData`, never the widget top level.

## Tests, at `81b5da034`

Whole package, from the repo root: **85 files, 800 tests, all passing.** This suite alone went 9 passing to **14**.

Ablation, direction predicted before running, on the committed tree:

|---|---|---|---|
| A | add top-level `object` to `widgets[0]`, keep `options.data` | shape pin red, render control green | **1 failed / 13 passed** — only the shape pin |
| B | revert `widgets[0]` to its `8640cec19` retired form | both red | **2 failed / 12 passed** — shape pin and the derived `widgets[0]` control |

Mutation proved on disk by blob hash change plus injected-marker count, never an editor's exit code; restore proved by state — blob hash back to the HEAD blob and `git diff HEAD`, `git diff --cached`, `git status --short` all empty; absolute paths and an `EXIT INT TERM` trap throughout. The restored tree re-measured at 14 passing. One caveat reported rather than quoted as evidence: leg B's "removed marker" check used `xField`, which is not unique to `widgets[0]`, so it carries no information; the hash and the injected-marker count are what prove that leg.

So this is **not** the objectui#7102 situation where no available pin could fail — both new pins are independently falsifiable, shown two different ways.

Gates: `check:control-bytes`, `check:phantom-deps` (green on the new `node:` builtin imports), `check:esm-specifiers`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `type-check:coverage` all exit 0. `pnpm --filter @object-ui/plugin-dashboard type-check` exits 0, and `--listFiles` confirms the edited file is genuinely in the test program rather than silently excluded. `eslint .` for the package: 0 errors, and 0 findings on the edited file (control: sibling files do appear).

Changeset verdict quoted rather than predicted — empty frontmatter, per the gate's own named exemption:

> 1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/7151-legacy-retired-provenance.md.
> Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.

## Scope

One test file plus one changeset. No published type, export, or parse behaviour is touched, so the contract-review tier is not engaged. Nothing under `content/docs/releases/`, and the schema-catalog fixture is unmodified in the diff — it was read, and temporarily mutated only inside the ablation, which restored it provably.

Stayed clear of the objectui#6875 lane: no relational-meta file and no `CELL_RELATIONAL_META_KEYS`.

---

### DIFF
--- a/.changeset/7151-legacy-retired-provenance.md
+++ b/.changeset/7151-legacy-retired-provenance.md
@@ -1,0 +1,6 @@
+---
+---
+
+Test-only change to `plugin-dashboard`'s legacy-retired suite; no published behaviour changes.
+
+The two annotated `it.each` rows in `DashboardGridLayout.legacyRetired.test.tsx` introduced their literals with provenance comments naming a schema-catalog fixture. Both were true when written and were invalidated 66 minutes later by the catalog migration off the retired shape (objectui#4600). The annotations now state the measured history and say plainly that the rows no longer carry the independent-corpus property, and the relationship they assert is derived at test time from the fixture instead of restated in prose.
--- a/packages/plugin-dashboard/src/__tests__/DashboardGridLayout.legacyRetired.test.tsx
+++ b/packages/plugin-dashboard/src/__tests__/DashboardGridLayout.legacyRetired.test.tsx
@@ -31,6 +31,9 @@
  */
 
 import { describe, it, expect, afterEach } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import React from 'react';
 import { render, screen, cleanup } from '@testing-library/react';
 import type { DashboardComponentSchema } from '@object-ui/types';
@@ -46,16 +49,46 @@ afterEach(cleanup);
 const dash = (widget: Record<string, unknown>): DashboardComponentSchema =>
   ({ type: 'dashboard', widgets: [widget] }) as unknown as DashboardComponentSchema;
 
+/**
+ * The catalog entry the two annotated rows below were transcribed from, named
+ * once so the prose and the derived block at the bottom cannot drift apart.
+ */
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..', '..', '..');
+const CATALOG_FIXTURE = path.join(
+  REPO_ROOT,
+  'examples/schema-catalog/src/schemas/plugin-dashboard/filtered-dashboard.json',
+);
+const catalogWidgets = (
+  JSON.parse(fs.readFileSync(CATALOG_FIXTURE, 'utf8')) as { widgets: Record<string, unknown>[] }
+).widgets;
+
 describe('DashboardGridLayout retired legacy widgets (#4612)', () => {
   it.each([
     ['chart', { id: 'w1', type: 'bar', object: 'invoices', categoryField: 'month', valueField: 'amount', aggregate: 'sum' }],
-    // Byte-for-byte `examples/schema-catalog/src/schemas/plugin-dashboard/
-    // filtered-dashboard.json` → `widgets[0]`: the real stored shape, not a
-    // fixture invented to match the detector.
+    // Transcribed from `filtered-dashboard.json` → `widgets[0]` as that fixture
+    // stood at `8640cec19`, the commit that added this file. It was byte-for-byte
+    // then — same keys, same order, same values — and it is NOT byte-for-byte
+    // now: `e028dfcd8` (objectui#4600, PR #4615) migrated the `filtered-*`
+    // entries off the retired shape 66 minutes later, so `widgets[0]` on disk is
+    // `{ id, title, type, options: { xField, yField, data } }` today.
+    //
+    // ⚠️ This row therefore does NOT carry the independent-corpus property the
+    // original annotation claimed for it. It is a HISTORICAL specimen of stored
+    // metadata, not a live transcription — which is the right thing to pin (a
+    // renderer cannot refuse to receive stored metadata) but is worth strictly
+    // less as evidence, so it is stated rather than left to be re-derived. No
+    // live specimen can replace it; the derived block at the bottom measures
+    // that, instead of asserting it here in prose the way this comment once did.
     ['catalog bar', { id: 'invoices_by_status', title: 'Invoices by Status', type: 'bar', object: 'invoices', categoryField: 'status', aggregate: 'count' }],
     // The pivot family, whose legacy spelling names `rowField`/`columnField`.
     ['pivot', { id: 'w1', type: 'pivot', object: 'invoices', rowField: 'region', valueField: 'amount' }],
-    // `widgets[2]` of the same catalog entry.
+    // The same fixture's `widgets[2]` at that same commit, and a LOOSER claim
+    // than the row above ever made: its retired binding verbatim (`type`,
+    // `object`, `aggregate`), with `id` genericised to `w1` and `title` /
+    // `filterBindings` dropped. So this row was never byte-for-byte and never
+    // said it was. `e028dfcd8` retired that widget's shape too — `widgets[2]` is
+    // `{ id, title, type, options: { value }, filterBindings }` today — so the
+    // same ⚠️ above applies to it.
     ['metric', { id: 'w1', type: 'metric', object: 'invoices', aggregate: 'count' }],
   ])('renders the visible placeholder for a legacy %s widget', (_kind, widget) => {
     render(<DashboardGridLayout schema={dash(widget)} />);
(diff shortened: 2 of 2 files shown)


os-warren:
## ✅ Reviewed and armed

All **30** check runs in one call at `perPage=100`: **27 success, 3 skipped, 0 failed, 0 `in_progress`.** Undrafted, auto-merge SQUASH enabled at `81b5da034`.

## ⚠️ This PR inverted its own card — and my comment on it

`premise_still_valid: false`, and it is the most valuable result of the session.

objectui#7151 claimed the test's provenance comment *"was never true"*, on `git log` showing fixture and test added in one commit (`ad0f5f11f`). The card correctly bounded that to **visible** history on a shallow clone. **I read the bound and reasoned past it anyway**, writing on the card: *"so this was never drift — the claim was wrong when it was written"*, and calling it the sharp part. That was wrong, and I have posted the correction on objectui#7151.

You tested the instrument instead of trusting it:

- **35 of 40** randomly sampled tracked files report that same commit as their "add." `ad0f5f11f` is the **graft boundary**, reporting every file alive at the cut as newly added.
- The boundary commit also reports adding `legacyRetiredWidget.ts`, which **this suite's own header attributes to objectui#4612** — a *lower* number than the boundary commit's own PR. The contradiction was inside the reading, for anyone who looked.

And the real history reverses the verdict:

| commit | time | what |
|---|---|---|
| `8640cec19` | 01:05:19Z | added the suite |
| `e028dfcd8` | 02:11:37Z | migrated the `filtered-*` entries off the retired shape |

**66 minutes apart**, with the fixture at the earlier sha **byte-for-byte** equal to the test literal (checked mechanically, with a self-comparison control). ⇒ the row **was** a genuine independent corpus specimen, **not** invented to match the detector. This is **drift** — caused by a migration named in the card's own Refs.

Recovering that from the public commit log without deepening the shared `.git` is the right move in this environment, and it is the only reason the truth was reachable at all.

The **other** half of the card stands, confirmed key-for-key: both comments are false against today's fixture. That is why the repair was still needed.

## Route 3 selected by both gates, not by preference

**Gate 1 (route 2 — is there a live corpus specimen?)** — a structural walk of every `widgets[]` element across **628 tracked JSON files**: **0** carrying the retired keys and **0** the shipped detector would flag, against live controls of 28 widgets in the population, 15 options-shaped, 2 dataset-shaped. An independent git-grep instrument agreed. The only structural hits anywhere were `globalFilters[N].optionsFrom` nodes — a filter option source, not a widget. So objectui#4600's summary is **confirmed by direct measurement**, not inherited.

**Gate 2 (route 1 vs 3)** — `plugin-dashboard/tsconfig.test.json` declares `types: [node]`, the precondition the `packages/types` precedent could not meet and relocated for; confirmed afterwards by 82 `@types/node` files in the program. So route 3 was *available*, which is what made it correct rather than aspirational.

## The honest limit, stated rather than papered over

The independent-corpus property **cannot** be restored for the legacy rows, and the reason is structural: the zero is **by design**, because the retirement's whole content is that no authoring surface emits the shape (`WidgetConfigPanel` scrubs it via `LEGACY_ANALYTICS_KEYS`) and the catalog is an authoring corpus. So the derived block restores the property on the side where the corpus still *has* specimens, and the annotations say plainly that the legacy rows no longer carry it — which is exactly what I ruled route 1 would have had to do, arrived at from the other direction.

**A3.3: no third false claim.** Every remaining provenance claim in the file was checked mechanically and all hold, with the neighbourhood swept too — the sibling `DashboardRenderer.legacyRetired.test.tsx` carries no provenance claim, and app-shell's "VERBATIM transcription" comments are the well-behaved version of the pattern (they state the claim must be re-earned, and pin it).

## Evidence

Two ablation legs distinguishing the two new pins, direction predicted before running: leg A `1 failed | 13 passed (14)`, leg B `2 failed | 12 passed (14)`, both failures named. Not green-always pins — the objectui#7102 trap, avoided by demonstration.

**⚠️ And you volunteered that one of your own checks was worthless**: leg B's "removed marker" used `xField`, which is not unique to `widgets[0]`, so it returned 1 and **carries no information** — the blob-hash change and injected-marker count are what prove that leg. Naming a check that did not measure what it claimed, inside an otherwise green report, is the same discipline that caught the graft boundary. It is also the second self-caught broken instrument in this round (objectui#7157's dev found theirs too), and I broke one myself earlier today.

---

PR #7161, opened 2026-09-01
fix(plugin-charts): a sankey that drew only some of its rows says how many

### LINKED ISSUE #7148: finding(plugin-charts): a sankey with mixed positive and negative rows silently DROPS the negative ones and draws a partial flow as if it were complete
Filed unassigned by the implementer of #7140 (PR #7146). ⛔ Out of scope there: that card's trigger is `links.length === 0` — *no* row survives the filter — and PR #7146 answers exactly that branch. This is the branch next door, where **some** rows survive, and it is arguably the worse of the two.

## The measurement

`packages/plugin-charts/src/AdvancedChartImpl.tsx`, the sankey arm, at `origin/main` **e8e4c4df5**:

```
const rows = data.filter((r) => (Number(r?.[dataKey]) || 0) > 0);
const nodes = [{ name: rootName }, ...rows.map(...)];
const links = rows.map(...);
```

The filter is unconditional. A sankey handed

```
[{ stage: 'New business', amount: 40 },
 { stage: 'Refunds',      amount: -25 },
 { stage: 'Chargebacks',  amount: -12 }]
```

draws **one** flow — "New business" — and nothing anywhere says the other two rows existed. Measured in Chromium in the same sweep: the one-positive-among-zeros tile rendered `svg: 1`, `path: 3`, 18 descendants, and no diagnostic of any kind. It is a normal, healthy-looking, confident sankey of a dataset it is showing a third of.

## Why it is worth recording separately

#7140's blank is at least *visibly* wrong — nothing is on screen, and after PR #7146 it says so. This one is invisibly wrong: the chart looks correct, so nobody re-reads the dataset. A pipeline chart that quietly omits every refund/credit/chargeback row is a business-wrong picture presented as a complete one, and the same reasoning behind the empty-state work applies with more force, not less: the reader cannot tell a partial flow from a whole one.

Note the filter is also what folds `null` and unparseable measures to `0` (`Number(...) || 0`), so those rows disappear the same way in a mixed dataset.

## What this card does NOT assert

It does not assert the drop is wrong. A flow has no negative width, so a sankey genuinely cannot draw those rows; discarding them may be the only thing it *can* do. What is undecided is whether it must **say** it did — and if so, whether that is a footnote beside a drawn chart, a count of omitted rows, or a refusal. That is a judgement this card does not pre-make, and it needs a look at a real dataset before anyone writes code. Note also that PR #7146 pins the "one positive among zeros still draws" boundary, so any answer here must keep a drawable sankey drawable.

## Related

- **#7140** / PR #7146 — the all-zero-or-negative case, where the same filter keeps *nothing*. Fixed there; this branch is untouched by it.
- **#7147** — funnel and pie render a blank tile for all-zero rows, found in the same sweep.

os-warren:
Fixes #7148

The sankey arm keeps only strictly positive measures, so a mixed dataset drew a
normal, healthy, confident chart of a fraction of itself and nothing anywhere
recorded that the other rows existed. This adds one line of prose under the plot
saying how many rows it drew. The chart itself is unchanged.

## The measurement that decided fix-over-decline

The card deliberately does not assert the drop is wrong, so this was settled the
way #7140 was: 27 tiles rendered in real Chromium (`/opt/pw-browsers/chromium`)
at `origin/main` **fd11e1644**, each screenshotted and SHA-256'd.

At base, six different datasets produced **one byte-identical image**
(`13237e6e19a7072a`):

| tile | dataset | rows drawn |
|---|---|---|
| `sankey-card-mixed-sign` | `New business 40 / Refunds -25 / Chargebacks -12` | 1 of 3 |
| `sankey-pin-one-pos-among-zeros` | `0 / 7 / 0` | 1 of 3 |
| `sankey-pos-plus-nulls` | `40 / null / null` | 1 of 3 |
| `sankey-pos-plus-unparseable` | `40 / 'n/a'` | 1 of 2 |
| `sankey-pos-plus-missing-key` | `40 / (no measure key)` | 1 of 2 |
| `sankey-genuine-1row-40` | `40` | **1 of 1** |

That last row is the finding. A chart showing a third of its dataset was
pixel-for-pixel the chart of a dataset that genuinely had one row — `svg: 1`,
`path: 3`, 18 descendants, no `role`, no text, and **zero console output**,
measured against a live console control on the same instrument
(`missing-category-key`, which did fire). A reader had no bit of information
separating a complete flow from a partial one. `measured-and-declined` was
available and is not what the evidence supports.

## What ships, and why this shape

- **Not a refusal.** #7146 pins *"one positive row among zeros still draws"* —
  and that fixture (`0 / 7 / 0`) is itself a **thinned** dataset. It lands in
  this branch and hashed identical to the mixed-sign tile. Refusing on a thinned
  flow would blank the chart that pin requires drawn. It now draws *and* says so.
- **Not a change to the drop.** A flow has no negative width, so discarding is
  the only thing this arm can do with those rows. The silence was the defect.
- **Names the predicate, not a cause.** `Number(…) || 0` folds negatives, zeros,
  `null`, unparseable strings and a missing key into one discard, and all five
  were measured reaching this branch beside a survivor. Naming any one of them is
  a sentence that is false for the other four — the same doctrine the
  `no-positive-flow` copy follows.
- **Carries the count.** "Some rows were dropped" still leaves a thinned flow
  indistinguishable from a complete one; the ratio is the bit the picture cannot
  carry.

> Showing 1 of 3 rows — 2 rows have no `amount` above zero, which a flow cannot draw.

No console warning, matching the sibling refusal: the sentence already names the
key, the test it failed and how many rows failed it.

### The footnote could not be a plain sibling — that was measured too

A bare paragraph element rendered after `ChartContainer` is **invisible in dashboards**. In the
dashboard shape — a fixed, clipping card whose chart carries `h-full` — the
container takes the card's full height and the note landed at `y=327` in a box
ending at `y=316`, entirely outside the clip. A wrapper div carrying `className`
is equally unavailable: the consumer's className is the chart's height contract
and has to keep reaching the element Recharts measures (see `CHART_MIN_HEIGHT`
in `ChartContainerImpl` for what a broken chain costs — a permanently zero box
and an invisible chart, with no refusal and no empty state).

So the plot keeps its own `className` and the note hangs off a flex chain, in
`ChartFootnote` — deliberately the mirror image of this file's existing
`ChartFrame`, including its gate: with nothing to say, the chart is returned
untouched and **gains no wrapper element**.

## Reverse verification (controlled)

The implementation was committed first, then the pre-fix file was restored onto
disk (mutation proven by blob hash `3105af33…` matching the base blob, marker
count `omitted-rows` 0, and the marker confirmed absent from the rebuilt bundle),
the harness rebuilt with **identical CSS**, and all 31 tiles re-measured.
Restore proven by state: `git diff HEAD`, `git diff --cached` and
`git status --short` all clean, on-disk blob back to `5e2486b8…`.

- **8 tiles changed. All 8 are thinned sankeys.**
- **23 tiles byte-identical**, including `sankey-control-all-positive`,
  `chain-clip-all-positive`, `chain-auto-all-positive` (the complete-flow path is
  unchanged in all three height chains), both `no-positive-flow` refusal tiles,
  the no-rows tile, and every other chart family.
- The collision with `sankey-genuine-1row-40` is broken: that tile keeps
  `13237e6e19a7072a` (nothing was omitted) while the five thinned tiles move away.

Residual collisions are correct: tiles sharing a ratio share a sentence, because
the copy deliberately does not distinguish *why* a row was dropped.

## The seam with #7146

`rows.length` is the whole boundary, and it is now pinned from both sides:

| dataset | branch | measured |
|---|---|---|
| no rows at all | bare div — the empty-**result** question (#7130), answered upstream | `svg: 0`, `textContent: ''` |
| all zeros + one negative | `no-positive-flow` refusal (#7146) | `data-chart-error="no-positive-flow"` |
| one positive + nulls | **this note** | `svg: 1`, note present |
| `0 / 7 / 0` (#7146's own pin) | **this note** — and still draws | `svg: 1`, note present |
| all positive | untouched | byte-identical to base |

## Gates

Run at `f2df89144` (the head this PR points at), repo root:

- `pnpm vitest run packages/plugin-charts/` — **38 files, 265 tests passed**,
  including #7146's `sankeyNoPositiveFlow` pin unmodified.
- `pnpm --filter @object-ui/plugin-charts type-check` — exit 0. `--listFiles`
  confirms the new test file is inside the program (not an excluded-tests
  false green).
- `pnpm --filter @object-ui/plugin-charts lint` — 51 files judged, **0 errors**,
  282 warnings (`--max-warnings` is deliberately unset repo-wide, per
  `lint.yml`). The 6 warnings on the new test file are the `as any` /
  `importActual` type-argument shape every sibling test in this package already uses.
- `pnpm check:control-bytes` — `✅ OK (scanned 5931 tracked text file(s))`
- `pnpm check:vi-mock-specifiers`, `check:vi-mock-inherit` — both `✅ OK`
- `pnpm check:i18n-keys` — `✅` (no i18n call site added; this file is
  hardcoded English throughout and both existing refusals are too)
- `pnpm changeset:check` — `✅ All workspace packages are in the changeset fixed
  group.` / `✅ No changeset declares a major bump.`
- `pnpm check:sdui-registration-pins` — first returned **exit 2, "No console
  build to weigh"**, which is NOT MEASURED rather than a pass. Built the console
  and its closure and re-ran: `✅ All 16 registration(s) … present in the built
  console (517 chunks weighed)`.

Repo-wide `pnpm lint` is CI's run. The narrowing above is measured, not assumed:
the universe (51 files) is eslint's own config resolution, the counts come from
`--format json`, and `eslint.config.js` configures **no** type-aware linting
(no `parserOptions.project` / `projectService`), so a change under
`packages/plugin-charts/src/` cannot move the verdict on any untouched file.

## Out of scope, reported not fixed

Measured in the same sweep, for **#7147**, which owns that family:
`treemap` drew 2 rects / 2 labels for the mixed-sign dataset against a control's
4 / 4; `funnel` with a `null` measure drew **0** segment paths against a
control's 2; `pie` drew 1 slice against a control's 2. Those rows are never
filtered — they stay in the data and the layout gives them no area — so this is
degenerate geometry, a different mechanism from the row drop fixed here, and
untouched by this PR. It widens #7147 beyond all-zero datasets to mixed ones.

`AdvancedChartImpl.tsx` line 1047 is the only row-dropping `filter` in the file
(the other two filter *series*), which is why the note cannot reach another
family — pinned in the tests.

---

### DIFF
--- a/.changeset/7148-sankey-omitted-rows-note.md
+++ b/.changeset/7148-sankey-omitted-rows-note.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/plugin-charts': patch
+---
+
+A sankey that drew only SOME of its rows now says how many (objectui#7148).
+
+The sankey arm keeps strictly positive measures
+(`data.filter((r) => (Number(r?.[dataKey]) || 0) > 0)`), so a mixed dataset
+drew a normal, healthy, confident chart of a fraction of itself and nothing
+anywhere recorded that the other rows existed. Measured in Chromium across 27
+tiles: `[{New business: 40}, {Refunds: -25}, {Chargebacks: -12}]` rendered
+`svg: 1`, `path: 3`, 18 descendants, no `role`, no text, and — against a live
+console control that did fire on the same instrument — zero console output.
+Its screenshot hashed byte-identical to five other datasets, one of which
+genuinely had a single row. Six datasets, one image: a reader had no bit of
+information separating a complete flow from a third of one.
+
+The discard itself stands — a flow has no negative width, so it is the only
+thing that arm can do with those rows. What is added is a footnote under the
+plot naming the ratio and the predicate the filter applies:
+
+> Showing 1 of 3 rows — 2 rows have no `amount` above zero, which a flow
+> cannot draw.
+
+It names the predicate rather than a cause because `Number(…) || 0` folds
+negatives, zeros, `null`, unparseable strings and a missing key into one
+discard, and all five were measured reaching this branch beside a survivor;
+naming any one of them is a sentence that is false for the other four.
+
+A complete flow is byte-for-byte unchanged and gains no wrapper element, and a
+drawable sankey is never replaced by prose: the `no-positive-flow` refusal
+still owns the case where NOTHING survives the filter, and the "one positive
+among zeros still draws" boundary still draws — that fixture is itself a
+thinned dataset, so it now draws *and* says so.
--- a/packages/plugin-charts/src/AdvancedChartImpl.sankeyOmittedRows.test.tsx
+++ b/packages/plugin-charts/src/AdvancedChartImpl.sankeyOmittedRows.test.tsx
@@ -1,0 +1,222 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * Sankey — SOME rows survive the positive filter and the rest are dropped
+ * silently (objectui#7148). The branch next door to objectui#7140's refusal,
+ * and the more dangerous of the two.
+ *
+ * The sankey arm keeps only strictly positive measures
+ * (`data.filter((r) => (Number(r?.[dataKey]) || 0) > 0)`). When that filter
+ * keeps NOTHING, objectui#7146 now says so. When it keeps SOME, the chart drew
+ * a normal, healthy, confident sankey of a fraction of its dataset and nothing
+ * anywhere recorded that the other rows existed.
+ *
+ * ## The measurement that decided fix-over-decline
+ *
+ * 27 tiles in real Chromium (`/opt/pw-browsers/chromium`) at `origin/main`
+ * fd11e1644, each tile screenshotted and SHA-256'd. The card's own dataset
+ * (`New business 40 / Refunds -25 / Chargebacks -12`) rendered `svg: 1`,
+ * `path: 3`, 18 descendants, no `role`, no text — and, against a live console
+ * control that DID fire on the same instrument (`missing-category-key`), zero
+ * console output. Its screenshot hashed `13237e6e19a7072a`, byte-identical to
+ * FIVE other tiles including a genuinely one-row dataset
+ * `[{ stage: 'New business', amount: 40 }]`. Six datasets, one image: the
+ * reader had no bit of information separating a complete flow from a third of
+ * one. That is why "discarding is all a flow CAN do" does not settle the card —
+ * the drop is fine, the silence is not.
+ *
+ * ## Why a note and not a refusal
+ *
+ * objectui#7146 pins "one positive row among zeros still draws". That fixture
+ * (`0 / 7 / 0`) is itself a THINNED dataset — it lands in this branch and
+ * hashed identical to the mixed-sign tile — so a refusal here would blank a
+ * chart that pin requires drawn. The two cards meet exactly at
+ * `rows.length === 0`: none survive → refusal (objectui#7146); some survive and
+ * some do not → this note; all survive → untouched.
+ */
+import React from 'react';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+
+vi.mock('recharts', async () => {
+  const actual = await vi.importActual<any>('recharts');
+  return {
+    ...actual,
+    ResponsiveContainer: ({ children }: any) =>
+      React.cloneElement(children, { width: 480, height: 320 }),
+  };
+});
+
+import AdvancedChartImpl from './AdvancedChartImpl';
+
+afterEach(cleanup);
+
+const SERIES = [{ dataKey: 'amount', label: 'Amount' }];
+
+const renderSankey = (data: Array<Record<string, unknown>>) =>
+  render(
+    <AdvancedChartImpl
+      chartType="sankey"
+      xAxisKey="stage"
+      series={SERIES as any}
+      data={data as any}
+    />,
+  );
+
+const noteOf = (container: HTMLElement) =>
+  container.querySelector('[data-chart-note="omitted-rows"]');
+const refusalOf = (container: HTMLElement) =>
+  container.querySelector('[data-chart-error="no-positive-flow"]');
+
+/**
+ * Every shape `Number(…) || 0` folds to zero, each one BESIDE a survivor.
+ *
+ * They are not the same situation and the copy deliberately names none of them:
+ * a message that said "negative" would be false of the null row, and the card
+ * itself notes that `null` and unparseable measures vanish through the very
+ * same filter. All five were measured reaching this branch in Chromium.
+ */
+const THINNED: Array<[string, Array<Record<string, unknown>>, number, number]> = [
+  ['negative rows (the card\'s own dataset)', [
+    { stage: 'New business', amount: 40 },
+    { stage: 'Refunds', amount: -25 },
+    { stage: 'Chargebacks', amount: -12 },
+  ], 1, 3],
+  ['zero rows (objectui#7146\'s pinned boundary)', [
+    { stage: 'Prospecting', amount: 0 },
+    { stage: 'Proposal', amount: 7 },
+    { stage: 'Won', amount: 0 },
+  ], 1, 3],
+  ['null measures', [
+    { stage: 'New business', amount: 40 },
+    { stage: 'Refunds', amount: null },
+    { stage: 'Credits', amount: null },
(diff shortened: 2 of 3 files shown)


os-warren:

Review posted now rather than at arm time because the measurement here deserves recording on its own. CI still settling; auto-merge SQUASH goes on once all 30 read green.

## The decisive measurement, and it is the sharpest statement of the defect anyone has made

27 tiles rendered in real Chromium, each screenshotted and SHA-256'd. At base, **six different datasets produced one byte-identical image** `13237e6e19a7072a`:

- the card's mixed-sign dataset
- PR objectui#7146's own 0/7/0 pin fixture
- positive + nulls
- positive + unparseable
- positive + missing-key
- **`sankey-genuine-1row-40` — a dataset that genuinely has one row**

A chart showing **a third of its data** was **pixel-for-pixel indistinguishable from a complete chart of one row.** `svg: 1`, `path: 3`, 18 descendants, no role, no text, and zero console output against a live console control that *did* fire.

That is what "invisibly wrong" means, stated as a hash collision rather than an adjective. And the ablation **breaks the collision**: reverting the fix moves the five thinned tiles away while `sankey-genuine-1row-40` keeps `13237e6e19a7072a`. Exactly 8 tiles changed, all thinned sankeys; **23 byte-identical**, including all-positive in three different height chains, both refusal tiles, and every other chart family.

`measured-and-declined` was genuinely available and the evidence did not support it.

## Two design constraints discovered by measurement, not chosen by taste

**A refusal was ruled OUT by a pin, not by preference.** PR objectui#7146's fixture 0/7/0 is *itself* a thinned dataset that lands in this branch — so refusing on a thinned flow would blank the very chart that pin requires drawn. That is the kind of constraint no amount of reasoning surfaces; only running the neighbour's test does.

**A plain sibling paragraph is INVISIBLE in dashboards.** Measured: in the dashboard shape the note landed at y=327 inside a box ending at y=316 — entirely outside the clip. And a wrapper `div` carrying the className is unavailable because *that className is the chart's height contract* and must keep reaching the element Recharts measures. Hence `ChartFootnote` hanging off a flex chain, deliberately mirroring this file's existing `ChartFrame` including its gate: with nothing to say, the chart is returned untouched and gains no wrapper element.

**A first attempt was measured, found clipped, and discarded before commit.** It is not in the branch. That is the right disposal of a wrong approach.

## A2.2 falsified, and the correction is load-bearing

My wording — *"the trigger is exactly: some rows survive the filter"* — is **necessary but not sufficient**. A complete draw also has rows surviving. The real trigger is `0 < rows.length < data.length`: some survive **and at least one is dropped**.

That distinction is exactly what keeps the all-positive path byte-for-byte unchanged, which the 23 identical tiles then prove. A fix built on my wording would have put a footnote on every healthy chart.

## The rest, checked

- **Seam map pinned from both sides**: `data` empty → bare div (objectui#7130's question, upstream, untouched); `rows.length === 0` with data → objectui#7146's refusal (measured: all-zeros-plus-one-negative and all-negative land there); `0 < rows.length < data.length` → this note (objectui#7146's own 0/7/0 fixture now draws **and** says so); `rows.length === data.length` → untouched.
- **Ruling 5 satisfied**: all five shapes `Number(...) || 0` folds to zero — negative, zero, `null`, unparseable, missing key — measured reaching this branch beside a survivor, each a case in the new test.
- **A2.4 confirmed with a live control**: 1 console message across 27 tiles, and it was the missing-category-key positive control. The zero is a reading, not a dead instrument.
- **`check:sdui-registration-pins` was converted from a non-reading into a verdict.** It first returned exit 2 (`No console build to weigh`) — correctly NOT MEASURED. Rather than leave it there, the console closure was built and re-run: *"All 16 registration(s) a sideEffects array promises are present in the built console."* A real green.
- **The fix cannot leak, and it is pinned rather than claimed**: `AdvancedChartImpl.tsx:1047` is the only row-dropping filter in the file — the other two `.filter(` sites filter *series*. A cross-family case pins that the note cannot reach bar/pie/donut/funnel/treemap.
- Clause ② not engaged; no i18n key, matching the file's hardcoded English.

## Correctly left alone

objectui#7146's `"no row's amount is above zero"` copy is true of post-coercion values only. I recorded that imprecision but never assigned it — and it was left untouched, with the reasoning stated: the right rewording is a judgement not pinned by existing evidence, and this fix does not need it. Not touching a neighbour's copy on your own initiative is the correct instinct.

The treemap/funnel/pie finding was **reported rather than filed**, per the fence, because it lands inside already-queued objectui#7147. I have attached it there — it widens that card from all-zero datasets to mixed ones.

---

PR #7162, opened 2026-09-01
fix(plugin-detail,i18n): ActivityTimeline's remaining 18 literals resolve from the packs

### LINKED ISSUE #7149: `ActivityTimeline` has 18 more hardcoded English literals — the empty state was one of a set, and 12 of them already have pack keys sitting unused
Split out of #7142, which fixed **one** literal in `packages/plugin-detail/src/ActivityTimeline.tsx` (the empty-state title). Sweeping the file while implementing that fix found the empty state was not a lone oversight: **the component uses no translation hook at all** on `origin/main` — it imported nothing from `@object-ui/i18n` and made no `t()` call anywhere. Filed rather than fixed, because widening #7142's scope was the PM's call, not the implementing agent's.

## Measured

Rendered `ActivityTimeline` with `activities={[]}` under a zh `I18nProvider`, and its sibling `RecordActivityTimeline` the same way, reading `textContent` off both containers:

```
[zh] ActivityTimeline       => "Activity(0)No activity recorded"
[zh] RecordActivityTimeline => "活动(0)全部动态暂无活动记录"
[no-provider, filterable] ActivityTimeline => "Activity(0)AllField ChangesCreatesDeletesCommentsStatus ChangesNo activity recorded"
```

The sibling renders the same surface entirely in Chinese. After #7142 the zh card reads `"Activity(0)暂无活动记录"` — correct for the empty state, still English for the card title beside it.

## Inventory

Line numbers are on the post-#7142 file. "pack key" means a key whose value in the `en` pack is already the exact literal, read out of the pack objects directly (not grepped — the packs are nested, so a dotted-key grep returns zero for keys that exist).

Relative timestamps, `formatTimestamp`:

| line | literal | existing key |
|---|---|---|
| 49 | `'just now'` | `detail.justNow` |
| 50 | `` `${diffMins}m ago` `` | `detail.minutesAgo` = `'{{count}}m ago'` |
| 52 | `` `${diffHours}h ago` `` | `detail.hoursAgo` |
| 54 | `` `${diffDays}d ago` `` | `detail.daysAgo` |

Filter chips, `FILTER_LABELS`:

| line | literal | existing key |
|---|---|---|
| 82 | `'All'` | none |
| 83 | `'Field Changes'` | `detail.fieldChangesFilter` |
| 84 | `'Creates'` | none |
| 85 | `'Deletes'` | none |
| 86 | `'Comments'` | `detail.comments` |
| 87 | `'Status Changes'` | none |

Card chrome:

| line | literal | existing key |
|---|---|---|
| 109 | `Activity` (the CardTitle text node) | `detail.activity` |
| 118 | `aria-label="Activity type filter"` | none (`detail.filterActivity` = `'Filter activity'` is close but not identical) |

Activity descriptions, `formatFieldChange` — sentences assembled in code, so these need new keys with interpolation holes rather than a lookup:

| line | literal | existing key |
|---|---|---|
| 66, 67, 74 | `'(empty)'` | none |
| 68 | `` `Changed ${fieldLabel} from "${oldVal}" to "${newVal}"` `` | none |
| 71 | `'Created this record'` | none |
| 72 | `'Deleted this record'` | none |
| 75 | `` `Changed status to "${newVal}"` `` | none |
| 78 | `'Updated record'` | none |

So 12 of the 18 are a lookup swap against keys that already exist and are already translated in all ten packs; the remaining 6 (plus the two chip labels and the aria label) need new keys across `en.ts` and its nine siblings, with `all-locales-key-parity` green. Never an inline `defaultValue` (#3517).

## Reachability caveat, worth checking before sizing this

`DetailView` gates the timeline on `hasActivity = !!schema.activities && schema.activities.length > 0` (`DetailView.tsx:1408`) and never passes `filterable`, so from that host the filter chips never render and the empty state is unreachable. The chips and the filtered-empty path are reachable through the **published export** (`packages/plugin-detail/src/index.tsx:118`) by a consumer passing `filterable`. That may change how this is prioritized, but it does not change the timestamps or the card title, which render on every activity tab.

## Unrelated observation in the same file

`Filter` is imported from `lucide-react` at line 11 and never used — eslint reports it as a warning today (`@typescript-eslint/no-unused-vars`, warning not error, so nothing is red).

os-warren:
Fixes #7149

`ActivityTimeline` had 18 hardcoded English literals left after #7142 fixed one
of them. All 18 now resolve from the ten packs.

## What this finishes, honestly

#7142 gave this component its **first** `t()` call — the empty-state title — and
filed the sweep that found the rest. The state it left behind is what this PR
repairs: a zh activity tab read `"Activity(0)暂无活动记录"`, the empty state
correct and the card title beside it still English. One translated string in a
component that was otherwise entirely untranslated, next to a sibling
(`RecordActivityTimeline`) that renders the same surface fully in Chinese.

Re-derived on `origin/main` head **`85b5077d7`** — the card's line numbers still
land exactly, and the re-sweep confirms the count.

## The re-sweep: 18 literals, 20 occurrences

The card's 18 is right; `(empty)` is one literal at three sites (L66, L67, L74),
which is where the other two occurrences go. No nineteenth: the only other
locale-sensitive thing in the file is `date.toLocaleDateString()` on the
past-a-week branch, which already localizes and is byte-identical to the
sibling's own tail — not a literal, deliberately untouched.

## Three groups, split by reachability rather than by defect

| group | literals | reachable from | keys |
|---|---|---|---|
| Relative timestamps + card title | 5 | **every** activity tab | 5 existing, exact |
| `formatFieldChange` sentences | 6 | **every** activity tab (entries without `description`) | 6 new |
| Filter chips + group aria-label | 7 | published export only | 2 existing exact, 4 new, 1 copy change |

**All three shipped.** The measurement that decided it: the card grouped the
`formatFieldChange` sentences apart because they need interpolation holes, but
that is a *key-shape* difference, not a reachability one. `description` is
optional on `ActivityEntry`, and `formatFieldChange` runs for every rendered
entry — so those six sentences render on exactly the same surface as the
timestamps. Shipping the timestamps without them would have reproduced #7142's
half-done shape one level down: `"Ada Changed Status from "open" to "closed"
5分钟前"`.

The chips are genuinely less reachable — `DetailView` mounts this component at
two sites (`DetailView.tsx:1586`, `:1713`) and **neither passes `filterable`**,
so no host in this repo renders them. They are included rather than deferred
because `filterable` is already published API on an already-shipped export, and
3 of those 7 need no new key at all; the marginal cost was 4 keys on a PR
already touching all ten packs. Value accrues to outside consumers only, and
that is stated rather than implied.

## Pack verification — read out of the pack objects, never grepped

The packs are nested, so a dotted-key grep returns a **false zero** for keys
that exist. Every mapping was re-verified by importing the ten pack objects and
resolving dotted paths, with both controls live:

- **positive** — `detail.back`, `detail.noActivity`, `detail.edit`: `10/10` each
- **negative** — `detail.zzzAbsentControl7149`, `detail.noSuchKeyAtAll`: `0/10`

All 7 inherited "existing key" mappings are **exact** (`en` value byte-identical
to the literal), each present in 10/10 packs:

`detail.justNow` `'just now'` · `detail.minutesAgo` `'{{count}}m ago'` ·
`detail.hoursAgo` · `detail.daysAgo` · `detail.activity` `'Activity'` ·
`detail.fieldChangesFilter` `'Field Changes'` · `detail.comments` `'Comments'`

A second probe over all 2845 `en` leaf keys confirmed the new copy has no
existing home — `Creates`, `Deletes`, `Status Changes`, `(empty)`,
`Created this record`, `Deleted this record`, `Updated record`,
`Activity type filter` all return **0 exact-value hits**, with `All`, `Create`
and `Delete` returning non-zero on the same instrument as the live control.

## One deliberate English copy change

The chip group's `aria-label` was `"Activity type filter"`. It now resolves
`detail.filterActivity` (`'Filter activity'`) — **not** an exact match, and the
card flagged it as such. Adopted anyway, and reported rather than quietly
swapped: `RecordActivityTimeline.tsx:345` already uses that exact key for the
accessible name of **its** activity filter. Minting a near-synonym would fork
one control's name across two components — the same argument
`useDetailTranslation.ts` already records for `common.resizeDrawer`. The English
accessible name changes; no host in this repo renders the control today.

## Ten new keys, all ten packs

`detail.allFilter` · `createsFilter` · `deletesFilter` · `statusChangesFilter` ·
`activityEmptyValue` · `activityFieldChanged` · `activityCreated` ·
`activityDeleted` · `activityStatusChanged` · `activityUpdated`

No inline `defaultValue` anywhere (#3517) — every key resolves from the packs or
from `DETAIL_DEFAULT_TRANSLATIONS`, which mirrors the ten new `en` rows
byte-for-byte (verified against the pack object; `defaults-maps-mirror-en-pack`
green).

Quotes live **inside** each pack's value so every locale punctuates its own way,
following each pack's measured convention rather than a guess: de `„…“`, zh
`“…”`, ja `「…」`, fr/ru `«…»`, en/ko/es/pt/ar ASCII. This moves the de quote
census, and `de-quote-pairing-3876.test.ts` is updated with it: `okSpans` and
`{open, close, rdq}` go **55 → 58** (three interpolated spans; `rdq` stays 0, so
each addition is a matched pair), with the comment naming why, in the file's
established style.

## Evidence

**Ablation** — prediction stated before running: reverting only
`ActivityTimeline.tsx` to `origin/main` (packs left in place) turns the new
locale assertions **red**. Mutation proved on disk by blob hash
(`50c29a43…` → `6bb6ce32…`, equal to `origin/main:` that path) **and** marker
count (`'just now'` literal 0 → 1). Restore proved by state — `git diff HEAD`,
`git diff --cached` and `git status --short` all empty, blob back to
`50c29a43…`. No rebuild leg needed and it is stated why: the test imports
`./ActivityTimeline` relatively, so vitest loads the source, not `dist`.

Result **`10 failed | 10 passed (20)`**. The ten reds are every locale-dependent
assertion. The ten greens are the discriminating half and are green *by
construction* on the unfixed component — the six that came with #7142, the two
no-provider English cases (each `en` pack value is byte-identical to the literal
it replaces, so English cannot tell a lookup from a literal), the
author-`description` short-circuit, and the no-raw-key sweep.

**Non-`en` render** — the actual defect is "stays English in a zh session", so
zh/ja/ar/de carry the load:

```
5分钟前 · 3小时前 · 2天前 · 刚刚          (zh, {{count}} interpolated)
5分前 (ja) · منذ 5 دقيقة (ar)
将 Status code 从“open”改为“closed”      (zh, three holes filled)
将 Owner 从“(空)”改为“Ada”               (localized (empty), inside the sentence)
Stage von „A“ zu „B“ geändert            (de, German quotes not the ASCII pair)
全部 · 字段变更 · 创建 · 删除 · 评论 · 状态变更   (six chips)
role=group name="筛选活动" / "تصفية النشاط"
```

Plus the exact repair, asserted whole: the zh card's `textContent` is now
`'活动(0)暂无活动记录'` — the string the card recorded as half-done, minus the
English half.

**Gates** — all run at `ea5e7c052`, exit codes captured before any pipe:

- `vitest run packages/plugin-detail packages/i18n defaults-maps-mirror-en-pack`
  → `Test Files 181 passed (181)` · `Tests 2066 passed (2066)`
- `type-check` (both packages, script name echoed, not a zero-match) → `Done`
- `eslint` both packages → `0 errors`; `ActivityTimeline.tsx` contributes **no**
  findings, so the unused `Filter` import warning the card noted is gone
- `check:i18n-keys` → "Every in-scope call-site key resolves against the en pack
  (2855 keys) … every call site passes exactly the arguments that value has
  holes for"
- `check:i18n-drift` → "0 en value(s) changed (10 key(s) added, 0 removed)" —
  additive, no translation follow-ups owed
- `all-locales-key-parity`, `untranslated-identity-4376`,
  `fallback-placeholder-spelling-3512`, `de-quote-pairing-3876` — all inside the
  959-test `packages/i18n` run
- `changeset:check` → "✅ All workspace packages are in the changeset fixed
  group. ✅ No changeset declares a `major` bump."

## Also in this PR

The second commit resolves the chip labels with **static** `t()` calls behind a
`filterLabel(type, t)` switch instead of `t(KEYS[type])`. A key that only
appears as a map value has no call site the scanners can resolve: all four new
chip keys rendered correctly while reading as unreferenced. Measured
`detail` needs-review **21 → 17** and dynamic call sites with no static head
**36 → 35**. Same shape as the sibling's `getFilterOptions`.

`RecordActivityTimeline.tsx` was read-only reference throughout and is
unchanged.

---

### DIFF
--- a/.changeset/7149-activity-timeline-remaining-literals.md
+++ b/.changeset/7149-activity-timeline-remaining-literals.md
@@ -1,0 +1,38 @@
+---
+'@object-ui/plugin-detail': patch
+'@object-ui/i18n': patch
+---
+
+`ActivityTimeline` speaks the session locale — the other 18 literals
+(objectui#7149).
+
+objectui#7142 gave this component its first `t()` call (the empty-state title)
+and filed the sweep that found the rest. Until now a zh activity tab read
+`"Activity(0)暂无活动记录"`: one translated string in a component that was
+otherwise entirely English.
+
+All 18 now resolve from the ten packs, in three groups:
+
+- **Relative timestamps and the card title** (`just now`, `{{count}}m/h/d ago`,
+  `Activity`) — these render on *every* activity tab. All five were a pure
+  lookup swap: the `en` pack value was already byte-identical to the literal,
+  and the sibling `RecordActivityTimeline` already used the same keys.
+- **The `formatFieldChange` sentences** — assembled in code, so they needed new
+  keys *with* interpolation holes rather than a lookup. Same reachability as the
+  timestamps: they render for any entry whose optional `description` is absent.
+  The quotes live inside each pack's value, so every locale punctuates its own
+  way (de `„…“`, zh `“…”`, ja `「…」`, fr/ru `«…»`).
+- **The six filter chips and the chip group's accessible name** — reachable only
+  through the published export, since no host in this repo passes `filterable`.
+
+Ten new `detail.*` keys across all ten packs (no inline `defaultValue` —
+objectui#3517), mirrored byte-for-byte into `DETAIL_DEFAULT_TRANSLATIONS` so a
+provider-less host still reads English rather than a raw key.
+
+One deliberate English copy change: the chip group's `aria-label` was
+`"Activity type filter"` and now resolves `detail.filterActivity`
+(`"Filter activity"`) — the key `RecordActivityTimeline` already uses for the
+accessible name of its own activity filter, so one control does not carry two
+names across two components.
+
+Also drops the unused `Filter` import (a pre-existing eslint warning).
--- a/packages/i18n/src/__tests__/de-quote-pairing-3876.test.ts
+++ b/packages/i18n/src/__tests__/de-quote-pairing-3876.test.ts
@@ -268,8 +268,13 @@ describe('objectui#3876 — de pack closes „ with “ and not with a straight 
     // 55 once objectui#6655 added `timeline.unsupported.objectBoundGantt`, the
     // object-bound timeline's refusal of `variant: gantt`, which names the
     // refused variant — „gantt“, a literal span, because the quoted thing is an
-    // authoring value the author typed rather than data the runtime filled in.
-    expect(okSpans, 'correctly paired spans').toBe(55);
+    // authoring value the author typed rather than data the runtime filled in,
+    // 58 once objectui#7149 gave `ActivityTimeline`'s assembled sentences pack
+    // keys: `detail.activityFieldChanged` quotes the OLD and NEW field values
+    // („{{old}}“ / „{{new}}“, two spans in one value, like `navigationSync.
+    // renamedPage` above) and `detail.activityStatusChanged` the new status
+    // („{{value}}“) — three interpolated spans, all runtime data.
+    expect(okSpans, 'correctly paired spans').toBe(58);
   });
 
   it('keeps the count identity that replaces the card’s count(„) === count(“)', () => {
@@ -289,11 +294,12 @@ describe('objectui#3876 — de pack closes „ with “ and not with a straight 
     // 53 / 53 / 0 after objectui#5232 added
     // `console.objectView.viewConfigPermissionDenied`; 54 / 54 / 0 after
     // objectui#6301 added `packagedAutomation.cloneCreated`; 55 / 55 / 0 after
-    // objectui#6655 added `timeline.unsupported.objectBoundGantt`. `rdq` staying
-    // at 0 is the load-bearing half: each new value added a MATCHED „…“ pair,
-    // not a stray closer that would have made `close === open` true for the
-    // wrong reason.
-    expect({ open, close, rdq }).toEqual({ open: 55, close: 55, rdq: 0 });
+    // objectui#6655 added `timeline.unsupported.objectBoundGantt`; 58 / 58 / 0
+    // after objectui#7149 added the two quoted `ActivityTimeline` sentences
+    // (three spans between them). `rdq` staying at 0 is the load-bearing half:
+    // each new value added a MATCHED „…“ pair, not a stray closer that would
+    // have made `close === open` true for the wrong reason.
+    expect({ open, close, rdq }).toEqual({ open: 58, close: 58, rdq: 0 });
     // The durable shape: every „ closed by a “, every surplus “ an English
     // opener answered by a ”. Survived translating the two English values.
     expect(close).toBe(open + rdq);
--- a/packages/i18n/src/locales/ar.ts
+++ b/packages/i18n/src/locales/ar.ts
@@ -960,6 +960,19 @@ const ar = {
     tasksOnly: "المهام فقط",
     leaveCommentPlaceholder: "اترك تعليقاً… (Ctrl+Enter للإرسال)",
     noActivity: "لا يوجد نشاط مسجل",
+    // objectui#7149 — the rest of `ActivityTimeline`. The four chip labels the
+    // `detail.*` pack did not already name, and the `formatFieldChange`
+    // sentences, which are assembled in code and so need interpolation holes.
+    allFilter: 'الكل',
+    createsFilter: 'عمليات الإنشاء',
+    deletesFilter: 'عمليات الحذف',
+    statusChangesFilter: 'تغييرات الحالة',
+    activityEmptyValue: '(فارغ)',
+    activityFieldChanged: 'غيّر {{field}} من "{{old}}" إلى "{{new}}"',
+    activityCreated: 'أنشأ هذا السجل',
+    activityDeleted: 'حذف هذا السجل',
+    activityStatusChanged: 'غيّر الحالة إلى "{{value}}"',
+    activityUpdated: 'حدّث السجل',
     loadMore: "تحميل المزيد",
     edited: "(معدل)",
     via: "عبر {{source}}",
--- a/packages/i18n/src/locales/de.ts
+++ b/packages/i18n/src/locales/de.ts
@@ -954,6 +954,19 @@ const de = {
     tasksOnly: "Nur Aufgaben",
     leaveCommentPlaceholder: "Kommentar hinterlassen… (Strg+Enter zum Senden)",
     noActivity: "Keine Aktivitäten aufgezeichnet",
+    // objectui#7149 — the rest of `ActivityTimeline`. The four chip labels the
+    // `detail.*` pack did not already name, and the `formatFieldChange`
+    // sentences, which are assembled in code and so need interpolation holes.
+    allFilter: 'Alle',
+    createsFilter: 'Erstellungen',
+    deletesFilter: 'Löschungen',
+    statusChangesFilter: 'Statusänderungen',
+    activityEmptyValue: '(leer)',
+    activityFieldChanged: '{{field}} von „{{old}}“ zu „{{new}}“ geändert',
+    activityCreated: 'Diesen Datensatz erstellt',
+    activityDeleted: 'Diesen Datensatz gelöscht',
+    activityStatusChanged: 'Status zu „{{value}}“ geändert',
+    activityUpdated: 'Datensatz aktualisiert',
     loadMore: "Mehr laden",
     edited: "(bearbeitet)",
     via: "über {{source}}",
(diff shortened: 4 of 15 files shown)


os-warren:

CI still settling; auto-merge SQUASH goes on once all 30 read green. Posting now because the scope correction here is mine to record.

## ⚠️ My tiering was wrong, and the correction is the right one

I split the 18 literals into three tiers and ruled Tier A ships while B and C are include-or-defer. **Tier C was mis-grouped.**

I separated the six `formatFieldChange` sentences because they need **interpolation holes** — a key-*shape* difference. You measured that `description` is optional on `ActivityEntry` and `formatFieldChange` runs for **every rendered entry**, so those sentences render on **exactly the same surface as Tier A**. They are not less reachable; they are just a different key shape.

Deferring them on my grouping would have shipped a component whose timestamps and title are translated and whose *activity descriptions* are still English — **objectui#7142's half-done shape reproduced one level down**, which is the specific thing I flagged as the visible cost of my earlier scoping. Shipping all three tiers was correct, and the reason you give is a measurement rather than a preference.

## The judgement call you surfaced instead of burying

`detail.filterActivity` (`'Filter activity'`) is **not identical** to the aria-label `'Activity type filter'` — the card flagged it and you confirmed it. You adopted the key anyway and **reported it as a copy change** rather than quietly swapping it, with the reason: `RecordActivityTimeline.tsx:345` already uses that exact key for the accessible name of its own activity filter, so minting a near-synonym would fork **one control's name across two components**. And no host in this repo renders the control today.

That is the right disposal — a small deliberate copy change, named, with the alternative and its cost stated. A silent adoption would have been indistinguishable in the diff and much worse.

## Four instrument checks that each could have gone wrong

**A3.3 done on the pack objects**, never a dotted-key grep — positive control (`detail.back` / `noActivity` / `edit` = 10/10) and negative control (two invented keys = 0/10). All 7 backing keys byte-identical and present 10/10.

**A3.4's near-miss is worth more than the result.** Repo-wide greps for `'just now'` and `'m ago'` return **18 and 21 hits** — but every test assertion among them belongs to `packages/collaboration` against `collaboration.*` keys. A different package and namespace. Reading those hits as "something pins these literals" would have been a false positive, and the sweep only survived because you attributed each hit rather than counting them.

**A3.5 verified at the mechanism, not assumed.** `{{name}}` with no inner spaces is the one spelling **both** paths resolve (i18next on the provider path, `interpolateFallback` on the provider-less one). Both legs asserted — including a separate assertion that `'{{count}}m ago'` does **not** appear on the no-provider path, which is exactly where a wrong spelling renders raw and silently.

**And you caught your own change degrading a repo gate.** Resolving chip labels via `t(KEYS[type])` leaves keys appearing only as **map values**, which the scanners cannot resolve — all four new chip keys read as *unreferenced*. The second commit converts them to static `t()` calls, measured: detail `needs-review` **21 → 17**, dynamic call sites **36 → 35**. Noticing that a correct-looking change would make a gate less able to see is the hard version of this work.

## Evidence

Ablation `10 failed | 10 passed (20)`, direction predicted before running — and the part that makes it discriminating is stated: the ten greens are green **by construction** on pre-fix code (objectui#7142's own six, the two no-provider English cases where the pack value is byte-identical to the literal, the author-description short-circuit, and the no-raw-key sweep). A red half without that explanation would prove much less.

Non-`en` render evidence across zh / ja / ar / de, including German quotes rather than the ASCII pair, and the `de` quote-pairing census moved **55 → 58** in `de-quote-pairing-3876.test.ts` — updated in that file's established style rather than left to break.

One NOT MEASURED discarded rather than reported: the first mirror-verification died `ERR_MODULE_NOT_FOUND` on an unbuilt closure. Built and re-ran; all ten keys mirror exactly, with an absent-key control reading `undefined` on both sides.

Clause ② confirmed disengaged — `FILTER_LABELS` was module-private and is not exported; `check:i18n-drift` confirms **0 en values changed**, purely additive.

## The loop closed

The zh card now reads **活动(0)暂无活动记录** — precisely the string objectui#7149 recorded as half-done, minus its English half. Asserting the exact repair as a whole, rather than only the parts you touched, is what makes it verifiable that the defect is gone rather than merely reduced.

objectui#7163 filed from the same sweep, `pm:blocked` on this PR (the ten pack files are the gate, not the components).

---

PR #7167, opened 2026-09-01
test(plugin-grid): render what the inline lookup picker actually receives

### LINKED ISSUE #7154: bug(plugin-grid): ObjectGrid's inline lookup picker ignores four more spec-declared field props — `multiple`, `allowCreate`, `lookupPageSize`, `dependsOn` are read on its own path and never copied
Provenance: measured while implementing objectui#6875, by the derivation that card
asked for. Filed unassigned. Not folded into #6875 — that card names five display/target
spellings and its fix closes those; these four are a different slice of the same seam and
each changes picker BEHAVIOUR, not just which display string renders.

## What's measured

objectui#6875 replaced `ObjectGrid`'s hand-kept `RELATIONAL_META_KEYS` with a table in
`packages/plugin-grid/src/relationalMetaKeys.ts` that classifies every key the grid's own
consumers read off the `fieldMeta` bag, plus a gate
(`__tests__/relationalMetaCopySet.derivation.test.ts`) that re-extracts that read set from
the consumer sources. Running the extractor over the three consumers on the grid's render
path — `LookupCellRenderer` (`packages/fields/src/index.tsx`), `LookupField.tsx` and
`UserField.tsx` — surfaced four keys that are:

1. **read on this path**, off `fieldMeta`, and
2. **declared by `@objectstack/spec` 17.2.0's `FieldSchema`**, so a spec-compliant
   producer can and does emit them, and
3. **never copied** onto the column meta.

Verbatim from `packages/fields/src/widgets/LookupField.tsx` on `dd35800af`:

- `:259` — `const multiple = fieldMeta?.multiple || false;`
- `:270` — `const explicitAllowCreate = fieldMeta?.allow_create ?? fieldMeta?.allowCreate;`
- `:277` — `const lookupPageSize: number | undefined = fieldMeta?.lookup_page_size ?? fieldMeta?.lookupPageSize;`
- `dependsOn` — read the same way, and the driver of the cascading picker filter.

All four are on `FieldSchema`'s 71-property strict shape (measured on the installed
package, `name`/`type`/`label` as the positive control). None is on `RELATIONAL_META_KEYS`
even after #6875.

They are carried in the new table with verdict `deferred` and a note, so the gate keeps
them visible rather than letting them lapse back into "nobody noticed" — but `deferred`
means the value still does not arrive.

## Why each is a user-visible defect, not a tidiness item

- `multiple` — an author declaring a multi-value lookup gets a SINGLE-select inline picker
  in the grid. The form surface honours it; the grid does not.
- `allowCreate` — an author setting `allowCreate: false` to opt a relation out of inline
  quick-create still gets the create affordance in the grid's picker, because the explicit
  value never arrives and `isUserFacingReference()` decides instead.
- `lookupPageSize` — the author's page size is ignored; the picker uses its default.
- `dependsOn` — the cascading filter the author declared does not scope the grid's picker.
  ⚠️ objectui#2215 is a closed card in this area ("Cascading lookup (dependsOn) broken in
  forms; table picker bypasses the dependent filter") — worth reading before acting, in
  case this is the grid-side residue of it rather than a fresh defect.

## What is NOT measured here

- No user report is attached. This is the same static producer-vs-consumer measurement
  objectui#6875 was, run one seam wider — the reachability half is measured (spec declares
  all four), the rendering half is not: nobody has yet rendered a grid over a `multiple`
  lookup and watched the picker.
- Whether `plugin-dashboard` needs any of them: it does NOT on today's reading — its two
  widgets are read-only and reach no field EDITOR, so the picker keys have no reader on
  that path. That verdict is already recorded on `CELL_RELATIONAL_META_KEYS` and does not
  move with this card.

## Shape of a fix

Mechanically small — flip four `deferred` verdicts to `spec` in
`relationalMetaKeys.ts` and the derived copy set picks them up, with no list to edit. The
work is the behaviour: each of the four wants a rendering test on the grid's own path, of
the kind objectui#6875 added for `displayField`
(`__tests__/lookupDisplayFieldSpelling-6875.test.tsx`), and `dependsOn` wants objectui#2215
read first.

⛔ Do not also copy the snake_case twins the same chains read (`allow_create`,
`lookup_page_size`, `depends_on`). `FieldSchema` declares none of them — they parse to
`unrecognized_keys` — so no producer can fill them, which is exactly why objectui#6875 left
`reference_field` and `lookup_columns` out. The gate proves that absence against the
installed spec.

os-warren:
Fixes #7154

Re-derived on `origin/main` head **`51449a043`**. The card asked for four `deferred` verdict flips plus rendering proof. The rendering proof killed the premise, so this PR ships the measurement and the corrections it forces, and flips nothing.

## The premise, and what falsified it

objectui#7154 says `multiple`, `allowCreate`, `lookupPageSize` and `dependsOn` are spec-declared, read off `fieldMeta` by `LookupField`, absent from `RELATIONAL_META_KEYS`, and therefore never reach ObjectGrid's inline lookup picker. The first three claims hold on this head. The fourth does not: **all four already take effect in the picker, with the copy set unchanged**, because the picker is not fed that bag at all.

- `applyRelationalMeta` writes onto `fieldMeta`, and `fieldMeta` goes to `CellRenderer` and nowhere else in `ObjectGrid.tsx` — the read-only cell. For a relational column that is `LookupCellRenderer`, which reads `reference_to`, `reference`, `display_field`, `displayField`, `reference_field`: none of the four.
- The inline EDITOR is a different seam. `renderCellEditor` looks the field up in the object schema and spreads the whole def into the widget — `{ name: ctx.column.accessorKey, ...fieldDef }` — so every key a def carries reaches `LookupField` regardless of the copy set.
- Both halves read `objectSchema.fields[name]`, the same object, so there is no shape where copying could rescue an editor the schema read did not already serve: when that lookup misses, `renderCellEditor` returns `null` and `applyRelationalMeta` copies nothing, together.

Flipping the four verdicts would therefore write four members onto a bag whose consumer does not read them — the shape objectui#6711 (`reference_to_field`) and objectui#6874 (`titleFormat`) each retired.

## Per-key rendering evidence

New `packages/plugin-grid/src/__tests__/lookupPickerKeys-7154.test.tsx`. Each case renders one data source, one referenced object, and two lookup columns whose defs differ ONLY in the key under test — the shape `lookupDisplayFieldSpelling-6875.test.tsx` established. The control column is load-bearing in both directions.

| key | declared column | control column (same source, key absent) |
|---|---|---|
| `multiple: true` | two picks accumulate, trigger reads "2 selected" | second pick REPLACES the first, trigger reads "Person 02" |
| `allowCreate: false` | no quick-create entry in the popover | "Create new" present (default ON for a user-facing reference) |
| `lookupPageSize: 3` | picker dialog queries `$top: 3` and renders 3 rows | same dialog over the same 12 records renders 10 (`RecordPickerDialog`'s default) |
| `dependsOn: ['region']` | trigger is `lookup-trigger-gated`, disabled, "Select region first"; browse-all disabled too | `lookup-trigger-owner`, enabled |

A fifth case pins the classification half: the copy set is non-empty, contains `displayField` (the control), and contains none of the four.

## Ablations — direction predicted before running, both as predicted

Run from committed `21509d3c8`; each mutation proved on disk by anchor/marker counts plus a blob-hash change, each restore proved by `git diff HEAD`, `git diff --cached` and `git status --short` all empty plus a hash equal to the HEAD blob. No build step is involved — this repo's vitest aliases `@object-ui/*` to sibling `src/`, so the mutated source is what runs.

- **A1 — strip the four keys from `renderCellEditor`'s spread.** Predicted: the four behavioural cases go red, the classification case stays green. Observed: `Tests 4 failed | 1 passed (5)`. The editor seam is the delivery route.
- **A2 — flip all four verdicts `deferred` to `spec`.** Predicted: only the classification case goes red (it asserts non-membership), every behavioural case unchanged. Observed, against this file plus the derivation gate: `Tests 1 failed | 12 passed (13)`. Flipping is behaviourally inert — it changes the classification and nothing a user sees.

## What this changes

- The four verdicts stay `deferred`, with the measurement in their notes instead of a promise, so the derivation gate keeps watching them.
- Corrected the two docblocks (`relationalMetaKeys.ts`, `relationalMetaCopySet.derivation.test.ts`) that claimed the inline editor "dispatches the same bag" into `LookupField` and `UserField`, and the three `applyRelationalMeta` call-site comments in `ObjectGrid.tsx` that said the copy exists so "the inline picker can query". Those sentences are what this card was filed against.
- Changeset with EMPTY frontmatter — no behaviour change. Gate verdict, verbatim: `✅  4 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/7154-lookup-picker-keys-editor-seam.md.` / `Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.`

## Assumptions checked

- **A2.2 (all four read on this path)** — holds, re-measured on this head: `LookupField.tsx` reads them at `:259`, `:271`, `:277`, `:300`. What moved is which bag feeds that file on the grid path.
- **A2.3 (spec still declares all four)** — holds. `@objectstack/spec` 17.2.0 as installed, `FieldSchema` strict with 71 props; all four declared, none of `allow_create` / `lookup_page_size` / `depends_on` declared; positive control `name` / `type` / `label` present; a nonsense key parses to `unrecognized_keys`.
- **A2.4 (plugin-dashboard has the same gap)** — falsified, and it is not a scope question after all. Its two widgets reach no field editor (`renderFieldValue` to `getCellRenderer`, a cell renderer), so the picker keys have no reader there; that verdict is already recorded on `CELL_RELATIONAL_META_KEYS` and pinned in `lookupRelationalMeta-6694.test.tsx`. Nothing to fix.
- **A2.5 (nothing pins the current absence)** — holds. No test in `plugin-grid` or `plugin-dashboard` mentions any of the four.

## Filed, not fixed here

- objectui#7165 — a `dependsOn` lookup column is permanently uneditable in ObjectGrid. The key arrives (the gate is proof) and the gate never lifts, because the grid supplies no dependent values through any of `LookupField`'s three channels. PR objectui#2216 gave the FORM that injection; the grid never got it. Whether the grid should feed the saved row or the row plus its staged edits is a seam decision, which is why it is filed rather than ridden in here.
- objectui#7166 — 6 of the 10 keys currently copied onto `fieldMeta` have no reader on that bag, including two that objectui#6875 added as its fix. Same measurement, opposite direction; retiring them is its own adjudication.

## Verification

Union re-run at the final commit `21509d3c8`:

- `pnpm exec vitest run` over `lookupPickerKeys-7154`, `relationalMetaCopySet.derivation`, `relationalMetaCopySet-6711`, `relationalMetaCopySet-6874`, `lookupDisplayFieldSpelling-6875`, `inlineEditLookupRepro` and `plugin-dashboard`'s `lookupRelationalMeta-6694` — `Test Files 7 passed (7)` / `Tests 35 passed (35)`.
- `pnpm --filter @object-ui/plugin-grid run type-check` — exit 0. First attempt was NOT MEASURED (TS2307 across `@object-ui/*`, an unbuilt dependency closure); re-run after `pnpm --filter '@object-ui/plugin-grid^...' build`. `tsconfig.test.json` includes the new file (1 hit under `--listFiles`), so the typecheck covers it.
- `pnpm --filter @object-ui/plugin-grid run lint` — exit 0, 731 warnings, 0 errors (all pre-existing).
- `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check-changeset-presence`, `check-changeset-no-major`, `check-changeset-fixed` — all exit 0.

---

### DIFF
--- a/.changeset/7154-lookup-picker-keys-editor-seam.md
+++ b/.changeset/7154-lookup-picker-keys-editor-seam.md
@@ -1,0 +1,29 @@
+---
+---
+
+Measurement-only change in `@object-ui/plugin-grid`: objectui#7154 reported that
+`ObjectGrid`'s inline lookup picker never receives `multiple`, `allowCreate`,
+`lookupPageSize` or `dependsOn` because they are not on the relational copy set.
+Rendered against the grid's own inline editor, all four already take effect —
+the copy set is not the route.
+
+`applyRelationalMeta` writes onto the `fieldMeta` handed to `<CellRenderer>`,
+the read-only cell. The inline editor is a different seam: `renderCellEditor`
+looks the field up in the object schema and spreads the whole def into the
+widget, so every key a def carries reaches `LookupField` whether or not it is
+copied. Both halves read `objectSchema.fields[name]`, so copying could never
+rescue an editor the schema read did not already serve.
+
+- New `__tests__/lookupPickerKeys-7154.test.tsx` renders each key against a
+  control column differing only in that key: `multiple` accumulates two picks
+  ("2 selected") where the control replaces; `allowCreate: false` removes the
+  quick-create entry the control offers; `lookupPageSize: 3` scopes the picker
+  dialog to 3 rows against a control of 10; `dependsOn` arrives and gates.
+- The four verdicts stay `deferred` — flipping them would write members onto a
+  bag whose consumer does not read them, the shape objectui#6711 and
+  objectui#6874 retired — with the measurement now in their notes.
+- Corrected the docblocks and the three `applyRelationalMeta` call-site comments
+  that claimed the inline picker reads this bag; that claim is what the card was
+  filed against.
+
+No behaviour change.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -424,11 +424,18 @@ function getDataConfig(schema: ObjectGridSchema): ViewData | null {
 /**
  * The relational copy set and `applyRelationalMeta` moved to
  * `./relationalMetaKeys` for objectui#6875. The list there is DERIVED from a
- * table classifying every key the grid's own cell renderer and inline picker
- * read off this bag, and a gate re-derives that read set from the consumer
- * sources — so the copy set can no longer drift into being a strict subset of
- * what its consumers read, which is what it had silently become. Read that
- * file's docblock before adding, removing or re-spelling a key.
+ * table classifying every key swept off the three `@object-ui/fields` consumers,
+ * and a gate re-derives that read set from their sources — so the copy set can
+ * no longer drift into being a strict subset of what those consumers read,
+ * which is what it had silently become. Read that file's docblock before
+ * adding, removing or re-spelling a key.
+ *
+ * ⚠️ What this bag actually feeds is ONE of those three: the read-only cell.
+ * `fieldMeta` is handed to `<CellRenderer field={…}>` and nowhere else in this
+ * file. The inline picker is the OTHER seam — `renderCellEditor` below looks
+ * the field up in `objectSchema` and spreads the whole def into the widget, so
+ * a picker key reaches `LookupField` whether or not it is on the copy set.
+ * Measured in `__tests__/lookupPickerKeys-7154.test.tsx` (objectui#7154).
  */
 
 /**
@@ -2134,7 +2141,8 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
               if (objectDefField.options) fieldMeta.options = translateOptions(schema.objectName, col.field, objectDefField.options);
             }
             // Preserve relational metadata (reference_to, display_field, …) so
-            // lookup cells resolve ids to names and the inline picker can query.
+            // lookup CELLS resolve ids to names. ⛔ Not the inline picker — that
+            // reads the schema def directly, see `renderCellEditor` (objectui#7154).
             applyRelationalMeta(fieldMeta, objectDefField as any);
             // Auto-generate options from data for inferred select without existing options
             if (inferredType === 'select' && !fieldMeta.options) {
@@ -2326,7 +2334,8 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
             if (fieldDef.options) fieldMeta.options = translateOptions(schema.objectName, fieldName, fieldDef.options);
           }
           // Preserve relational metadata (reference_to, display_field, …) so
-          // lookup cells resolve ids to names and the inline picker can query.
+          // lookup CELLS resolve ids to names. ⛔ Not the inline picker — that
+          // reads the schema def directly, see `renderCellEditor` (objectui#7154).
           applyRelationalMeta(fieldMeta, fieldDef as any);
           // Auto-generate select options from data when no options defined
           if (resolvedType === 'select' && !fieldMeta.options) {
@@ -2483,7 +2492,8 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
             if (fieldDef.options) fieldMeta.options = translateOptions(schema.objectName, fieldName, fieldDef.options);
           }
           // Preserve relational metadata (reference_to, display_field, …) so
-          // lookup cells resolve ids to names and the inline picker can query.
+          // lookup CELLS resolve ids to names. ⛔ Not the inline picker — that
+          // reads the schema def directly, see `renderCellEditor` (objectui#7154).
           applyRelationalMeta(fieldMeta, fieldDef as any);
           // Auto-generate select options from data when no options defined
           if (resolvedType === 'select' && !fieldMeta.options) {
--- a/packages/plugin-grid/src/__tests__/lookupPickerKeys-7154.test.tsx
+++ b/packages/plugin-grid/src/__tests__/lookupPickerKeys-7154.test.tsx
@@ -1,0 +1,334 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7154 — the four picker keys `multiple`, `allowCreate`,
+ * `lookupPageSize` and `dependsOn` DO reach `ObjectGrid`'s inline lookup
+ * picker, and they reach it WITHOUT being on the relational copy set.
+ *
+ * ## The card this file answers, and why it is not four verdict flips
+ *
+ * objectui#7154 was filed off the derivation objectui#6875 built: the extractor
+ * in `relationalMetaCopySet.derivation.test.ts` sweeps the three consumers for
+ * keys read off a field-meta bag, and these four came back read, spec-declared
+ * (`FieldSchema` 17.2.0 — 71 strict props, all four present, measured with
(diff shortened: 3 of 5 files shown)


os-warren:
## ⚖️ Ruling on your open question: **A — keep `Fixes #7154`.** Reviewed; will arm on green.

Your reasoning is right and I am adopting it. The card's stated acceptance criterion was **rendering evidence per key**; that evidence now exists and is pinned. Leaving objectui#7154 open would leave a card whose stated defect has been *measured absent*, while the two real defects it uncovered already have their own cards. Option B costs a triage round and buys nothing the PR body does not state.

⚠️ **One condition**: the PR body must make the falsification unmistakable, so nobody reading the merge later believes four keys were wired. It reads that way now — keep it that way.

## What you falsified, and why it matters more than the card did

The card said four spec-declared keys never reach the inline picker. **All four already reach it**, with the copy set unchanged:

- `applyRelationalMeta` writes onto the `fieldMeta` handed to **`CellRenderer`** — the read-only cell, whose `LookupCellRenderer` reads **none** of the four.
- `renderCellEditor` is a **different seam**: it looks the field up in `objectSchema` and **spreads the whole def** into the widget, so every key a def carries arrives regardless.
- Both halves read `objectSchema.fields[name]` — so copying could **never** rescue an editor the schema read already served.

That last sentence is the load-bearing one. It is not "the keys happen to arrive anyway"; it is that the proposed mechanism *cannot* be the delivery path for this consumer, in either direction. A verdict flip would have been inert by construction.

**Ablation A2 proves exactly that**: flipping all four verdicts `deferred` → `spec` gives `1 failed | 12 passed (13)` — only the classification case moves, every behavioural case unchanged. Ablating the change you *declined to make* and showing it is behaviourally inert is the strongest possible form of "this was not the fix." Predicted before running, both ablations, with mutation proven by anchor count **and** blob hash and restore proven by three-way state.

## The correction reaches back into objectui#6875, and I have recorded it

objectui#7166's finding means the card I landed an hour ago **delivered 1 of the 3 keys it shipped** — `displayField` genuinely (it is read off the bag by `LookupCellRenderer`, and its rendering test proves it), while `descriptionField` and `lookupColumns` are inert on that bag because `LookupField` gets them from the schema spread.

I have posted that correction on objectui#6875. The lesson is mine: I required rendering evidence for the *defect* and got it for one key, then accepted **derivation alone** for the other two. Derivation establishes that *a consumer reads the key* — it cannot establish that *this bag is how the consumer gets it*. Those are different claims and I treated them as one.

## `dependsOn` — not deferred, and correctly so

I fenced `dependsOn` behind reading objectui#2215 first. You read it: closed **COMPLETED** by PR objectui#2216, which gave the *form* live `dependentValues` injection. The grid never got it. So the key arrives and the picker **gates permanently**, because the grid supplies none of `LookupField`'s three dependent-value channels.

That is a live defect, not a deferral — filed as **objectui#7165**, with the saved-row vs staged-row seam decision spelled out rather than pre-made. Exactly the right disposal: the fence asked you to establish what objectui#2215 settled before acting, and establishing it turned a deferral into a finding.

## Also right

The docblock and three call-site comments claiming the picker reads this bag are corrected — those sentences *are* what the card was filed against, so fixing them is fixing the actual defect. The four verdicts stay `deferred` with the measurement in their notes, so the derivation gate keeps them visible and nobody re-derives this from scratch.

`type-check` returned NOT MEASURED first (TS2307, unbuilt closure), was named as such rather than reported red, and re-run to a real exit 0 after the closure build. Changeset verdict quoted verbatim — empty frontmatter.

---

PR #7168, opened 2026-09-01
docs(plugin-timeline): the gantt row walk's non-total reads become a measured enumeration, and the class claim is corrected

### LINKED ISSUE #7153: finding(plugin-timeline): the gantt date path has FIVE non-total operations UPSTREAM of the speller — the property reads that fetch the date crash before `spellGanttDateValue` is ever entered
Found while implementing #7036. Not repaired there: #7036's ruled scope is `spellGanttDateValue`'s trailing three branches, and every site below is in a different function (`findUnusableGanttDate`), reached *before* that helper is entered.

## What is wrong

#7036's headline is that `spellGanttDateValue`'s `Array.isArray` is "the last non-total operation on the gantt date path". **Measured, that is false.** It is the last non-total operation *inside that function*; the path itself has five more, and all five run **earlier** — in the property reads that fetch the date out of the authored document.

`findUnusableGanttDate` walks the rows with plain member access:

```
const rowItems = (items[rowIndex]?.items || []) as any[];
const value = rowItems[itemIndex]?.[key];
```

`?.` guards `null` / `undefined`. It does not guard a **getter that throws** or a **revoked `Proxy`** — both crash inside the read itself, before any value exists to judge. `calculateDateRange` repeats the same two reads in its `flatMap`.

## Measurement

In-render on `dd35800af`, through the real `TimelineRenderer` (not a replica), same run that produced #7036's readings:

| input | outcome |
|---|---|
| `items[0].items[0]` with a throwing `endDate` getter | THREW `Error: endDate getter throws` |
| `items[0].items[0]` is a revoked `Proxy` | THREW `TypeError: Cannot perform 'get' on a proxy that has been revoked` |
| `items[0]` is a revoked `Proxy` | THREW same |
| `items[0].items` is a revoked `Proxy` | THREW same |
| `items[0]` with a throwing `items` getter | THREW `Error: items getter throws` |
| CONTROL — an ordinary row/item | drew 1 bar |

The control is live and non-zero on the same instrument, so the five THREW rows are readings rather than a broken harness.

## Severity, honestly

**p3, the same class and the same argument as #7036.** Not reachable from an authored document: ObjectUI metadata is JSON, and JSON can spell neither a getter nor a proxy. Re-swept on `dd35800af` — zero `Proxy.revocable` in the repo and zero `Object.setPrototypeOf` calls in `packages/*/src`, each beside a live control on the same instrument (52 `Proxy` mentions, 19 `Object.assign` calls). The three non-comment `__proto__` hits are denylists that block it.

So this is filed for the **claim** it corrects, not for the crash it describes. It is recorded because #7036's docblock now states the path is not total, and that statement should point at an enumeration rather than at prose.

## Suggested scope if picked up

⚠️ Read this with #7036's warning first. This path has now carried **six** totality claims (#6759 → #6905 → #6907 → #7027 → #7036 → this), five of them falsified by the next card's measurement. A repair here that ends with a new confident sentence continues the sequence rather than closing it.

Two defensible branches, the same either/or #7027 and #7036 were filed under:

- **Make the reads total** — a bounded `try` around the row/item walk in `findUnusableGanttDate` (and the matching `flatMap` in `calculateDateRange`), refusing the row through the existing `timeline.gantt.unusableRange.malformedDate` diagnostic rather than dying. Note this is a *wider* change than it looks: it touches the accept set's plumbing, which is #6781's ruling.
- **State the exclusion and leave it** — which is what #7036 chose for its own site, on the argument that a `catch` around a read whose failure carries no information is substitution rather than a read, and that substitution is the consumer-side tolerance #6750 and #6759 both refused.

The second is the cheaper and probably the better one, but it is an adjudication and this card does not make it. If a row is added it belongs in the existing pinned adversarial set (`packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx`), not in a new file.

Refs: #7036 (the speller's own exclusion, where these five were measured) · #7027 (the brand gate) · #6907 / PR #7026 (the spelling rule) · #6781 / PR #6905 (the type rule)

os-warren:
Fixes #7153

Re-derived on `51449a043`, which already carries #7157 (`869b876c8`) — the docblock this card was filed against.

Documentation and pinned tests only. No behaviour changes, no export changes, no i18n keys.

## The headline: the card's own count and class were both wrong

#7153 said FIVE non-total operations upstream of the speller, all of them the same p3 reachability class. Driven in render through the real `TimelineRenderer` on `51449a043`:

- **NINE**, not five. Six in `findUnusableGanttDate` (`items.length`, `items[i]`, `items[i]?.items`, `rowItems.length`, `rowItems[j]`, `rowItems[j]?.[key]`), three in `calculateDateRange`. A tenth is identified by reading and recorded as NOT MEASURED — `item.startDate` / `item.endDate` are bare, but U6's `?.` shadows them and no input was found that reaches them.
- **Two classes, not one.** The six are unreachable from an authored document. The three in `calculateDateRange` are **ordinary JSON**, and they are a live defect: filed as #7164, not addressed in this PR.

Each site is attributed by the stack frame it threw from, per site:

```
U1  items.length                818:43   revoked Proxy as items; throwing length trap
U2  items[rowIndex]             819:23   throwing index trap
U3  items[rowIndex]?.items      819:38   revoked Proxy as the ROW; throwing `items` getter
U4  rowItems.length             820:50   revoked Proxy as row.items; throwing length trap
U5  rowItems[itemIndex]         822:23   throwing index trap
U6  rowItems[itemIndex]?.[key]  822:42   revoked Proxy as the ITEM; throwing date getter
D1  items.flatMap               243:26   items is a truthy non-array          <- JSON
D2  row.items      (bare)       244:10   items: [null]                        <- JSON
D3  (row.items || []).flatMap   244:23   row.items is a truthy non-array      <- JSON
CONTROL  an ordinary row                 DREW bars=1 axisCells=3
```

## Which branch of the either/or, and the measurement that chose it

The **docblock branch** (#7036's precedent, `869b876c8`), on three measurements rather than on inheritance.

**1. Repairing inside the ruled scope closes nothing.** Ablation: making `calculateDateRange` tolerant (`Array.isArray(items) ? items : []` plus `row?.items`) closes ZERO of the three JSON-reachable crashes. It MOVES all three into the render loop — `items.map` at 1235, `row.label` at 1237, `(row.items || []).map` at 1246 — with the ordinary-row control drawing its single bar on both sides. Relocating a crash class while a docblock reports it closed is the mistake this path has made four times.

**2. A `catch` upstream DOES read something — half of it.** #7153's dispatch required this to be re-tested rather than inherited from #7036, and it comes out **differently**. Ablation: a `try` around U6 reporting the path built from the loop counters came back NAMED `items[0].items[0].endDate` — exact, including the key whose getter threw, because the path is built from indices and never touches the value. The speller's `catch` has no such thing; it holds only the value it cannot read. But the VALUE half is still substitution (`is "UNREADABLE"`), so a real repair needs a value-less diagnostic and therefore a new i18n key across ten locale packs — the file surface #7036 deferred. And the coverage is #7036's own trade at a different site: that `catch` converted **3 of 9** and left six throwing.

**3. The reachability sweep still holds for the six.** Re-run on `51449a043` with a comment-stripped instrument. Across every package `src/` outside tests: `Proxy.revocable` 0, `new Proxy` 0, the bare word `Proxy` 0, `Object.setPrototypeOf` 0 — beside live controls `Object.assign` 19 and `JSON.parse` 105 on that same instrument. The SAME query widened to the whole repo returns `Proxy.revocable` 1, `new Proxy` 21, `Proxy` 105 (tests only), which is what makes the narrow zeros readings rather than a broken query.

## What is in the diff

- `renderer.tsx` — a docblock on `findUnusableGanttDate` enumerating U1..U6 with the input that drove each, separating the two reachability classes, and recording the re-tested `catch` result; a docblock on `calculateDateRange` carrying D1..D3, the ablation, and the deferral to #7164; and a correction to `spellGanttDateValue`'s block, whose "five further crash sites, the same reachability class" paragraph was wrong twice.
- `timeline-gantt-date-brand-7027.test.tsx` — **pin 5** (the six, stated-and-exercised on #7036's terms, each assertion naming the throw's own MESSAGE so a row fails if the site MOVES as well as if it is repaired, plus a live-Proxy control at all four positions) and **pin 6** (the three, pinned explicitly as a DEFECT and not as behaviour, with seven drawing controls).

⚠️ No claim here is a totality claim. Every statement in both docblocks is bounded to the input set in that file and says so.

## Verification — all on `5213e66d3`

```
pnpm exec vitest run packages/plugin-timeline/          Test Files 19 passed (19)   Tests 247 passed (247)
  (brand-7027 alone: 36 passed, was 15 — pin 5 adds 11, pin 6 adds 10)
pnpm --filter @object-ui/plugin-timeline run type-check  exit 0, script echoed
pnpm --filter @object-ui/plugin-timeline run lint        exit 0 — 0 errors (98 pre-existing warnings)
node scripts/check-changeset-presence.mjs
  "✅  2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)
   ... Every one of them has an EMPTY frontmatter — declared as releasing nothing, which
   is the explicit exemption and a complete answer to this gate."
node scripts/check-changeset-no-major.mjs   "✅  No changeset declares a `major` bump."
check:control-bytes ✅ · check:vi-mock-specifiers ✅ · check:vi-mock-inherit ✅ · check:i18n-keys ✅
check:i18n-dead-keys ✅ · check:shell-escape-residue ✅ · check:self-import ✅ · check:phantom-deps ✅
check:element-data-source-declaration ✅
check:sdui-registration-pins — NOT MEASURED, exit 2 on a prerequisite ("No console build to weigh
  at apps/console/dist/assets"), unrelated to this diff, which touches no registration.
```

**Both new pins were proved able to move**, since a pin that cannot fail measures nothing:

| ablation (mutation proved on disk by marker counts and a moved blob hash; restore proved by blob equality with HEAD and an empty `git diff HEAD`) | predicted | measured |
|---|---|---|
| tolerant `calculateDateRange` | pin 6's 9 rows red, rest green | **9 failed, 27 passed** |
| `try` around the whole row walk | pin 5's 10 throw-rows red, its control green | **10 failed, 26 passed** |

**Declared narrowing**: lint was run per-package rather than repo-wide. The population is eslint's own (26 files, counted from `--format json`, not guessed), and type-aware linting is not enabled — `languageOptions` carries only `ecmaVersion` and `globals`, no `project` / `projectService` — so a comments-and-tests diff inside one package cannot move any untouched file's verdict. CI runs the repo-wide scan regardless.

## Out of scope

#7164 is not addressed here and remains open. #7036's `Array.isArray` is untouched, and #6907's deliberate choice there was not re-opened. #6781's accept set does not move.

---

### DIFF
--- a/.changeset/gantt-row-walk-nontotal-reads.md
+++ b/.changeset/gantt-row-walk-nontotal-reads.md
@@ -1,0 +1,4 @@
+---
+---
+
+Documentation and pinned tests only; no published behaviour changes. `plugin-timeline`'s gantt row walk now carries the measured enumeration of its non-total reads — six in `findUnusableGanttDate`, three in `calculateDateRange` — in place of the prose count that was short, and the adversarial set exercises both classes instead of asserting them.
--- a/packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx
+++ b/packages/plugin-timeline/src/__tests__/timeline-gantt-date-brand-7027.test.tsx
@@ -37,9 +37,20 @@
  *      own headline was wrong: `Array.isArray` is not the last non-total
  *      operation on the PATH, only inside that function (objectui#7153).
  *
+ *   6. objectui#7153: the UPSTREAM reads. That card said five sites; driven in
+ *      render there are six in `findUnusableGanttDate` and three more in
+ *      `calculateDateRange`, and a tenth is NOT MEASURED. It took the same
+ *      stated-and-exercised branch for the six (`pin 5`) after re-testing the
+ *      `catch` argument rather than inheriting it — measured, a `catch`
+ *      upstream reads the PATH half and still substitutes the VALUE half, and
+ *      buys 3 of 9. And it falsified the class claim as well as the count: the
+ *      three in `calculateDateRange` are plain JSON, not exotica, so they are
+ *      pinned as a DEFECT (`pin 6`, objectui#7164) and not as an exclusion.
+ *
  * ⚠️ So this file's input set is not a claim of totality either. It is the
- * boundary that has actually been exercised, and the sixth card will be the
- * one that says where it ends.
+ * boundary that has actually been exercised, and the seventh card will be the
+ * one that says where it ends. Every entry above was written as the settled
+ * answer; five of the six were moved by the next card's measurement.
  *
  * ## What "a defined outcome" means here — two kinds, and never a third
  *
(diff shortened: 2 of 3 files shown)


os-warren:

## The finding that outranks the card

The card said five upstream sites. **It is nine** — six in `findUnusableGanttDate`, three in `calculateDateRange`, each attributed by the stack frame it threw from, plus a tenth identified by reading and honestly returned NOT MEASURED.

But the count is not the result. **They are not one reachability class**, and that is the thing five previous cards missed:

`calculateDateRange` re-walks the same rows **bare** — `items.flatMap`, `row.items`, no optional chaining. So `items: [null]`, or a row whose `items` is a truthy non-array, **crashes the render from ordinary JSON** that the declared zod mirror accepts and that `ObjectTimeline` passes through on a truthiness check alone.

⇒ The shared justification this whole family has been decided on — *"JSON cannot spell the input, therefore p3, therefore document it"* — is **sound for the exotic six and false for these three**. Filed as objectui#7164, which I have graded **p1 `bug`** and escalated as `needs-user-decision`.

Every card in the chain measured its own scope honestly. What nobody did until now was **split the sites by reachability class**, and doing so is what exposed that a correct-looking disposition was resting on a premise that only covered part of its population.

## My ruling was re-tested, and it came out differently in half

I said: ⛔ do not inherit objectui#7036's catch argument — **re-test it here**. You did, and it splits:

- A catch upstream **genuinely READS the path**. Because it is built from loop counters, the throwing-getter input came back **named exactly**: `items[0].items[0].endDate`. The speller's catch can never do that.
- But it still **substitutes** the value half, and buys **3 of 9**.

So the argument transfers in one half and fails in the other. Inheriting it wholesale would have been wrong, and so would rejecting it wholesale. That is precisely why the fence said re-test.

## The docblock branch, chosen by measurement rather than by precedent

Three readings, not one:

1. **The relocation ablation is the decisive one.** A repair confined to `calculateDateRange` — the whole of what this card scopes for that function — closes **zero** of the three and **moves all three into the render loop** at `1235/1237/1246`. *The exact relocation this path has performed four times.* Shipping it would have produced a sixth card and looked like progress.
2. The catch split above: 3 of 9, with the value half still substituted.
3. The reachability sweep still holds **for the six**.

Two pins, and the second is the important one: pin 6 records the three explicitly **as a DEFECT**, so objectui#7164's repair cannot land quietly. That is the right way to leave a known-bad path you are not fixing.

## ⚠️ You caught an instrument lying before believing its zero

`git grep -c | wc -l` counts **files**, and it scored prose inside `renderer.tsx`'s own docblock as a `Proxy` call site. You replaced it with a comment-stripped line counter **before any zero was believed**.

That is the third self-caught broken instrument in this seat today, and the most dangerous shape: it would have inflated a count rather than zeroing one, so it would have read as *more* evidence, not less. The corrected sweep then carries live controls on both sides — narrow zeros against `Object.assign` 19 / `JSON.parse` 105, and the same query widened repo-wide returning 1 / 21 / 105, so the narrow zeros are readings.

## Zone 2 honoured

I asked for no sixth confident totality sentence on a path with five claims and four falsifications. Four ablations, each predicting direction before running, each proving mutation by **deleted-marker count AND moved blob hash**, each proving restore by state, each asserting the **passing control beside the red** — including two marker-mobility runs (`9 failed | 27 passed`, `10 failed | 26 passed`) that establish the pins can actually move.

Changeset quoted verbatim: empty frontmatter. `check:sdui-registration-pins` named NOT MEASURED and correctly identified as unrelated to a diff that registers nothing.

Will arm on green.

---

PR #7172, opened 2026-09-01
fix(plugin-detail,i18n): RecordComments and PointInTimeRestore resolve their copy from the packs

### LINKED ISSUE #7163: `RecordComments` and `PointInTimeRestore` carry the same untranslated `formatTimestamp` `ActivityTimeline` just had — keys already exist in all ten packs
Found while implementing #7149 (PR #7162), which fixed this exact defect in `packages/plugin-detail/src/ActivityTimeline.tsx`. Filed rather than swept in: those files are outside that card's declared lane.

## Measured on `origin/main` head `85b5077d7`

Two more files in `packages/plugin-detail` hold a byte-identical copy of the same `formatTimestamp` helper, still returning hardcoded English:

`packages/plugin-detail/src/RecordComments.tsx:28-45`

```
if (diffMins < 1) return 'just now';
if (diffMins < 60) return `${diffMins}m ago`;
if (diffHours < 24) return `${diffHours}h ago`;
if (diffDays < 7) return `${diffDays}d ago`;
```

`packages/plugin-detail/src/PointInTimeRestore.tsx:38-48` — the same, minus the days branch (it falls through to `toLocaleString()` after hours).

## Why they are not equivalent to each other

- **`RecordComments.tsx` already uses the translation hook** — 11 `useDetailTranslation` / `t('detail.…')` references in the file. So it is precisely the `ActivityTimeline` shape #7149 described: a component that *is* wired to the packs, with one helper that never got converted. A zh session reads translated comment chrome next to an English `5m ago`.
- **`PointInTimeRestore.tsx` uses no translation hook at all** (0 references) — the wholly-unwired shape, so its scope is larger than a lookup swap.

## The keys already exist

No new keys are needed for the four relative-time strings — `detail.justNow`, `detail.minutesAgo` (`'{{count}}m ago'`), `detail.hoursAgo`, `detail.daysAgo` are present in **all ten packs** and already translated. Verified by reading the pack objects directly, not by a dotted-key grep (the packs are nested, so a grep for `detail.justNow` returns a false zero against the `en` pack that defines it); positive control `detail.back` / `detail.noActivity` / `detail.edit` = 10/10, negative control `detail.zzzAbsentControl7149` = 0/10.

Three call sites in this same package already consume them exactly this way — `RecordActivityTimeline.tsx:152-157`, `RecordMetaFooter.tsx:51-57`, `ThreadedReplies.tsx:37-42` — so the target shape is established and needs no design:

```
if (diffMins < 1) return t('detail.justNow');
if (diffMins < 60) return t('detail.minutesAgo', { count: diffMins });
```

`RecordComments` is therefore a pure lookup swap. `PointInTimeRestore` additionally needs the hook wired, and whoever takes it should sweep that file for its other literals rather than converting only the timestamps.

## Not in scope here, but worth a look together

`apps/console/src/pages/system/ApprovalsInboxPage.tsx:647-649` reaches the same copy through `tr('justNow', 'just now')` — an inline fallback rather than a bare key. Different pattern, possibly deliberate; noted only so a sweep does not miss it.

## Evidence bar for whoever takes this

An `en`-only test is green *before* the fix too — each key's `en` value is byte-identical to the literal it replaces — so English cannot discriminate a pack lookup from a hardcoded string. Assert in zh/ja/ar, and assert the provider-less path separately (it must yield the English default from `DETAIL_DEFAULT_TRANSLATIONS`, never a raw key, and it must interpolate `{{count}}`).

Never an inline `defaultValue` (#3517).

os-warren:
Fixes #7163

Measured on `origin/main` head **`d8ec8d6d4`** (PR #7162's landing sha — the card's line numbers predate it, so everything below was re-derived, not inherited). Final commit **`42f6c1480`**; every gate result quoted here was run on that tree with `git status --porcelain` empty.

## The card's "byte-identical" premise is FALSE — verified mechanically

The title and both PM comments say the two `formatTimestamp` copies are byte-identical. They are not:

```
md5  b36edffbf4d6719482f17e1743ac2a4a  RecordComments::formatTimestamp     (18 lines, 597 bytes)
md5  846a6b780ea858ee9650f7a0eadf6f6c  PointInTimeRestore::formatTimestamp (16 lines, 495 bytes)
diff exit=1 — 3 lines differ
```

`RecordComments` has a fourth `d ago` branch and falls through to `toLocaleDateString()`; `PointInTimeRestore` stops at hours and falls through to `toLocaleString()`. A revision list wants the time of day; a comment list does not. (The card *body* said as much — "the same, minus the days branch"; only the title and the dispatch generalised it.)

**This decides the sharing question by measurement**, per the scope ruling. Sharing is declined on two independent grounds: the two helpers have genuinely different tails, so unifying them would silently restyle one surface; and a shared helper needs a new export, which is Clause ② with the quota exhausted. Reported rather than routed around. The duplication is now three-way (`ActivityTimeline`, `RecordComments`, and this near-copy) and is filed separately as a finding.

## What shipped

**`RecordComments.tsx` — pure lookup swap, no new key, no copy change.** The file was already wired to the packs (11 `t('detail.…')` refs); only its helper was hardcoded. It now resolves `detail.justNow` / `minutesAgo` / `hoursAgo` / `daysAgo`. Each mapping was checked for **exactness**, not adopted on the card's word:

| literal | key | `en` pack value | verdict |
|---|---|---|---|
| `'just now'` | `detail.justNow` | `just now` | exact |
| `` `${diffMins}m ago` `` | `detail.minutesAgo` | `{{count}}m ago` | exact |
| `` `${diffHours}h ago` `` | `detail.hoursAgo` | `{{count}}h ago` | exact |
| `` `${diffDays}d ago` `` | `detail.daysAgo` | `{{count}}d ago` | exact |

**`PointInTimeRestore.tsx` — swept WHOLE, not partially wired.** It used no translation hook at all (0 refs), so per the ruling it was finished in one pass rather than becoming a third half-done component: card title, empty state, field-count line, preview heading, snapshot heading, both `(empty)` placeholders, the snapshot dash, the restore confirmation, and all three buttons. 17 static `t()` call sites; zero English literals remain (each removed anchor greps to 0).

Ten new keys land in **all ten packs** and in `DETAIL_DEFAULT_TRANSLATIONS`. Three existing keys are **reused rather than forked**: `detail.cancel`, `detail.activityEmptyValue` (the key `ActivityTimeline` already uses for an absent old/new value in a field-change diff — the same concept), and `detail.emptyValue`.

Two shape rules observed:
- The restore confirmation is **one key with a `{{when}}` hole**, not a sentence assembled around a JSX expression.
- The field-count plural is selected by a **static ternary over two literal keys**, never `t(KEYS[n])` — #7149 measured that a key visible only as a map value reads as unreferenced to the i18n scanners.

**One deliberate copy change, called out rather than quietly adopted:** the snapshot panel's null placeholder was an EN DASH (`–`, U+2013) written inline; it now resolves `detail.emptyValue`, which is an EM DASH (`—`, U+2014) in all ten packs — the glyph the rest of the detail package already uses.

## Pack verification method, with both controls

Read out of the **pack objects** by walking the nested path (`packages/i18n/src/locales/index.ts` -> `builtInLocales`), never a dotted-key grep — a grep for `detail.justNow` returns a false zero against the `en` pack that defines it, and the `detail` namespace is one of five in these packs that carry a `justNow` row.

- Loader liveness: `locales loaded: en,zh,ja,ko,de,fr,es,pt,ru,ar (count=10)`
- Positive controls: `detail.back` 10/10, `detail.noActivity` 10/10
- Negative control: `detail.zzzAbsentControl7163` **0/10**
- The four existing relative-time keys: **10/10 each**
- The ten new keys after the change: **10/10 each**, controls re-run in the same invocation and unchanged

## Reachability finding (this splits the card, as #7149's did)

`PointInTimeRestore` has **zero in-repo consumers**. Repo-wide, excluding `node_modules` and `dist`, it appears only in its own file and twice in the package barrel (`index.tsx:126` export, `:154` type export). Nothing renders it — not `DetailView`, not `apps/console`. `RecordComments`, by contrast, is rendered by `DetailView` at two call sites (`:1479`, `:1705`).

So the user-visible defect today is `RecordComments`; `PointInTimeRestore` is public API a downstream consumer can mount but nothing in this repo does. It was still swept whole rather than deferred: it is exported, so it is shippable surface, and finishing it now is what stops the #7142 shape from recurring.

## Nothing pinned the old literals — attributed, not counted

Repo-wide `grep -F` on `'just now'` and `'m ago'` returns hits across five packages. **Attributed** rather than counted: every test assertion among them belongs to `packages/collaboration` (a different package and namespace, `useCollaborationTranslation`) or to `ActivityTimeline.remainingLiterals.i18n.test.tsx` (#7162's suite, asserting on a different component). The two `plugin-detail` hits in `RecordMetaFooter.tsx` are **comments**, not literals — that file is already a correct `t()` call site. The `PointInTimeRestore` copy has exactly **one** hit per literal: the file itself. No test pinned either file's English.

## Evidence

**Ablation — direction predicted before running, mutation proven on disk, restore proven by state.**

Predicted: reverting only the two `.tsx` files to `d8ec8d6d4` (packs, defaults map and the new suites kept) turns the zh/ar assertions red while every `en` and provider-less assertion stays green **by construction**, since each `en` pack value is byte-identical to the literal being restored. Predicted 9 red / 8 green.

Observed, exactly:

```
Test Files  2 failed (2)
     Tests  9 failed | 8 passed (17)
```

- Mutation proven by **blob hash AND marker count**: `A(HEAD)=49c08e56…` vs `A(disk)=b5ce3ad9…`, `B(HEAD)=0db4d714…` vs `B(disk)=cac43cf7…`; injected marker `t('detail.justNow')` 0 in both, restored marker `return 'just now';` 1 in both, `Revision History` literal back to 1.
- Restore proven by **state**, not exit code: `git diff HEAD` produced **0 lines**, `git status --porcelain` **0 lines**, and both on-disk hashes match their HEAD blobs exactly. The script carried `trap restore EXIT INT TERM` with absolute paths resolved from `git rev-parse --show-toplevel`, and the restore leg names `HEAD` explicitly (a bare `git checkout -- path` restores *from the index*, which the mutation had already written).
- The 8 greens are green-by-construction and are kept deliberately: they guard that this was a lookup swap and **not** a copy change. Naming them: the `en` and provider-less legs of both suites, the two `{{count}}`-interpolation guards, and the two raw-key guards (a reverted file renders literals, not keys, so those cannot discriminate).

**Resolution path stated, since an ablation is only valid if it is measured:** the suites resolve **source**, not `dist/`. The components are same-package relative imports, and the root vitest config aliases `@object-ui/i18n` to `packages/i18n/src` (`vitest.config.mts:275`). No rebuild leg is required for this ablation, and the green run was measured against source too. The dependency closure was built anyway before the first run (`pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-detail^...' build`, exit 0).

**Green runs** (vitest paths root-relative per #3378, canonical invocation through the root config):

```
packages/plugin-detail/src/RecordComments.i18n.test.tsx
packages/plugin-detail/src/PointInTimeRestore.i18n.test.tsx
  Test Files  2 passed (2)       Tests  17 passed (17)

packages/plugin-detail/ packages/i18n/          (affected packages, whole)
  Test Files  182 passed (182)   Tests  2071 passed (2071)

all-locales-key-parity + de-quote-pairing-3876 + defaults-maps-mirror-en-pack
  Test Files  3 passed (3)       Tests  53 passed (53)
```

**Gates** — exit codes captured by redirect-then-capture, never read across a pipe; each verdict is the gate's own printed line:

- `check:i18n-keys` **exit 0** — *"Every in-scope call-site key resolves against the en pack (2865 keys), every literal inline defaultValue matches the value the pack serves, every call site passes exactly the arguments that value has holes for…"*
- `check:i18n-drift` **exit 0** — *"Compared the ten locale packs at d8ec8d6d4 (merge-base with origin/main) with the working tree: 0 en value(s) changed (10 key(s) added, 0 removed…)"*
- `check:i18n-dead-keys` **exit 0** (report, not a gate)
- `changeset:check` **exit 0** — *"All workspace packages are in the changeset fixed group."* / *"No changeset declares a `major` bump."*
- `check-changeset-presence.mjs` **exit 0** — *"15 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)"*
- `check-changeset-overwrite.mjs` **exit 0** — *"No pre-existing changeset was modified or deleted."*
- `check:control-bytes` **exit 0** — *"OK (scanned 5947 tracked text file(s); skipped 85 binary)"*; plus a direct `grep -naP` control-char scan over all 16 changed files: **0 hits**
- `type-check` (plugin-detail + i18n) **exit 0** — *"Scope: 2 of 47 workspace projects"* with both script names echoed, so not a zero-match silent pass
- `lint` (plugin-detail + i18n, plain per-package `eslint .`, no `--no-inline-config`) **exit 0** — warnings only, all pre-existing; the three on `PointInTimeRestore.tsx` are the `RevisionEntry` `any` fields at lines 25/27, which this diff does not touch

**Not NOT-MEASURED:** `type-check` runs `tsc -p tsconfig.test.json`, and `--listFiles` confirms both new suites are actually in that program (1 hit each; control `zzzNoSuchFile.tsx` = 0) — so the green genuinely covers the new test files rather than excluding them.

**The `de` quote-pairing census did not move.** The ⚠ carried forward from #7149 (a pack value with a quoted span shifts `de-quote-pairing-3876.test.ts`, 55 -> 58) does not apply: the German values added here contain no quote characters by construction, the census stays at **58**, and that file is untouched — verified by running it green rather than by assuming.

## Scope

Lane held: `packages/plugin-detail/src/{RecordComments,PointInTimeRestore}.tsx`, their new suites, and the ten locale packs. `ActivityTimeline.tsx` was read-only reference and is unmodified. No fenced package was entered. `content/docs/releases/` untouched.

---

### DIFF
--- a/.changeset/7163-detail-timestamp-i18n.md
+++ b/.changeset/7163-detail-timestamp-i18n.md
@@ -1,0 +1,42 @@
+---
+'@object-ui/plugin-detail': patch
+'@object-ui/i18n': patch
+---
+
+`RecordComments` and `PointInTimeRestore` resolve their copy from the locale
+packs instead of hardcoded English (objectui#7163).
+
+Both files carried their own `formatTimestamp` whose relative-time branches
+returned English literals, so a zh/ja/ar session read `5m ago` next to
+otherwise translated chrome — the defect objectui#7142/#7149 fixed one file
+over in `ActivityTimeline`.
+
+- **`RecordComments`** was already wired to the packs (11 `t('detail.…')`
+  references), so this is a pure lookup swap onto `detail.justNow` /
+  `minutesAgo` / `hoursAgo` / `daysAgo` — keys already present in all ten packs,
+  each `en` value byte-identical to the literal it replaces. **No new key, no
+  copy change.**
+- **`PointInTimeRestore`** used no translation hook at all, so it is swept
+  WHOLE rather than having only its timestamps converted: card title, empty
+  state, field-count line, preview panel, snapshot heading, restore
+  confirmation and all three buttons. Wiring one string into an otherwise
+  untranslated component is what shipped objectui#7142's visibly half-done zh
+  card; that is not repeated here.
+
+Ten new `detail.*` keys land in all ten packs and in
+`DETAIL_DEFAULT_TRANSLATIONS`: `revisionHistory`, `noRevisions`,
+`revisionFieldsChanged`, `revisionFieldsChangedOne`, `revisionPreview`,
+`revisionSnapshot`, `restoreConfirm`, `restoring`, `confirmRestore`,
+`restoreToPoint`. `detail.cancel`, `detail.activityEmptyValue` and
+`detail.emptyValue` are **reused**, not forked.
+
+The restore confirmation becomes one key with a `{{when}}` hole rather than a
+sentence assembled around a JSX expression, and the field-count line uses the
+repo's two-key plural convention selected by a static ternary over two literal
+keys — never `t(KEYS[n])`, which objectui#7149 measured as invisible to the
+i18n scanners.
+
+One deliberate copy change: the snapshot panel's null placeholder was an EN
+DASH (`–`, U+2013) written inline and now resolves `detail.emptyValue`, which
+is an EM DASH (`—`, U+2014) in all ten packs — the glyph the rest of the detail
+package already uses for an empty value.
--- a/packages/i18n/src/locales/ar.ts
+++ b/packages/i18n/src/locales/ar.ts
@@ -1059,6 +1059,21 @@ const ar = {
     showEmptyFields: "إظهار {{count}} حقل(حقول) فارغة",
     hideEmptyFields: "إخفاء الحقول الفارغة",
     noValue: "لا قيمة",
+    // objectui#7163 — PointInTimeRestore's revision-history chrome. The file
+    // used no translation hook at all, so every one of these read English in
+    // every session; swept in one pass rather than converting the timestamps
+    // alone. `Cancel`, `(empty)` and the empty-value dash reuse the keys this
+    // namespace already has, so only these ten are new.
+    revisionHistory: 'سجل المراجعات',
+    noRevisions: 'لا توجد مراجعات مسجلة',
+    revisionFieldsChanged: 'تم تغيير {{count}} حقول',
+    revisionFieldsChangedOne: 'تم تغيير {{count}} حقل',
+    revisionPreview: 'معاينة المراجعة',
+    revisionSnapshot: 'حالة السجل في هذه النقطة',
+    restoreConfirm: 'سيؤدي هذا إلى استعادة السجل إلى حالته في {{when}}. هل تريد المتابعة؟',
+    restoring: 'جارٍ الاستعادة…',
+    confirmRestore: 'تأكيد الاستعادة',
+    restoreToPoint: 'الاستعادة إلى هذه النقطة',
   },
   chart: {
     loading: "جارٍ تحميل الرسم البياني…",
--- a/packages/i18n/src/locales/de.ts
+++ b/packages/i18n/src/locales/de.ts
@@ -1052,6 +1052,21 @@ const de = {
     showEmptyFields: "{{count}} leere Felder anzeigen",
     hideEmptyFields: "Leere Felder ausblenden",
     noValue: "Kein Wert",
+    // objectui#7163 — PointInTimeRestore's revision-history chrome. The file
+    // used no translation hook at all, so every one of these read English in
+    // every session; swept in one pass rather than converting the timestamps
+    // alone. `Cancel`, `(empty)` and the empty-value dash reuse the keys this
+    // namespace already has, so only these ten are new.
+    revisionHistory: 'Versionsverlauf',
+    noRevisions: 'Keine Versionen aufgezeichnet',
+    revisionFieldsChanged: '{{count}} Felder geändert',
+    revisionFieldsChangedOne: '{{count}} Feld geändert',
+    revisionPreview: 'Versionsvorschau',
+    revisionSnapshot: 'Datensatzstand zu diesem Zeitpunkt',
+    restoreConfirm: 'Dadurch wird der Datensatz auf seinen Stand von {{when}} zurückgesetzt. Fortfahren?',
+    restoring: 'Wird wiederhergestellt…',
+    confirmRestore: 'Wiederherstellung bestätigen',
+    restoreToPoint: 'Auf diesen Zeitpunkt zurücksetzen',
   },
   chart: {
     loading: "Diagramm wird geladen…",
--- a/packages/i18n/src/locales/en.ts
+++ b/packages/i18n/src/locales/en.ts
@@ -1202,6 +1202,21 @@ const en = {
     showEmptyFields: 'Show {{count}} empty fields',
     hideEmptyFields: 'Hide empty fields',
     noValue: 'No value',
+    // objectui#7163 — PointInTimeRestore's revision-history chrome. The file
+    // used no translation hook at all, so every one of these read English in
+    // every session; swept in one pass rather than converting the timestamps
+    // alone. `Cancel`, `(empty)` and the empty-value dash reuse the keys this
+    // namespace already has, so only these ten are new.
+    revisionHistory: 'Revision History',
+    noRevisions: 'No revisions recorded',
+    revisionFieldsChanged: '{{count}} fields changed',
+    revisionFieldsChangedOne: '{{count}} field changed',
+    revisionPreview: 'Revision Preview',
+    revisionSnapshot: 'Record state at this point',
+    restoreConfirm: 'This will restore the record to its state at {{when}}. Continue?',
+    restoring: 'Restoring…',
+    confirmRestore: 'Confirm Restore',
+    restoreToPoint: 'Restore to this point',
   },
   chart: {
     loading: 'Loading chart…',
--- a/packages/i18n/src/locales/es.ts
+++ b/packages/i18n/src/locales/es.ts
@@ -1056,6 +1056,21 @@ const es = {
     showEmptyFields: "Mostrar {{count}} campos vacíos",
     hideEmptyFields: "Ocultar campos vacíos",
     noValue: "Sin valor",
+    // objectui#7163 — PointInTimeRestore's revision-history chrome. The file
+    // used no translation hook at all, so every one of these read English in
+    // every session; swept in one pass rather than converting the timestamps
+    // alone. `Cancel`, `(empty)` and the empty-value dash reuse the keys this
+    // namespace already has, so only these ten are new.
+    revisionHistory: 'Historial de revisiones',
+    noRevisions: 'Sin revisiones registradas',
+    revisionFieldsChanged: '{{count}} campos modificados',
+    revisionFieldsChangedOne: '{{count}} campo modificado',
+    revisionPreview: 'Vista previa de la revisión',
+    revisionSnapshot: 'Estado del registro en este punto',
+    restoreConfirm: 'Esto restaurará el registro a su estado del {{when}}. ¿Continuar?',
+    restoring: 'Restaurando…',
+    confirmRestore: 'Confirmar restauración',
+    restoreToPoint: 'Restaurar a este punto',
   },
   chart: {
     loading: "Cargando gráfico…",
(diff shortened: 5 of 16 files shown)


os-warren:

**Verdict: ACCEPT.** Arming for the queue on green. Three things in the report are the PM's to rule on rather than the implementer's, so I am ruling them here rather than leaving them in the PR body as open observations.

### ⭐ The ablation is the best-formed one this seat has reviewed this week

Direction predicted **before** running — 9 red / 8 green — and observed *exactly* (`Tests 9 failed | 8 passed (17)`). What makes it worth calling out is not the match, it is the **reason** the split was predictable: every `en` pack value is byte-identical to the literal it replaced, so `en` and provider-less assertions stay green **by construction** while `zh`/`ar` go red. That turns the 8 passes into a real instrument — they are the guard proving this was a **lookup swap and not a copy change**, which is precisely the claim an i18n PR most needs to establish and most easily fakes. Keeping them deliberately, and saying why, is the discipline working as intended.

Two details I want on the record because they are traps this seat has actually fallen into:

- **`git checkout -- <path>` restores from the INDEX, not HEAD** — and the mutation had already written the index. Naming `HEAD` explicitly is the difference between a restore and a no-op that reports success. This is the same failure family as this seat's own instrument failure #11 (a malformed control that printed **empty** and was read as a `0`).
- Restore proven **by state** (`git diff HEAD` 0 lines, `git status --porcelain` 0 lines, on-disk hashes equal to HEAD blobs), never by an editor's exit code; mutation proven by blob hash **and** marker count on both files. Correct on both legs.

The NOT-MEASURED check is also right: `--listFiles` confirms both new suites are **in** the type-check program (1 hit each, control `zzzNoSuchFile.tsx` = 0), so the green covers the new files rather than silently excluding them. And the rebuild leg was **measured** unnecessary (source aliasing at `vitest.config.mts:275`), not assumed — then the closure was built anyway. That is the right order.

### Ruling 1 — the whole-file sweep of a component with ZERO consumers: **upheld**

`PointInTimeRestore` is barrel-exported and rendered by nothing — not `DetailView`, not `apps/console`. The implementer swept all 17 sites anyway and flagged it as the kind of measurement that legitimately re-splits a card. Correct to flag; and the call to sweep stands, for the reason given: it is **exported public API**, and a third #7142-shaped half-done component is a worse outcome than translating a few strings nobody currently renders.

⚠️ But the measurement is more consequential than a scope note, and it does not belong buried in this PR. A zero-consumer exported component is exactly the **ADR-0049 enforce-or-remove** question — and this PR has just executed the *enforce* half without anyone asking whether *remove* was the right answer. That is not a reason to hold this PR (the work is done, correct, and cleanly revertible as one file plus its keys), but it is a reason to file the question. **Filing it now**, and linking it back here.

### Ruling 2 — the EN DASH → EM DASH copy change: **accepted, and reporting it was the right move**

`PointInTimeRestore`'s snapshot null placeholder was an inline EN DASH (U+2013) and now resolves `detail.emptyValue`, an EM DASH (U+2014) across all ten packs. This is a **user-visible copy change riding an i18n card**, which is normally exactly what I would send back.

It stands because the alternative is worse: forking a `detail.emptyValueEnDash` key to preserve one glyph would put a second spelling of one concept into ten packs — the multi-spelling class this repo is actively paying down (#7021 found the record-title key in three spellings). Reusing the key converges on the glyph the rest of the package already uses. ⭐ Reporting it rather than quietly adopting it is what makes it acceptable; an unreported glyph swap inside a 15-file i18n diff is invisible to review.

### Ruling 3 — `premise_still_valid: false` scoped to A3.2 only: **correct handling**

The "byte-identical copies" claim was false; A3.1, A3.3 and A3.5 confirmed; the implementable core was sound and shipped. Falsifying one PM assumption without stopping, and saying precisely which one, is the behaviour the three-zone order exists to produce. Zone 2 is there to be falsified.

### Out-of-scope handling: correct on both

- **#7173** (a fifth copy in `plugin-chatbot`, `Math.round` not `Math.floor`, 45s/30d thresholds) — filed rather than ridden, and the note that this is *genuinely different behaviour, not drift to normalise away* is the observation that keeps someone from "fixing" it into a bug later.
- **#7174** — and this is the call I most want to endorse: sharing one helper across packages **is Clause ② by construction** (a helper importable across packages is a new export), the fable quota is exhausted, so it was **recorded, not acted on**. That is the fence holding exactly where it should. The附带 observation that `Intl.RelativeTimeFormat` may beat sharing the hand-rolled version at all is worth more than the dedup itself.

### On the report's own accounting

`mcp_calls` reported as 9 with the note that the GitHub copy says 8 because the verifying read-back was necessarily the 9th. Trivial in itself; not trivial as a habit. A report that corrects its own count against its own artifact is a report whose other numbers I can spend less time re-deriving.

---

Nothing to change. Undrafting and arming auto-merge SQUASH once the four test shards, Type Check, Lint and Bundle Analysis report green — currently `in_progress`, nothing red.

---

PR #7111, opened 2026-09-01
feat(types)!: retire ChartDataSeries.data, correct categories' prose (#6896)

os-warren:
Fixes #6896

⛔ **Draft, and it stops here.** Clause ② applies: this narrows the accept set of a published
`@object-ui/types` surface, so it waits on contract review. Not ready, not enqueued, no
auto-merge — and the dispatch tier for this build was **opus** under the maintainer's standing
quota exemption (2026-08-13) after fable returned HTTP 429, so the merge-queue prohibition is
unchanged. `needs:contract-review` attached with this PR, not before it.

Implements the maintainer's ruling of 2026-08-31 (决裁批 #20 ③, 「其他同意」). Both halves land;
the ruling named half (a) alone as not a valid landing.

## (a) `ChartDataSeries.data` — RETIRED, announced

`data: number[]` was **required** on every authored series and read by nothing.
`normalizeSeries` (`packages/plugin-charts/src/normalizeChartSchema.ts:232`) reads
`dataKey`/`name`, `label`, `chartType`/`type`, `variant`, `opacity`, `dashArray`, `stack`,
`yAxis`, `color`. `data` is not among them. Rows come from the chart-level `data` — a key
`ChartSchema` never declared at all, surviving only because `BaseSchema` carries an index
signature, which is why no gate ever saw this.

`data: number[]` → **`data?: never`** plus `retirementTombstone()` on the Zod mirror. Deleting
the member was the option not taken: `ChartDataSeriesSchema` is a non-strict `z.object`, so a
deletion strips an authored value in silence — one silent no-op traded for another. Immediate
tombstone, no dual-reading window.

The accept set moves in **both** directions, both deliberate:

| | before | after |
|---|---|---|
| `{ name: 'Revenue', data: [1,2,3] }` | accepted | **refused**, named, with the remedy |
| `{ name: 'Revenue' }` | refused (`data` required) | **accepted** — the shape the renderer reads |

## (b) `categories` — live, with corrected prose

Not retired. It is read as an **alternative series list**, consulted only when `series` is
absent, each entry normalized as `{ dataKey }` — so the strings name columns to plot. The
category axis comes from `xAxisKey` / `xAxis` (`:290-292`). The docblock said "X-axis
labels/categories", so an author following the documentation got a different chart from the
documented one. Prose follows machine: the `ChartSchema.categories` docblock, the
`ChartDataSeries` header (which promised numbers "positionally aligned with the chart's
[...]

PR #7066, opened 2026-08-31
feat(types)!: retire DataTableSchema.toolbar — declared everywhere, read nowhere (objectui#6881)

os-sam:
Fixes #6881

Executes the maintainer ruling of 2026-08-31 (director batch 5, item 2), quoted verbatim from the card:

> **裁定:退役 `DataTableSchema.toolbar` —— 两个发布面(TS `data-display.ts:674` + zod 镜像)同笔墓碑化,⛔ 不接线。** 平台原则原文适用:已发布零消费的能力不因沉没成本获得豁免;要做 toolbar,须带着 enforcing reader 重新设计提案,⛔ 不在本卡复活。

CONTRACT REVIEW: this PR narrows a published validator's accept-set (ruling clause 4: "**条款②:yes**(收窄已发布接受集)——PR 建立时挂 `needs:contract-review` 走复审链;changeset 带 BREAKING 记号"). The label is applied at creation per that clause.

## Premises re-verified on this PR's base (85b495795), per the executor prerequisites

Line numbers re-located with git grep, not quoted from the card; the read-point census re-run with the card's positive control, not cited from it.

| Claim | Measured | Holds |
|---|---|---|
| TS face declares the key | `packages/types/src/data-display.ts:675` — `toolbar?: SchemaNode[]`, inside `DataTableSchema` (interface opens at line 650) | yes |
| Zod mirror declares it wider | `packages/types/src/zod/data-display.zod.ts:244` — union of one node or an array of nodes, `.optional()` | yes |
| Renderer reads it nowhere | `data-table.tsx` contains the word at lines 986 and 1211 only — both prose comments; no `schema.toolbar` read repo-wide (remaining hits are i18n keys such as `gantt.toolbar.refresh`, ARIA roles, and same-named React props: `PreviewShell` `toolbar`, `toolbarAddon`, `toolbarActions`) | yes |
| Positive control finds a real read | `emptyAction` — same interface, same SchemaNode shape — hits `data-table.tsx:2118-2119`, mounted through `SchemaRenderer`; the zero above is a reading, not a blind query | yes |
| No parity ledger watches it | `zod-mirror-parity.test.ts` — zero occurrences of the word across all four ledgers | yes |
| The key validates today | Measured before this change: array spelling ACCEPTED, single-node spelling ACCEPTED (see accept-set delta below) | yes |

## What changed — both faces, one stroke

[...]

PR #7176, opened 2026-09-01
fix(data-objectstack): declare the batch write-warning's unattributed-strip placeholder instead of a bare empty string

### LINKED ISSUE #7160: finding(data-objectstack): the batch write-warning names the object as an empty string when the response's `index` resolves to no operation
Filed unassigned by the `domain:ui` dev seat (session `session_012wwHa4aaFybxXrfmfHioDM`) while implementing objectui#6889. Scoped **out** of PR objectui#7159 deliberately — it is pre-existing, not introduced there, and it is a different fact from the two over-claims that card names. Recording so it does not die with the branch. Not graded here; grading is the triage seat's.

**Dedup before filing.** One targeted semantic search over this repo for the write-warning batch path, the empty resource placeholder and the wire entry's `object` returned four issues: objectui#6889, objectui#4934, objectui#3935 and objectui#7008. The first three are this topic's immediate neighbourhood, so the query reaches it — that hit set is the positive control, and the absence of an issue about the empty-string object name is a reading rather than a broken query.

## What

`ObjectStackAdapter.notifyBatchDroppedFields` resolves which object a cross-object batch strip belongs to as, in effect:

    the wire entry's `object` when it is a string,
    else the originating operation's `object`,
    else the empty string

The last arm is reachable: `op` is only looked up when the response entry carries a numeric `index`, and a response that omits `index` — or carries one that addresses no operation in the request — leaves `op` undefined. If the wire entry also omitted `object`, the emitted `WriteWarningEvent` names its `resource` as `''`.

After PR objectui#7159 that same empty string also lands on the notice's `object`, because the notice and the event now share one spelling of "which object is this". That is a strict improvement over the two expressions disagreeing, and it is what makes the placeholder worth a card: `DroppedFieldsEvent.object` is `z.ZodString` in the spec, so `''` satisfies the declared type while naming no object at all. Its `resource` twin has had the same property since the batch path was written.

## Why it is a finding and not a bug

- **No reader is harmed today.** `app-shell`'s `writeWarningToast` branches on `adapter && ev.resource`, so an empty resource simply skips label resolution and names fields by their api key — a documented, truthful fallback. Measured while implementing objectui#6889; the empty-resource arm was not driven directly, so treat that last clause as reasoned rather than executed.
- **Pre-existing.** The expression predates both objectui#4934 and objectui#7159.
- **Same family as objectui#6889, one level out.** That card is about a declared type claiming more than a runtime check establishes. This is about a runtime value that satisfies its declared type while meaning nothing — which the type system cannot see, so it needs a decision rather than a gate.

## Not measured

Whether any server actually returns a batch `droppedFields` entry with neither `index` nor `object`. Unanswerable from this repo — a host application outside it is invisible to any measurement runnable here. Do not let a triage pass infer it from server code that is not here.

## The shape a fix would take, if triage rules it worth doing

Three candidates, no recommendation attached — the pull question belongs to triage:

1. **Drop the entry.** Truthful, but it recreates objectui#3484's silence for a strip the server really did report.
2. **Keep the entry and stop claiming an object.** Needs `object` to become optional on the notice, which moves a published type and therefore engages the contract tier.
3. **Leave it and say so in a comment.** Cheapest; records that `''` means "no object named" so the next reader does not treat it as a real api name.

### Related

objectui#6889 (the over-claim card this was found under) · PR objectui#7159 (where the two expressions were unified) · objectui#4934 (the parent boundary card) · objectui#3484 (the silence that arm must not recreate)

os-warren:
Fixes #7160

Union of gates below re-run on the final commit `f05427d91`.

## The defect

`ObjectStackAdapter.notifyBatchDroppedFields` resolves the object a cross-object write-strip is about from the wire entry's own `object`, else from the operation its `index` addresses, else `''`. That last arm is a value satisfying the spec's required `object: string` while naming no object at all, and since PR objectui#7159 unified the two expressions it lands on both the notice's `object` and the event's `resource`.

## The gating measurement — driven, not reasoned

Driven end to end through the real chain: `ObjectStackAdapter.batchTransaction` (stubbed `client.data`) into a real `onWriteWarning` subscriber, into `app-shell`'s `emitWriteWarning` with the **real** `t` (a `createI18n` instance under `I18nextProvider`) and the **real** `fieldLabel` (`useSafeFieldLabel`, taken out of a rendered probe). Two-op batch: `account` create + `invoice` update.

| case | event | `getObjectSchema` called | what the user reads |
|---|---|---|---|
| `index: 99`, no wire `object` | `{"operation":"create","resource":"","droppedFields":[{"fields":["tax_rate"],"reason":"readonly_when","object":""}]}` | `[]` | `Saved — but some fields did not take effect` / `Not editable in this record's current state, so it did not take effect: tax_rate` |
| no `index` at all | identical to the above | `[]` | identical |
| **CONTROL** `index: 1` | `{"operation":"update","resource":"invoice","id":"inv1",...,"object":"invoice"}` | `[["invoice"]]` | `... did not take effect: Localized Tax rate` |

The control is alive and disagrees on every axis on the same instrument, so the empty call list is a reading and not a broken harness. `index: -1`, `1.5`, `0.5` and `NaN` all behave as `index: 99`.

**So the empty string is load-bearing, not inert.** `writeWarningToast` gates label resolution on `adapter && ev.resource`; the empty resource is *falsy*, so the schema lookup is skipped and fields are named by their api key — the documented truthful fallback. The card recorded that clause as reasoned rather than executed; it is now executed, and it holds.

## Reachability from a real response

Not reachable from a spec-conformant one. The spec's `CrossObjectBatchDroppedFieldsSchema` (`@objectstack/spec@17.2.0`) declares **both** `object: z.string()` and `index: z.number()` REQUIRED, and `CrossObjectBatchResponseSchema` documents `results` as index-aligned with the request's `operations`. Reaching the placeholder needs one entry that omits (or non-strings) `object` **and** carries an index naming no operation — off-spec twice over.

Unlike objectui#6889's exotic case this is not structurally impossible: the payload arrives as parsed JSON and a non-conformant server can send it. Whether a deployed backend does is not answerable from this repo, and is reported as NOT MEASURED.

## Which side of the refuse/tolerate line — and why it is neither

objectui#6889's repair is deliberately asymmetric with PR objectui#6884's `reason` arm:

- a `reason` from the future is the producer running **ahead** of us: expected version skew, so it is tolerated on an explicit arm carrying the wire value verbatim;
- a non-string `fields` element is off-spec input that would reach a consumer typed as a field name: refused here, fixed at the producer.

An unattributable strip is a third case, and the reason it is third is that **there is no producer value to keep or drop.** The response supplied nothing; the only question is what *we* write. So the answer is **declare**, not refuse and not widen:

**Refuse was measured and rejected.** Dropping the entry would replace a truthful, useful, user-visible warning — the save acknowledgement, the field list and the reason sentence all survive today — with silence for a strip the server really did report. That is objectui#3484's failure, and it is the stated reason neither `object` nor `reason` is gated on in the shared boundary.

**Widen was rejected as the banned lenient fallback.** Letting the notice say "no object" means making `object` optional on `DroppedFieldsNotice`, whose canonical arm *is* the spec's `DroppedFieldsEvent` (objectui#3160). That writes "servers may omit `object`" into our published client type in order to accommodate a producer violating two REQUIRED spec fields — precisely the consumer-side tolerance AGENTS.md #0.1 bans, fossilising the producer's bug into a second de-facto contract. The contract-first repair for an off-spec response is at the producer.

## What this PR does

- Replaces the bare `''` with a module-private, documented `UNATTRIBUTED_STRIP_OBJECT`, carrying the reachability argument, the refuse/widen rejections, and the reason the value **must stay falsy**: a namespaced sentinel in the style of `UNRECOGNIZED_DROP_REASON` would be truthful but *truthy*, and would send the sole consumer to `getObjectSchema('objectui:...')`. It is deliberately not exported — no consumer should branch on its identity; the falsiness check is the whole correct handling.
- Adds `packages/data-objectstack/src/droppedFieldsUnattributed.boundary.test.ts`, third sibling to the `reason` (objectui#4934) and shape (objectui#6889) boundary suites, pinning that the warning still reaches the subscriber, that the placeholder is falsy on both the notice and the event, that a wire `object` still wins, and that the single-record path cannot reach the placeholder at all. Every zero carries a live control.
- Corrects a comment that contradicted the code it sat on.

**The shared predicate was not extended, and did not need to be.** `isWireDroppedFieldsEntry` answers "is this wire value an entry at all", and this defect is not about the wire's shape — the entry is well-formed. It is about a *resolution* step downstream of the gate. Gating `object` in the shared predicate would also have wrongly dropped the single-record case, where a truthful heal exists. No third predicate was spelled.

## No published surface moves — proven

`packages/data-objectstack/dist/index.d.ts` built at the base sha and at this branch is **byte-identical** (`sha256 4796dca7c1a6c74...`), 389 declaration lines each, declaration-line diff empty.

Because the full files are identical, the usual "the full files do differ" control was unavailable, so a **live control on the same instrument** was run instead: appending one temporary exported const (marker count 0 to 1, blob `737d2dc` to `f0da4b0`) and rebuilding produced a 6-line full-file diff and a 5-line declaration-line diff. The instrument detects a real surface change; "identical" is therefore a measurement. The control was removed and `dist` rebuilt back to the head hash.

## Behaviour is unchanged — proven both ways

- **Ablation.** Direction predicted before running: mutating the placeholder to a truthy value turns exactly the two tests that read it red, leaving the file's other 6 and the 3 sibling suites green. Mutation confirmed on disk by marker count (`1` to `0` deleted / `0` to `1` injected) **and** blob hash (`737d2dc37c` to `3b658c6d6e`). Result: **`Tests 2 failed | 35 passed (37)`** — the two predicted, and no others. Restore proven by state, not exit code: `git diff HEAD` empty, blob back to `737d2dc37c`, mutant residue `0`. No build step is in this loop — the suite imports `./index` by relative path, so the subject is the source, not `dist`.
- **The new suite passes identically against the pre-change source** (8/8 with the file swapped to the base blob `1b93160c80`, marker count `0` confirming the swap landed). The pin therefore records existing behaviour rather than blessing new behaviour.

## Scoped out, filed rather than folded

The same trigger fabricates a second value: with no operation to read, `operation` lands on `'create'` (measured; the control reports `'update'`). It is filed as objectui#7170 rather than repaired here — there is no correct value to fall back to, and `WriteWarningEvent.operation` is a REQUIRED union on a published type, so an honest answer there hits the same wall. This PR pins the current value beside the live control so whichever disposition triage picks arrives as a visible diff. objectui#7170 remains open and is not addressed by this branch.

## Gates run, on `f05427d91`

| gate | verdict |
|---|---|
| `pnpm exec vitest run packages/data-objectstack/` | `Test Files 53 passed (53)` / `Tests 724 passed (724)` |
| `pnpm --filter @object-ui/data-objectstack type-check` | exit 0 — and `tsc --listFiles` confirms the new test file is in the program (1 hit), so this is measured coverage, not an exclusion |
| `pnpm --filter @object-ui/data-objectstack lint` | `✖ 421 problems (0 errors, 421 warnings)`, exit 0 — the warnings are the `any`-typed harness the sibling boundary suites already use |
| `check:changeset-presence` | `✅  2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` (empty frontmatter — declared as releasing nothing) |
| `check:control-bytes` | `✅  check-control-bytes: OK (scanned 5945 tracked text file(s); skipped 85 binary).` |
| `check:self-import` | `✅  No package names itself inside its own src/.` |
| `check:spec-symbols` | `✅  spec symbol derivation: 1332 files scanned against 4959 spec export names` |
| `check:vi-mock-specifiers` | `✅  check-vi-mock-specifiers: OK` |
| `check:phantom-deps` · `check:esm-specifiers` · `check:entry-guard` · `check:vi-mock-inherit` · `check:side-effects-array` | exit 0 |
| `check:readme-exports` | **NOT MEASURED locally** — the gate's own verdict is `the population COLLAPSED -- this run proves nothing`, with 353 complaints all of the form "type entry is not on disk -- run `pnpm build` first" and **0** naming `data-objectstack`. It needs a full monorepo build this worktree does not have; CI builds and owns it. |

Assumption falsified: none of A2.1 to A2.4. `f75810e7c` is an ancestor of the base `d8ec8d6d4` and the file carries the landed objectui#6889 work; the trigger is as stated; the object name is computed once; and nothing pinned the empty-string behaviour before this branch.

---

### DIFF
--- a/.changeset/olive-pears-invent.md
+++ b/.changeset/olive-pears-invent.md
@@ -1,0 +1,27 @@
+---
+---
+
+Declare the batch write-warning's unattributed-strip placeholder in
+`@object-ui/data-objectstack`, and pin what it is (objectui#7160).
+
+`notifyBatchDroppedFields` resolves the object a cross-object strip is about
+from the wire entry's `object`, else from the operation its `index` addresses.
+When neither channel answers, it wrote a bare `''` — a value satisfying the
+spec's required `object: string` while naming no object at all. That literal is
+now the named, documented `UNATTRIBUTED_STRIP_OBJECT`, carrying the reachability
+argument, why such a strip is still emitted rather than refused, and why the
+placeholder must stay falsy. A stale comment claiming an unattributable strip
+"reads as an update" is corrected to what the code does (it lands on `create`,
+tracked as objectui#7170).
+
+No published behaviour changes and no released surface moves: the value is
+unchanged, and the new boundary suite passes identically against the pre-change
+source. The emitted `dist/index.d.ts` is byte-identical at both shas, with a
+live control (one temporary exported const) proving the comparison detects a
+real surface change.
+
+`node scripts/check-changeset-presence.mjs` on this tree, verbatim:
+
+> ✅  2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/olive-pears-invent.md.
+>     Every one of them has an EMPTY frontmatter — declared as releasing nothing, which
+>     is the explicit exemption and a complete answer to this gate.
--- a/packages/data-objectstack/src/droppedFieldsUnattributed.boundary.test.ts
+++ b/packages/data-objectstack/src/droppedFieldsUnattributed.boundary.test.ts
@@ -1,0 +1,226 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * What the batch write-warning says about a strip it CANNOT attribute to an
+ * operation (objectui#7160).
+ *
+ * `notifyBatchDroppedFields` names the object a cross-object strip is about
+ * from the wire entry's `object`, else from the operation its `index`
+ * addresses. When the wire named none AND the index addresses no operation in
+ * the request we sent, there is nothing left to name it with, and the adapter
+ * writes its `UNATTRIBUTED_STRIP_OBJECT` placeholder onto both the notice's
+ * `object` and the event's `resource`.
+ *
+ * Third sibling to `droppedFieldsReason.boundary.test.ts` (objectui#4934) and
+ * `droppedFieldsShape.boundary.test.ts` (objectui#6889), and deliberately a
+ * DIFFERENT disposition from both, because the fact is different:
+ *
+ *  - `reason` from the future is the producer running AHEAD of us — expected
+ *    version skew, so it earns an explicit arm carrying the wire value verbatim.
+ *  - a non-string `fields` element is off-spec input that would reach a
+ *    consumer typed as a field name — refused here, fixed at the producer.
+ *  - an unattributable strip supplies NO producer value at all. There is
+ *    nothing to keep or drop; the only question is what we write in its place.
+ *
+ * So this suite pins the placeholder rather than removing it, and pins the two
+ * properties that make it safe:
+ *
+ *  - the warning is still EMITTED. Refusing it would trade a truthful,
+ *    user-visible warning about fields the server really did strip for silence
+ *    — objectui#3484's failure. Measured end to end for the PR: the user still
+ *    gets the save acknowledgement, the field list and the reason sentence,
+ *    with fields named by api key instead of label.
+ *  - the placeholder is FALSY. That is load-bearing, not incidental: the sole
+ *    consumer (`app-shell`'s `writeWarningToast`) gates label resolution on
+ *    `adapter && ev.resource`, so an empty resource skips the schema lookup and
+ *    names fields by their api key. A truthful-but-truthy sentinel would send
+ *    that consumer to `getObjectSchema('objectui:...')` instead. Nothing pinned
+ *    this before, so a "cleanup" replacing `''` could have broken the consumer
+ *    with every adapter test still green.
+ *
+ * REACHABILITY. Only from a response that is off-spec twice over: the spec's
+ * `CrossObjectBatchDroppedFieldsSchema` declares BOTH `object: z.string()` and
+ * `index: z.number()` required, and documents `results` as index-aligned with
+ * the request's `operations`. Nothing in this repo emits that shape, and
+ * whether a deployed backend does is not answerable from here. What is pinned
+ * is the DISCRIMINATION, not a population — every zero below keeps a live
+ * control beside it so a green run cannot be a silently-empty one.
+ */
+import { describe, it, expect, vi } from 'vitest';
+import { ObjectStackAdapter } from './index';
+import type { WriteWarningEvent } from './index';
+
+function makeDS(stub: Record<string, any>) {
+  const ds: any = new ObjectStackAdapter({
+    baseUrl: 'http://test.local',
+    fetch: vi.fn(async () =>
+      new Response(JSON.stringify({ success: true, data: { capabilities: {}, routes: {} } }), {
+        status: 200,
+        headers: { 'Content-Type': 'application/json' },
+      }),
+    ),
+  });
+  ds.connected = true;
+  ds.connectionState = 'connected';
+  ds.client = { data: stub };
+  ds.atomicBatchCapability = true;
+  return ds;
+}
+
+/**
+ * One two-op batch — the shape the console's record form saves a master-detail
+ * record with. Operation 0 creates an `account`, operation 1 updates an
+ * `invoice`, so a resolvable `index` has a distinct object to name.
+ */
+async function emitOnBatch(droppedFields: unknown[]): Promise<WriteWarningEvent[]> {
+  const batchTransaction = vi.fn().mockResolvedValue({
+    results: [{ id: 'acc1' }, { id: 'inv1' }],
+    droppedFields,
+  });
+  const ds = makeDS({ batchTransaction });
+  const events: WriteWarningEvent[] = [];
+  ds.onWriteWarning((e: WriteWarningEvent) => events.push(e));
+  await ds.batchTransaction([
+    { object: 'account', action: 'create', data: { name: 'Acme' } },
+    { object: 'invoice', action: 'update', id: 'inv1', data: { tax_rate: 7 } },
+  ]);
+  return events;
+}
+
+describe('an unattributable batch write-strip (#7160)', () => {
+  it('is still emitted — it does not recreate #3484 silence', async () => {
+    const events = await emitOnBatch([
+      { fields: ['tax_rate'], reason: 'readonly_when', index: 99 },
+    ]);
+
+    expect(events).toHaveLength(1);
+    // Everything the response DID establish survives: the user is told which
+    // fields did not take effect, and why.
+    expect(events[0].droppedFields).toHaveLength(1);
+    expect(events[0].droppedFields[0].fields).toEqual(['tax_rate']);
+    expect(events[0].droppedFields[0].reason).toBe('readonly_when');
+  });
+
+  it('names the object with the falsy placeholder on BOTH the notice and the event', async () => {
+    const events = await emitOnBatch([
+      { fields: ['tax_rate'], reason: 'readonly_when', index: 99 },
+    ]);
+
+    // One spelling for both, as objectui#6889 unified them.
+    expect(events[0].resource).toBe('');
+    expect(events[0].droppedFields[0].object).toBe('');
+    // The property the only consumer actually depends on. `writeWarningToast`
+    // gates label resolution on `adapter && ev.resource`, so a truthy
+    // placeholder would send it to `getObjectSchema(<placeholder>)`.
+    expect(events[0].resource).toBeFalsy();
+    expect(events[0].droppedFields[0].object).toBeFalsy();
+    // Still a string: `DroppedFieldsEvent.object` is `z.string()` in the spec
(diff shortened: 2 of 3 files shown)


os-warren:

**Verdict: ACCEPT.** Undrafting and arming on green. One open question is ruled below; it is the PM's, not the implementer's.

### ⭐ The control I want on the record, because it is the hard case

Proving `dist/index.d.ts` is **byte-identical** at base and head is how this PR establishes it never entered `CONTRACT_REVIEW_TIER`. The problem is that a byte-identical comparison has **no natural positive control** — the usual "the full files DO differ while the declaration lines do not" reassurance is unavailable precisely when the answer is "identical", which is exactly when you most need to know the instrument works.

The implementer noticed that and built one: append a temporary exported const (marker 0 → 1, blob `737d2dc` → `f0da4b0`), observe a 6-line full-file and 5-line declaration-line diff, then remove it and rebuild back to the head hash. That converts *"the diff was empty"* into *"the diff was empty **and** this instrument produces a non-empty diff when a change exists."*

This is the same failure family as this seat's own instrument failure #11 — a control that printed **empty** and was read as a `0`. And the report shows the lesson landed twice over: *"the first attempt printed EMPTY marker counts from broken shell quoting and was discarded as void rather than read as zero."* Discarding a void reading instead of scoring it is the whole discipline in one sentence.

### On the disposition: REFUSE / TOLERATE / **DECLARE**

The card offered two dispositions and the answer was neither. That is the right kind of falsification — not "the premise is wrong" but "the option set was incomplete." The reasoning holds up:

- There is **no producer value to keep or drop**. The response supplied nothing; the only question is what *we* write. That genuinely is a third case, and it is why refuse/tolerate both mis-frame it.
- **Refuse was measured and rejected**, not waved off — dropping the entry trades a truthful, user-visible warning for objectui#3484's silence. The gating measurement drove the full chain with the real `t` and real `fieldLabel` and showed the user still reads the complete toast with fields named by api key, `getObjectSchema` never called (call list `[]`), against a live control at index 1 resolving `invoice` / `update` / `Localized Tax rate`. That is a behavioural reading, not an argument.
- **Widen was rejected on principle rather than cost**, which is the part I most want to endorse. Making `object` optional on `DroppedFieldsNotice` — whose canonical arm *is* the spec's `DroppedFieldsEvent` — would write *"servers may omit `object`"* into our published client type to accommodate a producer violating two REQUIRED spec fields. That is the lenient consumer-side fallback AGENTS.md #0.1 bans, and it would cost the canonical arm its identity as THE spec type, which objectui#3160 fought for. ⭐ Note it would *also* have crossed Clause ② and been unlandable at today's tier — but the report reaches the right answer without leaning on that, which is the stronger position.
- Keeping the placeholder **falsy** is load-bearing and now documented: `writeWarningToast` gates on `adapter && ev.resource`, so a namespaced sentinel would be truthful but truthy and would send the consumer to `getObjectSchema` on a non-object. That is the kind of constraint that gets silently broken by a future "cleanup", which is exactly why declaring it beats leaving a bare `''`.

### Ruling on the open question: **A — leave as landed**

*Does any deployed backend actually return a batch `droppedFields` entry that omits `object` and carries an index naming no operation?* **NOT MEASURED**, and correctly reported as such rather than guessed.

Option **A** stands, for the implementer's reasons plus one of mine: **B is unavailable at this seat's tier regardless of merit** (Clause ②, and the `CONTRACT_REVIEW_TIER` quota is currently exhausted — it is what is holding #6896/PR #7111, #6881/PR #7066 and the retirement pile), and **C deletes a warning that measurement shows is fully truthful today**. A expands no published surface, adds no declared capability, and the declared-plus-pinned form closes the actual hazard — which was never the empty string, but that it was **undeclared**.

⚠️ **Carrying the implementer's warning into the record verbatim, because it is the part most likely to be lost:** *"Do not let a later triage pass infer the population from server code that is not in these repos."* The placeholder is unreachable from a **conformant** response, but — unlike objectui#6889's case — it is not structurally impossible, because the payload arrives as parsed JSON. A future pass must not upgrade "the spec forbids it" into "no server does it" without reading a server. That distinction is the whole content of this NOT MEASURED, and it is not resolvable from either repo.

### Also correct

- **Shared predicate not extended, and the report explains why that is right** rather than treating non-reuse as a gap: `isWireDroppedFieldsEntry` answers *"is this a wire entry at all"*, and this entry is well-formed — the defect is in a resolution step downstream of the gate. Gating `object` there would have wrongly dropped the single-record case, where a truthful heal exists. **No third predicate was spelled**, which is the outcome objectui#6889 was landed to protect.
- **Behaviour neutrality proven the right way round**: the new suite passes 8/8 against the *pre-change* source (file swapped to base blob `1b93160c80`, marker count 0 confirming the swap landed). The pin records existing behaviour instead of blessing new behaviour — and that is what makes it a pin rather than a spec.
- **#7170 filed, not folded.** The fabricated `operation: 'create'` has no correct fallback value (`WriteWarningEvent.operation` is a REQUIRED published union), and the code comment above that line asserts the opposite of what the measurement shows. Pinning the current value beside the control so any future disposition lands as a visible diff is better than either fixing it here or leaving it unrecorded.
- Ablation: direction predicted **before** running, mutation proven by marker count **and** blob hash, restore proven by **state** (`git diff HEAD` empty, blob back to `737d2dc37c`, residue 0), `Tests 2 failed | 35 passed (37)` — the two predicted and no others, with the passing count asserted alongside the red. No build in the loop, and that was **stated as measured** (the suite imports `./index` by relative path, so the subject is source) rather than assumed.

---

Nothing to change.

---

PR #7180, opened 2026-09-01
fix(components): page:header resolves declared action ids

### LINKED ISSUE #6252: page:header implements the declared action-id lookup — resolve `actions: string[]` like `record:quick_actions` does (objectstack#11592 ruling)
Blocked-by: objectstack-ai/objectstack#12580

Implementation of the objectstack#11592 ruling (maintainer, 2026-08-25, verbatim 「全部同意」 on recommendation B; full record = objectstack#11592 comment 5405065025). Filed by PM session `session_01KWRU3s15AJz7PGW7a7wdCh` per the confirmed checklist. Parent: objectstack#11592 (tracking).

**Step 1 — the equivalence measurement, and it gates the rest**: compare the exemplar's 16 inline ActionDef objects (hotcrm, 4 record pages, all `locations: ['record_header', …]`) against their object-level registered actions. If any inline def carries page-level customization an id-resolved action cannot express, ⛔ stop and report back to objectstack#11592 — that evidence reopens the ruling; do not silently absorb it with an override semantic.

**Content**: the canonical `page:header` renderer resolves `PageHeaderProps.actions` as **action ids** against the object's own metadata — the contract the spec declares (`z.array(z.string())`) and two sibling renderers already implement (`layout:page-header` delegating to `record:quick_actions`; the resolver pattern lives in `packages/plugin-detail/src/renderers/record-quick-actions.tsx`). Reuse that pattern, do not invent a second resolver. Keep the existing object-shape handling working during the transition if it is cheap; the contract of record is ids.

**Acceptance criterion**: a `page:header` authored with action ids renders the same buttons the object-shape authoring renders today (filtered by `actionRendersAt(a, 'record_header')`, honouring `requiredPermissions` / `visible` / `order`); the renderer's tests gain the id-shaped cases; the id path carries no `body.source` into the built artifact.

**Downstream**: hotcrm migrates its 16 inline objects to id references once this lands (its lane's follow-up; the conformance-sweep exemption naming objectstack#11592 comes off then); hotcrm#1279 unlocks via the parent.

os-warren:
Fixes #6252

Implements the objectstack#11592 ruling (maintainer, 2026-08-25, on recommendation B): the canonical `page:header` renderer resolves `PageHeaderProps.actions` as **action ids**.

## The defect

`@objectstack/spec` declares `actions: z.array(z.string()).optional().describe('Action IDs to show in header')` (verified on the installed pin, `@objectstack/spec@17.2.0`, `src/ui/component.zod.ts`). The renderer read that array as `ActionDef` **objects** and resolved nothing, so a header authored the way the published contract declares rendered **zero buttons** — satisfying the schema deleted the header.

Two sibling surfaces already read it as ids: `record:quick_actions` (`packages/plugin-detail/src/renderers/record-quick-actions.tsx`) resolves a string-valued `actions` out of the object's metadata, and `layout:page-header` reaches the same resolver by delegating its `actions` into a `record:quick_actions` node (`packages/layout/src/PageHeader.tsx:196-208`). So one authoring key meant two different things depending on which renderer drew the header.

## What changed

`packages/components/src/renderers/layout/containers.tsx`:

- Each element of `actions` is normalised at the **top** of the pipeline — a string is resolved against the object's own `actions` metadata, an object passes through. Resolution goes through the **same `useMetadataItem` entry** `record:quick_actions` uses; no second resolver, no second lookup path.
- Because normalisation happens above the filters, **one** chain still runs: `actionRendersAt` placement (`record_header` / `record_more`), the `requiredPermissions` capability gate, `visible` / `hidden`, `order`, and the inline/overflow split are untouched by this change. That is what makes the equivalence claim testable at all.
- An id that resolves to nothing renders nothing and **warns once**, naming the object's declared action names. Silently dropping it would reproduce the silent-loss class this repo keeps re-filing; it also makes a mistyped id indistinguishable from a correctly hidden one. The warning is suppressed while the metadata read is in flight, so a correctly-authored page does not warn on first paint.
- The registered `actions` input description now says ids.

The one deliberate difference from `record:quick_actions`: that renderer switches on the whole array (`rawActions.every(a =&gt; typeof a === 'string')`); this one normalises **per element**, so a half-migrated `['convert', { … }]` resolves the id and passes the object through. Same mechanism, wider arity — a page mid-migration is exactly the state this card creates.

**Clause ② is not engaged.** The declared type and its zod mirror are untouched — `actions` stays `z.array(z.string())`. The object shape survives as **renderer tolerance**, exactly as it does today, undeclared. No accept/reject behaviour of the contract moves.

Docs: `content/docs/guide/slotted-pages.md` now teaches the ids contract on the canonical header (it previously showed only the object form), with the inline-object form named as migration tolerance.

## Acceptance criteria — evidence

Run at `af40cf5c2` (working tree clean at that commit, so every run below read exactly this tree).

**1. An id-authored header renders the same buttons as the object-shape authoring.**

The same action metadata is authored twice and the two renders compared. The object-shape render is the **live control**: each equivalence case asserts it is non-empty first, so an id-side green can never come from two empty headers agreeing.

The population covers every filter the criterion names, so one equivalence assertion measures all of them: `list_only` (`list_item` — must not render), `archive` (`record_more` — must land in the ⋯ menu, not inline), `gated` (`requiredPermissions` the user lacks), `closed_only` (a CEL `visible` false for this record), and `qualify` (`order: 1`) rendering **before** `convert` (`order: 2`), which is the reverse of the order both authorings list them in. Both the ordered button-name list and a normalised `innerHTML` projection are asserted equal.

**2. The renderer's tests gain the id-shaped cases.** `packages/components/src/__tests__/page-header-action-ids.test.tsx` — 8 cases (equivalence, the filter chain, `record_more` routing, unresolved-id warn-once with a resolvable sibling as control, no-warn-while-loading, mixed arrays, the `properties.actions` bridge spelling, and criterion 3).

```
pnpm exec vitest run --project dom-heavy packages/components/src/__tests__/page-header-action-ids.test.tsx
  Test Files  1 passed (1)        Tests  8 passed (8)
```

**Ablation** — direction predicted before running: red on the id path, green on the object-shape control. Mutation = feeding `rawHeaderActions` back into the filter chain (the minimal deletion of the fix). Proven on disk by marker count **and** blob hash, never by an editor's exit code:

```
PRE  target-count=1  inject-count=0   PRE_HASH=3dafeb94… (== HEAD blob)
POST target-count=0  inject-count=1   POST_HASH=f87a37c9…
Tests  7 failed | 1 passed (8)
```

7 failed / 1 passed matched the prediction exactly: the only case that cannot move is "does not warn while the metadata lookup is still in flight", which is true with or without resolution. Restore proven by **state**, not exit code: restored blob `3dafeb94…` equals the HEAD blob, and `git diff HEAD`, `git diff --cached`, `git status --short` are all empty. The mutation and restore legs need no rebuild here: the root `vitest.config.mts` aliases `@object-ui/components` to `packages/components/src`, so this suite resolves source, not `dist`.

**3. The id path carries no `body.source` into the built artifact.**

Measured two ways, and the built-artifact one is the load-bearing half.

*Against the built artifact.* `pnpm --filter @object-ui/components build`, then a run that mounts the **built** `packages/components/dist/index.js` (an explicit relative import — the workspace alias would otherwise redirect to src) with an id-authored header whose resolved action carries `body: { language: 'js', source: '…OS6252_DIST_BODY_MARKER…' }`. The id resolved through the built renderer, the marker appears nowhere in the rendered DOM, and the authored node's `JSON.stringify` is byte-identical before and after render. **Live control:** with that one import removed the same file fails `expected undefined to be truthy` — `page:header` is unregistered in the light `dom` project, which proves the passing run measured the built bundle and nothing else. This measurement is **not committed**: turbo's `test` task is `dependsOn: ["^build"]` (dependencies only, not the package's own build), so a committed `dist`-importing test would be `NOT MEASURED` in CI rather than a pin. Filed as a follow-up.

*As a committed pin.* `never writes a resolved def — and so no body.source — back onto the authored node` asserts the authored node is unchanged by rendering and the marker is absent from the DOM, with the object-shape authoring as the live control: `expect(JSON.stringify(objectAuthored)).toContain(BODY_MARKER)` — the inline form **does** drag the handler body into what it serializes, which is what proves this assertion can fail at all.

Grep of the built bundle, with controls: `did not resolve` = 1 and `page:header` = 7 (present), `OS6252_HANDLER_BODY_MARKER` = 0.

## Other verification

| Run | Result |
|---|---|
| `vitest run packages/components scripts/__tests__/vitest-invocation-guard.test.ts` | 224 files, **2070 passed** |
| all `page-header*` + `RecordDetailView.header*` / predicate suites | 16 files, **184 passed** |
| console registry/contract parity (5 files incl. `registry-inputs-spec-parity`, `public-contract`) | **156 passed** |
| `pnpm --filter @object-ui/components run type-check` | exit 0 — and it covers the new test file (`tsc -p tsconfig.test.json` reported errors *in* it before they were fixed) |
| `pnpm check:control-bytes` | ✅ 5945 tracked text files |
| `pnpm check:doc-types` | ✅ every documented component type registered |
| `pnpm check:doc-fences` | ✅ |
| `pnpm check:doc-snippets` | ✅ 272/272 blocks judged, 0 failed (after building its declared closure; its own resolution/sentinel/positive controls printed) |
| `pnpm check:sdui-registration-pins` | ✅ all 16 registrations present (after building the console + its closure) |

`eslint` on all three changed source files: **0 errors** (5 pre-existing-style `no-explicit-any` warnings in the new test, matching its sibling page-header tests; `--max-warnings` is deliberately unset in this repo). The lint run is narrowed and the narrowing is measured, not assumed: the population is the diff itself (`git diff --name-only`), the count comes from `--format json` (3 files linted), and the config is not type-aware — `eslint.config.js` contains zero `projectService` / `parserOptions` / `project:` occurrences against a live control of 10 `rules` hits in the same file — so no untouched file's verdict can move under this diff, which touches no config. The repo-wide farm is CI's run either way.

## Not in this PR

hotcrm's migration of its 16 inline `record_header` objects to id references is downstream of this landing, as the card records. Nothing here touches it.

---

### DIFF
--- a/.changeset/6252-page-header-action-ids.md
+++ b/.changeset/6252-page-header-action-ids.md
@@ -1,0 +1,35 @@
+---
+'@object-ui/components': minor
+---
+
+`page:header` resolves its `actions` as declared ACTION IDS (objectui#6252,
+implementing the objectstack#11592 ruling — maintainer, 2026-08-25, on
+recommendation B).
+
+`@objectstack/spec`'s `PageHeaderProps.actions` is
+`z.array(z.string()).describe('Action IDs to show in header')` and has been for
+as long as the key has existed. The canonical renderer read that array as
+`ActionDef` objects and resolved nothing, so a header authored the way the
+published contract declares rendered **zero buttons** — satisfying the schema
+deleted the header. Two sibling surfaces already read it as ids
+(`record:quick_actions`, and `layout:page-header` by delegating to it), so one
+authoring key meant two different things depending on which renderer drew the
+header.
+
+Each id is now resolved against the object's own `actions` metadata through the
+same `useMetadataItem` entry `record:quick_actions` uses. Resolution happens at
+the top of the actions pipeline, so the existing chain — `record_header` /
+`record_more` placement, the `requiredPermissions` capability gate, `visible` /
+`hidden`, `order`, and the inline/overflow split — runs unchanged over
+uniformly-shaped defs: an id-authored header and an object-authored one converge
+before a single filter runs.
+
+- Inline `ActionDef` objects keep rendering, per element, so a half-migrated
+  `['convert', { … }]` array resolves the id and passes the object through. This
+  is renderer tolerance for the migration and stays undeclared — the contract is
+  ids.
+- An id that resolves to no action renders nothing and warns **once**, naming
+  the object's declared action names. A mistyped id is no longer indistinguishable
+  from a correctly hidden one.
+- Nothing is written back onto the authored node, so an id-authored page carries
+  no `ActionDef` — and no `body.source` handler body — into what it serializes.
--- a/content/docs/guide/slotted-pages.md
+++ b/content/docs/guide/slotted-pages.md
@@ -103,9 +103,47 @@ Delete) as a button row. Up to **`maxVisible`** actions render inline,
 side by side (default **3** on desktop, **`mobileMaxVisible`**, default
 **1**, on mobile); the rest collapse into a `⋯` "More actions" menu.
 
+### Naming the actions: `page:header.actions` holds IDS
+
+`PageHeaderProps.actions` is a list of **action ids** — the `name` of an
+action declared on the object's own metadata. The header resolves each id
+against that object, which is the same lookup `record:quick_actions`
+performs, and keeps the definitions in exactly one place: change an action
+once on the object and every page that names it follows.
+
+```ts
+// on the object: the definitions
+actions: [
+  { name: 'convert_lead', label: 'Convert', locations: ['record_header'] },
+  { name: 'export_pdf',   label: 'Export',  locations: ['record_more'] },
+]
+```
+
+<!-- doc-snippet: fragment — a metadata excerpt: the bare slots: key is one fragment of a page schema, shown without the document that would contain it -->
+```ts
+// on the page: the header names them
+slots: {
+  header: {
+    type: 'page:header',
+    properties: {
+      title: '{name}',
+      actions: ['convert_lead', 'export_pdf'],
+    },
+  },
+}
+```
+
+An id that names no action on the object renders nothing and says so once in
+the console — a mistyped id is not a silently shorter header.
+
+> Inline `ActionDef` objects in this array still render, so pages written
+> before the ids contract are not stranded. That is renderer tolerance for the
+> migration, not a second declared shape: the contract is
+> `z.array(z.string())`, and only ids are validated.
+
 ### Declaring placement
 
-An authored action renders here only if its **`locations`** declares
+A named action renders here only if its **`locations`** declares
 `record_header` (inline) or `record_more` (straight into the `⋯` menu).
 There is no default: an action that declares no location renders in **no**
 located surface — not here, not the list toolbar, not the row menu. That
@@ -113,13 +151,6 @@ one rule is shared by every surface that places actions by location
 (`action:bar`, `action:group`, `record:quick_actions`, related lists, the
 metadata-admin toolbars and the action engine), so an action behaves the
 same wherever it is drawn.
-
-```ts
-actions: [
-  { name: 'convert_lead', label: 'Convert', locations: ['record_header'] },
-  { name: 'export_pdf',   label: 'Export',  locations: ['record_more'] },
-]
-```
 
 Two placements come from somewhere other than `locations`, and neither
 needs an entry here:
--- a/packages/components/src/__tests__/page-header-action-ids.test.tsx
+++ b/packages/components/src/__tests__/page-header-action-ids.test.tsx
@@ -166,6 +166,9 @@ const objectAuthored = { type: 'page:header', title: 'Lead', actions: AUTHORED.m
 
 describe('page:header — declared action-id lookup (objectui#6252)', () => {
   let warn: ReturnType<typeof vi.spyOn>;
+  /** Every string this render passed to `console.warn`, first argument only. */
+  const warnMessages = (): string[] =>
+    (warn.mock.calls as unknown[][]).map((c) => String(c[0]));
 
   beforeEach(() => {
     warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
@@ -234,8 +237,7 @@ describe('page:header — declared action-id lookup (objectui#6252)', () => {
     expect(buttonNames(container)).toEqual(['Convert Lead']);
 
     await waitFor(() => expect(warn).toHaveBeenCalled());
-    const messages = warn.mock.calls.map((c) => String(c[0]));
-    const hits = messages.filter((m) => m.includes('covert_lead'));
+    const hits = warnMessages().filter((m) => m.includes('covert_lead'));
     expect(hits.length).toBe(1);
     expect(hits[0]).toContain('[page:header]');
     // The message names what the object DOES declare — the fix is usually one
@@ -255,7 +257,7 @@ describe('page:header — declared action-id lookup (objectui#6252)', () => {
     await Promise.resolve();
     await waitFor(() => expect(container.querySelector('h1, [data-page-actions-slot]')).toBeTruthy());
     expect(buttonNames(container)).toEqual([]);
-    expect(warn.mock.calls.map((c) => String(c[0])).filter((m) => m.includes('[page:header] action id'))).toEqual([]);
+    expect(warnMessages().filter((m) => m.includes('[page:header] action id'))).toEqual([]);
   });
 
   it('keeps the inline object shape working during the transition, including mixed arrays', async () => {
--- a/packages/components/src/__tests__/page-header-action-ids.test.tsx
+++ b/packages/components/src/__tests__/page-header-action-ids.test.tsx
@@ -1,0 +1,304 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `page:header` resolves `actions` as ACTION IDS (objectui#6252, implementing
+ * the objectstack#11592 ruling — maintainer, 2026-08-25, 「全部同意」 on
+ * recommendation B).
+ *
+ * `@objectstack/spec`'s `PageHeaderProps.actions` is
+ * `z.array(z.string()).describe('Action IDs to show in header')`. This renderer
+ * used to read that array as `ActionDef` OBJECTS and resolve nothing, so
+ * metadata satisfying the published contract rendered ZERO buttons — the defect
+ * the parent issue reports.
+ *
+ * The shape of the proof is deliberate: the same action metadata is authored
+ * TWICE — once as ids, once as the inline objects authors write today — and the
+ * two renders are compared. That is what makes "renders the same buttons" a
+ * measurement rather than a restatement, and the object-shape render is the
+ * LIVE CONTROL: every equivalence case asserts it is non-empty first, so an
+ * id-side green can never come from two empty headers agreeing.
+ *
+ * Population covers each filter the acceptance criterion names, so one
+ * equivalence assertion measures all of them at once:
+ *   - `actionRendersAt` placement — `list_only` (list_item) must not render;
+ *     `archive` (record_more) must land in the ⋯ menu, not inline
+ *   - `requiredPermissions` — `gated` is denied for this user
+ *   - `visible` — `closed_only` is a CEL predicate that is false for this record
+ *   - `order` — `qualify` (order 1) renders BEFORE `convert` (order 2), which is
+ *     the reverse of the order both authorings list them in
+ */
+
+import * as React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { ComponentRegistry } from '@object-ui/core';
+import { ActionProvider, MetadataCtx, RecordContextProvider } from '@object-ui/react';
+import type { MetadataContextValue } from '@object-ui/react';
+
+/**
+ * A script body on one of the resolved actions. The id path must never carry
+ * this into anything serialized — see the `body.source` case at the bottom.
+ */
+const BODY_MARKER = 'OS6252_HANDLER_BODY_MARKER';
+
+const ACTIONS: Record<string, any> = {
+  convert: { name: 'convert', label: 'Convert Lead', type: 'flow', locations: ['record_header'], order: 2 },
+  qualify: { name: 'qualify', label: 'Qualify', type: 'api', locations: ['record_header'], order: 1 },
+  scripted: {
+    name: 'scripted',
+    label: 'Run Script',
+    type: 'script',
+    locations: ['record_header'],
+    order: 3,
+    body: { language: 'js', source: `return { marker: '${BODY_MARKER}' };` },
+  },
+  archive: { name: 'archive', label: 'Archive', type: 'api', locations: ['record_more'] },
+  list_only: { name: 'list_only', label: 'List Only', type: 'api', locations: ['list_item'] },
+  gated: {
+    name: 'gated',
+    label: 'Gated Action',
+    type: 'api',
+    locations: ['record_header'],
+    requiredPermissions: ['nobody_holds_this'],
+  },
+  closed_only: {
+    name: 'closed_only',
+    label: 'Closed Only',
+    type: 'api',
+    locations: ['record_header'],
+    visible: 'record.status == "closed"',
+  },
+};
+
+/** Authoring order — deliberately NOT the rendered order (see `order` above). */
+const AUTHORED = ['convert', 'qualify', 'scripted', 'archive', 'list_only', 'gated', 'closed_only'];
+
+const OBJECT_META = { name: 'lead', label: 'Lead', actions: AUTHORED.map((n) => ACTIONS[n]) };
+
+const RECORD = { id: 'rec-1', name: 'Ada', status: 'open' };
+const USER = { id: 'u1', systemPermissions: ['setup.access'] };
+
+const getItem = vi.fn(async (type: string, name: string) =>
+  type === 'object' && name === 'lead' ? OBJECT_META : null,
+);
+
+/**
+ * Hand-rolled context value held at MODULE level on purpose: `getItem` is an
+ * effect dependency of `useMetadataItem`, so a value rebuilt per render spins
+ * that hook forever (the loop `NO_METADATA_PROVIDER` was frozen to fix).
+ */
+const METADATA: MetadataContextValue = {
+  apps: [],
+  objects: [OBJECT_META] as any,
+  dashboards: [],
+  reports: [],
+  pages: [],
+  loading: false,
+  error: null,
+  refresh: async () => {},
+  invalidate: () => {},
+  ensureType: async () => [],
+  getItem: getItem as unknown as MetadataContextValue['getItem'],
+  getItemsByType: () => [],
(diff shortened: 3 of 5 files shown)


os-warren:
os-dev-report

Same report as objectstack-ai/objectui#6252 (comment 5492213936); posted here too because the dispatch asked for it on the PR.

```json
{
  "issue": 6252,
  "status": "done",
  "branch": "claude/issue-6252-page-header-action-ids",
  "pr": "https://github.com/objectstack-ai/objectui/pull/7180",
  "premise_still_valid": true,
  "summary": "The canonical page:header renderer (packages/components/src/renderers/layout/containers.tsx) now resolves each element of `actions` as a declared action id against the object's own metadata, through the same useMetadataItem entry record:quick_actions uses — no second resolver. Normalisation sits at the TOP of the actions pipeline, so the existing single filter chain (actionRendersAt placement, requiredPermissions capability gate, visible/hidden, order, inline/overflow split) runs unchanged over uniformly-shaped defs; that is what makes the equivalence claim measurable. Inline ActionDef objects keep working per element, so a half-migrated ['convert', {def}] array resolves the id and passes the object through — renderer tolerance for the migration, still undeclared. An id that resolves to nothing renders nothing and warns ONCE naming the object's declared action names (suppressed while the metadata read is in flight, so a correct page never warns on first paint); silently dropping it would reproduce the silent-loss class this seat keeps filing. Clause 2 fence honoured: the spec type and its zod mirror are untouched, actions stays z.array(z.string()). Docs updated (content/docs/guide/slotted-pages.md taught only the object form) plus a minor changeset. ZONE 2 VERDICTS, all re-measured on this repo's head: A2.1 CONFIRMED (the pre-fix chain read .locations/.requiredPermissions/.visible/.order straight off each array element via actionRendersAt, no lookup anywhere). A2.2 CONFIRMED — exactly ONE canonical page:header registration, ComponentRegistry.register('header', PageHeaderRenderer, {namespace:'page', skipFallback:true}) in containers.tsx; packages/layout/src/index.ts deliberately does NOT re-register it and says so, and the legacy `page-header` alias already resolves ids by delegating to record:quick_actions — so the multi-spelling hazard the PM flagged does not fire here, and the sibling that already implements the ruling is the alias, not a second canonical mount. A2.3 CONFIRMED, cheap. A2.4 CONFIRMED — nothing pinned the rendered result of a string-authored page:header; the only two id-shaped fixtures in the repo pin VALIDATION only (examples/schema-catalog pins that actions:['export'] validates clean, which supports ids), so nothing had to be updated or deleted. A2.5 CONFIRMED with no seam: @object-ui/components already depends on @object-ui/react and already imports eight hooks from it, so useMetadataItem was reachable with no new dependency and no second context scope.",
  "tests": "All runs at final commit af40cf5c2 with a clean working tree (git status --short empty), so every figure below read exactly that tree. Heavy runs went through the container's shared verify lock; verdicts read from its VERDICT line, exit codes captured before any pipe. (1) NEW PIN packages/components/src/__tests__/page-header-action-ids.test.tsx, registered in vitest.config.mts heavyDomTests (required — it renders through the ComponentRegistry): 'pnpm exec vitest run --project dom-heavy ...' gives Test Files 1 passed (1), Tests 8 passed (8). Shape of the proof: the SAME action metadata authored twice, once as ids and once as inline objects, both renders compared on the ordered button-name list AND a normalised innerHTML projection; the object-shape render is the LIVE CONTROL and is asserted non-empty first, so an id-side green cannot come from two empty headers agreeing. The population makes one equivalence assertion cover every filter the acceptance criterion names: list_item-only must not render, record_more must route to the overflow menu and never inline, a requiredPermissions the user lacks is denied, a CEL visible false for the record hides, and qualify (order 1) renders BEFORE convert (order 2) — the reverse of the order both authorings list them in. (2) ABLATION, direction predicted before running (red on the id path, green on the object-shape control). Mutation = feeding rawHeaderActions back into the filter chain. Proven on disk by BOTH marker count and blob hash, never by an editor's exit code: PRE target-count=1 inject-count=0 PRE_HASH=3dafeb94 equal to the HEAD blob; POST target-count=0 inject-count=1 POST_HASH=f87a37c9. Result: Tests 7 failed | 1 passed (8) — passing count asserted alongside the red, and 7/1 matched the prediction exactly (the one immovable case is 'does not warn while the metadata lookup is still in flight', true with or without resolution, so this is a targeted break and not a collapsed population). Restore proven by STATE not exit code: restored blob 3dafeb94 equals the HEAD blob, and git diff HEAD / git diff --cached / git status --short are all empty; the script carried an EXIT/INT/TERM trap with an absolute path. No rebuild needed on either leg and this was measured, not assumed: the root vitest.config.mts aliases @object-ui/components to packages/components/src, so the suite resolves SOURCE, never dist. (3) ACCEPTANCE CRITERION 3, measured against the BUILT ARTIFACT as the card demands. pnpm --filter @object-ui/components build, then a run mounting the built packages/components/dist/index.js by explicit relative import (the workspace alias would otherwise redirect to src) with an id-authored header whose resolved action carries body:{language:'js',source:'...OS6252_DIST_BODY_MARKER...'}: the id resolved through the BUILT renderer, the marker appears nowhere in the rendered DOM, and the authored node's JSON.stringify is byte-identical before and after render. LIVE CONTROL: with that one import removed the same file fails 'expected undefined to be truthy' (page:header is unregistered in the light dom project), which proves the passing run measured the built bundle and nothing else. That measurement is deliberately NOT COMMITTED — turbo's test task is dependsOn ['^build'], dependencies only, so a committed dist-importing test would be NOT MEASURED in CI rather than a pin; filed as objectui#7183. The committed half is the pin 'never writes a resolved def — and so no body.source — back onto the authored node', with the object-shape authoring as its live control: expect(JSON.stringify(objectAuthored)).toContain(BODY_MARKER) — the inline form DOES drag the handler body into what it serializes, which is what proves the id-side assertion can fail at all. Built-bundle greps with controls: 'did not resolve'=1 and 'page:header'=7 present, 'OS6252_HANDLER_BODY_MARKER'=0. (4) REGRESSION SURFACE: 'vitest run packages/components scripts/__tests__/vitest-invocation-guard.test.ts' gives 224 files, 2070 passed; all page-header* plus RecordDetailView.header*/predicate suites give 16 files, 184 passed; console registry/contract parity (registry-inputs-spec-parity, public-contract, ga-honoured-inputs-author-reach, component-input-union-specimens, record-block-record-reach) gives 156 passed. (5) TYPES: 'pnpm --filter @object-ui/components run type-check' exit 0 — note the script is spelled type-check (hyphen) in this repo — and it demonstrably COVERS the new test file: tsc -p tsconfig.test.json reported four TS7006 errors inside it before they were fixed, so this is not a typecheck that excludes tests. (6) GATES: check:control-bytes OK (5945 tracked text files); check:doc-types OK; check:doc-fences OK; check:doc-snippets OK, 272/272 blocks judged 0 failed, after building the closure it names (its first run was PRECONDITION NOT MET exit 2 — recorded as NOT MEASURED, then re-run, never as a pass); check:sdui-registration-pins OK, all 16 registrations present, after building the console plus its dependency closure (its first two runs were exit 2 'no console build to weigh' and a TS2307 unbuilt-closure error — both NOT MEASURED, not results). (7) LINT: eslint over all three changed source files, 0 errors (5 no-explicit-any WARNINGS in the new test matching its sibling page-header tests; lint.yml deliberately sets no --max-warnings). The narrowing is measured, not assumed, all three pieces: population is the diff itself via git diff --name-only; the count comes from --format json (3 files linted); and the config is not type-aware — eslint.config.js has zero projectService / parserOptions / project: occurrences against a live control of 10 'rules' hits in the same file — so with no config file in the diff no untouched file's verdict can move. The repo-wide farm is CI's run either way. CI convergence not awaited, per the standing dispatch contract.",
  "mcp_calls": "12 — issue_read get + get_comments, create_pull_request, pull_request_read (body read-back), search_issues x2 (the second was the required control and it FAILED to hit, see open_questions), issue_write create x2, and this report's comments plus one read-back. Bulk reads went to git and to the zero-quota web payload channel instead.",
  "open_questions": [
    {
      "question": "MECHANISM, not a code decision — the MCP `search_issues` channel appears blind in this container, and this seat's dedupe had to route around it. A targeted search for this card's own near-verbatim title returned total_count 0, i.e. the known-must-hit CONTROL failed, so its earlier empty result for my dedupe query was not a reading. Unauthenticated REST is also 403 here (both /repos/.../issues and /search/issues). Dedupe was completed on the zero-quota GitHub web payload channel with a live control that DID hit (query 'action-id lookup' returned 8 issue numbers including 6252; the two dedupe queries returned disjoint populations with no duplicate of either finding), so both findings are filed properly — but the next seat should know that search_issues answering 0 here means nothing.",
      "options": [
        "A — treat MCP search_issues as unavailable for this seat/container and make the web payload channel the documented dedupe route, control word mandatory",
        "B — investigate why search_issues returns 0 for a query matching an issue it can read directly via issue_read (indexing lag vs. a scope/permission difference between the read and search paths)",
        "C — do nothing; each dev re-discovers it per run"
      ],
      "recommendation": "A plus B. A is cheap and immediate — the web channel is measured working here, costs no GraphQL quota, and the control discipline the dispatch already mandates is exactly what caught this. B matters because a silently-empty search is the worst failure shape available to a dedupe step: it looks like 'no duplicate exists' and it is unfalsifiable without a control, so a seat that skips the control files duplicates and never learns."
    }
  ],
  "out_of_scope_findings": [
    "filed as objectui#7182: the action-id lookup now exists twice (record-quick-actions.tsx and containers.tsx) and the two copies DISAGREE on mixed arrays — quick_actions switches on the whole array (every element a string), page:header normalises per element — so during the migration this card opens, one authored array means two things depending on which renderer draws it. Not fixed here: unifying them changes record:quick_actions behaviour on a renderer this card does not own, and wants a ruling on whether a mixed array is a supported migration state.",
    "filed as objectui#7183: no committed test can observe its own package's BUILT artifact — the root vitest alias sends every workspace specifier to src, and turbo's test task is dependsOn ['^build'] (dependencies only), so a dist-importing test is MODULE_NOT_FOUND / NOT MEASURED in CI. This card's acceptance criterion 3 explicitly asks for a built-artifact measurement; it was run by hand with a live control and passes, but it cannot be left behind as a standing pin."
  ]
}
```

---

os-warren:

**Verdict: ACCEPT.** Arming on green. Two judgement calls are ruled below, and one finding in this report is bigger than this card.

### ⭐ The finding that outranks the PR: `search_issues` is blind in this container, and its zero is not a reading

> *"A targeted search for this card's own near-verbatim title returned `total_count 0`, i.e. the known-must-hit CONTROL failed, so its earlier empty result for my dedupe query was not a reading."*

This is the most valuable thing in the report and it is not about `page:header` at all. **A silently-empty search is the worst failure shape available to a dedupe step**: it renders as *"no duplicate exists"*, it is indistinguishable from a real negative, and it is unfalsifiable without a control. A seat that skips the control files duplicates and never finds out.

It is also the seat's own instrument rule — *a control that returns zero is no control; an empty output is not a `0`* — applied one level up, to a **channel** rather than a grep. That generalisation is worth more than the rule it came from, and I am adopting it: **the control belongs on the channel, not only on the query.** Recording it as its own card rather than leaving it in a PR body, because it applies to every dev in this container and nothing in the repo currently says so. Dedupe here was completed on the zero-quota web payload channel with a control that **did** hit (`action-id lookup` → 8 issue numbers including #6252), so both out-of-scope findings are properly filed.

⚠️ Noting for the record that this does **not** impugn `search_code`, which this seat used successfully an hour ago (13 hits on `13855`, with the returned paths independently confirmed against a local checkout). The blindness is specific to the issue-search path, which is exactly why a per-channel control is the answer rather than a blanket distrust.

### A2.2 was my low-confidence assumption, and confirming it was worth the cost

I flagged a possible second `page:header` mount because #7021 found the record-title key in three spellings. Measured: **exactly one** canonical registration (`ComponentRegistry.register('header', PageHeaderRenderer, {namespace:'page', skipFallback:true})`), `packages/layout/src/index.ts` deliberately does **not** re-register it *and says so*, and the legacy `page-header` alias already resolves ids by delegating to `record:quick_actions`.

⭐ That last clause reframes the card: **the sibling that already implements this ruling is the alias, not a second canonical mount.** The acceptance criterion was not under-specified, and the hazard did not fire. A confirmed assumption that closes a named risk is a result, not a formality.

### Ruling 1 — the built-artifact measurement was run and deliberately NOT committed: **upheld**

Acceptance criterion 3 (*"the id path carries no `body.source` into the built artifact"*) was measured properly — the built `packages/components/dist/index.js` mounted by explicit relative import, an id-authored header whose resolved action carries an `OS6252_DIST_BODY_MARKER` body, marker absent from the rendered DOM, authored node byte-identical before and after render, and a **live control** proving the passing run measured the bundle (removing that one import fails `expected undefined to be truthy`, because `page:header` is unregistered in the light dom project).

And then it was **not committed**, because turbo's `test` task is `dependsOn ['^build']` — *dependencies only* — so a committed dist-importing test would be `MODULE_NOT_FOUND` / NOT MEASURED in CI.

⭐ **That is the right call and I want the reasoning preserved: a test that reports NOT MEASURED in CI is worse than no test, because it renders as coverage.** Shipping it would have put a permanently-void assertion into the suite under the name of this card's hardest acceptance criterion. Filed as **#7183** instead, with the committed half being the pin that *can* fail — *"never writes a resolved def back onto the authored node"* — carrying the object-shape authoring as its live control, since the inline form genuinely **does** drag the handler body into what it serializes. That control is what makes the id-side assertion falsifiable at all.

### Ruling 2 — #7182, the divergence this card *created*: **file, do not unify. Grading raised.**

The action-id lookup now exists in two places and they **disagree on mixed arrays**: `record:quick_actions` switches on the whole array (every element a string), `page:header` normalises per element. So during the migration this card opens, one authored array means two things depending on which renderer draws it.

I asked for "no second resolver" and that fence held — the metadata read goes through the same `useMetadataItem` entry. What diverges is the **array normalisation policy**, which is a different thing and was not covered. ⚠️ But it is a real new dialect in a repo actively paying down dialects (#6771 is retiring one this hour), and unlike most findings it was **introduced here rather than discovered**. Declining to unify was still correct: it would change `record:quick_actions` behaviour on a renderer this card does not own, and whether a mixed array is a supported migration state is a ruling, not an implementation detail.

⇒ I am treating #7182 as the **completion condition of this migration**, not a loose end, and will grade it accordingly rather than letting it sit as an ordinary finding.

### Also correct

- **Clause ② fence honoured** — spec type and zod mirror untouched, `actions` stays `z.array(z.string())`. The renderer catches up to a declaration that was already public; nothing widens.
- **Unresolvable id renders nothing and warns ONCE**, naming the object's declared action names, **suppressed while the metadata read is in flight** so a correct page never warns on first paint. That is the loud-over-silent option, and the in-flight suppression is the detail that makes it usable rather than noisy.
- **Equivalence proved by the same metadata authored twice** — ids and inline objects — compared on the ordered button-name list *and* a normalised innerHTML projection, with the object-shape render asserted **non-empty first**, so an id-side green cannot come from two empty headers agreeing. One assertion covers every filter the acceptance criterion names, including `qualify` (order 1) rendering **before** `convert` (order 2) — the reverse of the order both authorings list them in.
- **Ablation**: `Tests 7 failed | 1 passed (8)`, matching the 7/1 prediction exactly, with the single immovable case identified (*"does not warn while the metadata lookup is still in flight"* is true with or without resolution) — which is what distinguishes a targeted break from a collapsed population.
- Two gates correctly reported **NOT MEASURED** on first run (`check:doc-snippets` exit 2 PRECONDITION NOT MET; `check:sdui-registration-pins` exit 2 and a TS2307 unbuilt closure) and re-run to real verdicts rather than scored as passes.
- Docs updated at `content/docs/guide/slotted-pages.md`, which taught only the object form. ⛔ `content/docs/releases/` untouched, correctly.

---

Nothing to change.

---

PR #7186, opened 2026-09-01
fix(plugin-grid): retire the three inert `fieldMeta` copies the cell never reads

### LINKED ISSUE #7166: finding(plugin-grid): 6 of the 10 keys ObjectGrid copies onto a column's `fieldMeta` have no reader on that bag — the picker keys reach the editor by a different seam
Provenance: measured while implementing objectui#7154, by the same rendering measurement that card asked for. Filed unassigned. Not folded into objectui#7154 — that card is about four keys NOT on the copy set; this is about keys that ARE on it.

## What is measured

On `51449a043`, `RELATIONAL_META_KEYS` derives to 10 keys:

`reference`, `reference_to`, `displayField`, `display_field`, `descriptionField`, `description_field`, `lookupColumns`, `lookupFilters`, `lookup_filters`, `id_field`

`applyRelationalMeta` writes them onto `fieldMeta`, and `fieldMeta` in `ObjectGrid.tsx` is handed to `CellRenderer` as the `field` prop and to nothing else — grepped, all six occurrences are that one JSX prop across the three column-building paths. `getCellRenderer` sends a relational column to `LookupCellRenderer` (`packages/fields/src/index.tsx`) and a `user` column to `UserCellRenderer`.

- `LookupCellRenderer` reads exactly `reference_to`, `reference`, `display_field`, `displayField`, `reference_field` off that prop (plus `options`).
- `UserCellRenderer` destructures `{ value }` only — it reads no field meta at all.

Intersecting: **6 of the 10 copied keys have no reader on the bag they are copied onto** — `descriptionField`, `description_field`, `lookupColumns`, `lookupFilters`, `lookup_filters`, `id_field`.

That is the shape objectui#6711 (`reference_to_field`) and objectui#6874 (`titleFormat`) each retired: a member written from the schema def on every column build that nothing on the receiving path reads.

## Why the docblocks say otherwise

Both `relationalMetaKeys.ts` and `relationalMetaCopySet.derivation.test.ts` stated that the grid's inline editor "dispatches the same bag" into `LookupField` and `UserField`. It does not. `ObjectGrid.renderCellEditor` looks the field up in the object schema and spreads the WHOLE def into the widget:

```
const fieldDef = (objectSchema)?.fields?.[ctx.column?.accessorKey];
let field: any = { name: ctx.column.accessorKey, ...fieldDef };
```

So every key a def carries reaches `LookupField` whether or not it is copied. objectui#7154's PR corrected those two docblocks and pinned the consequence by rendering: four spec-declared picker keys that are NOT on the copy set (`multiple`, `allowCreate`, `lookupPageSize`, `dependsOn`) all take effect in the inline picker with the table unchanged, and an ablation that strips them from the editor's spread turns 4 of those 5 assertions red while the copy-set assertion stays green.

## Why this matters beyond tidiness

The keys are not equally inert.

- `id_field`, `lookup_filters`, `description_field` were copied for the inline picker, on the docblock's stated reasoning ("the extra keys drive the inline picker's query"). The picker gets them from the def instead, so the copies are dead writes — the objectui#6711 class, and retirable on the same reasoning.
- `descriptionField` and `lookupColumns` were ADDED by objectui#6875 as its fix, on the reading that they were "read, never copied" spellings the picker needed. Under this measurement that half of objectui#6875 delivered nothing: their only reader is `LookupField`, which was already receiving them off the def. objectui#6875's third key, `displayField`, is genuinely delivered — `LookupCellRenderer` reads it, and `lookupDisplayFieldSpelling-6875.test.tsx` renders the difference. So objectui#6875 is 1-for-3, not wrong.

## What is NOT measured

- Whether the DERIVATION should be re-scoped. The gate sweeps three consumers, but only one of them is fed this bag; a read set that mixes the cell's bag with the editor's def is why a "read, never copied" verdict can be reached for a key that needed no copying. Rebuilding it around the cell alone (with the editor widgets classified separately) is a design change to objectui#6875's mechanism, not a call to make from here.
- Whether any host `DataSource` outside these two repos feeds `fieldMeta` to something else. The sweep bounds `packages/` and `apps/` in this repo, as objectui#6875's own docblock notes for its verdicts.
- Whether retiring the four legacy spellings is safe on back-compat grounds. `id_field`, `lookup_filters`, `description_field` are recorded as `legacy-alias` precisely because a host outside these repos may hand-feed them; that argument is about producers, and is untouched by this reader-side measurement.

os-warren:
Fixes #7166

`ObjectGrid` no longer copies `descriptionField`, `lookupColumns` or `lookupFilters` onto a relational column's `fieldMeta`. **No behaviour change** — all three still reach the inline lookup picker, by a different seam.

Copy set: **10 keys to 7**. Re-measured on `47035ce79` (the card's own numbers were taken on `51449a043`, before PR #7167 landed into these files).

## Ruling item 3 — the gating measurement: `RELATIONAL_META_KEYS` is INTERNAL

Clause 2 is **not** engaged, measured three ways, each with a live control:

| probe | relationalMetaKeys | control (`useRowColor`) |
|---|---|---|
| re-export in `src/index.tsx` barrel | absent (no `export *` either) | line 30 |
| built `dist/index.d.ts` | 0 occurrences of any of its symbols | present |
| built `dist/index.js` export list | not exported | present |

Every import of `./relationalMetaKeys` in the repo lives inside `packages/plugin-grid/src/` — the implementation plus two of its own tests. The cross-package hits are prose in docblocks; `plugin-dashboard` keeps its own independent `CELL_RELATIONAL_META_KEYS` literal rather than importing this one. Narrowing the copy set therefore moves no published surface.

## What was measured

`applyRelationalMeta` writes onto the `fieldMeta` that `generateColumns` hands to `CellRenderer` as the `field` prop — 6 JSX passes across the three column-building paths, and nowhere else (as a whole object it otherwise only reaches `applyRelationalMeta`, which mutates it).

- `LookupCellRenderer` reads exactly `reference_to`, `reference`, `display_field`, `displayField`, `reference_field`, plus `options`.
- `UserCellRenderer` has **zero** occurrences of `field` — control: `LookupCellRenderer` has 8.

Stated by receiver rather than by count: `packages/fields/src/index.tsx`, the file holding **every** cell renderer, contains **zero** occurrences of the three retired keys, against a control of **22** occurrences of the `display_field` / `displayField` / `reference_to` spellings the cell does read.

Their only readers off a field meta are `LookupField` and `UserField` — the two **editor** widgets — and the editor is not fed this bag. `renderCellEditor` looks the field up in the object schema and spreads the whole def into the widget (`{ name: ctx.column.accessorKey, ...fieldDef }`), so every key a def carries reaches `LookupField` whether or not it is copied. The copies were dead writes: the #6711 / #6874 class arriving from the opposite direction — those keys had no *declaration*, these have no *reader on this path*.

## The evidence bar this card exists to raise

Two of the three, `descriptionField` and `lookupColumns`, were **added by #6875** as its fix, classified from a read-set derivation alone. Its third key `displayField` is genuinely delivered and stays copied — and it is the one that arrived with a rendering test.

⭐ **A derivation establishes that a consumer READS a key. It does not establish that a given BAG is how the consumer gets it.** Recorded in both docblocks.

So this is pinned by **rendering**, in both directions — new `__tests__/relationalMetaCopySet-7166.test.tsx`:

- **The cell.** Four lookup columns over one referenced record, differing only in the key under test, all resolve the same text; the `displayField` control column resolves something else. The control is what makes the three zeros readings rather than a fixture that never reached the lookup path.
- **The editor.** Each retired key still takes effect in the inline picker with the copy set no longer carrying it — `descriptionField` drives the secondary line, `lookupColumns` shapes the picker columns, `lookupFilters` scopes the candidates — each against a sibling control column that declares nothing. The referenced schema is deliberately two fields, so `deriveLookupColumns` derives nothing and the controls stay load-bearing.

## Scope: 3 keys, not 6

The card measures six keys with no reader on this bag. They are two populations. The three snake_case `legacy-alias` spellings — `description_field`, `lookup_filters`, `id_field` — are **deliberately kept**. They are recorded as legacy aliases precisely because a host `DataSource` outside these repos may hand-feed them; that is a **producer-side** argument and this measurement is entirely **reader-side**. Their reader-side verdict is now recorded on the table with the open producer question stated, so the next pass inherits a measurement instead of a silence.

## ⛔ Reported, not fixed: the derivation gate cannot see this class

The gate's read set is a **union over three consumers, two of which are not fed this bag**. Membership means "some consumer reads this key", never "this bag is how that consumer gets it" — and only the second claim justifies a copy. All three retired keys **remain** in the extracted read set, so every derived assertion passes whichever verdict they carry.

Measured, not asserted (ablation 2 below): re-adding `lookupFilters` to the copy set leaves **every derived assertion in the gate green** and turns red only the hand-written pins. Re-scoping the derivation around the cell alone, with the editor widgets classified separately, is a design change to #6875's mechanism and is **filed rather than made here**.

The gate is **strengthened, not weakened**: every `deferred` verdict is now mechanically proved spec-declared against the installed `FieldSchema`, and the sibling pins turn the retired `lookupFilters` into a live negative assertion on a fixture that still declares it.

## Tests

All runs are the canonical repo-root invocation (the repo's vitest guard refuses a package-cwd run as a false green). Union run on the final commit `f4f299e23`:

```
pnpm exec vitest run packages/plugin-grid/src packages/fields/src/__tests__/lookupCellDisplayField.test.tsx
  Test Files  106 passed (106)
       Tests  956 passed (956)
```

`pnpm --filter @object-ui/plugin-grid run type-check` — clean (first attempt reported `TS2307: Cannot find module '@object-ui/*'`, which is an unbuilt dependency closure and NOT a result; re-run after `pnpm --filter '@object-ui/plugin-grid^...' build`). The new test file is confirmed present in the `tsconfig.test.json` program via `--listFiles`, so that green does cover it.

`eslint . --no-inline-config` was run repo-wide, not narrowed: **4100 files**, 75 files with errors, **all pre-existing and none in this diff**; the changed files are E0.

### Ablations

Both targets resolve from **source** (the root config aliases every `@object-ui/*` to `src/`, and both are relative imports), so no rebuild leg applies. Each mutation was proven on disk by **marker count and blob hash** — never an editor's exit code — and each restore by **state**: `git diff HEAD`, `git diff --cached` and `git status --short` all empty, blob back to the HEAD blob, marker gone.

1. **Break the editor's schema spread** (predicted before running: the editor half goes red, the cell half stays green) — `Tests 3 failed | 2 passed`. Exactly B1/B2/B3 red; test A and the premise test green. The retirement's safety claim is load-bearing.
2. **Re-add `lookupFilters` to the copy set** (predicted: derived gate assertions stay green, hand-written pins go red) — `Tests 8 failed | 19 passed`. The only derivation-gate failure is the hand-written #7166 pin; every derived assertion stayed green. The cell and editor tests also stayed green, which is itself the finding: the copy makes no observable difference.

---

### DIFF
--- a/.changeset/7166-retire-inert-fieldmeta-copies.md
+++ b/.changeset/7166-retire-inert-fieldmeta-copies.md
@@ -1,0 +1,56 @@
+---
+'@object-ui/plugin-grid': patch
+---
+
+`ObjectGrid` no longer copies `descriptionField`, `lookupColumns` or `lookupFilters` onto a
+relational column's `fieldMeta` (objectui#7166). No behaviour change — all three still reach
+the inline lookup picker, by a different seam.
+
+`applyRelationalMeta` writes the copy set onto the `fieldMeta` that `generateColumns` hands
+to `<CellRenderer>` as the `field` prop — six JSX passes across the three column-building
+paths, and nowhere else. For a relational column that resolves to `LookupCellRenderer`,
+which reads exactly `reference_to`, `reference`, `display_field`, `displayField`,
+`reference_field` and `options`; a `user` column resolves to `UserCellRenderer`, which
+destructures `{ value }` and reads no field meta at all. Measured by receiver rather than by
+count: `packages/fields/src/index.tsx`, the file holding **every** cell renderer, contains
+**zero** occurrences of the three retired keys, against a control of 22 occurrences of the
+`display_field` / `displayField` / `reference_to` spellings the cell does read.
+
+Their only readers off a field meta are `LookupField` and `UserField` — the two **editor**
+widgets — and the grid's inline editor does not receive this bag. `renderCellEditor` looks
+the field up in the object schema and spreads the whole def into the widget
+(`{ name: ctx.column.accessorKey, ...fieldDef }`), so every key a def carries reaches
+`LookupField` whether or not it is copied. The copies were dead writes: the objectui#6711
+(`reference_to_field`) and objectui#6874 (`titleFormat`) class, arriving from the opposite
+direction — those keys had no *declaration*, these have no *reader on this path*.
+
+Two of the three, `descriptionField` and `lookupColumns`, were **added** by objectui#6875 as
+its fix, classified from a read-set derivation alone. Its third key, `displayField`, is
+genuinely delivered and stays copied — and it is the one that arrived with a rendering test.
+⭐ The generalisable lesson, now recorded in both docblocks: **a derivation establishes that
+a consumer READS a key; it does not establish that a given BAG is how the consumer gets it.**
+
+- New `__tests__/relationalMetaCopySet-7166.test.tsx` renders both directions. The **cell**:
+  four lookup columns over one referenced record, differing only in the key under test, all
+  resolve the same text, while the `displayField` control column resolves something else —
+  the control is what makes the three zeros readings rather than a fixture that never reached
+  the lookup path. The **editor**: each retired key still takes effect in the inline picker
+  with the copy set no longer carrying it — `descriptionField` drives the secondary line,
+  `lookupColumns` shapes the picker columns, `lookupFilters` scopes the candidates — each
+  against a sibling control column that declares nothing.
+- The three snake_case `legacy-alias` spellings — `description_field`, `lookup_filters`,
+  `id_field` — have the same reader-side verdict and are **deliberately kept**. They are
+  recorded as legacy aliases precisely because a host `DataSource` outside these repos may
+  hand-feed them; that is a producer-side argument, untouched by this reader-side
+  measurement. Their verdict is now recorded on the table with the open producer question
+  stated, so the next pass inherits a measurement instead of a silence.
+- ⛔ The derivation gate cannot enforce this retirement, and now says so. Its read set is a
+  union over three consumers, two of which are not fed this bag, so all three retired keys
+  remain in it and every derived assertion passes whichever verdict they carry. Their absence
+  is pinned by an explicit hand-written assertion plus the rendering test. Re-scoping the
+  derivation around the cell alone is a design change to objectui#6875's mechanism and is
+  filed, not made here.
+- The gate is strengthened, not weakened: every `deferred` verdict is now mechanically proved
+  spec-declared against the installed `FieldSchema`, and the sibling pins
+  (`relationalMetaCopySet-6711` / `-6874`) turn the retired `lookupFilters` into a live
+  negative assertion on a fixture that still declares it.
--- a/packages/plugin-grid/src/__tests__/relationalMetaCopySet-6711.test.tsx
+++ b/packages/plugin-grid/src/__tests__/relationalMetaCopySet-6711.test.tsx
@@ -68,11 +68,11 @@ const MANAGER_DEF = {
   reference_to_field: 'MUST_NOT_BE_COPIED',
 };
 
-/** The seven keys that survive both retirements — the control. */
+/** The six keys that survive every retirement so far — the control. */
 const SURVIVING_KEYS = [
   'reference_to', 'reference',
   'display_field', 'id_field', 'description_field',
-  'lookup_filters', 'lookupFilters',
+  'lookup_filters',
 ] as const;
 
 const ROWS = [{ id: 'r1', name: 'Tower T1', manager: 'u1' }];
@@ -161,13 +161,18 @@ describe('objectui#6711 — ObjectGrid no longer copies `reference_to_field` ont
       expect(meta).not.toHaveProperty('reference_to_field');
     });
 
-    it(`still copies the seven surviving relational keys (${name})`, async () => {
+    it(`still copies the six surviving relational keys (${name})`, async () => {
       const meta = await renderAndCaptureMeta(schemaExtra);
       for (const key of SURVIVING_KEYS) {
         expect(meta).toHaveProperty(key);
       }
       expect(meta.reference_to).toBe('users');
       expect(meta.display_field).toBe('name');
+      // objectui#7166 retired `lookupFilters` from the copy set — its only
+      // reader is an editor widget, which `renderCellEditor` feeds from the
+      // schema def. The fixture above still declares it, so this absence is a
+      // reading and not a fixture that never offered the key.
+      expect(meta).not.toHaveProperty('lookupFilters');
     });
   }
 });
--- a/packages/plugin-grid/src/__tests__/relationalMetaCopySet-6874.test.tsx
+++ b/packages/plugin-grid/src/__tests__/relationalMetaCopySet-6874.test.tsx
@@ -79,11 +79,11 @@ const MANAGER_DEF = {
   reference_to_field: 'x',
 };
 
-/** The seven keys that survive both retirements — the control. */
+/** The six keys that survive every retirement so far — the control. */
 const SURVIVING_KEYS = [
   'reference_to', 'reference',
   'display_field', 'id_field', 'description_field',
-  'lookup_filters', 'lookupFilters',
+  'lookup_filters',
 ] as const;
 
 const ROWS = [{ id: 'r1', name: 'Tower T1', manager: 'u1' }];
@@ -172,13 +172,18 @@ describe('objectui#6874 — ObjectGrid no longer copies `titleFormat` onto field
       expect(meta).not.toHaveProperty('titleFormat');
     });
 
-    it(`still copies the seven surviving relational keys (${name})`, async () => {
+    it(`still copies the six surviving relational keys (${name})`, async () => {
       const meta = await renderAndCaptureMeta(schemaExtra);
       for (const key of SURVIVING_KEYS) {
         expect(meta).toHaveProperty(key);
       }
       expect(meta.reference_to).toBe('users');
       expect(meta.display_field).toBe('name');
+      // objectui#7166 retired `lookupFilters` from the copy set — its only
+      // reader is an editor widget, which `renderCellEditor` feeds from the
+      // schema def. The fixture above still declares it, so this absence is a
+      // reading and not a fixture that never offered the key.
+      expect(meta).not.toHaveProperty('lookupFilters');
     });
   }
 });
--- a/packages/plugin-grid/src/__tests__/relationalMetaCopySet-7166.test.tsx
+++ b/packages/plugin-grid/src/__tests__/relationalMetaCopySet-7166.test.tsx
@@ -1,0 +1,322 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7166 — `descriptionField`, `lookupColumns` and `lookupFilters` leave
+ * `ObjectGrid`'s relational copy set, and NOTHING a user can see changes.
+ *
+ * ## Why this file renders instead of asserting the copy set
+ *
+ * This card exists because a DERIVATION was mistaken for a delivery proof.
+ * objectui#6875 swept its three consumers, found `descriptionField` and
+ * `lookupColumns` read and not copied, and added them — onto a bag their only
+ * reader never consults. Its third key, `displayField`, was genuinely
+ * delivered, and it is the one that arrived with a rendering test
+ * (`lookupDisplayFieldSpelling-6875.test.tsx`, `ACME-42` vs `Wrong Name`).
+ * One-for-three, and the difference between the hit and the misses is exactly
+ * the evidence bar.
+ *
+ * ⭐ A derivation establishes that a consumer READS a key. It does NOT
+ * establish that a given BAG is how that consumer gets it. So this file proves
+ * the second claim the only way it can be proved — by rendering — in both
+ * directions:
+ *
+ *   A. THE CELL. Four lookup columns over one referenced record, differing only
+ *      in the key under test, all resolve the SAME text. The `displayField`
+ *      column is the control: it differs, which is what makes the other three
+ *      zeros readings rather than a fixture that never reached the lookup path.
+ *
+ *   B. THE EDITOR. Each retired key STILL takes effect in the inline picker
+ *      with the copy set no longer carrying it, because `renderCellEditor`
+ *      looks the field up in the object schema and spreads the whole def into
+ *      the widget. This is the half that makes the retirement
+ *      behaviour-preserving rather than merely tidy, and each test carries a
+ *      sibling control column that declares nothing.
+ *
+ * ## The fixture choice that makes the controls load-bearing
+ *
+ * The referenced object's schema declares ONLY `id` and `name`. That matters:
+ * `deriveLookupColumns` builds a picker column set from the referenced schema
+ * whenever a field declares no `lookup_columns`, and `effectiveDescriptionField`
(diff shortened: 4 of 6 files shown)


os-warren:
os-dev-report

```json
{
  "issue": 7166,
  "status": "done",
  "branch": "claude/issue-7166-inert-fieldmeta-copies",
  "pr": "https://github.com/objectstack-ai/objectui/pull/7186",
  "premise_still_valid": true,
  "summary": "RULING ITEM 3 FIRST, and it came back INTERNAL, so I proceeded: `RELATIONAL_META_KEYS` is not re-exported from the plugin-grid barrel (fully enumerated export list, no `export *`), the built `dist/index.d.ts` holds zero occurrences of any of the module's symbols, and it is absent from the built bundle's export list — each with `useRowColor` as a live control that hits in all three probes. Clause 2 not engaged. A2.1 re-derived on 47035ce79 and it still holds: the copy set was 10, six had no reader on the bag. Retired the 3 modern spellings (`descriptionField`, `lookupColumns`, `lookupFilters`) through the derivation table by flipping their verdict to `deferred`, the verdict this repo already uses for the same class; copy set now derives to 7, verified at runtime. Kept the 3 snake_case legacy aliases with their reader-side verdict recorded on the table and the unanswered producer question stated beside it. A2.3, A2.4 and A2.5 all confirmed on head by re-measurement, not inherited: 6 `field={fieldMeta}` JSX passes across the three column-building paths and no other whole-object use except `applyRelationalMeta`; `UserCellRenderer` has zero occurrences of `field` against a control of 8 in `LookupCellRenderer`; and `packages/fields/src/index.tsx`, which holds every cell renderer, has zero occurrences of all six keys against a control of 22 for the spellings the cell does read. The gate was strengthened, not weakened. Note for the PM: the platform appended a second copy of the signature footer to the PR body; I left it, because a PATCH edit would downgrade both footers to the bare form and lose the session URL.",
  "tests": "All runs used the canonical repo-root vitest invocation (this repo's guard refuses a package-cwd run as a false green) and went through the shared verify lock. UNION on the final commit f4f299e23: `pnpm exec vitest run packages/plugin-grid/src packages/fields/src/__tests__/lookupCellDisplayField.test.tsx` gave `Test Files 106 passed (106)` / `Tests 956 passed (956)`, VERDICT command-exit 0. Baseline before any edit was 5 files / 27 tests; the relational suites after the change are 6 files / 34 tests. `pnpm --filter @object-ui/plugin-grid run type-check` clean — the FIRST attempt returned TS2307 'Cannot find module @object-ui/*' across files I never touched, which is an unbuilt dependency closure and NOT MEASURED rather than a red; re-ran after `pnpm --filter '@object-ui/plugin-grid^...' build` (13 packages) and it passed with the script name echoed, so it was not a zero-match no-op. Confirmed the new test file is actually IN the tsconfig.test.json program with `--listFiles` (1 hit, control `lookupPickerKeys-7154` also 1, 1770 files). LINT: ran `eslint . --no-inline-config` REPO-WIDE rather than narrowing — 4100 files by eslint's own count, 75 files with errors, every one pre-existing and none in this diff; my files are E0 (one unused import it caught was fixed). Control-character self-scan clean over all 6 changed files, with the pattern itself verified against 0x01 and 0x0b after my first control (a tab) proved nothing because 0x09 is deliberately outside the class. ABLATIONS — both targets resolve from SOURCE (the root config aliases every @object-ui/* to src/, and both are relative imports), so no rebuild leg applies and none was needed; each mutation was proven on disk by marker count AND blob hash versus the HEAD blob, never by an editor's exit code, and each restore proven by STATE with `git diff HEAD`, `git diff --cached` and `git status --short` all empty plus the blob back to the HEAD blob and the marker gone. Both ran against the COMMITTED implementation. (1) Break `renderCellEditor`'s schema spread — direction predicted before running as 'the editor half goes red, the cell half stays green' — result `Tests 3 failed | 2 passed`: exactly B1/B2/B3 red, test A and the premise test green. (2) Re-add `lookupFilters` to the copy set — direction predicted as 'the derived gate assertions stay green, only the hand-written pins go red' — result `Tests 8 failed | 19 passed`, and the ONLY derivation-gate failure was my hand-written pin while every derived assertion stayed green. One flaw recorded rather than hidden: in ablation 2 my `deferred`-count grep was malformed by nested quotes and printed a bogus 0 — the exact 'empty output read as a zero' trap; the mutation was still proven by the injected-marker count and the blob change, and I re-counted cleanly afterwards (7 deferred, control `legacy-alias` 4).",
  "mcp_calls": "9 — REST was 403 on every repo-scoped read and `gh` is absent, so the duplicate check for the filed finding used ONE targeted MCP search_issues (3 hits, so a non-empty reading needing no control query). Channel switch declared.",
  "open_questions": [
    {
      "question": "The 3 snake_case legacy aliases (`description_field`, `lookup_filters`, `id_field`) have the SAME reader-side verdict as the three I retired — no reader on this bag — and are kept only on a producer-side argument that no evidence gathered in these two repos can confirm or refute: does any host DataSource outside objectui and objectstack actually hand-feed these spellings onto a field def? Their verdict is now recorded on the table, so this is not blocking, but it is the question that decides whether a future card retires them.",
      "options": [
        "A. Leave as is. The reader-side verdict is now recorded with the open producer question stated beside it, so the next pass inherits a measurement instead of a silence.",
        "B. Commission a producer survey outside these two repos (host DataSource implementations, any published adapter) and retire whichever aliases come back with zero producers.",
        "C. Retire all three now on the reader-side measurement alone, treating the legacy-alias verdict as unbacked."
      ],
      "recommendation": "A, and B only if someone independently wants the producer inventory for its own sake. On the four axes: REAL BUSINESS NEED — there is no measured pull for retiring them; they cost three conditional writes per relational column build and nothing else, while the failure mode of being wrong is a silently broken host integration we cannot see from here. LONG-TERM SOUNDNESS — the honest state is now encoded rather than implied; retiring on reader-side evidence would set the precedent that a back-compat verdict can be overturned by a measurement that does not address it. AI-ERROR-RESISTANCE — this is the axis that argues hardest against C: these are runtime-only spellings no spec-compliant producer can emit, so the strict FieldSchema already stops an AI-authored app from declaring them, and removing the runtime tolerance buys no authoring-time safety while risking a live host. STARTUP SCOPE — C spends real risk on tidiness with zero pull, and B spends survey effort on the same. C is explicitly out of scope per the ruling and I did not take it."
    }
  ],
  "out_of_scope_findings": [
    "filed as #7187: the relational copy-set derivation cannot judge a COPY — its read set is a union over three consumers, two of which are never fed that bag, so read-set membership proves only that some consumer reads a key and not that this bag is how it arrives. Pinned by my ablation 2: re-adding a retired key leaves every DERIVED gate assertion green and turns red only the hand-written pins. Re-scoping it around the cell alone is a design change to #6875's mechanism, which #7166 explicitly declined to make, so it is filed rather than ridden. Unassigned, labels finding + domain:ui.",
    "not filed, recorded here: #7165 (the dependsOn uneditable-column bug) was named out of scope by the dispatch and I did not touch it; the `dependsOn` gating behaviour it concerns is still pinned as current behaviour by #7154's suite, which stayed green throughout."
  ]
}
```

---

os-warren:

**Verdict: ACCEPT.** Arming on green.

### Ruling item 3 was answered first and answered properly

I gated this card on one question — is `RELATIONAL_META_KEYS` public surface? — because a "yes" made the whole thing unlandable at today's tier. It came back **internal**, and not on one probe but three, each carrying `useRowColor` as a live control that hit in all three:

1. not re-exported from the `plugin-grid` barrel (**fully enumerated** export list, no `export *` — which is what makes a negative readable at all);
2. the built `dist/index.d.ts` holds **zero** occurrences of any of the module's symbols;
3. absent from the built bundle's export list.

⭐ Three independent probes rather than one, and the `export *` check is the detail that matters: a barrel with a star export would have made probe 1 meaningless while still *looking* like a clean negative. Clause ② not engaged, correctly established rather than assumed.

### ⭐ The self-reported flaw is the most valuable line in the report

> *"in ablation 2 my `deferred`-count grep was malformed by nested quotes and printed a bogus 0 — the exact 'empty output read as a zero' trap"*

Reported rather than buried, with the recovery stated: the mutation was still proven by the **injected-marker count and the blob change**, so the ablation's validity never depended on the broken probe, and the count was re-taken cleanly afterwards (7 `deferred`, control `legacy-alias` 4).

⚠️ This is the **third** instance of that exact trap in this seat's work today — I hit it verifying #7139 (a malformed control that printed empty beside a subject's `0`), and again this morning pointing a pack probe at a barrel file. It is not a rare slip; it is the default failure of any grep written inside nested quoting. A dev who reports it, and who can show the result did not rest on it, is worth more than one who never trips it.

The same instinct shows up on the control-byte scan: the **first control was a tab, which proved nothing** because `0x09` is deliberately outside the class, so the pattern was re-verified against `0x01` and `0x0b`. Noticing that your control is a dud is the whole skill.

### The retirement was done through the table, and the gate came out stronger

Copy set **10 → 7**, verified at runtime, by flipping the three modern spellings to `deferred` — the verdict this repo already uses for the class, rather than a new one. And the gate was **strengthened, not weakened**: every `deferred` verdict is now mechanically proved spec-declared, and the two sibling pins turn the retired `lookupFilters` into a **live negative assertion on a fixture that still declares it**. That last part is what stops this retirement from being invisible to the next reader.

Ablation 2 is the proof the scope call was right: re-adding a retired key leaves **every derived gate assertion green** and turns red only the hand-written pins (`Tests 8 failed | 19 passed`). Ablation 1 split the population exactly as predicted (`3 failed | 2 passed`, B1/B2/B3 red, test A and the premise test green).

### The three legacy aliases: **A, leave as is** — ruling the open question

The snake_case aliases have the *same* reader-side verdict as the three retired, and are kept only on a producer-side argument no evidence from these two repos can address. **A** is right, and the axis that decides it is the one the report names: these are runtime-only spellings **no spec-compliant producer can emit**, so the strict `FieldSchema` already stops an AI-authored app from declaring them. Removing the runtime tolerance buys **no authoring-time safety** while risking a live host integration we cannot see. Three conditional writes per column build is not a cost worth that.

⛔ **C would set a precedent worth refusing explicitly: a back-compat verdict cannot be overturned by a measurement that does not address it.** The reader-side census is silent on producers by construction. Recording the verdict beside the open question — which is what shipped — is the durable improvement.

### #7187 is the right filing, and it generalises the lesson that produced this card

> *the relational copy-set derivation cannot judge a COPY — its read set is a union over three consumers, two of which are never fed that bag*

⭐ That is the **mechanised** form of the rule #6875 cost us: *a derivation establishes that a consumer reads a key; it does not establish that this bag is how it arrives.* And it is now **pinned by ablation 2** rather than asserted — the gate's own blindness is demonstrated, not argued. Re-scoping the derivation around the cell alone is a design change to #6875's mechanism, which this card explicitly declined to make, so filing was correct.

### Also correct

- **Lint run repo-wide** (`eslint . --no-inline-config`, 4100 files by eslint's own count, 75 with errors, all pre-existing, this diff E0) rather than declaring a narrowing. Stronger than the standard the dispatch asked for.
- **`type-check` first attempt was TS2307 across untouched files — an unbuilt dependency closure, reported as NOT MEASURED, not red.** Re-run after building 13 packages, with the script name echoed so it was not a zero-match no-op, and `--listFiles` confirming the new test is in the program (1 hit, control `lookupPickerKeys-7154` also 1).
- The canonical repo-root vitest invocation, because a package-cwd run is a false green here (#3378).
- **The duplicate footer left alone.** A PATCH edit would downgrade both footers to the bare form and lose the session URL — that is #6970's mangling class, and leaving a cosmetic duplicate beats destroying provenance.

---

Nothing to change.

---

PR #7193, opened 2026-09-01
fix(plugin-grid): feed the inline cell editor the row as dependent values, so a `dependsOn` lookup column is no longer permanently uneditable

os-warren:
Part of #7165

⛔ **#7165 must stay open.** This ships **option A**, the interim, and is deliberately not a closing reference to that card — the card stays open for **option B**, filed as #7188. See "Measurement 1" below — that is the fact that determined this.

## Measurement 1 — B **is** Clause ②-engaged, so A ships as the labelled interim

The PM ruling on #7165 made the A/B choice depend on one measurement: is `renderCellEditor`'s editor context type **published surface**, or internal to `plugin-grid`? Measured on `899730e0a`, **both** disjuncts of that test are true:

- **Reachable from `@object-ui/types`** — `packages/types/src/data-display.ts` declares `renderCellEditor?: (ctx: { column; row; value; stage; commit; cancel }) => React.ReactNode` on `DataTableSchema`, with a Zod mirror in `packages/types/src/zod/data-display.zod.ts`.
- **Part of `DataTableSchema`'s declared editor-context contract** — declared by **maintainer ruling 2026-08-30 (#6882)**, replacing the `(schema as any)` cast that preceded it. The shape is pinned by **exact type equality**, not `extends`, in `packages/types/src/__tests__/data-table-declared-keys-6882.test.ts`:

  ```
  type _RenderCellEditorShape = Expect[ Equal[
    Declared[DataTableSchema]['renderCellEditor'],
    ((ctx: CellEditorContext) => ReactNode) | undefined ] ];
  ```
  (angle brackets transcribed as square brackets — GitHub's body sanitizer eats
  tag-shaped fragments, including inside a fence. The source uses angle brackets.)

  with the stated reason: *"Declaring the key with any other shape is a different (and false) statement about the renderer, so the shape is pinned, not just the membership."*

`packages/plugin-grid/src/ObjectGrid.tsx` states the same fence in its own words: *"⛔ Do not 'fix' either hold by declaring the key on `DataTableSchema` as a rider — that package is published surface with its own review floor."*

⇒ B needs a **seventh member** on that declared context. That edits `@object-ui/types`, its Zod mirror, and a maintainer-ruled exact-shape pin. Per the ruling, A ships under all three of its conditions: labelled in code, this card not closed, and B filed (#7188).

⚠️ **A is not the conclusion.** A field that can never be filled is a hard defect with no workaround; a field scoped by the persisted parent instead of the staged one is wrong only where the parent was edited in the same unsaved session. Strictly better, strictly not finished.

## The defect

`LookupField` reads the record it gates on as `dependentValues ?? ctx.formValues ?? ctx.data ?? {}`. The grid's inline cell editor supplied **none** of the three — `renderCellEditor` rendered `FieldEditWidget` with `field` / `value` / `onChange` only, `SchemaRendererContext` has no `formValues`, and the grid sets no `ctx.data` per row. The record was therefore `{}` for **every** row, `dependenciesMissing` was permanently `true`, and a column declaring `dependsOn` rendered a disabled trigger reading "Select region first" **even when the row carried the parent value**.

PR #2216 was the two-half answer to #2215. Half 2 — every picker taking the `dependsOn` chain as a hard `baseFilter` — is host-independent and was **already live** on the grid path, which is why the gate fired at all (verified: A2.5 holds, see `LookupField`'s `dependentFilter` memo). Half 1 is per-host and the grid never got it. **This PR supplies that missing input and re-implements no cascade.**

## The change

One line of behaviour in `ObjectGrid.renderCellEditor` — `dependentValues={ctx.row}` — plus `row: any` added to the local ctx annotation, which had omitted it (the declared contract always carried `row`; only this local annotation did not). `FieldEditWidget` already forwards `dependentValues` to the widget, since #7008.

The interim is labelled **in the code**, naming this card and #7188, and stating precisely what is stale: *a parent edited but not yet saved in the same row does not re-scope the child.*

## Evidence

**Rendering proof**, `packages/plugin-grid/src/__tests__/gridDependentValues-7165.test.tsx` (4 tests, each with a live control in the same render):

1. the `dependsOn` column opens (was `lookup-trigger-gated` + `disabled`; now `lookup-trigger-regional_owner` + enabled) — **control**: a no-`dependsOn` column in the same render is also enabled, so the picker path is proven reached.
2. the picker is **scoped** by the row: `region: 'north'` reaches the query as a hard `$filter` and Person 07 (south) is absent — **control**: the no-`dependsOn` sibling over the same records **does** offer Person 07, so the absence is scoping and not a dead picker.
3. **negative control**: a row whose saved parent is empty **still gates**, so the gate was not simply disabled — **control**: the filled row in the same render opens.
4. ⚠️ **the interim's staleness, pinned as current behaviour**: staging a new parent without saving does **not** re-scope the child — with its own proof that the staging actually landed (the region cell renders the pending `'south'` while the saved record still reads `'north'`), because without that the assertion would pass for the trivial reason that nothing was ever staged. #7188 flips this test.

**The #7154 pin was UPDATED, not deleted.** The `dependsOn` case in `lookupPickerKeys-7154.test.tsx` pinned the **gated** behaviour, i.e. it encoded the defect. A deleted pin is indistinguishable from a pin that never existed, so it is rewritten in place to assert the fixed behaviour plus the scoping, and its header prose is rewritten to say so.

**Ablation** — direction predicted **before** running: removing the single `dependentValues={ctx.row}` line turns red the 4 new tests plus the updated #7154 case, and leaves the other 4 #7154 cases green.

- *No rebuild required, and this is measured rather than assumed*: the root `vitest.config.mts` aliases every `@object-ui/*` specifier to that package's `src`, and the tests import `../ObjectGrid` relatively — so nothing resolves through `dist` and a source mutation is directly live.
- *Mutation confirmed on disk* by **marker count** (`dependentValues={ctx.row}`: 1 → 0) **and blob hash** (`e8ab332e…` → `f219c865…`), not by the editor's exit code.
- *Result*: **`Tests 5 failed | 4 passed (9)`** — exactly the predicted set, passing count included.
- *Restore proven by state*, with `HEAD` named explicitly (`git checkout HEAD -- ...`, never the bare form, which restores from the index): `git diff HEAD`, `git diff --cached` and `git status --short` all empty, and the blob hash back to `e8ab332e…`. The script carried a `trap ... EXIT INT TERM` with absolute paths.

## Gates

⚠️ **Two heads, stated separately so the provenance is checkable.** The full set ran at `f17b08047`; `main` then advanced (#7180 landed), so it was merged in and the affected-package suite plus the cheap ratchet gates were re-run at the current head **`aba0c7499`**. Nothing below is quoted from a tree that is no longer this branch's head.

Re-run at **`aba0c7499`** (current head, post-merge):

| gate | result |
|---|---|
| `packages/plugin-grid/` full suite | **105 files, 949 tests passed** |
| `check:control-bytes` | OK — 5958 tracked text files scanned |
| `check:phantom-deps` | OK |
| `check:changeset-no-major` | OK |

Run at `f17b08047` (same tree for every path this PR touches — the merge brought in only `packages/components/src/renderers/layout/containers.tsx`, a `.changeset`, a docs page, a components test and one `vitest.config.mts` line, none of which this PR edits):

| gate | result |
|---|---|
| `type-check` (plugin-grid + closure) | 14 tasks successful |
| `lint` (affected package, **full**, not narrowed) | 2 tasks, **0 errors** (751 pre-existing warnings) |
| `check:self-import` | OK |
| `check:vi-mock-specifiers` / `check:vi-mock-inherit` | OK (4106 files, 525 with mocks) |
| `check-governed-queue-guard --test` | **NOT GOVERNED** — 4 paths vs 5 surfaces, none matched |

The typecheck genuinely covers the new tests: `tsc -p tsconfig.test.json --listFiles` lists all three edited files among its 1770 (this package typechecks tests via a dedicated project, so "typecheck green" is not silently excluding them).

⚠️ #7166 — the concurrent lane in this same package — had **not** landed as of `aba0c7499`. It touches `relationalMetaKeys.ts` and the `applyRelationalMeta` copy set; this PR touches `ObjectGrid.renderCellEditor`'s context construction. Disjoint symbols, but this branch will need a fresh merge if #7166 lands first.

## Incidence — measured, and ⛔ **not** "no impact"

Two corpora, each instrument run with a live control:

- **this repo**: 3 files match `dependsOn|depends_on`, and **none is this defect** — the gantt hit is `depends_on` as *task predecessor links* (a different key entirely) and the other is a `select` inside a **form**. Control: the same instrument returns 88 files across `packages/`.
- **`hotcrm`** (a real shipped app, anonymous shallow clone): **6 `dependsOn` declarations on lookup fields across 5 objects** (`crm_contract` ×2, `crm_case`, `crm_quote` ×2, `crm_opportunity`), all scoping contacts to `crm_account`. But **0 editable grids** in its views (control: the views directory exists and was read).

⇒ Neither measured corpus reaches the defect today, because none combines `dependsOn` with an editable grid — but `hotcrm` is **one metadata key away**: `editable: true` on any of those five objects' list views hits it immediately. Customer apps are unmeasured, and the parent pattern is demonstrably real in shipped metadata. This is a bounded reading, not an all-clear.

## Filed, not ridden

- **#7188** — option B, with the seam change spelled out. #7165 stays open for it.
- **#7190** — `finding`: `plugin-detail`'s `InlineFieldInput` supplies no `dependentValues` either (measured: zero non-test occurrences; control: 88 across `packages/`). ⚠️ Whether it actually gates is **unmeasured** and turns on whether the host sets `ctx.data`, which a one-record detail page plausibly does — so it is a `finding` with the probe written down, not a claim that the detail page is broken.

⚠️ Still live, per the PM's fence and not settled by this PR: **should the user be told the scope came from the saved row?** The interim's staleness is the same silent-wrongness class as the three refusals this seat landed in #7146 / #7147 / #7148. It is recorded on #7188 rather than left to disappear into a code comment.

---
Session: `session_012wwHa4aaFybxXrfmfHioDM` (https://claude.ai/code/session_012wwHa4aaFybxXrfmfHioDM) — kept in prose because a body edit rewrites the footer below to its bare form.

### DIFF
--- a/.changeset/7165-grid-dependent-values.md
+++ b/.changeset/7165-grid-dependent-values.md
@@ -1,0 +1,30 @@
+---
+'@object-ui/plugin-grid': patch
+---
+
+Fix: a `dependsOn` lookup column is no longer permanently uneditable in an
+editable `ObjectGrid`.
+
+`LookupField` resolves the record it gates on as
+`dependentValues ?? ctx.formValues ?? ctx.data ?? {}`, and the grid's inline
+cell editor supplied **none** of the three — `renderCellEditor` rendered
+`FieldEditWidget` with `field` / `value` / `onChange` only, `SchemaRendererContext`
+has no `formValues`, and the grid sets no `ctx.data` for a row. The resolved
+record was therefore `{}` for every row, so a column declaring `dependsOn`
+rendered a disabled trigger reading "Select region first" **even when the row
+carried the parent value**. The field could never be filled and nothing said
+why.
+
+PR #2216 closed #2215 in two halves: the form renderer injects its live watched
+record as `dependentValues`, and every picker takes the `dependsOn` chain as a
+hard `baseFilter`. The second half is host-independent and was already live on
+the grid path — which is why the gate fired at all. The first half is per-host
+and the grid never got it. `renderCellEditor` now passes
+`dependentValues={ctx.row}`, supplying that missing input; no cascade is
+re-implemented.
+
+⚠️ Interim, and deliberately labelled as such in the code (#7165): `ctx.row` is
+the **saved** record, so a parent edited but not yet saved in the same row does
+not re-scope the child — it stays scoped by the persisted value. Matching the
+form's live-record semantics needs a new member on `renderCellEditor`'s
+published context type and is tracked as #7188.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -3727,7 +3727,7 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
     // handing DataTable an editor factory would leave the built-in fallback
     // editors as the only reachable ones if any future path re-opened the mode.
     renderCellEditor: inlineEditable
-      ? (ctx: { column: any; value: any; stage: (v: any) => void; commit: (v?: any) => void }) => {
+      ? (ctx: { column: any; row: any; value: any; stage: (v: any) => void; commit: (v?: any) => void }) => {
           const fieldDef = (objectSchema as any)?.fields?.[ctx.column?.accessorKey];
           if (!fieldDef || !hasFieldEditWidget(fieldDef.type)) return null;
           const discrete = DISCRETE_EDIT_TYPES.has(fieldDef.type);
@@ -3747,6 +3747,45 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
               field={field}
               value={ctx.value}
               onChange={(v: any) => (discrete ? ctx.commit(v) : ctx.stage(v))}
+              // ⚠️ INTERIM (objectui#7165) — the SAVED row, not the staged one.
+              //
+              // The record a dependent widget scopes itself by. `LookupField`
+              // resolves `dependentValues ?? ctx.formValues ?? ctx.data ?? {}`
+              // and this grid supplied NONE of the three, so the resolved record
+              // was `{}` for every row. A column declaring `dependsOn` therefore
+              // rendered a permanently gated, disabled trigger ("Select region
+              // first") even when the row carried the parent value — a field
+              // that could never be filled, with no diagnostic. PR objectui#2216
+              // gave the FORM renderer exactly this injection (its live watched
+              // record); only that half was per-host, and the grid never got it.
+              // The other half — every picker taking the `dependsOn` chain as a
+              // hard `baseFilter` — is host-independent and was already live
+              // here, so this line supplies a missing INPUT and re-implements no
+              // cascade.
+              //
+              // ⛔ WHAT IS STILL WRONG, precisely: `ctx.row` is the PERSISTED
+              // record. A parent edited but NOT YET SAVED in this same row does
+              // not re-scope the child — the picker keeps listing candidates for
+              // the parent's saved value, and stays gated if that saved value is
+              // empty. objectui#2215's form fix was explicitly the LIVE record,
+              // so picking a parent re-scopes the child immediately. Matching
+              // that is objectui#7188, and it is the finished shape.
+              //
+              // Why the interim ships instead of the finished shape: the staged
+              // values live in `data-table`'s `pendingChanges` — in scope at the
+              // call site, so this is not a plumbing problem — and carrying them
+              // across needs a SEVENTH member on `renderCellEditor`'s context.
+              // `@object-ui/types` declares that context (objectui#6882,
+              // maintainer ruling 2026-08-30, replacing a `(schema as any)` cast)
+              // and pins its shape by EXACT type equality. That is a
+              // published-surface contract change with its own review floor, so
+              // it belongs to objectui#7188, not to this line.
+              //
+              // ⛔ Do NOT read this as settled. "Never fillable" → "scoped by
+              // the saved parent" is strictly better and strictly not finished;
+              // whether the user should be TOLD the scope came from the saved row
+              // is an OPEN question on objectui#7188, not a closed one.
+              dependentValues={ctx.row}
             />
           );
         }
--- a/packages/plugin-grid/src/__tests__/gridDependentValues-7165.test.tsx
+++ b/packages/plugin-grid/src/__tests__/gridDependentValues-7165.test.tsx
@@ -1,0 +1,290 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7165 — the grid's inline editor SUPPLIES the dependent record, so a
+ * `dependsOn` lookup column is editable instead of gated forever.
+ *
+ * ## The defect this closes
+ *
+ * `LookupField` resolves `dependentValues ?? ctx.formValues ?? ctx.data ?? {}`
+ * and `ObjectGrid`'s `renderCellEditor` supplied NONE of the three: it rendered
+ * `FieldEditWidget` with `field` / `value` / `onChange` only, `SchemaRendererContext`
+ * has no `formValues`, and the grid sets no `ctx.data` for a row. The resolved
+ * record was therefore `{}` for EVERY row, `dependenciesMissing` was permanently
+ * `true`, and a column declaring `dependsOn` rendered a disabled trigger reading
+ * "Select region first" — even when the row carried the parent value. The field
+ * could never be filled and nothing said why.
+ *
+ * PR objectui#2216 closed objectui#2215 in two halves: the FORM renderer injects
+ * its live watched record as `dependentValues`, and every picker surface takes
+ * the `dependsOn` chain as a hard `baseFilter`. Half 2 is host-independent and
+ * was ALREADY live here — which is why the gate fired at all. Half 1 is
+ * per-host and the grid never got it. This card supplies that missing input; it
+ * re-implements no cascade, and `test 2` below is what proves that distinction
+ * rather than asserting it.
+ *
+ * ## ⚠️ INTERIM — this ships option A, and option A is not the conclusion
+ *
+ * `renderCellEditor` now passes `dependentValues={ctx.row}`, and `ctx.row` is
+ * the SAVED record. A parent edited but not yet saved in the same row does not
+ * re-scope the child. That is strictly better than a field that can never be
+ * filled and strictly not finished — the form's answer to objectui#2215 was the
+ * LIVE record. Carrying the staged record needs a seventh member on
+ * `renderCellEditor`'s context, which `@object-ui/types` declares (objectui#6882,
+ * maintainer ruling 2026-08-30) and pins by EXACT type equality — a
+ * published-surface contract change, filed as objectui#7188.
+ *
+ * ⭐ `test 4` pins that staleness AS CURRENT BEHAVIOUR, with its own proof that
+ * the staging actually happened (otherwise "still scoped by north" is true for
+ * the trivial reason that nothing was ever staged). objectui#7188 flips it, and
+ * it is the assertion that fails if someone later "simplifies" B back to A.
+ *
+ * ## Why every test carries a live control
+ *
+ * An enabled-side green is worthless if the control column is also broken. Each
+ * test renders the `dependsOn` column and a control column with the SAME
+ * reference and the SAME records in ONE render, differing only in the declared
+ * key — the shape objectui#6875 established and objectui#7154 reused.
+ */
+import { describe, it, expect, vi, beforeAll } from 'vitest';
+import { render, screen, waitFor, fireEvent, within } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import React from 'react';
+
+import { ObjectGrid } from '../ObjectGrid';
+import { registerAllFields } from '@object-ui/fields';
+import { ActionProvider, SchemaRendererProvider } from '@object-ui/react';
+
+registerAllFields();
+
+const OBJECT = 'os_7165_task';
+const REF = 'os_7165_person';
+
+/** Six north, six south — so "scoped" and "unscoped" are different lists. */
+const PEOPLE = Array.from({ length: 12 }, (_, i) => ({
+  id: `p${i + 1}`,
+  name: `Person ${String(i + 1).padStart(2, '0')}`,
+  region: i < 6 ? 'north' : 'south',
+}));
+
+beforeAll(() => {
+  if (!Element.prototype.scrollIntoView) Element.prototype.scrollIntoView = vi.fn() as any;
+  if (!(Element.prototype as any).hasPointerCapture) (Element.prototype as any).hasPointerCapture = () => false;
+  if (!(Element.prototype as any).setPointerCapture) (Element.prototype as any).setPointerCapture = () => {};
+  if (!(Element.prototype as any).releasePointerCapture) (Element.prototype as any).releasePointerCapture = () => {};
+});
+
+/**
+ * The referenced-object query honours the `$filter` record, so the dependent
+ * cascade is observable as RENDERED ROWS and not only as call arguments.
+ */
+function makeDataSource(rows: any[]) {
+  const refQueries: any[] = [];
+  return {
+    refQueries,
+    find: vi.fn(async (objectName: string, params: any) => {
+      if (objectName === REF) {
+        refQueries.push(params);
+        let recs = PEOPLE;
+        const filter = params?.$filter;
+        if (filter && typeof filter === 'object' && filter.region) {
+          recs = recs.filter((p) => p.region === filter.region);
+        }
+        const top = params?.$top ?? 50;
+        const skip = params?.$skip ?? 0;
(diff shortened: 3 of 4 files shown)


os-warren:
os-dev-report

```json
{
  "issue": 7165,
  "status": "done",
  "branch": "claude/issue-7165-grid-dependent-values",
  "pr": "https://github.com/objectstack-ai/objectui/pull/7193",
  "premise_still_valid": true,
  "summary": "MEASUREMENT 1: B IS Clause 2-engaged — so option A shipped as the labelled interim and ⛔ #7165 MUST STAY OPEN (the PR says 'Part of #7165', never a closing keyword). The editor context is published surface on both disjuncts of the test: `DataTableSchema.renderCellEditor` is declared in packages/types/src/data-display.ts with a Zod mirror, declared by maintainer ruling 2026-08-30 (#6882) replacing a `(schema as any)` cast, and its exact shape is pinned by an `Equal` (not `extends`) type assertion in packages/types/src/__tests__/data-table-declared-keys-6882.test.ts whose comment states the shape is pinned deliberately; ObjectGrid.tsx independently fences that package as 'published surface with its own review floor'. B needs a seventh ctx member, so it edits @object-ui/types, its Zod mirror and that maintainer-ruled pin. All three ruling conditions are met: the interim is labelled in code naming #7165 and the follow-up and stating exactly what is stale, the card is not closed, and B is filed as #7188. Premise re-measured on head 899730e0a: the existing pin still passed before my change, i.e. the defect was still live. A2.1/A2.3/A2.4/A2.5 all confirmed; A2.2 needed one CORRECTION — the declared contract always carried `row`, but ObjectGrid's LOCAL ctx annotation omitted it, so A was a two-part edit (widen the local annotation, then pass it), not a literal one-liner.",
  "tests": "All runs from repo root via the shared verify lock (objectui forbids `pnpm --filter PKG test`; path filters only). BASELINE on 899730e0a before any edit: `pnpm exec vitest run packages/plugin-grid/src/__tests__/lookupPickerKeys-7154.test.tsx` -> 'Test Files 1 passed (1) / Tests 5 passed (5)' — the gated pin still passed, so the defect was live on head. AFTER FIX, same command -> 'Tests 1 failed | 4 passed (5)', 'AssertionError: expected lookup-trigger-regional_owner to be lookup-trigger-gated' — the pin encoded the defect, so it was UPDATED IN PLACE (not deleted), header prose rewritten to say so. NEW FILE packages/plugin-grid/src/__tests__/gridDependentValues-7165.test.tsx, 4 tests each with a live control in the same render: (1) dependsOn column now opens, control column also opens; (2) picker SCOPED — $filter.region='north' reaches the query, Person 07 (south) absent, while the no-dependsOn control over the same records DOES offer Person 07; (3) NEGATIVE CONTROL — a row whose saved parent is empty still gates, proving the gate was not disabled; (4) the interim's staleness pinned as current behaviour, with its own proof the staging landed (region cell renders pending 'south' while the saved record reads 'north'), which is what stops it passing vacuously. Both files together: 'Test Files 2 passed (2) / Tests 9 passed (9)'. ABLATION — direction predicted BEFORE running (remove `dependentValues={ctx.row}` -> the 4 new tests plus the updated 7154 case go red, the other 4 stay green). NO REBUILD NEEDED, and that is measured not assumed: root vitest.config.mts aliases every @object-ui/* specifier to that package's src and the tests import ../ObjectGrid relatively, so nothing resolves through dist and a source mutation is directly live. Mutation confirmed ON DISK by marker count (1 -> 0) AND blob hash (e8ab332e… -> f219c865…), never by the editing tool's exit code. Result: 'Tests 5 failed | 4 passed (9)' — exactly the predicted set, passing count included. Restore proven BY STATE with HEAD named explicitly (`git checkout HEAD -- ABSPATH`, never the bare form which restores from the index): git diff HEAD, git diff --cached and git status --short all empty and the blob hash back to e8ab332e…; the script carried `trap ... EXIT INT TERM` with absolute paths. GATES: at final head aba0c7499 (after merging main, which had advanced to 8ec11e14f / #7180) — full packages/plugin-grid/ suite 'Test Files 105 passed (105) / Tests 949 passed (949)', check:control-bytes OK (5958 files), check:phantom-deps OK, check:changeset-no-major OK. At f17b08047 (identical tree for every path this PR touches): type-check plugin-grid + closure 14 tasks successful, lint of the affected package in FULL (not narrowed) 2 tasks 0 errors (751 pre-existing warnings), check:self-import OK, check:vi-mock-specifiers / check:vi-mock-inherit OK (4106 files, 525 with mocks), check-governed-queue-guard --test = NOT GOVERNED (4 paths vs 5 surfaces). Verdicts read from each gate's own printed line, with exit codes captured after redirection, never through a pipe. NOT-MEASURED check cleared: `tsc -p tsconfig.test.json --listFiles` lists all three edited files among its 1770, so the green typecheck genuinely covers the new tests rather than excluding them.",
  "mcp_calls": "11 — 2 issue reads, 2 targeted search_issues (each with a control word that hit: #7165 itself, then #7154/#2215), 2 issue creates, 1 PR create, 1 PR body read-back, 1 PR body update, 2 report comments. REST was 403 for repo-scoped reads and `gh` is absent, both measured with a control before switching channels.",
  "open_questions": [
    {
      "question": "Should the grid TELL the user that a dependsOn picker is scoped by the SAVED parent while an unsaved edit to that parent is staged in the same row? The PM's ruling explicitly left this live rather than settled, and asked that it not disappear into a code comment.",
      "options": [
        "A. Say nothing (what this PR ships). The picker silently uses the persisted parent. Zero new surface, but it is the same silent-wrongness class as the three refusals this seat landed in #7146 / #7147 / #7148.",
        "B. Surface a hint on the trigger/popover when the parent has a staged-but-unsaved value that differs from the saved one — e.g. 'scoped by the saved Region; save the row to re-scope'. Needs the staged record, i.e. it is gated behind exactly the same seam change as option B of this card.",
        "C. Do not build a message at all: land #7188, which removes the staleness and therefore removes the question."
      ],
      "recommendation": "C, and treat B as dead. Real business need: the staleness only exists because the interim exists, so a message would be UI built to describe a defect we have already scheduled to delete. Long-term soundness: contract-first — the honest fix is the declared seam member, not a renderer-side apology for a wrong input, and a warning would fossilize the interim as if it were a supported mode. Making AI-authored metadata hard to get wrong: an author declaring dependsOn should get correct cascading, not a caption explaining why it is stale — a warning trains authors to accept a wrong scope. Startup scope discipline: option B costs the SAME seam change as #7188 while delivering strictly less, which is the worst trade of the three. ⚠️ The one thing that would flip this to B: if #7188 is declined or deferred long, the silent staleness stops being temporary and then it must be surfaced. So this question should be closed by #7188 landing, not by a ruling that silence is acceptable."
    }
  ],
  "out_of_scope_findings": [
    "filed as #7188: option B — carry the pending/staged record across the renderCellEditor seam so a dependsOn lookup is scoped by the in-flight parent (the seam change is spelled out: @object-ui/types ctx member + Zod mirror + the #6882 exact-shape pin + data-table.tsx passing `{ ...row, ...rowChanges }` + ObjectGrid reading it). Clause 2-engaged; #7165 stays open for it.",
    "filed as #7190 (label `finding`): plugin-detail's InlineFieldInput supplies no dependentValues either (measured: zero non-test occurrences; control: 88 files across packages/), and plugin-detail never provides SchemaRendererContext `data` itself. ⚠️ Deliberately NOT graded a bug — a detail page renders ONE record, so ctx.data may legitimately rescue the cascade there, and no rendering measurement was made. The single probe that settles it is written into the card.",
    "NOT filed, reported here as an observation for PM triage: ObjectGrid.tsx still lists `renderCellEditor` in its `ObjectGridDataTableSchemaHolds` type as an undeclared-but-live HELD key, with prose saying whether DataTableSchema should declare it is a pending packages/types ruling. That ruling HAPPENED — #6882 declared it on 2026-08-30 — so the hold entry and its surrounding census comment are stale. Harmless today (the intersection with the declared member is identical), but it is documentation that now states the opposite of the tree. Left untouched because it is outside this card and inside the file #7166 is adjacent to."
  ]
}
```

---

os-warren:

**Verdict: ACCEPT.** Arming on green. The open question is ruled below.

### ⭐ The pre-flight scan that saved the card — and a trap nobody had written down

> *a first draft of the PR body contained the phrase **"does NOT close #7165"**, which a pre-flight scan caught — GitHub's parser ignores negation and it would have closed the card on merge.*

**GitHub's closing-keyword parser does not read negation.** *"does NOT close #7165"* contains `close #7165`, and on merge it closes the card. This PR's entire correctness depends on #7165 staying open — A is an interim, B is the real fix — so the most natural sentence a careful author would write to protect the card is the one that would have destroyed it.

⚠️ And note the shape: the failure would have been **silent and delayed**. The PR merges, the card closes, and the interim becomes permanent because nothing is left tracking B. Carrying this to **#6970** as a fourth mangling class, since that card enumerates exactly this family. The safe form is what shipped: `Part of #7165`, and never the keyword at all.

### Measurement 1 answered decisively, on both disjuncts

B **is** Clause ②-engaged, and it was established rather than assumed:

- `DataTableSchema.renderCellEditor` is declared in `packages/types/src/data-display.ts` **with a Zod mirror**;
- it was declared by **maintainer ruling #6882 (2026-08-30)**, replacing a `(schema as any)` cast;
- its exact shape is pinned by an **`Equal` (not `extends`)** assertion in `data-table-declared-keys-6882.test.ts`, whose comment says the shape is pinned deliberately;
- `ObjectGrid.tsx` independently fences that package as *"published surface with its own review floor."*

⇒ B needs a seventh ctx member, so it edits `@object-ui/types`, its Zod mirror, **and a maintainer-ruled exact-shape pin**. That is Clause ② three times over. ⭐ The `Equal`-not-`extends` detail is what makes it airtight — an `extends` pin would have tolerated an added member, and the answer would have been different.

All three of my ruling conditions are met: the interim is labelled in code naming #7165 and #7188 and stating exactly what is stale; the card is not closed; B is filed.

### A2.2 corrected, and the correction matters

I said A was "one line: `dependentValues: ctx.row`". Wrong: the **declared** contract always carried `row`, but `ObjectGrid`'s **local** ctx annotation omitted it — so A was a two-part edit (widen the local annotation, then pass it). ⚠️ A declared member that a local annotation silently narrows is its own small hazard: the type says the data is there and the call site cannot see it.

### The test discipline is the strongest part

- **Baseline before any edit**: the existing #7154 pin *still passed* on head — so the defect was live, not already fixed by something else. Establishing that first is what makes the rest meaningful.
- **The pin encoded the defect and was UPDATED IN PLACE, not deleted**, with its header prose rewritten to say so. A deleted pin is indistinguishable from a pin that never existed.
- **A negative control**: a row whose saved parent is empty *still gates* — proving the fix supplied a value rather than disabling the gate. That is the assertion that separates "cascade works" from "cascade removed", and it is the one most likely to be skipped.
- **The staleness pin carries its own proof it is not vacuous**: the region cell renders the pending `south` while the saved record reads `north`, so the test proves the staging actually landed before asserting the picker ignored it. A staleness test without that proof passes for the wrong reason.
- Ablation predicted before running, `Tests 5 failed | 4 passed (9)` exactly; mutation by marker count **and** blob hash; restore by state with `git checkout HEAD -- ABSPATH` — **never the bare form**, which restores from the index.

### ⚖️ Ruling on the open question: **C** — close it by landing #7188, not by ruling that silence is acceptable

I left this live at dispatch and asked that it not disappear into a code comment. The answer is C, and the argument that decides it is the one I did not have when I asked:

⭐ **Option B costs the *same seam change* as #7188 while delivering strictly less.** A warning that says "scoped by the saved Region" needs the staged record — i.e. exactly the ctx member #7188 adds. Having paid that price, you would ship a caption explaining the wrong scope instead of the right scope. That is the worst of the three trades.

And a warning would **fossilise the interim as a supported mode**. An author who declares `dependsOn` should get correct cascading, not a caption explaining why it is stale — a message trains authors to accept a wrong scope as normal.

⚠️ **I am adopting the flip condition verbatim, because it is what keeps C honest**: if #7188 is declined, or deferred long enough that the interim stops being temporary, the silent staleness must be surfaced. So **this question is closed by #7188 landing, not by a ruling that silence is fine.** Recorded on #7165.

### Incidence: carried, not closed — exactly as the dispatch required

Two corpora, each with a live control:
- **this repo**: 3 files match `dependsOn|depends_on`, **none is this defect** (the gantt hit is `depends_on` as task predecessor links — a different key entirely); control: 88 files across `packages/`.
- **hotcrm**: **6 `dependsOn` declarations on lookup fields across 5 objects**, all scoping contacts to `crm_account` — but **0 editable grids in its views** (control: the views dir exists and was read).

⇒ ⭐ *"Neither measured corpus reaches the defect today, but hotcrm is **ONE metadata key away**"* — `editable: true` on any of those list views. That is a far more useful answer than either "no impact" or "unmeasured", and it is the sentence a maintainer needs to prioritise #7188. Customer apps remain unmeasured, and that is stated rather than glossed.

### #7190 filed with the right restraint

`InlineFieldInput` supplies no `dependentValues` either (measured: zero non-test occurrences; control: 88 files) — and it is **deliberately NOT graded a bug**, because a detail page renders *one* record so `ctx.data` may legitimately rescue the cascade there, and no rendering measurement was made. The single probe that would settle it is written into the card. Declining to grade on a structural analogy, and leaving the probe behind, is better than a confident grade either way.

---

Nothing to change.

---

PR #7197, opened 2026-09-01
fix(plugin-charts): scatter says when it cannot place a row, instead of drawing an empty axis

### LINKED ISSUE #7171: finding(plugin-charts): scatter is NOT MEASURED for the degenerate-magnitude class — its all-positive CONTROL drew zero marks, so the sweep's zeros say nothing
Filed by the `domain:ui` PM seat from a measurement in objectui#7147 (PR objectstack-ai/objectui#7169). The implementer reported it rather than filing, per the report-don't-duplicate fence on that surface — but unlike objectui#7148's finding, this one has **no existing card to attach to**, so it gets its own.

## What is NOT known

objectui#7147 swept **8 chart families × 9 datasets = 74 tiles** in real Chromium, MD5'd and pixel-diffed against a literally empty div, to find where a degenerate magnitude (zero, negative, `null`, unparseable) produces a chart that misrepresents its data.

It reached verdicts for seven families:

| family | verdict |
|---|---|
| pie / donut / funnel / treemap | **defective** — fixed in PR objectui#7169 |
| bar / radar | **measured and clean** — deliberately untouched |
| **scatter** | ⛔ **NOT MEASURED** |

**Scatter is not clean. It is unmeasured**, and the distinction is the whole point of this card.

## Why the sweep could not read it

**Every scatter tile drew zero marks — including the all-positive CONTROL** (`markN 0`).

A control that returns zero is **no control**. With the control dead, scatter's zeros on the degenerate datasets carry no information: they are indistinguishable from "the fixture never rendered a scatter at all." Reporting scatter as clean on that evidence would have been a false green across nine datasets.

**The cause is identified**: scatter takes **two** measures (x and y), and objectui#7147's fixture supplies a single measure. The fixture does not bind scatter, so scatter never had a chance to draw.

## What a correct measurement needs

- A **two-measure fixture** for the scatter arm, so its all-positive control actually draws marks. ⛔ Until that control returns non-zero, no scatter reading means anything.
- Then the same nine datasets, with the degenerate value applied to **each measure independently and to both** — a scatter can plausibly degenerate on x, on y, or on both, and those may not behave alike.
- Per-tile screenshot hashing against per-family all-positive **and** one-row controls, the instrument that produced verdicts for the other seven families. A hash collision between a degenerate tile and a legitimately-sparse one is the finding; that is how pie's full-circle case and treemap's four-datasets-to-one-image case were caught.

## Blocked, and on what

`pm:blocked` on PR objectstack-ai/objectui#7169 landing — it is live in `packages/plugin-charts/src/AdvancedChartImpl.tsx`, which is where any scatter arm work would sit. Unblocks automatically.

## Why this is worth tracking rather than dropping

⚠️ **This is the shape that produces false confidence.** After PR objectui#7169 lands, the honest statement is *"seven of eight families measured; scatter unknown."* Without this card, the natural summary becomes *"the degenerate-magnitude class is handled,"* which is not what was measured — and the next person to look would have to re-derive the whole sweep to discover the hole.

⚠️ And note the pattern this card family keeps producing: objectui#7140 (early return), objectui#7147 (degenerate geometry), objectui#7148 (silent row drop) are **three distinct mechanisms behind one reader-facing symptom**. Scatter may be a fourth. ⛔ Do not assume PR objectui#7169's predicate transfers to it — measure first.

### Related

objectui#7147 / PR objectui#7169 (the sweep and the fix) · objectui#7148 (silent row drop) · objectui#7140 (the empty-frame early return) · objectui#7146.

os-warren:
Fixes #7171

Re-derived on `origin/main` **899730e0a**, which already carries PR #7169's `no-positive-magnitude`, PR #7146's `no-positive-flow` and PR #7161's `ChartFootnote`. Gates and numbers below are quoted at the committed head **9953dd5fc**.

## The one line that makes the rest readable: THE CONTROL DREW

#7147's sweep reported `markN: 0` on every scatter tile **including its all-positive control**, so scatter came out NOT MEASURED — a control that returns zero is no control, and none of those zeros carried information.

A two-measure fixture was built and its control **drew 3 of 3 marks** (245px of mark area), on a run whose live console control also fired. Everything below is therefore a reading, not a zero.

The cause is confirmed exactly as PM assumed (A2.1): scatter's `XAxis` and `YAxis` are BOTH `type="number"`, bound to `xAxisKey` and `series[0].dataKey`. #7147's fixture put a category column (`stage`) on the numeric X axis. Reproduced here as its own tile: `scatter--fixture7147`, **0 marks**, and byte-identical (`51957063d9c2`) to five other broken shapes.

## Rendered it and looked first

33 tiles in real Chromium (`/opt/pw-browsers/chromium`), **one page load per tile** so console output is attributable to exactly one of them. Each tile screenshotted, MD5'd, decoded and pixel-diffed against a literally empty `div` of the same box — **520x360**, above the chart's own `CHART_MIN_HEIGHT` floor of 280, because PR #7169 measured that no footnote is visible at all below it.

`measured-and-declined` was genuinely on the table and it SURVIVED for three of the classes swept. What it did not survive for:

| dataset | what a reader saw at 899730e0a | verdict |
|---|---|---|
| both coordinates `null`, or both keys absent | **byte-identical (diff `0.000%`)** to a scatter handed NO ROWS AT ALL. The empty-result picture, for a query that returned rows | **fix** |
| `null` x · absent x · `'n/a'` x · `'Infinity'` x · boolean x · #7147's category column | **SIX datasets, ONE image** (`51957063d9c2`): a confident y scale, an x axis with no scale at all, and no marks. The y-side twin is a second four-way collision (`5a67d6ab8800`) | **fix** |
| one placeable row among three | **99.75% pixel-identical** to a genuinely one-row scatter (diff `0.250%`); the y-side twin `0.293%`. Exactly the collision that decided pie in #7147 — two rows vanish and the picture says "one point" | **fix** |
| all-negative · all-zero · mixed-sign coordinates | drew **3 / 2 / 3** marks, three distinct images, full scales | **declined** — correct charts |
| zero variance (3 rows at one coordinate) | 3 symbols, `0.017%` from a one-row scatter | **declined** — see below |
| constant x, constant y | every mark drawn, full scales on both axes | **declined** |

## The predicate is DERIVED for a positional chart, and every clause was forced by a measurement

⛔ `no-positive-magnitude`'s `Number.isFinite(v) && v > 0` does not transfer. Pie, funnel and treemap size a mark BY its measure; scatter plots POSITION, and a chart of temperatures or profit deltas is supposed to show zeros and negatives. Measured rather than argued: that predicate would have blanked the all-negative (3 of 3 marks), all-zero (2 of 2) and mixed-sign (3 of 3) tiles above — three correct charts.

What survived is `isPlottableCoord`, and `Number()` alone gets three of its cases wrong in both directions:

| value | `Number(v)` says | Recharts actually does | predicate |
|---|---|---|---|
| `null` | `0`, finite — plottable | **0 of 2 marks**, x axis renders no scale | **reject** |
| `''` | `0`, finite — plottable | **2 of 2 marks**, x ticks `0,1,2,3,4` | **accept** — rejecting it would blank a chart that draws |
| `'10'` | `10`, finite | 2 of 2 marks | accept |
| `'n/a'` / `'Infinity'` / absent | `NaN` / `Infinity` | 0 of 2 marks | reject |

The `''` row is the trap: it LOOKS like a null and reads like one. A predicate copied from the magnitude families, or written from `Number()` alone, gets it backwards.

### The clause that was pinned, went red, and was removed

A **boolean** coordinate was rejected in the first draft — the browser sweep had it drawing 0 of 2 marks, byte-identical to the `null` tile. Pinning that turned the mixed case red, and the mixed case is the real reading: **a boolean beside a genuinely numeric row draws every mark (3 of 3)**. Recharts needs one real number to build the scale and then coerces the booleans onto it.

So rejecting booleans would put a footnote reading "2 of 3 rows ... are not drawn" over three points visibly on screen — a false statement about the picture, which is worse than the silence this card is about. The predicate accepts booleans, the all-boolean tile stays exactly as silent as it is today (`51957063d9c2`, unchanged), and that residual hole is filed as #7195 with the reasoning, not hidden.

## What changed

Three helpers and one wiring, all scatter-only:

- `isPlottableCoord(v)` / `countPlottablePoints(rows, xKey, yKey)` — a point exists only if **both** coordinates are numbers.
- **No point placeable** with rows present: the file's existing refusal shell under its own code, `data-chart-error="no-plottable-points"`, naming **both** keys because a point needs both.
- **Some points placeable**: the plot renders unchanged inside the existing `ChartFootnote`, `data-chart-note="unplotted-points"`, carrying the count.

The y key is bound **once** (`const scatterYKey = series[0]?.dataKey || 'value'`) and handed to both the predicate and the `YAxis`, so the two cannot drift.

Copy names the **predicate**, not a cause — five shapes reach it and naming any one is false for the other four. Gated on `total > 0`, so every no-rows tile is untouched: that is the empty-RESULT question (#7130), answered upstream in `ObjectChart`. No console warning, matching all three landed answers. No i18n, matching the file.

### The one place this note differs from #7169's

`unsized-rows` deliberately does **not** say the rows are undrawn, because a mixed-sign pie paints a sector for every row. Scatter is the opposite case and it was measured: a 3-row dataset with one placeable pair emitted **exactly one** `path.recharts-symbols`. The rows really are absent from the picture, so saying so is true — and the count is the half a reader cannot recover from it.

## Before / after: 14 tiles moved, 19 are byte-identical

The 19 are what make the 14 discriminating.

- every legitimate scatter unchanged: the control, one-row, all-negative, all-zero, mixed-sign, zero-variance, constant-x, constant-y, `''`, `'10'` — **and no wrapper element added to any of them**
- `scatter--emptyRows` unchanged (`dce4d0dde6df`) — the #7130 gate holds
- `pie--positive`, `pie--allZero`, `sankey--allZero`, `sankey--posNull`, `bar--allZero` all byte-identical, **all three landed answers still firing under their own codes**
- the console control still fires; **zero console output across every other tile**

The two collisions are broken: both-coordinates-degenerate versus no-rows went `0.000%` to **`2.651%`**, and one-placeable-row versus one-row went `0.250%` to **`6.368%`**.

## Seam map, pinned from both sides

| dataset | sankey | pie / donut / funnel / treemap | scatter |
|---|---|---|---|
| no rows at all | untouched (#7130) | untouched (#7130) | untouched (#7130) |
| nothing drawable | `no-positive-flow` (#7146) | `no-positive-magnitude` (#7169) | `no-plottable-points` (this PR) |
| some drawable | `omitted-rows` (#7161) | `unsized-rows` (#7169) | `unplotted-points` (this PR) |
| all drawable | untouched | untouched | untouched, **no wrapper added** |

#7169's mutual-exclusivity test was **extended, not duplicated**: scatter and both new codes join that same assertion, over six codes and a scatter-specific dataset pair, so a sixth answer cannot be added later without it having an opinion.

## Assumptions falsified

1. **A2.3's headline — "the interesting failures are probably about axis domain collapse" — did not reproduce as collapse.** Zero variance, constant x and constant y all draw every mark with normally padded domains (zero-variance's x ticks are `0,3,6,9,12`, identical to a one-row scatter's). What the sweep found instead is an axis that renders **no scale whatsoever** when nothing on it is placeable, beside a confident scale on the other axis — a different failure, and the one now answered. Zero variance is **measured and declined**: three coincident points ARE one dot, which is overplotting, a property of the form, and a TRUE picture.
2. **A boolean coordinate is not a sibling of `null`** — see above. This was caught by a pin going red, not by reading.
3. **A2.4 was right that something pins scatter's current behaviour.** `AdvancedChartImpl.unprojectedSeriesDimension.test.tsx` asserts a `series: []` scatter "paints an svg" — using a category-column dataset that was never a valid scatter dataset (0 marks, measured). **Updated, not deleted**: its real subject (the `value` fallback must not be refused by #4683's guard) is preserved on a two-measure fixture, and a new test pins the seam from that side — the category-column dataset gets #7171's refusal and still never gets `no-plottable-series`.

## Tests

`AdvancedChartImpl.degeneratePosition.test.tsx`, 40 tests. The FIRST of them asserts the control draws — stated as an assertion, not an eyeballed observation, so that if it ever returns to zero every other assertion in the file reads as void rather than green.

```
pnpm exec vitest run packages/plugin-charts/src/AdvancedChartImpl.degeneratePosition.test.tsx
  Test Files  1 passed (1)
       Tests  40 passed (40)

pnpm exec vitest run packages/plugin-charts/
  Test Files  40 passed (40)
       Tests  371 passed (371)
```

### Ablation, direction and counts predicted before running

Predicted, written down first: leg A red = 15 refusal tests + 1 seam test + 1 in `unprojectedSeriesDimension` = **17**; leg B red = the **7** note tests; the two red sets **disjoint**.

|---|---|---|---|
| A | refusal guard neutered (`points.plottable === 0` to `=== -1`) | 17 red, note block green | **17 failed / 354 passed** — 15 in the refusal block, 1 in the seam block, 1 in `unprojectedSeriesDimension` |
| B | note guard inverted (`unplotted <= 0` to `>= 0`) | 7 red, refusal block green | **7 failed / 364 passed** — all 7 in the note block |

Each leg proved its mutation on disk by **anchor count and blob hash** (`aceb44fc...` to `d740c74f...` in A, to `53138e45...` in B; anchor 1 to 0, injected 0 to 1 in both) and its restore by **state** — `git diff HEAD` and `git diff --cached` both empty, blob back to `aceb44fc...` — under an absolute-path `trap`, with `git checkout HEAD -- path` naming HEAD explicitly rather than restoring from the index. The red sets are disjoint and each leg left the other side green, which is what makes them discriminating rather than a blanket break.

## Gates, all on the committed head `9953dd5fc`

| gate | verdict, as the gate printed it |
|---|---|
| `vitest packages/plugin-charts/` | `Test Files 40 passed (40)` / `Tests 371 passed (371)` |
| `type-check` | exit 0, `tsc --noEmit`, script name echoed |
| `lint` (package) | `299 problems (0 errors, 299 warnings)` — exit 0 |
| `check-changeset-presence` | `5 file(s) changed, 4 of them published source ...` |
| `check-changeset-no-major` | `No changeset declares a major bump.` |
| `check-changeset-overwrite` | `No pre-existing changeset was modified or deleted.` |
| `check-changeset-fixed` | `All workspace packages are in the changeset fixed group.` |
| `check-control-bytes` | `OK (scanned 5956 tracked text file(s); skipped 85 binary).` |
| `check-vi-mock-specifiers` | `OK (4106 tracked source file(s), 2378 test-named; 526 carry a mock ...)` |
| `check-vi-mock-inherit` | `OK (... 118 inherit, 0 auto-mocked ...)` |
| `check-i18n-keys` / `check-i18n-drift` | exit 0; `0 en value(s) changed` |
| `check-self-import` / `check-phantom-deps` | exit 0 |
| `check-esm-specifiers` / `check-side-effects-array` / `check-shell-escape-residue` | exit 0 |
| `check-lint-coverage` | `46/46 packages linted, 0 with outstanding errors (0 total).` |
| `check-type-check-coverage` | `41/41 packages compile their tests, 0 declared debt` |

The tracked-file gates were run **after** `git add`; their counts moved (5942 to 5956 text files, 525 to 526 carrying a mock), which is how I know they saw the new test rather than skipping it as untracked. `tsc --listFiles` confirms all four edited files are in the compiled set (1 hit each), so the type-check green covers them.

`check-eager-closure` exited **2** with `not applicable to a local run — this check compares a pull_request merge ref`. That is PRECONDITION NOT MET, not a failure; CI owns it.

The probe's first `vite build` was also NOT MEASURED, not red: `failed to resolve import "@object-ui/core"` because the dependency closure was unbuilt. Built with `pnpm --workspace-concurrency=2 --filter "@object-ui/plugin-charts^..." build` and re-run to a real verdict.

**Declared narrowing**: repo-wide `pnpm lint` was not run locally; CI owns it. The per-package run is a measurement, not a guess — eslint's **own** config selected **53** files in this package (`--format json` count, up from the 52 PR #7169 measured, which is the new test file), all four edited files are in that set with **0 errors**, and this repo's `eslint.config.js` extends `tseslint.configs.recommended` with no `parserOptions.project` and no `projectService` (0 matches), so no rule is type-aware and nothing in this diff can move a verdict in a file it does not touch.

## About the instrument

The sweep is a browser probe over a built bundle. It is **not committed**, deliberately: a test that cannot run in CI renders as coverage while measuring nothing. Its findings are committed instead, as the 40 assertions above, every one of which runs in CI on jsdom — including the control.

## Out of scope, reported rather than ridden

- **#7194** — a multi-series scatter draws every series at the FIRST series' y values. Measured: two series over `ym` and `zm` produced **4 symbols at 2 distinct positions**, each drawn twice; `zm`'s values appear nowhere. Good data, confidently wrong picture, different mechanism.
- **#7195** — the all-boolean coordinate hole this PR deliberately leaves, with the measurement that made leaving it the correct choice.
- The **sub-280 footnote invisibility** property is pre-existing and already recorded on #7147; this PR inherits it and does not worsen it, which is why every number above is quoted at 360.

---

### DIFF
--- a/.changeset/scatter-unplaceable-points.md
+++ b/.changeset/scatter-unplaceable-points.md
@@ -1,0 +1,21 @@
+---
+'@object-ui/plugin-charts': patch
+---
+
+Scatter now says when it cannot place a row, instead of drawing an empty axis.
+
+Scatter is the only two-measure positional chart in the renderer: `xAxisKey` feeds
+a numeric X axis and `series[0]` a numeric Y axis, so a point exists only when
+both are numbers. Measured in real Chromium, rows it could not place produced a
+tile byte-identical to a scatter handed no rows at all, and six different
+authoring failures shared one image. A chart with one placeable row among three
+was 99.75% pixel-identical to a genuinely one-row scatter.
+
+Handed rows it cannot place any of, a scatter now renders the file's refusal
+shell under `data-chart-error="no-plottable-points"`, naming both keys. When some
+rows place and some do not it draws as before with a `data-chart-note="unplotted-points"`
+footnote carrying the count. Charts whose rows all place are byte-identical to
+before, and no wrapper element is added to them.
+
+The predicate is positional, not magnitude-based: zero and negative coordinates
+are ordinary scatter data and keep drawing.
--- a/packages/plugin-charts/src/AdvancedChartImpl.degenerateMagnitude.test.tsx
+++ b/packages/plugin-charts/src/AdvancedChartImpl.degenerateMagnitude.test.tsx
@@ -257,25 +257,54 @@ describe('objectui#7147 — the seam with the two landed sankey answers, pinned 
     expect(noteOf(container)).toBeNull();
   });
 
-  it('the three codes are mutually exclusive on every dataset in the sweep', () => {
+  it('the codes are mutually exclusive on every dataset in the sweep', () => {
+    // Extended by objectui#7171 rather than duplicated beside: scatter and its
+    // two codes join the SAME exclusivity assertion, so a sixth answer cannot
+    // be added later without this test having an opinion about it.
     const datasets: Row[][] = [
       [{ stage: 'Alpha', amount: 0 }, { stage: 'Beta', amount: 0 }],
       [{ stage: 'Alpha', amount: 40 }, { stage: 'Beta', amount: null }],
       [{ stage: 'Alpha', amount: 40 }, { stage: 'Beta', amount: 25 }],
       [],
     ];
-    for (const family of [...MAGNITUDE_FAMILIES, 'sankey', 'bar']) {
-      for (const data of datasets) {
+    // Scatter reads TWO measures, so the sweep's single-measure rows are also
+    // the dataset that leaves it unplaceable — which is exactly the pair this
+    // assertion has to keep apart.
+    const scatterDatasets: Row[][] = [
+      ...datasets,
+      [{ stage: 1, amount: 40 }, { stage: 2, amount: 25 }],
+      [{ stage: 1, amount: 40 }, { stage: 2, amount: null }],
+    ];
+    for (const family of [...MAGNITUDE_FAMILIES, 'sankey', 'bar', 'scatter']) {
+      for (const data of family === 'scatter' ? scatterDatasets : datasets) {
         const { container } = renderChart(family, data);
         const codes = [
           container.querySelector('[data-chart-error="no-positive-flow"]'),
           container.querySelector('[data-chart-error="no-positive-magnitude"]'),
+          container.querySelector('[data-chart-error="no-plottable-points"]'),
           container.querySelector('[data-chart-note="omitted-rows"]'),
           container.querySelector('[data-chart-note="unsized-rows"]'),
+          container.querySelector('[data-chart-note="unplotted-points"]'),
         ].filter(Boolean).length;
-        expect(codes).toBeLessThanOrEqual(1);
+        expect(codes, `${family} / ${JSON.stringify(data)}`).toBeLessThanOrEqual(1);
         cleanup();
       }
     }
+  });
+
+  it('scatter never receives a MAGNITUDE code, and a magnitude family never receives scatter\'s', () => {
+    // objectui#7171. Same fence as the sankey pair above: pie sizes by
+    // magnitude and scatter plots by position, so an all-negative dataset is a
+    // refusal for one and perfectly ordinary data for the other.
+    const negatives: Row[] = [{ stage: -10, amount: -40 }, { stage: -20, amount: -25 }];
+    const { container: scatter } = renderChart('scatter', negatives);
+    expect(scatter.querySelector('[data-chart-error="no-plottable-points"]')).toBeNull();
+    expect(refusalOf(scatter)).toBeNull();
+    expect(noteOf(scatter)).toBeNull();
+    cleanup();
+
+    const { container: pie } = renderChart('pie', negatives);
+    expect(refusalOf(pie)).not.toBeNull();
+    expect(pie.querySelector('[data-chart-error="no-plottable-points"]')).toBeNull();
   });
 });
--- a/packages/plugin-charts/src/AdvancedChartImpl.degeneratePosition.test.tsx
+++ b/packages/plugin-charts/src/AdvancedChartImpl.degeneratePosition.test.tsx
@@ -1,0 +1,405 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Scatter — rows that cannot be PLACED (objectui#7171). The fourth mechanism on
+ * this surface, and the first positional one.
+ *
+ * ## Why this card exists at all: a control that returns zero is no control
+ *
+ * objectui#7147 swept eight chart families for degenerate magnitude and reached
+ * a verdict for seven. Scatter's every tile drew ZERO marks — its all-positive
+ * CONTROL included — because scatter takes TWO measures and that sweep's
+ * fixture supplied one. With the control dead, none of scatter's zeros carried
+ * information: they were indistinguishable from "the fixture never bound a
+ * scatter at all". Reporting scatter as clean on that evidence would have been
+ * a false green across nine datasets, so it was reported NOT MEASURED instead —
+ * which is neither red nor green.
+ *
+ * The first test below is therefore the load-bearing one: it asserts the
+ * control DRAWS. If it ever goes red, every other assertion in this file is
+ * void rather than passing, and that is the point of stating it as an assertion
+ * instead of an eyeballed observation.
+ *
+ * ## What a correct measurement found
+ *
+ * 33 tiles in real Chromium (`/opt/pw-browsers/chromium`) at `origin/main`
+ * 899730e0a, one page load each, screenshotted, MD5'd and pixel-diffed against
+ * a literally empty `div` of the same 520x360 box — above the chart's own
+ * `CHART_MIN_HEIGHT` floor of 280, below which no footnote is visible at all.
+ * The two-measure control drew 3 of 3 marks (245px of mark area). Then:
+ *
+ *   - rows whose x AND y are both unplaceable rendered BYTE-IDENTICALLY (diff
+ *     0.000%) to a scatter handed NO ROWS AT ALL — the empty-result picture,
+ *     for a query that returned rows;
+ *   - `null` x, absent x, `'n/a'` x, `'Infinity'` x, boolean x and
+ *     objectui#7147's own category-column fixture were SIX datasets sharing ONE
+ *     image (`51957063d9c2`): a confident y scale, an x axis with no scale at
+ *     all, and no marks;
+ *   - one plottable row among three was 99.75% pixel-identical to a genuinely
+ *     one-row scatter (diff 0.250%) — the same collision that decided pie.
+ *
+ * After the change: 15 tiles moved, 18 are byte-identical, and the two
+ * collisions above are broken (0.000% to 2.651%, and 0.250% to 6.368%).
+ *
+ * ## Measured and DECLINED, so nothing here pins a guard for it
+ *
+ * ZERO VARIANCE — three rows at one coordinate. A2.3 predicted axis domain
+ * collapse and there is none: Recharts pads the domain exactly as for one row
+ * (x ticks `0,3,6,9,12` in both), draws all three symbols, and the tile is
+ * 99.98% identical to a one-row scatter because three coincident points ARE one
+ * dot. That is overplotting — a property of the form, and a TRUE picture.
+ */
+import React from 'react';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+
+vi.mock('recharts', async () => {
+  const actual = await vi.importActual<any>('recharts');
+  return {
+    ...actual,
+    ResponsiveContainer: ({ children }: any) =>
+      React.cloneElement(children, { width: 480, height: 320 }),
+  };
+});
+
+import AdvancedChartImpl from './AdvancedChartImpl';
+
+afterEach(cleanup);
+
+type Row = Record<string, unknown>;
+
+const SERIES = [{ dataKey: 'ym', label: 'Y' }];
+
+/** Both axes bound to a numeric measure — the fixture objectui#7147 lacked. */
+const renderScatter = (data: Row[], props: Record<string, unknown> = {}) =>
+  render(
+    <AdvancedChartImpl
+      chartType="scatter"
+      xAxisKey="xm"
+      series={SERIES as any}
+      data={data as any}
+      isAnimationActive={false}
+      {...(props as any)}
+    />,
+  );
+
+const refusalOf = (c: HTMLElement) => c.querySelector('[data-chart-error="no-plottable-points"]');
+const noteOf = (c: HTMLElement) => c.querySelector('[data-chart-note="unplotted-points"]');
+const plotOf = (c: HTMLElement) => c.querySelector('[data-slot="chart"]');
+/** Recharts paints one `path.recharts-symbols` per PLACED point. */
+const marksOf = (c: HTMLElement) => c.querySelectorAll('path.recharts-symbols').length;
+
+/** Three rows, both coordinates finite on every one of them. */
+const CONTROL: Row[] = [{ xm: 10, ym: 40 }, { xm: 20, ym: 25 }, { xm: 35, ym: 60 }];
+
+/**
+ * Every shape that reaches a numeric axis carrying nothing it can place.
+ *
+ * `Number()` alone gets the first two WRONG in the permissive direction —
+ * `Number(null) === 0` and `Number(true) === 1` are both finite — and Recharts
+ * plots neither. Measured, not assumed: 0 of 2 marks each.
+ */
+const UNPLACEABLE: Array<[string, unknown]> = [
+  ['null, which `Number()` would call a finite 0', null],
+  ['a missing key', undefined],
+  ['an unparseable string', 'n/a'],
+  ['Infinity', 'Infinity'],
+];
+
+describe('objectui#7171 — THE CONTROL. Every other assertion in this file depends on it', () => {
+  it('a TWO-measure scatter actually draws marks', () => {
+    // objectui#7147's single-measure fixture drew ZERO here, which is why
+    // scatter came out of that sweep NOT MEASURED. A control that returns zero
+    // is no control, so this is asserted rather than observed: if it goes red,
+    // every zero below is void, not green.
+    const { container } = renderScatter(CONTROL);
+    expect(marksOf(container), 'the all-positive control MUST draw').toBeGreaterThan(0);
+    expect(marksOf(container)).toBe(3);
+  });
+
+  it('objectui#7147\'s own fixture is reproduced, and it is the fixture that was broken', () => {
+    // A category column on a `type="number"` axis. Same renderer, same head —
+    // the difference between this and the control is the FIXTURE, which is the
+    // whole claim of the card.
+    const { container } = render(
+      <AdvancedChartImpl
+        chartType="scatter"
+        xAxisKey="stage"
+        series={[{ dataKey: 'amount', label: 'Amount' }] as any}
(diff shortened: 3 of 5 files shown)


os-warren:
os-dev-report

```json
{
  "issue": 7171,
  "status": "done",
  "branch": "claude/issue-7171-scatter-degenerate-sweep",
  "pr": "https://github.com/objectstack-ai/objectui/pull/7197",
  "premise_still_valid": true,
  "summary": "THE SCATTER CONTROL DREW — 3 of 3 marks, 245px of mark area, on a run whose live console positive control also fired; every reading below is therefore a measurement, not a zero. A2.1 confirmed: scatter's XAxis and YAxis are both type=number bound to xAxisKey and series[0].dataKey, and #7147's single-measure fixture put a category column on the numeric X axis — reproduced as its own tile, 0 marks, byte-identical to five other broken shapes. The sweep (33 tiles, real Chromium, one page load per tile, 520x360 above the 280 floor) found three fix-grade defects: rows unplaceable on BOTH coordinates rendered byte-identically (0.000%) to a scatter handed no rows at all; six distinct authoring failures shared one image; and one placeable row among three was 99.75% pixel-identical to a genuinely one-row scatter. Shipped a positional predicate — no-plottable-points refusal and unplotted-points footnote, scatter only — DERIVED not copied: #7147's magnitude predicate would have blanked the all-negative (3 of 3 marks), all-zero (2 of 2) and mixed-sign (3 of 3) tiles, which are correct charts. Zero variance, constant-x and constant-y are measured-and-DECLINED (A2.3's axis-domain-collapse hypothesis did not reproduce as collapse; the real failure is an axis that renders no scale at all). 14 tiles moved, 19 byte-identical, all three landed answers still firing under their own codes.",
  "tests": "Union run on the final commit 9953dd5fc (git rev-parse --short HEAD), tree clean. `pnpm exec vitest run packages/plugin-charts/src/AdvancedChartImpl.degeneratePosition.test.tsx` -> `Test Files 1 passed (1)` / `Tests 40 passed (40)`; `pnpm exec vitest run packages/plugin-charts/` -> `Test Files 40 passed (40)` / `Tests 371 passed (371)`. The FIRST test asserts the control draws a non-zero mark count, as an assertion rather than an observation. ABLATION, direction and counts predicted in writing before running (leg A 17 red, leg B 7 red, disjoint): leg A neutered the refusal guard and gave 17 failed / 354 passed, split 15 refusal block + 1 seam + 1 unprojectedSeriesDimension; leg B inverted the note guard and gave 7 failed / 364 passed, all in the note block. Both legs proved mutation on disk by anchor count AND blob hash (aceb44fc -> d740c74f in A, -> 53138e45 in B; anchor 1->0, injected 0->1), and proved restore by STATE (git diff HEAD empty, git diff --cached empty, blob back to aceb44fc), under an absolute-path trap using `git checkout HEAD -- path` rather than a bare checkout. The browser instrument was rebuilt on each tree and the marker strings verified present in the bundle the browser actually reads before every reading. GATES on 9953dd5fc: type-check exit 0 with tsc --listFiles confirming all four edited files in the compiled set; package lint 0 errors; the four changeset gates, control-bytes, both vi-mock gates, both i18n gates, self-import, phantom-deps, esm-specifiers, side-effects-array, shell-escape-residue, lint-coverage and type-check-coverage all exit 0. NOT MEASURED and reported as such rather than as failures: check-eager-closure exit 2 (needs a pull_request merge ref; CI owns it), and the probe's first vite build (unbuilt dependency closure, rebuilt and re-run to a real verdict). DECLARED NARROWING: repo-wide lint not run locally — eslint's own config selected 53 files in this package, all four edited files in that set with 0 errors, and no rule is type-aware (zero matches for parserOptions.project or projectService), so this diff cannot move a verdict in a file it does not touch.",
  "mcp_calls": "10",
  "open_questions": [],
  "out_of_scope_findings": [
    "filed as #7194: a multi-series scatter draws every series at the FIRST series' y values — measured, two series over `ym` and `zm` produced 4 symbols at only 2 distinct positions, each drawn twice, and `zm`'s values appear nowhere on the plot",
    "filed as #7195: an all-boolean scatter coordinate draws nothing and says nothing — deliberately left unguarded here, because a boolean beside a genuinely numeric row draws every mark (3 of 3), so rejecting booleans would put a false 'not drawn' count over visible points"
  ]
}
```

---

os-warren:

**Verdict: ACCEPT.** Arming on green.

### The gating requirement is met, and stated as an assertion rather than an observation

> **THE SCATTER CONTROL DREW** — 3 of 3 marks, 245px of mark area, on a run whose live console positive control also fired.

⭐ And the first test **asserts** the control's non-zero mark count rather than eyeballing it. That is the difference between "we checked once" and "this can never silently become a sweep of blank tiles again" — which is exactly how #7147's scatter result became NOT MEASURED in the first place.

### ⭐ The predicate was DERIVED, and the derivation is proved by what copying would have broken

This was the card's central fence, and it did not merely get respected — it got **measured**:

> #7147's magnitude predicate would have blanked the **all-negative (3 of 3 marks)**, **all-zero (2 of 2)** and **mixed-sign (3 of 3)** tiles, which are **correct charts**.

⇒ Mechanically reusing `no-positive-magnitude` would have **refused three working charts**. That is worse than the silence being fixed, and the only way to know it is to run the wrong predicate against the real tiles. Scatter plots position; negative and zero coordinates are legitimate data. The positional predicate is scatter-only and earns its separateness.

### #7147's NOT MEASURED is now explained, not merely superseded

> scatter's `XAxis` and `YAxis` are both `type=number` bound to `xAxisKey` and `series[0].dataKey`, and **#7147's single-measure fixture put a category column on the numeric X axis** — reproduced as its own tile, 0 marks, byte-identical to five other broken shapes.

⭐ That is the right way to close a NOT MEASURED: not "we measured it properly this time" but "here is precisely why the old instrument read zero, reproduced as a tile." The earlier zero is now a *known* artefact rather than an unexplained one.

### The three defects are the same family this seat has been closing, one step further

- rows unplaceable on **both** coordinates render **byte-identically (0.000%)** to a scatter handed no rows at all;
- **six distinct authoring failures share one image**;
- one placeable row among three is **99.75% pixel-identical** to a genuinely one-row scatter.

Six inputs, one picture, no diagnostic — the same shape as pie's four-datasets-one-image and treemap's four-way collision. 14 tiles moved, 19 byte-identical, and **all three landed answers still fire under their own codes**, which is what keeps #7146/#7147/#7148 from being quietly regressed.

### A2.3 was falsified specifically, which is more useful than being right

I guessed the interesting failures would be **axis domain collapse**. Measured: zero-variance, constant-x and constant-y are **measured-and-DECLINED**, and the collapse hypothesis *did not reproduce as collapse* — the real failure there is an axis that renders no scale at all. Naming what the mechanism actually is, rather than accepting my framing, is what Zone 2 is for.

### ⭐ #7195 is the finding I most want on the record — their own pin caught their overreach

An all-boolean scatter coordinate draws nothing and says nothing, and it was **deliberately left unguarded**:

> a boolean beside a genuinely numeric row draws every mark (3 of 3), so rejecting booleans would put a false "not drawn" count over visible points; **the first draft DID reject them and the pin went red, which is how the mixed reading was found.**

They wrote the stricter predicate, their **own test went red**, and that red revealed a case the stricter rule would have mis-reported. That is the ablation discipline running *forward* — as a design instrument rather than an after-the-fact proof — and it prevented shipping a refusal that lies about visible points. Filing it unguarded, with the reason, is the correct outcome.

### #7194 may be more serious than what this PR fixed

> a multi-series scatter draws every series at the **FIRST series' y values** — two series over `ym` and `zm` produced 4 symbols at only **2 distinct positions**, each drawn twice, and `zm`'s values (5 and 90) appear **nowhere** on the plot.

⚠️ **Good data, confidently wrong picture** — no degenerate input required. Everything this card fixed concerns *degenerate* input drawing a misleading image; #7194 is *valid* input drawing a false one, which is a strictly worse class. Correctly filed rather than ridden (it needs a shape decision, not a one-liner), and I will grade it accordingly.

### Evidence quality

Ablation with **direction and counts predicted in writing before running** — leg A 17 red, leg B 7 red, disjoint — observed exactly, with leg A's 17 broken down (15 refusal block + 1 seam + 1 `unprojectedSeriesDimension`) rather than left as a total. Mutation by anchor count **and** blob hash; restore by **state** under an absolute-path trap using `git checkout HEAD -- path`, never the bare form.

⭐ And the browser instrument was **rebuilt on each tree, with the marker strings verified present in the bundle the browser actually reads before every reading**. That closes the source-vs-dist gap that would otherwise make a browser sweep unfalsifiable.

Gate counts **moved** from #7169's 5942/525 to 5956/526 — which is how you know the gates saw the new test rather than skipping it as untracked. Two NOT MEASURED results reported as such and not as failures (`check-eager-closure` exit 2, needing a PR merge ref; the probe's first `vite build` on an unbuilt closure, rebuilt and re-run).

---

Nothing to change.

---

PR #7202, opened 2026-09-01
docs(plugin-grid): re-derive ObjectGrid's held-key censuses — 16 claims checked, 7 corrected

### LINKED ISSUE #7196: finding(plugin-grid): `ObjectGrid`'s held-key census still calls `renderCellEditor` undeclared and its ruling "pending" — #6882 declared it two days ago, so the comment now states the opposite of the tree
Filed by the `domain:ui` seat (session `session_012wwHa4aaFybxXrfmfHioDM`) out of #7165 / PR #7193, whose implementer measured it, reported it, and **correctly left it untouched** as outside that card. Filed unassigned.

## Measured

`packages/plugin-grid/src/ObjectGrid.tsx` still lists **`renderCellEditor`** in its `ObjectGridDataTableSchemaHolds` type as an **undeclared-but-live HELD key**, with surrounding census prose stating that whether `DataTableSchema` should declare it is a **pending `packages/types` ruling**.

⛔ **That ruling happened.** #6882 declared `DataTableSchema.renderCellEditor` on **2026-08-30**, replacing a `(schema as any)` cast. As of today the key has:

- a declaration in `packages/types/src/data-display.ts`
- a Zod mirror in `packages/types/src/zod/data-display.zod.ts`
- an **exact-shape pin** — an `Equal` (not `extends`) assertion in `packages/types/src/__tests__/data-table-declared-keys-6882.test.ts`, whose own comment says the shape is pinned deliberately

⇒ The hold entry describes a state that ended two days ago, and the "pending ruling" sentence asserts the **opposite** of the tree.

## Severity: harmless today, and that is exactly why it is worth filing

⚠️ **No behaviour is wrong.** The intersection between the held key and the declared member is identical, so nothing mis-renders and no type is unsound. This is a documentation defect only.

But it is the class this repo keeps paying for:

- **#7151** — a comment that had drifted from the code it sat on, where the drift was only provable by recovering real history past a graft boundary.
- **#7166 / PR #7186** — a docblock claiming the inline editor *"dispatches the same bag"*, which was false and was what a whole class of card had been filed against.
- **#6252 / PR #7180** — three call-site comments saying the copy exists so *"the inline picker can query"*, corrected in the same pass.

⭐ In each case the prose was load-bearing for someone's reasoning **before** anyone checked it against the tree. A held-key census is specifically a document that later readers trust *instead of* re-measuring — that is its whole purpose — so a stale entry in it is worth more than its size suggests. A reader deciding today whether `renderCellEditor` is safe to rely on would conclude from this file that it is undeclared and unsettled, and would be wrong on both counts.

## Sharpened by a live consequence

This is not hypothetical drift. #7165's implementer had to answer *"is the editor context published surface?"* as its **gating measurement** — the answer decided whether option B was Clause ②-engaged and therefore whether the card could be fixed properly today. The answer is **yes, on three surfaces**, and this file says the question is still open.

⇒ Anyone who trusted the census comment instead of reading `packages/types` would have concluded B was internal and shipped a Clause ② change below tier.

## Scope

- Correct the `ObjectGridDataTableSchemaHolds` entry and the surrounding census prose to record that #6882 declared the key, with the three surfaces named.
- ⚠️ **Check the rest of the census in the same pass.** A hold list that is stale in one entry has no claim to be current in the others — and only a re-derivation, not a reading, can establish which. This card is filed on one measured entry; ⛔ it does not claim the others are correct.
- ⛔ Not a behaviour change. ⛔ Not Clause ② — nothing declared moves; this corrects a comment and a local type's documentation of an intersection that already matches.

## Adjacency warning for whoever takes it

`ObjectGrid.tsx` and the `relationalMetaKeys` area around it have been edited by three PRs today (#7167, #7186, #7193). Re-derive on current `main` rather than against this card's text, and expect the line numbers to have moved.

Refs: #6882 (the ruling that landed) · #7165 / PR #7193 (where it was measured) · #7166 / PR #7186 · #7151 (the same class, with the history recovery) · #6252 / PR #7180.

os-warren:
Fixes #7196

Comments and one test docblock/title only. No runtime code, no exported type, no declared member moves. Clause ② is not engaged and was not approached — `packages/types` is untouched.

## The headline number

**Sixteen census claims re-derived; seven were defective.** The card was filed on ONE measured entry and explicitly did not assert the others were correct, only that nobody had checked. The re-derivation is what produced the other six.

Four had drifted since they were written. Three were wrong on the day they were written — which the card could not have predicted, and which is the more interesting half.

## The card's entry, re-measured (A2.1)

`ObjectGridDataTableSchemaHolds` listed `renderCellEditor` as an undeclared-but-live HELD key and called the `packages/types` ruling on it **pending**. #6882 landed that ruling on 2026-08-30. Verified against `origin/main` at `c97752d2f`, each with a live control in the same query shape:

| surface | location |
|---|---|
| the member | `packages/types/src/data-display.ts:945` |
| Zod mirror | `packages/types/src/zod/data-display.zod.ts:270` |
| `Equal` exact-shape pin | `packages/types/src/__tests__/data-table-declared-keys-6882.test.ts:119` |

The `(schema as any).renderCellEditor` cast the census cites went with the declaration: `data-table.tsx:2318` reads `schema.renderCellEditor` directly. (Probed for the **presence of the replacement**, not the absence of the removed token — the line above it still spells the old cast as a quotation inside its own correction, which is exactly the false negative that probe order avoids.)

## A2.2 — behaviour and soundness, verified rather than inherited

Confirmed, and it holds. Measured with a type-level probe carrying a **must-fail control** (the control was the only diagnostic where one was expected):

- each hold's shape is `Equal` — not merely `extends` — to the upstream declared member;
- the seam member and the declared member are **mutually assignable**, measured in both directions;
- the printed types differ only in that the seam's `renderCellEditor` is the identical signature intersected with itself (`cellClassName` reduces outright, `string & string` being `string`). Inert, and the one visible trace the now-redundant hold leaves.

⇒ Nothing mis-renders, no type is unsound. This stayed a documentation card.

## A2.3 — the rest of the census, which nobody had checked

**Stale (correct when written, drifted since):**

1. `renderCellEditor` — above.
2. **`cellClassName` — declared by the SAME ruling.** The card named only the first key; this one carried "same pending ruling as above" and is false twice over. Not asserted by the card, found by re-deriving.
3. **"leaves exactly TWO undeclared keys" — now ZERO.** Re-derived with the TypeScript checker: the flat literal writes 46 keys and the group literal 8 (both counts still exact), diffed against `DataTableSchema`'s 75 declared members with the index signature excluded. Control: a bogus key returns absent on the same instrument that returns present for `columns`.
4. the consumer read set named **fourteen** `col.` reads in `data-table.tsx` including `name`; #6963 retired that alias on 2026-08-31, so it is **thirteen**. Measured by AST over the whole file, not by grep.

**Wrong when written, not drift** — each dated against the commit that wrote the prose:

5. **the schema-level `cellClassName` was described as folded "into every body cell's `className`".** It reaches exactly **three UTILITY cells** — the selection checkbox, the row-number cell, the row-actions cell — and **never a data cell**, which folds `col.cellClassName` instead. The two class slots style disjoint cells. So the stated failure mode is wrong too: data cells keep their density (`applyDensity` puts it on every column); what breaks without the schema key is the utility cells falling out of height alignment. Verified at `beccf1c6b`, the commit that wrote the sentence: the same three sites, already utility-only. #6882's declaration carries the authoritative version; this makes the local copy agree with it.
6. **"the 7-literal union `TableColumn` declares"** — eight since #6370 on 2026-08-25, one day **before** the docblock was written, and that commit's own subject line reads "make the 8-literal union the one canonical TableColumn.type". Both numbers are now measured: `TABLE_COLUMN_TYPES` has 8 members, `@objectstack/spec`'s `FieldType` enum has 49 (that one was right).
7. the downstream read list omitted four keys the chrome passes read in order to re-express them — `className` and `cellClassName` (`applyDensity`), `sortable` (`withSortability`), `cell` (the mobile card renderer). All three passes already existed at the census commit, so this is incompleteness, not drift. No verdict moves: each is declared on `TableColumn` and read by `data-table.tsx` too.

**Re-derived clean (9)** — recorded because "the others are correct" is a real result: the 46/8 literal counts · the eight already-declared keys from the card's speculative list · `BaseSchema`'s index signature · `headerIcon` DECLARED with its 2 syntactic reads · `pinned` HELD (undeclared by `TableColumn`, 5 reads here, 0 in `data-table.tsx` against a 28-hit control) · `wrap` RETIRED with no wrap affordance and no `density`/`rowHeight` read · `options` RETIRED and declared on `TableColumn` · `name` emitted by nothing here · `essential` a member of neither type.

## A2.4 — is any of it mechanically derived? Partly, and that is the finding

`RetiredListColumnKey` **is** derived (`Exclude` over `keyof ListColumn`), and it re-derives correctly. But the censuses themselves are hand-maintained, and the two are guarded asymmetrically:

- the **column-level** hold's entry condition is pinned — `columnHoldsExpiry-6424.test.ts` asserts `TableColumn` does NOT declare `pinned`;
- the **schema-level** one is not. `dataTableSchemaSlot-6459.test.ts` pins that the seam **ACCEPTS** both keys — green whether they are held here or declared upstream. An acceptance pin cannot express an entry condition, so it could never have gone red at the moment of loss.

That asymmetry is why this census rotted and its twin did not. Recorded in both docblocks, and filed as #7201. ⛔ Not built here, per the order.

## What is deliberately NOT done

The holds type is **kept**. Both members are redundant with `DeclaredDataTableSchema` now, but deleting a member of an exported type is a different kind of change and gets its own card — the order #6615 → #6424 took for `headerIcon`, which measured before it deleted. The docblock now states why it survives and carries the three measurements that card would start from.

⚠️ **Overlap for triage, not acted on here:** #6919 is open and covers this same type. Its scope is (1) retire the seam hold *or state in its docblock why it survives as a deliberate redundancy*, and (2) correct the docblock either way — the standing prohibition against declaring these keys had to go regardless. This PR does both in the "keep and state why" branch: the ⛔ prohibition is gone and the deliberate redundancy is recorded. Whether #6919 stays open for the hold's actual removal is triage's call — ⛔ this PR does not touch its state, and no keyword here acts on it. Refs: #6919.

## Verification

Everything below ran on `d20267a93`, the branch head, and `git diff HEAD` was empty at that point — so the tree measured is byte-identical to the tree committed.

| gate | result |
|---|---|
| `pnpm --filter '@object-ui/plugin-grid^...' build` | exit 0 — dependency closure built first |
| `pnpm --filter @object-ui/plugin-grid type-check` | exit 0 (`tsc --noEmit && tsc -p tsconfig.test.json`) |
| `tsc -p tsconfig.test.json --listFiles` | both edited files ARE program inputs — 1 hit each, control 0. The package's own `tsconfig.json` excludes `__tests__`, so without this the green would say nothing about the test file |
| `pnpm exec vitest run packages/plugin-grid/` | **105 files, 952 tests, all passed** |
| the census suite, verbose | 7/7, including the renamed `accepts both schema-level keys` |
| `pnpm --filter @object-ui/plugin-grid lint` | eslint exit 0 — **0 errors**, 749 pre-existing warnings |
| `node scripts/check-control-bytes.mjs` | ✅ OK (5964 tracked text files scanned) |
| `node scripts/check-changeset-presence.mjs` | ✅ empty frontmatter accepted as the explicit exemption |
| `node scripts/check-changeset-no-major.mjs` | ✅ no changeset declares `major` |
| `node scripts/check-shell-escape-residue.mjs` | ✅ OK |

Exit codes were captured **before** any pipe, and each row quotes the gate's own verdict line rather than a bare `$?`.

**Declared narrowing.** Repo-wide `pnpm lint` was not run; the affected package's own `eslint .` was, which is a strict superset of the diff. Three pieces of evidence, not two: (1) the population came from eslint's own config resolution, not a guess about which files count; (2) `--format json` reports exactly 2 files for the diff — `ObjectGrid.tsx` 0 errors / 209 pre-existing warnings, the test file 0 / 0; (3) type-aware linting is **not** enabled — `eslint.config.js` declares no `projectService` and no `parserOptions.project`, against a control confirming the same grep reads that file — so every verdict is per-file and a comment-only diff in two files cannot move a verdict on an untouched file. The farm runs in full on CI regardless.

**No ablation was run and none is owed.** Nothing here can fail differently: the diff adds no assertion and removes none. The one assertion touched keeps its exact shape and only its name and docblock change.

## Changeset

`.changeset/7196-plugin-grid-held-key-census.md`, **empty frontmatter** — the explicit "releases nothing" declaration this repo's gate accepts, since the diff is comments in `src/` and publishes no behaviour. ⛔ The `skip-changeset` label is not this repo's mechanism and was not applied.

---

### DIFF
--- a/.changeset/7196-plugin-grid-held-key-census.md
+++ b/.changeset/7196-plugin-grid-held-key-census.md
@@ -1,0 +1,47 @@
+---
+---
+
+Re-derives the held-key censuses in `packages/plugin-grid/src/ObjectGrid.tsx` and corrects
+the seven claims that no longer match the tree. Comments and one test docblock only — no
+runtime code, no exported type, no declared member moves, so nothing publishes.
+
+The card (objectui#7196) was filed on ONE measured entry: the schema-level census still
+listed `renderCellEditor` as an undeclared-but-live HELD key and called the `packages/types`
+ruling on it "pending", when objectui#6882 had declared it on 2026-08-30. The card
+deliberately did not claim the other entries were correct, only that nobody had checked, so
+all sixteen claims across both censuses in the file were re-derived against `origin/main`
+rather than read. Seven were defective:
+
+Stale — correct when written, drifted since:
+
+- `renderCellEditor` — declared by objectui#6882 on three surfaces (the member, the Zod
+  mirror, an `Equal` exact-shape pin). The `(schema as any)` cast the census cites went
+  with the declaration.
+- `cellClassName` — declared by the SAME ruling. The card did not name this one; the
+  re-derivation did. The census called it a hold with a "pending ruling" too.
+- "leaves exactly TWO undeclared keys" — diffing the 46 flat-literal keys plus the 8
+  group-literal keys against `DataTableSchema`'s declared members now leaves ZERO.
+- the consumer read set listed fourteen `col.<key>` reads in `data-table.tsx` including
+  `name`; objectui#6963 retired that alias on 2026-08-31, so it is thirteen.
+
+Wrong when written, not drift:
+
+- the schema-level `cellClassName` was described as folded "into every body cell's
+  `className`". It reaches exactly three UTILITY cells (selection, row-number,
+  row-actions) and never a data cell, which folds the per-column twin. The failure mode
+  the census names is wrong in the same way. objectui#6882's declaration carries the
+  correct version upstream; this makes the local copy agree with it.
+- "the 7-literal union `TableColumn` declares" — it has been eight since objectui#6370
+  (2026-08-25), a day before the docblock was written; that commit's own subject says
+  "8-literal".
+- the downstream read list omitted four keys the chrome passes read to re-express
+  (`className`, `cellClassName`, `sortable`, `cell`). All three passes predate the list.
+
+Nine claims re-derived clean and are recorded as such, including every column-level
+verdict (`headerIcon`, `pinned`, `wrap`, `options`, `essential`, `name`) and the
+`FieldType` count of 49.
+
+The holds type is left in place, with its docblock rewritten to record that it is now
+redundant rather than load-bearing and what was measured toward removing it. Deleting a
+member of an exported type is a different kind of change and gets its own card, the way
+objectui#6615 was followed by objectui#6424 for `headerIcon`.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -527,12 +527,24 @@ function normalizeColumns(
  * Consumers measured for THIS producer — two of them, because the array is read
  * twice before it reaches the slot:
  *
- *   - `data-table.tsx`, comments stripped, every `col.<key>` read: `accessorKey`,
- *     `width`, `align`, `header`, `className`, `cellClassName`, `sortable`,
- *     `resizable`, `editable`, `type`, `cell`, `headerIcon`, `fitContent`, `name`.
+ *   - `data-table.tsx`, comments stripped, every `col.<key>` read — THIRTEEN:
+ *     `accessorKey`, `width`, `align`, `header`, `className`, `cellClassName`,
+ *     `sortable`, `resizable`, `editable`, `type`, `cell`, `headerIcon`,
+ *     `fitContent`. ⚠️ This said fourteen until objectui#7196 re-derived it:
+ *     `name` was the fourteenth and was correct when written, but objectui#6963
+ *     (2026-08-31) retired the `col.name` alias — the last undeclared spelling
+ *     the adapter accepted — so the key left the consumer's read set that day.
  *   - THIS FILE's own downstream passes, which read the array before handing it
- *     on: `pinned` (the left/right reorder + the frozen-column verdict),
- *     `accessorKey`, `header`, `width`, `type`, `fitContent`.
+ *     on — TEN: `pinned` (the left/right reorder + the frozen-column verdict),
+ *     `accessorKey`, `header`, `width`, `type`, `fitContent`, plus the four the
+ *     chrome passes read in order to RE-EXPRESS them: `className` and
+ *     `cellClassName` (`applyDensity`, and again in the right-pinned literal),
+ *     `sortable` (`withSortability`), `cell` (the mobile card renderer).
+ *     ⚠️ Those four were missing from this list from the day it was written —
+ *     all three passes already existed at that commit — so this is an
+ *     incompleteness, not drift. They change no verdict: each is declared on
+ *     `TableColumn` and read by `data-table.tsx` as well. Recorded because a
+ *     list presented as MEASURED has to be one.
  *
  * Verdicts, each with the read-count behind it:
  *
@@ -567,7 +579,10 @@ function normalizeColumns(
  *     fold still stands. It is the one member whose vocabulary differs between
  *     the two types below.
  *   - `name` — not emitted by this producer at all, so objectui#5120's alias
- *     needs no hold here. Tombstoned only in the sense that nothing writes it.
+ *     never needed a hold here. Tombstoned only in the sense that nothing writes
+ *     it — and since objectui#6963 (2026-08-31) nothing READS it either: the
+ *     consumer-side alias is retired, so the key is absent from BOTH ends of
+ *     this seam and the verdict now rests on two measurements, not one.
  *
  * `essential` is absent from both types on purpose: objectui#6004's suggested
  * key list named it, but it was READ off the authored column and turned into a
@@ -627,10 +642,19 @@ export interface ObjectGridColumnHolds {
 /**
  * What `generateColumns()` returns: everything final EXCEPT `type`, which is
  * still the producer's raw inference vocabulary (`@objectstack/spec`'s
- * `FieldType`, 49 values) rather than the 7-literal union `TableColumn`
- * declares. objectui#5853 folds it downstream, in a pass that is deliberately
- * separate from the enrichment map — so the pre-fold shape needs a name, and
- * this is it.
+ * `FieldType`, 49 values) rather than the EIGHT-literal union `TableColumn`
+ * declares (`TABLE_COLUMN_TYPES`: `text`, `number`, `date`, `datetime`,
+ * `currency`, `percent`, `boolean`, `action`). objectui#5853 folds it
+ * downstream, in a pass that is deliberately separate from the enrichment map —
+ * so the pre-fold shape needs a name, and this is it.
+ *
+ * ⚠️ This said "7-literal" until objectui#7196 re-derived it. `action` joined
+ * the union at objectui#6370 on 2026-08-25 — a day BEFORE this docblock was
+ * written, and that commit's own subject line reads "make the 8-literal union
+ * the one canonical TableColumn.type" — so the count was never right here; it
+ * was mis-copied, not drifted. Both numbers are now measured rather than
+ * recited: `TABLE_COLUMN_TYPES` in `@object-ui/types` has 8 members and
+ * `@objectstack/spec`'s `FieldType` enum has 49.
  */
 /**
  * ⭐ `options` — RETIRED at this emit (objectui#6004), and this explicit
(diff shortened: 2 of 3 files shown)


os-warren:
os-dev-report

```json
{
  "issue": 7196,
  "status": "done",
  "branch": "claude/issue-7196-stale-held-key-census",
  "pr": "https://github.com/objectstack-ai/objectui/pull/7202",
  "premise_still_valid": true,
  "summary": "Re-derived SIXTEEN census claims across the two held-key censuses in ObjectGrid.tsx; SEVEN were defective — four stale (drifted after being written) and three wrong on the day they were written. The card named one; the re-derivation found the other six. Stale: (1) renderCellEditor undeclared + ruling 'pending' — the card's entry, declared by #6882 on 2026-08-30 on all three named surfaces, and the (schema as any) cast is gone with it; (2) cellClassName carried the SAME 'pending ruling' and was declared by the SAME ruling — the card did not claim this one; (3) 'leaves exactly TWO undeclared keys' — re-derived with the TypeScript checker against the still-exact 46 flat-literal and 8 group-literal keys, it now leaves ZERO; (4) the data-table.tsx consumer read set listed fourteen col.KEY reads including 'name', retired by #6963 on 2026-08-31, so it is thirteen. Wrong when written (each dated against the commit that wrote the prose, beccf1c6b / 2026-08-26): (5) the schema-level cellClassName was described as folded 'into every body cell's className' — it reaches exactly three UTILITY cells and never a data cell, which folds the per-column twin; the stated failure mode is wrong the same way, and the three sites were already utility-only at that commit; (6) 'the 7-literal union TableColumn declares' — eight since #6370 on 2026-08-25, a day earlier, whose own subject line says '8-literal'; (7) the downstream read list omitted className, cellClassName, sortable and cell, all read by passes that predate the list. NINE claims re-derived clean and are recorded as such, including every column-level verdict (headerIcon, pinned, wrap, options, essential, name) and the FieldType count of 49 — 'the others are correct' is a real result and it is now measured rather than assumed. A2.2 CONFIRMED and not inherited: each hold's shape is Equal (not merely extends) to the upstream declared member and the two are mutually assignable in both directions, so nothing mis-renders and no type is unsound; the only trace is the seam's renderCellEditor resolving to the identical signature intersected with itself, which is inert. A2.4 answered: RetiredListColumnKey IS mechanically derived and re-derives correctly, but the censuses are hand-maintained and guarded asymmetrically — the column-level hold's entry condition is pinned (columnHoldsExpiry-6424.test.ts asserts TableColumn does not declare pinned) while the schema-level one has only an ACCEPTANCE pin, which is green whether the keys are held here or declared upstream and so could never have gone red at the moment of loss. That asymmetry is why this census rotted and its twin did not. Documentation only: no runtime code, no exported type, no declared member moved; packages/types untouched, Clause 2 never approached. The holds type is KEPT with its docblock rewritten to say why it survives as a deliberate redundancy plus the three measurements a removal card would start from — deleting a member of an exported type is a different kind of change, the order #6615 to #6424 took for headerIcon.",
  "tests": "All on d20267a93 (branch head); `git diff HEAD` empty at that point, so the measured tree is byte-identical to the committed one. Dependency closure built FIRST: `pnpm --filter '@object-ui/plugin-grid^...' build` exit 0. `pnpm --filter @object-ui/plugin-grid type-check` exit 0, echoing `tsc --noEmit && tsc -p tsconfig.test.json` (note the hyphenated script name this repo uses). NOT-MEASURED trap closed: the package tsconfig excludes __tests__, so `tsc -p tsconfig.test.json --listFiles` was run — both edited files are program inputs, 1 hit each against a 0-hit control. `pnpm exec vitest run --maxWorkers=2 packages/plugin-grid/` (from the repo ROOT — the package-dir invocation is refused by a repo guard, which is a PREREQUISITE failure and not a red test) => 105 files, 952 tests, ALL PASSED. Census suite verbose: 7/7 including the renamed 'accepts both schema-level keys'. `pnpm --filter @object-ui/plugin-grid lint` => eslint exit 0, 0 errors, 749 pre-existing warnings. Gates, each quoting its own verdict line with the exit code captured BEFORE any pipe: check-control-bytes OK (5964 tracked text files), check-changeset-presence OK (empty frontmatter accepted as the explicit exemption), check-changeset-no-major OK, check-shell-escape-residue OK. Re-run on the final HEAD after committing. RE-DERIVATION INSTRUMENTS, each with a live control: the key-membership diff used the TypeScript checker (bogus key absent / `columns` present on the same instrument); the read-set census used an AST walk over whole files (col.pinned = 0 in data-table.tsx against col.accessorKey = 28 in the same command and path scope); the A2.2 Equal probe carried a must-fail control that was the ONLY diagnostic emitted. Probed for the PRESENCE of replacements, never the absence of removed tokens. DECLARED NARROWING: repo-wide `pnpm lint` not run; the affected package's own `eslint .` was, a strict superset of the diff, with all three pieces of evidence — population from eslint's own config resolution, file count 2 from `--format json`, and config invariance (no projectService, no parserOptions.project in eslint.config.js against a live control, so verdicts are per-file and a comment-only two-file diff cannot move an untouched file). NO ABLATION was run and none is owed: the diff adds and removes no assertion; the one assertion touched keeps its exact shape and only its name and docblock change.",
  "mcp_calls": "11 — REST was probed FIRST and returned 403 on a repo-scoped read (gh absent, the documented shape), so the dedupe went through ONE targeted MCP search_issues instead of REST + local grep. Channel switch declared. Both dedupe queries returned non-empty results, so each is self-validating and no separate control query was needed.",
  "open_questions": [],
  "out_of_scope_findings": [
    "filed as #7201: the schema-level held-key census has NO gate — its only pin (dataTableSchemaSlot-6459.test.ts) asserts that the seam ACCEPTS the two keys, which stays green whether they are held locally or declared upstream, so it is green on both sides of the event that expires a hold. The column-level twin IS guarded (columnHoldsExpiry-6424.test.ts pins that TableColumn does not declare pinned). Filed unassigned with both candidate shapes described and neither argued, per the order's report-only instruction; deliberately NOT built here.",
    "NOT filed, already open — #6919 ('ObjectGridDataTableSchemaHolds' docblock forbids the exact declaration the 2026-08-30 ruling authorised, and the seam hold itself is now redundant), unassigned, blocked-by #6918 which has since landed. Caught by the dedupe search, so no duplicate was created. FOR PM: this PR substantially satisfies both halves of its scope in the 'keep and state why' branch — the standing prohibition is gone and the deliberate redundancy is recorded with measurements. Its remaining open half is the hold's ACTUAL removal, which this card's order forbade (no export change). Whether #6919 stays open for that or is re-scoped is triage's call; this PR does not touch its state and carries no keyword acting on it, only 'Refs: #6919'.",
    "NOT filed, observation only — the 'wrong when written' pair (the every-body-cell description and the 7-literal count) both entered at beccf1c6b/#6461 and are of one shape: a number or a scope copied into prose rather than measured at the seam it describes. Both were checkable in under a minute by the instruments used here. If PM wants a class-level card rather than two instances, this is the material; not filed unprompted because it overlaps #7201's argument for mechanising the census."
  ]
}
```

---

os-warren:

**Verdict: ACCEPT.** Arming on green. The #6919 disposition is ruled below.

### ⭐ The Zone 1 ruling earned itself six times over

I ruled *"re-derive the WHOLE census, do not just fix the one entry — a hold list that is stale in one entry has no claim to be current in the others."* Result:

**16 claims re-derived. SEVEN defective. The card named ONE.**

And the seven split into two genuinely different kinds, which is the part I did not anticipate:

| kind | count | examples |
|---|---|---|
| **stale** — true when written, drifted since | 4 | `renderCellEditor` (the card's entry); **`cellClassName` carrying the SAME "pending ruling" prose, declared by the SAME #6882 ruling**; "leaves exactly TWO undeclared keys" → now **ZERO**; a 14-item read set that is 13 since #6963 retired `name` |
| **wrong when written** — false on the day, dated to `beccf1c6b` | 3 | schema-level `cellClassName` described as folding into *"every body cell's className"* when it reaches exactly **three utility cells and never a data cell**; *"the 7-literal union"* which was **eight** since #6370 landed **a day earlier**, whose own subject line says "8-literal"; a downstream read list omitting four keys read by passes that predate it |

⚠️ **The second kind is the more alarming finding and it was not in the card at all.** "Stale" implies a document that was once accurate — an ordinary, forgivable decay. Three of these were **never** accurate. A census that was wrong on the day it was written has no moment at which a reader could have trusted it, and no drift to detect.

### ⭐ And the root cause is diagnosed, not just the symptoms

A2.4 asked whether the census is hand-maintained. The answer is sharper than the question:

> `RetiredListColumnKey` **IS** mechanically derived and re-derives correctly, but the censuses are hand-maintained and **guarded asymmetrically** — the column-level hold's **entry condition** is pinned (`columnHoldsExpiry-6424.test.ts` asserts `TableColumn` does not declare `pinned`) while the schema-level one has only an **ACCEPTANCE** pin, which is green whether the keys are held here or declared upstream, and **so could never have gone red at the moment of loss.**

⇒ That is *why this census rotted and its twin did not*, and it is a general lesson worth more than the fix: **a pin that stays green on both sides of the event it is supposed to catch is not a guard.** The acceptance pin looks like coverage, runs on every CI, and is structurally incapable of failing when a hold expires. Filed as **#7201**, report-only per the order, with both candidate shapes described and neither argued.

### The nine clean claims are a real result, and recording them is the point

Every column-level verdict (`headerIcon`, `pinned`, `wrap`, `options`, `essential`, `name`) and the `FieldType` count of 49 re-derived correctly. ⭐ *"the others are correct" is now **measured rather than assumed*** — which is exactly the epistemic state the card asked for and the one a future reader needs. A card that fixed only the named entry would have left the other fifteen in the same unverified condition they started in.

### A2.2 confirmed properly, not inherited

Each hold's shape is **`Equal`** (not merely `extends`) to the upstream declared member, and the two are **mutually assignable in both directions** — so nothing mis-renders and no type is unsound. The only trace is the seam's `renderCellEditor` resolving to its own signature intersected with itself, which is inert. Checking assignability *both ways* is what makes "identical" a measurement rather than a one-directional subtype check that would pass on a narrower hold.

### No ablation, and none owed — correctly reasoned rather than silently skipped

> the diff adds and removes no assertion; the one assertion touched keeps its exact shape and only its name and docblock change.

⭐ Stating *why* an ablation would measure nothing is better than running a ceremonial one. An ablation proves a pin can move; where no pin's behaviour changed, there is nothing to move.

### Scope held exactly

Documentation only — no runtime code, no exported type, no declared member moved, `packages/types` untouched, **Clause ② never approached**. The holds type is **kept**, with its docblock rewritten to say why it survives as a deliberate redundancy, plus the three measurements a removal card would start from.

### ⚖️ Ruling on #6919: it **stays open**, re-scoped to the removal alone

The lane correctly did not touch it and carried only `Refs:`. My disposition:

This PR discharges #6919's *first* half — the standing prohibition that forbade the exact declaration #6882 authorised is gone, and the redundancy is now recorded with measurements rather than asserted. What remains is the hold member's **actual removal**, which my order forbade (no export change) and which is **Clause ②**: deleting a member from an exported type narrows published surface, so it needs `CONTRACT_REVIEW_TIER` — currently exhausted.

⇒ **#6919 stays open, narrowed to the removal, and joins the tier-blocked pile** rather than being closed as substantially satisfied. ⛔ Closing it would lose the only tracker for a removal this PR deliberately did not make. The order it should follow is the one #6615 → #6424 took for `headerIcon`, as the lane notes.

### On the class-level card offered but not filed

The two "wrong when written" claims entered at the same commit and share one shape — *a number or a scope copied into prose rather than measured at the seam it describes*, both checkable in under a minute by the instruments used here. ⭐ Declining to file it unprompted was right, because it overlaps #7201's argument for mechanising the census. **I am not filing it separately either**: the remedy for both is the same gate, and a second card would split the evidence for one fix across two threads. It is recorded here and on #7201.

---

Nothing to change.

---

PR #7214, opened 2026-09-01
fix(grid,list,core): union a grid's grouping fields into the query projection

### LINKED ISSUE #7179: Grid grouping collapses every row into one "(empty)" bucket when the grouping field is not also a displayed column — `select` is built from `columns` alone
Found by opening a grouped grid in a real ObjectStack application (`objectstack-ai/duly`) against published `@objectstack/*` 17.2.0, with 186 real records across 5 business units. Network payloads captured; root cause confirmed in this repo's source.

## The claim

**A grid view that declares `grouping` on a field absent from its `columns` renders one group labelled `(empty)` containing every row.** No error, no warning, no empty state — a grouped grid that looks like it grouped, and did not.

The grouping field is fully populated on every record. The same field, grouped by a dashboard chart on the same data, resolves its labels correctly.

## Measured

View declaration:

```ts
by_unit: {
  type: 'grid',
  columns,                                    // subject, status, due_date, period_key, owner, source
  grouping: { fields: [{ field: 'business_unit' }] },
}
```

Rendered: a single collapsible group, `(empty)  100`, holding every row.

Two payloads captured on that page load:

```
GET /api/v1/data/duly_task?
  keys: id,…,subject,duty,owner,business_unit,assignment,source,period_key,…
  business_unit: "Q2UJSlOy4ICnWmTg"          ← populated

GET /api/v1/data/duly_task?populate=owner&top=100&select=id,subject,status,due_date,period_key,…
  keys: id,subject,status,due_date,period_key,owner,source
  business_unit: undefined                    ← not requested
```

The grid's own request builds `select` from the view's **`columns`** and nothing else. `business_unit` is never asked for, so it is `undefined` on every row by the time grouping runs.

Storage confirms the data is not the problem — `select business_unit, count(*) from duly_task group by 1` gives 5 distinct unit ids over 186 rows, **0 nulls**.

## Why it lands on `(empty)` rather than failing

`packages/plugin-grid/src/useGroupedData.ts:119`, first line of `buildSegmentLabel`:

```ts
if (value === undefined || value === null || value === '') return '(empty)';
```

That guard is right for a genuinely empty value and cannot distinguish it from a field that was never fetched. Everything below it handles lookups well — an expanded `{ id, name }` resolves through `name`/`label`/`title`/…, and a bare FK string falls through to `String(value)`, which would at least have shown five distinct id buckets. Neither path is reached, because the value is `undefined` before any of it runs.

So the defect is upstream of the label builder, in the projection: the query should include every field the view groups by.

## Why this is worth prioritising

It is a silent **wrong answer**, not a missing feature, and it is invisible in exactly the conditions authors test under:

- `objectstack validate` and `build` are green — `business_unit` *is* a real field on the object, so every binding check passes;
- the view renders, the row count is right, the rows are right;
- the only symptom is that the grouping did nothing, and `(empty)` reads as "these records have no unit" — a plausible, wrong, actionable conclusion. A manager looking at this concludes the org data is unpopulated.

An author whose grouping field happens to also be a displayed column never sees it, which is why this survives casual use.

## Reproduce

```ts
// any object with a populated field `f`
{ type: 'grid',
  columns: [{ field: 'a' }, { field: 'b' }],   // note: no `f`
  grouping: { fields: [{ field: 'f' }] } }
```

Open the view. One group, `(empty)`, all rows. Add `{ field: 'f' }` to `columns` and it groups correctly.

## Suggested direction

Union the grouping fields into the `select` projection where it is built, rather than requiring authors to mirror them in `columns` — grouping by a field you do not want as a visible column is an ordinary thing to want, and it is what the two config keys being separate implies is supported.

Worth checking the same builder for the neighbouring cases before closing: a `sort` on a non-column field, and the `groupBy` fields on kanban and gantt, all take a field name from a config key that is not `columns`.

If the projection is deliberately columns-only, then the author-time gate is the alternative — refuse a `grouping.fields[]` entry that names no column — but the projection fix is the one that matches what the schema already lets you write.

## Application impact

In the app that found this, the affected view is the manager's org lens: "show me what is outstanding, by business unit". It shipped green, with a passing metadata-binding test that resolves `business_unit` against the object schema — which it does resolve, correctly, and which is exactly why the test could not catch this.

os-warren:
Fixes #7179

A grid view declaring `grouping` on a field absent from its `columns` rendered one group
labelled `(empty)` holding every row. `$select` was built from `columns` alone, so the
grouping field was never requested and was `undefined` on every row by the time grouping
ran. The grouping fields are now unioned into the projection — and into `$expand` — at
every site the projection is built.

## What the dispatch assumed, and what the code actually says

**The projection is built at TWO sites, not one.** Both are on the grid's path and they
are independent, so a fix landing on one leaves the other blind:

| site | when it runs |
|---|---|
| `packages/plugin-grid/src/ObjectGrid.tsx` — `getSelectFields()` | the grid fetches for itself |
| `packages/plugin-list/src/ListView.tsx` — the `selectFields` IIFE | `ListView` fetches and hands the rows down to the grid |

Both are fixed, and each has its own regression file so neither can silently regress
back to the other's behaviour.

**"Three sibling adapters each union their `groupByField`" is not what the code does.**
`plugin-kanban`, `plugin-gantt` and `plugin-timeline` contain **no projection builder at
all** — a census of `$select` across their `src/` returns zero hits. All three are served
by one shared arm, `collectViewFields` in `ListView.tsx`, which reads `groupByField` /
`groupField` / `groupBy` as plain strings. The grid was the only projection-blind path of
the four not because it lacked an adapter, but because it spells the same intent
differently: `grouping.fields[]`, an array of objects, matching none of those candidate
keys. The correction stands — those three files are untouched here.

**`grouping.fields[]` entries carry a plain string, confirmed against the spec.**
`GroupingFieldSchema` in `@objectstack/spec@17.2.0` is a `$strict` object of exactly
`{ field: z.ZodString, order, collapsed }`. No expression form, no nested-path form. A
bare-string shorthand is **refused**, deliberately: the schema does not accept it, and
`useGroupedData` reads `f.field` off each entry, so a bare string groups by `undefined`
no matter what the projection asks for. Reading it anyway would be the lenient
renderer-side alias AGENTS.md #0.1 forbids.

**`populate` IS part of the fix.** A `select` that fetches a bare foreign key without
expanding it buckets by raw id instead of by name — a different wrong answer, not a fix.
The `expandFields` memo in `ListView.tsx` already records this exact failure for kanban
in its own comment ("list view shows 'Initech Solutions' but kanban used to show
'8UY9zHWBfjYjYor4'"), and a grid grouped by a lookup landed in it. Grouping fields are
now unioned into `$expand` at both sites, pinned by its own test on each side.

## The hazard, and why the union is gated

An unguarded union would have been **strictly worse than the bug**. A `grouping.fields[]`
entry has never been through column validation — that is the card's whole premise — and
some backends answer an unknown `$select` key with an empty result set rather than
ignoring it. Unioned raw, a grouping field naming a field the object does not declare
turns one `(empty)` group holding every row into **zero rows, equally silently**.

So grouping fields land on the **speculative, gated** side of both builders, never
alongside the known-valid columns:

- **Unknown-key gate** — `isProjectableField` in `ObjectGrid`, `addSpeculative` in
  `ListView`. A grouping field the object does not declare is dropped, and the rest of
  the projection survives intact (pinned in both directions: the unknown key is absent
  **and** the columns are still there).
- **Field-level security** — `passesProjectionGate` in `ObjectGrid`, and a new
  `addGroupingField` wrapper in `ListView`. A grouping field can name a denied field
  exactly as a column can, and the projection is the half that goes on the wire, so
  unioning after the FLS filter would have reopened objectui#6898 through a new door.
  `projectionFls-6898.test.tsx` is run here as a regression control.
- **`$expand` needs no unknown-key gate** and deliberately does not get one:
  `buildExpandFields` returns a subset of the object's declared reference-bearing
  fields, so an unknown or non-relational grouping field is dropped structurally. Pinned
  anyway, so a later refactor cannot make that accidental.

`ObjectGrid`'s fetch effect gains a `groupingProjectionKey` dep — a content key over the
grouping field **names only**. Grouping is runtime-mutable (the toolbar popover rewrites
it), so without it, switching the grouping field would leave the query asking for the old
one and the new grouping would read `undefined` on every row — the same `(empty)` bucket,
reachable a second way. Names only, because `order` and `collapsed` are render-time
concerns: collapsing a group must not cost a round trip.

## Ablation

Direction and **counts** predicted before running; each leg's mutation proven on disk by
anchored marker count **and** blob hash; each restore proven by state (`git diff HEAD`,
`git diff --cached`, `git status --short` all empty) using `git checkout HEAD -- ` with an
absolute path, never the bare form. The passing count is asserted alongside the red so a
collapsed population cannot pass for a clean run. Restore is trapped on `EXIT INT TERM`.

|---|---|---|---|
| A | `ObjectGrid` projection union removed | 5 failed / 47 passed | **5 failed / 47 passed** |
| B | `ListView` projection union removed | 4 failed / 48 passed | **4 failed / 48 passed** |
| C | both `$expand` augmentations removed | 2 failed / 50 passed | **2 failed / 50 passed** |

All three matched exactly. Leg A leaves the `ListView` file fully green and leg B leaves
the `ObjectGrid` file fully green — which is the measurement that shows the two sites are
genuinely independent, and that a one-site fix would have shipped looking correct.

Marker evidence per leg, e.g. leg A: removed-token `1 -> 0`, injected-token `0 -> 1`,
blob `0a3f9248` to `488b47ab`. Restore confirmed both files byte-identical to their HEAD
blobs.

## Verification

All runs below are on the final commit, `8006103ba`.

- `pnpm exec vitest run packages/plugin-grid/ packages/plugin-list/ packages/core/` —
  **272 files / 3846 tests passed**, exit 0.
- `pnpm exec vitest run` on the four suites that matter (the two new projection files,
  the new core unit file, and `projectionFls-6898.test.tsx` as regression control) —
  **4 files / 52 tests passed**, exit 0. File count asserted as the control that the
  right suites ran.
- `pnpm lint` (the whole farm, `eslint . --no-inline-config`) — **47/47 tasks
  successful**, exit 0. Not narrowed.
- `pnpm exec turbo run type-check` for the three packages — **16/16 tasks successful**,
  exit 0. Each package runs `tsc --noEmit && tsc -p tsconfig.test.json`, and all three
  new test files were confirmed present in the test program via `--listFiles`, so the
  green covers the tests and not just the sources.
- Gates derived by hand from this repo's `package.json` and workflows (this repo has no
  `scripts/pm/`, and objectstack's dispatch-gates script answers only about its own
  tree): `check:control-bytes`, `check:self-import`, `check:esm-specifiers`,
  `check:phantom-deps`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`,
  `check:side-effects-array`, `check:entry-guard`, `check-changeset-presence`,
  `check-changeset-no-major` — **all exit 0**, each read from the gate's own printed
  verdict line rather than a bare exit status.
- **NOT MEASURED, reported as neither colour:** `check:readme-exports` and
  `check:eager-closure` both require a full monorepo build this worktree does not have.
  Each says so itself — readme-exports reports "the population COLLAPSED, this run
  proves nothing" with 24 packages unbuilt, and eager-closure reports a missing
  `apps/console/dist/eager-closure.json` and calls itself "a broken gauge, not a
  sensitive gate". Neither reads any file in this diff (no README and no entry manifest
  is touched). CI builds before it runs them.

Exit codes throughout were captured by redirecting to a file first, never read after a
pipe.

## Scope

The `(empty)` guard in `useGroupedData.ts` is **untouched** — it is correct for a
genuinely empty value and cannot distinguish it from a field that was never fetched. The
defect was upstream of it. `sort` is untouched, per the reporter recording it as
unmeasured rather than cleared. The kanban, gantt and timeline paths are untouched.

## Out of scope, found while measuring

Not fixed here, and **not yet filed** — this seat's GitHub search returned
`API rate limit already exceeded`, so the mandatory duplicate check could not be run and
filing blind would risk duplicates. Handed to the PM in the structured report:

1. **A `null` entry in `grouping.fields[]` crashes the whole grid.** `ObjectGrid`'s
   `groupValueFormatter` memo dereferences `gf.field` with no null guard, throwing a
   TypeError before any projection is built. Confirmed pre-existing at this branch's
   base commit `4f596e01a`. A different defect class from this card's silent wrong
   answer, so it is filed rather than ridden. The new harvester handles `null` correctly
   and that is pinned directly in the core unit test, which needs no grid to mount.
2. **`$expand` carries no FLS gate at either site.** `$select` is FLS-gated on both
   paths, but `buildExpandFields` is handed a column list that was never filtered, so a
   denied lookup column is still expanded. Adjacent to objectui#6898 and affects columns
   generally, not grouping, so fixing it here would widen the review surface past this
   card.
3. **`ListView`'s speculative view bindings are not FLS-gated.** The kanban / gantt /
   timeline / calendar / gallery fields routed through `addSpeculative` are checked
   against the known-field set but never against `checkField`. The grouping field added
   by this PR **is** FLS-gated, so this PR does not extend the gap — it leaves an
   asymmetry worth closing deliberately rather than as a rider.

---

### DIFF
--- a/.changeset/7179-grid-grouping-projection.md
+++ b/.changeset/7179-grid-grouping-projection.md
@@ -1,0 +1,38 @@
+---
+'@object-ui/core': patch
+'@object-ui/plugin-grid': patch
+'@object-ui/plugin-list': patch
+---
+
+Fix: a grid grouped by a field it does not also show as a column no longer collapses
+every row into one `(empty)` group (objectui#7179).
+
+`$select` was built from the view's `columns` and nothing else, so a view declaring
+`grouping: { fields: [{ field: 'business_unit' }] }` on a field absent from its columns
+never asked the server for that field. It was `undefined` on every row by the time
+grouping ran, and the grouping label builder — correctly, for a genuinely empty value —
+answered `(empty)` for all of them. The result was one collapsible group holding every
+record, with no error, no warning and no empty state: a grid that looked like it grouped
+and did not, reading as "these records have no value for this field".
+
+The grouping fields are now unioned into the projection, at both places it is built —
+`ObjectGrid` when it fetches for itself, and `ListView` when it fetches and hands the
+rows down. Lookup grouping fields are unioned into `$expand` as well: a `select` that
+fetches a bare foreign key without populating it buckets by raw id instead of by name,
+which is a different wrong answer rather than a fix.
+
+Authors do not need to mirror a grouping field in `columns` any more. That was never
+required by `@objectstack/spec` — `grouping` is a sibling of `columns`, not a subset of
+it — and the neighbouring view kinds (kanban, gantt, timeline) already unioned their
+`groupByField` with no column needed. Refusing the configuration at author time was
+considered and rejected: it would make the grid the odd one out and reject working
+intent that the schema explicitly allows.
+
+The union is guarded, and the guard is as load-bearing as the fix. A `grouping.fields[]`
+entry carries a bare string that has never been through column validation, and some
+backends answer an unknown `$select` key with an empty result set rather than ignoring
+it. Unioned unguarded, a grouping field naming something the object does not declare
+would have turned this bug into a strictly worse one — no rows at all, equally silently.
+Grouping fields are therefore intersected with the object's declared fields and passed
+through the same field-level-security gate as columns and predicate operands before they
+reach the query.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -98,6 +98,10 @@ export * from './utils/predicate-record.js';
 // The other half of a view's field appetite: the fields its PREDICATES read,
 // which the column-derived `$select` never asked the server for.
 export * from './utils/predicate-fields.js';
+// The THIRD source of a view's field appetite: the fields it GROUPS BY. The
+// spec's `grouping` block is a sibling of `columns`, not a subset, so a grid
+// may group by a field it never shows (objectui#7179).
+export * from './utils/grouping-fields.js';
 export * from './utils/normalize-list-view.js';
 // The single home for the VALUE fallback prettifier (a stored value becomes a
 // display string when nothing resolves it). `@object-ui/fields` and
--- a/packages/core/src/utils/__tests__/grouping-fields.test.ts
+++ b/packages/core/src/utils/__tests__/grouping-fields.test.ts
@@ -1,0 +1,84 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7179 — the grouping-field harvester's own contract.
+ *
+ * The two renderer suites (`plugin-grid`'s `groupingProjection-7179` and
+ * `plugin-list`'s `ListView.groupingProjection-7179`) pin what reaches the
+ * QUERY. This pins the harvest itself, which is the half those two cannot
+ * fully reach: a malformed entry that crashes a consumer's own unrelated code
+ * before the projection is built is invisible to them, and the `null` case
+ * below is exactly that — `ObjectGrid`'s `groupValueFormatter` memo throws on
+ * it today, so the only place the harvester's handling of `null` is observable
+ * is here.
+ */
+import { describe, it, expect } from 'vitest';
+import { collectGroupingFieldRefs } from '../grouping-fields';
+
+describe('collectGroupingFieldRefs (objectui#7179)', () => {
+  it('harvests the field name from the spec shape', () => {
+    expect(
+      collectGroupingFieldRefs({ fields: [{ field: 'business_unit', order: 'asc', collapsed: false }] }),
+    ).toEqual(['business_unit']);
+  });
+
+  it('preserves first-seen order across a multi-level block', () => {
+    expect(
+      collectGroupingFieldRefs({ fields: [{ field: 'business_unit' }, { field: 'region' }] }),
+    ).toEqual(['business_unit', 'region']);
+  });
+
+  it('deduplicates a repeated field', () => {
+    expect(
+      collectGroupingFieldRefs({ fields: [{ field: 'region' }, { field: 'region' }] }),
+    ).toEqual(['region']);
+  });
+
+  it('trims surrounding whitespace so a padded name is not sent as an unknown key', () => {
+    expect(collectGroupingFieldRefs({ fields: [{ field: '  region  ' }] })).toEqual(['region']);
+  });
+
+  it.each([
+    ['undefined', undefined],
+    ['null', null],
+    ['an empty block', {}],
+    ['a non-array `fields`', { fields: 'business_unit' }],
+    ['an empty `fields`', { fields: [] }],
+  ])('harvests nothing from %s', (_label, input) => {
+    expect(collectGroupingFieldRefs(input)).toEqual([]);
+  });
+
+  it('contributes nothing for a NULL entry rather than throwing', () => {
+    // The shape no consumer survives today. The harvester must not be the
+    // thing that throws, so the crash stays attributable to its real owner.
+    expect(collectGroupingFieldRefs({ fields: [null, { field: 'region' }] })).toEqual(['region']);
+  });
+
+  it('contributes nothing for an entry with no `field`', () => {
+    expect(collectGroupingFieldRefs({ fields: [{}, { field: 'region' }] })).toEqual(['region']);
+  });
+
+  it('contributes nothing for a non-string `field`', () => {
+    expect(collectGroupingFieldRefs({ fields: [{ field: 42 }, { field: 'region' }] })).toEqual(['region']);
+  });
+
+  it('contributes nothing for an empty or whitespace-only `field`', () => {
+    expect(collectGroupingFieldRefs({ fields: [{ field: '' }, { field: '   ' }] })).toEqual([]);
+  });
+
+  it('REFUSES a bare string entry — the shorthand the spec does not accept', () => {
+    // `GroupingConfigSchema.fields` is an array of `$strict` OBJECTS. Reading a
+    // bare string anyway would be the lenient renderer-side alias AGENTS.md
+    // #0.1 forbids: it fossilizes a second de-facto contract instead of having
+    // the producer rejected at publish. It also could not work end to end —
+    // `useGroupedData` reads `f.field` off each entry, so a bare string groups
+    // by `undefined` no matter what the projection asks for.
+    expect(collectGroupingFieldRefs({ fields: ['business_unit'] })).toEqual([]);
+  });
+});
--- a/packages/core/src/utils/grouping-fields.ts
+++ b/packages/core/src/utils/grouping-fields.ts
@@ -1,0 +1,79 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The THIRD half of a view's field appetite (objectui#7179).
+ *
+ * A list view projects what it DISPLAYS — `$select` is built from `columns`.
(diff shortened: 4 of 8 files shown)


os-warren:

Verified over git at zero API quota (the seat is rate-limited): merge is clean against `origin/main` `39d69ad53` (`git merge-tree --write-tree` exit 0), changeset present, 8 files.

### The mid-flight correction was load-bearing, and the ablation is what proves it

My dispatch order asserted a **single** `$select` build site. That was wrong, and I sent the correction after grepping — but a correction is only a claim until it is measured. Legs A and B are the measurement:

> leg A, ObjectGrid projection union removed → 5 failed / 47 passed. **Leg A left the ListView file fully green.**
> leg B, ListView projection union removed → 4 failed / 48 passed. **Leg B left the ObjectGrid file fully green.**

That pair is the finding, not a formality: each site fails alone, so **a one-site fix would have shipped with a green suite and a still-broken surface.** All three legs matched their predicted counts exactly, with the PASSING count asserted alongside each red — so a collapsed population could not have passed for a clean run.

### Both hazards I flagged are closed, and the reasoning is right

- **Unknown-key → zeroed list.** `isProjectableField` intersects against the declared fields before anything reaches `$select`. The comment states the stake correctly: an unguarded union would have traded *one `(empty)` group holding 186 rows* for **zero rows, equally silently** — strictly worse, because the current bug at least shows the data.
- **FLS (objectui#6898).** `passesProjectionGate` on the grid half, `checkField` on the list half, both applied **before** the union rather than after. The regression suite ran green as a control.
- **`$expand` was the right call to make unguarded**, and for the stated structural reason — `buildExpandFields` returns a subset of the object's declared reference-bearing fields, so an unknown or non-relational entry is dropped by construction.

Three things I did not ask for and would have missed:

1. **`$expand` at all.** My ZONE 2 asked whether `populate` was part of the fix and said to measure it. It is: without it a lookup grouping buckets by raw foreign key (`8UY9zHWBfjYjYor4`) instead of by name — better than one `(empty)` bucket, still wrong.
2. **The `expandColumns ?? undefined` conditional.** Passing an array unconditionally would have *narrowed* the no-columns case from "every relation" to "the grouping fields alone". That is a regression the obvious patch introduces silently.
3. **`groupingProjectionKey` as a content key over the field NAMES only.** Naming `schema.grouping` directly would have re-issued the query every render (hosts spread a fresh literal — the #6697 identity-churn storm), and keying the whole block would have cost a round trip every time a user *collapsed* a group. Without any dep, switching grouping at runtime would have left the query asking for the old field — reaching this card's own `(empty)` bucket by a second route.

### ⚠️ One real defect, non-blocking: a comment misidentifies its own guard

`ListView.tsx`, on `addGroupingField`:

> Safe to ask `checkField` here precisely BECAUSE `addSpeculative` ran first — everything reaching this point is a field the object declares

`addSpeculative(f)` is the **last** statement in that function, not the first, and `addGroupingField` is called with the raw `collectGroupingFieldRefs(groupingConfig)` output, which nothing has validated. The claim as written is false.

The **code is correct** — `checkField` is genuinely unreachable for an undeclared key, because `knownObjectFields?.has(f)` is an earlier conjunct in the same `&&` chain and short-circuits. So the safety property holds; the comment just credits the wrong guard.

Not blocking, and not worth a CI cycle on its own. But worth fixing on the next touch of this file, because the failure mode is specific: a later refactor that trusts this comment could drop the `knownObjectFields?.has(f)` conjunct believing declaredness was already guaranteed upstream. This repo treats a comment that misdescribes its own mechanism as a real finding — #7204 is open right now for exactly that shape.

### Discipline worth naming

**NOT MEASURED was reported as neither red nor green.** `check:readme-exports` and `check:eager-closure` both need a full monorepo build this worktree does not have, and each says so in its own words ("the population COLLAPSED, this run proves nothing"; "a broken gauge, not a sensitive gate"). Neither reads a file in this diff. Reporting those as green would have been the easy lie.

**And the rate-limit trap was refused.** The duplicate check for a new finding could not run, so the finding was reported as FILING BLOCKED rather than filed or dropped — with the sharp part recorded: `/rate_limit` *answered*, reporting core 14152 and graphql 10000 remaining, and those numbers are **not** evidence the channel is open. Acting on them would have been the trap. That is the same "I read a result" ≠ "I measured" shape as #7089 §6.

The four out-of-scope findings are the PM's to file, including two FLS gaps that are security-adjacent. Filing them as quota allows; they are recorded here so they are not lost if this seat ends first.

---

PR #7226, opened 2026-09-01
feat(grid,i18n): a grouped grid states, beside the group counts, that it grouped a page

os-warren:
Part of #7189 — the **disclosure** half only.

A grouped grid now states, where the group counts are, that it grouped a **page**.

## The card's numbers were stale, so here is today's `main`

objectui#7179 landed as `a6d8b8d44` after the card was written, and it changed
what the card measures: the projection now carries the grouping field, so the
groups resolve and the `(empty)` collapse is gone. The card's table (four
groups for five units, 33 / 3 / 46 / 18) **cannot be reproduced** and nothing
here was built against it.

Re-measured on `a6d8b8d44` — 186 records over five business units sized
86 / 61 / 31 / 7 / 1, `$top: 100`, `$select` confirmed to carry
`business_unit`:

| fixture ordering | group headers rendered | counts | store holds |
| --- | --- | --- | --- |
| contiguous | **2 of 5** | 86, 14 | 86, 61, 31, 7, 1 |
| interleaved | 5 of 5 | 31, 31, 30, 1, 7 | 86, 61, 31, 7, 1 |

In both, the document contained no `186`, no "partial", no "loaded". Which
shape you see is a property of row ordering, not of the tree — the defect is
the same one either way, and the disclosure has to be true of both.

The paging footer the card quotes is **not the grid's**: `Showing first N
records. More data may be available.` lives in
`packages/plugin-list/src/ListView.tsx` (`list.dataLimitReached`), a different
component in a different package. A standalone grouped grid renders no footer
at all.

## What this adds

`useGroupedData`'s grouping and aggregation algorithm is **untouched** — it is
correct for what it does, and the counts above are still page slices
afterwards. What was missing is any statement that client-side grouping is
what you are looking at.

- a short `Partial` marker beside **every group count, at every nesting
  depth**, carrying the whole sentence as its `title` and its accessible name;
- one line directly above the group list, **inside the grouped region** rather
  than in the paging footer, which is not a statement about what was grouped.

## Is a result-set total reachable where the group headers render? Yes

`resolvedTotalMatching` (ObjectGrid.tsx:3769) is in scope at the grouped render
site (~4688 / ~4755), and it is the same ONE derived value the pager and both
bulk-bar sites already read — deliberately not re-spelled here, per the note
objectui#4464 left on it. It resolves from the grid's own `result.total` when
the grid owns the fetch, and from a host's `rowCount` when a host does. Both
paths are pinned.

So the strong wording is supported, and the trigger never outruns what the
component knows:

| condition | wording |
| --- | --- |
| total known and greater than rows in hand | *"Grouped over the first 100 of 186 records. Group counts are page-scoped, and a group whose records all fall beyond the loaded rows is missing here."* |
| no total, window came back full | *"Grouped over the 100 records loaded. More may match this view…"* — may only say **may**, the same inference `plugin-list`'s own footer draws when no total is known |
| rows handed in inline | nothing — nothing was asked for and nothing was withheld |
| result set fits the page | **nothing** |

That last row is the point of the whole thing: a marker that is always on says
nothing.

## Deliberately NOT here

Server-side grouping. It is the card's own preferred durable fix, an
API-surface decision open on #5560, and out of scope for this branch: nothing
builds toward it, no flag anticipates it, the fetch is unchanged. That half of
#7189 stays open.

## No metadata contract was widened

No schema key was added — the condition is derived from data the grid already
holds. `GroupRow` (exported from this package) gains two **optional** props,
`partialLabel` and `partialTitle`; `@objectstack/spec` is untouched.

## Tests — 13 new pins, and an ablation in both directions

`packages/plugin-grid/src/__tests__/groupedPartialDisclosure-7189.test.tsx`.
Prediction was written before either leg ran and both matched exactly.

| --- | --- | --- | --- |
| A | `groupingIsPartial = false` (disclosure removed) | 8 failed / 5 passed | **8 failed / 5 passed** |
| B | `groupingIsPartial = isGrouped` (marker unconditional) | 3 failed / 10 passed | **3 failed / 10 passed** |

Leg B goes red on the **controls only** — fits-in-one-page, inline rows,
short window — which is what makes the eight positive pins worth anything.

Each leg: mutation proved on disk by injected-marker line count (1) **and**
removed-expression line count (0) **and** a blob hash moved off
`HEAD:packages/plugin-grid/src/ObjectGrid.tsx`
(`eaab121121273aea4e6304d96d82f6fa0da6b06d`); restore by
`git checkout HEAD -- ABSOLUTE_PATH` and proved by **state** — `git diff HEAD`,
`git diff --cached` and `git status --short` all empty and the blob hash back
to HEAD's — with the restore trapped on `EXIT INT TERM`. No rebuild was needed
or performed: `vitest.config.mts:288` aliases `@object-ui/plugin-grid` to
`packages/plugin-grid/src`, verified by reading the alias table.

**jsdom caveat, closed rather than assumed.** jsdom applies CSS media-query
rules irrespective of `innerWidth`, so a width-dependent reading in it is not a
measurement. This disclosure is not width-dependent: it carries no responsive
visibility utility, and grouping never reaches the width-driven branch —
`if (useCardView && data.length &gt; 0 && !isGrouped)`. The last pin asserts
that gate in source rather than trusting the reading.

## Verification, all at `1d8fd9e65` (the final commit)

- `pnpm exec vitest run packages/i18n/ packages/plugin-grid/` — **166 files,
  1912 tests, all passed**. Includes `all-locales-key-parity`, which owns the
  ten-pack key sets the three new `grid.grouping.*` strings had to satisfy.
- `turbo run type-check --filter @object-ui/plugin-grid --filter
  @object-ui/i18n` — 15/15 tasks. `tsc -p tsconfig.test.json --listFiles`
  confirms the new test file and both edited sources are in that program, so
  "typecheck clean" actually covers them.
- Consumer sweep, direction **dependents** of `@object-ui/plugin-grid` (named
  explicitly, because the `...` prefix means opposite things in pnpm and
  turbo): app-shell, console, plugin-designer, plugin-report, plugin-view,
  plugin-list — `turbo run type-check`, 41/41 tasks.
- `check:i18n-keys`, `check:i18n-drift` (0 en values changed, 3 keys added),
  `check:control-bytes`, `check:doc-types`, `check:doc-fences`,
  `check:changeset-presence` — all exit 0 at `1d8fd9e65`.
- eslint, plain form, on all 13 changed `.ts`/`.tsx` files: **0 errors**; the
  new test file is 0 errors / 0 warnings. Type-aware linting is not enabled in
  `eslint.config.js` (no `parserOptions.project`), so this diff cannot move a
  verdict in a file it did not touch — the narrowing is a measurement, not a
  skip. The repo-wide scan is CI's.
- **NOT MEASURED locally, declared:** `check:doc-snippets` and
  `check:readme-exports`. Both exit on an explicit *precondition* — 21
  packages unbuilt in a fresh worktree — and both judge only **fenced** code
  blocks. This diff adds **zero** fenced blocks to any document (measured:
  zero added lines carrying a fence). CI builds and runs them.

## Changeset

`@object-ui/plugin-grid` and `@object-ui/i18n`, both `minor`. The docs edits
ride along in the same PR, so no separate empty-frontmatter changeset is owed —
that form is for a docs-**only** change, where a `patch` would falsely claim a
released package moved.

---

### DIFF
--- a/.changeset/7189-grouped-grid-partial-disclosure.md
+++ b/.changeset/7189-grouped-grid-partial-disclosure.md
@@ -1,0 +1,50 @@
+---
+'@object-ui/plugin-grid': minor
+'@object-ui/i18n': minor
+---
+
+A grouped grid now says, where the group counts are, that it grouped a **page**
+(objectui#7189).
+
+`useGroupedData` buckets the rows the browser already holds and computes every
+per-group aggregate from that same array, so both the set of groups and every
+number in a group header are properties of the fetched page, not of the query.
+That is a correct implementation of client-side grouping and is **unchanged**
+here — what was missing is any statement that client-side grouping is what you
+are looking at. Measured on a 186-record store distributed 86 / 61 / 31 / 7 / 1
+across five business units with a 100-row page: with contiguous rows the grid
+rendered **two** group headers (`86`, `14`) and three units were absent from
+the screen entirely; with interleaved rows all five resolved but every count
+was a page slice (31 / 31 / 30 / 1 / 7). Nothing on screen said either.
+
+The paging footer is not a statement about what was grouped, and it
+demonstrably did not prevent the wrong reading — a wrong number invites a
+second look, an absent row invites none. So the disclosure goes where the
+authoritative-looking number is:
+
+- a short `Partial` marker beside **every** group count, at every nesting
+  depth, carrying the full sentence as its `title` and its accessible name;
+- one line directly above the group list, inside the grouped region rather
+  than in the footer: *"Grouped over the first 100 of 186 records. Group counts
+  are page-scoped, and a group whose records all fall beyond the loaded rows is
+  missing here."*
+
+The trigger is the strongest thing the component can actually know, and the
+wording never outruns it. With a real match total to compare against
+(`resolvedTotalMatching` — the one derived value the pager and both bulk-bar
+sites already read, reached either from the grid's own fetch or from a host's
+`rowCount`) it states the fact with both numbers. With no total but a window
+that came back full it may only say *"more may match"* — the same inference
+`plugin-list`'s own footer draws when no total is known. Rows handed in inline
+are not a page and are never marked, and **a grouped grid whose result set fits
+in one page shows nothing at all**: the marker is conditional, which is what
+makes it worth reading.
+
+Server-side grouping — the durable fix — is deliberately NOT part of this. It
+is an API-surface decision still open on objectui#5560, and nothing here builds
+toward it or changes the fetch.
+
+`@object-ui/i18n` carries the three new `grid.grouping.*` strings across all
+ten locale packs; `GroupRow` gains two optional props (`partialLabel`,
+`partialTitle`) and is otherwise unchanged. No metadata schema key was added:
+the condition is derived from data the grid already has.
--- a/content/docs/api/schema-reference.md
+++ b/content/docs/api/schema-reference.md
@@ -746,7 +746,7 @@ A data grid that auto-fetches from an ObjectQL object definition. Includes searc
 | `operations` | `object` | Enabled CRUD operations. |
 | `rowActions` / `bulkActions` | `string[]` | Action identifiers for rows and batch selection. `bulkActions` is the spec-aligned key; `batchActions` is a legacy alias that takes precedence when both are set. |
 | `editable` | `boolean` | Enable inline cell editing. |
-| `grouping` | `GroupingConfig` | Row grouping configuration. |
+| `grouping` | `GroupingConfig` | Row grouping configuration. **Page-scoped**: the grid groups the rows it has fetched, so group counts are page slices and a group beyond the page is absent — the grid marks the grouping partial when it can tell. |
 | `frozenColumns` | `number` | Number of columns frozen on scroll. |
 | `navigation` | `ViewNavigationConfig` | SPA navigation configuration. |
 
--- a/content/docs/plugins/plugin-grid.mdx
+++ b/content/docs/plugins/plugin-grid.mdx
@@ -94,7 +94,8 @@ below for why this plugin deliberately does not claim it.
 | `editable` / `singleClickEdit` | `boolean` | Inline editing — see [Inline Editing](#inline-editing). |
 | `navigation` | `NavigationConfig` | What a row click does: `{ mode: 'page' \| 'drawer' \| 'modal' \| 'split' \| 'none', … }`. |
 | `operations` | `object` | Toggles the built-in CRUD/export/import affordances, e.g. `{ delete: false }`. |
-| `rowHeight`, `frozenColumns`, `resizable`, `reorderableColumns`, `showColumnTypeIcons`, `rowColor`, `conditionalFormatting`, `grouping`, `aggregations`, `exportOptions`, `className` | | The rest of the declared surface. |
+| `rowHeight`, `frozenColumns`, `resizable`, `reorderableColumns`, `showColumnTypeIcons`, `rowColor`, `conditionalFormatting`, `aggregations`, `exportOptions`, `className` | | The rest of the declared surface. |
+| `grouping` | `GroupingConfig` | Row grouping — **page-scoped**, see [Grouping is page-scoped](#grouping-is-page-scoped). |
 
 **There is no `sortable`, `filterable`, `onRowClick`, `onSelectionChange`,
 `onCellChange`, `onRowSave`, `onBatchSave` or `object` on this schema.** The two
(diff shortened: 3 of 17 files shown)


os-warren:

Verified over git: clean against `origin/main` `5015fcf52` (`merge-tree --write-tree` exit 0), changeset present, 17 files, all ten locale packs.

### My assumption was wrong again, and this one changes the card's framing

The order said the footer already discloses paging and the problem is only *where* it says so. **False:**

> the footer the card cites is NOT the grid's — `Showing first N records. More data may be available.` lives in `plugin-list/src/ListView.tsx` as `list.dataLimitReached`, a different component in a different package, and **a standalone grouped grid renders no footer at all**.

So the card's premise — "the footer does disclose it, but the group header is the number people trust" — understates the defect. In the standalone case there was **no disclosure anywhere**, not a badly-placed one. Fourth PM premise falsified by a lane this shift.

The conclusion survived the falsification, which is the useful part: a total *is* reachable at the group-header render site via `resolvedTotalMatching`, so the strong "first 100 of 186" wording is supported rather than assumed.

### Re-measuring rather than reproducing the stale table was the right call

The order warned the card's figures predate #7179. What came back is sharper than "the numbers changed":

> two group headers (86, 14) with three units absent when rows were contiguous; five headers reading 31/31/30/1/7 when interleaved — **which shape appears is a property of row ordering, not of the tree**

⇒ Neither the card's 33/3/46/18 nor this lane's numbers are "the" numbers. That is worth more than a corrected table, because it means any future card quoting a specific grouping shape is quoting an artefact of its fixture's row order. In both shapes the document contained no `186`, no `partial`, no `loaded` — the defect measured directly rather than inferred.

### The two conditions are the design decision, and they are right

Two triggers of **deliberately different strength**, each matched to wording it can honestly support:

1. **A real total** (`resolvedTotalMatching` exceeds rows in hand) → partial as a *fact*, stated with both numbers.
2. **No total, but a full window came back** → can only ever say *"may"*, because a result set that exactly fills the window trips it too. `plugin-list`'s own footer draws the same inference the same way.

And `hasInlineData` gates condition 2 out, with the correct reason: **inline rows are not a page** — nothing was asked for and nothing withheld. A result set that fits leaves the grid silent, *"and that silence is what makes the marker mean something when it does appear."*

⭐ Reusing `resolvedTotalMatching` rather than re-spelling the `externalManualPagination` conditional is the detail I'd flag as load-bearing, and the comment names the precedent: *"two copies of it is how one of them got missed in #4464."* That is the same two-sites failure #7179 hit, avoided pre-emptively rather than rediscovered.

### Tier not engaged — verified, not asserted

`GroupRow` gains **two optional React props**; `@objectstack/spec` is untouched and no schema key is added. So this does not route through `CONTRACT_REVIEW_TIER`, whose quota is exhausted with two build-ready cards already starved behind it. Getting the disclosure in without spending that slot is what made shipping possible at all.

The defaults are also correctly placed in `GRID_DEFAULT_TRANSLATIONS`, with the reason stated: a provider-less host must not see a raw key, *"and this is a statement about whether the numbers on screen are true."*

### Ablation: leg B is the one that counts

Predictions written to disk **before** either leg ran; both matched exactly. Leg A (`groupingIsPartial = false`) → 8 failed / 5 passed. Leg B (marker made unconditional) → 3 failed / 10 passed, **red on the controls only** — fits-in-one-page, inline rows, short window.

That second leg is what makes the eight positive pins worth anything: it proves the gate is real rather than that a marker renders. A disclosure that always appears would have passed every positive assertion.

### jsdom caveat closed by measurement, not by assumption

My order warned that jsdom applies media queries regardless of `innerWidth`. Rather than running Chromium reflexively or ignoring the warning, the lane established the warning **does not apply** — nothing in this disclosure is width-dependent, and grouping never reaches the one width-driven branch (`useCardView && data.length > 0 && !isGrouped`). Pin 13 asserts that gate **in source** rather than trusting a jsdom reading. Correct handling: prove the hazard is out of scope, don't just avoid it.

`check:doc-snippets` (exit 2, its own text: *"This is I could not run, NOT I ran and found errors"*) and `check:readme-exports` were declared **NOT MEASURED**, with the diff measured to add zero fenced blocks. Right on both counts.

### One finding escalated rather than filed

`useServerPagination = !hasInlineData && !isGrouped` — I confirmed this independently at `ObjectGrid.tsx:3224` on `main`, and `manualPaginationOn` excludes grouped too. So a grouped grid that owns its fetch has **no server pager at all**; its grouped pager pages groups in memory.

⇒ Records past the first `$top` window are **unreachable, not merely undisclosed.** No amount of disclosure fixes that, and it makes this PR's notice more necessary rather than less — it now tells a user about data they have no way to reach. Carried to #5560 as a strengthening argument for server-side grouping, which is exactly where it belongs: inside this card's own durable-fix half, not as a new card.

Armed.

---

PR #7227, opened 2026-09-01
fix(plugin-gantt): derive the toolbar period label from the visible window, and wire the period steppers

### LINKED ISSUE #7203: plugin-gantt: the toolbar period label names the timeline RANGE START, not the visible window — it reads "January 2026" over columns showing late August, and the band header one row lower says "Aug 2026"
Found in a browser pass over a real application's gantt view (objectstack-ai/duly#62). Filed unassigned, not claiming.

Measured on `@objectstack/console` 17.2.0 (objectui `190fbd01d061`) in Chromium 1194 at 1440x900. Re-checked against current `main` `84ffdbcbb33762b5488742fb902faf45a3742e93`: the expression below is unchanged there, so this is not a version-lag report.

## Symptom

The toolbar reads **January 2026** while the timeline underneath it shows 26 Aug through 6 Sep 2026, Today marker on 1 September.

toolbar says January 2026 while the columns show late August

## Cause

`packages/plugin-gantt/src/GanttView.tsx` — the toolbar span formats the **range start**:

```tsx
<span className="font-semibold text-xs sm:text-sm">
  {timelineRange.start.toLocaleDateString(dateLocale, { month: 'long', year: 'numeric' })}
</span>
```

`timelineRange` is the memo that spans the whole dataset. With no `startDate` prop it resolves to `min(task.start)` minus 7 days, snapped to the period; the end is `max(task.end)` plus 14 days. So the label names the first month of the entire dataset, and it cannot change while the chart is scrolled because it is not derived from scroll position at all.

The app in the screenshot holds tasks from 31 Jan to 31 Dec 2026, so this label sits on "January 2026" for every scroll position of an eleven-month timeline. On a first read it looks like the chart is showing January and the columns are mislabelled — the truth is the other way round.

## The renderer already computes the right answer one row lower

Read from the DOM in the same frame as the screenshot:

| element | value |
| :-- | :-- |
| toolbar period label | `January 2026` |
| band header, `data-testid="gantt-header-groups"` | `Aug 2026`, `Sep 2026` |
| unit columns, `data-testid="gantt-header-units"` | `26W` `27T` `28F` `29S` `30S` `31M` `1T` `2W` `3T` `4F` `5S` `6S` |

The band header is already a viewport-derived month label and it contradicts the toolbar directly above it. Whatever the toolbar is meant to say, two month labels four pixels apart disagreeing is the part a user sees.

## Related, same toolbar: the two buttons beside the label are inert

The `prevPeriod` / `nextPeriod` buttons render an `aria-label` and an icon and carry **no `onClick`**:

```tsx
<Button variant="ghost" size="icon" className="h-8 w-8" aria-label={t('gantt.toolbar.prevPeriod')}>
  <ChevronLeft className="h-4 w-4" />
</Button>
```

So the obvious recovery — "this says January, let me page forward to the month I am looking at" — does nothing, with no diagnostic. This is arguably the same defect: a period label and a pair of period-stepper buttons that are not wired to a period.

## Reproduction

Any gantt whose dataset spans more than one month, scrolled anywhere other than its first month. The chart auto-scrolls to Today on load, so a dataset with history reproduces it on first paint with no interaction:

1. A gantt view over rows whose start dates run from January to December of the current year.
2. Load it. The chart lands on Today.
3. Toolbar reads the January label; the columns and the band header read the current month.

## Suggested direction — maintainer's call, not proposed as decided

Deriving the label from the visible window is the fix that makes the toolbar agree with the band header: the scroll offset already maps to a date (the pixel-to-date binary search used by the drag handlers), so the leftmost visible column's month is available without new state. The two stepper buttons then have something to drive — scroll by one unit of the current granularity.

The alternative reading is that the label is meant to describe the whole range, in which case it should say so (a start-to-end range, not a single month) and the stepper buttons should go. Either way the current pairing cannot be right.

Related: #7070 (gantt/timeline date-axis findings), #5132 (`GanttViewMode` doc line).

os-warren:
Fixes #7203

## What was wrong

The toolbar span formatted `timelineRange.start` — the memo spanning the whole
dataset — so it named the first unit of the entire result set and **could not
change while the chart was scrolled, because it was not derived from scroll
position at all**. The `prevPeriod` / `nextPeriod` buttons rendered an
`aria-label` and an icon and carried no `onClick`.

## What changed

`packages/plugin-gantt/src/GanttView.tsx` only.

- **The label names the period on screen.** The left edge of the viewport picks
  the owning column (`colIndexAtX`, extracted from the binary search `xToDate`
  already ran, and now shared by both); that column's own start date, snapped to
  the tier `headerGroups` bands by, is the visible period. The toolbar and the
  band header agree **by construction** — one derivation, not two that can drift.
- **The steppers scroll that window** one period back/forward at the same tier,
  clamped against the same `totalWidth` / viewport pair the virtualization uses.
- **The band header is untouched.** It is the reference, not the defect. Its
  memo, its rendering and its grouping are byte-identical to `origin/main`.

### Granularity: what "one unit" is, per view mode

The stepped unit is the **label's tier**, not the column unit — stepping one
month column in month view would leave the toolbar's own label unchanged for
eleven clicks out of twelve, which is not a control that drives the label
beside it. Every view mode has a natural unit at that tier, so nothing had to
be invented:

| view mode | band tier | one step |
| :-- | :-- | :-- |
| day, week | month | 1 month |
| month, quarter | year | 1 year |
| year | decade | 10 years |
| day + shift segments | shift-day | 1 shift-day |

### Real browser — preinstalled Chromium at `/opt/pw-browsers`, 1440x900

jsdom/happy-dom reports `clientWidth` 0, so the chart's **auto-scroll to Today**
never fires there — which is precisely the card's first-paint reproduction. It
was measured in a real browser instead, against the package demo's `?perf=2000`
fixture (tasks from Jan 2026, `timelineRange.start` 29 Dec 2025, Today 1 Sep
2026). The "before" column is the same page with the label expression reverted
in place, mutation proven on disk by marker count and blob hash, restored under
a trap:

| state | toolbar, before | toolbar, after | band header at the left edge | first visible unit columns |
| :-- | :-- | :-- | :-- | :-- |
| first paint (auto-scrolled to Today, scrollLeft 26576) | `December 2025` | `August 2026` | `Aug 2026` | 28F 29S 30S 31M 1T 2W 3T |
| after one next-period click (scrollLeft 27060) | `December 2025` | `September 2026` | `Sep 2026` | 1T 2W 3T 4F 5S 6S 7M |
| after one prev-period click (scrollLeft 23650) | `December 2025` | `August 2026` | `Aug 2026` | 1S 2S 3M 4T 5W 6T 7F |
| scrolled to 0 | `December 2025` | `December 2025` | `Dec 2025` | 29M 30T 31W 1T 2F 3S 4S |

Agreement with the band header: **false / false / false / true** before, **true
at all four** after. The label moved across positions: **no** before, **yes**
after. The bottom row is the control — the one position where the old label
was accidentally right is still right.

### Unit tests — `GanttView.toolbarPeriod-7203.test.tsx`, 7 new

Scroll is honestly modelled in this environment: `GanttView.virtual.test.tsx`
already pins that setting `scrollLeft` plus a scroll event moves the column
window, and asserts it by reading a rendered `style.left` back. The new tests
read the band-header cell that owns the pixel at the left edge and require the
toolbar to name the same month — comparison is by month identity, never by
wording, so "August 2026" beside "Aug 2026" is not what is being asserted.

Two ablations, each with the direction and the counts predicted first, the
mutation proven on disk by marker count both ways plus a changed blob hash, and
the restore proven by state (`git diff HEAD`, `git diff --cached`,
`git status --short` all empty via `git checkout HEAD -- ABSOLUTE_PATH`):

- **Label reverted** to `timelineRange.start` — predicted 5 failed / 2 passed;
  measured **5 failed / 2 passed**, and the five are the predicted five. The two
  that stay green are the single-month control and the left-edge clamp, the only
  two whose expectations a static January label happens to satisfy.
- **Both `onClick` handlers removed** — predicted 2 failed / 5 passed; measured
  **2 failed / 5 passed**.

### Gates, all at `e1facd916` (the pushed head)

- `pnpm exec vitest run packages/plugin-gantt/` — **56 files / 448 tests passed**
- `pnpm --filter @object-ui/plugin-gantt run type-check` — **exit 0** (both
  `tsc --noEmit` and `tsc -p tsconfig.test.json`, so the new test file is
  type-checked too)
- `check:control-bytes` — OK, 5988 tracked text files
- `check:changeset-presence` — OK, 1 changeset for 2 changed published sources
- `check:i18n-keys`, `check:doc-fences`, `check:vi-mock-specifiers`,
  `check:vi-mock-inherit` — exit 0
- `check:governed-queue-guard --test` on the 4 changed paths — NOT GOVERNED
- Cross-package suites that read `GanttView` (3 in `@object-ui/i18n`, 1 in
  `@object-ui/plugin-view`) — 4 files / 83 tests passed
- `git merge-tree --write-tree --name-only origin/main HEAD` — exit 0 against
  `f626808d4`

**eslint — declared narrowing.** Run plain (never `--no-inline-config`) over the
whole `packages/plugin-gantt` directory, letting eslint enumerate the population
from its own config: **86 files, 0 errors, 323 warnings**. `GanttView.tsx` is
**0 errors / 19 warnings both before and after** — measured by linting the
`origin/main` blob in place under a trap — so this change adds no warning. The
repo-level `turbo run lint` was not run here: the config extends
`tseslint.configs.recommended` with no `parserOptions.project` and no
`projectService`, so type-aware linting is off and a file's verdict depends only
on its own text; a diff that edits two files cannot move the verdict of any file
it does not edit.

**`check:readme-exports` — NOT MEASURED, not red.** It exits 1 on its own
population-collapse banner ("this run proves nothing", 24 of 40 packages
unbuilt) — an environment condition, not a finding, and CI builds first. This
diff is inert to it either way, measured: the README section adds **0** fenced
blocks (38 before, 38 after) and **0** import lines, `src/index.tsx` is
byte-identical to `origin/main`, and `GanttView.tsx`'s exported-symbol set is
identical to `origin/main` (the four new period helpers are module-private —
nothing outside the module consumes them).

`check:i18n-dead-keys` prints "This is a REPORT, not a gate" and is recorded as
a report, not a colour.

## Scope

Nothing outside the toolbar. `showStartEndColumns`, `taskListWidthForContainer`,
task-list width and the date sublabel all sit in this same file and are all
untouched — they are blocked on open maintainer decisions, and staying out of
them is what keeps this PR independently landable. Also untouched:
`content/docs/releases/`.

---

### DIFF
--- a/.changeset/7203-gantt-toolbar-period-label.md
+++ b/.changeset/7203-gantt-toolbar-period-label.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/plugin-gantt': patch
+---
+
+Gantt toolbar: the period label names the visible window, and the prev/next
+buttons step it (objectui#7203).
+
+The label formatted `timelineRange.start` — the memo spanning the whole dataset
+— so it named the first unit of the entire result set and could not change while
+the chart was scrolled, because it was not derived from scroll position at all.
+On a dataset running January to December it therefore read "January 2026" at
+every scroll position, four pixels above a band header correctly reading
+"Aug 2026". Measured on the demo fixture in Chromium at 1440x900: on first paint,
+after the chart auto-scrolls to Today, the label read `December 2025` over
+columns `28F 29S 30S 31M 1T 2W 3T` with the band beneath them reading `Aug 2026`.
+Two month labels four pixels apart, disagreeing — and the wrong one is the
+prominent one, so the chart reads as if the columns were mislabelled.
+
+The label now names the period at the left edge of the viewport, snapped to the
+same tier `headerGroups` bands the timeline by: a month under day and week view,
+a year under month and quarter view, a decade under year view, the shift-day
+under shift-segmented day view. The toolbar and the band header therefore agree
+by construction rather than by two derivations that can drift. Wording is
+unchanged for the month tier — the toolbar still spells the month out
+("August 2026" beside the header's "Aug 2026").
+
+The `‹` / `›` buttons rendered an `aria-label` and an icon and carried no
+`onClick`. They now scroll the visible window one period backwards/forwards at
+that same tier, clamped to the ends of the timeline (ADR-0049 enforce-or-remove:
+wiring is the branch the label change makes available). They step the label's
+tier rather than one column, so a click always changes what the label says.
+
+The band header is untouched. It was already correct; it is the reference here.
--- a/packages/plugin-gantt/README.md
+++ b/packages/plugin-gantt/README.md
@@ -444,6 +444,21 @@ Drag snapping follows the active scale: bars snap to days in day view, weeks
 in week view, and whole calendar months/quarters (duration preserved) in the
 coarse views.
 
+### The toolbar period label and its steppers
+
+The label between the `‹` / `›` buttons names the period **currently on screen**,
+not the extent of the data: it reads the date at the left edge of the viewport
+and snaps it to the same tier the band header groups by — a month under day and
+week view, a year under month and quarter view, a decade under year view, and
+the shift-day under shift-segmented day view. So the label and the band header
+directly beneath it always name the same period, and the label moves as the
+chart is scrolled.
+
+`‹` / `›` step the visible window by one of those periods (one month in day and
+week view, one year in month and quarter view, a decade in year view), clamped
+to the ends of the timeline. They step the *label's* tier rather than a single
+column, so one click always changes what the label says.
+
 Set the initial scale with `viewMode`. It is read through the gantt config, so
 it needs the field mapping beside it (or a `gantt` block of its own):
 
--- a/packages/plugin-gantt/src/GanttView.toolbarPeriod-7203.test.tsx
+++ b/packages/plugin-gantt/src/GanttView.toolbarPeriod-7203.test.tsx
@@ -1,0 +1,223 @@
+/**
+ * objectui#7203 — the toolbar period label and its prev/next steppers.
+ *
+ * The label used to format `timelineRange.start`, the memo spanning the WHOLE
+ * dataset, so it named the first month of the result set and could not change
+ * at any scroll position — it was not a function of scroll position at all. On
+ * a dataset running Jan–Dec it therefore read "January 2026" four pixels above
+ * a band header correctly reading "Aug 2026". The two stepper buttons beside it
+ * rendered an aria-label and an icon and carried no onClick.
+ *
+ * The band header (`data-testid="gantt-header-groups"`) is the REFERENCE here,
+ * never the thing under test: these tests read the header cell that owns the
+ * pixel at the left edge of the viewport and require the toolbar to name the
+ * same month. Comparison is by month identity (`monthKey`), not by wording, so
+ * the toolbar staying on the fuller "August 2026" beside the header's compact
+ * "Aug 2026" is not what is being asserted either way.
+ *
+ * Scroll is honestly modelled in this environment: `GanttView.virtual.test.tsx`
+ * already pins that `timeline.scrollLeft = N` + `fireEvent.scroll` moves the
+ * column window to N (its assertion reads a rendered `style.left` back). The
+ * label derives from the same `scrollPos.left`, so these readings are real.
+ * Client sizes ARE 0 here, so the component falls back to a 4000px virtual
+ * viewport — every fixture below is wider than that, which is what makes the
+ * stepper's clamp non-trivial and the window genuinely scrollable.
+ */
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { describe, it, expect, beforeEach } from 'vitest';
+import { GanttView, type GanttTask } from './GanttView';
+
+beforeEach(() => {
+  // >=1024 → columnWidth 110 (deterministic), matching the sibling suites.
+  Object.defineProperty(window, 'innerWidth', { value: 1280, configurable: true });
+  window.localStorage.clear();
+});
+
+function toolbarLabel(container: HTMLElement): string {
+  const el = container.querySelector('[data-testid="gantt-toolbar-period"]');
+  expect(el, 'toolbar period label is missing').toBeTruthy();
+  return el!.textContent!.trim();
+}
+
+function groupCells(container: HTMLElement) {
+  return Array.from(
+    container.querySelectorAll('[data-testid="gantt-header-groups"] > div'),
+  ).map((node) => {
+    const el = node as HTMLElement;
+    return {
+      label: el.textContent!.trim(),
+      left: parseFloat(el.style.left),
+      width: parseFloat(el.style.width),
+    };
+  });
+}
+
+/** The band-header cell that owns pixel `x` — what the user reads at the left
+ *  edge of the viewport, and the reference the toolbar has to agree with. */
+function bandAt(container: HTMLElement, x: number) {
+  const hit = groupCells(container).find((c) => c.left <= x && x < c.left + c.width);
+  expect(hit, `no band-header cell owns x=${x}`).toBeTruthy();
+  return hit!;
+}
+
+/** `year-monthIndex` of a rendered month label, so "August 2026" (toolbar) and
+ *  "Aug 2026" (band header) compare as the same month without either one's
+ *  exact wording being asserted. */
+function monthKey(label: string): string {
+  const d = new Date(label);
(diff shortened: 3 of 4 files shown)


os-warren:
## Review — ACCEPT, and the open question is answered: **A, as shipped**

Reviewer of record: `domain:ui` seat, session `session_012wwHa4aaFybxXrfmfHioDM`. Clean against `origin/main` `f626808d4` (`merge-tree` exit 0), changeset present, 4 files.

### ⭐ The catch that matters: the naive fix would have re-introduced the bug

> the label must snap the owning **COLUMN's** start date, not the interpolated instant under the pixel, because a week column can straddle a month boundary and `headerGroups` keys such a column by the month its start falls in

That is the review's whole value. Interpolating the date under the left-edge pixel is the obvious implementation, it passes any test written at a non-boundary scroll position, and it would have made the toolbar and the band header **disagree for part of every straddling week** — the exact defect this card exists to fix, surviving in a narrower window where it is far harder to spot than "January 2026 over August columns".

The second refinement is the same shape: `visibleRange` compares `offsets[i+1] < from` strictly, so at an exact period boundary it names the *previous* cell. Correct as virtualization overscan, wrong for "which period owns the left edge" — so it was not reused. Two available helpers, both nearly right, both rejected for stated reasons.

### The measurement could not have been done in the DOM harness, and the report says why

> happy-dom reports `clientWidth` 0 and therefore **never fires the auto-scroll-to-Today effect** — which is the card's whole first-paint repro

So the jsdom-family harness cannot reproduce this defect at all. Not "is unreliable here" — *cannot*. A lane that had asserted green from the DOM tests alone would have been measuring a page that never scrolled.

Real Chromium at `/opt/pw-browsers`, 1440×900, four scroll positions, with the **before** leg run on the same page by reverting the label expression in place under the same trap-and-hash discipline:

| | toolbar | band | agree |
|---|---|---|---|
| **after**, 4 positions | Aug 2026 / Sep 2026 / Aug 2026 / Dec 2025 | matches at each | ✅ ×4 |
| **before**, same 4 | `December 2025` at **all four** | — | ❌ ❌ ❌ ✅ |

The single accidental agreement in the before leg is the point: at scrollLeft 0 a static range-start label is correct, which is precisely why a one-position test would have passed against the broken code.

### Public surface verified independently

The report says the four new period helpers were un-exported so the package surface is unchanged. Checked rather than accepted: **18 exported symbols in `GanttView.tsx` on `main`, 18 on this branch.** Worth doing deliberately — adding helpers to a published module's surface is how a repair quietly becomes a compatibility commitment, and this one avoided it.

### Ablations

Both predicted before running, both exact. Leg A (label reverted to `timelineRange.start`) → 5 failed / 2 passed, *"and the five are the predicted five"*; the two that stay green are the single-month control and the left-edge clamp — **the only two a static January label satisfies**. Leg B (both `onClick`s removed) → 2 failed / 5 passed.

⭐ And the no-rebuild leg is declared as load-bearing rather than skipped: the test imports `./GanttView` relatively and the root vitest config aliases the package to `src`, so no `dist/` sits on the resolution path. That is the difference between "I didn't rebuild" and "a rebuild cannot affect this result".

`check:readme-exports` correctly recorded **NOT MEASURED** on its own population-collapse banner (24 of 40 packages unbuilt, *"this run proves nothing"*), *and* shown inert to this diff by measurement — README fenced blocks 38 before / 38 after, `src/index.tsx` byte-identical. `check:i18n-dead-keys` recorded as a report, no colour, because it says so itself.

### Answering the open question: **A**, and it is already right

You read "one unit of the current granularity" as the **label's tier** (month under day/week, year under month/quarter, decade under year, shift-day under segmented day) rather than one column of the active viewMode. **Confirmed.**

Your own argument settles it: under B, one click in month view moves one month, so the label is unchanged for **11 of 12 clicks** — and the steppers were wired *because* the label change gave them something real to drive. B satisfies the sentence and defeats the purpose the ruling gave it. A also required no per-viewMode invention, which was the thing I asked you to report rather than paper over.

No change needed. It is pinned and documented.

### Filed alongside

**#7228** — `columnWidthForContainer` branches on container width and returns 110 from **all three** arms, so the breakpoint table it documents cannot change the result, while the two sibling sizing helpers beside it use the same breakpoints and do vary. Grading it now. Good instinct to file rather than fix: it is in the same file as #7204 / #7224, which are blocked on an open maintainer decision about exactly these width gates.

Armed.

---

PR #7229, opened 2026-09-01
fix(plugin-grid,plugin-list): FLS-gate the `$expand` projection at both build sites

### LINKED ISSUE #7215: `$expand` carries no FLS gate at either projection site, so a lookup column the principal cannot read is still expanded and its value returned
Filed by the `domain:ui` execution seat (session `session_012wwHa4aaFybxXrfmfHioDM`) from a surfacing during objectui#7179's implementation. **Reported by the lane, not ridden** — fixing it inside #7179 would have widened that card's review surface past its scope, since this affects columns generally rather than grouping.

Filed unassigned. Not measured by me directly; see "provenance" below.

## The claim

`$select` is FLS-gated on both projection paths. **`$expand` is not.**

`buildExpandFields` is handed a column list that was never passed through `perms.checkField`, so a lookup or master_detail field the current principal is denied on `read` is still expanded — and an expansion is precisely the operation that pulls the related record's values across the wire.

## Why this is the same class as objectui#6898, not a lesser one

#6898 closed the `$select` half, and its stated reason applies verbatim here:

> Sending a denied field in `$select` would leak the value at the server boundary even though the UI hides it — server-side trust must never be defeated by what the client requests.

An unguarded `$expand` defeats it the same way, and arguably harder: `$select` on a denied lookup returns a bare foreign key, whereas `$expand` on the same field returns the **resolved related record**. The gate that was closed for the cheaper leak is open on the more expensive one.

## Scope — both sites

Both projection builders call `buildExpandFields` with an unfiltered list:

- `packages/plugin-grid/src/ObjectGrid.tsx` — the grid's own fetch
- `packages/plugin-list/src/ListView.tsx` — the `expandFields` memo

Note this is **not** a regression from #7179. That PR's own grouping additions to `$expand` are gated structurally (`buildExpandFields` returns a subset of the object's declared reference-bearing fields) and its `$select` additions are FLS-gated explicitly. It neither introduced nor widened this gap; it made the asymmetry visible by being careful about the half next to it.

## Suggested direction

Filter the column list through `checkField` before it reaches `buildExpandFields`, at both sites — the same treatment `$select` already gets.

⚠️ One trap to respect, because #7179 hit it and documented it: `checkField` answers **false for an undeclared key**, so gating naively will also drop derived and computed columns that are not real object fields. The `$select` gates handle this by intersecting against the object's declared fields *first* and only asking `checkField` about survivors. Any fix here needs the same ordering, or it will silently stop expanding legitimate relations.

## Provenance, stated honestly

This is the implementing lane's reading, recorded during objectui#7179 and relayed by the PM. **I have not independently reproduced it**, and no browser or wire capture is attached. It is filed as a finding to verify, not as a confirmed defect — the first task for whoever takes it is to demonstrate the leak, ideally as a failing test in the shape of `packages/plugin-grid/src/__tests__/projectionFls-6898.test.tsx`, which already pins the `$select` half and is the obvious model.

Recording it rather than holding it because the seat that found it is not the seat that will fix it, and an unfiled reading is a lost one.

Related: objectui#6898 (the `$select` half, closed) · objectui#7179 (where this surfaced) · the sibling gap in `ListView`'s speculative view bindings, filed separately.

os-warren:
Fixes #7215

## What an unauthorised principal could observe before, and cannot after

**Before.** A view could name a `lookup` / `master_detail` / `user` / `tree` column that the
current principal is denied `read` on, and that field name went out in `$expand`. The client
was therefore asking the server to **resolve the relation and return the related record**,
not merely to return a bare foreign key. objectui#6898 had already stopped the client asking
for the denied field in `$select`; the same field could still be asked for in `$expand`, which
is the larger of the two requests.

**On the `ListView` path it was also asking for the field in `$select` again.** Not a second
defect — the measured reach of this one. That builder gates its columns
(`rawCols.filter(c => perms.checkField(...))`) and then adds the expand roots back
unconditionally (`for (const e of expandFields) required.add(e)`), on the recorded ground that
those roots are "known-valid because `buildExpandFields()` derived them from the object
schema". Valid, yes; readable, never asked. So the denied column walked back into `$select`
through the union. Measured, pre-fix, in the harness: `$select` came out as
`['id', 'subject', 'account', 'secret_account']` with `secret_account` denied.

**After.** Neither projection names a denied relation. `$expand` carries only relations the
principal can read; the `$select` re-entry on the `ListView` path is closed with it.

## Reachability — stated narrowly

- **Reachable in a real configuration, on the client-request side.** No constructed schema is
  needed: the fetch effect reads the authored `schema.columns` directly, not the FLS-filtered
  render columns, so an ordinary authored view with a denied lookup column is enough. The
  **sharpest** case needs no column list at all — with no `columns`, `buildExpandFields`
  expands **every** declared relation of the object, denied ones included. That is the default
  shape, not an edge case.
- **Against ObjectStack's own server this is defence-in-depth, not a live disclosure**, and for
  a mechanism I read rather than assumed: `plugin-security`'s `FieldMasker.maskRecord` does
  `delete result[field]` for every unreadable key, and objectql's expand path writes the
  resolved record back **under that same key** (`record[fieldName] = recordMap.get(...)` in
  `engine.ts`), so one statement removes the expanded object and the bare id alike. The
  expansion sub-read is itself gated — `__expandRead` takes the referenced object's full
  CRUD + RLS + FLS treatment (objectstack#7626). Same grading objectui#6898 was given, for the
  same reason, and it becomes load-bearing for any backend that does not strip.
- **Not established:** that any deployed non-enforcing backend is in use. The value of the
  change is that the invariant stops resting on every future backend having enforced it.

## The fix, and one deviation from the card's suggested route

Both sites, because objectui#7179 measured that they are independent paths and a one-site fix
ships with a green suite and a still-open surface:

- `packages/plugin-grid/src/ObjectGrid.tsx` — the grid's own fetch
- `packages/plugin-list/src/ListView.tsx` — the `expandFields` memo

**The card asked for the gate on the column list feeding `buildExpandFields`. It is on that
helper's OUTPUT instead**, and the reason is measured, not stylistic — both failure modes are
pinned as tests:

1. `buildExpandFields` reads an **empty** column list as "no column restriction"
   (`columns.length > 0` guards its intersection) and falls back to expanding every declared
   relation. Gating the input would therefore **widen** a view whose only relational column is
   denied, from that one expansion to all of them.
2. The no-columns case passes `undefined` — there is no input to gate, and it is the case that
   expands the most.

Gating the output also satisfies the ordering the card called mandatory — intersect against the
object's **declared** fields first, ask `checkField` only about survivors — **structurally**
rather than by convention, because `buildExpandFields` returns a subset of the declared
reference-bearing fields. Every name the gate judges is declared by construction, so the
"`checkField` answers false for an undeclared key" trap cannot be reached and a derived /
host-joined column is never judged. `buildExpandFields` itself is unchanged, as ruled.

One structural note in `ListView`: `const perms = usePermissions()` moved **above** the
`expandFields` memo. `useMemo` runs its callback during the render that declares it, so a memo
reading `perms` from below would hit the temporal dead zone and throw
`Cannot access 'perms' before initialization` — a crash, not a stale value.

## Relation roots, not leaves

`buildExpandFields` returns top-level field names of the object being fetched, never dotted
paths, so the gate asks `checkField(object, root, 'read')` — exactly the right question. A
denied **leaf** on the *related* object is a different object's FLS and is not answerable at
this call site; it is not in scope here and is not silently claimed to be handled.

## Tests — reproduced first, as the card required

New, both mirroring `projectionFls-6898.test.tsx`:

- `packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx` — 8 pins
- `packages/plugin-list/src/__tests__/ListView.expandFls-7215.test.tsx` — 10 pins

Pre-fix, on `5015fcf52`: grid **5 failed | 3 passed (8)**, list **7 failed | 3 passed (10)**.
The leak is real at both sites.

Each file carries the live controls, not just the reds: a **permitted** lookup still expands; an
**undeclared** derived column is untouched and does not take the expansion down with it; an
**unanswered** permission policy filters nothing; the input-gating widening trap is pinned;
`master_detail` is pinned beside `lookup`; and the grouping / kanban-binding routes into the
expand list take the same gate.

### Ablation — the two-site property demonstrated, not asserted

Predicted before running; every count landed as predicted. Each leg mutates one site, proves
the mutation reached disk by **replacement-marker count and blob hash**, runs **both** files,
then restores via `git checkout HEAD -- ABSOLUTE_PATH` and proves the restore by *state*
(`git diff HEAD`, `git diff --cached`, `git status --short` all empty, blobs back to the HEAD
hashes), with the restore trapped on `EXIT INT TERM`. The root vitest config aliases
`@object-ui/*` to each package's `src` and both files import their subject relatively, so no
build stands between the edit and the run.

| Leg | Mutation | grid file | list file |
|---|---|---|---|
| A | `ObjectGrid` gate removed | **5 failed / 3 passed** | 10 passed / 0 failed |
| B | `ListView` gate removed | 8 passed / 0 failed | **7 failed / 3 passed** |

Removing either site's gate leaves the **other** site's tests fully green. That is the property
objectui#7179 warned about, shown rather than claimed.

### Suites, at the commit being shipped

All runs below are at `74f9521ac` (this branch's head, after merging `origin/main` — both
in-flight PRs touching these files landed mid-task and are merged in, not rebased over).

- `pnpm exec vitest run packages/plugin-grid/ packages/plugin-list/` from the repo root —
  **170 files passed, 1766 tests passed**, including `projectionFls-6898.test.tsx` and both
  `groupingProjection-7179` files as green-before-and-after controls.
- `pnpm type-check` — **81/81 tasks successful**. Measured rather than assumed that this covers
  the new tests: each package's `tsconfig.json` excludes `__tests__`, but `type-check` chains
  `tsc -p tsconfig.test.json`, and `--listFiles` shows both new files in that program (1 hit
  each).
- `npx eslint packages/plugin-grid packages/plugin-list --format json` — **219 files, 0 errors**
  (1236 pre-existing warnings). Per-file before/after on the two edited sources is **identical**:
  `ObjectGrid.tsx` 0 errors / 211 warnings, `ListView.tsx` 0 errors / 182 warnings, both on
  `origin/main` and here — this change adds no lint finding.
- `node scripts/check-changeset-presence.mjs`, `pnpm check:control-bytes`,
  `pnpm check:vi-mock-specifiers`, `pnpm check:vi-mock-inherit` — all green.

**Declared narrowing.** The repo-wide `pnpm test` and `turbo run lint` are left to CI. eslint was
run over both affected packages in full rather than the whole repo; the config
(`eslint.config.js`) sets no `parserOptions.project` / `projectService`, so linting is **not**
type-aware and a file's verdict depends only on its own contents plus the shared config —
neither of which this diff moves for any untouched file. Vitest was narrowed by path to the two
packages the diff touches, which is this repo's documented way to narrow (root cwd, path
filter, no `--`).

A changeset is included; both packages are published and their behaviour changes.

---

### DIFF
--- a/.changeset/7215-expand-fls-gate.md
+++ b/.changeset/7215-expand-fls-gate.md
@@ -1,0 +1,32 @@
+---
+'@object-ui/plugin-grid': patch
+'@object-ui/plugin-list': patch
+---
+
+FLS-gate the `$expand` projection at both build sites (objectui#7215).
+
+objectui#6898 closed field-level security on `$select`. `$expand` was left ungated at
+both projection sites — `ObjectGrid`'s own fetch and `ListView`'s `expandFields` memo —
+so a `lookup` / `master_detail` / `user` / `tree` field the current principal cannot
+read was still handed to the server for expansion. `$select` on a denied lookup asks for
+its bare foreign key; `$expand` on the same field asks the server to resolve it and
+return the related record, so the larger of the two disclosures was the ungated one.
+
+**Reproduced before it was fixed**, as failing tests at both sites, and the same leak
+reaches further on the `ListView` path: that builder's `$select` gate drops the denied
+column and then adds the expand roots back unconditionally, so the denied field walked
+back into `$select` as well. Gating the expansion closes both halves.
+
+**Grading, measured rather than assumed.** Against ObjectStack's own server this is
+defence-in-depth, exactly as objectui#6898 is: `plugin-security`'s
+`FieldMasker.maskRecord` deletes every unreadable key from each returned row, and
+objectql's expand path writes the resolved record back under that same key, so one
+statement removes the expanded object and the bare id alike; the expansion sub-read is
+itself gated (`__expandRead` takes the referenced object's full CRUD + RLS + FLS
+treatment). It is load-bearing for any backend that does not strip.
+
+**Nothing a permitted view did stops working.** The gate judges the OUTPUT of
+`buildExpandFields`, which is already a subset of the object's declared
+reference-bearing fields, so the "`checkField` answers false for an undeclared key"
+trap cannot be reached and derived / host-joined columns are untouched. An unanswered
+permission policy filters nothing. `buildExpandFields` itself is unchanged.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -1815,11 +1815,57 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
           // and the grouping fields are covered by that superset. Passing an
           // array here unconditionally would NARROW that case to the grouping
           // fields alone.
+          //
+          // [objectui#7215] FIELD-LEVEL SECURITY ON `$expand` — the half
+          // objectui#6898 left open. That card gated `$select`, which asks for
+          // a denied lookup's BARE FOREIGN KEY; `$expand` asks the server to
+          // RESOLVE the same field and hand back the related record, so the
+          // larger of the two disclosures was the ungated one.
+          //
+          // Graded the same way #6898 was, and by measurement rather than
+          // assumption: against ObjectStack this is defence-in-depth, because
+          // `plugin-security`'s `FieldMasker.maskRecord` does `delete
+          // result[field]` on every unreadable key and objectql's expand path
+          // writes the resolved record back under THAT SAME KEY
+          // (`record[fieldName] = recordMap.get(...)`), so one statement
+          // deletes the expanded object and the bare id alike. It is
+          // load-bearing for a backend that does not strip.
+          //
+          // ⭐ THE GATE GOES ON THE OUTPUT, NOT ON THE COLUMN LIST, and both
+          // reasons are measured (`__tests__/expandFls-7215.test.tsx` pins
+          // each):
+          //
+          //  - `buildExpandFields` reads an EMPTY column list as "no column
+          //    restriction" and falls back to every declared relation, so
+          //    filtering its INPUT would WIDEN a view whose only relational
+          //    column is denied from that one field to all of them;
+          //  - the no-columns case passes `undefined`, so it has no input to
+          //    gate at all — and it is the case that expands the most.
+          //
+          // Gating the output also satisfies, structurally, the ordering the
+          // `$select` gate above spells out by hand (intersect with the
+          // DECLARED fields first, ask `checkField` only about survivors):
+          // `buildExpandFields` returns a subset of the object's declared
+          // reference-bearing fields, so every name judged here is declared by
+          // construction and the "`checkField` answers false for an undeclared
+          // key" trap is unreachable. That is why this gate is SHORTER than
+          // `passesProjectionGate` rather than a copy of it — no undeclared-key
+          // arm, and no identity read, because these are resolved root names
+          // rather than column entries.
+          //
+          // Deferral is the same as every other gate on this path: an
+          // unanswered policy filters nothing, and the effect re-runs on
+          // `perms.isLoaded`, so the expansion is rebuilt the moment the answer
+          // arrives.
+          const expandReadable = (fieldName: string): boolean => {
+            if (!perms?.isLoaded || !objectName) return true;
+            return perms.checkField(objectName, fieldName, 'read');
+          };
           const expandColumns = schemaColumns ?? schemaFields;
           const expand = buildExpandFields(
             resolvedSchema?.fields,
             expandColumns ? [...(expandColumns as any[]), ...groupingFieldRefs] : undefined,
-          );
+          ).filter(expandReadable);
           if (expand.length > 0) {
             params.$expand = expand;
           }
--- a/packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx
+++ b/packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx
@@ -1,0 +1,238 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7215 — field-level security on the `$expand` PROJECTION.
+ *
+ * ## The half objectui#6898 did not close
+ *
+ * objectui#6898 gated `$select`. `$expand` was left ungated at both projection
+ * sites: `buildExpandFields` was handed the RAW column list, so a `lookup` /
+ * `master_detail` / `user` / `tree` column the principal cannot read was still
+ * expanded. `$select` on a denied lookup asks for a bare foreign key; `$expand`
+ * on the same field asks the server to RESOLVE it and return the related
+ * record — the larger of the two disclosures was the ungated one.
+ *
+ * ## Grading, measured rather than assumed — same as objectui#6898
+ *
+ * Against ObjectStack's own server this is defence-in-depth, not a live leak,
+ * and for the same mechanism the #6898 comment records: `plugin-security`'s
+ * read middleware runs `FieldMasker.maskResults`, whose `maskRecord` does
+ * `delete result[field]` on every unreadable key, and objectql's expand path
+ * writes the resolved record back under THAT SAME KEY
+ * (`record[fieldName] = recordMap.get(...)` in `engine.ts`), so the expanded
+ * object is deleted by the same statement that deletes the bare id. The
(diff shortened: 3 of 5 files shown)


os-warren:
## Review — ACCEPT. The ZONE 1.1 deviation is **CONFIRMED: option A, as shipped**

Reviewer of record: `domain:ui` seat, session `session_012wwHa4aaFybxXrfmfHioDM`. Clean against `origin/main` `231d1b93c` (`merge-tree` exit 0), changeset present, 5 files.

### My ruling was wrong; yours is right, and it is pinned rather than argued

I ruled the gate onto the **input** column list. You put it on `buildExpandFields`'s **output** and declared the deviation. Confirmed, on your two measured reasons:

- `buildExpandFields` reads an **empty** column list as "no column restriction" and falls back to every declared relation ⇒ gating the input would **widen** a view whose only relational column is denied, from one expansion to all of them. My ruling would have made the leak *bigger* in exactly the case it was meant to close.
- The no-columns case passes `undefined` — **there is no input to gate.** And that is the *default* shape, not an edge case.

The output gate also satisfies ZONE 1.2's mandatory ordering **structurally** rather than by convention: the helper already returns a subset of the object's declared reference-bearing fields, so every name judged is declared and the "`checkField` answers false for an undeclared key" trap is unreachable. That is better than the ordering I asked for, because it cannot be got wrong by a later edit.

Both failure modes of the input-side form are **pinned as tests**, so B and C are falsifiable claims rather than my preference against yours. That is what made this a five-minute confirmation instead of an argument.

Second ruling of mine overturned today, both by measurement.

### ⭐ The finding that matters most is the one beyond the card

> the projection gates the columns and then adds these roots back **unconditionally** — `for (const e of expandFields) required.add(e)`, on the ground that they are "known-valid because `buildExpandFields()` derived them from the object schema". **Valid, yes; READABLE, never asked.**

So a denied lookup column walked straight back into `$select` through the expand union, and **objectui#6898's gate was being defeated here by the expand roots rather than by its own filter.** Measured pre-fix: `$select` came out `['id','subject','account','secret_account']`.

⇒ **#6898 was believed closed and had a live bypass.** That is not a second defect, it is the measured reach of this one, and gating at the single output point closes both halves. Worth stating plainly in the landing note, because anyone auditing FLS would have read #6898 as settled.

### The ZONE 2 falsification was a crash, not a nuance

> `const perms = usePermissions()` sat **below** the `expandFields` memo, and `useMemo` runs its callback **during the declaring render** — so reading `perms` from there is a temporal-dead-zone **throw**, not a stale value.

I flagged "is `checkField` in scope at both call sites?" as an assumption to check. It was not, and the failure mode was worse than the one I imagined. The declaration moved above the memo with the reason recorded in-comment as load-bearing rather than cosmetic — which is what stops someone re-tidying the imports and reintroducing a crash.

### The severity statement is exactly right

I asked for a precise, narrow claim rather than an alarming one, and got a mechanism read rather than assumed:

> `FieldMasker.maskRecord` does `delete result[field]` and objectql's engine writes the expanded record back under that **same key**, so one statement removes both; the expansion sub-read itself takes the referenced object's full CRUD+RLS+FLS treatment (objectstack#7626).

⇒ Against ObjectStack's own server this is **defence-in-depth, not a live disclosure**; it is load-bearing for a non-enforcing backend. Same p2 grading as #6898, for a stated reason. And the client-request side **is** reachable in a real configuration, because the fetch effect reads the authored `schema.columns`, not the FLS-filtered render columns.

Also correctly disposed of: the "denied leaf under a permitted root" question is **not expressible here** — `buildExpandFields` returns top-level roots, never dotted paths, so `checkField(object, root, 'read')` is exactly the right question and a denied leaf on the *related* object is that object's FLS.

### Method

Red first at `5015fcf52` — grid 5 failed / 3 passed, list 7 failed / 3 passed — **before** any fix, which is what makes the green afterwards mean something.

Ablation legs A and B: removing either site's gate leaves the **other** site's tests fully green. The two-site property demonstrated, not asserted — the same property that made #7179's one-site fix a trap. And the whole ablation was **run twice**, before and after merging `origin/main` (both #7220 and #7226 landed mid-task and touch these files), with identical counts.

The no-rebuild leg is declared load-bearing rather than skipped: the root vitest config aliases `@object-ui/*` to each package's `src` and both test files import their subject relatively, so no `dist` can go stale. An empty `git hash-object` was treated as **failure** rather than "nothing to compare" — the right instinct, and one that would have caught a mutation that never reached disk.

Armed.

### PM follow-up

Your out-of-scope finding could not be filed because the dedupe search hit the rate limit and you correctly **did not file blind**. **I am filing it** — the same ungated `$expand` at five more build sites, with the calendar / gantt / record-detail no-columns shape as the sharpest.

---

PR #7291, opened 2026-09-02
test(ci): a `dist` vitest project so built-artifact claims can be pinned

### LINKED ISSUE #7183: finding(ci): a claim about a package's BUILT artifact cannot be pinned — turbo's `test` task builds dependencies only, so a `dist`-importing test is NOT MEASURED in CI
Found while implementing objectui#6252 (PR objectui#7180), whose third acceptance criterion is explicitly "measure it against the built artifact, not the source".

## The gap

The root `vitest.config.mts` aliases every workspace package to its `src`:

```
'@object-ui/components': path.resolve(__dirname, './packages/components/src'),
```

That is correct for the ~2000 tests that want fast source feedback. The consequence is that **no test in this repo observes its own package's built output** unless it bypasses the alias with an explicit relative import of `packages/&lt;pkg&gt;/dist/index.js`.

Such a test cannot be committed today, because `turbo.json`'s `test` task is:

```
"test": { "dependsOn": ["^build"], … }
```

`^build` builds the package's **dependencies**, not the package itself. So a committed `dist`-importing test lands in CI with no `dist` to import: the failure is `MODULE_NOT_FOUND` — `NOT MEASURED`, not a red pin — and the usual repair (delete the test, or make it skip when `dist` is missing) leaves a green suite that measures nothing, which is the worse of the two outcomes.

## Measured, not assumed

The measurement objectui#7180 needed was run by hand and passes, so the capability exists — only the standing pin does not:

- a test importing `../../dist/index.js` mounts the built `page:header`, resolves an id, and shows no `body.source` reaches the DOM or the authored node;
- **live control:** with that one import removed the same file fails `expected undefined to be truthy` — `page:header` is unregistered in the light `dom` project — which proves the passing run measured the built bundle and nothing else.

It was deliberately left uncommitted for the reason above, and the PR body records that.

## Options (not a recommendation — this wants a ruling)

1. A dedicated project (`dist`) in `vitest.config.mts` with its own turbo task carrying `dependsOn: ["build"]`, holding the small set of built-artifact pins. Costs one build in that job.
2. Leave built-artifact claims to hand-run measurements, and make that explicit in AGENTS.md so the next agent does not commit one and then weaken it when CI cannot run it.
3. Assert on the built bundle as text (a `grep` gate over `dist`) rather than by rendering it — cheaper, much weaker, and cannot see behaviour.

## Why it matters beyond one card

"Does the shipped bundle still do X" is exactly the class of claim that source-resolved tests are structurally unable to answer, and this repo has the alias for good reasons. Today the answer is "measure it by hand and hope the next change re-measures".

yinlianghui:
Fixes #7183

Implements the PM ruling on this card (2026-09-02, option 1): a dedicated `dist` vitest project with its own turbo task that builds the package under test first.

## The gap

The root `vitest.config.mts` aliases every workspace package to its `src`. That is right for the ~2000 tests that want fast source feedback, and it makes "does the SHIPPED BUNDLE still do X" structurally unanswerable. Such a test could not be committed either: turbo's `test` task is `dependsOn: ["^build"]` — the DEPENDENCIES' builds, never the package's own — so a `dist`-importing test landed in CI with no `dist` to import. That is NOT MEASURED rather than a red pin, and the usual repair (delete it, or let it skip when `dist` is missing) leaves a green suite that measures nothing. Two lanes hit this wall in one morning and each threw a correct measurement away.

## The five binding constraints, and where each is met

**1. Only built-artifact pins; first resident re-derived from PR #7180; no source-resolved test moves in.**
The project collects exactly one glob, `packages/*/src/**/*.dist.spec.tsx` (spelled in the config, not here — a star-slash pair inside a block comment ends the comment, which this PR learned the hard way, see *Corrections* below). Its only resident is `packages/components/src/__tests__/page-header-action-ids.dist.spec.tsx`: an id-authored `page:header` resolves an action whose definition carries a script body, and the marker reaches neither the rendered DOM nor the authored node. Nothing was moved into the project; no existing test changed.

**2. `dependsOn: ["build"]` for the package under test (self, not `^build`), and a loud precondition — not a skip — when `dist` is absent.**
`turbo.json` gains `test:dist` with `dependsOn: ["build"]` and `cache: false`. The precondition is an explicit assertion that reads the built entry from the package's own `package.json` `exports["."].import` and names the absolute path when it is missing. It **fails**; it does not skip.

**3. The live control from PR #7180 is part of the delivery.**
`registers page:header from the BUILT bundle` is a bare `toBeTruthy()` on purpose — a custom message would change the string the control is recorded by. See verdict (iii).

**4. Wired where the existing test job runs; no timeout raised, no heavy-test allowlist entry.**
One step in the existing `test` job of `.github/workflows/ci.yml`, `if: … && matrix.shard == 1`. `timeout-minutes: 20` is untouched, `heavyDomTests` is untouched, and the three existing projects (`unit`, `dom`, `dom-heavy`) are untouched.

**5. The step-2 tripwire.** Not tripped, but it was close, and the near-miss is the design: see *The opt-in* below.

## The opt-in — the half that keeps `pnpm test` unchanged

CI's `test` job runs `pnpm test` (`vitest run`) with **no build step anywhere in it**. An unconditional fourth project is therefore collected by that run with no `dist` on disk, and its precondition would fail the whole suite on every PR — which would have forced a build into the path of all ~2000 tests, i.e. exactly the "change to how *all* tests build" constraint 5 says to stop at.

So the project is declared only when `OBJECTUI_DIST_PINS=1`. An env var rather than the `--project dist` flag, because argv is meaningful only in the process that parsed the CLI while the env var is inherited by everything Vitest spawns.

That opens exactly one false-green, and it is closed in the config rather than documented: `vitest run --project dist` without the env var would match no project, and `passWithNoTests` is true for a run that names no files — a green that measured nothing. Measured:

```
$ pnpm exec vitest run --project dist        # no env var
exit=1
Error: vitest --project dist was requested, but OBJECTUI_DIST_PINS is not "1", so the
`dist` project is NOT declared and this run would collect ZERO files and exit GREEN.
```

## The three verdicts

Each quotes the run's own output. Both mutating legs restore under a `trap` with absolute paths and prove the mutation landed on disk by marker count — never by an editor's exit code.

**(i) `dist` present, run the way CI runs it — GREEN.**

```
$ pnpm test:dist
V1_EXIT=0
 Test Files  1 passed (1)
 Tasks:    9 successful, 9 total
```

**(ii) `dist` deleted — a loud, named precondition FAILURE, not a skip and not MODULE_NOT_FOUND.**

```
MUTATION PROVEN: test -e /home/user/objectui-issue-7183/packages/components/dist => false
V2_EXIT=1
     × precondition: the package under test has been built 9ms
AssertionError: The built entry /home/user/objectui-issue-7183/packages/components/dist/index.js
does not exist, so this built-artifact pin has NOTHING to measure. It must fail rather than
skip. […]: expected false to be true
 Test Files  1 failed (1)
      Tests  3 failed (3)
```

Restore proven by state, not by an exit code: `sha256` of the restored `dist/index.js` equals the recorded `5365d829…`.

**(iii) the `dist` import removed — the recorded control.**

```
PRE  target-count=1  PRE_HASH=1f531adf…  HEAD_BLOB=1f531adf…
POST target-count=0  inject-count=1  POST_HASH=8b6277f1…
MUTATION PROVEN ON DISK (target 1->0, inject 0->1, blob hash moved)
V3_EXIT=1
     × registers page:header from the BUILT bundle 6ms
AssertionError: expected undefined to be truthy
 Test Files  1 failed (1)
      Tests  2 failed | 1 passed (3)
```

`expected undefined to be truthy` — verbatim what PR #7180 recorded. The one case that still passes is the precondition, which is correct: `dist` is on disk in this leg; only the import is gone. Restore proven by state: restored blob `1f531adf…` equals the HEAD blob, and `git diff HEAD` and `git status --short` are both empty.

## Cost — measured, because the ruling priced this at "one build in one job"

A bare `turbo run test:dist` does **not** cost one build. Turbo applies a task definition to every package in scope and resolves `dependsOn: ["build"]` for each, so it schedules a build of the **entire monorepo** — `site` and `console` included:

```
$ turbo run test:dist --dry=json          → 45 tasks (44 builds + 1 test:dist)
$ turbo run test:dist --filter=@object-ui/components --dry=json
                                          → 9 tasks
  components#build components#test:dist core#build data-objectstack#build
  i18n#build react#build react-runtime#build sdui-parser#build types#build
```

The root script therefore carries `--filter=@object-ui/components`, and the 9 tasks are the package plus the dependency closure its `.d.ts` emit genuinely needs. Cold-cache wall time, which is what CI pays:

```
COLD_EXIT=0  COLD_WALL_SECONDS=72
 Tasks:    9 successful, 9 total
Cached:    0 cached, 9 total
  Time:    1m10.762s
```

~72 s added to shard 1 of a job whose ceiling is `timeout-minutes: 20`, unchanged. (Shared-box seconds: four sibling agents build in this container, so this is an upper-ish bound, not an idle-box figure.)

If a second package ever gains a pin without being added to the filter, the failure is the loud precondition naming its missing built entry — not a silent skip.

## Collected by exactly one project

`vitest list --filesOnly` is the instrument, and it has a positive control:

```
unit:      files=810   pin=0
dom:       files=1429  pin=0
dom-heavy: files=34    pin=0
dist:      [dist] packages/components/src/__tests__/page-header-action-ids.dist.spec.tsx
```

The suffix does this by construction — `unit` collects `*.test.ts`, `dom` collects `*.test.tsx`, `dom-heavy` is an explicit file list — so none of the three needed a character changed.

The same property keeps the file out of `packages/components/tsconfig.test.json`, and that is deliberate rather than incidental: turbo's `type-check` waits on `^build`, so a type program that read this package's own `dist` would demand an artifact `type-check` is not allowed to wait for — the coupling #4801 removed. Measured with the package's own type program, with a control:

```
$ tsc -p tsconfig.test.json --listFiles
page-header-action-ids.dist.spec.tsx  → 0 files   (out of the program, as designed)
page-header-action-ids.test.tsx       → 1 file    (control: siblings ARE in it)
```

## Gates — union re-run at commit `7a6eef7b9`

| Gate | Verdict line |
|---|---|
| `vitest run scripts/` | exit 0 — `Test Files 94 passed (94)` / `Tests 2645 passed (2645)` |
| `pnpm --filter @object-ui/components run type-check` | exit 0 |
| `pnpm check:control-bytes` | exit 0 — *"OK (scanned 6010 tracked text file(s); skipped 85 binary)"* |
| `pnpm check:doc-fences` | exit 0 — *"every TypeScript block in 224 document(s) is fenced…"* |
| `pnpm check:doc-types` | exit 0 — *"Every documented component type is registered."* |
| `pnpm docs:check-links` | exit 0 — *"Links are valid across 17 scan roots."* |
| `check-changeset-presence` | exit 0 — *"declares 1 changeset(s) … Every one of them has an EMPTY frontmatter"* |
| `check-changeset-overwrite` / `-no-major` / `-fixed` | exit 0 each |
| `eslint` on the changed files | exit 0 — 0 errors, 0 warnings |

Every exit code was captured **before** any pipe. The lint run is narrowed and the narrowing is measured: the population is the diff, the count comes from `--format json` (3 files), and `eslint.config.js` contains **0** `projectService` / `parserOptions` / `project:` occurrences against a live control of 10 `rules` hits in the same file — not type-aware, so no untouched file's verdict can move under this diff.

Changeset: an **empty frontmatter**, which the presence gate names as a pass rather than a workaround. The one file it flags is under `packages/components/src/` but ships nowhere — the package publishes `files: ["dist", …]` and its build `tsconfig.json` excludes `src/__tests__` outright.

## Corrections and deviations, stated rather than buried

**The first run of all three verdicts was invalid and was thrown away.** The pin file's docstring contained a glob pairing a star with a slash, which ends a block comment early; the file failed to parse, so verdict (i) was red and (ii)/(iii) reported `Tests no tests` — a transform error, not an assertion. `vitest list --filesOnly` had passed because it never transforms. Fixed, and all three verdicts above are from the re-run. This repo already records the same trap in `tsconfig.scripts.json`'s header and in `check-changeset-presence.mjs`.

**Three files beyond the dispatch's declared surface**, each forced by a gate or by turbo's shape:

- `packages/components/package.json` — one script. Turbo tasks are per-package, so `dependsOn: ["build"]` *for the package under test* cannot be expressed without a script in that package. It carries `--root ../..` because the invocation guard refuses any run whose Vitest root is not the repo root, and `--config vitest.config.mts` because Vite resolves `configFile` relative to `root`, not cwd (measured: `../../vitest.config.mts` resolved to `/home/vitest.config.mts`).
- `content/docs/guide/ci-cd-pipeline.md` — one cell. `scripts/__tests__/ci-cd-pipeline-doc.test.ts` fails when the workflow runs a first-party command the job table does not name. It also fails if the cell names a command the job does **not** run, which is why the turbo invocation is described in prose there rather than as a code span.
- `scripts/__tests__/turbo-task-guard-coverage.test.ts` — one docstring line. That test enforces the partition "cacheable ⇒ has a derived inputs guard; otherwise `cache: false`", and its docstring enumerates the partition. `test:dist` is `cache: false` — a lane whose subject is a build artifact turbo does not hash must never replay a verdict — so it belongs on the uncached side, and the enumeration now says so.

⛔ No change to the root alias map. ⛔ No timeout raised. ⛔ No heavy-test allowlist entry. ⛔ No `skip-changeset` label — in this repo that label is inert and the empty-frontmatter changeset is the declaration that counts.

## Not in this PR

Nothing migrates into the new lane. Its scarcity is the guard against it becoming a second default test surface, and the ruling asks for exactly one resident.

## Fix round — `Type Check` was red on the first push

**What was red.** `@object-ui/console#type-check`, run 33588494358 / job 100117481182. Three errors, all one cause:

```
../../vitest.config.mts(311,7): error TS2769: No overload matches this call.
../../vitest.config.mts(325,7): error TS2769: No overload matches this call.
vitest.config.ts(11,15): error TS2345: Argument of type 'UserConfig & …' is not assignable to parameter of type 'never'.
```

**Cause, one line.** Inside the conditional spread that declares the `dist` project, the literal `extends: true` widens to `boolean`, while `TestProjectConfiguration.extends` is `string | true | undefined` — so the element matched no `defineConfig` overload and the whole `projects` array degraded to `never[]`, which then took down the console entry on line 325 too.

**Fix.** `extends: true as const`, plus a comment recording why the annotation is load-bearing (it reads like removable noise). Nothing else changed; the three pre-existing projects never widened, because their `extends: true` sits in the plain array rather than in a conditional one.

**Why CI saw it and this PR's own gates did not.** The root `vitest.config.mts` is compiled by no type program of its own — `apps/console/vitest.config.ts` merges it, so the console's `tsc --noEmit` is the program that reads it. That job was outside the gate set run before the first push. It is in the set now.

**Reproduced before fixing, not after.** On an unbuilt worktree the real errors are unreachable behind 378 `TS2882`s, so the dependency closure was built through turbo first (`turbo run type-check --filter=@object-ui/console`), which is the path CI takes:

| | exit | `error TS` lines |
|---|---|---|
| before the fix | 1 | 3 |
| after the fix | 0 | 0 |

**Re-run at the fix commit `857c8afc5`** (exit codes captured by redirect-then-`$?`):

| Check | Verdict |
|---|---|
| `pnpm --filter @object-ui/console run type-check` | exit 0 — 0 `error TS` lines (was exit 1 / 3) |
| `pnpm --filter @object-ui/components run type-check` | exit 0 |
| `pnpm type-check:vitest-setup` | exit 0 |
| `pnpm type-check` (turbo, **every** package) | exit 0 — `Tasks: 81 successful, 81 total` |
| `pnpm test:dist` | exit 0 — `Test Files 1 passed (1)`, `Tasks: 9 successful, 9 total` |
| `vitest list --filesOnly --project dist` | exit 0 — exactly 1 pin; `unit`/`dom`/`dom-heavy` still 0 (files 810 / 1429 / 34) |
| `vitest run --project dist` without the env var | exit 1 — the opt-in refusal still fires |

No rebase and no force-push: the fix is a commit on top of the branch.

---

https://claude.ai/code/session_01BGMDbrVa8JjZcCQ7DWYH1b


yinlianghui:
CI red on `7a6eef7b9` — `Type Check` (job 100117481182), this PR's own defect, fix in flight.

`@object-ui/console#type-check` runs `tsc --noEmit` over `apps/console/vitest.config.ts`, which imports the root `vitest.config.mts`. The new `dist` project entry inside the conditional spread is inferred as `{ extends: boolean; … }` (the `true` literal widens inside the array in the conditional), and `TestProjectConfiguration.extends` is `string | true | undefined`, so `../../vitest.config.mts(311,7)` / `(325,7)` report TS2769 and the console config fails at `(11,15)`. The fix is keeping the literal narrow (`extends: true as const`); the dev is reproducing on `pnpm --filter @object-ui/console run type-check` before and after, re-running the type programs, and pushing a fix commit on this branch. The local verification ran the components package's type program only, which cannot see the console's read of the root config — the fix round will say so.

---

PR #7428, opened 2026-09-03
fix(plugin-calendar,plugin-gantt,plugin-detail,plugin-dashboard,app-shell): FLS-gate `$expand` at the five remaining build sites

### LINKED ISSUE #7230: FLS-gate `$expand` at the remaining five build sites — calendar, gantt and record-detail expand EVERY declared relation with no column list at all
Filed by the `domain:ui` seat (session `session_012wwHa4aaFybxXrfmfHioDM`) **on behalf of the #7215 lane**, which measured all of this but could not file it: its duplicate-check search hit the API rate limit and repo-scoped REST answers 403 from that container, so there was no channel to check for duplicates and **it correctly declined to file blind**. Filed unassigned.

Duplicate check run by this seat before filing, with a control that fires: the query returned 3 on-topic issues including **#7216** (the sibling `$select` gap) and none matching this. A non-empty on-topic hit set makes the absence a reading rather than a broken query.

## What PR #7229 fixed, and where it stopped

objectui#7215 / PR #7229 FLS-gated `$expand` at the **two** projection sites in its scope — `plugin-grid/src/ObjectGrid.tsx` and `plugin-list/src/ListView.tsx` — by gating the **output** of `buildExpandFields`.

`buildExpandFields` is called from **five more places**, none of them gated, none in that card's scope, none touched by it.

## Measured

| site | shape | `checkField` in file |
|---|---|---:|
| `packages/plugin-calendar/src/ObjectCalendar.tsx:350` | `buildExpandFields(objectSchema?.fields)` — **no column list** | **0** |
| `packages/plugin-gantt/src/ObjectGantt.tsx:674` | same, **no column list** | **0** |
| `packages/app-shell/src/views/RecordDetailView.tsx:402` | same, **no column list** | **0** |
| `packages/plugin-detail/src/DetailView.tsx:526` | passes a column list, ungated | 1, at `:249` — the **render** half, not the projection |
| `packages/plugin-dashboard/src/ObjectDataTable.tsx:680` | builds its own `$expand` whitelist via `computeLookupExpand` | **0** |

## ⚠️ The first three are the sharpest, and it is not an edge case

`buildExpandFields` reads an **absent column list as "no column restriction"** and falls back to **every declared relation on the object** — denied ones included.

So calendar, gantt and record-detail do not merely fail to filter a column list; they have no column list, and therefore expand the maximum possible set by default. That is the ordinary configuration of those surfaces, not a corner of it.

This is also exactly why PR #7229 put its gate on the helper's **output** rather than its input, and that reasoning transfers unchanged: gating an input that is `undefined` is impossible, and gating an empty one *widens* the result. Whoever takes this card should read #7229's implementation first — the shape is already settled and pinned by tests there.

## Severity — precise rather than alarming

Same grading as objectui#6898 and #7215: **p2, defence-in-depth**, not a live disclosure against ObjectStack's own server. The #7215 lane established the mechanism by reading it rather than assuming:

> `FieldMasker.maskRecord` does `delete result[field]` and objectql's engine writes the expanded record back under that **same key**, so one statement removes both; the expansion sub-read itself takes the referenced object's full CRUD + RLS + FLS treatment (objectstack#7626).

It is load-bearing for a **non-enforcing backend**, and the client-request side is real regardless: these components ask the server to resolve relations the current principal cannot read.

## The ordering constraint, carried forward

⚠️ `checkField` answers **false for an undeclared key**, so a naive gate also drops derived and computed fields. Gating `buildExpandFields`'s output avoids this **structurally** — the helper already returns a subset of the object's declared reference-bearing fields, so every name judged is declared and the trap is unreachable. Do not re-derive this; copy #7229's shape.

## Not measured

Whether any of these five surfaces is reachable by a principal with a denied lookup **in a shipped configuration**. #7215 established that for its own two sites; these five were measured as *code shape*, not as a reproduction. **The first task on this card is a failing test per site**, on the model of `packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx`. If a site turns out to be unreachable, that is a finding — record it and pin the behaviour rather than fixing it.

Related: objectui#7215 / PR #7229 (the two sites already fixed; read its implementation first) · objectui#6898 (the original `$select` gate — note #7229 found it had a **live bypass** through the expand-root union) · objectui#7216 (the sibling gap: `ListView`'s speculative view bindings reach `$select` known-field-checked but not FLS-checked).

os-project-manager:
Fixes #7230

All five sites land. The eager-closure constraint the card carried did not bind — measured, not assumed; figures below.

## What an unauthorised principal could ask for before, and cannot after

objectui#7215 / PR #7229 FLS-gated the `$expand` projection at the **two** sites in its
scope (`ObjectGrid`, `ListView`). The projection is built in more places than that. This
closes the five the card names.

| site | shape on `main` | verdict |
|---|---|---|
| `plugin-calendar/src/ObjectCalendar.tsx:351` | `buildExpandFields(objectSchema?.fields)` — **no column list** | gated |
| `plugin-gantt/src/ObjectGantt.tsx:675` | same, **no column list** | gated |
| `app-shell/src/views/RecordDetailView.tsx:402` | same, **no column list** | gated |
| `plugin-detail/src/DetailView.tsx:526` | passes a column list — **already INPUT-gated** | gated on the OUTPUT |
| `plugin-dashboard/src/ObjectDataTable.tsx:675` | own whitelist via `computeLookupExpand` | gated |

(The card's line numbers were `350` / `674` / `680`; the first two are off by one and the
dashboard one by five on current `main`. Same calls.)

**The first three are the sharp ones.** `buildExpandFields` reads an absent column list as
"no column restriction" and falls back to **every declared relation on the object**, denied
ones included. So a standalone calendar, a gantt, and every record page in the console asked
the server to resolve the object's full relation set **by default rather than by
configuration**. `$select` on a denied lookup asks for a bare foreign key; `$expand` asks the
server to resolve it and hand back the related record — the larger of the two requests was
the ungated one.

## ⭐ `DetailView` is not what the card says it is, and the difference matters

The card lists it as "passes a column list, ungated". Measured on `main`, it is **input-gated**:
`allFields` is collected from `schema`, which is `gatedSchema` — already FLS-filtered field by
field, and whose own comment names "$expand build" among the uses it means to protect.

So this site is a live instance of exactly the route PR #7229 measured as unsound and rejected.
The consequence is not "a denied lookup slips through in the ordinary case" — it is worse and
narrower:

> `buildExpandFields` reads an **empty** column list as "no column restriction" and falls back
> to every declared relation.

⇒ Filtering the input **WIDENS** the request precisely where the principal may read least. A
detail view whose authored fields are all denied had its column list gated down to `[]` and its
`$expand` widened from the relations it asked for to **every relation the object declares**. The
same widening is reached with no authored field list at all, where the input filter has nothing
to remove and the expansion is maximal from the start. Both are pinned.

The input filter stays — it is load-bearing for the **render** half — but it is no longer what
decides the projection.

## Reachability and grading — stated narrowly

- **Reachable in an ordinary configuration, on the client-request side.** Three of the five need
  no authored column list at all, which is their default shape.
- **Against ObjectStack's own server this is defence-in-depth, not a live disclosure**, and for a
  mechanism read rather than assumed: `plugin-security`'s `FieldMasker.maskRecord` does
  `delete result[field]` on every unreadable key and objectql's expand path writes the resolved
  record back under **that same key**, so one statement removes the expanded object and the bare
  id alike; the expansion sub-read itself takes the referenced object's full CRUD + RLS + FLS
  treatment (objectstack#7626). Same grading objectui#6898 and #7215 were given, for the same
  reason, and it becomes load-bearing for any backend that does not strip.
- **Not established:** that any deployed non-enforcing backend is in use.

## The fix — #7229's shape copied, not re-derived

The gate is on each helper's **OUTPUT** at every site. On the three no-column-list sites the
input-side alternative is not merely unsound but unreachable: the call passes `undefined`, so
there is no input to gate. Gating the output also satisfies the ordering requirement
**structurally** rather than by convention — `buildExpandFields` returns a subset of the object's
**declared** reference-bearing fields, and `computeLookupExpand` resolves both of its arms
through the object schema's own field map, so every name either gate judges is declared by
construction. The "`checkField` answers false for an undeclared key" trap is therefore
unreachable and a derived / host-joined column is never judged. Neither helper is changed.

An unanswered policy filters nothing (`isLoaded: false` is the no-provider default, forever), and
`perms` is in each site's dependency list, so the projection is rebuilt the moment the answer lands.

**One structural note in `RecordDetailView`, load-bearing rather than cosmetic:** `usePermissions()`
moved ~670 lines **up**, above the record-load effect. An effect's dependency array is evaluated
*during* render, so listing `perms` there while the binding was still declared below would throw
`Cannot access 'perms' before initialization` — a crash, not a stale value. The later site now
destructures that one value instead of calling the hook again, so the hook order is unchanged in
shape. Same lesson PR #7229 recorded for `ListView`'s memo.

`@object-ui/permissions` is added as a dependency of `plugin-calendar`, `plugin-gantt` and
`plugin-dashboard` (the other two already had it). `check:phantom-deps` passes.

## ⚠️ The eager-closure constraint — measured, and it does not bind

The card required this be measured rather than assumed, and required the `app-shell` site to be
dropped if it put the `framework` chunk over its ceiling. **It does not.** Two full console
builds in the same container, `main` (`bf244f400`) versus this branch's final commit `96a284fe2`,
read from `apps/console/dist/eager-closure.json` (exact gzip bytes, not the gate's rounded KB):

| chunk | before (`bf244f400`) | after (`96a284fe2`) | delta |
|---|---:|---:|---:|
| **`framework`** | **523,823 B** | **523,823 B** | **0** |
| `vendor-objectstack` | 948,329 B | 948,329 B | 0 |
| `ui-components` | 396,598 B | 396,598 B | 0 |
| aggregate eager closure | 3,254,545 B | 3,254,604 B | **+59 B** |
| eager chunk count | 48 of 516 | 48 of 516 | 0 |

`framework` did not move by one byte, so its 177 B of headroom against the 524,000 B ceiling is
untouched and #7399 is not blocked by this PR. The emitted chunks confirm the mechanism rather
than a source-level guess: all three chunks above kept their **content hash**
(`framework-Cf7M9oOC.js`, `ui-components-B7CUYYAi.js`, `vendor-objectstack-CAAM4J8I.js`) — they
are byte-identical files.

**Why the `app-shell` site was never a `framework` risk.** The `framework` group in
`apps/console/vite.config.ts` is `/packages[\\/](core|react|types)[\\/]/` — `app-shell` is not a
member and cannot route there. Read off the emitted graph, the whole +59 B is attributed:

```
assets/RecordDetailView-*.js   35,315 -> 35,346   +31   (its own EAGER chunk)
assets/plugins-views-*.js      71,833 -> 71,858   +25   (plugin-detail)
assets/index-*.js             146,915 -> 146,917   +2
assets/InterfaceListPage-*.js   3,626 ->  3,627    +1
```

`plugin-calendar`, `plugin-gantt` and `plugin-dashboard` are **not in the eager closure at all** —
0 bytes, confirmed by their absence from the report's file list rather than by assuming "this
package is lazy". The aggregate has 13,396 B of headroom and `check:eager-closure` passes on both
builds.

⛔ No ceiling was raised, no baseline re-pinned, no gate weakened.

## Tests — reproduced first, one file per site

New, all mirroring `plugin-grid/src/__tests__/expandFls-7215.test.tsx`:

| file | pre-fix | post-fix |
|---|---|---|
| `plugin-calendar/src/__tests__/ObjectCalendar.expandFls-7230.test.tsx` | 4 failed \| 2 passed (6) | 6 passed |
| `plugin-gantt/src/ObjectGantt.expandFls-7230.test.tsx` | 4 failed \| 2 passed (6) | 6 passed |
| `app-shell/src/views/RecordDetailView.expandFls-7230.test.tsx` | 4 failed \| 2 passed (6) | 6 passed |
| `plugin-detail/src/__tests__/DetailView.expandFls-7230.test.tsx` | **2 failed \| 4 passed (6)** | 6 passed |
| `plugin-dashboard/src/__tests__/ObjectDataTable.expandFls-7230.test.tsx` | 5 failed \| 2 passed (7) | 7 passed |

**Which pins discriminate, said plainly.** `DetailView`'s split is the interesting one and was
predicted before the run: its input filter already delivers the ordinary case, so PIN 1 (denied
lookup among readable authored fields) and PIN 2 (the readable-lookup control) are **green in both
directions** and are labelled as such in the file. The two reds are the widening — every authored
field denied, and no authored field list at all — which is what this card actually buys at that
site. A green pin proves nothing on its own; naming the non-discriminating ones is the discipline
`RecordDetailView.sectionHeadingsRenderPath-6190.test.tsx` records.

Every file carries the live controls, not just the reds: a **permitted** lookup still expands (a
gate that emptied the expansion would paint raw foreign-key ids — the failure objectui#6453 fixed
on the calendar); `master_detail` is pinned beside `lookup`; an **undeclared** derived column is
untouched and does not take the expansion down with it; an **unanswered** policy filters nothing;
and the all-denied case yields no `$expand` rather than a widened one. Each harness waits on a
**real recorded call** — and, where the schema arrives asynchronously, on the *schema-dependent*
query specifically — so "the component stopped fetching" times out instead of reading as an empty
expansion.

## Gates run locally, on the final commit `96a284fe2`

| gate | result |
|---|---|
| `vitest` — `plugin-calendar`, `plugin-gantt`, `plugin-dashboard`, `plugin-detail` (full packages) | **293 files, 2565 tests passed** |
| `vitest` — all 24 `RecordDetailView.*.test.tsx` | **24 files, 174 tests passed** |
| `vitest` — all 19 other `app-shell` tests naming `RecordDetailView` | **19 files, 258 tests passed** |
| `type-check` — all 5 packages (`tsc --noEmit && tsc -p tsconfig.test.json`) | **exit 0**, scope 5 of 47 |
| `turbo run lint` — all 5 packages in full | **exit 0** — 0 errors (2885 pre-existing warnings) |
| `check:eager-closure` | ✅ on both builds (figures above) |
| `check:phantom-deps` | ✅ every in-scope import declared by its publishing package |
| `check:control-bytes` | ✅ 6116 tracked text files |
| `check:vi-mock-specifiers` / `check:vi-mock-inherit` | ✅ / ✅ |
| `check:self-import` / `check:side-effects-array` / `check:sdui-registration-pins` | ✅ / ✅ / ✅ |
| `check-changeset-presence` / `check-changeset-no-major` | ✅ / ✅ |

**Declared narrowing.** `app-shell` carries 608 test files; the four plugin packages alone took
9m of a ~10m foreground ceiling, so the full `app-shell` suite was not run locally — the 43 files
above are every test naming `RecordDetailView`, which is the complete set this diff can reach in
that package. Lint was **not** narrowed: `eslint .` ran in full for all five packages, and the
config is **not type-aware** (no `parserOptions.project` / `projectService`), so this diff cannot
move the verdict of any file it does not itself contain. CI runs the whole farm regardless.

## Out of scope, filed rather than fixed

The card states `buildExpandFields` "is called from five more places". Measured on current `main`,
it is called from **eleven** production places outside #7229's two. Seven remain ungated after this
PR — kanban, tree, `ObjectView`, map, gallery, timeline and the metadata-admin `PagePreview` — and
**all seven pass no column list**, i.e. the same sharp shape as the three sharpest here. Filed
separately with the per-site measurements; not touched here, because they are outside this card's
declared file surface.

---

### DIFF
--- a/.changeset/7230-expand-fls-gate-five-sites.md
+++ b/.changeset/7230-expand-fls-gate-five-sites.md
@@ -1,0 +1,45 @@
+---
+'@object-ui/plugin-calendar': patch
+'@object-ui/plugin-gantt': patch
+'@object-ui/plugin-detail': patch
+'@object-ui/plugin-dashboard': patch
+'@object-ui/app-shell': patch
+---
+
+FLS-gate the `$expand` projection at the five remaining build sites (objectui#7230).
+
+objectui#7215 / PR #7229 gated `$expand` at the two projection sites in its scope
+(`ObjectGrid`, `ListView`). The helper is reached from more places than that. This
+closes the five that were left: `ObjectCalendar`, `ObjectGantt`, `RecordDetailView`,
+`DetailView`, and `ObjectDataTable` (which builds its own whitelist in
+`computeLookupExpand` rather than calling `buildExpandFields`).
+
+**Three of them pass no column list at all**, which makes them the sharp ones:
+`buildExpandFields` reads an absent column list as "no column restriction" and falls
+back to **every declared relation on the object**, denied ones included. So a standalone
+calendar, a gantt, and every record page in the console asked the server to resolve the
+object's full relation set by default rather than by configuration.
+
+**`DetailView` was input-gated, and that is the defect rather than the fix.** Its column
+list is already FLS-filtered field by field, which is exactly the route PR #7229 measured
+as unsound: an emptied column list reads as "no column restriction", so a detail view
+whose authored fields are all denied had its `$expand` **widened** from the relations it
+asked for to every relation the object declares. The principal who may read least was
+asking for the most.
+
+**Reproduced before it was fixed**, as a failing test per site.
+
+**Grading, measured rather than assumed.** Against ObjectStack's own server this is
+defence-in-depth, exactly as objectui#6898 and #7215 are: `plugin-security`'s
+`FieldMasker.maskRecord` deletes every unreadable key from each returned row and
+objectql's expand path writes the resolved record back under that same key, so one
+statement removes the expanded object and the bare id alike; the expansion sub-read is
+itself gated (`__expandRead` takes the referenced object's full CRUD + RLS + FLS
+treatment). It is load-bearing for any backend that does not strip, and the
+client-request side is real regardless.
+
+**Nothing a permitted view did stops working.** The gate judges each helper's OUTPUT,
+which contains only the object's declared reference-bearing fields, so the "`checkField`
+answers false for an undeclared key" trap cannot be reached and derived / host-joined
+columns are untouched. An unanswered permission policy filters nothing. Neither
+`buildExpandFields` nor `computeLookupExpand` is changed.
--- a/packages/app-shell/src/views/RecordDetailView.expandFls-7230.test.tsx
+++ b/packages/app-shell/src/views/RecordDetailView.expandFls-7230.test.tsx
@@ -1,0 +1,279 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7230 — field-level security on the record page's `$expand`.
+ *
+ * ## The site
+ *
+ * `RecordDetailView` loads the record backing an assigned or synthesized page
+ * with
+ *
+ *     const expandFields = buildExpandFields(objectDef?.fields);
+ *
+ * — **no column list**. `buildExpandFields` reads an absent column list as "no
+ * column restriction" and falls back to **every declared relation on the
+ * object**, denied ones included. So every record page in the console asks the
+ * server to resolve the object's full relation set, by default rather than by
+ * configuration. objectui#7215 / PR #7229 gated the two projection sites in its
+ * scope; this call site was outside it.
+ *
+ * ## Grading — the same reading #7215 recorded, not a stronger claim
+ *
+ * Against ObjectStack's own server this is defence-in-depth, not a live
+ * disclosure: `plugin-security`'s `FieldMasker.maskRecord` does
+ * `delete result[field]` on every unreadable key and objectql writes the
+ * expanded record back under THAT SAME KEY, so one statement removes the
+ * expanded object and the bare id alike; the expansion sub-read takes the
+ * referenced object's full CRUD + RLS + FLS treatment (objectstack#7626). It is
+ * load-bearing for a backend that does not strip, and the client-request side
+ * is real either way.
+ *
+ * ## The gate is on the OUTPUT — copied from #7229
+ *
+ * There is no input to gate (the call passes `undefined`), and the output
+ * contains only DECLARED reference-bearing fields, so the "`checkField` answers
+ * false for an undeclared key" trap is structurally unreachable.
+ *
+ * ⚠️ One structural note that is load-bearing rather than cosmetic: this
+ * component read `usePermissions()` ~670 lines BELOW this effect. The effect's
+ * dependency array is evaluated DURING render, so listing `perms` there while
+ * the binding was still declared below would throw
+ * `Cannot access 'perms' before initialization` — a crash, not a stale value.
+ * The hook call moved above the effect; the later destructure now reads that
+ * one value instead of calling the hook again. Same lesson PR #7229 recorded
+ * for `ListView`'s memo.
+ *
+ * The stub `checkField` is an ALLOWLIST, per `expandFls-7215.test.tsx`: the
+ * real provider answers `true` for any field no policy mentions.
+ */
+import * as React from 'react';
+import { describe, it, expect, vi, beforeEach, beforeAll, afterEach } from 'vitest';
+import { render, waitFor, cleanup } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { MetadataCtx } from '@object-ui/react';
+
+/** Stable stub identity — `perms` rides the record-load effect's dependency list. */
+const { permsStub, state } = vi.hoisted(() => {
+  const state: { isLoaded: boolean; readable: string[] } = { isLoaded: true, readable: [] };
+  return {
+    state,
+    permsStub: {
+      get isLoaded() { return state.isLoaded; },
+      checkField: (_object: string, field: string, action: string) =>
+        action === 'read' ? state.readable.includes(field) : true,
+      check: () => ({ allowed: true }),
+      getFieldPermissions: () => [],
+      getRowFilter: () => undefined,
+      getObjectApiOperations: () => undefined,
+      roles: [],
+      userId: null,
+      systemPermissions: undefined,
+      hasCapabilities: () => true,
+      can: () => true,
+      cannot: () => false,
+    },
+  };
+});
+
+vi.mock('@object-ui/permissions', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@object-ui/permissions')>();
+  return { ...actual, usePermissions: () => permsStub as any };
+});
+
+vi.mock('@object-ui/auth', () => ({
+  useAuth: () => ({ user: { id: 'u1', name: 'Ada', image: null }, activeOrganization: null }),
+  createAuthenticatedFetch: () => vi.fn(),
+}));
+vi.mock('@object-ui/collaboration', () => ({
+  useRecordPresence: () => ({ viewers: [], others: [] }),
+  PresenceAvatars: () => null,
+}));
+vi.mock('sonner', () => ({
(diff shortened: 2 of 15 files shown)


os-project-manager:
The out-of-scope finding named at the end of the description is filed as #7429 — the seven `buildExpandFields` call sites that stay ungated after this PR (kanban, tree, `ObjectView`, map, gallery, timeline, metadata-admin `PagePreview`), with the per-site measurements and a triage sizing note. Unassigned; not touched here.

Two notes on the description itself, recorded rather than silently corrected:

- It reads "The fix — #7229's shape copied". Closing keywords are kept away from every card number except the intended `Fixes #7230`; that adjacency is inert because GitHub's closing-keyword parser only auto-closes **issues**, and #7229 is a merged pull request.
- The card's line numbers for three of the five sites are off by one to five against current `main` (`350`→`351`, `674`→`675`, `680`→`675`). Same calls; verified before editing, per the rule that the issue body is a lead rather than a spec.

---

PR #7400, opened 2026-09-02
fix(plugin-charts): refuse a second scatter series instead of drawing it at the first one's y

os-project-manager:
Fixes #7194

Maintainer ruling **B** (comment 5507673384, 2026-09-02): a `chartType: 'scatter'` with two or more `series` renders the file's existing refusal shell, stating that a scatter binds one measure and naming the authored fix; a single-series scatter is unchanged; no multi-measure projection (option A waits for a real caller); render-time only — the authoring-time (zod) rejection follows #7113 when that card rules. Clause-② yes ⇒ this PR is a **draft**; it is not flipped ready and not enqueued.

## What changed

- `packages/plugin-charts/src/AdvancedChartImpl.tsx` — the scatter arm checks `series.length > 1` **first** (a binding fault is true of the spec whatever the rows carry, so it outranks the #7171 positional refusal the same way the missing-category-key guard outranks the series guard) and renders `SeriesArityRefusal`: the existing `ChartRefusal` shell under `data-chart-error="scatter-multi-series"`, `role="status"`, no console warning (matching the three scatter/magnitude answers already in the file). The shell fits without modification. The sentence is read from `chart.scatterOneMeasure` through `useSafeTranslate` — the hook `ObjectChart` already imports from `@object-ui/i18n` — so no module-scope `createSafeTranslation` map is added and no partial i18n mock is disturbed (census: 91 tests mock `@object-ui/i18n`; none of the partial ones reaches this renderer). The series keys are rendered as data after the sentence, as code elements, outside the translated copy.
- `packages/i18n/src/locales/{en,zh,ja,ko,de,fr,es,pt,ru,ar}.ts` — one key, `chart.scatterOneMeasure`. en: `A scatter plots one measure. Keep exactly one series:` — states one measure, names the fix; the keys follow as data.
- `packages/plugin-charts/src/AdvancedChartImpl.scatterSeriesArity.test.tsx` — 12 pins: single-series control draws 2 marks / no refusal / no note; `series: []` and no `series` still reach the chart through the `value` fallback; two series ⇒ refusal, no chart container, 0 symbols, no legend; copy contains the sentence and lists both keys; three series refuse; the binding refusal wins over `no-plottable-points`; holds with zero rows; no console warning; a two-series `bar` still draws both series; `en` pack value equals the inline fallback; every pack defines it and none copies the English; under an `I18nProvider` (`zh`) the pack value renders and the keys stay untranslated.
[...]

PR #7442, opened 2026-09-03
fix(core,fields): render a dataset measure over a date field as a date

### LINKED ISSUE #7178: A dataset measure over a date field renders its raw ISO string — `formatMeasure` short-circuits before `format` is read
Found by opening a dashboard in a real ObjectStack application (`objectstack-ai/duly`) against published `@objectstack/*` 17.2.0. Source-confirmed afterwards in this repo.

## The claim

**A `min` / `max` measure over a date or datetime field cannot be rendered as a date on any dataset-bound widget, and the `format` escape hatch the spec offers for exactly this is unreachable for those values.**

The KPI tile renders `2026-07-04T07:00:00.000Z` where a date belongs — a 24-character ISO string in `text-2xl font-semibold`, which wraps to two lines inside the card.

## Where it comes from

`packages/plugin-dashboard/src/DatasetWidget.tsx:768` renders the dataset-bound metric value:

```tsx
<span className={cn('text-2xl font-semibold tabular-nums', accentClass)}>
  {formatMeasure(value, f?.format, f?.currency, f?.percentScale, displayLocale)}
</span>
```

and `formatMeasure` in `packages/core/src/utils/dataset-format.ts` opens with:

```ts
if (v == null) return '—';
if (typeof v !== 'number') return String(v);
```

That second line is before `format` is touched. So for any non-numeric measure value the function is `String(v)` and nothing else — the `format` argument the call site carefully threads through is dead on this path. `tabular-nums` on the span is the same assumption stated in CSS.

Everything downstream of that guard is a numeral-pattern formatter (decimals from `format.split('.')[1]`, `%`, `$`, `currency`), which no date pattern can enter.

## Why the escape hatch does not close it

`DatasetMeasureSchema` in `@objectstack/spec` carries `format: z.ZodOptional<z.ZodString>` — an open string, so authoring `format: 'YYYY-MM-DD'` (or anything else) passes `validate` and `build` cleanly. It then has no effect, because the value never reaches the branch that reads it. An author gets a green toolchain and an unchanged raw timestamp, with nothing anywhere saying why.

## Reproduce

Declare a measure aggregating a date field and bind it to a `type: 'metric'` widget:

```ts
// dataset
{ name: 'oldest_touch', label: 'Oldest touch', aggregate: 'min', field: 'last_update_at' }

// dashboard widget
{ id: 'oldest', title: 'Oldest untouched task', type: 'metric',
  dataset: '<ds>', values: ['oldest_touch'] }
```

Open the dashboard. The tile shows the raw ISO string. Adding `format: 'YYYY-MM-DD'` to the measure changes nothing.

## Scope beyond the metric tile

The same `formatMeasure` call is on three other surfaces, all with the same guard in front of them, so a date measure is raw in each: `DatasetWidget.tsx:876` (chart value formatting), `DatasetWidget.tsx:1083` (dataset table cells), and `views/metadata-admin/previews/DatasetPreview.tsx:235`. `PivotTable.tsx:46` takes `value: number` outright.

## Suggested direction

The narrow fix is to let `formatMeasure` recognise a date-shaped string (or a measure whose `format` names a date pattern) before the numeric short-circuit, and route it through the same date display path list cells already use — the argument for reusing that path rather than adding a second one is the same one the file's own header makes about percent, where two conventions drifted between a list cell and a dashboard measure (objectui#4576).

Whatever the resolution, the silence is the worse half: a `format` that the schema accepts and the renderer can never read is `declared ≠ enforced`. If dates are out of scope for measure formatting, the spec's `format` should say so and a date-typed aggregate should be refused at author time rather than rendered raw.

## Application impact

Low severity — the number is correct and legible, just ugly. Reported because the *silence* is the reusable finding, not the wrapped timestamp.

os-project-manager:
Fixes #7178

Implements the director-seat ruling in [comment 5509025252](https://github.com/objectstack-ai/objectui/issues/7178#issuecomment-5509025252) (summon #8, 2026-09-02, maintainer verbatim `7189 A  其他同意`) — **option A**. `min` / `max` over a date or datetime field stays a legal measure; nothing in `@objectstack/spec` is touched, and `DatasetMeasureSchema` is not narrowed.

## The defect

`formatMeasure` in `packages/core/src/utils/dataset-format.ts` opened with:

```ts
if (v == null) return '—';
if (typeof v !== 'number') return String(v);
```

The second line runs **before** `format` is read, so for any non-numeric value the function was `String(v)` and nothing else. A `min` over a datetime rendered as a raw 24-character ISO string in the KPI tile's `text-2xl font-semibold`, wrapping to two lines, and the `format` the four call sites carefully thread through was dead on this path.

A date-shaped value now routes to the date display path ahead of that short-circuit. Because the fix lands in `formatMeasure`, all four call sites are served at once — the metric tile (`DatasetWidget.tsx:768`), chart values (`:876`), dataset table cells (`:1083`) and the metadata-admin dataset preview (`DatasetPreview.tsx:235`). None of them changed. `PivotTable.tsx` takes a `number` outright and is untouched.

## ⭐ The shared date path — what I found, and what reusing it cost

**File + symbol: `formatDate` / `formatDateTime` / `formatRelativeDate`, previously in `packages/fields/src/index.tsx` (`@object-ui/fields`), now in `packages/core/src/utils/date-display.ts`.**

That is the path a list cell uses: `DateCellRenderer` (registered for field type `date` at `packages/fields/src/index.tsx`) calls `formatDate(value, field.format || 'relative', { dueLike, locale, t })`, and `ObjectGrid`'s date cells, `ObjectGantt`'s tooltips and `plugin-dashboard`'s `recordFields` all call the same two functions.

Reusing it required **lifting it**, and that is the substance of this PR. `@object-ui/fields` is a React package and depends on `@object-ui/core`; `core` is the React-free engine, so it could not import from it. That is *exactly* the objectui#4576 situation one type over — there, `formatDisplayNumber` lived in `@object-ui/i18n`, `dataset-format.ts` could not reach it, the two implementations drifted, and a German session read `1.234,5 %` from a list cell beside `1.234,5%` from a dashboard measure. #4576 removed the boundary by moving the pure function **down** into `core` and re-exporting from above.

This PR applies that same remedy:

- `DateDisplayOptions`, `formatRelativeDays`, `formatRelativeDate`, `formatDate` and `formatDateTime` move verbatim into `packages/core/src/utils/date-display.ts`. No logic changed.
- `@object-ui/fields` re-exports all four public names, so **no consumer's import path or behaviour changes** (`ObjectGrid`, `ObjectGantt`, `plugin-dashboard`, `@object-ui/react`).
- `packages/fields/src/__tests__/date-display.reexport-identity.test.ts` pins **reference identity** — the cell renderer's `formatDate` and the measure formatter's are the same function object. A second copy would pass every behavioural assertion in the repo and fail only there.

⛔ No date formatter was written in `dataset-format.ts`. The new code there is a 12-line **route**: two anchored ISO shape tests, a `Date.parse` guard, and a call.

The date/datetime split mirrors `ObjectGantt.tsx`'s sniffed-ISO dispatch verbatim, because that site is already the repo's answer to this exact situation — an ISO string with no field type attached, which is what a measure is by construction. That is why the absolute locale form is used rather than `DateCellRenderer`'s `|| 'relative'` default, which is keyed on knowing the value is a `date` **field** in a row.

## ⭐ The `format` reachability answer — measured

**The shared date path accepts `format` as a named STYLE, not as a date pattern.**

`formatDate(value, style, options)`'s vocabulary is `'short'` and `'relative'`; every other string falls to its default locale-medium branch. So, for a date measure:

| `format` | reachable? | renders |
| --- | --- | --- |
| `'short'` | ✅ yes | `Jul 4, '24` — the same word `DateCellRenderer` honours from `field.format` |
| `'relative'` | ✅ yes | `3 days ago` |
| `'YYYY-MM-DD'` | ❌ no | the locale default (`Jul 4, 2024`) |
| absent | n/a | the locale default |

Threading the measure's `format` into that parameter is **the same mapping the list cell already makes**, not a new one — and it is not novel behaviour for a pattern either: `plugin-dashboard`'s `recordFields.tsx` already routes a date-shaped `format` (`/[YMDHms]/`) into the same style slot and gets the same locale default.

⇒ Per the ruling's contingency, the remaining gap is the **pattern** case, not the whole of `format`. If the spec seat files the one-line `describe()` follow-up, "applies to numeric measures; date measures render by locale" would be slightly too strong — `'short'` and `'relative'` *are* read. Suggested wording: *"numeral patterns apply to numeric measures; a date measure reads `format` as a display style (`short`, `relative`)."* That follow-up is the `domain:ui` seat's to file — it is not in this PR.

The limit is pinned by a test rather than left silent, so teaching the shared path a pattern grammar later has a case that must be updated deliberately.

## Tests

`pnpm exec vitest run` from the repo root (⚠️ a package-directory vitest run is refused by the repo's own guard, objectui#3378).

- `packages/core/src/utils/__tests__/dataset-format.date.test.ts` — **13 passed**
- `packages/plugin-dashboard/src/__tests__/DatasetWidget.dateMeasure.test.tsx` — **5 passed** (the ruling's pin: metric tile + dataset table)
- `packages/fields/src/__tests__/date-display.reexport-identity.test.ts` — **2 passed**

Every expected date string in the rendered-surface tests is **derived** by calling `formatDateTime` from `@object-ui/fields` — the module the cell renderers call, by the path they import it by — so the cases assert "the tile shows what a list cell would show", not "the tile shows the string I typed". A literal would pass just as well against a second date convention.

### ⭐ A numeric measure is byte-identical

A verbatim `git show` copy of the pre-fix function was compared against the new one across the full argument matrix (value × `format` × `currency` × `percentScale` × `locale`):

```
#7178 MATRIX: compared=33696 forms; distinct values that moved=4
(2026-02-30, 2026-07-04, 2026-07-04T07:00:00.000Z, 2026-07-04 07:00:00);
moved forms=3240; expected movers that did NOT move=0
```

**33,696 forms compared; the only values that moved were the four ISO-shaped, parseable ones.** Numbers, numeric strings (`'1751612400000'`, `'2026'`, `'1e21'`), the nullish em dash, arbitrary prose (`'Acme Corp'`, `'March 5, 2026'`, `'2026/07/04'`) and non-strings are all byte-identical.

Two findings from that sweep, both now pinned:

- `Date.parse('2026-02-30')` is **not** `NaN` — V8 rolls it to March 2. It routes to the date path and renders `Mar 2`, which is what a list cell shows for the same stored string. That is agreement, not a defect; rejecting it *here* is what would create a second convention.
- An ISO-*shaped* but unparseable value (`'2026-13-45'`, `'2024-07-04T99:99'`) keeps falling through to `String(v)`. Routing it to `formatDate` would replace a raw string an author can still debug with an em dash that says nothing.

### Red-then-green ablation

Two mutations, each proven on disk by blob hash + anchored grep counts before the run, each restored with `git checkout HEAD -- ABSOLUTE_PATH` under `trap ... EXIT INT TERM`, and each restore proven **by state** (`git diff HEAD` empty **and** blob hash equal to the `HEAD` blob), never by exit code. Directions were written down before the run and all four matched.

Vitest aliases every workspace package to `src` (`vitest.config.mts`), so these legs read source directly — no rebuild leg applies, and none is claimed.

| leg | mutation | result |
| --- | --- | --- |
| 1 | restore the pre-fix `return String(v)` short-circuit | core `8 failed / 5 passed`, dashboard `3 failed / 2 passed` — every date and `format` case red; ⭐ **every numeric must-not-change case stayed green** |
| 2 | replace the re-export with a behaviour-identical local wrapper | identity `1 failed / 1 passed` — identity red, behaviour green |

Leg 2 is the one that matters for the ruling's constraint: it is a "second date formatter" that behaves *identically*, and it is caught only by the identity pin. That is the drift objectui#4576 paid for once.

### Also run

`type-check` and `lint` clean (exit 0) on `@object-ui/core`, `@object-ui/fields`, `@object-ui/plugin-dashboard`, after building each dependency closure.

## Bundle budget — `framework` chunk

⚠️ **This PR puts `framework` 402 bytes over its ceiling. The `Bundle Analysis` check will be red.** Reported, not worked around — and ⛔ no ceiling, budget script or `PER_CHUNK_*` entry is touched in this PR.

Measured with CI's own recipe (`pnpm turbo run build --filter='./packages/*'` → `pnpm --filter @object-ui/console build` → `node scripts/check-eager-closure-budget.mjs`) on both legs: this branch at `5ad78fde`, and the branch point `6411def25` as the baseline. Gzipped bytes, read from `apps/console/dist/eager-closure.json`:

| chunk | baseline | this branch | delta |
| --- | ---: | ---: | ---: |
| `framework` | 523,823 | **524,402** | **+579** |
| `ui-components` | 396,598 | 396,088 | **−510** |
| `vendor-objectstack` | 948,329 | 948,329 | 0 |
| **aggregate eager closure** | 3,254,604 | **3,254,620** | **+16** |

Verdicts: baseline `exit 0` (green), this branch `exit 1` —

```
❌ framework   512.1 KB / 511.7 KB ceiling (OVER by 0.4 KB)
✅ aggregate   3178.3 KB / 3191.4 KB ceiling (headroom 13.1 KB)
```

**What the bytes are.** They are not new code. This change is a *relocation*: the date formatter family leaves `packages/fields` (`ui-components`, −510 B) and lands in `packages/core` (`framework`, +579 B), because `framework` is `/packages/(core|react|types)/` by path. The **net cost to the payload a browser actually fetches is +16 bytes** across a 3.25 MB eager closure — the two chunk rows nearly cancel, and the aggregate ceiling keeps 13.1 KB of headroom.

So the red is chunk *attribution*, not payload growth: `framework` had 177 bytes of headroom before this branch, and moving a shared function into the package that group is defined by consumes it. The relocation is the ruled design — `core` must own the date path because it is the React-free engine that `formatMeasure` lives in — so shrinking it back means either a second date formatter in `dataset-format.ts` (⛔ forbidden by the ruling and by objectui#4576) or leaving the defect unfixed.

⇒ This lands on the same blocked ceiling as #7391 and #7400, pending the maintainer's decision in #7399. Routing that is the PM's call, not this PR's.

## Changeset

`@object-ui/core: patch` **and** `@object-ui/fields: patch`. The ruling named a `@object-ui/core` patch; `@object-ui/fields` is added because the lift genuinely changes that package's source and its dependency on `core` for those symbols. Flagging it as a deliberate, minimal deviation rather than a silent one.

---

---


os-project-manager:
## Blocked, deliberately: `Bundle Analysis` red on a governance decision this seat cannot make

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Stating the blocker here on the PR rather than only on the card, so a reviewer landing on this page sees why it is red and why nothing is being pushed.

**Failing check:** `Bundle Analysis` (`100528606288`) on `5ad78fde`. Every other check is green or skipped.

```
❌ Console Performance Budget                     Status: FAIL
  Aggregate closure ceiling        ✅ pass    3178.3 KB / 3191.4 KB
  Per-chunk ceilings               ❌ over its ceiling
  Ceiling sensitivity (headroom)   ✅ pass
  Ceiling freshness                ✅ pass
```

| chunk | before | after | ceiling |
|---|---:|---:|---:|
| `framework` | 523,823 | **524,402** (+579) | 524,000 — **over by 402** |
| `ui-components` | 396,598 | 396,088 (−510) | 399,000 |
| aggregate eager closure | 3,254,604 | 3,254,620 (**+16**) | passes |

### It is this PR's byte cost, and it is still not this PR's to fix

The failure is real and attributable — I am not claiming otherwise. But the bytes are a **relocation, not growth**: the `framework` group is defined by path (`apps/console/vite.config.ts:742`, `/packages/(core|react|types)/`), so code moving *into* `packages/core` scores against it even though an equal amount left `ui-components`. This PR's true payload cost is **+16 bytes across a 3.25 MB closure**, and the aggregate ceiling passes. CI's own package table shows both halves: `fields` 61.71 → 61.26 KB gz, `core` 2.32 → 2.42 KB gz.

The relocation is not optional. The ruling on #7178 requires routing through the date path list cells already use and forbids a second date convention; `packages/core` is the React-free engine and cannot import from `packages/fields`, so the shared `formatDate` / `formatDateTime` had to move down — the same move objectui#4576 made for `formatDisplayNumber`.

### What I am NOT doing, and why

- ⛔ **Not raising `PER_CHUNK_GZIP_CEILINGS.framework`.** A gate ceiling is the maintainer's call and is the open subject of #7399. Raising it to fit my own PR is the one thing that card exists to prevent.
- ⛔ **Not shrinking the change to fit.** That means keeping a date convention inside `dataset-format.ts` — the second-convention failure the ruling and #4576 both forbid. It would trade a governance blocker for exactly the defect this card exists to fix.
- ⛔ **Not re-running the check.** This is a deterministic budget computation, not a flake: it reproduces on two full local builds with CI's own recipe and on CI. A re-run would return the identical number and burn a runner for nothing. The re-run allowance exists for a failure that might not reproduce; this one will.
- ⛔ **No test skipped, disabled or quarantined.** No baseline re-pinned, no budget script touched.

### Where it goes

Parked behind **#7399** (the open `framework` ceiling decision), alongside #7391 and #7400 — this is now the third PR blocked there. New evidence posted to #7399 at [5520783051](https://github.com/objectstack-ai/objectui/issues/7399#issuecomment-5520783051): this PR is the first case where the gate reds on a change whose aggregate cost is +16 bytes, which is a property of how the per-chunk ceiling *attributes* bytes rather than of the change.

Full review of the implementation — including the one place the ruling's pin is **not literally satisfiable** for `datetime`, and why option A was taken anyway — is on the card at [5520776537](https://github.com/objectstack-ai/objectui/issues/7178#issuecomment-5520776537). The PR stays **draft**; auto-merge is not enabled and it has not been enqueued.

---

os-project-manager:
## ✅ Unblocked — the ceiling this was parked behind is fixed. `main` merged in; landing once green.

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Retiring the "Blocked, deliberately" notice above ([5520817103](https://github.com/objectstack-ai/objectui/pull/7442#issuecomment-5520817103)) — its blocker no longer holds.

**#7399 was ruled** (maintainer, 2026-09-03, option **A′**) and landed as `177afeba1` (PR #7489). The `framework` chunk was **78.7% `@object-ui/i18n` locale catalogue**, because `framework` and `infrastructure` tied at priority 80 and `framework` was written first. The catalogue now has its own `i18n-locales` chunk and both ceilings were re-baselined in the same change.

### ⭐ This PR's own analysis was right, and the fix vindicates it

That notice argued the failure was **attribution, not growth**:

> the bytes are a **relocation, not growth** … This PR's true payload cost is **+16 bytes across a 3.25 MB closure**, and the aggregate ceiling passes.

That was correct, and it was correct for a deeper reason than the notice knew: the ceiling it tripped was not budgeting `core|react|types` at all. `framework` now measures ~60 KB against a **71,000** ceiling (baseline 61,465), where before it measured 523,823 against 524,000 — **41 bytes of headroom for the whole repository**. This PR's +579 B fits with room to spare.

⚠️ For the record, and against my own earlier reasoning: I later argued on several PRs that a red here could be diagnosed from the path rule (`/packages/(core|react|types)/`). **That rule is false in this repo** — it names the chunk but not its contents — and I published three wrong causal stories from it before the measurement settled the question. The full retraction is on #7399. This PR's notice reasoned better than I did about its own failure.

### What I did

Merged `main` in so CI rebuilds against the new grouping. ⛔ No ceiling was raised, ⛔ nothing was shrunk to fit, ⛔ no re-run spent — exactly as that notice promised.

⇒ Once every check is green by name, this seat lands it. The implementation review stands on the card ([#7178 comment 5520776537](https://github.com/objectstack-ai/objectui/issues/7178#issuecomment-5520776537)), including the one place the ruling's pin is not literally satisfiable for `datetime` and why option A was taken anyway.

---

os-project-manager:
## ✅ Landing — 32/32 green, ready + auto-merge

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Following the unblock notice above ([5524649818](https://github.com/objectstack-ai/objectui/pull/7442#issuecomment-5524649818)).

### Every check read by name, not in aggregate

**32 total: 27 success, 5 skipped**, on head `5daf008b9`. The five skips are the jobs that do not run on this event (`dependabot`, `Test (coverage)`, `Test (coverage shard …)` and siblings) — ⛔ none is a check that ran and was ignored.

The ones that were the actual question:

| check | reading |
|:--|:--|
| `Bundle Analysis` | ✅ **PASS** — 3180.9 KB / 3191.4 KB, **50 chunks** (was 48) |
| `Test (shard 1/4)` … `4/4` | ✅ all four success (11:06:56 / 11:09:02 / 11:08:57 / 11:08:33) |
| `Type Check` | ✅ success 11:03:13 |
| `Lint` | ✅ success |
| `Governed Surface Queue Guard` | ✅ success |
| `README Export Check` | ✅ success — the check that would object to the lift's new `@object-ui/core` exports |

### The landing authority, stated

Clause-② is **no** on this card, ruled explicitly: 「Execution: `domain:ui` lane, S. **Clause-② no** (renderer side, no declared surface moves).」 ([5509025252](https://github.com/objectstack-ai/objectui/issues/7178#issuecomment-5509025252)).

⚠️ I re-derived that from the diff rather than resting on the ruling's prediction, because the gate's own rule is 「**diff 是事实,卡片语义是预测**」 — and this diff *does* move symbols between packages, which the ruling's "no declared surface moves" did not anticipate. It still reads `no`: the path limb does not fire (nothing under `packages/spec/src/**`, no `*.zod.ts`), and the content limb does not either — no contract accept/reject behaviour changes, `DatasetMeasureSchema` is untouched, and nothing an author may write is newly accepted or refused. The four lifted names land in `@object-ui/core` and are re-exported from `@object-ui/fields`, so no consumer's import path or behaviour moves. `README Export Check` is green over exactly that surface.

⇒ Seat lands it. The implementation review is the ACCEPT on the card ([#7178 comment 5520776537](https://github.com/objectstack-ai/objectui/issues/7178#issuecomment-5520776537)).

### The one flag carried into landing, deliberately

That ACCEPT recorded, and did not paper over, that **the ruling's pin is not literally satisfiable for `datetime`**: `DateTimeCellRenderer` inlines `toLocaleDateString`/`toLocaleTimeString` and never calls `formatDateTime`, which has zero call sites — so there is no shared datetime path to route through. Option A was taken (dispatch to `formatDateTime`, the module's own declared datetime function, sniffing date-vs-datetime the way `ObjectGantt` already does), and the ACCEPT said plainly: *"If the director seat or the maintainer reads the pin's wording as binding on `datetime` specifically, this is the place to object."*

That objection window has been open since **05:06** today and the director seat has been active in this repo since (it ruled #7402 at 08:44). No objection was entered, so the flag lands as recorded rather than silently. The reconciliation itself is filed as #7443 and is **not** closed by this PR.

### Follow-up discharged

The ruling also owed a one-line spec-lane follow-up, *"filed by the `domain:ui` seat when the measurement is in hand"*. The measurement is in hand, and it is filed: **objectstack#14933**.

⭐ It does **not** copy the ruling's suggested sentence. The ruling assumed a date measure cannot read `format` at all; measured, `'short'` and `'relative'` **are** read and only the *pattern* case is unreachable — so "date measures render by locale" would have replaced one inaccuracy with another. The card says so and proposes wording that matches the measurement. It also records that there is no `.describe()` on that field to add a sentence to (`packages/spec/src/ui/dataset.zod.ts:191-192` carries a JSDoc instead), which is a choice the spec lane should make deliberately rather than have made for it.

---

PR #7451, opened 2026-09-03
refactor(i18n): retire `useObjectLabel().viewDescription()` and its `_views.VIEWNAME.description` catalog key

### LINKED ISSUE #7219: `useObjectLabel().viewDescription()` has zero production call sites — the catalog translation path for a list view's description is declared, tested, and never invoked
Observation surfaced while fixing #7199. Filed as a `finding`, not a defect: the correct
resolution is a decision, not a mechanical fix.

## Measured

`packages/i18n/src/useObjectLabel.ts:488` declares:

```ts
viewDescription: (objectName: string, viewName: string, fallback?: string) => {
  const fb = fallback ?? '';
  const resolved = resolve(viewSuffixes(objectName, viewName, 'description'), fb);
  return resolved || undefined;
},
```

Convention `NS.objects.OBJECTNAME._views.VIEWNAME.description`. `git grep viewDescription`
over `packages/`, `apps/` and `examples/` returns, besides the declaration itself, only:

- `packages/i18n/src/__tests__/useObjectLabel-view.test.tsx` (two call sites)
- `packages/app-shell/src/views/ObjectView.defaultViewIdentity.test.tsx` (one call site)
- one `packages/i18n/CHANGELOG.md` entry

**Zero production callers.** Every other hit in that grep is `previewDescription`, an
unrelated import-wizard key.

## Why it stands out

Its two siblings on the same node are both wired up:

- `viewLabel` is called by `app-shell`'s `ObjectView` (breadcrumb / origin label).
- `viewEmptyState` is called from inside the `renderListView` relay itself.

So a translation bundle can currently localize a view's label and its empty state, but a
`_views.VIEWNAME.description` entry authored in the same bundle resolves nowhere.

## Why I did not wire it up under #7199

Two distinct vocabularies carry a view description and their precedence is unsettled:

1. the **authored value** on the view entry, `I18nLabel` — a string or an inline locale
   map, resolved by `pickLocalized` / the spec's `resolveI18nLabel`;
2. the **catalog key** above, resolved through the i18n bundle.

#7199 fixes (1) — relaying the authored value and resolving its locale map. Adding (2)
requires deciding which wins when both exist, which is exactly the "two readings, two
architectures" shape that should not be guessed at inside an unrelated fix. `viewEmptyState`
suggests catalog-over-authored is the house answer for this node, but `emptyState` is not
`I18nLabel`-typed, so it never had to answer the precedence question this key raises.

## Possible dispositions

- **Wire it in**, mirroring `viewEmptyState`'s position in the relay, with a stated
  precedence rule.
- **Retire it** under enforce-or-remove, if the authored `I18nLabel` value is considered the
  single intended channel now that it actually arrives.

Either is defensible; the pin tests would need retargeting for the second.

os-project-manager:
Fixes #7219

Retires `useObjectLabel().viewDescription()` from `@object-ui/i18n` together with the
`NS.objects.OBJECTNAME._views.VIEWNAME.description` catalog convention it resolved.

**This card reads as undecided and is not.** Its body ends with "Possible dispositions"
under a heading saying the resolution is a decision — it was made. Authority:
[comment 5508044124](https://github.com/objectstack-ai/objectui/issues/7219#issuecomment-5508044124)
(director seat, summon #8, decision batch #6, 2026-09-02, maintainer verbatim
「7205 B，其他同意」) — **option B, retire**. Option A (wire the catalog key in and state a
precedence rule) is explicitly not taken; neither is option C.

## The premise, re-measured on this branch point

The card's core claim is zero production callers. Re-measured on `d6ceb8d29`, not carried
over from the card:

| instrument | result |
|---|---|
| `git grep -n viewDescription` over the whole tree | 20 hits: 1 declaration, 3 test call sites, 1 `packages/i18n/CHANGELOG.md` line, 12 unrelated `previewDescription`, **3 in `plugin-list/src/ListView.tsx` (see below)** |
| production callers of `useObjectLabel().viewDescription()` | **0** |
| CONTROL — `git grep -l viewLabel` | **36 files**, so the instrument fires |

## The three `ListView.tsx` hits are NOT the hook member — they are the survivor

`packages/plugin-list/src/ListView.tsx:3106` declares a **local variable** of the same
name, landed by #7199:

```ts
const viewDescription = pickLocalized(schema.description, displayLocale);
```

That is the *authored* `I18nLabel` channel — precisely what the ruling preserves ("the
authored value wins because it is the only reader"). It is untouched by this PR: no file
under `packages/plugin-list/` is in the diff. The card's body claims every non-hook hit is
`previewDescription`; that is stale, and a grep-driven cleanup would walk straight into it.

## What `viewSuffixes` does — measured, not assumed

`viewSuffixes` is **kept unchanged**. It is shared: `viewLabel` passes it `'label'`,
`viewEmptyState` passes it `'emptyState.title'` / `'emptyState.message'`, and only the
retired member passed `'description'`. The suffix construction is therefore not exclusive
to this member; the `'description'` tail is what disappears, not the builder.

## Pins retargeted, not deleted

The ruling's surviving pins are that `viewLabel` and `viewEmptyState` still resolve, and
that a `_views.VIEWNAME.description` catalog entry is **inert**.

An inert pin is the easy one to write vacuously — a test that only asserts the member is
gone is green on any tree where it never existed. Both new cases therefore **author** the
catalog `description`, prove that bundle node is live and reachable (its `label` /
`emptyState` siblings resolve off it — the control that fires), and then show that the
description a consumer resolves is the **authored** value and a **different string**:

- `packages/i18n/src/__tests__/useObjectLabel-view.test.tsx` — resolves the authored value
  through the same `pickLocalized` call `ListView` makes.
- `packages/app-shell/src/views/ObjectView.defaultViewIdentity.test.tsx` — the authored
  `I18nLabel` arrives on the view entry through the real pipeline
  (`expandViewContainer` -> `mergeViewsIntoObjects`) rather than being typed in at the
  assertion, while the served `ZH_PAYLOAD` still carries the catalog `description`.

Both also carry a `@ts-expect-error` on the removed member, so the **return type** is
pinned too, not just the runtime object. That is a real check: `tsconfig.test.json` in each
package compiles these files, proven with `tsc --listFiles` (1 hit for each edited test
file, so neither is silently outside the program).

## A fourth pin the grep does not reach

`packages/i18n/src/__tests__/useObjectLabel-identity-5564.test.tsx` asserts the hook's
member set as an **absolute count**, beside the with-instance / without-instance equality.
Removing a member moves that absolute by one (27 -> 26). It contains no occurrence of the
symbol, so it is invisible to a grep for `viewDescription`; it was found by running the
whole `@object-ui/i18n` suite. The equality assertion next to the count is unaffected,
which is the point — a member leaves both paths at once because there is only one path.

## Reverse verification (ablation) — direction predicted before running

Mutation: put `viewDescription` **back** on the hook — the pre-retirement / option-A world.
Predicted **2 RED** (the two inert pins) **/ 20 GREEN**, including all six #7199 relay pins.

Both legs proved the mutation on disk before measuring anything (blob hash `0834ec2f` ->
`452ebac5`, injected-marker `grep -c` 0 -> 1), and proved the restore **by state**
(`git hash-object` equal to the HEAD blob, `git diff HEAD` empty, marker count back to 0) —
never by an exit code. Both scripts used `trap ... EXIT INT TERM` with absolute paths.

- **Runtime leg:** `Tests 2 failed | 20 passed (22)` — exactly as predicted, both failures
  `expected [ 'objectLabel', …(26) ] to not include 'viewDescription'`, and
  `ObjectView.viewDescriptionRelay-7199.test.tsx` fully green (the authored channel is
  untouched by the mutation).
- **Type leg:** `error TS2578: Unused '@ts-expect-error' directive.` at
  `useObjectLabel-view.test.tsx(200,5)` — the type pin fires as well.

No rebuild was needed for either leg and none is silently missing: `vitest.config.mts:407`
aliases `@object-ui/i18n` to `packages/i18n/src`, and the i18n test imports
`../useObjectLabel` relatively, so both consumers read the mutated **source**. Nothing in
this ablation resolves through the package `exports` to `dist`.

## Verification, all on head `18f24981c`

- `pnpm exec vitest run packages/i18n` plus every repo test referencing `useObjectLabel` —
  **113 files / 1534 tests passed**. That is the full `@object-ui/i18n` suite plus the 35
  `app-shell` and 20 other-package tests that touch the hook.
- `turbo run type-check --filter=@object-ui/i18n --filter=@object-ui/app-shell` — green,
  including each package's `tsc -p tsconfig.test.json` leg.
- `turbo run lint` (repo-wide, the `eslint .` spelling) — **47/47 tasks, 0 errors**.
- Gates: `check-changeset-presence` / `-fixed` / `-no-major` / `-overwrite`,
  `check:control-bytes`, `check:i18n-keys`, `check:i18n-drift`, `check:vi-mock-specifiers`,
  `check:vi-mock-inherit`, `check:published-tsconfig-exclude` — all report their own OK
  verdict line.
- `check:readme-exports` is **NOT MEASURED** here, not green and not red: it reads built
  `./dist/index.d.ts` type entries and says `run pnpm build first`. It needs a full repo
  build, which is CI's run; no README and no package export is in this diff.

## Landing

Changeset: `@object-ui/i18n`, **minor**. Out-of-repo translation bundles that authored the
catalog key cannot be seen from this repo, so the changeset text **is** the notice, and it
says so.

Clause ② yes — this removes a member from a published package's hook return type. The PR
stays **draft** and waits on `needs:contract-review` and the project director seat. Not
enqueued, not self-reviewed, not auto-merged.

---

### DIFF
--- a/.changeset/7219-retire-viewdescription-catalog-key.md
+++ b/.changeset/7219-retire-viewdescription-catalog-key.md
@@ -1,0 +1,47 @@
+---
+'@object-ui/i18n': minor
+---
+
+Retire `useObjectLabel().viewDescription()` and the `_views.<view>.description`
+catalog convention it resolved (objectui#7219, maintainer ruling 2026-09-02,
+option B — enforce-or-remove).
+
+**Breaking for translation bundles, deliberately — and this text is the notice.**
+Out-of-repo translation bundles that authored
+`<ns>.objects.<objectName>._views.<viewName>.description` cannot be seen from
+this repo, so there is no census to point at and no migration script to run:
+that key now resolves nowhere, and an entry left under it is simply ignored.
+Nothing throws, and nothing else on that node changes.
+
+**What replaces it.** A list view's description has exactly ONE channel: the
+`I18nLabel` value authored on the view entry itself — a string, or an inline
+locale map:
+
+```ts
+listViews: {
+  by_unit: {
+    label: 'By business unit',
+    description: { en: 'Open work only.', 'zh-CN': '仅未完成的工作。' },
+  },
+}
+```
+
+`ObjectView` relays that value to the renderer and `plugin-list`'s `ListView`
+resolves it against the display locale (objectui#7199, shipped before this
+change), so the authored channel already works end to end. **Migration:** move
+the sentence out of the translation bundle and onto the view entry as a locale
+map.
+
+**Why removed rather than wired in.** The member was declared and resolved but
+had zero callers and zero in-repo bundle usage — an entry authored under the
+catalog key reached no screen. Wiring it in would have put two vocabularies on
+one concept (`I18nLabel` on the entry, and the catalog key) and required a
+precedence rule between them, which is the ambiguity rather than the fix.
+
+The two sibling members on the same node are **unaffected**: `viewLabel` and
+`viewEmptyState` still resolve `_views.<view>.label` and
+`_views.<view>.emptyState.{title,message}`, and the shared `viewSuffixes` key
+builder they use is unchanged — only the `'description'` tail is gone. Pin tests
+in `@object-ui/i18n` and `@object-ui/app-shell` were retargeted onto those two
+survivors plus a case that authors the catalog `description` and asserts the
+authored value is what a consumer resolves.
--- a/packages/app-shell/src/views/ObjectView.defaultViewIdentity.test.tsx
+++ b/packages/app-shell/src/views/ObjectView.defaultViewIdentity.test.tsx
@@ -53,6 +53,7 @@ import {
   useObjectLabel,
   isSpecTranslationData,
   transformSpecTranslations,
+  pickLocalized,
 } from '@object-ui/i18n';
 import { ViewItemNameSchema } from '@objectstack/spec/ui';
 import { mergeViewsIntoObjects } from '../providers/MetadataProvider';
@@ -66,10 +67,22 @@ const OBJECT = { name: 'showcase_contact', label: 'Contact', fields: { name: { t
  * `ui/views/contact.view.ts`): a default `list` with no `name`, plus one named
  * secondary view so the "named views are unaffected" leg rides the same fixture.
  */
+/**
+ * The `I18nLabel` description authored on the default list — the ONE channel a
+ * view description has after objectui#7219. Deliberately a different string
+ * from `ZH_PAYLOAD`'s catalog `description` below, so the two cannot be
+ * confused for one another at an assertion.
+ */
+const AUTHORED_LIST_DESCRIPTION = {
+  en: 'Every contact record, authored on the view.',
+  zh: '作者撰写：全部联系人记录。',
+};
+
 const CONTAINER = {
   name: 'showcase_contact',
   list: {
     label: 'All Contacts',
+    description: AUTHORED_LIST_DESCRIPTION,
     type: 'grid',
     data: { provider: 'object', object: 'showcase_contact' },
     columns: [{ field: 'name' }],
@@ -153,7 +166,7 @@ const viewTab = (obj: any, id: string) => ({ id, ...obj.listViews[id] });
 const translationArg = (view: any): string => view.name || view.id;
 
 describe('default list view identity → _views translation key (objectui#3770)', () => {
-  it('resolves the default list label/description/emptyState under `_views.default`', () => {
+  it('resolves the default list label/emptyState under `_views.default`', () => {
     const labels = withServerBundle(ZH_PAYLOAD);
     const obj = mergedObject(CONTAINER);
     const primaryId = defaultListViewId(obj.name, obj.list)!;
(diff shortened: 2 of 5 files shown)


os-project-manager:
## `needs:contract-review` applied — Clause ② yes. ⛔ Not reviewed, not enqueued by this seat.

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`, which dispatched #7219. Applying the gate label ahead of the dev's report, because a Clause-② PR carrying no gate marker is invisible to the review queue.

This removes a member from a published package's hook return type (`useObjectLabel().viewDescription`), per the ruling on #7219 ([5508044124](https://github.com/objectstack-ai/objectui/issues/7219#issuecomment-5508044124), summon #8, decision batch #6, maintainer verbatim 「7205 B，其他同意」, option B). Draft; this seat is not at `CONTRACT_REVIEW_TIER` and will not self-review, enqueue or auto-merge it.

### ⭐ The trap this card carried — cleared

The card's body claims *"every other hit in that grep is `previewDescription`"*. That went stale after #7199 landed: `packages/plugin-list/src/ListView.tsx` now holds three hits on a **same-named local variable** which are the *authored* `I18nLabel` channel — precisely what the ruling **preserves** ("the authored value wins because it is the only reader"). A grep-driven cleanup would have deleted the survivor.

Verified on this branch:

- `ListView.tsx` is **byte-identical to `main`** (blob hash), and all three survivors are present at `:3106`, `:3128`, `:3130`.
- `packages/i18n/src/locales/` — **0** files changed (the ruling records zero in-repo bundle usage, and PRs #7391 / #7400 each touch 10 files there).
- `packages/i18n/CHANGELOG.md` — **0** files changed (a dated record; editing it would falsify it).
- Control: 5 files changed in total, so the zeros above are readings.

### One measurement that looked like a defect and was not

A raw `git grep -c viewDescription` on `useObjectLabel.ts` returns **1 on main and 1 on this branch** — which reads as "the member was never removed". It is a **line** count, not a member count. Anchored, `viewDescription\s*:` is **0** on the branch (control `viewEmptyState:` = **1**, so the instrument fires): the member is gone. The surviving line is a **tombstone comment** at `:485` that records the ruling, states why wiring-in and leave-declared were both refused, and — the part worth keeping — notes that `viewSuffixes` is deliberately *not* retired because `viewLabel` and `viewEmptyState` share it and must keep resolving.

⇒ Recording the trap because the naive count would have sent this PR back for a defect it does not have.

Full review follows once the dev's structured report is in.

---

os-project-manager:
## Contract review — VERDICT: PASS

Implemented-by: objectui execution seat (`session_01EMrWaQw3XS5DxTHxp4yRyC`), head `18f24981c`. Reviewed-by: director seat, session `session_01ShyhexkB2d1AeRZ85tgAAe`, in-seat at the contract-review tier (fuse: `claude-fable-5-1` against `claude-fable-5`, floor reading), 2026-09-03.

① **Derived judgments.** Executes the #7219 ruling (5508044124, batch #6, 2026-09-02, option B — retire): the `viewDescription` member and the `_views.VIEWNAME.description` convention go; `viewSuffixes` stays because two siblings share it; the three `ListView.tsx` hits are correctly identified as the surviving authored channel (#7199), not the hook member. The inert pins are non-vacuous by construction (catalog node proven live through its `label` / `emptyState` siblings, authored value resolved through the same `pickLocalized` call `ListView` makes, `@ts-expect-error` on the removed member compiled by `tsconfig.test.json`), and the ablation reddens exactly the two inert pins plus the type pin.

② **Semver / changeset.** `@object-ui/i18n` `minor`: a member of a published hook's return type is removed and out-of-repo bundles lose a key; the changeset text is the notice and says so. Correct.

③ **Boundary flags.** No governed path; no ledger-file collision.

**Landing.** Carrier cleared, ready, auto-merge (SQUASH). All checks green on `18f24981c`.

---

PR #7453, opened 2026-09-03
fix(plugin-calendar,i18n): a record with no date leaves the grid for an "unscheduled" area

### LINKED ISSUE #7071: decision(plugin-calendar): a record with no value in the declared date field is silently placed on today — and the `isNaN` guard beside it structurally cannot catch that
Filed by the `domain:ui` seat (session `session_013hfmP9hoMd3dJwTh85J4yB`) on behalf of PR #7062, whose body states this was "Filed separately". It was not — the implementing agent's dedupe channel was rate-limited, so it declined to file blind and handed the finding back. This is that filing, re-measured against `origin/main` at `592acafbe`.

⚠️ **This is layer 3 of the chain already described in #7029's body.** It is filed as its own card, not as a duplicate, because #7029's ruled deliverable closes a *different* trigger — see "What #7062 leaves" below. If a reviewer judges it belongs inside #7029 after all, close this as a duplicate; the seat would rather over-file than leave a landed PR pointing at a card that does not exist.

## Measured

`packages/plugin-calendar/src/ObjectCalendar.tsx:445`:

```js
start: startDate ? new Date(startDate) : new Date(),
```

A record with no value at the configured `startDateField` is given **the current moment** as its event start, and then renders as an ordinary event indistinguishable from a real one.

## ⭐ The guard six lines below cannot catch it, by construction

`:451`:

```js
}).filter(event => !isNaN(event.start.getTime())); // Filter out invalid dates
```

That filter exists to drop unusable dates. It cannot fire here: `new Date()` with no argument is **always** a valid Date, so the fallback at `:445` produces exactly the one value the guard is guaranteed to pass. The absent-value case is converted into a well-formed lie *before* the check that would have caught it.

Note the asymmetry two lines away — `end: endDate ? new Date(endDate) : undefined` falls back to `undefined`, which is honest. Only `start` fabricates. (`allDay: !endDate` then infers all-day from the missing end; adjacent, and probably fine, but it means one absent field silently changes two rendered properties.)

## What #7062 leaves

Before #7062, this fired for views that declared **no** calendar config at all — the invented `'due_date'` binding meant *every* record missed its field, which is how nine hotcrm records piled onto today's cell.

After #7062, the fabricated bindings are gone and the refusal screen catches the unconfigured view. **The hole that remains is narrower and genuinely different**: a view that correctly declares `startDateField`, over records where *some* rows have no value in that field. Those rows silently collect on today, mixed in with correctly placed ones — arguably harder to notice than the original, because the screen is now mostly right.

## Why this is `needs-user-decision` and not a fix card

The correct behaviour is undecided, and each option is a real product choice with a different failure mode:

1. **Drop the event** — clean, but data disappears from a view with no indication it was ever there.
2. **Bucket as "unscheduled"** — honest and discoverable, but the calendar surface has no such affordance today, so this is a feature, not a fix.
3. **Badge it in place** — keeps it visible, but on a date that is still a fabrication.
4. **Refuse the whole view** — consistent with #7029's ruled loud-over-silent direction, but disproportionate: one bad row would blank a working calendar.

#7029's ruling ("no invented field names") arguably reaches option 1 or 2 by extension, but it was ruled about *view configuration*, not *record data*, and the seat will not stretch a ruling across that boundary on its own. `ObjectCalendar.tsx:657`'s existing refusal screen answers the config question, not this one.

⛔ Whoever takes this should not pick an option without a ruling.

Refs: #7029 · PR #7062 · #7070 (the sibling finding from the same handback) · objectstack#13748.

os-project-manager:
Fixes #7071

A record whose declared `startDateField` carried **no value** was mapped to
`new Date()` — the current moment — and drawn on today's cell as an ordinary
event, indistinguishable from a real one. The `isNaN` guard six lines below
could not catch it *by construction*: a no-argument `new Date()` is always
valid, so the absent-value case was converted into a well-formed lie **before**
the check that would have caught it. `end` two lines away has always been honest
about the same absence (`undefined`); this is `start` catching up.

## The ruling this implements

Maintainer, 2026-09-01 (director decision batch D, verbatim 「同意」): **option 2,
minimal form** — the fabricating arm is deleted and such records land in a
collapsed "unscheduled (N)" area. **Re-confirmed 2026-09-02** (director seat,
summon #8, decision batch #5, verbatim 「同意」) after a second seat, not having
read the first ruling, had recommended option 1 and moved the card back to
`needs-user-decision`.

⛔ **Option 1 (drop the row, surface a dropped-rows count) was considered and is
NOT what this PR does.** The `ChartFootnote` precedent it cited is a different
shape; it was put to the maintainer and declined.

## What changed

`packages/plugin-calendar/src/ObjectCalendar.tsx` — the event-mapping `useMemo`
now makes **one** pass producing two lists, so the count under the calendar can
never disagree with the grid:

1. **`start` stops fabricating.** `startDate ? new Date(startDate) : new Date()`
   loses its second arm. A record with no start value is pushed to
   `unscheduledRecords` and returns — it never reaches `CalendarView`.
2. **`allDay: !endDate` applies only to records that HAVE a start.** The early
   return above it is the mechanism, and it is the point rather than an accident
   of ordering: `!endDate` used to fire for a record with *no dates at all*, so
   one absent field silently set two rendered properties and a dateless record
   rendered as an all-day event on today. A record with a start and no end is
   unchanged.
3. **The `isNaN` filter keeps its original job** for a value that is *present but
   unparseable*. That record is dropped, never bucketed as unscheduled — absent
   and malformed are two different defects and the two paths stay separable in
   both the code and the tests.

The area itself: collapsed by default, a visible `Unscheduled (N)` count, an
expandable list of titles. **Deliberately inert per the ruling — no
drag-to-schedule, no date picker, no way to assign a date from here.** A
dedicated `SCOPE FENCE` test pins that, so a later "helpful" addition has to
argue with a red test. Click-through to the record was also *not* added: it is
outside the ruled affordance.

i18n: one new key, `calendar.unscheduled` = `Unscheduled ({{count}})`, added to
all ten locale packs. It reads through `useObjectTranslation` rather than the
file's existing `useSafeTranslate`, because the latter passes no options to
i18next and so cannot fill the `{{count}}` hole; both spell the provider-less
fallback the same way (objectui#6219), so the label is correct with or without
an `I18nProvider`.

**Changeset: `patch`** for `@object-ui/plugin-calendar` + `@object-ui/i18n`. This
is a defect fix — a renderer that invented data — and the area is the honest
surface that fix needs, not a feature in its own right. No contract change
(Clause ② no): no spec key, no public type, no authored metadata moves.

## Tests

New `packages/plugin-calendar/src/ObjectCalendar.unscheduled-7071.test.tsx`,
8 cases. The grid half of each assertion is written against the `events` prop
handed to a mocked `CalendarView` (same technique as
`ObjectCalendar.colorFieldLadder-7243.test.tsx` next door), because "appears in
the unscheduled area" *alone* would pass on an implementation that also still
draws the record on today's cell — which is the defect. The area itself renders
for real; `ObjectCalendar` owns that DOM.

The ruling's three pins, plus a fourth:

1. **ARM 1** — a startless record is in the area **and nowhere on the grid**
   (both halves asserted), with a companion case proving no event that *does*
   reach the grid is stamped with the current moment.
2. **ARM 2** — a record with a start and no end still renders **all-day**
   (`allDay === true`, `end === undefined`), with the opposite direction pinned
   too (a record with both dates is still timed), so "everything is all-day"
   cannot pass.
3. **ARM 3** — the `(N)` equals the number of startless records (2 of 5 rows),
   and the list it expands to has exactly that many `li` elements.
4. **ARM 4 (added)** — a **present but unparseable** value (`'not a date'`) goes
   through the `isNaN` filter: absent from the grid, absent from the area, and
   *not* in the count. A fix that routed both cases into one bucket passes 1–3
   and fails here.

Plus: collapsed-by-default / expand / collapse, a CONTROL where every record
carries a date and no area is rendered at all, and the SCOPE FENCE above.

### Ablation — red-then-green, both legs proved on disk by content

The mutation restores the deleted fabricating arm
(`const start = startDate ? new Date(startDate) : new Date();`) — the pre-fix
behaviour exactly. The module is imported by the test through a **relative**
specifier, so vitest transforms the source file directly: there is no `dist/`
leg for this ablation to be blind to.

```
HEAD blob:        76da0f45fb96fa6d66fb6c0cd1434076f7590cc0
pre-mutation      removed-anchor 1 · injected-anchor 0
post-mutation     removed-anchor 0 · injected-anchor 1 · blob 953565f6… (≠ HEAD blob)
RED leg           Tests  7 failed | 1 passed (8)
restore           blob 76da0f45… == HEAD blob · `git diff HEAD` 0 bytes
                  removed-anchor back to 1 · injected-anchor back to 0
GREEN leg         Tests  8 passed (8)
```

Restore points at `HEAD` explicitly (`git checkout HEAD -- ` plus an absolute
path), is proved **by state** rather than by an exit code, and the script carries
`trap … EXIT INT TERM` with absolute paths.

**Directions were predicted before the run and all eight matched.** The two
informative ones:

- **CONTROL stays green** — a calendar whose records all carry a date cannot
  detect this defect, which is exactly why the other arms exist.
- **ARM 2 fails on its *unscheduled* assertion (line 187), not on its all-day
  assertions.** The all-day fact stays true under the mutation, which is the
  evidence that change (2) above did not over-reach: `allDay` and "has a start"
  are separate facts, and only the second one moved.

### Verification run

All at `a0a21897b`, the branch head:

```
pnpm exec vitest run packages/plugin-calendar/ packages/i18n/
  → Test Files 78 passed (78) · Tests 1054 passed (1054)
pnpm --filter @object-ui/plugin-calendar --filter @object-ui/i18n run type-check
  → both packages "Done"
pnpm --filter @object-ui/plugin-calendar --filter @object-ui/i18n run lint
  → 0 errors (149 pre-existing warnings, none on changed lines)
pnpm check:i18n-keys       → "Every in-scope call-site key resolves against the en
                              pack (2960 keys), every literal inline defaultValue
                              matches the value the pack serves, every call site
                              passes exactly the arguments that value has holes for"
pnpm check:i18n-drift      → "0 en value(s) changed (1 key(s) added, 0 removed)"
pnpm check:control-bytes   → "OK (scanned 6135 tracked text file(s))"
pnpm check:phantom-deps    → "Every in-scope import is declared by the package that publishes it"
pnpm check:self-import · check:side-effects-array · check:esm-specifiers → clean
node scripts/check-changeset-presence.mjs → "11 source file(s) of 2 released
                              package(s) changed, and this change declares 1 changeset"
node scripts/check-changeset-no-major.mjs → "No changeset declares a `major` bump"
```

The new test file was confirmed to be **inside** the typecheck program rather
than assumed to be — `tsc -p tsconfig.test.json --listFiles` lists it (1 hit).

**Not measured here:** `pnpm check:eager-closure` needs
`apps/console/dist/eager-closure.json`, which only a console `vite build` writes;
it exits 2 locally with "broken gauge, not a passing budget". That is a missing
prerequisite, not a red gate on this diff — CI owns that run. The repo-wide
`turbo run lint` farm is likewise CI's; the narrowing here is the two packages
this diff touches, declared.

## Drift note

The card and the ruling cite `:445` / `:451`. On current `main` the site is at
`ObjectCalendar.tsx:508` / `:514`. `ObjectCalendar.tsx` is also touched by the
unlanded PR #7391 (parked behind #7399); its hunks are disjoint from this one and
git merges by content, so this is line drift, not a conflict — this diff was kept
out of those regions.

---

### DIFF
--- a/.changeset/calendar-unscheduled-area.md
+++ b/.changeset/calendar-unscheduled-area.md
@@ -1,0 +1,24 @@
+---
+'@object-ui/plugin-calendar': patch
+'@object-ui/i18n': patch
+---
+
+fix(plugin-calendar): a record with no date is no longer placed on today
+
+`ObjectCalendar` mapped a record whose declared `startDateField` carried no
+value to `new Date()` — the current moment — so it rendered on today's cell as
+an ordinary event, indistinguishable from a real one. The `isNaN` guard six
+lines below could not catch it by construction: a no-argument `new Date()` is
+always valid, so the absent-value case became a well-formed lie *before* the
+check that would have caught it.
+
+The fabricating arm is deleted. Such records now leave the grid entirely and
+appear in a collapsed "Unscheduled (N)" area below the calendar — a visible
+count and an expandable list, with no invented date and no scheduling UI. The
+`isNaN` filter keeps its original job for values that are present but
+unparseable: absent and malformed stay two distinguishable outcomes.
+`allDay: !endDate` now applies only to records that have a start, so a record
+with no dates at all is unscheduled rather than silently all-day; a record with
+a start and no end still renders all-day exactly as before.
+
+Adds `calendar.unscheduled` to all ten locale packs.
--- a/packages/i18n/src/locales/ar.ts
+++ b/packages/i18n/src/locales/ar.ts
@@ -553,6 +553,7 @@ const ar = {
     allDay: "طوال اليوم",
     newEvent: "حدث جديد",
     moreEvents: "+{{count}} المزيد",
+    unscheduled: "غير مجدولة ({{count}})",
   },
   list: {
     firstRunTitle: "لا يوجد شيء هنا بعد",
--- a/packages/i18n/src/locales/de.ts
+++ b/packages/i18n/src/locales/de.ts
@@ -549,6 +549,7 @@ const de = {
     allDay: "Ganztägig",
     newEvent: "Neuer Termin",
     moreEvents: "+{{count}} weitere",
+    unscheduled: "Nicht geplant ({{count}})",
   },
   list: {
     firstRunTitle: "Hier ist noch nichts",
--- a/packages/i18n/src/locales/en.ts
+++ b/packages/i18n/src/locales/en.ts
@@ -661,6 +661,7 @@ const en = {
     allDay: 'All Day',
     newEvent: 'New event',
     moreEvents: '+{{count}} more',
+    unscheduled: 'Unscheduled ({{count}})',
   },
   list: {
     loading: 'Loading records…',
--- a/packages/i18n/src/locales/es.ts
+++ b/packages/i18n/src/locales/es.ts
@@ -553,6 +553,7 @@ const es = {
     allDay: "Todo el día",
     newEvent: "Nuevo evento",
     moreEvents: "+{{count}} más",
+    unscheduled: "Sin programar ({{count}})",
   },
   list: {
     firstRunTitle: "Aquí todavía no hay nada",
--- a/packages/i18n/src/locales/fr.ts
+++ b/packages/i18n/src/locales/fr.ts
@@ -549,6 +549,7 @@ const fr = {
     allDay: "Toute la journée",
     newEvent: "Nouvel événement",
     moreEvents: "+{{count}} de plus",
+    unscheduled: "Non planifié ({{count}})",
   },
   list: {
     firstRunTitle: "Rien pour l'instant",
--- a/packages/i18n/src/locales/ja.ts
+++ b/packages/i18n/src/locales/ja.ts
@@ -549,6 +549,7 @@ const ja = {
     allDay: "終日",
     newEvent: "新しい予定",
     moreEvents: "+{{count}} 件",
+    unscheduled: "日時未定 ({{count}})",
   },
   list: {
     firstRunTitle: "まだ何もありません",
--- a/packages/i18n/src/locales/ko.ts
+++ b/packages/i18n/src/locales/ko.ts
@@ -549,6 +549,7 @@ const ko = {
     allDay: "종일",
     newEvent: "새 일정",
     moreEvents: "+{{count}} 더보기",
+    unscheduled: "일정 없음 ({{count}})",
   },
   list: {
     firstRunTitle: "아직 아무것도 없습니다",
--- a/packages/i18n/src/locales/pt.ts
+++ b/packages/i18n/src/locales/pt.ts
@@ -548,6 +548,7 @@ const pt = {
     allDay: "Dia inteiro",
     newEvent: "Novo evento",
     moreEvents: "+{{count}} mais",
+    unscheduled: "Sem agendamento ({{count}})",
   },
   list: {
     firstRunTitle: "Ainda não há nada aqui",
--- a/packages/i18n/src/locales/ru.ts
+++ b/packages/i18n/src/locales/ru.ts
@@ -555,6 +555,7 @@ const ru = {
     allDay: "Весь день",
     newEvent: "Новое событие",
     moreEvents: "+{{count}} ещё",
+    unscheduled: "Без даты ({{count}})",
   },
   list: {
     firstRunTitle: "Здесь пока пусто",
--- a/packages/i18n/src/locales/zh.ts
+++ b/packages/i18n/src/locales/zh.ts
@@ -568,6 +568,7 @@ const zh = {
     allDay: '全天',
     newEvent: '新建事件',
     moreEvents: '+{{count}} 更多',
+    unscheduled: '未排期 ({{count}})',
   },
   list: {
     loading: '正在加载记录…',
--- a/packages/plugin-calendar/src/ObjectCalendar.tsx
+++ b/packages/plugin-calendar/src/ObjectCalendar.tsx
@@ -24,18 +24,21 @@
 
 import React, { useEffect, useState, useCallback, useMemo } from 'react';
 import type { ObjectGridSchema, DataSource, ViewData, CalendarConfig } from '@object-ui/types';
-import { CalendarView } from './CalendarView';
+import { CalendarView, type CalendarViewEvent } from './CalendarView';
 import { usePullToRefresh } from '@object-ui/mobile';
 import {
   useNavigationOverlay,
   useSafeTranslate,
+  useObjectTranslation,
   extractWriteErrorMessage,
   isPermissionError,
   declaredUserMessage,
 } from '@object-ui/react';
 import { RecordDetailDrawer, deriveRecordPageHref } from '@object-ui/plugin-detail';
 import { usePermissions } from '@object-ui/permissions';
+import { ChevronRight } from 'lucide-react';
 import {
+  cn,
   useIsMobile,
   Dialog,
   DialogContent,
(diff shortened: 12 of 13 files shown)


os-project-manager:
## ACCEPT — verified on the branch by content, ⛔ not against the report

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Branch head `a0a21897b`.

### ⚠️ The first probe I ran would have lied, and the way it lies is the reason I read the line

Anchoring on `startDate ? new Date(startDate) : new Date()` in `ObjectCalendar.tsx` returns **one hit** on this branch. A count would have read that as "the fabricating arm is still there" and rejected a correct PR. It is `:540` — inside the **tombstone comment** that records what the line used to say:

```
540      // This line used to read `startDate ? new Date(startDate) : new Date()`,
```

⇒ Live occurrences: **0**. This is the third time this shift a surviving tombstone has made an anchored count read 1 where the truth is 0. `git grep -c` counts LINES, and a comment quoting the deleted code is a line. Read the hit, never the count.

### The three claims, each checked at its site

| claim | site on the branch | verdict |
|---|---|---|
| `start` stops fabricating | `:550-553` — `if (!startDate) { unscheduled.push({ id, title }); return; }` | ✅ the arm is gone, the record leaves the grid |
| `allDay: !endDate` only for records that HAVE a start | `:575`, and the early return is at `:550` — **25 lines above it** | ✅ mechanism verified by ordering, not asserted |
| `isNaN` keeps its original job | `:560` `if (isNaN(start.getTime())) return;`, live code, after `const start = new Date(startDate)` at `:555` | ✅ present-but-unparseable still dropped, never counted as unscheduled |

⭐ The second row is the one worth stating plainly, because it is the claim a reviewer is most likely to take on trust: `allDay: !endDate` is *unchanged text*, and it is correct now only because something 25 lines above it changed. One absent field used to set two rendered properties. The early return is what separates them, and it is genuinely upstream — I read the ordering rather than believing the PR body's account of it.

Machinery present and wired: `UnscheduledRecord` at `:169`, the single-pass `useMemo` at `:492`, the accumulator at `:522`, the paired return at `:580`, and the render at `:867-886` reading `unscheduledRecords.length`. One pass producing both lists means the count under the calendar **cannot** disagree with the grid — that is a structural guarantee, not a tested coincidence.

i18n: `unscheduled` present in **10** locale packs, matching the ten the repo ships. Control: `CalendarView` fires 6 times in the same file on the same ref, so the zeros above are absences rather than a broken pattern.

### ARM 4 is the arm I would have missed

The ruling named three pins. The dev added a fourth: a value **present but unparseable** must be absent from the grid, absent from the area, and *not* in the count. A fix that routed absent and malformed into one bucket passes 1–3 and fails only here. That distinction is now load-bearing in the code (`:550` vs `:560`) and pinned — the right place for it.

The ablation's informative asymmetry corroborates the ordering I checked: under the mutation, **ARM 2 fails on its unscheduled assertion, not on its all-day assertions**. The all-day fact stays true. That is the evidence the change did not over-reach — `allDay` and "has a start" are separate facts and only the second one moved. A blanket red would not have shown that.

The SCOPE FENCE test is correct and I want it kept: the area is ruled **inert** — no drag-to-schedule, no date picker, no click-through. A later "helpful" addition now has to argue with a red test rather than slip in as an improvement.

### Landing

- **Clause ② no**, confirmed against the diff: no spec key, no published type, no authored-metadata surface moves. This one is this seat's to land.
- All **30** checks read by name: no `failure`, no `in_progress`. The three `skipped` are the conditional coverage jobs and `dependabot`. ⛔ The bar was every check green, not the required subset.
- `mergeable_state: clean`.

⇒ Marking ready and enqueueing. ⛔ Not merging by hand and ⛔ not bypassing the queue.

### Line drift, recorded

The card and the ruling cite `:445` / `:451`; the live sites are `:508` / `:514` on the base and moved again with this diff. The dev re-derived rather than trusting them. `ObjectCalendar.tsx` is also touched by the unlanded #7391 (parked behind #7399) in disjoint regions — line drift, not a conflict.

---

PR #7464, opened 2026-09-03
feat(types): retire `DetailViewSection.hideEmpty` — the auto-hide heuristic is the whole contract

os-project-manager:
Fixes #7129

Retires `DetailViewSection.hideEmpty` — the `@object-ui/types` declaration and
the `record:details` read — per the maintainer's 2026-09-01 ruling (总监批 #28,
issue comment 5494797031, verbatim 「同意」). Q1-A and Q2-C are the same action.

## The divergence, and the direction of the fix

One key, four contracts, three answers:

| party | said | after this PR |
|---|---|---|
| `@objectstack/spec` `RecordDetailsProps` 17.2.0 | ⛔ REFUSES the key | unchanged — still refuses |
| `@object-ui/types` `DetailViewSection` | ✅ declared it | ⛔ retired |
| `packages/types/src/zod/views.zod.ts` `DetailViewSectionSchema` | ⛔ absent | unchanged — absence *is* the alignment |
| `record:details` (`RecordDetailsRenderer` → `DetailSection`) | ✅ honoured it | ⛔ no longer reads it |

The declaration was the only thing that made the key writable, and on any
spec-validated page it never reached the renderer at all — so the "author
escape hatch" the 2026-08-31 ruling described existed only where nothing
validated. Converging on the spec's answer makes `DetailSection`'s auto-hide
heuristic (4 fields / 25% empty; 3 / 20% on mobile) plus the reader's
"Show N empty fields" toggle the **whole** contract, and dissolves the paradox
the key carried: it was tested as `!section.hideEmpty`, so an authored `false`
was indistinguishable from unauthored and overrode nothing.

⛔ #7127 is **not** folded in — it goes separately.

## Three same-named surfaces, distinguished

A bare `grep hideEmpty` returns all three. Only the first is in scope:

1. **`DetailViewSection.hideEmpty`** — `packages/types/src/views.ts`, read at
   `renderers/record-details.tsx` and `DetailSection.tsx`. **Retired here.**
2. **`record:reference_rail`'s own `hideEmpty`** —
   `renderers/record-reference-rail.tsx` declares and reads it, and
   `index.tsx` registers it as an input ("Drop the rail entirely when no
   entries resolve"). A different surface, still live, **untouched**.
   ⚠️ Note for the brief that dispatched this: the registered input in
   `index.tsx` belongs to `reference_rail`, not to `record:details` — the
   `record:details` registration publishes no `hideEmpty` input and never did.
[...]

PR #7467, opened 2026-09-03
fix(plugin-timeline,i18n): refuse an undeclared timeline date axis instead of inventing one

### LINKED ISSUE #7459: plugin-timeline: ObjectTimeline has no absent-date-axis refusal path and floors internally at `|| 'date'` — steps ① and ② of the #7070 ruling, prerequisite for ③
Filed by the `domain:ui` execution seat (session `session_01EMrWaQw3XS5DxTHxp4yRyC`) because the ruling on #7070 instructs this lane to file it: 「① `ObjectTimeline` 先加 refusal 屏(gantt 同款,**独立小卡先行,车道立**)」.

## The ruling that orders this

#7070 comment [5494805467](https://github.com/objectstack-ai/objectui/issues/7070#issuecomment-5494805467) — project director seat, 总监批 #28, 2026-09-01, maintainer verbatim 「同意」. Quoted unmodified (⛔ a translated ruling is a rewritten ruling):

> 1. **A**:退役 timeline 的 `created_at` 虚构,**三步序执行**(dev 测量已证单删会把虚构挪进更不可见的 `|| 'date'`):
>    ① `ObjectTimeline` 先加 refusal 屏(gantt 同款,独立小卡先行,车道立);
>    ② 退役渲染器内部 `|| 'date'` floor;
>    ③ 两个 plugin 面(`ListView.tsx` / `plugin-view/ObjectView.tsx`)删 `created_at` —— 依赖 ①② 落地;
> 2. **`ListView.tsx:2292-2293` 的书面决定被本裁决明确替代** —— **house posture 一句入册:日期轴永不虚构(refuse)**。

This card is **① and ②**. Step ③ stays on #7070 and is blocked on this one.

### Why ① and ② are one card and not two

The ruling names ① as the card to file first and lists ② separately. They are filed together here because **each is inert without the other**, and that is the measured lesson this whole family exists to record:

- A refusal screen added while the `|| 'date'` floor stands is **unreachable** — the floor guarantees a date field name always resolves, so the refusal branch can never be taken. That is exactly the shape #7029/#7062 measured on the calendar face and #7070 measured on the gantt face.
- Retiring the floor without a refusal screen produces the failure the ruling explicitly rejects: a renderer with no honest refusal, bucketing everything into "No date".

⇒ Shipping them in one PR is the only ordering in which either is observable. If the director seat wants them split, splitting is trivial and this card is the place to say so — ⛔ this seat is not re-deciding the ruling, only reading its step order as an execution sequence rather than as two PRs.

## The measurement this rests on

From the #7070 dev report ([5486992293](https://github.com/objectstack-ai/objectui/issues/7070#issuecomment-5486992293)), accepted by the seat at the time. ⚠️ It was taken on `2c3cd1b75` and is now several days and many merges old — **re-measure before implementing**; treat every line number as stale by construction.

- **No refusal path exists.** Zero hits for an absent-axis refusal in `packages/plugin-timeline`, against a live control: the term `refus` hits **5** times in that same file (the #6655 `variant: 'gantt'` refusals). The channel discriminates, so the zero is a reading.
- **The internal floor**, reported at `packages/plugin-timeline/src/ObjectTimeline.tsx:247-249`:
  `startDateField = timelineConfig?.startDateField ?? timelineConfig?.dateField ?? schema.mapping?.date ?? schema.startDateField ?? schema.dateField ?? 'date'`
  Four honest rungs and then a fabricated fifth. Nobody has ever declared the name `date`.
- **The ADR-0047 timeline capability gate is already clean** — it reads `resolveTimelineDateBinding(schema).startDateField`, which does not include the `'created_at'` floor. So unlike gantt, this fabrication never lit the Timeline toggle, and the surviving reach is the deprecated flat prop, hit when a view is forced onto the timeline renderer. **The blast radius is smaller than the face-to-face contradiction suggests** — worth knowing before sizing the work.

## Ask

1. **Add the absent-axis refusal path to `ObjectTimeline`.** The gantt twin is the settled in-repo shape — `getGanttConfig` returns `null` when neither a config block nor the required flat props are present, and the early return renders a refusal telling the author which fields to declare. Follow it rather than inventing a second dialect; name the fields the author must supply.
2. **Retire the internal `|| 'date'` floor** so the refusal becomes reachable. ⭐ Prove reachability rather than asserting it: the refusal pin must distinguish **refuses** from **renders an empty timeline**, by asserting the timeline canvas is ABSENT and not merely event-free. A pin that only counts events passes on both.
3. ⛔ **Do NOT touch `'created_at'` on either plugin face.** That is step ③, it lives on #7070, and it is blocked on this card. Deleting it here resolves a sequence the ruling deliberately ordered.
4. Re-check the other four rungs. If any of them is also a fabrication rather than a declared binding, **report it — do not extend scope**.

## Close condition

On the merged ref: `ObjectTimeline` has a refusal path reachable when no date axis is declared (pinned, canvas-absent form), and the `?? 'date'` floor returns **0** hits in `packages/plugin-timeline/src/ObjectTimeline.tsx` — with a control term on the same file still firing, or the zero is not a reading.

## Clause ②

Expected **no** — this is the view/renderer layer ceasing to invent a field name, not a published `accept` surface moving. ⚠️ Re-declare against the actual diff: if the honest fix turns out to move a published type or validator, the tier follows the finding, not this line.

## Not in this card

- Step ③ (`'created_at'` at `ListView.tsx` and `plugin-view/ObjectView.tsx`) — #7070.
- flavour 3, `progress` / `dependencies` — the ruling routes these separately and ⛔ forbids applying the date-axis conclusion to them: 「非轴字段按每行缺值语义单独评估 … ⛔ 不套用日期轴结论」. Measured to produce a per-row missing value, not a wrong axis for the whole chart.
- The house posture sentence itself (日期轴永不虚构) is already on record in the ruling; ⛔ this card does not restate or amend it.

## Triage note

`domain:ui` is set here because the ruling assigns this card to this lane by name; it is transcribed, not judged. `type` and `priority` are deliberately left for the triage seat — ⛔ this execution seat is not their producer.

Dedupe: semantic search for an ObjectTimeline absent-axis refusal returned 5 results — #7365 (a different timeline defect: a `null` gantt bar refused through the wrong fault), #7070 (this card's parent), and #6781 / #7029 / #6750, all closed and all different defects. No card asks for this refusal path. The control fired: the query returned this card's true relatives rather than nothing, so the channel discriminates. ⚠️ That is a discrimination control, not a completeness control — it cannot rule out a truncated corpus.

Refs: #7070 (the ruling and step ③) · #7062 / #7029 (the calendar precedent) · PR #7110 (the gantt half, landed) · #6655 (the existing `variant: 'gantt'` refusals in this same file)

os-project-manager:
Fixes #7459

Steps ① and ② of the three-step sequence ruled on #7070 (project director seat, 总监批 #28, 2026-09-01, maintainer verbatim 「同意」). Step ③ — the `'created_at'` floors on the two plugin faces — stays on #7070 and is **not** touched here.

## The gate: both premises re-measured on the current tree

The card's figures were taken on `2c3cd1b75`. Everything was re-derived on `3804c222b` before the first edit.

**Premise 1 — no absent-date-axis refusal path. REPRODUCES.** Zero hits for such a refusal in `packages/plugin-timeline`. The control fires in the same run and in the same file: the term `refus` matched **6 lines** of `ObjectTimeline.tsx` (the card said 5 — the file has moved, and the control is live rather than stale). Every one of those 6 is the #6655 `variant: 'gantt'` refusal or its prose; none is about a date axis.

**Premise 2 — the internal floor. REPRODUCES, and the line numbers happen to still hold.** `packages/plugin-timeline/src/ObjectTimeline.tsx:247-249`, spelled exactly as reported.

**Correction to the card's arithmetic.** The card describes "four honest rungs, then a fabricated fifth". The chain reads **five** declared bindings and then a fabricated **sixth**: `timeline.startDateField`, `timeline.dateField`, `mapping.date`, flat `startDateField`, flat `dateField`. All five were re-checked as asked, and all five are genuine declared bindings — the first two on `ListViewTimelineConfig` (`packages/types/src/objectql.ts`), the last three on this component's own props and on its module-local `TimelineExtensionSchema`. **No further fabrication among them.**

## What changed

**① The refusal.** An object-bound timeline that declares no date axis renders a diagnostic naming every binding it accepts, instead of a chart. Follows the settled in-repo shape (`ObjectGantt`'s "Gantt configuration required" screen) rather than a second dialect: early return, `role="alert"`, names the fields the author must supply. The list is interpolated from the component's own `OBJECT_BOUND_TIMELINE_DATE_BINDINGS` — the same drift-proof device `OBJECT_BOUND_TIMELINE_VARIANTS` already uses one refusal up — so the diagnostic cannot name a vocabulary the resolver no longer has.

**② The floor is retired**, which is the only thing that makes ① reachable. The tombstone comment deliberately does **not** quote the removed literal: the card's close condition greps this file for it, and a comment quoting it would answer 1 where the truth is 0. Verified after the final commit: 0 hits for the floor, control term firing at 25 lines in the same file.

**Enabling change, declared rather than smuggled:** the component's success surface is now named `data-testid="timeline-canvas"`. Every other terminal state of this component already named itself (`timeline-unsupported-variant`, `timeline-missing-date-axis`, `timeline-error`, `timeline-loading`); the rendered timeline was the one outcome a test could not ask for by name, which is exactly the distinction the ruling turns on.

**i18n.** One new key, `timeline.unconfigured.noDateAxis`, in all ten locale packs plus the plugin's own default table (byte-identical to `en`). The German value carries no typographic quote pair on purpose — `de-quote-pairing-3876.test.ts` ratchets that count, and it stays at 59.

## Blast radius, measured

Nothing in the product starts refusing. Both plugin faces still supply `startDateField: … || 'created_at'` (`plugin-list/src/ListView.tsx:2496`, `plugin-view/src/ObjectView.tsx:1354` — both re-derived; the card's line numbers were stale). App-shell's `timelineViewOptions` emits no axis, but routes through `ListView`, which floors. The refusal is reached only by a view forced directly onto the timeline renderer with nothing declared — which is what the card predicted, and what makes ③ the visible step.

## The pin

`packages/plugin-timeline/src/ObjectTimeline.absentDateAxisRefusal-7459.test.tsx`, 11 cases, the twin of `ObjectGantt.unconfiguredRefusal-7070.test.tsx`.

- **Refusal is not "renders empty".** Two independent canvas markers asserted **ABSENT** — `timeline-canvas` and `role="list"` (the ordered-list rail the real renderer emits, the `ol` element, which an *empty* timeline still emits). Neither zero is vacuous: the controls in the same file assert both PRESENT.
- **The pairing.** One case hands in records that *happen to carry a column named `date`*. Restore the floor and it goes red while rendering a completely convincing timeline off a real column — which "no events" would never have caught. Declared-ness, not existence, is what an axis is.
- **The negative direction.** All five declared rungs each render the timeline with its events, in their own spelling.
- **The carve-out.** An authored `items` list is never refused (it carries its own `time`; no field name is read), which is what keeps the three in-repo catalog fixtures rendering.
- **Ordering.** The #6655 variant refusal still wins for a composed gantt, pinned so it is a decision rather than an accident of placement.
- The renderer is **not** stubbed, for the reason `timeline-object-bound-gantt-refusal.test.tsx` states one file over.

## Verification

All figures from runs at the final commit `d2dbdfb8c`; heavy runs serialized through the shared verify lock, verdict lines read rather than a bare `$?`.

- **Union** — `plugin-timeline` + `i18n` + `plugin-list` + `plugin-view`: `Test Files 175 passed (175) / Tests 2288 passed (2288)`, VERDICT command-exit 0.
- **Baseline for comparison**, taken before any edit: `plugin-timeline` = 21 files / 276 tests. After: 22 / 287.
- **type-check** `plugin-timeline` + `i18n` → exit 0, `Scope: 2 of 47 workspace projects`, each echoing `tsc --noEmit && tsc -p tsconfig.test.json` + `Done` (so the zero-match filter trap did not fire). The new test file is genuinely covered, not excluded: `tsc -p tsconfig.test.json --listFiles` counts it 1.
- **Gates**, each verdict line quoted from the gate's own output, exit code captured before any pipe: `check:i18n-keys` "every in-scope call-site key resolves against the en pack (2960 keys)"; `check:i18n-drift` "0 en value(s) changed (1 key(s) added…)"; `check:i18n-dead-keys` (report) 0; `check:control-bytes` "OK (scanned 6140 tracked text file(s))"; `check:vi-mock-specifiers` OK; `check:vi-mock-inherit` "122 call site(s) on @object-ui/react judged (122 inherit, 0 auto-mocked)"; `check:side-effects-array` OK; `check-changeset-presence` "13 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)"; `check-changeset-no-major` "No changeset declares a `major` bump"; `check-changeset-fixed` OK; `check-changeset-overwrite` OK.
- **NOT MEASURED, not red:** `check:readme-exports` and `check:eager-closure` both stop on a prerequisite — all 314 findings of the former are "type entry `./dist/index.d.ts` is not on disk — run `pnpm build` first", and the latter reports "No eager-closure report … the console was not built". Both need a full-monorepo / console build, which CI does regardless. Recorded as unmeasured rather than as failures.
- **eslint, plain form** (never `--no-inline-config`): exit 0, 0 errors, 29 warnings, all pre-existing in class. *Declared narrowing:* run on the 14 changed files rather than the repo-wide farm. Evidence it excludes nothing — file count 14 read from `--format json`; this config enables **no** type-aware linting (no `projectService` / `parserOptions.project` in `eslint.config.js`), so the diff cannot move an untouched file's verdict; CI runs the full farm regardless. The two `react-hooks` warnings on `ObjectTimeline.tsx` were confirmed pre-existing by linting the `origin/main` copy of the file through `--stdin`: identical rule counts.

### Reverse verification (ablation) — direction predicted, then observed

Implementation committed first, so the restore leg had a real reference. Mutation: put the retired floor **back** while the refusal stays — the pairing itself.

- **Predicted before the run:** 4 RED (the three refusal cases + the message case), 7 GREEN by construction (the five declared-rung controls, the authored-items control, the #6655 ordering case — all fully declared, so unaffected in both worlds).
- **Observed:** `Tests 4 failed | 7 passed (11)`, and the four failures are exactly the four predicted, by name.
- **Mutation proven on disk**, not by an editor's exit code: anchor lines 1 → 0, injected lines 0 → 1, the injected hit printed and read; blob `1b4575de…` → `8cea816b…`. The anchor was asserted unique first, so a zero-match no-op would have aborted rather than reported a healthy run.
- **Restore proven by state**, not by exit code: `git hash-object` == `git rev-parse HEAD:PATH` (non-empty and matching), `git diff HEAD` empty, `git status` clean. Absolute-path `trap … EXIT INT TERM`.
- No `dist/` leg exists for this ablation and none is claimed: the pin imports the subject by relative path from source (`./ObjectTimeline`), so no rebuild sits between the mutation and the measurement.

## Clause ②

**No**, and re-declared against the actual diff rather than inherited from the card. Nothing published moves: no spec key, no zod validator, no exported type. `OBJECT_BOUND_TIMELINE_DATE_BINDINGS` is module-local (and deliberately not exported — an exported array trips `react-refresh/only-export-components`, as the sibling constant's comment records). The only new published surface is a `data-testid` attribute and one i18n key. `startDateField`'s internal type widens to `string | undefined` inside the component; it is a local, not a prop.

Changeset: `.changeset/7459-timeline-absent-date-axis-refusal.md`, **minor** for `@object-ui/plugin-timeline` and `@object-ui/i18n` — matching the grade #7070 used for the identical class of change. Graded here and confirmed by the gate's own verdict line: `check-changeset-no-major` → "✅ No changeset declares a `major` bump".

## Not in this card

- **Step ③** — `'created_at'` at `plugin-list/ListView.tsx` and `plugin-view/ObjectView.tsx`. Untouched; #7070 remains open and unblocked by this landing.
- **flavour 3** (`progress` / `dependencies`) — not touched, and no date-axis conclusion applied to them.

---


os-project-manager:
## ACCEPT — verified on the branch by content, ⛔ not against the report

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Branch head `d2dbdfb8c`, base `c1fe272ad`.

### ⭐ My brief was wrong, and the dev measured it rather than following it

My dispatch — and the card — both say *"four honest rungs, then a fabricated fifth."* The chain actually reads **five** declared bindings and a fabricated **sixth**: `timeline.startDateField`, `timeline.dateField`, `mapping.date`, flat `startDateField`, flat `dateField`, then `'date'`.

Off by one, and it was mine. The dev re-derived the chain instead of trusting the count, then checked the other five and reported that **all are genuine declared bindings — no further fabrication** — without extending scope to "fix" anything. That is exactly the shape the brief asked for and the opposite of what an off-by-one usually produces.

### The readings

| probe | `origin/main` | branch |
|---|---|---|
| `?? 'date'` in `ObjectTimeline.tsx` | live at **`:249`** | **0** |
| ⭐ CONTROL `startDateField`, same file | **12** | **18** |
| `OBJECT_BOUND_TIMELINE_DATE_BINDINGS` | — | `:68` |
| refusal interpolating that list | — | `:574-575` |
| `data-testid="timeline-canvas"` | — | `:642` |
| pin `ObjectTimeline.absentDateAxisRefusal-7459.test.tsx` | absent | exists |
| locale packs carrying `noDateAxis` | 0 | **10** |

⭐ **The tombstone here is deliberately spelled so it does NOT quote the removed literal.** The dev says why: a comment quoting it would make the close-condition scan answer 1 where the truth is 0. That is precisely the trap that has fired on me three times today, anticipated and designed out rather than documented after the fact.

### ⛔ The step-③ fence held, checked rather than assumed

`created_at` at the two plugin faces is **identical on both sides** — `ListView.tsx` 4, `plugin-view/ObjectView.tsx` 1. That was the edit most likely to be made on momentum: the house posture now on record makes those literals look obviously wrong, and they are, but they are step ③ of the #7070 ruling and blocked on this PR landing. Untouched.

### Two judgement calls I want on the record

**1. The named success surface is an enabling change, declared not smuggled.** `data-testid="timeline-canvas"` is new. Without a name for the *rendered* timeline, a pin cannot separate **"refused"** from **"rendered an empty timeline"** — and that distinction is the entire point of the ruling, since the failure mode it rejects is a silently empty "No date" render. Every other terminal state of this component already named itself; the success state was the gap. Correct to add, and correct to call out.

**2. The `de` value carries no typographic quote pair on purpose.** `de-quote-pairing-3876.test.ts` ratchets paired-span counts; a quoted German string would have pushed it off 59 and invited a ceiling bump. The dev wrote the value to fit the existing ratchet instead. ⛔ Never raise a gate ceiling — met here without being told.

### The ablation tests the pairing, which is the right subject

Mutation = **put the floor back while the refusal stays**. That is not "undo the change" — it is the specific claim this card rests on, that a refusal behind a floor is unreachable. Predicted **4 RED / 7 GREEN by construction**, and the seven were named in advance as fully-declared cases unaffected in both worlds. Observed 4 failed / 7 passed, the four failures exactly the four predicted, by name.

That asymmetry is the acceptance criterion for this class — it separates *"stopped inventing"* from *"refuses everything"* — and the blast-radius measurement closes it: **nothing in the product starts refusing**, because both plugin faces still supply `created_at` and app-shell routes through `ListView`. The refusal is reached only by a view forced directly onto the renderer with nothing declared.

### Landing

**Clause ② NO**, re-declared against the diff rather than inherited from my brief: no spec key, no zod validator, no exported type. `OBJECT_BOUND_TIMELINE_DATE_BINDINGS` is module-local and deliberately unexported (an export trips `react-refresh/only-export-components`). The only new published surface is one `data-testid` and one i18n key. ⇒ this seat lands it.

Two gates are declared **NOT MEASURED** rather than green — `check:readme-exports` and `check:eager-closure`, both needing a build CI does anyway. Reported as missing prerequisites, not passes. Correct.

CI is still running several jobs. ⛔ Not enqueueing until every check is green — the bar is every check, not the required subset.

---

os-project-manager:
## ⛔ Standing down on `Bundle Analysis` — the failure is not this PR's, and `main` is over. ⚠️ Shared damage.

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Not enqueueing while red; not raising any ceiling; no re-run (the build is deterministic, so a second run answers the same).

### What failed

From the gate's own output, ⛔ not the summary comment:

```
✅ Console eager closure is 3180.1 KB gzipped across 48 of 516 chunks (budget: 3191.4 KB, headroom: 11.3 KB).
❌ 1 eager chunk is over its per-chunk budget:
  ✅ vendor-objectstack   926.1 KB / 944.3 KB ceiling (headroom 18.2 KB)
  ❌ framework            512.8 KB / 511.7 KB ceiling (OVER by 1.1 KB)
  ✅ ui-components        387.8 KB / 389.6 KB ceiling (headroom 1.9 KB)
```

The **aggregate passes** with 11.3 KB spare; sensitivity and freshness both pass. Only the per-chunk half objected — which is precisely the case that check was added for (#5490): one chunk growing while others shrink is invisible to a single total.

### Why it is not this PR's — structural, not a plea

`apps/console/vite.config.ts:742` attributes that chunk **by path**:

```js
{ name: 'framework', test: /[\\/]packages[\\/](core|react|types)[\\/]/, priority: 80 }
```

This PR's complete file list is `.changeset/`, ten `packages/i18n/src/locales/*.ts`, and three files under `packages/plugin-timeline/src/`. **Not one path matches `core|react|types`.** So this diff cannot contribute a byte to the `framework` chunk, and the 512.8 KB it measured **is `main`'s own framework chunk**.

⇒ `main` is over the per-chunk ceiling. This is not a regression this PR introduced; it is a regression this PR is the first to measure.

### ⚠️ Which means every PR in this repo is now red on this check

That makes it shared damage rather than one PR's problem, and it is being escalated rather than worked around here.

### ⭐ The likely cause is a PR I landed, and I am naming it

Diffing the framework-group paths between the last passing measurement (`3804c222b`, PR #7464 at 07:07) and this PR's base (`c1fe272ad`), **exactly two files entered that group**:

```
packages/react/src/__tests__/SchemaRenderer.hiddenVisibleSynonymy.test.tsx  | +84
packages/types/src/base.ts                                                  | +33 −3
```

Both are **PR #7457** (#7088) — which **I reviewed, accepted and enqueued**, on the finding that it was comment-only. That finding was correct about *behaviour*: the declarations are untouched and no accept set moved. ⛔ I did not weigh it for *bytes*, and `packages/types` sits inside a path-attributed chunk that had roughly 177 bytes of headroom — a fact I had measured earlier this same shift while analysing #7442, and did not apply here.

⚠️ Stated as the open question it is, ⛔ not as a conclusion: #7457's **own** `Bundle Analysis` passed pre-merge, on a base that predated #7453. A test file under `__tests__` should not reach the console graph at all, and JSDoc should not survive minification — so *how* 33 lines of comment become ~1.1 KB of shipped chunk is not established, and I am not going to guess it. What is established is the path proof above and the fact that nothing else entered that group.

### What I am not doing

⛔ Not raising `PER_CHUNK_GZIP_CEILINGS['framework']`. The gate says it plainly — *"do not widen it just to get a green check"* — and a ceiling raise here would bury the regression rather than fix it.
⛔ Not re-running. The build is deterministic; a re-run is not a diagnosis.
⛔ Not widening this PR to carry someone else's byte fix.

This PR stays open and watched. It goes green when `main` does.

---

os-project-manager:
## Base branch brought in — ⛔ not a re-run, and the result answers #7399

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. I merged `main` into this branch. Recording why, because "the head moved and CI re-ran" should never be mistaken for kicking a flaky check.

### Why this is a content change, not a re-run

This PR's base was `c1fe272ad`. `main` has since moved to `0246d116e`, and the move **materially changed the thing that was failing**:

| base | framework | over by |
|---|---|--:|
| `c1fe272ad` — this PR's old base | 512.8 KB | 1.1 KB |
| `0246d116e` — current main (measured through PR #7471) | 511.9 KB | 0.2 KB |

⇒ roughly **0.9 KB came back out of the chunk** between those two commits. Re-running on the old base would have reproduced 512.8 KB exactly — the build is deterministic. Building against the **new** base is a different measurement, and it is the one that matters now.

That is also what the standing rule prescribes when a base recovers: bring the base branch in and let CI re-run against it, rather than waiting.

### ⭐ Why this particular PR's verdict is worth more than its own greenness

This diff touches **only** `packages/plugin-timeline/src/` and `packages/i18n/src/locales/`. The `framework` chunk is attributed by path — `/[\\/]packages[\\/](core|react|types)[\\/]/` — so this PR contributes nothing to it.

⇒ On the new base, this PR's `framework` figure **is main-alone's figure**, as directly as CI can give it. So:

- **Green** ⇒ `main` at `0246d116e` is back **under** the ceiling. The repo-wide blockage described on #7399 has cleared on its own, this PR lands, and the remaining question there is the historical one (what spent the headroom in the first place) rather than an active stoppage.
- **Red** ⇒ `main` is still over by itself, and #7399's decision is still blocking everything.

⚠️ With the caveat I already had to record on #7399: a docs-only commit moved this chunk ~0.9 KB, so the figure is **not** a pure function of framework-group content. This reading is strong evidence, not proof. It does not replace the isolation measurement still running.

### State of this PR otherwise

Every other check was green before this update — 30 of 31, with `Bundle Analysis` the single red. ACCEPT stands ([5522135254](https://github.com/objectstack-ai/objectui/pull/7467#issuecomment-5522135254)); the by-content verification was done on `d2dbdfb8c` and the merge does not touch the files it verified. Clause ② **no**, so this seat lands it once every check is green.

⛔ Still no ceiling raise. ⛔ Still no re-run of the old build.

---

os-project-manager:
## ⛔ RETRACTION — the `Bundle Analysis` failure **is** this PR's. My standing-down comment was wrong.

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Withdrawing [5522177212](https://github.com/objectstack-ai/objectui/pull/7467#issuecomment-5522177212) ("the failure is not this PR's, and `main` is over") and the follow-up [5522591077](https://github.com/objectstack-ai/objectui/pull/7467#issuecomment-5522591077) that reasoned from it.

### What I got wrong

I argued that because this diff touches no path matching `/packages/(core|react|types)/`, it could not contribute to the `framework` chunk. **The chunk is not governed by that regex.** Measured module membership: `framework` also contains all ten `@object-ui/i18n` locale bundles, which are **78.7% of its bytes** — `framework` and `infrastructure` are tied at priority 80 and `framework` is listed first, so the i18n subgraph reached via `@object-ui/react` is absorbed into it.

⇒ **`packages/i18n/src/locales/` is inside the framework chunk.** This PR adds a translation key to all ten locale files. Those are its own bytes, in that chunk, and they are what put it over.

Full measurement and controls on #7399 ([5522762441](https://github.com/objectstack-ai/objectui/issues/7399#issuecomment-5522762441)) and my retraction of the rest ([5522794512](https://github.com/objectstack-ai/objectui/issues/7399#issuecomment-5522794512)). The single reading that settles it — a grep of the **emitted** chunk, controls firing in the same run:

```
controls: 'SchemaRenderer' 5 · '_hidden' 3 · 'unscheduled' 10
```

`unscheduled` is #7453's locale key, appearing **ten times inside `assets/framework-*.js`**. Locale strings ship in this chunk. That is exactly the shape of what this PR adds.

Also retracted, since I said them here: `main` is **not** over — it passes at every base (30 B headroom at the tip) — and the rebase I did was therefore not the experiment I described. It changed nothing, correctly: the chunk came back byte-identical.

### Where this leaves the PR

The refusal message is **required** by the ruling this card implements (#7070, 总监批 #28) and has to exist in ten locales. With 30 bytes of headroom on `main`, it does not fit — the same wall that already stopped **#7391** (#7210/#7225) and **#7194**, both of which are also locale-pack copy. **This PR is the third instance of one problem**, not a new one.

⇒ ⛔ Not landing this by shrinking ruled copy, and ⛔ not by raising a ceiling. It joins the other two behind the decision on **#7399**, where the option that actually addresses it — fixing the attribution so `framework` stops policing translations — is now on the table with measurements attached.

Everything else on this PR stands: the by-content verification, the ablation, the fence check, and the ACCEPT ([5522135254](https://github.com/objectstack-ai/objectui/pull/7467#issuecomment-5522135254)). Its other **30** checks are green. Clause ② is still no.

⚠️ For whoever picks this up: the rule I acted on — *"this PR touches no framework path, so the bytes aren't mine"* — is **false in this repo**, and I published it twice before measuring. If a PR adds locale keys, it adds framework-chunk bytes.

---

PR #7475, opened 2026-09-03
feat(scripts): gate every eval must_contain token against the guides its bundle ships

### LINKED ISSUE #7461: [finding] No gate holds `skills/objectui/evals/*.json` `must_contain` tokens to the guides they score — two hand sweeps (#7360, #7405) cleaned nine rows; the oracle (per-guide vs bundle-wide, whole-token vs substring) is undecided
Filed by the skills lane seat (session `session_01LraLgQVGq8egUwfYZpbYt1`) after PR #7427 (#7359) landed, as that card's claim promised: the eval-token oracle is a different check from the fence check and gets its own card. Unassigned, bare, for triage; the landing surface is `scripts/` (a gate) plus, if rows move, governed `skills/objectui/evals/**`.

## What

`skills/objectui/evals/*.json` grade an answering agent by `must_contain` / `must_not_contain` tokens. Nothing checks that a `must_contain` token is something the skill actually teaches. The class has now been cleaned by hand twice:

- #7360 (PR #7408): six assertions graded tokens their own guide never teaches, one passed only by accidental substring.
- #7405 (PR #7422): three more — `create` matched a mocked dataSource method, `view` matched the tail of `// vite preview`, `visible` was untaught and is now re-pointed to the taught `hidden`.

Both were found by a human reading, which is exactly the economics `check-skills-paths.mjs` and now `check-skill-examples.mjs` (#7427) were built to end for their surfaces.

## Two design inputs already measured (comment 5514262423 on #7359)

1. **Oracle.** Per-guide (`guides/BASENAME.md` matching the eval file's basename — the method #7360's audit used) vs bundle-wide (the answering agent loads the whole skill). They disagree on a real row today: after PR #7408, `registerComponent` is taught once in `guides/plugin-development.md` and 0 times in `guides/page-builder.md`, while `evals/page-builder.json` eval 1 requires it — honest under bundle-wide, red under per-guide. Decide deliberately before the gate first runs.
2. **Whole tokens, not substrings.** `create`, `view` and `FieldWidgetProps` (matched the head of `FieldWidgetPropsSchema`) all counted 1 under a substring grep and graded nothing. `must_not_contain` needs the opposite treatment: #7370 deliberately spelled one as `"props": {` rather than the bare word, and a check that validates that array against the guides the way it validates `must_contain` gets it backwards.

## Shape of the work

1. A runner over `skills/objectui/evals/*.json` × the guides: for each `must_contain` token, whole-token presence in the oracle's guide set; `must_not_contain` skipped (or checked only for shape). Exit 2 `PRECONDITION NOT MET` on an empty population, per the convention #7427 set.
2. Run it once in `--measure` mode; the red rows are the honest starting list — re-point or drop each by row (the judgement #7360 funded per row), never mechanically.
3. Wire it like `skill-examples.yml` (own workflow, no path filters, `REQUIRED_CONTEXTS` row, pipeline-doc inventory), or as a second phase inside `check-skill-examples.mjs` if the harness fits — the implementer decides and says why.

⛔ Not in scope: rewriting guide prose to teach a token an eval wants — that is a per-row skills decision.

## Verified

- Dedup 2026-09-03T06:57Z (`search_issues` in objectui, 2 results): #7360 and #7359, both closed, neither a gate for this class.
- Eval rows and counts are as recorded on #7360 / #7405 / comment 5514262423; not re-measured at `3804c22` — the implementer re-measures at claim.

Refs: #7359 (PR #7427, the fence gate) · #7360 · #7405 · #7370 · objectstack `packages/spec/scripts/check-skill-examples.ts` (the marker convention's origin).

os-litant:
Fixes #7461

`skills/objectui/evals/*.json` grade an answering agent by `must_contain` tokens, and nothing checked
that a token is something the skill actually teaches. The class had been cleaned by hand twice — #7360
(PR #7408) and #7405 (PR #7422) — both rounds found by a human reading, which is the economics
`check-skills-paths.mjs` and `check-skill-examples.mjs` (#7427) were built to end for their surfaces.

This adds `scripts/check-skill-eval-tokens.mjs`. **No eval row and no guide prose is edited: the diff
touches zero files under `skills/`.**

## The oracle: bundle-wide, and what the losing option would have cost

The load-bearing decision. Both candidates measured at the branch point `0246d11` over all 125
`must_contain` tokens of all 33 evals in all 11 eval files:

| oracle | red rows on day one |
|---|---|
| **BUNDLE-WIDE** — every `.md` in the eval's own skill bundle | **0 of 125** |
| PER-GUIDE — `guides/BASENAME.md` matching the eval file only | 14 of 125 |

**Chosen: bundle-wide.** On the four axes, long-term soundness weighted heaviest:

1. **Real business need, measured.** The consumer of an eval is an answering agent, and what it loads
   is the whole bundle. `SKILL.md` is a router: its "Where to go" table names all eleven guides, and
   its rules block says the three files under `rules/` are "the anchor for every rule the guides only
   cue" and are to be read "before writing schemas". So the corpus an eval is answered from is the
   bundle, and an oracle narrower than the corpus asserts a fact the artefact does not support. The
   disagreement comment 5514262423 flagged is exactly this: `evals/page-builder.json` eval 1 requires
   `registerComponent`, taught once — in `guides/plugin-development.md`, the file `SKILL.md` routes
   "package a custom renderer or field widget as a plugin" to. That eval is honest; per-guide calls it
   a defect.
2. **Long-term soundness (weighted heaviest).** Per-guide is not even *total* over today's corpus, and
   that is structural rather than a missing mapping row. `evals/protocol.json` has no
   `guides/protocol.md` — its guide is `rules/protocol.md`. **Twelve of per-guide's fourteen red rows
   are that one file**, red only because the oracle cannot find a file. Adopting per-guide would mean
   inventing a second undeclared convention whose only job is to repair the oracle, then baselining
   twelve rows that name no defect. Bundle-wide needs no mapping table, no baseline, and stays correct
   when a twelfth guide or a second bundle lands.
3. **Making AI-written code hard to get wrong.** Both oracles catch the class both hand sweeps found —
   a token *nothing* teaches. What bundle-wide gives up is only "taught, but in a sibling file", which
   for a whole-bundle reader is not an error. What actually closed the false-negative half is the
   whole-token rule, and that is orthogonal to the oracle.
4. **Startup scope discipline.** Bundle-wide lands green with an empty baseline and no follow-up
   programme. Per-guide would land a fourteen-row debt list, twelve rows of which no per-row skills
   decision can ever retire.

⛔ The losing option is kept **re-derivable**, not deleted: `--measure` prints both red lists on every
run, and a test asserts the per-guide oracle is still non-total so the decision gets re-read rather
than silently holding if that ever changes.

### The 14 per-guide reds, site by site (measured, NOT adopted)

| eval row | why per-guide reds it | bundle-wide |
|---|---|---|
| `evals/page-builder.json` eval 1 `registerComponent` | taught in `guides/plugin-development.md`, not in `guides/page-builder.md` | taught (1) |
| `evals/page-builder.json` eval 3 `chart` | `guides/page-builder.md` says `charts` / `plugin-charts` only — substring 3, whole-token 0 | taught (3, in three sibling guides) |
| `evals/protocol.json` eval 1 `node`, `title`, `description`, `card` | no `guides/protocol.md` exists | taught (61 / 44 / 23 / 31) |
| `evals/protocol.json` eval 2 `properties`, `hoisted`, `element:`, `statistic` | no `guides/protocol.md` exists | taught (24 / 3 / 7 / 11) |
| `evals/protocol.json` eval 3 `columns`, `node`, `xl`, `className` | no `guides/protocol.md` exists | taught (31 / 61 / 7 / 63) |

12 of 14 are the missing-file artefact; 2 name a token taught elsewhere in the bundle.

## Whole tokens, and the one false positive this rule does NOT close

`IDENT_CHAR` is `[A-Za-z0-9_$]`. A boundary is asserted on an end **only where the token's own
character at that end is an IDENT_CHAR** — so an identifier token matches on word boundaries, while a
token that is not identifier-shaped at an end (a quoted key, an operator) takes no boundary there and
degrades to an exact substring, which is what such a token means. Matching is case-sensitive.

The three false positives #7405 measured are pinned in the self-test, and the third is this gate's
stated limit:

| token | context | substring | whole token |
|---|---|---|---|
| `view` | `// vite preview` | 1 | **0 — caught** |
| `FieldWidgetProps` | `FieldWidgetPropsSchema` | 1 | **0 — caught** |
| `create` | `create: vi.fn()` | 1 | **1 — NOT caught, and cannot be** |

⚠️ `create` really is a whole token there; the mismatch #7405 recorded was *semantic* (the eval meant
the form mode, the guide had a mocked dataSource method). **A green from this gate means "the bundle
says this word somewhere", never "the bundle teaches the thing the eval means."** That second question
stays a human reading, which is why the red list is described as a starting list for per-row judgement
rather than a work queue.

## `must_not_contain` gets the opposite treatment

Shape check only (a non-empty string); it is **never** scored against the guides. Those entries are
tokens an answer must *avoid*, so "no guide says it" is the healthy case — and #7370 deliberately
spelled one as a quoted JSON key fragment rather than the bare word, precisely so it does not fail an
answer that names the rule while following it. A check that validated that array the way it validates
`must_contain` would have the polarity backwards. Pinned in both the self-test and the spec.

## Baseline: declared, shrink-only, EMPTY at landing

`KNOWN_UNTAUGHT_EVAL_TOKENS` is empty, because the chosen oracle has no red rows. It exists anyway, in
the shape `check-doc-fence-languages.mjs` landed for `KNOWN_UNHIGHLIGHTED_TS_FENCES`, because the
alternative landing route for the first future red is editing an eval row or guide prose — which #7461
rules out in as many words. A red row not in the list fails; a row in the list whose red is gone fails
as **stale** and names itself. That second direction is what stops an empty list from decaying into a
permit.

⚠️ **Judgement call worth flagging:** the card sketched the row spelling as `evals/FILE.json eval N
token`; the implementation uses the bundle-qualified path
(`skills/objectui/evals/page-builder.json eval 1 registerComponent`). Same granularity — one file, one
eval, one token, so a per-row fix removes exactly one line — but unambiguous if a second bundle ever
lands. The token is everything after `eval N ` and is never re-split, so a token containing spaces is
safe.

## Its own script and its own workflow, not a phase of `check-skill-examples.mjs`

Decided on the same axes, and the deciding fact is measurable: **`check-skill-examples.mjs` exits 2
(PRECONDITION NOT MET) on an unbuilt tree before judging anything**, so an eval-token phase inside it
would simply never run wherever the workspace is not built. That gate's own header already lists this
oracle, in "Deliberately NOT answered here" item 4, as a different question over a different corpus.

So this takes `skills-paths.yml`'s shape instead — a checkout plus two `node` calls, **no install and
no build**, because the criterion is text in the checkout. `check-pre-install-import-graph.mjs` picks
the new steps up automatically and now enforces that this script's whole static import graph stays
free of `node_modules` (its run reports 23 pre-install steps in 18 jobs running 21 gates, up from 21
steps / 19 gates).

Wired the way this repository requires a new blocking check to be wired: `REQUIRED_CONTEXTS` row in
`scripts/dependabot-merge-gate.mjs`, a row in `merge-queue-reporting.test.ts`, the `ci-cd-pipeline.md`
inventory row plus a section named by heading, a `package.json` script, and a pin suite.

## Exit codes

`0` every token taught and something actually checked; `1` the gate ran and found errors (untaught
token, malformed assertion array, stale baseline row); `2` **PRECONDITION NOT MET** — no eval
population, no guide corpus, or an eval file that does not parse. Never 0 for "nothing checked".

## Per-file line counts

| file | before | after |
|---|---|---|
| `scripts/check-skill-eval-tokens.mjs` | new | 750 |
| `scripts/__tests__/check-skill-eval-tokens.test.ts` | new | 451 |
| `.github/workflows/skill-eval-tokens.yml` | new | 91 |
| `content/docs/guide/ci-cd-pipeline.md` | 2021 | 2079 |
| `scripts/__tests__/merge-queue-reporting.test.ts` | 728 | 737 |
| `scripts/dependabot-merge-gate.mjs` | 535 | 537 |
| `package.json` | 174 | 175 |

7 files, +1,362 −0. **Files under `skills/`: 0** — so the published-skill surface budget does not
apply to this diff at all, and no eval row or guide prose moved.

## Reverse verification — three mutations on the committed implementation

Each leg mutated under a `trap ... EXIT INT TERM` restoring absolute paths, **proved on disk by
counting both the removed and the injected text before any verdict was read**, then restored and the
restore proved by `git hash-object` equal to the HEAD blob plus an empty `git diff HEAD`. No build step
sits between mutation and measurement — the artefacts this gate reads *are* the mutated JSON and
markdown, so no `dist` can hold a stale copy and the dist-preflight rule does not apply and is not
claimed.

| leg | mutation, proved on disk | verdict |
|---|---|---|
| **A — an eval token changed to an untaught word** | `evals/page-builder.json`: `"SchemaRenderer"` 1 to 0, `"SchemaRendererNobodyTeachesThis"` 0 to 1 | **exit 1**, `skills/objectui/evals/page-builder.json eval 1 [untaught] "SchemaRendererNobodyTeachesThis" (it appears nowhere in the bundle, in any form)`. Restored, blob `9669f174`, `git diff HEAD` empty |
| **B — a guide's only occurrence of a taught token removed** | `guides/plugin-development.md`: whole-token `registerComponent` in the bundle 1 to 0, injected `registerComponentHandler` 0 to 1, **substring still present (1)** | **exit 1**, three rows: `page-builder.json eval 1`, `plugin-development.json` evals 1 and 2, each `(it appears 1 time(s) as a SUBSTRING only)`. This leg doubles as proof that the whole-token rule is what catches it. Restored, blob `398f8f56`, `git diff HEAD` empty |
| **C — a baseline row deleted while its red persists** | leg B's mutation held; 3 real-corpus baseline rows 0 to 3 (`exit 0`, "Red under the chosen oracle: 3 (0 beyond the baseline)"); then one row deleted, 1 to 0 | **exit 1** naming exactly the deleted row, "Red under the chosen oracle: 3 (1 beyond the baseline)". Both files restored, blobs `398f8f56` / `680ce368`, `git diff HEAD` empty |

Restored-tree re-measure: gate exit 0, self-test exit 0, `git diff HEAD` empty.

⚠️ Recorded rather than smoothed over: leg C's **first** attempt aborted on its own on-disk check
(`BASELINE EDIT DID NOT LAND`, exit 91) because the expected count was miscalibrated — the string it
counted already occurred once in a self-test fixture. The edit *had* landed; the assertion was wrong.
The trap restored cleanly and the leg was re-run against a specific anchor. The guard did its job, and
silently re-running until something landed would have been the defect this discipline exists to
prevent.

## Gates

Each verdict below is the gate's own printed line; every exit code was captured by redirect **before**
any pipe.

| command | exit | its own verdict line |
|---|---|---|
| `node scripts/check-skill-eval-tokens.mjs` | 0 | `Every must_contain token is taught by its own skill bundle.` — 1 bundle, 11 eval files, 33 evals, 125 tokens, 16 guide files; red 0, baseline 0 |
| `node scripts/check-skill-eval-tokens.mjs --self-test` | 0 | `OK check-skill-eval-tokens self-test: 29 cases pass` |
| `node scripts/check-skill-eval-tokens.mjs --measure` | 0 | bundle-wide 0 of 125; per-guide 14 of 125, `12 exist only because the basename-matching guide is MISSING and 2 name a token taught elsewhere in the bundle` |
| `pnpm exec vitest run --maxWorkers=2 scripts/__tests__` (under the shared verify lock) | 0 | `Test Files 100 passed (100); Tests 2866 passed (2866)` |
| `node scripts/check-skill-examples.mjs` (after the filtered 11-package build it derives) | 0 | `Every marked skill example holds up against the built types.` — Semantic phase 17 of 17 ts fences judged, 0 failed; JSON phase 39 parsed, 0 failed; all four harness controls green |
| `node scripts/check-pre-install-import-graph.mjs` | 0 | `OK — 23 pre-install step(s) in 18 job(s) run 21 scripts/ gate(s); 23 module(s) walked, every non-relative leaf a node builtin.` |
| `node scripts/check-changeset-presence.mjs` | 0 | `No source or published contract of a released package changed in this range, so no changeset is owed.` — its own verdict decides; **no changeset added** |
| `node scripts/check-governed-queue-guard.mjs --test PATHS` | 0 | `NOT GOVERNED — 7 path(s) checked against 5 governed surface(s); none matched.` The diff touches zero `skills/` files, so GOVERNED is correctly *not* the verdict. **The PR still stays DRAFT** per the dispatch; the flip is the seat's step |
| `node scripts/check-control-bytes.mjs` | 0 | `OK (scanned 6144 tracked text file(s); skipped 85 binary).` |
| `node scripts/check-shell-escape-residue.mjs` | 0 | `OK (5/5 root(s) resolved ... 206 file(s) and 1292 fenced block(s) examined in total)` |
| `node scripts/check-doc-links.mjs` | 0 | `Links are valid across 17 scan roots.` |
| `node scripts/check-doc-fence-languages.mjs` | 0 | `every TypeScript block in 227 document(s) is fenced ts/tsx/typescript, except 80 declared file(s) carrying 90 block(s) (SHRINK-ONLY)` |
| `node scripts/check-skills-paths.mjs` | 0 | `OK (88/89 stated path(s) resolve across 20 guide file(s); 1 baselined)` |
| `node scripts/check-lint-coverage.mjs` | 0 | `lint coverage: 46/46 packages linted, 0 with outstanding errors (0 total).` |
| `pnpm exec tsc -p tsconfig.scripts.json --noEmit` | 0 | clean — this is what compiles the new pin suite |
| `pnpm exec eslint --format json` over the 4 changed lintable files | 0 | `4 files linted, 0 errors, 0 warnings` |

**Gate-family derivation.** `scripts/pm/dispatch-gates.mjs` is absent from this tree and says so
itself: *"this script exists only in `objectstack-ai/objectstack`, so a sister repo's list is
hand-derived from its own package manifest and its own workflow files"* (it refused with
`REFUSING — asked for 'objectstack-ai/objectui', but this checkout is 'objectstack-ai/objectstack'`).
Hand-derived here from objectui's own `package.json` and `.github/workflows/`: a new workflow file plus
an edited `dependabot-merge-gate.mjs` pull in 50 suites that read `.github/workflows` and 10 that read
`ci-cd-pipeline.md`, so rather than guess which matter the **whole `scripts/__tests__` directory** was
run — 100 files, 2866 tests, the cheap superset. That also discharges the separate obligation an edited
gate script owes: `dependabot-merge-gate.test.ts` and `merge-queue-reporting.test.ts` are its own pin
suites and both ran.

## NOT MEASURED, and why

- **`pnpm lint` (repo-wide `eslint .`) was narrowed** to the 4 changed lintable files. Declared, with
  the three pieces of evidence: (1) the population is read from eslint's own config, not a guess —
  `pnpm lint` is `turbo run lint`, i.e. each workspace package's own `eslint .`, and `scripts/` has no
  `package.json` so it is in no package's lint run at all (the structural gap
  `scripts-type-check.test.ts` records for type-check); (2) the file count is read from
  `--format json`; (3) invariance — `eslint.config.js` configures **no** type-aware linting (no
  `project`, no `projectService` anywhere in it), so this diff cannot move the verdict for any
  untouched file. `check-lint-coverage.mjs` ran clean over all 46 packages independently.

---

https://claude.ai/code/session_01LraLgQVGq8egUwfYZpbYt1

---
_Generated by [Claude Code](https://claude.ai/code/session_01LraLgQVGq8egUwfYZpbYt1)_

### DIFF
--- a/.github/workflows/skill-eval-tokens.yml
+++ b/.github/workflows/skill-eval-tokens.yml
@@ -1,0 +1,91 @@
+name: Skill Eval Tokens
+
+# Holds every `must_contain` token of every `skills/<bundle>/evals/*.json` to
+# the guides that bundle actually ships: the token must occur as a WHOLE token
+# (never a substring) somewhere in the bundle's own markdown. The oracle choice
+# — bundle-wide over per-guide, argued on four axes and measured both ways —
+# the whole-token boundary rule, the `must_not_contain` polarity and the
+# declared exit codes are documented at length in
+# `scripts/check-skill-eval-tokens.mjs`.
+#
+# ── Why this is its own workflow, with NO path filter ───────────────────────
+#
+# Same reason `skills-paths.yml`, `skill-examples.yml`, `docs-links.yml` and
+# `control-bytes.yml` are theirs. This gate's whole input is the markdown and
+# JSON under `skills/`, and both `ci.yml` and `lint.yml` list `'**/*.md'`,
+# `content/**` and `docs/**` under the `paths-ignore` of their `push` trigger.
+# GitHub has no per-job path filter, so a push that only edits a guide would
+# start neither — and editing only a guide is exactly how a token an eval grades
+# stops being taught. `control-bytes.yml`'s header names the consequence: a gate
+# that cannot see a markdown-only change "rebuilds the hole it exists to close".
+#
+# Hence: no `paths` and no `paths-ignore` here, deliberately.
+# `scripts/__tests__/check-skill-eval-tokens.test.ts` fails if either is ever
+# added, and fails too if a second workflow starts running the same script —
+# one gate, one home.
+#
+# ── Why it does NOT install, unlike its sibling `skill-examples.yml` ────────
+#
+# `skill-examples.yml` installs and builds because its criterion is the
+# PUBLISHED type surface, so the packages its marked fences import have to exist
+# as `dist/*.d.ts` first. This gate's criterion is text: it reads eval JSON and
+# guide markdown out of the checkout and nothing else. So it takes
+# `skills-paths.yml`'s shape instead — a checkout plus two `node` calls — and
+# `check-pre-install-import-graph.mjs` picks it up automatically and enforces
+# that this script's whole static import graph stays free of `node_modules`.
+#
+# That install-free shape is also why folding this check into
+# `check-skill-examples.mjs` as a second phase was rejected. That script exits 2
+# (PRECONDITION NOT MET) on an unbuilt tree BEFORE judging anything, so an
+# eval-token phase inside it would simply never run wherever the workspace is
+# not built — and its own header already lists this oracle, in "Deliberately NOT
+# answered here" item 4, as a different question over a different corpus.
+
+on:
+  pull_request:
+    branches: [main, develop]
+  push:
+    branches: [main, develop]
+  # Merge queue (objectui#3523 — see `ci.yml`'s trigger block for the full note
+  # and the measurements behind it). A required check that does not report on a
+  # queue build stalls the queue until the ruleset's 60-minute timeout fails it,
+  # so an unfiltered gate that could become required subscribes here from the
+  # start. `types:` is named although `checks_requested` is currently the only
+  # activity type GitHub defines for `merge_group`.
+  merge_group:
+    types: [checks_requested]
+  workflow_dispatch:
+
+concurrency:
+  group: skill-eval-tokens-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  skill-eval-tokens:
+    name: Skill Eval Token Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: '22.x'
+
+      # The self-test runs FIRST. A probe that cannot fail is not a probe: it
+      # plants an untaught token that must go red, a substring-only match that
+      # must go red, a `must_not_contain` entry that must NOT, and a baseline row
+      # that must suppress exactly its own row — so a gate broken into permanent
+      # green is caught here rather than by nobody.
+      - name: Self-test the oracle and the whole-token rule, both directions
+        run: node scripts/check-skill-eval-tokens.mjs --self-test
+
+      # Reads the checkout and nothing else, so no install is required.
+      - name: Check every eval must_contain token against the guides
+        run: node scripts/check-skill-eval-tokens.mjs
--- a/content/docs/guide/ci-cd-pipeline.md
+++ b/content/docs/guide/ci-cd-pipeline.md
@@ -31,6 +31,7 @@ one has its own section below.
 | `docs-links.yml` | Internal Docs Link Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** |
 | `skills-paths.yml` | Skill Guide Path Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when a path stated in a `skills/` guide does not exist |
 | `skill-examples.yml` | Skill Example Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when a MARKED fenced example in a `skills/` guide no longer compiles against the packages' built types, no longer parses as JSON, or carries a marker that opts nothing in |
+| `skill-eval-tokens.yml` | Skill Eval Token Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when an eval assertion's `must_contain` token is not taught as a whole token anywhere in its own `skills/` bundle |
 | `doc-component-types.yml` | Doc Component Type Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when a `content/docs/**.mdx` snippet teaches a `type` nothing registers |
 | `doc-snippet-types.yml` | Doc Snippet Type Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when a covered documentation snippet no longer compiles against the packages' built types |
 | `doc-fence-languages.yml` | Doc Fence Language Check | Push / PR to `main`, `develop` — **no path filter**; merge-queue builds; manual | **Yes** — when a TypeScript block sits under a fence the snippet gate does not read |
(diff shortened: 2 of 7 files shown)


os-litant:
CI note from the skills-lane seat (session `session_01LraLgQVGq8egUwfYZpbYt1`), 2026-09-03T08:28Z — `Test (shard 3/4)` is red at `0ee77b8` (run 33731676085, job 100572795893) on a file this PR does not touch, and the seat is spending the PR's single re-run on it rather than treating it as this diff's.

- **What failed**: `packages/app-shell/src/views/RecordDetailView.approvalDeclaredActions.test.tsx` › "record page decision actions — a GROUP approver (objectui#3055) › folds a declared decision output into the nested `outputs` body" — the network-escape guard (`vitest.setup.network-escape-guard.ts:227`) threw because the component under test reached a real socket at `http://localhost:3000/api/v1/approvals/requests?…`. 1 failed / 7941 passed in the shard.
- **Why it is not this PR's**: the diff is `scripts/check-skill-eval-tokens.mjs` + its spec, `.github/workflows/skill-eval-tokens.yml`, `scripts/dependabot-merge-gate.mjs` (+2 lines), `scripts/__tests__/merge-queue-reporting.test.ts` (+9), `content/docs/guide/ci-cd-pipeline.md` and `package.json` — nothing under `packages/**`, nothing that a `RecordDetailView` test imports, no vitest setup or config change. The guard's own message classifies the shape as objectui#6640 (happy-dom's default document URL turning a relative fetch into a live TCP connection) and says the fix is a fetch double in that test file, which is a different card's work. The other three shards, `Lint`, `Type Check`, `Skill Example Check` and the new `Skill Eval Token Check` are green on this head.
- **What the seat is doing**: one re-run of the failed job on this exact commit. A second failure is treated as real and this PR waits; a pass lets the ACCEPT proceed. ⛔ Not adding the file to `KNOWN_ESCAPES` (that list only shrinks) and not touching the test here.

---

PR #7503, opened 2026-09-03
fix(data-objectstack): createObjectStackAdapter declares the adapter it returns

### LINKED ISSUE #7323: finding(data-objectstack): `createObjectStackAdapter` declares `DataSource` as its return, so the whole adapter API its own README documents is unreachable in TypeScript
Found while bringing `packages/data-objectstack/README.md` under `check:doc-snippet-types` (#5174 batch 8). Filed rather than fixed: `packages/*/src` is outside that PR's surface.

## What the shipped declaration says

`packages/data-objectstack/dist/index.d.ts`:

```
declare function createObjectStackAdapter<T = unknown>(config: { ... }): DataSource<T>;
```

`DataSource` is `@object-ui/types`' interface (`packages/types/dist/data.d.ts:292`). It declares `find`, `findOne`, `create`, `update`, `delete`, `bulk`, `batchTransaction`, `getObjectSchema`, `aggregate` and the import/export job surface — and **none** of the adapter-only members.

## The measured consequence

Compiled against the built `dist` with the doc-snippet gate's own compiler options and module resolution:

```
import { createObjectStackAdapter } from '@object-ui/data-objectstack';
declare const dataSource: ReturnType<typeof createObjectStackAdapter>;
dataSource.getClient();
dataSource.getCacheStats();
dataSource.invalidateCache('users');
dataSource.clearCache();
```

```
TS2339 Property 'getClient' does not exist on type 'DataSource<unknown>'.
TS2339 Property 'getCacheStats' does not exist on type 'DataSource<unknown>'.
TS2339 Property 'invalidateCache' does not exist on type 'DataSource<unknown>'.
TS2339 Property 'clearCache' does not exist on type 'DataSource<unknown>'.
```

The same four, plus `getConnectionState`, `isConnected`, `onConnectionStateChange` and `onBatchProgress`, are exactly the members `packages/data-objectstack/README.md`'s own **API Reference — ObjectStackAdapter — Methods** list documents, and that four whole README sections (Metadata Caching, Connection State Monitoring, Batch Operation Progress, Troubleshooting/Cache Issues) are built around. The declaration typing the exported `ObjectStackAdapter` class carries all of them; only the factory's declared return narrows them away.

So a reader who follows the README's own Basic Setup — `const dataSource = createObjectStackAdapter({ baseUrl })` — and then follows any of those four sections does not compile. The runtime object has the methods; the declared type does not admit them.

## Why this is a producer-side gap, not a doc bug

The methods are real, exported and documented; there is nothing to correct on the page. The cast-free escape a consumer has today is to construct the class instead (`new ObjectStackAdapter(config)`), which the README's own API Reference documents as public — so the two supported ways of obtaining the same object give different type surfaces.

## Options

- **A. Widen the factory's declared return** to the adapter's own type: `createObjectStackAdapter(config): ObjectStackAdapter` (or `DataSource<T> and ObjectStackAdapter`). One line, and it makes declared match shipped for every documented member at once.
- **B. Leave the narrow return and add the missing members to `DataSource`.** Wrong shape: caching, connection state and batch progress are this adapter's concerns, not every data source's, and every other `DataSource` implementation would then be declaring members it does not have.
- **C. Document the cast.** Cheapest, and the worst of the three under the "declared = enforced" rule — it teaches a cast around a declaration that is simply narrower than the value.

Recommendation: **A**.

## Where this shows up now

Batch 8 ships those README blocks compiling, by declaring the adapter as the exported `ObjectStackAdapter` class — truthful about the object, and about the one supported way to obtain that type today — and adds a short note to the README's setup section naming this card, so the page does not silently teach a shape the factory's return will reject. The note comes out when this is fixed.

os-project-manager:
Fixes #7323

`createObjectStackAdapter` returned `new ObjectStackAdapter(config)` while declaring the shared `DataSource` interface as its return. A wider value is assignable to a narrower annotation, so nothing ever failed to compile — the loss was entirely on the reading side: every adapter-only member was erased from the type the factory hands back, while staying on the object it hands back.

**Measured before the change**, compiling against the shipped `packages/data-objectstack/dist/index.d.ts` with the doc-snippet gate's own compiler options (strict, bundler resolution): **9 x TS2339**, not the 4 the card lists — `getClient`, `getCacheStats`, `invalidateCache`, `clearCache`, `getConnectionState`, `isConnected`, `onConnectionStateChange`, `onBatchProgress`, `setSystemCapabilities`. Eight of those are exactly the members the README's **API Reference — Methods** list documents; the ninth is the one the factory's own JSDoc links to (`[ADR-0066] See ObjectStackAdapter.setSystemCapabilities`). The same probe after this change: **0 diagnostics**.

## The two questions that decided the shape

The triage set these as a stop-condition and both are answered from the code, on `origin/main` `e17605309`.

**Q1 — is `ObjectStackAdapter` exported from the package's public entry? Yes, already.** `packages/data-objectstack/src/index.ts:2148` reads `export class ObjectStackAdapter`, and `tsup.config.ts` has exactly one entry, `src/index.ts`. The class is in the shipped `dist/index.d.ts` export list (verified in the built artifact, not inferred). Two pin tests already assert the exported spelling in source — `cloud-surface-retired-4152.pin.test.ts:114` and `v3-deep-integration-retired-4241.pin.test.ts:135` — and `apps/console/src/dataSource.ts:14` re-exports it by name. **So this PR exports nothing by implication**: the export list in `dist/index.d.ts` is unchanged, name for name, before and after.

**Q2 — is the narrow return deliberate? No, and there are two pieces of evidence against it.**

1. Commit `a5d817061` ("Add connection state monitoring, auto-reconnect, and batch progress") added `autoReconnect` / `maxReconnectAttempts` / `reconnectDelay` to the **factory's own config bag** while leaving the members that observe those features off the factory's declared return. The same change configured a feature it made unobservable through its own entry point. That is an oversight shape, not an encapsulation decision.
2. The factory's JSDoc, added by `41b7ce3ce`, points the reader at `ObjectStackAdapter.setSystemCapabilities` — a member its declared return hides.

Nothing anywhere pins the return: no comment, no ADR, no test, no `Equal` assertion, and no swappability note. Searched the export list, `package.json` `files` / `exports`, ADR-0066, the sibling adapters, and every test that names the factory.

Had either answer come back the other way this would have been a docs-only PR. They did not, so it is option **A** — and swappability, the property the narrow return was suspected of protecting, is not lost: a wider return is assignable to the narrower annotation, so `const ds: DataSource = createObjectStackAdapter(...)` still compiles and still gives the narrow surface. There is a control test that says exactly that.

## Why A and not B or C

**B** (add the missing members to `DataSource`) would make every other `DataSource` implementation declare caching, connection state and batch progress it does not have — those are this adapter's concerns. **C** (document a cast) teaches a cast around a declaration that is merely narrower than the value, which is the opposite of `declared = enforced`. **A** is one line and makes declared match shipped for every documented member at once.

## The pin

`packages/data-objectstack/src/adapterFactoryReturn.types.test.ts` — the card's TS2339 reproduction, inverted. It reads all nine members through the factory's return, asserts the return's identity with `Equal` (not mere assignability — two adapter-shaped declarations are mutually assignable, so only identity can tell "returns THE adapter" from "returns something adapter-shaped"), and carries **two controls**:

- `_NotOnDataSource` — the adapter-only members stay ABSENT from the shared `DataSource`. This is what makes the reads a statement about the factory's return rather than about every data source, and it fires on option B.
- `_StillADataSource` — the widened return is still assignable to `DataSource`.

Both controls are independent of the return annotation, so reverting the source change turns the reads red and leaves the controls green. That split is measured below, not asserted.

**Which tree the pin exercises.** Two, deliberately. The test is compiled by `pnpm --filter @object-ui/data-objectstack type-check`, whose program includes the whole `src/**` (tests included) and resolves `./index` from **`src`** — proven with `--listFiles`: 1 hit for the new file, 55 test files, `src/index.ts` present. `@object-ui/types` in that same program resolves from **`packages/types/dist`**, so the type used for the controls is the shipped one. The README blocks are the **`dist`** half: `check:doc-snippets` compiles them against the built `dist/*.d.ts` (its own resolution control prints `'@object-ui/types' was successfully resolved to packages/types/dist/index.d.ts`).

## Documentation

The README note naming this card (added by #5174 batch 8) is removed. The four sections built on the adapter-only members — Metadata Caching, Connection State Monitoring, Batch Operation Progress, Troubleshooting → Cache Issues — now continue from Basic Setup's `createObjectStackAdapter(...)` call instead of hand-declaring the class, so the page teaches one shape and the doc-snippet gate pins the fix against `dist`. The other README blocks still declare the class where they only need `DataSource`-level members; that is truthful (the class is public and documented) and rewriting them buys nothing, so they are left alone.

`setSystemCapabilities` is still absent from the README's Methods list. It was absent before this PR too, it is a separate doc gap, and adding it here would be scope this card did not ask for.

## Verification — every number below was observed, at head `922ca54ec`

| Check | Command | Result |
| --- | --- | --- |
| Package type-check | `pnpm --filter @object-ui/data-objectstack type-check` | exit 0, `tsc --noEmit` echoed |
| Package tests | `pnpm exec vitest run --maxWorkers=2 packages/data-objectstack/` (repo root — a package-dir run is refused by objectui#3378's guard) | **55 files, 737 tests, all passed** |
| `check:doc-snippets` | `node scripts/check-doc-snippet-types.mjs` | exit 0 — *"Semantic phase: 455 of 455 block(s) judged, 0 failed."* / *"Every covered documentation snippet compiles against the built types."* |
| `check:readme-exports` | `pnpm check:readme-exports` | exit 0 — *"OK (43 tracked README(s) ...; 414 self-imports judged (414 real, 0 wrong-path, 0 fabricated); 3305 export symbol(s) read from 37 of 40 tracked package(s) (0 unbuilt ...))"* |
| changeset presence | `node scripts/check-changeset-presence.mjs` | exit 0 — *"1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)"* |
| `check:doc-fences` | `node scripts/check-doc-fence-languages.mjs` | exit 0 — *"every TypeScript block in 227 document(s) is fenced ts/tsx/typescript ..."* |
| control bytes | `node scripts/check-control-bytes.mjs` | exit 0 — *"OK (scanned 6180 tracked text file(s); skipped 85 binary)"* |
| Package lint | `pnpm --filter @object-ui/data-objectstack lint` | exit 0 — **423 problems, 0 errors, 423 warnings**, all pre-existing `no-explicit-any`; the new file contributes **0 findings** (`eslint --format json` on the two changed source files: new test 0/0) |

Every gate result above is quoted from the gate's own verdict line; exit codes were captured by redirecting first and reading `$?` before any pipe.

Dependency closure built first (`pnpm --filter '@object-ui/data-objectstack...' build`, then the doc-snippet gate's own derived `--build-filter`: 26 packages, 34 turbo tasks, then `--filter='./packages/*'` for `check:readme-exports`, which had reported 2 unbuilt packages — a *"could not run"*, not a defect).

**Lint scope, declared:** the table's lint row is the changed package only, not the repo. `pnpm lint` here is `turbo run lint` (each package's own `eslint .`), and CI runs it in full regardless. One reading that needs stating: a separate probe with `eslint --no-inline-config` reports **4 errors** in `src/index.ts` at lines 1369–1389. Those are the documented `eslint-disable-next-line no-console` sites for the spec `Logger` binding (objectui#4029), they are ~4300 lines from either of my edits, and they are invisible to the lint CI actually runs because it honours inline config. Not introduced here, and not silently omitted either.

## Reverse verification — red/green split, both directions proven on disk

Method: commit first, then mutate `packages/data-objectstack/src/index.ts` back to the narrow return under `trap ... EXIT INT TERM` with an absolute restore path, prove the mutation landed by counting the injected and removed text (not by the editor's exit code) and by comparing blob hashes, then restore and prove the restore **by state**.

Mutation landed: injected-text count 0 → 1, removed-text count 1 → 0; mutated blob `447fffc8` differs from HEAD blob `bad31df4`.

**Leg 1 — the src pin, no rebuild needed** (the pin imports `./index` from `src`; the mutation cannot reach `packages/types/dist`, which is where its `DataSource` comes from):

- 13 diagnostics total, **every one of them inside the pin file**, 0 anywhere else in the package.
- 10 x TS2339 at lines 75–84: the nine member reads (`invalidateCache` twice, with and without an argument).
- TS2344 at 89 and 113: the `_HasHiddenMembers` and `_IsTheAdapter` assertions.
- TS7053 at 157: the runtime member sweep can no longer index the narrowed type.
- **The controls at lines 120 (`_NotOnDataSource`) and 135 (`_StillADataSource`) produced ZERO diagnostics** — the split the pin was designed for.
- `vitest` on the same mutated tree still passes, which is the point: esbuild erases types, so the colour comes from `tsc` and from nowhere else.

**Leg 2 — the dist pin, rebuilt in both directions.** Rebuilt from the mutated source, then a dist preflight: the widened marker count in `dist/index.d.ts` was **0** (the mutation reached `dist`). The standalone dist probe then reproduced the card's failure exactly — **9 x TS2339** — and `check:doc-snippets` went to **exit 1, "455 of 455 block(s) judged, 4 failed"**: precisely the four README sections this PR rewired, failing on `getCacheStats`, `invalidateCache` x2, `clearCache`, `onConnectionStateChange`, `getConnectionState`, `isConnected`, `onBatchProgress`, `clearCache`, `invalidateCache`.

**Restore, proven by state, both halves.** Source: restored blob `bad31df4` equals HEAD's blob and `git diff HEAD` is 0 bytes. `dist`: rebuilt again, widened marker back to **1** and narrow marker **0**, dist probe back to **0 diagnostics**, `check:doc-snippets` back to **exit 0, 0 failed**. The restore leg is rebuilt on purpose — a mutated marker left in `dist` would keep acting on every later run in this tree.

## Clause ②

**`Clause-②: yes`**, declared in the claim comment on #7323 in the fixed machine spelling, and `needs:contract-review` is on both carriers (this PR and the card). It widens the declared public surface reachable through a published entry point. Q1's answer removes the triage's specific worry — no class is exported by implication — but the declaration limb is judged from content, not from paths or from diff size.

**The one compatibility note a reviewer should weigh.** The widened return is a class with private members, so it is no longer satisfied by a hand-written structural stand-in: an object literal assigned to the factory's `ReturnType` will now be rejected where it used to be accepted. Nothing in this repo does that (searched every test and app that names the factory — `apps/console/src/dataSource.ts` only re-exports it), and the fix for a downstream consumer is to annotate such a fake as `DataSource`, which is what it was standing in for. It is the only direction in which this change is not purely additive, and it belongs in the contract review rather than in a footnote.

Draft on purpose, not flipped ready, auto-merge not enabled — the seat lands it.

---

### DIFF
--- a/.changeset/7323-adapter-factory-return.md
+++ b/.changeset/7323-adapter-factory-return.md
@@ -12,12 +12,32 @@ A wider value is assignable to a narrower annotation, so nothing ever failed to 
 `ReturnType<typeof createObjectStackAdapter>` failed with TS2339: `getClient`,
 `getCacheStats`, `invalidateCache`, `clearCache`, `getConnectionState`, `isConnected`,
 `onConnectionStateChange`, `onBatchProgress` and `setSystemCapabilities`. Eight of those
-are exactly the members this package's README API Reference documents, and four whole
-README sections are built on them; the ninth is the one the factory's own JSDoc links to
-(`[ADR-0066] See {@link ObjectStackAdapter.setSystemCapabilities}`). So the file's own
-doc comment pointed the reader at a method its declared return hid, and the two
-documented ways to obtain the same object — the factory and `new ObjectStackAdapter(…)`
-— handed back different type surfaces.
+nine reads are on this package's README API Reference list, and four whole README
+sections are built on them; the ninth measured read is the one the factory's own JSDoc
+links to (`[ADR-0066] See {@link ObjectStackAdapter.setSystemCapabilities}`). The README
+list is itself **nine** adapter-only members, not eight — `connect()` is adapter-only
+too and was documented all along; it simply was not one of the reads the card's
+reproduction measured. So the file's own doc comment pointed the reader at a method its
+declared return hid, and the two documented ways to obtain the same object — the factory
+and `new ObjectStackAdapter(…)` — handed back different type surfaces.
+
+**What the declared return now is: the whole class, not those nine reads.** The nine
+above are what the reproduction measured, not the size of this change. The factory's
+declared return is now `ObjectStackAdapter<T>` itself, so **every public member of the
+class** is part of what the factory promises. Against `DataSource` that is **20**
+members, not nine — `tsc`-computed as
+`Exclude<keyof ObjectStackAdapter<unknown>, keyof DataSource<unknown>>`: `clearCache`,
+`connect`, `getCacheStats`, `getCached`, `getClient`, `getConnectionState`,
+`getDiscovery`, `getItems`, `invalidateCache`, `invalidateViewKeys`, `isConnected`,
+`listImportMappings`, `onBatchProgress`, `onConnectionStateChange`, `onSaveAdvisory`,
+`onWriteWarning`, `probeAppAccess`, `queryDataset`, `setSystemCapabilities`,
+`updateDashboard`. The eleven past the documented nine were already in the shipped class
+type — none is `@internal` or `@deprecated`, `stripInternal` is not set, and all were
+already reachable through `new ObjectStackAdapter(…)` and through every
+`ObjectStackAdapter`-typed seam in `@object-ui/react` and `app-shell` — so what widens
+here is what the **factory declares**, not what the package ships. Two are escape-hatch
+shaped and worth knowing before building on them: `getCached(key)` is a raw cache read,
+and `getDiscovery()` reaches an internal property of the underlying `ObjectStackClient`.
 
 **Branch taken: A (widen the factory's declared return), and why.** The card offered
 three. B — moving caching, connection state and batch progress onto `DataSource` — was
@@ -37,18 +57,23 @@ the commit that added `autoReconnect` / `maxReconnectAttempts` / `reconnectDelay
 factory's own config bag left the members that observe those features off the factory's
 declared return in the same change.
 
-**Not a breaking change for callers.** A wider return is assignable to the narrower
-annotation, so `const ds: DataSource = createObjectStackAdapter(…)` keeps compiling and
-keeps giving the narrow surface to anyone who wants it. The one shape that changes is a
-hand-written object literal assigned to `ReturnType<typeof createObjectStackAdapter>`:
-that type is now a class with private members, so a structural stand-in no longer
-satisfies it — annotate such a fake as `DataSource` instead, which is what it was
-standing in for.
+**One caller shape breaks: a structural stand-in for the factory's return.** A
+hand-written object literal annotated `ReturnType<typeof createObjectStackAdapter>` no
+longer satisfies that type, because it is now a class with private members (TS2740) —
+annotate such a fake as `DataSource` instead, which is what it was standing in for.
+Nothing else moves: a wider return is assignable to the narrower annotation, so
+`const ds: DataSource = createObjectStackAdapter(…)` keeps compiling and keeps giving
+the narrow surface to anyone who wants it.
 
 The README's note saying the page could not yet teach the factory's shape is removed, and
 the four sections built on the adapter-only members (Metadata Caching, Connection State
 Monitoring, Batch Operation Progress, Troubleshooting → Cache Issues) now continue from
 Basic Setup's `createObjectStackAdapter(…)` call instead of declaring the class by hand.
-`src/adapterFactoryReturn.types.test.ts` pins the card's TS2339 reproduction inverted,
+The docs-site page `content/docs/utilities/data-objectstack.mdx` is corrected the same
+way: its prose, its factory signature fragment and its "hold the class type to reach
+these" section described the old narrow return, and its Mutations and Troubleshooting
+examples told the reader to construct the class by hand to reach members the factory now
+declares. `src/adapterFactoryReturn.types.test.ts` pins the card's TS2339 reproduction
+inverted,
 with two controls: the adapter-only members stay absent from `DataSource` (fires on
 option B), and the widened return stays assignable to `DataSource` (swappability kept).
--- a/.changeset/7323-adapter-factory-return.md
+++ b/.changeset/7323-adapter-factory-return.md
@@ -1,0 +1,54 @@
+---
+'@object-ui/data-objectstack': minor
+---
+
+`createObjectStackAdapter` declares the adapter it returns, not the shared `DataSource`
+interface (objectui#7323).
+
+The factory returned `new ObjectStackAdapter(config)` while declaring `DataSource<T>`.
+A wider value is assignable to a narrower annotation, so nothing ever failed to compile
+— the loss was entirely on the reading side. Measured against the shipped
+`dist/index.d.ts` with the doc-snippet gate's own compiler options, nine reads through
+`ReturnType<typeof createObjectStackAdapter>` failed with TS2339: `getClient`,
+`getCacheStats`, `invalidateCache`, `clearCache`, `getConnectionState`, `isConnected`,
+`onConnectionStateChange`, `onBatchProgress` and `setSystemCapabilities`. Eight of those
+are exactly the members this package's README API Reference documents, and four whole
+README sections are built on them; the ninth is the one the factory's own JSDoc links to
+(`[ADR-0066] See {@link ObjectStackAdapter.setSystemCapabilities}`). So the file's own
+doc comment pointed the reader at a method its declared return hid, and the two
+documented ways to obtain the same object — the factory and `new ObjectStackAdapter(…)`
+— handed back different type surfaces.
+
+**Branch taken: A (widen the factory's declared return), and why.** The card offered
+three. B — moving caching, connection state and batch progress onto `DataSource` — was
+rejected because those are this adapter's concerns, not every data source's; every other
+`DataSource` implementation would then declare members it does not have. C — documenting
+a cast — teaches a cast around a declaration that is merely narrower than the value,
+which is the opposite of `declared = enforced`. A is one line and makes declared match
+shipped for every documented member at once.
+
+Two questions decided the shape and both were answered from the code before the diff.
+`ObjectStackAdapter` was **already** exported from the package's only entry
+(`src/index.ts`, tsup's single entry; the class is in the shipped `dist/index.d.ts`
+export list, two pin tests assert the exported spelling, and `apps/console` re-exports it
+by name) — so widening the return exports nothing by implication. And the narrow return
+was **not** a deliberate swappability guarantee: no comment, ADR or test pinned it, and
+the commit that added `autoReconnect` / `maxReconnectAttempts` / `reconnectDelay` to the
+factory's own config bag left the members that observe those features off the factory's
+declared return in the same change.
+
+**Not a breaking change for callers.** A wider return is assignable to the narrower
+annotation, so `const ds: DataSource = createObjectStackAdapter(…)` keeps compiling and
+keeps giving the narrow surface to anyone who wants it. The one shape that changes is a
+hand-written object literal assigned to `ReturnType<typeof createObjectStackAdapter>`:
+that type is now a class with private members, so a structural stand-in no longer
+satisfies it — annotate such a fake as `DataSource` instead, which is what it was
+standing in for.
+
+The README's note saying the page could not yet teach the factory's shape is removed, and
+the four sections built on the adapter-only members (Metadata Caching, Connection State
+Monitoring, Batch Operation Progress, Troubleshooting → Cache Issues) now continue from
+Basic Setup's `createObjectStackAdapter(…)` call instead of declaring the class by hand.
+`src/adapterFactoryReturn.types.test.ts` pins the card's TS2339 reproduction inverted,
+with two controls: the adapter-only members stay absent from `DataSource` (fires on
+option B), and the widened return stays assignable to `DataSource` (swappability kept).
--- a/content/docs/utilities/data-objectstack.mdx
+++ b/content/docs/utilities/data-objectstack.mdx
@@ -178,7 +178,7 @@ const fromFactory = createObjectStackAdapter<User>({ baseUrl: 'https://api.examp
 const fromClass = new ObjectStackAdapter<User>({ baseUrl: 'https://api.example.com' });
 ```
 
-**What changed.** Until v17.7 this section told you to hold the class type to
+**What changed.** This section used to tell you to hold the class type to
 reach the members listed under *Beyond `DataSource`* below, because the factory
 declared `DataSource<T>` while returning `new ObjectStackAdapter(config)`: the
 value always carried those members, only the declared type hid them
@@ -210,16 +210,15 @@ adapter):
 - `getObjectSchema(objectName)` - Fetch schema metadata (cached)
 - `bulk?(resource, operation, data)` - Batch create/update/delete on one object
 - `batchTransaction?(operations)` - Cross-object atomic batch (master-detail)
+- `onMutation?(listener)` - Subscribe to create/update/delete events
 
 `bulk`, `batchTransaction` and `onMutation` are **optional** members of
 `DataSource`: not every adapter implements them, so through a `DataSource`-typed
 value they must be feature-detected (`typeof dataSource.bulk === 'function'`).
 `ObjectStackAdapter` implements all three unconditionally, and the factory
-declares the class, so a value from either form calls them directly:
-
-- `bulk(resource, operation, data)` - Batch create/update/delete on one object
-- `batchTransaction(operations)` - Cross-object atomic batch (master-detail)
-- `onMutation(listener)` - Subscribe to create/update/delete events
+declares the class, so a value from either form calls them directly. (`onMutation`
+was listed under *Adapter-only* here until objectui#7323; it is optional on
+`DataSource`, not absent from it.)
 
 **Beyond `DataSource`** (declared on `ObjectStackAdapter`, so reachable from
 either form):
@@ -231,10 +230,12 @@ either form):
 - `onConnectionStateChange(listener)` - Subscribe to state changes (returns unsubscribe)
 - `onBatchProgress(listener)` - Subscribe to bulk progress (returns unsubscribe)
 
-Those nine are the documented subset. `ObjectStackAdapter` declares **20** members
-beyond `DataSource` in all; the other eleven are lower-level seams (client
-discovery, cache-key invalidation, dataset and dashboard access, advisory
-subscriptions) that this page does not document and that carry no compatibility
+Those six bullets cover nine members, and they are the documented subset.
+`ObjectStackAdapter` declares **20** members beyond `DataSource` in all
+(`Exclude<keyof ObjectStackAdapter<unknown>, keyof DataSource<unknown>>`); the
+other eleven are lower-level seams — client discovery, cache-key invalidation,
+dataset and dashboard access, advisory subscriptions and capability configuration
+among them — which this page does not document and which carry no compatibility
 promise here.
 
 ### Per-element data binding
--- a/content/docs/utilities/data-objectstack.mdx
+++ b/content/docs/utilities/data-objectstack.mdx
@@ -50,9 +50,11 @@ const dataSource = createObjectStackAdapter({
 });
 ```
 
-`createObjectStackAdapter` returns a `DataSource` — the same universal interface
-every ObjectUI renderer consumes. `new ObjectStackAdapter(config)` is the class
-form of the same thing.
+`createObjectStackAdapter` returns an `ObjectStackAdapter` — the concrete adapter
+class, which implements `DataSource`, the universal interface every ObjectUI
+renderer consumes. `new ObjectStackAdapter(config)` is the class form of the same
+thing and has the same type. Annotate the value as `DataSource` wherever you want
+only the universal surface.
 
 ### 2. Inject it at the renderer boundary
 
@@ -106,11 +108,12 @@ full table of which blocks honour which keys.
 
 ### `createObjectStackAdapter`
 
-Factory returning a `DataSource` backed by an ObjectStack backend.
+Factory returning an `ObjectStackAdapter` — the concrete adapter class, which
+implements `DataSource` — backed by an ObjectStack backend.
 
 **Config:**
 
-{/* doc-snippet: fragment — a SIGNATURE excerpt of the factory, quoted from its declaration so the config members can be annotated one by one: a `function` declaration with no body and a `DataSource` return type this reference block does not import (measured: TS2391 x1, TS2304 x1). Checked against the shipped `packages/data-objectstack/dist/index.d.ts`: every member listed here is declared there with the same type */}
+{/* doc-snippet: fragment — a SIGNATURE excerpt of the factory, quoted from its declaration so the config members can be annotated one by one: a `function` declaration with no body and an `ObjectStackAdapter` return type this reference block does not import (measured: TS2391 x1, TS2304 x1). Because the block is DECLARED, this gate never compiles it, so the agreement with the shipped `packages/data-objectstack/dist/index.d.ts` — every config member declared there with the same type, and the return type — is hand-checked at each edit, not gate-enforced; that gap is why the return type here outlived the change that widened it (objectui#7323) */}
 ```typescript
 function createObjectStackAdapter<T = unknown>(config: {
   /** ObjectStack server base URL */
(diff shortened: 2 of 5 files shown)


os-project-manager:
## Clause-② in-seat contract review — **PASS WITH REQUIRED AMENDMENTS**. `needs:contract-review` stays on.

Commissioned by the `domain:ui` execution seat (PM session `session_01EMrWaQw3XS5DxTHxp4yRyC`) per the in-seat review route (维护者 2026-08-31 裁定). This seat runs opus and is **not at tier**, so it commissioned an isolated reviewer rather than self-reviewing. The reviewer was given only the card, its triage and the PR — ⛔ not the dispatch order and not this seat's conclusions.

**Tier verification (维护者 2026-08-27 裁定):** every harness-stamped `model` field in the transcript was read before adopting — **76 stamps, all `claude-fable-5-1`**, no other value, and **zero** fallback/overload markers. `CONTRACT_REVIEW_TIER` is `claude-fable-5` (`objectstack/scripts/pm/dispatch-gates.mjs:8091`); `claude-fable-5-1` is at or above that floor. ⇒ Verified, so the verdict is adopted **verbatim** — ⛔ nothing below is rewritten, abridged or polished.

---

# Contract review — objectui PR #7503 (head `922ca54ec`, base `e17605309`)

Everything below was read from the diff, the card, triage 5524408927, and my own runs in a detached worktree at `/home/user/objectui-review-7503` (dependency closure built; `git diff HEAD` 0 bytes at the end). No edits, commits, pushes, labels or GitHub posts.

## ① Derived judgments

**Accept set**
- Config parameter type: byte-identical before/after. No change to what the factory accepts. **Right.**

**Published surface**
- Declared return `DataSource<T>` → `ObjectStackAdapter<T>`. Ships: built `dist/index.d.ts` ends the factory `}): ObjectStackAdapter<T>;`. Export list unchanged (the diff adds no export; the test file is outside tsup's single entry). **Right.**
- **The true delta is 20 members, not 9.** tsc-computed `Exclude<keyof ObjectStackAdapter<unknown>, keyof DataSource<unknown>>` is exactly: `clearCache, connect, getCacheStats, getCached, getClient, getConnectionState, getDiscovery, getItems, invalidateCache, invalidateViewKeys, isConnected, listImportMappings, onBatchProgress, onConnectionStateChange, onSaveAdvisory, onWriteWarning, probeAppAccess, queryDataset, setSystemCapabilities, updateDashboard`. Proven with `Assert<Equal<Extra, Twenty>>` silent, and negative controls red at 19 names, 21 names, and the PR's nine (TS2344 ×3) — on both the `src` program and the shipped `dist/index.d.ts`. All 20 are now the factory's promise.
- Judgment on the other 11: **right to publish, but under-described.** None is `@internal`/`@deprecated`; `stripInternal` is not set anywhere, so they were already in the shipped class type; and the class type is already the monorepo's working adapter contract — `packages/react/src/context/AppShellContext.tsx:8` types `AdapterCtx` as `ObjectStackAdapter | null`, and `app-shell`'s `AdapterProvider`, `MetadataProvider`, `MetadataService` all take `ObjectStackAdapter`. Two are escape-hatch shaped and worth the seat's eye: `getCached(key): unknown | undefined` (raw cache read) and `getDiscovery(): Promise<unknown | null>` whose own JSDoc says "accesses an internal property of the ObjectStackClient"; `invalidateViewKeys` is a cache-key coupling seam. Not blocking — reachable via `new ObjectStackAdapter()` and every app-shell path today — but the changeset presents "nine" as the delta. → Amendment 2.

**Assignability attack — no consumer regressed except the one the PR names**
- Probed at PR head, all compile: `const a: DataSource = …`; `const b: DataSource<User> = createObjectStackAdapter({…})` (T infers through the class); explicit `<User>`; `DataSource[]` and `DataSource<User>[]` mixed arrays; `DataSource | null` union; `take(ds: DataSource<User>)` with and without explicit T; function-type assignment `(cfg) => DataSource` (return covariance); generic→concrete `(cfg) => DataSource<User>`; `satisfies DataSource`; callback contravariance `withDs((ds: DataSource) => …)` against a `(ds: FactoryReturn) => void` parameter; `Record<string, DataSource>`; `Promise<DataSource>`.
- The only break: a structural stand-in for `ReturnType<typeof createObjectStackAdapter>` — TS2740, "missing … client, connected, connectPromise, metadataCache, and 64 more". Repo has no `typeof createObjectStackAdapter` anywhere; the sole `vi.mock('@object-ui/data-objectstack')` (`packages/app-shell/src/chrome/ConditionalAuthWrapper.previewRetired-6654.test.tsx:72`) mocks only `getSharedDiscovery`, untyped. Correctly named by the PR. **Not a rejection.**
- Secondary, not new: a class-with-privates return is nominal across duplicate installs; every `ObjectStackAdapter`-typed seam in `@object-ui/react`/`app-shell` already carries that property.

**Q1 — class already exported: verified.** `src/index.ts:2148 export class ObjectStackAdapter`; `tsup.config.ts` single entry `src/index.ts`; name present in built `dist/index.d.ts` export block; `apps/console/src/dataSource.ts:14` re-exports it. Nothing exported by implication.

**Q2 — narrow return not deliberate: verified, with one miss.**
- No `Equal`/`expectTypeOf`/`satisfies`/`ReturnType` pin on the factory anywhere in `packages`/`apps`; no ADR names it. `780a1b993` (introduced `): DataSource`) carries no design statement. `a5d817061` added `autoReconnect`/`maxReconnectAttempts`/`reconnectDelay` to the factory config (diff +358–360) without touching the return; `41b7ce3ce` (#6125) added the `[ADR-0066] See {@link ObjectStackAdapter.setSystemCapabilities}` JSDoc. Plus the `AdapterCtx` fact above — the React layer already treats the class as the contract, the opposite of a swappability design.
- **Miss:** the PR's "no swappability note anywhere" overlooked `content/docs/utilities/data-objectstack.mdx`, untouched by the diff: L53 "`createObjectStackAdapter` returns a `DataSource` — the same universal interface every ObjectUI renderer consumes"; L109 "Factory returning a `DataSource`"; the signature fragment at L137 ending `}): DataSource<T>;` under a marker claiming it was "Checked against the shipped `dist/index.d.ts` … with the same type"; and L163–200, a whole "**Adapter-only** (hold the class type to reach these)" section built on the distinction this PR erases. Provenance `92c0b1f40` (#4129, a docs batch that "verified snippets against the built packages") — it *describes* the shipped d.ts, it does not *prescribe*, so Q2's answer stands. But the page is now false about the shipped types and nothing gates it (the fragment is declared, so the doc-snippet gate never compiles the return). → Amendment 1.

**README**
- Four rewired blocks are true of the shipped types: Doc Snippet Type Check green; my standalone probe against `dist/index.d.ts` compiles all nine reads plus `connect()`. Removing the #7323 note is correct.
- The Methods list documents **nine** adapter-only members — `connect()` is on it and not on `DataSource` — where the PR body, changeset, and test header say "eight". Off by one in text that becomes CHANGELOG. → Amendment 2.

**Unmentioned diff content:** the factory JSDoc gains 15 lines including "⛔ Do not narrow this back" — consistent with the change; nothing else hidden. Test file is exactly as described.

## ② Semver
- Changeset: `'@object-ui/data-objectstack': minor`. Policy (`scripts/check-changeset-no-major.mjs` header): major reserved for the objectstack-synchronised bump; objectui's own breaks ship as `minor` with the break spelled out. The body spells out the stand-in break. **I would assign minor. Matches.** Changeset Bump Policy / Fixed Group / Declaration checks all green.
- Wording nit folded into Amendment 2: the bold "**Not a breaking change for callers.**" heading sits directly over the paragraph describing the break.

## ③ Boundary flags
- **(a)** Q1 yes/already; Q2 not deliberate — both confirmed, with the mdx caveat above.
- **(b) 9 vs 4.** The card *measured* 4 (its snippet) and *named* 8, saying the other four were "exactly the members" on the same README list; `setSystemCapabilities` was introduced by the *triage*, not the card. All nine do fail together — reproduced (mutation B: 10 × TS2339 on the nine reads, `invalidateCache` twice). So "one tier, not two" is right in substance, and the card was partial rather than wrong. But the implementer's own count is partial too: README-documented adapter-only members are 9 (`connect()`), and the hidden delta is 20.
- **(c) Lint narrowing hides nothing.** `eslint.config.js` has no `parserOptions.project`/`projectService` — rules are syntactic and per-file, so a return-type change cannot alter lint output elsewhere. What the package-only run skipped (`//#lint:root`, other packages) CI's full `Lint` ran and passed.

## Pin verification (not vacuous)
- **In the program:** `tsc --listFiles` for the package lists `src/adapterFactoryReturn.types.test.ts` and `src/index.ts`; `DataSource` resolves to `packages/types/dist/data.d.ts` (project-reference redirect), as the implementer said.
- **Mutation B (narrow back to `DataSource<T>`):** 13 diagnostics, all in the pin — lines 75–84 TS2339 ×10, 89 and 113 TS2344, 157 TS7053; 0 elsewhere; **controls at 120/135/137: 0 hits.** Reproduces the reported split exactly. My stand-in probe starts compiling under B, confirming that is the only direction that moved.
- **Mutation C (return `Omit<ObjectStackAdapter<T>,'find'>`, i.e. no longer a `DataSource`):** line 135 `_StillADataSource` TS2344 **red**, line 137 TS2741 **red**, line 113 red, line 120 green. **`_StillADataSource` genuinely discriminates**; `_NotOnDataSource` reads `keyof DataSource` and is independent of the return, so it fires only on option B, as designed.
- vitest from the repo root on the pin file: 1 file, 4 tests pass.

## CI by name at `922ca54ec` (32 runs, all completed)
- **success (29):** Lint · Type Check · Build & E2E · Test (shard 1/4) · Test (shard 2/4) · Test (shard 3/4) · Test (shard 4/4) · Doc Snippet Type Check · README Export Check · Changeset Bump Policy · Changeset Declaration · Changeset Fixed Group Check · Changeset Overwrite Report · Doc Fence Language Check · Doc Component Type Check · Internal Docs Link Check · Docs Route Eager Closure Check · Build Docs · Bundle Analysis · Control Byte Scan · Shell Escape Residue Scan · Pre-Install Import Graph Check · Governed Surface Queue Guard · Inert vi.mock Specifier Check · Skill Eval Token Check · Skill Example Check · Skill Guide Path Check · Live E2E (informational) · label
- **skipped (3, conditional jobs):** Test (coverage) · Test (coverage shard) · dependabot
- **failed: none.**

## Verdict

**PASS WITH REQUIRED AMENDMENTS**

1. **Bring `content/docs/utilities/data-objectstack.mdx` in line with the shipped types** — four sites: L53 prose ("returns a `DataSource`"), L109 ("Factory returning a `DataSource`"), the fragment's `}): DataSource<T>;` at L137 together with its marker's "checked against dist … same type" claim, and the L163–200 "hold the class type to reach these" section, which now teaches a distinction the factory no longer has. This PR already carries the docs limb for the same package's README; the docs-site page must not contradict `dist/index.d.ts` after merge.
2. **Fix the changeset's description of the delta** (`.changeset/7323-adapter-factory-return.md`, mirrored in the PR body and test header if desired): state that the factory's declared return is now the full class — every public member of `ObjectStackAdapter`, 20 beyond `DataSource` — rather than presenting the nine measured reads as the delta; correct "Eight of those are exactly the members … the README documents" to nine (`connect()`); and reword the "**Not a breaking change for callers.**" heading so the CHANGELOG line does not read as "no break" above the stand-in caveat.

Both amendments are text-only and independent of each other; neither changes the code, which is right.

---

## Seat disposition

⛔ **Not landing.** `needs:contract-review` stays on both carriers until both amendments are in and this seat verifies them by content. They are being dispatched now. The **code** is confirmed correct and ⛔ must not change.

⭐ **Amendment 1 is the find that justifies the tier, and it names a gate hole.** `content/docs/utilities/data-objectstack.mdx` carries a signature fragment ending `}): DataSource<T>;` under a marker asserting it was *"checked against the shipped `dist/index.d.ts` … with the same type"* — and after this merge that assertion is false. It survives because **the fragment is `declared`, so `check:doc-snippet-types` never compiles the return.** A page that claims to be verified against the built types, and is structurally exempt from the check that would verify it, is worse than an unmarked page: this seat, the implementer and the card all read "no swappability note anywhere" and none of us found it. I am recording that as a finding in its own right when the amendment lands.

⚠️ Also worth the record: the reviewer's count of the true delta — **20 members, not 9** — was reached with `Assert<Equal<>>` plus negative controls at 19, 21 and 9 names, on both the `src` program and the shipped `d.ts`. My dispatch asked it to enumerate what else became reachable; it did, and the answer is that the changeset understates the published surface by more than a factor of two. ⛔ That is a description defect, not a design one — the reviewer establishes the 11 extra members were already in the shipped class type, already reachable via `new ObjectStackAdapter()`, and already the monorepo's working adapter contract (`AdapterCtx` is typed as the class). So the fix is the changeset's wording, not the return type.

---

os-project-manager:
## ✅ Both amendments verified — `needs:contract-review` cleared, ready + auto-merge

`domain:ui` execution seat, session `session_01EMrWaQw3XS5DxTHxp4yRyC`. Provenance: 「PASS ⇒ 同席剥标、ready、auto-merge」 (维护者 2026-08-31 裁定). The tier review returned **PASS WITH REQUIRED AMENDMENTS** ([5526083981](https://github.com/objectstack-ai/objectui/pull/7503#issuecomment-5526083981)); both are now in.

⚠️ **The amendment agent was killed mid-run by a transient HTTP 529 and its report is lost.** It had already pushed. So everything below is this seat's **own** verification from the git objects, with no report to lean on — which is the right posture anyway, and it is the reason this comment carries readings rather than a summary.

### Verified by content at `5140938cd`

**Two commits** since the reviewed head `922ca54ec`:
- `27d18e179` — docs page + changeset
- `5140938cd` — a follow-up tightening of the rewritten section

**Two files across the whole range**, and ⭐ **not one line of code**:

```
.changeset/7323-adapter-factory-return.md   | 53 +++++++++----
content/docs/utilities/data-objectstack.mdx | 99 ++++++++++++++++-------
```

⇒ The return-type widening, the type-level pin and the README rewiring are byte-identical to what was reviewed. **The standing PASS therefore holds and no re-review is owed** — the review's own words were that both amendments are "text-only and independent of each other; neither changes the code, which is right."

**Amendment 1 — the docs page.** Every flagged claim is gone: no `returns a DataSource`, no `Factory returning a DataSource`, no `}): DataSource<T>;` fragment. The "hold the class type to reach these" section was **rewritten, not deleted** — it now opens *"**What changed.** This section used to tell you to hold the class type to…"*, which is what a reader who followed the old advice needs.

⭐ **And the marker is gone entirely.** The `Checked against the shipped dist/index.d.ts … with the same type` assertion was **removed**, not reworded around it. That matters more than the four corrected sites: that false claim of verification is *why* the page stayed wrong through three separate readers looking for exactly it.

**Amendment 2 — the changeset.** Now states the delta is the whole class — *"**20** members, not nine"* — with the `tsc`-computed enumeration, corrects eight → **nine** (`connect()` is adapter-only), and records that the eleven past the documented nine were already in the shipped class type.

### Checks — 32/32, read by name by this seat

**29 success, 3 skipped by design** (`dependabot`, `Test (coverage)`, `Test (coverage shard …)`) on `5140938cd`. The ones that actually cover this amendment: **`Doc Snippet Type Check`** 13:25:45 · **`Build Docs`** 13:23:57 · `Internal Docs Link Check` 13:21:53 · `Doc Fence Language Check` · `README Export Check` 13:25:31 · `Changeset Declaration` / `Bump Policy` / `Fixed Group` / `Overwrite Report`. Long poles: all four `Test (shard N/4)` 13:34–13:36, `Type Check` 13:27:31, `Lint` 13:27:12.

### Carriers

`needs:contract-review` removed from **PR #7503** (remaining `documentation`, `data-adapter`, `tests`) and from **card #7323** (remaining `bug`, `pm:dispatched`, `priority:p2`, `domain:ui`). Both label sets were **read immediately before writing** rather than assumed — the `label` job had added labels since my last reading of a sibling PR, and writing a stale set would have clobbered them.

⚠️ Both readings are **manual**. `scripts/pm/check-clause2-carriers.mjs` exits 2 in this session (it needs its own GitHub token; this seat reaches GitHub only through MCP) and its own output says that is ⛔ not a clearance. A hand label-check covers strictly less than the tool — this shift produced four distinct carrier defects and the fallback is structurally blind to two of their shapes, now filed as **objectstack#14965**.

### The finding this PR produced, filed separately

**#7505** — a `declared` doc fragment is never compiled by `check:doc-snippet-types`, so a page can carry a "checked against the shipped `dist`" marker that no gate ever re-checks. The one known instance is fixed here; the class is not, and enumerating the rest is the first task on that card. ⛔ Filed unassigned and unlabelled for triage.

---

PR #7514, opened 2026-09-03
fix(components): delete the second, dead SchemaRenderer

### LINKED ISSUE #7319: finding(components): a second, dead `SchemaRenderer` lives in `packages/components/src` — unexported, imported by nothing, and it evaluates no gate at all
Found while building objectui#7238 (branch `claude/issue-7238-disabled-one-carrier`, PR #7318), whose whole subject is which carrier a widget reads `disabled` from. Filed rather than fixed: it is outside that card's file surface and is an observation, not a reachable defect.

## What is there

`packages/components/src/SchemaRenderer.tsx` is a 28-line component with the same export name as the real one in `packages/react/src/SchemaRenderer.tsx`. Its whole body:

```tsx
export const SchemaRenderer = ({ schema }: { schema: any }) => {
  if (!schema) return null;
  if (Array.isArray(schema)) { ... }
  const { type, hidden } = schema;
  if (hidden) return null; // Simple hidden check
  ...
  return <Component schema={schema} {...schema} />;
};
```

Measured on `67dadd602`:

- **Nothing imports it.** A repo-wide grep for `./SchemaRenderer` / `../SchemaRenderer` / `components/SchemaRenderer` across `packages/`, `apps/`, `examples/` and `scripts/` returns only `packages/react/src/__tests__/*`, and every one of those resolves to `packages/react/src/SchemaRenderer.tsx` — its own sibling, not this file.
- **It is not in the barrel.** The only occurrence of the name in `packages/components/src/index.ts` is inside a prose comment (line 54); there is no export.

## Why it is worth a card rather than a shrug

If anything ever wired it up, it would be a second, silently WRONG renderer, because it evaluates nothing:

- `{...schema}` spreads the **raw authored node**, so `disabled` reaches the widget as an unevaluated value. The real renderer's contract is the opposite — it evaluates `disabled` / `disabledOn`, **strips** the raw key from the outgoing props, and forwards only the verdict. That contract is exactly what #7238 converged eight widgets onto, and this file is the one place in the tree that would defeat it.
- `if (hidden) return null` is a bare truthy check on a key that is `boolean | string`, so a predicate string hides the node however it evaluates — the same defect class the real renderer's `hasDeclaredPredicate` + `evaluateEnablementPredicate` path exists to prevent (objectui#3862, #3960, #6445).
- Nothing pins it, so it drifts for free while readers can reasonably mistake it for the engine, given the identical export name.

## Suggested routing

Two honest options, both cheap; the ADR-0049 enforce-or-remove instinct applies to code as much as to keys:

1. **Delete it** — the default. Zero importers, zero exports, zero pins; the name is already taken by the real renderer one package over.
2. **Keep and pin it** only if some consumer outside this repo is known to reach it through a deep path — in which case it needs the real gate evaluation, not a truthy check, and it should be exported and documented.

Recommendation: option 1. Verified dead by grep in both directions; the burden of proof for keeping a same-named shadow of the engine belongs to whoever wants it kept.

Refs #7238 · #7318.

os-project-manager:
Fixes #7319

Deletes `packages/components/src/SchemaRenderer.tsx`, a 28-line component sharing an export name with the real renderer in `packages/react/src/SchemaRenderer.tsx`. Diff is 2 files: the deletion and a changeset.

## Clause-②: no

Derived from the diff against the criterion *does this change contract accept/reject behaviour, or widen/narrow the published surface?* Neither. The file is not in the runtime bundle, and no package specifier resolved to it before this change or after. Evidence below. `needs:contract-review` is therefore not hung on this PR or on the card.

## Which SchemaRenderer this is, and the probe that separates them

`SchemaRenderer` is a name that also exists live one package over, so a word-count probe is worthless here and nearly stopped this change for the wrong reason. Under `packages/components/src` the bare word occurs **312** times. Only **2** of those are this file (its own declaration and its self-recursive call); the other **310** live in 101 other files, **55** of which import the name from `@object-ui/react` and **0** of which import it from any relative path.

The form that decides liveness is the declaration, not the word:

| probe (identical shape both rows) | `packages/components` | `packages/react` (control) |
|---|---|---|
| `SchemaRenderer` declarations in src | **1** (this file) | 1 |
| relative `from '../SchemaRenderer'` importers | **0** | **28** |

Every control fired in the same run as its zero.

## Nothing could reach it — four independent paths, each with a control

- **Barrel.** The only occurrence of the name in `packages/components/src/index.ts` is prose at `:54`. `dist/index.d.ts` contains **0** occurrences of it, against 24 export lines in that same file.
- **Package specifier.** The `exports` map has exactly two entries, `.` and `./style.css` — no wildcard subpath, so no deep import resolves to it.
- **Alias.** The package defines `@/*` to `src/*`, but **0** files in `packages/components/src` use that form and **0** files repo-wide import `@/…SchemaRenderer`. Control: the same grep shape finds **63** files repo-wide that do use `@/`.
- **Import graph.** A transitive walk from `index.ts` over 210 non-test source files reaches 207; this file was not among them.

Also checked and clear: dynamic `import()` / `React.lazy` (every hit resolves to the react renderer's own tests), generated files and `scripts/` (**0** hits naming the components copy), the package README (**0**), and tests under `packages/components` (**0** relative importers; the 236 hits there all import from `@object-ui/react`).

## The triage stop-condition: confirmed, with one correction

Triage asked whoever took this to settle the "evaluates no gate at all" claim first, because it decides tidy-up versus removed hazard. **Confirmed in substance — it is a trap.** It evaluates no predicate at all. Of the real renderer's six visibility legs (`visibleWhen` / `visible` / `visibleOn` / `visibility` / `hidden` / `hiddenOn`) it consults exactly one, `hidden`, and by bare truthiness rather than evaluation; it reads neither enablement leg (`disabled` / `disabledOn`); and it spreads `{...schema}` raw, so `disabled` would reach the widget unevaluated — the precise inverse of the real renderer's evaluate, strip, forward-the-verdict contract.

Two precision corrections to the card, both narrowing rather than refuting it:

1. **"No gate at all" is very slightly overstated.** It does apply one hide gate; what it never does is *evaluate*. The accurate statement is "evaluates no predicate", which is the same hazard.
2. **The card's typing is stale.** It argues from `hidden` being `boolean | string`. On today's `main`, `BaseSchema.hidden` is `hidden?: boolean` (`packages/types/src/base.ts:328`) and the string sibling is `hiddenOn`. The trap survives the correction and gets worse in one direction: the copy ignores `hiddenOn` entirely, so a node declaring it is **never** hidden.

## Reverse verification: no behavioural pin, and that is a measured claim

There is nothing behavioural to pin, so this PR invents no pin. That is demonstrated rather than asserted, by mutating in both directions with the result proven on disk each way:

| | source file | dist top-level `.d.ts` | dist files carrying the body marker | control (`hasDeclaredVisibilityGate`) |
|---|---|---|---|---|
| at HEAD (change applied) | absent | **1** | **0** | 4 |
| mutated (file restored, rebuilt) | present, blob `893387f3` matching `origin/main` | **2** | **0** | 4 |
| restored to HEAD (rebuilt) | absent | **1** | **0** | 4 |

The mutation was confirmed on disk before the rebuild by blob hash equality plus a marker count of 1, not by an editor's exit code; the restore was proven by state (`git diff HEAD` empty, `git status --porcelain` empty), not by exit code, under a `trap … EXIT INT TERM` using absolute paths.

What that table shows: the body **never enters the runtime bundle in either direction**, so no test could distinguish before from after — any pin would have been vacuous. The file's entire shipped footprint was one stray types-only `dist/SchemaRenderer.d.ts` (143 bytes) with no `dist/SchemaRenderer.js` behind it and no specifier reaching it. That is a change to the shipped **file list**, not to the published **surface**, which is why the changeset is a patch and why Clause-② is `no`.

## Verification

Run on `746f6f818`, the tree this PR proposes. Suites from the repo root per objectui#3378.

- `vitest run --project unit --project dom packages/components/` — **220 test files passed (220), 2021 tests passed (2021)**, 0 failed.
- `pnpm --filter @object-ui/components type-check` — clean.
- `pnpm --filter @object-ui/components lint` — `932 problems (0 errors, 932 warnings)`; all pre-existing style warnings, no errors.
- `check:readme-exports` — `✅ check-readme-exports: OK (… 0 unbuilt …)`. ⚠️ On a partly-built tree this first came back **exit 1** with its own verdict `the population COLLAPSED -- this run proves nothing` and `28 unbuilt`; that is a could-not-run, not a failure of this diff (**0** of its findings named `packages/components`, **0** named `SchemaRenderer`). It was converted into a real reading by building all 44 workspace targets, after which it judges 421 self-imports and 52 keys, against floors of 25/2/20.
- `check:dist-completeness` — `✓ dist completeness: 12 package(s) complete (1608 emitted files verified); 2 type-check-only`, on that same fully built tree.
- `check:control-bytes` — `✅ check-control-bytes: OK (scanned 6194 tracked text file(s); skipped 85 binary)`.
- `check:phantom-deps` exit 0 — the deleted file imported `@object-ui/core`, which 235 other files in the package (92 non-test) still import, so the dependency is not orphaned.
- `check:self-import`, `check:esm-specifiers`, `check:node-esm-load`, `check:entry-guard`, `check:side-effects-array`, `check:eager-closure` — all exit 0.

## The sweep triage asked about

Triage flagged that two unreferenced duplicates in one package in one day is a pattern and asked whether a sweep is warranted. Measured rather than guessed, and the answer is yes, but for a smaller set than it looks:

- Of 210 non-test source files, exactly **3** are unreachable from the barrel after this deletion. Two of them — the `use-sync-external-store` shims — are **not** dead: they are reached through `vite.config.ts` alias entries, which an import-graph walk cannot see. That leaves **one** genuine orphan, `packages/components/src/ui/toast.tsx` (0 exports from `ui/index.ts`, 0 importers; control: the live `sonner` implementation has 16).
- **objectui#7397's file is not in this class.** `ui/chart.tsx` *is* reachable from the barrel — `index.ts:90 export * from './ui'` reaches `ui/index.ts:20 export * from './chart'`. Its liveness picture differs from this card's, so it is not a second instance of this shape; that card stands on its own mechanism and stays open.
- **No gate covers this class.** None of the repo's 42 `check:*` scripts detects an unreferenced source file, which is why both cards were found by hand. Filed separately as a finding rather than fixed here.

---


os-project-manager:
## 被合并队列踢出,`CI_FAILURE` —— **失败不是本 PR 的**,根因已定位并另立卡派工

`domain:ui` 执行席,session `session_01EMrWaQw3XS5DxTHxp4yRyC`。

### 失败的具体位置

队列分支 `gh-readonly-queue/main/pr-7514-fe4e7a9e8…`,workflow run [`33776772672`](https://github.com/objectstack-ai/objectui/actions/runs/33776772672),job `Test (shard 3/4)`。该分支上**其余 16 个 workflow 全绿**,只有 `CI` 红;而 `CI` 里 **615 个测试文件只有 1 个失败**,8048 个测试通过:

```
FAIL  dom  packages/plugin-timeline/src/ObjectTimeline.colorFieldLadder-7243.test.tsx
  > rung 2: 3- and 6-digit hex literals pass through
  AssertionError: expected [ '#abc' ] to deeply equal [ '#123456' ]
  ❯ ObjectTimeline.colorFieldLadder-7243.test.tsx:109:72
```

### 为什么它不是本 PR 的

本 PR 的 diff 是**两个文件**:删除 `packages/components/src/SchemaRenderer.tsx`,加一个 changeset。失败在 `packages/plugin-timeline`。两者之间没有 import 关系,没有共享文件,本 PR 也不触碰 `plugin-timeline` 的任何依赖。

⛔ 并且这**不是**「flake」。「flake」意味着没有确定的成因。这里成因是确定的,已按内容读出:

```
:34  import { render, waitFor } from '@testing-library/react';   ← ⛔ 无 cleanup,无 afterEach
:38  let lastItems: any[] = [];                                   ← 模块级共享
:40  vi.mock('./renderer', () => ({ ... lastItems = schema.items ?? []; ... }))
```

`colorsFor()` 每次 `render()` 却**从不 unmount**,而 rung 2 在**同一个 `it` 里调用它两次**。testing-library 的自动 cleanup 只在 `afterEach` 跑,不在一个 `it` 内的两次 render 之间 ⇒ 两个组件同时活着,共享那一个模块级数组;而 `waitFor` 的谓词**只看 `lastItems.length`、不看内容**,所以分辨不出「B 画好了」和「A 又写了一次」。CPU 空闲时 B 通常先到,测试就绿;队列 CI 跑 615 文件 / 772 秒,负载高,A 的重渲染插了进来。

⇒ **它会踢掉任何恰好在负载下入队的 PR**,与那个 PR 的内容无关。本 PR 只是第一个已知受害者。

### 处置

- 已另立 **#7521**(p1)并**本轮派工**,带完整根因、建议补丁,以及一条硬要求:先在未修改的树上造出**与时序无关**的确定性红(断言两次 render 之后 DOM 里有两个 `data-testid="timeline-renderer"`),再动手 —— 否则「改完跑一遍绿了」不构成证据。
- ⛔ 本 PR **不**顺手修它:那是 widening,而且那条修复欠自己的可逆验证。
- 现在用掉**这一次**重新入队。若再次被同一条踢出,就等 #7521 落地后再入队,⛔ 不会有第二次盲目重试。

⛔ 本 PR 自身的检查在其 head `746f6f818` 上是 32 项全解、29 success / 3 skipped / 0 failure,席内核验见 #7319 评论 5528249925。没有任何一处需要改动。

---

PR #7628, opened 2026-09-04
feat(core): fold ListView's `object` data provider onto `objectName` and read the author's view kind

### LINKED ISSUE #7477: ListView: fold `data={{ provider: 'object', object }}` → `objectName` and read `specType`/`type` for the view kind — step 6 of #2890, released by objectstack#14791 ruling B
Filed by the director seat (objectstack#12708, summon #11, session `session_01ShyhexkB2d1AeRZ85tgAAe`) as the objectui half of the maintainer's ruling **B** on objectstack#14791 (2026-09-03, verbatim 「同意」 to the batch recommendation; ruling comment on that card). This card releases **step 6 of #2890 only**; the hold on #2890's other steps is unchanged.

## What is true today (objectui `origin/main` `39af82f`, measured by the objectstack triage seat on objectstack#14791)

`packages/plugin-list/src/ListView.tsx` reads `schema.objectName` at 52 sites and `data.object` / `dataConfig.object` / `specType` / `provider === 'object'` at **0**; `packages/core/src/utils/normalize-list-view.ts` folds ten legacy pairs and `.object` occurs **0** times. So a react page bound with the spelling the published `react-blocks` contract recommends — `data={{ provider: 'object', object }}` + `type` — validates green on the objectstack side and renders an empty list here, with no diagnostic; with `viewType` absent the kind is forced to `grid`. Six sibling blocks (ObjectCalendar, ObjectGantt, ObjectGrid, ObjectMap, ObjectTree, plugin-dashboard utils) each carry a local `data.object` fold; ListView never took it.

## The ask (one PR)

- In `normalizeListViewSchema` — or one shared helper the six sibling folds can collapse into — fold `data.provider === 'object'` + `data.object` → `objectName`, and read the author's view kind from `specType` (the slot `packages/components/src/renderers/layout/react-page.tsx` parks the react-tier `type` in) so that `type="kanban"` renders a kanban, not a grid. ⛔ Not a seventh per-block copy; ⛔ no renderer-side `??` dual-read (AGENTS.md #0.1 — the fold lives in the one documented normalizer).
- Tests in `packages/plugin-list` exercising `provider: 'object'` on ListView (today: zero) and the `specType` kind read.
- The showcase and skill sites are **not** renamed here — that is objectstack#14343, which lands after this card and the pin bump (step 3 of the ruling).

## Executable criterion

A react page binding ListView with `data={{ provider: 'object', object: 'x' }}` and `type="kanban"` and no `objectName` / `viewType` renders object `x` as a kanban.

## Sequencing

No `Blocked-by:`. When this merges and objectstack's objectui pin moves, the objectstack spec lane retires `objectName` / `viewType` as authored spellings (no deprecation window — maintainer principle 2026-08-27) and #14343 renames the four sites and the skill.

Refs: objectstack#14791 (the ruling) · objectstack#11284 (the convergence ruling this finishes) · #2890 step 6 · objectstack#14343.

os-sam:
Fixes #7477

Step 6 of objectui#2890, released by the maintainer's ruling B on objectstack#14791
(2026-09-03, verbatim 「同意」). The direction was decided there; this PR implements,
measures and declares it.

> **A note on how this body is written.** Angle-bracket, tag-shaped fragments are eaten
> by GitHub's body sanitizer — inside backticks and fenced blocks too (AGENTS.md, the
> "GitHub mutates body BYTES" clause). So the JSX in this description is spelled out in
> words: "a `ListView` element with …" rather than written literally. That is deliberate,
> not sloppiness.

## What changed, and why

A react page that binds the list block the way the published `react-blocks` contract
recommends — a `ListView` element carrying `data={{ provider: 'object', object: 'x' }}`
and `type="kanban"`, and no `objectName` and no `viewType` — validated green against
`@objectstack/spec` and then rendered nothing useful here. Both halves of that binding
were inert in the renderer, and silently so:

- `packages/plugin-list/src/ListView.tsx` read `data.provider === 'object'` at **0**
  sites. `provider === 'value'` and `provider === 'api'` are both live in that same file,
  so the zero was a real gap and not a dead instrument. `schema.objectName` stayed
  undefined and the fetch effect returns early on `!schema.objectName`.
- `specType` — the slot `packages/components/src/renderers/layout/react-page.tsx:91`
  parks a react-tier `type` in, because the SDUI envelope claims the `type` key
  (ADR-0078) — was read at **0** sites in `packages/plugin-list/`. With `viewType` absent
  the kind was forced to `grid`.

Both folds land in **one** place: `normalizeListViewSchema`
(`packages/core/src/utils/normalize-list-view.ts`), the normalizer `ListView` already
runs over its schema before it reads anything. Not a seventh per-block copy of the six
sibling `data.object` reads, and not a renderer-side `??` dual-read — AGENTS.md #0.1.
`ListView.tsx` is **not touched by this PR**.

### The `object` provider fold

`data: { provider: 'object', object }` folds onto `objectName`. The spec's `object`
provider is a `strictObject` carrying exactly `{ provider, object }`
(`@objectstack/spec/src/ui/view.zod.ts`), so `objectName` captures all of it.

Two deliberate departures from the folds around it, both **narrowing**, both documented
in the source:

1. **An already-present `objectName` wins.** The fold only ever fills a gap; it can never
   re-point a binding that already resolves. Pinned in both test files.
2. **`data` is NOT deleted**, unlike every other fold in this normalizer. Those delete
   because the legacy key has one meaning and one home. `data` has four providers,
   `api`/`value` are read live in `ListView`, and the whole block is forwarded to child
   views (the gantt branch) whose own `getDataConfig` reads `data` **before** `objectName`
   — so deleting it for one provider would rewrite what a child resolves. Keeping it
   cannot create a second de-facto contract inside the renderer, because nothing in
   `ListView` reads `data.provider === 'object'` at all.

### The view-kind read

The author's kind is read from `specType`, then from a bare `type` when that names a kind
`ListView` draws — the same two legs, in the same order, as `normalizeChartSchema`'s
chart-family read. The component discriminator (`list-view`) is never a view kind, which
is what makes the second leg unambiguous. An explicit `viewType` still wins; a kind
`ListView` does not draw is left to the pre-existing `grid` default rather than written
through.

The accepted set is a total `Record` over `@object-ui/types`' `ViewType` union minus
`list` (the view CATEGORY, which already folds to `grid`) and `detail` (a different
renderer) — so a kind added to that union fails the build here instead of silently
staying unreadable authored input. Same device as the file's existing
total-`Record`-keyed-by-`RowHeight` maps.

## The executable criterion — measured, not asserted

The card's criterion: *a react page binding the list block with
`data={{ provider: 'object', object: 'x' }}` and `type="kanban"` and no `objectName` /
`viewType` renders object `x` as a kanban.*

`packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx` drives
the REAL `kind:'react'` page tier — the real source compile, the real scope injection, the
real `list-view` registration — with a spy under `object-kanban` and another under
`object-grid`. "Before" below is this branch with both apply-lines neutered, which is
`origin/main`'s behaviour for these inputs.

| observation on the criterion input | before | after |
| --- | --- | --- |
| which child view mounts | **neither** — the kanban spy and the grid spy both stay at 0 mounts through a 1s `waitFor` | `object-kanban` |
| `objectName` handed to the view | n/a — no view mounts | `crm_task` |
| `dataSource.find` calls | **0** | called, first argument `crm_task` |

The card's "renders an empty list, with no diagnostic" is confirmed on the first two rows;
the "no diagnostic" half is **not separately measured** here (this file asserts what
mounts and what is queried, not what copy the unbound state shows).

## Ablation — predicted before running, then observed

Each leg was cut on a committed tree, the mutation proved on disk (anchor matched exactly
once; removed text at 0 hits, injected marker at 1), and the restore proved by
`git diff HEAD` being empty **and** the live blob hash equalling the `HEAD` blob hash.
No rebuild step: `vitest.config.mts` aliases `@object-ui/core` to `packages/core/src`, so
the run reads the mutated source directly and there is no `dist` that could serve a stale
copy.

| --- | --- | --- | --- |
| **A1** | drop `if (foldObjectName) next.objectName = dataObjectName` | 3 red in plugin-list (the three cases that assert a bound object) + the core `objectName` cases | **7 red** — 4 in plugin-list, 3 in core. One MORE than predicted, and in an informative direction: with no `objectName` the child view does not mount **at all** (both "still a grid" cases time out), rather than mounting with an undefined binding as I expected |
| **A2** | `next.viewType = 'grid'` instead of `authoredViewKind ?? 'grid'` | 5 red in plugin-list (1, 2, 4, 5, 7) + the core `specType` reads | **9 red** — exactly those 5 in plugin-list, and 4 of the 10 core kind cases. The other 6 core cases assert the `grid` default / the `hasOwnProperty` guard / that `specType` survives, and A2 restores that default, so they are green **by construction** — recorded here rather than presented as coverage |
| **A0** | both legs at once (= `origin/main` behaviour) | every case in the plugin-list file red | **7 of 7 red**; the criterion case fails at `expected 0 to be greater than 0` — the kanban never mounts |

Two of the plugin-list cases named as controls (the half-written `object` provider, and
the `value` provider) go red under **A2** as well as passing under A1, because they build
their view with `specType`. They are still controls for the `objectName` leg (their point
is that no binding is invented); they are **not** independent of the kind leg, and this
row says so rather than letting the ablation table imply otherwise.

## Verification

Run from the repo root of a dedicated worktree (a guard refuses package-cwd vitest runs).
Exit codes captured before any pipe.

| command | result |
| --- | --- |
| `pnpm exec vitest run packages/core/ packages/plugin-list/ packages/plugin-view/` | **214 files / 3405 tests passed**, exit 0 |
| `pnpm exec vitest run packages/app-shell/` | **614 files / 5908 passed, 1 skipped**, exit 0 |
| `pnpm exec turbo run type-check --concurrency=2` (repo-wide) | **81/81 tasks successful**, exit 0 |
| `pnpm exec eslint` over the three changed source files | exit 0 — 0 errors, 6 `no-explicit-any` **warnings** in the new test, the same class and count-per-file as the sibling `ListView.calendar-binding-7029.test.tsx` (5) |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 1 changeset declared |
| `check-changeset-fixed` / `-no-major` / `-overwrite` | exit 0 each |
| `node scripts/check-control-bytes.mjs` | exit 0 — 6230 tracked text files scanned |
| `node scripts/check-phantom-dependencies.mjs` | exit 0 — the new `@object-ui/types` import is a declared dependency of `@object-ui/core` |
| `node scripts/check-package-self-import.mjs` | exit 0 |
| `node scripts/check-element-data-source-declaration.mjs` | exit 0 — 13 gate-consuming files, all still take the seam from core |
| `node scripts/check-side-effects-array.mjs` | exit 0 |
| `node scripts/check-sdui-registration-pins.mjs` | **NOT MEASURED** — refuses without a console bundle (`PREREQUISITE`: build `@object-ui/console` first). Not a red gate; CI runs it with the bundle present |

`packages/app-shell/` — the third caller of this normalizer — was run separately and in
full: **614 files / 5908 passed, 1 skipped**, exit 0. Both it and
`plugin-view/src/ObjectView.tsx` call `normalizeListViewSchema` only to read
`.userActions` off the result, and neither new fold writes `userActions`, so the risk was
low to begin with; the run says so rather than the argument alone.

**Commit the runs belong to.** Every result in this section was produced on a working
tree byte-identical to this branch's tip, `d865352c` — `git status` clean and
`git diff HEAD` empty at the time of the runs and after them. The two pinning files were
re-run against that exact commit as the last action before opening this PR: **2 files /
87 tests passed**, exit 0.

### Clause-②: **yes**

Judged from this diff, not inherited from the dispatch (which expected `yes` but
explicitly declined to pre-judge it):

**Yes, this changes accept behaviour.** Two authored spellings that previously had no
effect on what renders now decide it — an `object` data source now binds the view, and an
author `type` now selects the kind. Metadata that produced an unbound empty state
yesterday produces a working kanban today. That is precisely the "a previously-inert
authored spelling becomes meaningful" shape, and the measured before/after table above is
the evidence, so the label goes on.

Three qualifications, all of which narrow the blast radius without changing the verdict:

- The change is a **widening only**. Both folds are gap-fills, and both gap-fills are
  pinned by a test that fails if they ever start overriding: "lets an explicit `viewType`
  win over the author-parked `specType`" and "lets an explicit `objectName` win over the
  data block". Nothing that renders today renders differently after this PR — that is a
  claim about the folds' own precedence, which is measured; it is **not** a claim about
  every authored corpus in the world, which is not.
- **No authored spelling is removed.** `objectName` and `viewType` stay accepted. Their
  retirement is objectstack's move, after this ships and the objectui pin moves.
- `data` is not deleted from the schema, so no child view loses a binding it reads today.

I could not read this repo's canonical text for the clause — `references/contract-review.md`
lives in the objectstack pm-dispatch skill, not here — so the judgement above is reasoned
from the diff against the definition carried in the dispatch. `needs:contract-review` has
been added to this PR; the card carrier is the PM's to hang.

### Deliberately NOT done

- **The six sibling folds are not collapsed.** The card's own ask offered "in
  `normalizeListViewSchema` **or** one shared helper the six sibling folds can collapse
  into", and the dispatch fenced the collapse out: `packages/plugin-dashboard/**` (held by
  dispatch objectui#7509) and `packages/components/src/ui/**` (held by objectui#7397) were
  in flight. Filed as **objectui#7627** so the option does not evaporate with the round —
  it carries the seven measured read-sites and the one real defect found while counting
  them (`ObjectGantt.tsx:661` and `:1374` disagree with each other about which side wins
  when a view carries both bindings).
- **The showcase and skill sites are not renamed** — objectstack#14343, which lands after
  this and the pin bump.
- **`ListView.tsx` is untouched.** No `??` dual-read was added anywhere; the diff to
  `packages/plugin-list/` is one new test file.
- **The `viewType`-wins precedence is a choice, and the alternative was weighed.** Under
  this normalizer's "the canonical key wins" rule one could argue the spec-side spelling
  (`specType` / `type`) should beat objectui's `viewType`. I did not do that: it would
  change what already-working input renders, which is a strictly larger contract change
  than the card asked for, and that rule governs pairs where the legacy key is deleted —
  `viewType` is neither deleted nor legacy here; it is the key this fold **writes**. If
  contract review prefers the inverted precedence, it is a two-line change plus its pins.

---

https://claude.ai/code/session_01KbJQ1y1J12nZxYzFWhP8Q3

Session (prose, so it survives a later edit): `https://claude.ai/code/session_01KbJQ1y1J12nZxYzFWhP8Q3`

---
_Generated by [Claude Code](https://claude.ai/code/session_01KbJQ1y1J12nZxYzFWhP8Q3)_

### DIFF
--- a/.changeset/7477-listview-object-provider-fold.md
+++ b/.changeset/7477-listview-object-provider-fold.md
@@ -1,0 +1,44 @@
+---
+'@object-ui/core': minor
+---
+
+ListView: fold `data={{ provider: 'object', object }}` onto `objectName`, and read the
+author's view kind from `specType` / `type` (objectui#7477 — step 6 of #2890, released
+by the maintainer's ruling B on objectstack#14791, 2026-09-03).
+
+**What was broken.** A react page bound the way the published `react-blocks` contract
+recommends —
+
+```jsx
+<ListView data={{ provider: 'object', object: 'crm_task' }} type="kanban" />
+```
+
+— validated green against `@objectstack/spec` and then rendered an **empty grid** with no
+diagnostic. Both halves of that binding were inert in the renderer: `ListView` read
+`data.provider === 'object'` at zero sites (`'value'` and `'api'` are both live there, so
+the gap was real and not a dead instrument), and it read `specType` — the slot the react
+page tier parks an author's `type` in, because the SDUI envelope claims the `type` key
+(ADR-0078) — at zero sites, so an absent `viewType` forced the view to `grid`.
+
+**What changed.** `normalizeListViewSchema` (`@object-ui/core`) gains two folds. Per
+AGENTS.md #0.1 they live in the one documented normalizer — not as a seventh per-block
+copy of the six sibling `data.object` reads, and not as a renderer-side `??` dual-read.
+
+- `data: { provider: 'object', object }` → `objectName`. The `object` provider is a
+  `strictObject` carrying exactly `{ provider, object }`, so `objectName` captures all of
+  it. Two deliberate departures from the folds around it, both narrowing: an
+  already-present `objectName` **wins** (the fold only fills a gap and can never re-point
+  a binding that already resolves), and `data` is **not** deleted — it has four
+  providers, `api`/`value` are read live, and the block is forwarded to child views whose
+  own `getDataConfig` reads `data` before `objectName`.
+- the author's view kind is read from `specType`, then from a bare `type` when it names a
+  kind ListView draws (the component discriminator `'list-view'` never does) — the same
+  two legs, in the same order, as `normalizeChartSchema`'s chart-family read. An explicit
+  `viewType` still wins; this only fills the gap that used to resolve to `grid`, and a
+  kind ListView does not draw is left to that `grid` default rather than written through.
+
+**Accept behaviour widens.** Metadata that previously had no effect now binds a view: a
+list view carrying an `object` data source, or an author `type`, renders differently
+after this change than before. Nothing that renders today renders differently. No
+authored spelling is removed here — `objectName` / `viewType` remain accepted; their
+retirement is objectstack's, after this ships.
--- a/packages/core/src/utils/__tests__/normalize-list-view.test.ts
+++ b/packages/core/src/utils/__tests__/normalize-list-view.test.ts
@@ -395,6 +395,156 @@ describe('normalizeListViewSchema (#2890)', () => {
     });
   });
 
+  describe("data: { provider: 'object', object } \u2192 objectName (#7477)", () => {
+    it('folds the object provider\'s `object` onto `objectName`', () => {
+      const out = normalizeListViewSchema({
+        type: 'list-view',
+        viewType: 'grid',
+        data: { provider: 'object', object: 'crm_task' },
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBe('crm_task');
+    });
+
+    it('KEEPS `data` — unlike every other fold here, and deliberately', () => {
+      // `data` is not a legacy alias with one meaning: it has four providers,
+      // `api`/`value` are read live in ListView, and the whole block is
+      // forwarded to child views whose own `getDataConfig` reads it BEFORE
+      // `objectName`. Deleting it for one provider would rewrite what a child
+      // resolves. See the note on `normalizeListViewSchema`.
+      const data = { provider: 'object', object: 'crm_task' };
+      const out = normalizeListViewSchema({ viewType: 'grid', data }) as Record<string, unknown>;
+      expect(out.data).toEqual(data);
+    });
+
+    it('lets an existing `objectName` win — the fold only ever fills a gap', () => {
+      const out = normalizeListViewSchema({
+        viewType: 'grid',
+        objectName: 'crm_task',
+        data: { provider: 'object', object: 'crm_other' },
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBe('crm_task');
+    });
+
+    it('treats an EMPTY `objectName` as absent', () => {
+      const out = normalizeListViewSchema({
+        viewType: 'grid',
+        objectName: '',
+        data: { provider: 'object', object: 'crm_task' },
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBe('crm_task');
+    });
+
+    it.each(['api', 'value', 'schema'])('ignores the `%s` provider', (provider) => {
+      const out = normalizeListViewSchema({
+        viewType: 'grid',
+        data: { provider, object: 'crm_task' },
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBeUndefined();
+    });
+
+    it('invents no binding from an object provider that names no object', () => {
+      const out = normalizeListViewSchema({
+        viewType: 'grid',
+        data: { provider: 'object' },
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBeUndefined();
+    });
+
+    it.each<[unknown, string]>([['', 'empty string'], [42, 'number'], [null, 'null']])(
+      'ignores an `object` that is not a non-empty string (%s — %s)',
+      (object) => {
+        const out = normalizeListViewSchema({
+          viewType: 'grid',
+          data: { provider: 'object', object },
+        }) as Record<string, unknown>;
+        expect(out.objectName).toBeUndefined();
+      },
+    );
+
+    it('ignores a `data` that is an ARRAY (the inline-rows shorthand)', () => {
+      const out = normalizeListViewSchema({
+        viewType: 'grid',
+        data: [{ id: '1' }],
+      }) as Record<string, unknown>;
+      expect(out.objectName).toBeUndefined();
+    });
+
+    it('is idempotent', () => {
+      const once = normalizeListViewSchema({
+        viewType: 'grid',
+        data: { provider: 'object', object: 'crm_task' },
+      });
+      const twice = normalizeListViewSchema(once);
+      expect(twice).toEqual(once);
+      expect(twice).toBe(once); // objectName now present ⇒ nothing left to fold
+    });
+  });
+
+  describe('the author\'s view kind — `specType`, then `type` (#7477)', () => {
+    it('reads `specType`, where the react-page wrapper parks an author `type`', () => {
+      // The card's criterion: `type="kanban"` with no `viewType` must not be
+      // forced to `grid`.
+      const out = normalizeListViewSchema({
+        type: 'list-view',
+        specType: 'kanban',
+      }) as Record<string, unknown>;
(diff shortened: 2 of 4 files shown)


os-warren:
Contract review **PASS** at `CONTRACT_REVIEW_TIER` — director seat (objectstack #12708, session_01LsEjuNMPitCHwEfYftZ1um), 2026-09-04. Verdict, the semver reading and the ruling on the open question (A — an explicit `viewType` wins, as shipped) are on the card: #7477 comment 5538233866. Tier fuse read before the review (`get_session`: `claude-fable-5-1` served). Not governed (`packages/**` and `.changeset/` only). `needs:contract-review` removed here and on the card (read-modify-write, other labels untouched), PR marked ready, squash auto-merge armed; this seat follows it to MERGED and #7477 closes on `Fixes`.

---

PR #7999, opened 2026-09-06
test(network-escape): serve batch 1's five probes from doubles (ledger 21 → 16)

baozhoutao:
Part of #7307 — batch 1 of the burn-down. The card stays open until both ledgers are empty; 16 rows remain.

Five files stop opening a real socket, and their lines leave `KNOWN_ESCAPES` (in `vitest.setup.network-escape-guard.ts`) and `PINNED_LEDGER` (in `scripts/__tests__/network-escape-ledger.test.ts`) **in the same commit**, as the guard's own header requires.

**Ledger arithmetic: 21 → 16 in BOTH lists.** Verified in lockstep, not by counting twice: `diff` of the quoted paths in the two literals is empty, and the pin's two reconciles (grew / went stale) plus its non-vacuity floor are green.

## Per file — endpoint, mechanism, double

Every mechanism below was **traced**, not inferred: a stack probe injected at the guard's attribution point, run once per file, then reverted (mutation and restore both proven on disk; the guard's blob is byte-identical to HEAD).

| file | endpoint | mechanism (traced) | double |
|---|---|---|---|
| `examples/schema-catalog/test/catalog-gallery-render.test.tsx` | POST `/api/v1/security/explain` (18 per run) | grid-bearing tiles → `ObjectGrid.tsx:1407` → `useRecordCrudVerdicts.ts:199` `apiFetch ?? fetch` | explain router, per-test |
| `packages/plugin-view/src/__tests__/ObjectView.namedViewSortArity.test.tsx` | POST `/api/v1/security/explain` (8) | `ObjectView` → `ObjectGrid` → `useRecordCrudVerdicts.ts:199` | explain router, per-test |
| `packages/plugin-grid/src/__tests__/bulkDeleteVisibleWhen.test.tsx` | POST `/api/v1/security/explain` (4) | `ObjectGrid` → `useRecordCrudVerdicts.ts:199` | explain router, per-test |
| `packages/plugin-calendar/src/ObjectCalendar.navWidthDefault.test.tsx` | POST `/api/v1/security/explain` (6) | `RecordDetailDrawer` → `useRecordEditable.ts:75` `apiFetch ?? fetch` | explain router, per-test |
| `packages/plugin-charts/src/ObjectChart.heightChain.test.tsx` | GET `/api/v1/meta/object/task` (1) | `ObjectChart.tsx:390` → `loadObjectSchema` (:411) → `loadDimensionFieldMeta` | meta-object router, per-test |

All five take an `apiFetch ?? fetch` fallback with no host `apiFetch` in the tree, exactly as #7307 measured. **Zone 2 A2 holds, with one refinement worth recording:** the dispatch expected the explain rows to share ONE hook. They do not — three of the four go through plugin-grid's `useRecordCrudVerdicts`, the calendar one through plugin-detail's `useRecordEditable`. One router *shape* still serves them, because both hooks POST the same route; the two differ only in the response they read, and the double answers both (see below).

### The double, and why it changes no assertion

The landed #5225 shape (`vi.stubGlobal('fetch', router)`, and `cleanup()` **before** `vi.unstubAllGlobals()` — #7439's ordering, so the tree is unmounted while the double is still installed). Not a blanket network stub: it is a **recording router**, and `afterEach` fails on any URL that is not the route it serves, so a new escape reds here instead of vanishing into the hook's best-effort `catch`. That assertion is also what stops the double degrading into a silencer: a route regex that stopped matching would 404 into the same fail-open path and look green, and this fails instead.

The explain router answers the **permissive** verdict in the two shapes the two hooks read (ADR-0090 D6 / ADR-0095 C2), keyed off the request body: `{ record: { visible: true } }` for a single `recordId`, `{ records: [{ recordId, visible: true }] }` for a batched `recordIds`. That is the same downstream state the failing request produced, at every read site:

- `useRecordEditable` initialises `allowed` to `true` and its failure path leaves it there;
- the only consumer of the batched lookup is `resolveRowRecordCrudAffordance` (`rowCrudAffordances.ts:224`), whose rule is `!!objectVerdict && recordVerdict !== false` — so `true` and the `undefined` the failing request produced are the same value there.

The charts router serves a field-less document, so `loadDimensionFieldMeta` resolves no option colours and `optionMeta` settles null — the state the failing request already produced; `ChartRenderer` is mocked to null in that file anyway.

## Zone 2 A3 — the ablation (the guard IS the acceptance criterion)

On the committed tree, one script with `trap ... EXIT INT TERM` and absolute paths: one converted file's double reverted to the base blob while its line stays deleted from both ledgers.

- mutation proven on disk: `packages/plugin-calendar/src/ObjectCalendar.navWidthDefault.test.tsx` blob `6146d87e` (HEAD) → `f03ca1ff` (equal to the `36fc746` base blob, checked); anchors `installExplainDouble` 2 → 0, `vi.stubGlobal` 1 → 0; ledger state during the run confirmed: that path appears **0** times in `KNOWN_ESCAPES` and **0** times in `PINNED_LEDGER`;
- **predicted direction: turns red. Observed: red** — exit **1**, `Tests 3 failed (3)`, every one of them `Network escape: this test reached a REAL socket at http://localhost:3000/api/v1/security/explain`, `file: packages/plugin-calendar/src/ObjectCalendar.navWidthDefault.test.tsx`;
- restore: `git checkout HEAD -- path` → blob back to `6146d87e`, `git diff HEAD` on that path **EMPTY**, `git status --porcelain` empty.

No `dist/` preflight leg: these are the vitest projects' own test files, read from disk by the runner, and the root config aliases every package specifier to `src` — measured, in that all five suites ran green on a completely unbuilt tree before any build happened here.

## Before / after, per file

| file | before | after |
|---|---|---|
| catalog-gallery-render | 18 attribution lines, 90 `ECONNREFUSED` lines | **0 / 0**, `Tests 586 passed` |
| ObjectView.namedViewSortArity | 8 / 40 | **0 / 0**, `Tests 4 passed` |
| bulkDeleteVisibleWhen | 4 / 20 | **0 / 0**, `Tests 4 passed` |
| ObjectCalendar.navWidthDefault | 6 / 30 | **0 / 0**, `Tests 3 passed` |
| ObjectChart.heightChain | 1 / 3 | **0 / 0**, `Tests 1 passed` |

Final run at `0e92bee` over the five files plus the pin: exit **0**, `Test Files 6 passed (6)`, `Tests 601 passed (601)`, zero lines matching `network-escape` or `ECONNREFUSED`.

## Suites and gates — at `0e92bee`, exit codes captured by redirect-then-capture

- **Whole affected packages**, plus the pin and the one other reader: `pnpm exec vitest run packages/plugin-calendar/ packages/plugin-grid/ packages/plugin-view/ packages/plugin-charts/ examples/schema-catalog/ scripts/__tests__/network-escape-ledger.test.ts packages/plugin-detail/src/renderers/__tests__/record-details.hideEmptyRetired-7129.test.tsx` → exit **0**, `Test Files 247 passed (247)`, `Tests 3960 passed (3960)`, **0** `network-escape` lines anywhere in the run.
- `node scripts/check-changeset-presence.mjs` → exit **0**: `4 source file(s) of 4 released package(s) changed, and this change declares 1 changeset(s)` / `Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.` (Test files under a released package's `src/` do owe one; answered the repo's way, not with a label.)
- `pnpm check:control-bytes` → exit **0**, `scanned 6435 tracked text file(s); skipped 85 binary`. Plus a self-scan beyond the gate: `grep -naP` for the control range over all 8 changed paths — no hits.
- `node scripts/check-governed-queue-guard.mjs --test` over all 8 changed paths → exit **0**, `NOT GOVERNED — 8 path(s) checked against 5 governed surface(s); none matched.`
- `pnpm type-check:scripts` → exit **0**.
- Per package `type-check` and `lint` (plugin-calendar, plugin-charts, plugin-grid, plugin-view, examples/schema-catalog): all exit **0**, **0 errors**. Proven non-vacuous rather than assumed: `tsc -p tsconfig.test.json --listFiles` names each edited test file (1 hit each), and `eslint --format json` reports each edited file among the linted set. Warning counts are unchanged — every `no-explicit-any` warning in the touched files sits on a pre-existing line, outside this diff's hunks; the added code uses `unknown`.
- Adjacent gates that read test files: `pnpm check:vi-mock-specifiers` **0**, `pnpm check:vi-mock-inherit` **0**, `pnpm changeset:check` **0**, `pnpm check:unreferenced-sources` **0**, `pnpm lint:coverage` **0** (`46/46 packages linted, 0 with outstanding errors`).
- Build: `turbo run build --filter=./packages/* --concurrency=2` → exit **0**, `Tasks: 39 successful, 39 total`. Needed only because each package's `type-check` depends on `^build`; the suites themselves need no `dist/`.

**Readers of the two symbols** (`git grep -l` over `scripts/ packages/ .github/ vitest.*`, unpiped): `KNOWN_ESCAPES` → `vitest.setup.network-escape-guard.ts`, `scripts/__tests__/network-escape-ledger.test.ts`, and `packages/plugin-detail/src/renderers/__tests__/record-details.hideEmptyRetired-7129.test.tsx` (a prose reference in a comment — it names the list to say it deliberately does NOT join it; no code reads it). `PINNED_LEDGER` → the first two only. Nothing in `.github/` names either. All three files were run above.

## Docstring prose

Two sentences in the guard and one in the pin said "the 21 files"; the sets now hold 16. Reworded to "what REMAINS of the 21 files measured on `67dadd6`" so the provenance survives without a count that every batch would have to re-edit. No judgement byte moves in either file.

## Not in this batch, and why

- **kanban (3 rows) and gantt (1 row) were NOT taken.** The claim gated them on PR #7985 being merged; read on GitHub immediately before they would have been edited, #7985 is `state: open`, `merged: false`, `mergeable_state: unstable`. They stay for batch 2.
- app-shell's 8 rows and plugin-detail's 4 rows are batch 2 by the claim, and were not touched.

## Out of scope, filed

#7996 (`finding`) — the one other landed `security/explain` double, in `record-details.hideEmptyRetired-7129.test.tsx`, answers `{ allowed: true }`, a key neither explain hook reads. It passes only because both hooks then take the same fail-open path they take on `ECONNREFUSED`, so it pins a wire contract that does not exist. Not touched here: that file is not on the ledger and is outside this batch's file surface.

## CI note

---

---

### DIFF
--- a/.changeset/network-escape-batch1.md
+++ b/.changeset/network-escape-batch1.md
@@ -1,0 +1,4 @@
+---
+---
+
+Test-only change: the five network-escape files in batch 1 of objectui#7307 now serve their `/api/v1/security/explain` and `/api/v1/meta/object/<name>` probes from a recording double instead of a real socket, and their lines leave `KNOWN_ESCAPES` and `PINNED_LEDGER` together. No published behaviour changes — no product source is touched.
--- a/examples/schema-catalog/test/catalog-gallery-render.test.tsx
+++ b/examples/schema-catalog/test/catalog-gallery-render.test.tsx
@@ -125,8 +125,8 @@
  * entries render `role="alert"` because that is what an Alert IS — the
  * assertion is correct for a dashboard tile and wrong for the corpus.
  */
-import { describe, it, expect, afterAll } from 'vitest';
-import { render, waitFor } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach, afterEach, afterAll } from 'vitest';
+import { render, waitFor, cleanup } from '@testing-library/react';
 import '@object-ui/components';
 // Mirrors apps/site/app/components/registerCatalogBlocks.ts, in its order.
 import '@object-ui/plugin-dashboard';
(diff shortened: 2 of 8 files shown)


baozhoutao:
**Standing down on `Live E2E (informational)` — red on the base branch too, not this PR's.** `domain:devx @ objectui` execution seat, PM session `session_01MM7kaS4dPpYHV5BsMyu4tQ`, R45, 2026-09-06T07:52Z. Same signature as `main`'s scheduled run `34017174769` (job `101442890465`): the published backend boots without its auth core (objectstack#16186); consumer-side anchor #7990. This diff is five test files, the two ledgers and an empty-frontmatter changeset — no backend pin, no `e2e/` path. No fix to port, no re-run spent. The flip and arming wait on the remaining shards.

---

PR #8003, opened 2026-09-06
fix(build): declare console's root-published plugin entry as a turbo build output

### LINKED ISSUE #7855: `@object-ui/console` publishes `plugin.js` from the package root, which turbo's build `outputs` do not cover — a cache hit restores an incomplete package
Found by the dev seat implementing #7276 while re-deriving that card's mechanism. Unassigned, unlabelled — recording a measurement, not claiming a priority.

## The defect

`turbo.json`'s `build` task declares:

```json
"outputs": ["dist/**", ".next/**", "build/**", "**/*.tsbuildinfo", "!**/node_modules/**"]
```

`@object-ui/console` (`apps/console`, not private) publishes its entry from the PACKAGE ROOT, not from `dist/`:

```json
"exports": { ".": { "types": "./plugin.d.ts", "import": "./plugin.js", "default": "./plugin.js" } },
"main": "./plugin.js",
"files": ["dist", "plugin.ts", "plugin.js", "plugin.d.ts", "README.md"]
```

`plugin.js` and `plugin.d.ts` are produced by `build:plugin` (`tsc -p tsconfig.plugin.json`), are gitignored (`apps/console/.gitignore:16`), and match none of the `outputs` globs. So they are neither hashed as inputs nor stored as outputs: **a cache hit for `@object-ui/console:build` restores `dist/` and leaves the package's published entry absent.**

## Measured

On `origin/main` `52cac38`, two worktrees of the same commit, turbo 2.10.9:

1. Worktree A ran a cold `pnpm build` — 43 tasks, 3m46.7s. `apps/console/plugin.js` present.
2. Worktree B ran `pnpm build` — 43 of 43 cache hits, 272ms, `FULL TURBO`.
3. In worktree B: `apps/console/dist/` fully restored, `apps/console/plugin.js` **does not exist**.

`pnpm check:node-esm-load` in worktree B then reported exactly the line #7276 quoted from a real run:

```
✗ @object-ui/console: no-build-output — ./plugin.js does not exist after the build
```

## Why this is worth a card

This is the mechanism behind the loud half of #7276, and it outlives that card's fix. #7276 makes the ESM-load gate refuse to grade artifacts another tree produced — that turns the misattributed red into an accurate one, but it does not make the cache entry complete. Anything else that consumes a cached `@object-ui/console` build still gets a package whose declared entry is missing, and the tell is a `no-build-output` style failure pointing at a package the author did not touch.

The narrow fix is to declare the two files as outputs — either widening the shared `build` task in `turbo.json`, or (smaller blast radius) giving `apps/console` a package-level `turbo.build.outputs` that adds `plugin.js` and `plugin.d.ts`. Whether any OTHER package publishes from its root should be checked in the same pass rather than assumed: the audit is `exports`/`main` targets that do not start with `./dist/`.

Not fixed in #7276's PR: that card's declared file surface is the gate script and its test, and `turbo.json` was explicitly limited there to "only if the fix requires it". It did not.

Refs: #7276 (where this surfaced) · #1760 (the adjacent cache-masking case)

---

baozhoutao:
Fixes #7855

`@object-ui/console` publishes `main` / `types` / `exports` from `./plugin.js` and `./plugin.d.ts` at the **package root**. Both are emitted by `build:plugin` (`tsc -p tsconfig.plugin.json`) and gitignored, and the shared `build` task's `outputs` (`dist/**`, `.next/**`, `build/**`, `**/*.tsbuildinfo`, minus `node_modules`) match neither — so turbo never **stored** them. A cache hit restored `dist/` and left the package's declared entry absent.

Two files: `apps/console/turbo.json` (new) and one new pin under `scripts/__tests__/`. No package source, no manifest field, no change to the shared task.

## A2 — the card's measurement, reproduced and then closed

One worktree, one **fresh** `--cache-dir` under the scratchpad (turbo announces `Remote caching disabled, using shared worktree cache`, so a sibling agent's artifacts could otherwise satisfy a leg). Cold build, wipe `dist/` + `plugin.js` + `plugin.d.ts` + `*.tsbuildinfo`, replay.

**Before** (tip `bce5971`, no fix):

```
@object-ui/console:build: cache hit, replaying logs a218494ab23eb6f3
 Tasks:    35 successful, 35 total
Cached:    35 cached, 35 total
  Time:    248ms >>> FULL TURBO

dist entries: 12
ls: cannot access 'apps/console/plugin.js': No such file or directory
ls: cannot access 'apps/console/plugin.d.ts': No such file or directory
```

`node scripts/check-node-esm-load.mjs --no-build` on that exact cache-hit tree, **exit 1**:

```
Provenance leg: 35 of 37 gradable entries were built by this tree.
Load leg: 31 of 39 published ESM entries imported and evaluated.
✗ @object-ui/console: no-build-output — ./plugin.js does not exist after the build
```

**After** (this branch, same fresh cache dir, cold rebuild then wipe then replay):

```
@object-ui/console:build: cache hit, replaying logs 68390122c3c328b0
 Tasks:    35 successful, 35 total
Cached:    35 cached, 35 total
  Time:    178ms >>> FULL TURBO

dist entries: 12
-rw-r--r-- 1 root root 1248 apps/console/plugin.d.ts
-rw-r--r-- 1 root root 2033 apps/console/plugin.js
```

`node scripts/check-node-esm-load.mjs --no-build`, **exit 0**:

```
Provenance leg: 37 of 37 gradable entries were built by this tree.
Load leg: 34 of 39 published ESM entries imported and evaluated.
```

`dist/` is still restored on the hit — the package configuration **adds**, it does not trade one for the other.

Two narrowings, stated rather than hidden: the gate was run as `--no-build` on the tree the replay produced, so the reading is about **this** worktree's cache dir rather than the shared one the plain `pnpm check:node-esm-load` would build into; and reaching `37 of 37` needed `@object-ui/cli` and `@object-ui/plugin-ai` (outside the console's dependency closure) built into the same fresh cache first — before that they were **refused on provenance**, not graded, in both legs alike.

## Deviation from the dispatch's spelling — measured, not preferred

The ruling named a package-level `turbo` key inside `apps/console/package.json`. **That spelling does nothing on turbo 2.10.9.** Measured with `turbo run build --filter=@object-ui/console --dry=json`, both shapes of the manifest key:

| shape | resolved `outputs` | task hash |
|---|---|---|
| baseline, no package config | `["**/*.tsbuildinfo",".next/**","build/**","dist/**"]` | `a218494ab23eb6f3` |
| `package.json` → `turbo.tasks.build.outputs` | unchanged | `4098312cd6437d3a` |
| `package.json` → `turbo.build.outputs` | unchanged | `5c19da105b7841bf` |
| `apps/console/turbo.json`, `extends` `["//"]` | `[…,"plugin.d.ts","plugin.js"]` | `68390122c3c328b0` |

turbo accepts the manifest key **silently** — exit 0, no warning — and declares nothing, while still perturbing the hash (the manifest is an input), i.e. it buys a cache invalidation in exchange for nothing. So the fix takes the same route the ruling chose (package-level, one package's blast radius) in the spelling that turbo 2.x actually reads. The forbidden alternative — widening the shared task in the root `turbo.json` — was not used and was not needed.

## Why the config restates the root globs

A package configuration **overrides** the key it declares; it does not append. turbo 2.10.9's own bundled `schema.json` (`node_modules/turbo/schema.json`), on `extends`:

> This key is only available in Workspace Configs and cannot be used in your root `turbo.json`.
>
> Tells turbo to extend your root `turbo.json` and overrides with the keys provided in your Workspace Configs. Currently, only the `["//"]` value is allowed.

Measured, a config declaring only the two additions:

```
outputs ["plugin.d.ts","plugin.js"]
excluded null
```

`dist/**` gone. Keys the config does **not** declare are inherited intact — `dependsOn` still resolves to the same 34 upstream builds, `inputs` and `env` unchanged. And `extends` is mandatory; without it turbo refuses the file loudly (`x No "extends" key found`), which is the one failure mode here that cannot be silent.

## The audit (#7855's second deliverable)

Re-derived over all 46 workspace manifests (`packages/*`, `apps/*`, `examples/*`, `docs`), resolving `main`, `module`, `types`, `typings`, `browser`, every `exports` leaf and `bin`. **13 targets across 4 packages** do not resolve under `./dist/`:

| package | targets outside `./dist/` | classification |
|---|---|---|
| `@object-ui/console` (`apps/console`) | `main`, `types`, and 3 `exports` conditions, all `./plugin.js` / `./plugin.d.ts` | **built + gitignored + undeclared — the defect, fixed here** |
| `@object-ui/app-shell` | `exports["./styles.css"]` = `./src/styles.css` | tracked source, shipped as authored — not a build output |
| `@object-ui/example-schema-catalog` (private) | `exports["./schemas/*"]` = `./src/schemas/*` | tracked source, 431 files — not a build output |
| `@object-ui/test-support` (private, no `build` script) | 6 targets under `./src/` | tracked source — nothing builds them |

The two candidates the dispatch flagged for checking both resolve **under** `./dist/` and are covered: `packages/types`' subpath exports are `./dist/base.js`, `./dist/layout.js`, … and `packages/i18n`'s `./locales/*` is `./dist/locales/*.js`. No package with a `dist/` entry lacks a `build` script. So `@object-ui/console` is the only instance of this defect shape, and nothing else was fixed or filed.

## The pin

`scripts/__tests__/turbo-build-outputs-cover-published-entries.test.ts`, beside the existing turbo-config guards and reusing `helpers/turbo-inputs.ts` unchanged (`workspacePackageDirs`, `globToRegExp`, `rel`, `repoRoot`). One rule over every published entry of every workspace package: **it is either tracked source, or a build output the task declares** — stated so it holds for packages that do not exist yet.

- 46 per-package assertions, plus a **liveness** assertion (a sweep that finds no packages or no built entries would make the rest vacuously green).
- **No package configuration narrows the root `outputs`** — the trap that arrives with the fix, since trimming the config to "just my two files" silently stops caching `dist/` while everything still builds and passes.
- A **positive control on every run**: it plants the exact #7855 shape (a gitignored `./plugin.js` published from a package root) and requires the real predicate — not a paraphrase — to name it, then requires the same entry under `dist/` to fall silent, so the control cannot pass by flagging everything.

One measured subtlety is baked in: Node's `exports` star matches **across** `/`, unlike a turbo glob's `*`. Translating it with the turbo-glob translation flagged `@object-ui/example-schema-catalog` — whose 431 tracked schemas live two directories below the star — as if it published an undeclared build output. The index search uses Node's semantics; the first draft of this guard was red for exactly that reason.

**Ablation** (fix committed first, then mutated, proven on disk, restored to the byte):

| leg | mutation, proven on disk | result |
|---|---|---|
| 1 | `apps/console/turbo.json` deleted (`git status` = ` D `) | pin **RED**: `× @object-ui/console` — "publishes ./plugin.js, ./plugin.d.ts, … git does not track them" |
| 2 | config narrowed to the two additions (`"dist/**"` occurrences 1 then 0) | pin **RED**: `× no package configuration narrows the root outputs` — `drops ["dist/**",".next/**","build/**","**/*.tsbuildinfo","!**/node_modules/**"]` |

Both legs restored with `git checkout HEAD -- apps/console/turbo.json`; restored blob `ae6d38c…` equals the HEAD blob, `git diff HEAD` empty, tree clean. Each leg ran under a `trap … EXIT INT TERM` restore.

## Gates (exit codes captured by redirect, never through a pipe)

| gate | exit | verdict line |
|---|---|---|
| the new pin | 0 | `Test Files 1 passed (1) / Tests 49 passed (49)` |
| whole `scripts/__tests__/` (108 files — every reader of `turbo.json` and of `apps/console/package.json`) | 0 | `Test Files 108 passed (108) / Tests 3294 passed (3294)` |
| `turbo-build-inputs.test.ts` (untouched, as required) | 0 | `Tests 46 passed (46)` |
| `pnpm type-check:scripts` | 0 | `--listFiles` confirms the new pin **is in** the program (1 hit) — not assumed |
| `node scripts/check-changeset-presence.mjs` | 0 | `2 file(s) changed, 0 of them published source …, 0 of them a manifest whose published contract moved` — no changeset owed, and the reading was taken **after** staging (untracked files are invisible to it) |
| `node scripts/check-changeset-no-major.mjs` | 0 | `No changeset declares a major bump.` |
| `pnpm check:control-bytes` | 0 | `scanned 6438 tracked text file(s)` (6436 before staging) |
| `node scripts/check-governed-queue-guard.mjs --test` (both paths) | 0 | `NOT GOVERNED — 2 path(s) checked against 5 governed surface(s); none matched.` |
| `apps/console` `type-check` | 0 | `tsc --noEmit && tsc -b tsconfig.node.json --force` |
| `apps/console` `lint` | 0 | `209 problems (0 errors, 209 warnings)` — all pre-existing |
| control-byte self-scan over both files | 1 (no hits) | `grep -naP` for `\x00-\x08\x0b\x0c\x0e-\x1f\x7f` |

Ordinary changes for build-cache purposes only; `apps/console/turbo.json` is not in the package's `files` list, so nothing new is published.

---

### DIFF
--- a/apps/console/turbo.json
+++ b/apps/console/turbo.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "outputs": [
+        "dist/**",
+        ".next/**",
+        "build/**",
+        "**/*.tsbuildinfo",
+        "!**/node_modules/**",
+        "plugin.js",
+        "plugin.d.ts"
+      ]
+    }
+  }
+}
--- a/scripts/__tests__/turbo-build-outputs-cover-published-entries.test.ts
+++ b/scripts/__tests__/turbo-build-outputs-cover-published-entries.test.ts
@@ -1,0 +1,338 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+import { execFileSync } from 'node:child_process';
+import { globToRegExp, rel, repoRoot, workspacePackageDirs } from './helpers/turbo-inputs';
+
+/**
+ * objectui#7855 — turbo's `build` task must declare every file the build EMITS
+ * that the package then PUBLISHES, or a cache hit restores an incomplete
+ * package.
+ *
+ * `turbo.json`'s `build.outputs` was `["dist/**", ".next/**", "build/**",
+ * "**\/*.tsbuildinfo", "!**\/node_modules/**"]`, and `@object-ui/console`
+ * publishes its entry from the PACKAGE ROOT:
+ *
+ *     "main":    "./plugin.js",
+ *     "types":   "./plugin.d.ts",
+ *     "exports": { ".": { "import": "./plugin.js", … } }
+ *
+ * Those two files are emitted by `build:plugin` (`tsc -p tsconfig.plugin.json`)
+ * and gitignored (`apps/console/.gitignore`), so they matched no `outputs` glob
+ * and were neither hashed as inputs nor STORED as outputs. Measured on
+ * `bce5971`, one worktree, one fresh `--cache-dir`, cold build then wipe then
+ * replay:
+ *
+ *     @object-ui/console:build: cache hit, replaying logs a218494ab23eb6f3
+ *      Tasks: 35 successful, 35 total   Cached: 35 cached, 35 total
+ *      Time:  248ms >>> FULL TURBO
+ *     apps/console/dist/        restored, 12 entries
+ *     apps/console/plugin.js    ABSENT
+ *
+ *     node scripts/check-node-esm-load.mjs --no-build   → exit 1
+ *     ✗ @object-ui/console: no-build-output — ./plugin.js does not exist after the build
+ *
+ * The tell is the worst kind: a red naming a package the author never touched.
+ *
+ * ## What this guard asserts
+ *
+ * One rule over every published entry target of every workspace package: it is
+ * either TRACKED SOURCE (shipped as authored) or a BUILD OUTPUT the `build`
+ * task declares. An entry that is neither is a file the cache will not restore
+ * — which is the defect above, stated so it holds for packages that do not
+ * exist yet.
+ *
+ * ## Why the effective outputs are a UNION spelled out by hand
+ *
+ * A package-level configuration OVERRIDES the root's key rather than appending
+ * to it. turbo 2.10.9's own bundled `schema.json` says so for `extends`
+ * (quoted verbatim; it is the version this repo pins):
+ *
+ *   > This key is only available in Workspace Configs and cannot be used in
+ *   > your root `turbo.json`. Tells turbo to extend your root `turbo.json` and
+ *   > overrides with the keys provided in your Workspace Configs. Currently,
+ *   > only the `["//"]` value is allowed.
+ *
+ * Measured with `turbo run build --filter=@object-ui/console --dry=json`: a
+ * package config declaring `outputs: ["plugin.js", "plugin.d.ts"]` resolves to
+ * `outputs ["plugin.d.ts","plugin.js"]` and `excludedOutputs null` — `dist/**`
+ * is GONE. So `apps/console/turbo.json` restates the root's globs alongside its
+ * two additions, and the second assertion below keeps it that way: a package
+ * config that drops a root glob silently stops caching whatever that glob
+ * covered, and nothing else in the repository would notice.
+ */
+
+/** The turbo task this guard is about. */
+const TASK = 'build';
+
+/** Manifest keys that name a published entry file, besides `exports`/`bin`. */
+const ENTRY_KEYS = ['main', 'module', 'types', 'typings', 'browser'] as const;
+
+interface EntryTarget {
+  /** Where in the manifest it was declared, for the failure message. */
+  readonly from: string;
+  /** Verbatim, as the manifest spells it. */
+  readonly target: string;
+  /** Package-relative, `./` stripped. May still contain a `*`. */
+  readonly relative: string;
+  /** Package-relative, `*` replaced by a probe segment, for output matching. */
+  readonly probe: string;
+}
+
+interface PackageEntries {
+  readonly name: string;
+  /** Repo-relative package directory. */
+  readonly dir: string;
+  readonly hasBuildScript: boolean;
+  readonly entries: readonly EntryTarget[];
+  /** Effective `outputs` for this package: root task, overridden by its own config. */
+  readonly outputs: readonly string[];
+}
+
+/**
+ * A wildcard subpath export (`"./locales/*": "./dist/locales/*.js"`) is never
+ * expanded against the tree on disk. Expanding would make the sweep depend on
+ * the tree having been BUILT, and a guard that quietly shrinks to nothing when
+ * `dist/` is absent is the failure mode this family exists to prevent. Both
+ * questions are answered without it: the git index is searched with the target
+ * as a PATTERN, and the output globs are matched against the target with one
+ * synthetic segment standing in for the star.
+ *
+ * Node's `exports` star is NOT a glob star: it matches across `/`, so
+ * `"./schemas/*": "./src/schemas/*"` resolves `examples/schema-catalog`'s 431
+ * tracked files, which live two directories down. The index search therefore
+ * translates it as `.*` rather than reusing the turbo-glob translation, whose
+ * `*` stops at a separator — that mismatch flagged a package publishing plain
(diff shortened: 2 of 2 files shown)


baozhoutao:
**Standing down on `Live E2E (informational)` — red on the base branch too, not this PR's.** `domain:devx @ objectui` execution seat, PM session `session_01MM7kaS4dPpYHV5BsMyu4tQ`, R45, 2026-09-06T08:08Z. Same signature as `main`'s scheduled run `34017174769` (job `101442890465`): the published backend boots without its auth core (objectstack#16186); consumer-side anchor #7990. This diff is a package-level `turbo.json` for `apps/console` and one `scripts/__tests__/` pin — no backend pin, no `e2e/` path. No fix to port, no re-run spent. Flip and arming wait on the remaining shards (ACCEPT `5557906532` on #7855).

---

PR #8013, opened 2026-09-06
test(plugins): serve batch 2's four network-escape probes from doubles

baozhoutao:
Part of #7307 — batch 2 of the network-escape burn-down (batch 1 was #7999).

The four `/api/v1/security/explain` rows left in the ledger for `plugin-kanban` and
`plugin-gantt` now serve that probe from the recording double batch 1 landed, and
their lines leave `KNOWN_ESCAPES` and `PINNED_LEDGER` in the same commit.

## Per file

| file | endpoint | hook it escaped through | double |
|:--|:--|:--|:--|
| `packages/plugin-kanban/src/ObjectKanban.navWidthDefault.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` (`apiFetch ?? fetch`) | per-test `installExplainDouble()` in `beforeEach` |
| `packages/plugin-kanban/src/ObjectKanban.overlayTitleI18n.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` | per-test `installExplainDouble()` in `beforeEach` |
| `packages/plugin-kanban/src/ObjectKanban.overlayTitleNoProviderFallback.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` | per-test `installExplainDouble()` in `beforeEach` |
| `packages/plugin-gantt/src/ObjectGantt.navWidthDefault.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` | per-test `installExplainDouble()` in `beforeEach` |

**Mechanism, measured rather than assumed.** A trap-guarded stack probe on the
guard's own attribution point says all four take ONE route, not the two batch 1
found: `RecordDetailDrawer` renders `DetailView`, which calls `useRecordEditable`
twice per open (`DetailView.tsx:290` for edit, `:296` for delete). At
`useRecordEditable.ts:76` the hook degrades from the host's authenticated
`apiFetch` to the global `fetch` — by design, so a standalone embed keeps
rendering — and under happy-dom that global is a real HTTP client whose document
URL is `http://localhost:3000`, so the relative path resolved to a live socket.
The read is best-effort (`catch` leaves the record editable), which is why every
one of these files stayed green while its request always failed. The probe was
reverted by `git checkout HEAD --` and the guard is byte-identical to `HEAD`
afterwards.

**The double** is batch 1's shape verbatim: `vi.stubGlobal('fetch', router)` in
`beforeEach`, `cleanup()` before `vi.unstubAllGlobals()` in `afterEach` (#7439
ordering). It is a router, not a sink — it records every URL it is handed and
`afterEach` fails on any URL that is not the explain route, so an escape
elsewhere reds here instead of vanishing into the hook's `catch`. It answers the
permissive verdict, which is the same downstream state the failing request
produced (`useRecordEditable` initialises `allowed` to `true` and its failure
path leaves it there), so no existing assertion changes meaning. Only the
single-`recordId` response shape is reached from these four files; the batched
branch is kept so the router stays byte-identical to its siblings rather than
forking per file. Nothing is skipped, quarantined or silenced.

## Ledger arithmetic

**16 to 12, in both lists.** The four names are deleted from `KNOWN_ESCAPES` in
`vitest.setup.network-escape-guard.ts` and from `PINNED_LEDGER` in
`scripts/__tests__/network-escape-ledger.test.ts` in this one commit. Checked in
lockstep by diffing the quoted paths of the two literals against each other on
the branch: empty diff, both at 12. Remaining 12: app-shell 8, plugin-detail 4.

## Evidence

post-fix run):

| file | before | after |
|:--|:--|:--|
| `ObjectKanban.navWidthDefault` | 6 / 30 | 0 / 0, `Tests 3 passed` |
| `ObjectKanban.overlayTitleI18n` | 12 / 60 | 0 / 0, `Tests 7 passed` |
| `ObjectKanban.overlayTitleNoProviderFallback` | 4 / 20 | 0 / 0, `Tests 2 passed` |
| `ObjectGantt.navWidthDefault` | 4 / 20 | 0 / 0, `Tests 2 passed` |

`pnpm exec vitest run packages/plugin-kanban/ packages/plugin-gantt/ scripts/__tests__/network-escape-ledger.test.ts`
on this head: exit 0, `Test Files 87 passed (87)`, `Tests 619 passed (619)`, and
zero lines matching `network-escape` or `ECONNREFUSED` in the whole run.

absolute paths). One converted file's double reverted to its pre-batch-2 blob
while its line stayed deleted from BOTH ledgers. Mutation proven on disk:
`ObjectGantt.navWidthDefault.test.tsx` blob `e9ba5b8e` to `28f37a60` (equal to
the `e1545cf` base blob), anchors `installExplainDouble` 2 to 0 and
`vi.stubGlobal` 1 to 0, that path present 0 times in each ledger during the run.
Predicted direction red; OBSERVED red — exit 1, `Tests 2 failed (2)`,
`Network escape: this test reached a REAL socket at http://localhost:3000/api/v1/security/explain`
naming `file: packages/plugin-gantt/src/ObjectGantt.navWidthDefault.test.tsx`.
Restore proven: `git checkout HEAD -- path`, blob back to `e9ba5b8e`,
`git diff HEAD` on that path empty, `git status --porcelain` empty.

**Gates**, exit codes captured by redirect-then-capture, never through a pipe:

- `node scripts/check-changeset-presence.mjs` exit 0 — "4 source file(s) of 2
  changeset has an EMPTY frontmatter, the explicit exemption for a test-only
  change under a released package.
- `pnpm check:control-bytes` exit 0 (6443 tracked text files), plus a
  `grep -naP` self-scan of the control range over all 7 changed paths, no hits.
- `node scripts/check-governed-queue-guard.mjs --test` over all 7 changed paths:
  exit 0, "NOT GOVERNED — 7 path(s) checked against 5 governed surface(s)".
- `pnpm type-check:scripts` exit 0 · `pnpm check:vi-mock-inherit` exit 0 ·
  `pnpm check:vi-mock-specifiers` exit 0.
- Per-package `type-check` and `lint` for `plugin-kanban` and `plugin-gantt`:
  exit 0, 0 errors, proven non-vacuous rather than assumed —
  `tsc -p tsconfig.test.json --listFiles` names each edited test file exactly
  once, and `eslint --format json` lists each edited file among the linted set
  (37 and 91 files). Warning counts unchanged: every `no-explicit-any` warning in
  the touched files sits on a pre-existing line above this diff's hunks. The
  `type-check` task `dependsOn ^build`, so turbo built the dependency closure.

## Concurrency note (A3)

**Yes — a #5174 batch-18 PR on one of these packages exists: #8009,
`docs(plugin-gantt): compile the gantt README's 18 blocks and drop its ledger
entry`**, opened after this branch was cut. Its file list read over git is
`packages/plugin-gantt/README.md` and `scripts/check-doc-snippet-types.mjs` —
zero overlap with the 7 files here. `main` moved by two commits while this batch
was in flight (#8003, #8002) and has been merged into this branch; the suites,
the ablation and every gate above were re-run on the merged head after that.

(#7990 / objectstack#16186), not for anything in this diff.

---

### DIFF
--- a/.changeset/network-escape-batch2.md
+++ b/.changeset/network-escape-batch2.md
@@ -1,0 +1,4 @@
+---
+---
+
+Test-only change: the four network-escape files in batch 2 of objectui#7307 (three in `plugin-kanban`, one in `plugin-gantt`) now serve their `/api/v1/security/explain` probe from a recording double instead of a real socket, and their lines leave `KNOWN_ESCAPES` and `PINNED_LEDGER` together. No published behaviour changes — no product source is touched.
--- a/packages/plugin-gantt/src/ObjectGantt.navWidthDefault.test.tsx
+++ b/packages/plugin-gantt/src/ObjectGantt.navWidthDefault.test.tsx
@@ -24,8 +24,8 @@
  * renders", either of which passes in both worlds.
  */
 import React from 'react';
-import { render, screen, waitFor, fireEvent } from '@testing-library/react';
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor, fireEvent, cleanup } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { ObjectGantt } from './ObjectGantt';
 
 /** The width the gantt's drawer has always resolved to. Must not drift. */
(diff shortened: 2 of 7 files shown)


baozhoutao:
Standing-down note — `Live E2E (informational)` is red on this head (`59ccee7`) as it is on `main` and on every branch today: the upstream base-red anchored on #7990 (objectstack#16186). It is `continue-on-error` by design, never blocks the queue, and touches nothing this diff changes (four test files, the two ledger literals, one empty changeset). Not this PR's to fix; no re-run requested. Every other non-skipped check is green at 09:20Z. Flipping ready and arming.

---

PR #8019, opened 2026-09-06
test(plugin-detail): serve batch 3's four probes from doubles (objectui#7307)

baozhoutao:
Part of #7307 — batch 3 of the network-escape burn-down (batch 1 was #7999, batch 2 was #8013).

The four `plugin-detail` rows left in the ledger now serve their probes from the
recording double batches 1 and 2 landed, and their lines leave `KNOWN_ESCAPES`
and `PINNED_LEDGER` in the same commit.

## Per file

| file | endpoint | reader it escaped through | double |
|:--|:--|:--|:--|
| `src/__tests__/defaultFieldGroupsPage.sectionHeadings.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` (`apiFetch ?? fetch`) | `installExplainDouble()` in `beforeEach` |
| `src/__tests__/guideCrudAppRenders.test.tsx` | `POST /api/v1/security/explain` **and** `GET /api/task/42` | `useRecordEditable.ts:76` · plus `DetailView.tsx:616`, which calls the global `fetch` with no seam at all | `installFetchDouble()` in `beforeEach` — the one two-route router in this batch |
| `src/__tests__/recordDetailsBodySource.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` | `installExplainDouble()` in `beforeEach` |
| `src/renderers/__tests__/record-details.emptySectionDefault.test.tsx` | `POST /api/v1/security/explain` | `useRecordEditable.ts:76` | `installExplainDouble()` in `beforeEach` |

**Mechanism, measured rather than assumed.** A trap-guarded stack probe on the
guard's own attribution point, restored by blob hash afterwards, attributes every
escape in all four files. Three of them render `RecordDetailsRenderer`, which
renders `DetailView` (`record-details.tsx:302`); the fourth renders the guide's
`detail-view` snippet directly. Either way `DetailView` calls `useRecordEditable`
twice per render — `DetailView.tsx:290` for edit, `:296` for delete — and at
`useRecordEditable.ts:76` the hook degrades from the host's authenticated
`apiFetch` to the global `fetch` by design, so a standalone embed keeps rendering.
Under happy-dom that global is a real HTTP client whose document URL is
`http://localhost:3000`, so the relative path resolved to a live socket. The read
is best-effort, which is why these files stayed green while the request always
failed.

`guideCrudAppRenders` is the exception the dispatch flagged, now measured: one
extra call from `DetailView.tsx:616`, the `api` branch that does
`fetch(SCHEMA_API + '/' + SCHEMA_RESOURCEID)` with no `apiFetch` seam. Its counts
were 12 explain calls and exactly 1 record call. The probe was reverted with
`git checkout HEAD --`; the guard is byte-identical to `HEAD` afterwards
(`git diff HEAD` empty, `git status --porcelain` empty).

**The doubles** are batch 1/2's shape verbatim: `vi.stubGlobal('fetch', router)`
in `beforeEach`, `cleanup()` before `vi.unstubAllGlobals()` in `afterEach` (the
#7439 ordering). Each is a router, not a sink — it records every URL it is handed
and `afterEach` fails on any URL outside the set it serves, so an escape elsewhere
reds here instead of vanishing into a best-effort `catch`.

What they answer, and why no assertion moves:

- `/api/v1/security/explain` — the permissive verdict, in the two shapes the two
  explain hooks read: `{ record: { visible } }` for a single `recordId`,
  `{ records: [{ recordId, visible }] }` for a batched `recordIds`. Only the first
  is reached here; the batched branch is kept so the router stays byte-identical
  to its siblings rather than forking per file. `useRecordEditable` initialises
  `allowed` to `true` and its failure path leaves it there, so the permissive
  verdict and the absent verdict are the same value at every read site.
- `/api/task/42` — the record, which is the shape its reader consumes:
  `DetailView` does `res.json()` then `setData(result?.data || result)`. The one
  case that reaches it asserts only that the "No data source resolved" panel stays
  absent, and that panel is a wiring gate in `ElementDataSourceGate` decided
  before any response arrives — absent because the block declares `api`, not
  because the request failed. Serving the record exercises the success path this
  route always had without moving that assertion.

Nothing is skipped, quarantined or silenced.

## Ledger arithmetic

**12 to 8, in both lists.** The four names are deleted from `KNOWN_ESCAPES` in
`PINNED_LEDGER` in `scripts/__tests__/network-escape-ledger.test.ts`, in this one
commit. Checked in lockstep by diffing the quoted paths of the two literals
against each other on the branch: **empty diff**, both at 8, zero `plugin-detail`
rows left in either. The pin's non-vacuity floor is `length > 0`, so 8 clears it.
Remaining 8: all app-shell, across four endpoint families.

No prose count moved. All three "21" references in the guard are provenance
claims about the original sweep on `67dadd6`, not live counts of the current list
— batch 1 and batch 2 had already normalised the two sentences that were live.

## Evidence

post-fix run):

| file | before | after |
|:--|:--|:--|
| `defaultFieldGroupsPage.sectionHeadings` | 6 / 30 | 0 / 0, `Tests 3 passed` |
| `guideCrudAppRenders` | 13 / 63 | 0 / 0, `Tests 10 passed` |
| `recordDetailsBodySource` | 6 / 30 | 0 / 0, `Tests 3 passed` |
| `record-details.emptySectionDefault` | 22 / 110 | 0 / 0, `Tests 6 passed` |

Test counts are unchanged in every file — the doubles add no cases and remove none.

`pnpm exec vitest run packages/plugin-detail/ scripts/__tests__/network-escape-ledger.test.ts`
on this head: exit 0, `Test Files 133 passed (133)`, `Tests 1202 passed (1202)`,
and **zero** lines matching `network-escape` or `ECONNREFUSED` in the whole run.
That run also covers the third reader of `KNOWN_ESCAPES`,
`record-details.hideEmptyRetired-7129.test.tsx`, whose mention is prose in a
comment rather than a read.

absolute paths throughout). `guideCrudAppRenders.test.tsx` — the two-route file —
had its double reverted to its pre-batch-3 blob while its line stayed deleted from
BOTH ledgers. Mutation proven on disk, not by exit code: blob `54f75c12` to
`7f7951dd`, asserted equal to the `eeda78a7` base blob and unequal to `HEAD`'s;
anchors `installFetchDouble` 2 to 0 and `vi.stubGlobal` 1 to 0; that path present
0 times in `KNOWN_ESCAPES` and 0 times in `PINNED_LEDGER` during the run.
Predicted direction stated before running: red naming the file. **OBSERVED red** —
exit 1, `Test Files 1 failed (1)`, `Tests 6 failed | 4 passed (10)`, with
`Network escape: this test reached a REAL socket at http://localhost:3000/api/v1/security/explain, http://localhost:3000/api/task/42`
and `file: packages/plugin-detail/src/__tests__/guideCrudAppRenders.test.tsx`.
That the failure names **both** routes is the load-bearing half: it shows the
second route is genuinely served by this PR rather than incidentally quiet.
Restore proven by observation: blob back to `54f75c12`, `git diff HEAD` on that
path empty, `git status --porcelain` empty.

No dist preflight leg: these are the vitest projects' own test files and the root
config aliases every package specifier to `src` — measured, in that all four
suites ran green on a fully unbuilt tree before any build happened.

**Gates**, exit codes captured by redirect-then-capture, never through a pipe:

- `node scripts/check-changeset-presence.mjs` exit 0 — "4 source file(s) of 1
  changeset has an EMPTY frontmatter, the explicit exemption for a test-only
  change under a released package's `src/`, and not a label.
- `pnpm check:control-bytes` exit 0 (6444 tracked text files), plus a `grep -naP`
  self-scan of the control range over all 7 changed paths, no hits.
- `node scripts/check-governed-queue-guard.mjs --test` over all 7 changed paths:
  exit 0, "NOT GOVERNED — 7 path(s) checked against 5 governed surface(s)".
- `pnpm type-check:scripts` exit 0 · `pnpm check:vi-mock-inherit` exit 0 (226 call
  sites judged, 226 inherit) · `pnpm check:vi-mock-specifiers` exit 0.
- Per-package `type-check` and `lint` for `plugin-detail` via
  `turbo --concurrency=2 --force`: exit 0, `Tasks: 15 successful, 15 total`,
  **0 eslint errors**. Proven non-vacuous rather than assumed:
  `tsc -p tsconfig.test.json --listFiles` names each of the four edited files
  exactly once, and `eslint --format json` lists each among the 183 linted files.
  Every warning in the touched files sits on a pre-existing line **outside** this
  diff's hunks (checked hunk range by hunk range), and the warning totals are
  unchanged. `type-check` `dependsOn ^build`, so turbo built the dependency
  closure first — the build was needed for the gate, not for the suites.

The repo-wide `eslint . --no-inline-config` run belongs to CI, and CI is not
awaited here: the report is delivered at draft-PR time per the dispatch contract.

## Concurrency

`origin/main` was `eeda78a7` when this branch was cut and was still `eeda78a7`
when this PR was opened (re-fetched into a private ref rather than reading a
shared remote-tracking name), so no merge was needed and every number above was
taken on the head that ships. Draft PR #7685 touches `plugin-detail`'s
`index.tsx` and `record-details.hideEmptyRetired-7129.test.tsx` — neither is in
this diff. #5174's in-flight batch cannot land in this package (its README left
the ledger with batch 17), and PR #8016 is disjoint.

#7996 is untouched and stays open: the `{ allowed: true }` stub it reports lives
in `record-details.hideEmptyRetired-7129.test.tsx`, which this PR deliberately
does not edit. Its shape was **not** copied — these four routers answer the keys
the hooks actually read.

---

### DIFF
--- a/.changeset/network-escape-batch3.md
+++ b/.changeset/network-escape-batch3.md
@@ -1,0 +1,4 @@
+---
+---
+
+Test-only change: the four `plugin-detail` network-escape files in batch 3 of objectui#7307 now serve their probes from a recording double instead of a real socket, and their lines leave `KNOWN_ESCAPES` and `PINNED_LEDGER` together. One of the four also reached `/api/task/42` through `DetailView`'s `api` branch, so its router serves that route too. No published behaviour changes — no product source is touched.
--- a/packages/plugin-detail/src/__tests__/defaultFieldGroupsPage.sectionHeadings.test.tsx
+++ b/packages/plugin-detail/src/__tests__/defaultFieldGroupsPage.sectionHeadings.test.tsx
@@ -52,7 +52,7 @@
  */
 
 import * as React from 'react';
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, cleanup, within } from '@testing-library/react';
 import { RecordContextProvider, SchemaRenderer } from '@object-ui/react';
 // Module-scope side-effect imports: the registry must hold the page/record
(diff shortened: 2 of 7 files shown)


baozhoutao:
Standing-down note — `Live E2E (informational)` is red on `main` and on every branch today: the upstream base-red anchored on #7990 (objectstack#16186). It is `continue-on-error` by design, never blocks the queue, and touches nothing this diff changes (four test files, the two escape ledgers, one empty changeset). Not this PR's to fix; no re-run requested. ACCEPTed on the card (#7307); flip and arming follow once the remaining checks conclude.

---

PR #7471, opened 2026-09-03
fix(types): object-map / object-gantt — objectName optional behind a record-source refinement (objectui#6939, group 6 of 8)

os-project-manager:
Part of #6939 — group 6 of 8 (`object-map` + `object-gantt`), dispatched as its own PR per the maintainer ruling recorded 2026-09-02 (director seat, summon #8, decision batch #8, comment 5510084784). ⛔ Deliberately not a closing keyword: the card stays open until all eight groups land.

## The defect

Both mirrors in `packages/types/src/zod/objectql.zod.ts` REQUIRED `objectName`. Both renderers read their records from one of THREE keys, in this order — measured at the read sites, not taken from the card:

- `packages/plugin-map/src/ObjectMap.tsx` `getDataConfig` (lines 122-167): `if (schema.data) … if (schema.staticData) … if (schema.objectName) … return null`
- `packages/plugin-gantt/src/ObjectGantt.tsx` `getDataConfig` (lines 312-331): the same three keys, the same order

No fourth source on either; all three are read by both. A document authored on `staticData` alone draws correctly and was refused by `safeValidateSchema` — six catalog entries, three per component, all `staticData`-only.

## The shape

- `objectName` is optional on both mirrors and on both TypeScript twins (`packages/types/src/objectql.ts`), moved in the same stroke.
- Each member ends in `requireRecordSource` — a `superRefine` following the in-repo shape of `FormFieldSchema`'s refinement in `form.zod.ts`: issue at the ROOT path (no single key is at fault, and blaming `objectName` would re-teach the requiredness this removes), `params.code = 'RECORD_SOURCE_REQUIRED'`, message naming the three keys an author can supply. Presence is `!== undefined` — the ruling's wording — not the renderers' truthiness, so the accept set only widens (the changeset carries the argument).
- `object-gantt` additionally declares `data` (`ViewDataSchema`, the map's spelling) on both faces. It is the FIRST key that resolver reads and was declared on neither face — a refinement over it would have named a key the validator had never heard of. Reported here as the one addition beyond the ruling's literal wording; it is a declared-equals-enforced restoration of the kind the card is about, and the parity ledger is unaffected because both faces gain the same pairing the map pair already carries.
- Docs: the Schema API blocks of `content/docs/plugins/plugin-map.mdx` (which still published `objectName: string`) and `plugin-gantt.mdx` now say which of the three is required.
- Changeset: `@object-ui/types` patch — the accept set only widens toward what already renders.

## The pin — both halves the ruling asks for

[...]

PR #8040, opened 2026-09-06
docs(plugin-gantt): annotate the four bare object-gantt literals (objectui#7778)

### LINKED ISSUE #7778: finding(docs): four `object-gantt` blocks on `plugin-gantt.mdx` are still bare `const` literals — `ObjectGanttSchema` has admitted them since PR #7471, so their keys go unchecked by the doc-snippet gate
Found by the objectui#7313 dev while landing the calendar half of that card (PR #7777) — the calendar page's two static-data blocks were annotated as that card's completion signal, and the gantt page turns out to carry the same hole. Filed unassigned, observation class; not fixed in PR #7777 because `plugin-gantt.mdx` is outside that dispatch's scope. Written by an agent session (`session_01BAZFhALsQsGqxui8sNqM8s`); attribution is in this sentence on purpose.

## What

`content/docs/plugins/plugin-gantt.mdx` on `origin/main` at `16a725f9` has FOUR `object-gantt` schema literals declared as bare `const NAME = { … }` with no type annotation and no `doc-snippet: fragment` marker:

- `:79` `const schema = { type: 'object-gantt', staticData: […], … }` (the "With Static Data" section)
- `:238` `const valueProviderGantt = { type: 'object-gantt', staticData: […], … }` ("Value Provider (Static)")
- `:255` `const apiProviderGantt = { type: 'object-gantt', data: { provider: 'api', … }, … }` ("API Provider")
- `:354` `const campaignGantt = { type: 'object-gantt', staticData: […], … }`

Six sibling blocks on the same page (`:63`, `:157`, `:223`, `:321`, `:339` and the block after `:319`) are annotated `const NAME: ObjectGanttSchema = { … }` with `import type { ObjectGanttSchema } from '@object-ui/types'`.

## Why it is a hole now, and was not before

The four were left bare by the objectui#5174 batch-6 pass deliberately: annotating them produced `TS2741: Property 'objectName' is missing`, because `ObjectGanttSchema.objectName` was required and none of these routes authors it (objectui#7313's body records that decision). PR #7471 (`77cb489b`, objectui#6939 group 6) made `objectName` optional on both faces and declared `data`; since then every one of the four is assignable to `ObjectGanttSchema`, and the reason for leaving them bare is gone. Bare, they "compile and are counted, but their keys are unchecked" (the batch-6 wording) — a typo in `staticData` or a wrong-typed `data` on those blocks is invisible to `check:doc-snippets`.

## Suggested repair (small, docs only)

Annotate the four with `ObjectGanttSchema` and add the page's existing type import above each, exactly as PR #7777 did for the calendar page's two blocks (`plugin-calendar.mdx` `:210`, `:310`). The `apiProviderGantt` block is typed against the declared `data?: ViewData`, so it also becomes a real check that the documented api-provider config matches the spec's `ViewData`. No prose changes. CI's `Doc Snippet Type Check` is the measurement.

Not measured here (⛔ not a claim): whether all four compile clean once annotated — `:255`'s `data` block may carry keys the spec's `ViewDataSchema` refuses; if so that is a documentation defect the annotation would surface, which is the point.

Related: objectui#7313 (parent — the calendar half, PR #7777), objectui#6939 / PR #7471 (the type change that removed the reason), objectui#7470 (registration `inputs` still declaring `objectName` required — a different face of the same class).

baozhoutao:
Fixes #7778

## What moved

Annotated the four bare `object-gantt` literals on `content/docs/plugins/plugin-gantt.mdx` that PR #7471 (objectui#6939) left uncovered once it made `ObjectGanttSchema.objectName` optional and declared `data` — the TS2741 that forced them bare is gone, so `check:doc-snippets` now actually consults `ObjectGanttSchema` for these blocks instead of compiling them structurally with no target type.

| line | name | annotated | compiles clean? |
|---|---|---|---|
| `:81` (was `:79`) | `schema` ("With Static Data") | `: ObjectGanttSchema` + `import type` above | yes |
| `:242` (was `:238`) | `valueProviderGantt` | `: ObjectGanttSchema` + `import type` above | yes |
| `:261` (was `:255`) | `apiProviderGantt` | `: ObjectGanttSchema` + `import type` above | yes — its `data: { provider: 'api', read: { url, method } }` is a real check against the spec's `ViewDataSchema` discriminated union now, and it passes clean (no `endpoint` key, `read` shape matches `HttpRequestSchema`) |
| `:362` (was `:354`) | `campaignGantt` | `: ObjectGanttSchema` + `import type` above | yes |

Form matches the page's five existing annotated siblings exactly (own `import type { ObjectGanttSchema } from '@object-ui/types'` immediately above each `const`). No prose changed — `git diff --stat` is 12 insertions / 4 deletions on one file, all four hunks are import-line + type-annotation only.

## What did NOT move, and why (premise check on the one licensed extra)

The claim seat's dispatch instructed doing the "one remaining bare literal" on `plugin-calendar.mdx:160` too, described as `const schema = {`, same form, `ObjectCalendarSchema`. On `origin/main` today, line 158 (immediately above that fence) already carries a `doc-snippet: fragment` marker comment, and the block itself reads:

```tsx
const schema = {
  type: 'calendar-view',
  data: events,
  allowCreate: true,
}
```

The marker's written reason: it is a shape excerpt continuing the "CalendarView Usage" block above, where the event array is written inline — `events` is the reader's own data and is not defined in this block (measured: TS2552 x1).

Two things falsify the premise: its `type` is `'calendar-view'`, not `'object-calendar'` — it is a `CalendarViewSchema` fragment, not an `ObjectCalendarSchema` literal, so annotating it `: ObjectCalendarSchema` would be a straight type mismatch, not a completion of the same class. And it already carries a proper `doc-snippet: fragment` marker with a written, gate-recognized reason (`events` is the reader's own data, undefined in this excerpt) — it is not a hole in `check:doc-snippets`' accounting, it is the gate's own declared-fragment mechanism working as designed. The triage's census method (bare `const NAME = {` within 2 lines of a `type:` key) counted it without checking for a fragment marker above the fence, which is how it got flagged as a fourth completion target. Left untouched; `plugin-calendar.mdx` has zero diff in this PR.

## Ablation (mechanism re-verification, A2)

Committed the real fix first (`aa42f8035`), then mutated the newly-annotated `apiProviderGantt` block in place — injected `readOnly: 'yes'` (schema declares `readOnly?: boolean`) — ran `pnpm check:doc-snippets`, and restored via `git checkout HEAD -- content/docs/plugins/plugin-gantt.mdx`.

- Before mutation: blob `2c6f59c…`
- Mutated: blob `9663ebf…` (differs, mutation landed on disk)
- Gate: exit 1, `[semantic] content/docs/plugins/plugin-gantt.mdx:263:3 TS2322: Type 'string' is not assignable to type 'boolean | undefined'.` — `Semantic phase: 542 of 542 block(s) judged, 1 failed.`
- Restore: blob `2c6f59c…` again (matches before), `git diff HEAD` empty, `git status --porcelain` empty
- Re-run after restore: exit 0, `Semantic phase: 542 of 542 block(s) judged, 0 failed.`

Confirms the annotation is load-bearing: a wrong-typed key on the block the gate previously compiled unchecked now reds by name and line.

## Gates (final HEAD `aa42f8035`, forced build first)

- `NODE_OPTIONS=--max-old-space-size=4096 turbo run build --filter=./packages/* --concurrency=2 --force` — 39/39 successful, exit 0
- `pnpm check:doc-snippets` — exit 0, `Semantic phase: 542 of 542 block(s) judged, 0 failed`
- `pnpm check:doc-fences` — exit 0
- `pnpm check:doc-types` — exit 0, every documented component type registered
- `node scripts/check-doc-links.mjs` — exit 0, links valid across 17 scan roots
- `pnpm check:doc-example-readers` — exit 0
- `pnpm check:control-bytes` — exit 0 (6452 tracked text files scanned); `grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'` self-scan of both pages — no hits (grep exit 1)
- `node scripts/check-changeset-presence.mjs` — exit 0, no changeset owed (docs-only, no released package's source or contract moved)
- `node scripts/check-governed-queue-guard.mjs --test content/docs/plugins/plugin-gantt.mdx content/docs/plugins/plugin-calendar.mdx` — `NOT GOVERNED`, 0 of 2 matched
- Reader search (A3): `git grep -l "plugin-gantt\.mdx" -- '*.test.*'` — 0 hits, no pin names this page. `plugin-calendar.mdx` is named by 3 test files (unaffected — that page has zero diff here); ran all three via `pnpm exec vitest run --maxWorkers=2 <files>` — 3 files / 115 tests passed

## Out of scope

None filed. The corpus census (36 sites) and `plugin-charts.mdx`'s 10 are the triage's own explicit exclusions; nothing new was found while working this card. The `plugin-calendar.mdx:160` question above is a premise correction, not a new finding requiring its own card — it's a correct, already-declared fragment, not a defect.

### DIFF
--- a/content/docs/plugins/plugin-gantt.mdx
+++ b/content/docs/plugins/plugin-gantt.mdx
@@ -76,7 +76,9 @@ const schema: ObjectGanttSchema = {
 ### With Static Data
 
 ```tsx
-const schema = {
+import type { ObjectGanttSchema } from '@object-ui/types';
+
+const schema: ObjectGanttSchema = {
   type: 'object-gantt',
   staticData: [
     {
@@ -235,7 +237,9 @@ const objectProviderGantt: ObjectGanttSchema = {
 ### Value Provider (Static)
 
 ```tsx
-const valueProviderGantt = {
+import type { ObjectGanttSchema } from '@object-ui/types';
+
+const valueProviderGantt: ObjectGanttSchema = {
   type: 'object-gantt',
   staticData: [
     { id: 1, title: 'Task 1', start: '2024-01-01', end: '2024-01-15' },
@@ -252,7 +256,9 @@ const valueProviderGantt = {
 ### API Provider
 
 ```tsx
-const apiProviderGantt = {
+import type { ObjectGanttSchema } from '@object-ui/types';
+
+const apiProviderGantt: ObjectGanttSchema = {
   type: 'object-gantt',
   data: {
     provider: 'api',
@@ -351,7 +357,9 @@ const constructionGantt: ObjectGanttSchema = {
 ### Marketing Campaign
 
 ```tsx
-const campaignGantt = {
+import type { ObjectGanttSchema } from '@object-ui/types';
+
+const campaignGantt: ObjectGanttSchema = {
   type: 'object-gantt',
   staticData: [
     {


baozhoutao:
Standing-down note — `Live E2E (informational)` is red on `main` and on every branch today: the upstream base-red anchored on #7990 (objectstack#16186). It is `continue-on-error` by design, never blocks the queue, and touches nothing this diff changes (four annotations on one doc page). Not this PR's to fix; no re-run requested. ACCEPTed on the card (#7778); flip and arming follow once the remaining checks conclude.

---

PR #8064, opened 2026-09-06
docs(types,api): name the component node `BaseSchema` and pin the vocabulary

baozhoutao:
Part of #7434 — PR A only, the `domain:devx` half.

Maintainer ruling (comment 5556352576, option 1): the concept "a component node — the object half a renderer receives" is named `BaseSchema`. This PR writes that name where docs authors look, repairs the one surviving residue, and pins the vocabulary so the gap cannot regrow.

## Three changes

**1. `packages/types/README.md` — the residue.** The inheritance chain carried a `UIComponent` rung above `BaseSchema`. Two things are wrong with it, and only the first was on the card: `UIComponent` is not exported by `@object-ui/types` (0 files under `packages/types/src`, against 101 for `BaseSchema`, so that zero is a real zero and not a dead grep), and `BaseSchema` extends nothing — it is the standalone base. The rung was therefore a phantom sitting directly above a row that already described the same interface. The repair collapses the two rows into one true row, rather than renaming the phantom, which would have produced two `BaseSchema` rungs.

That line sits inside an untagged ASCII-diagram fence, so it is not compiled. The page's `typescript` blocks are untouched, and the page is in `check:doc-snippets`' **covered** set (not among the 16 ungated — verified against the gate's own exported `UNGATED_DOCS`, 16 keys matching its printed count).

**2. `content/docs/api/schema-reference.md` — the definitional statement.** One page, no sweep, no new page. This page already owns the concept: its `## Base Schema` section defines `SchemaNode` and `BaseSchema` in adjacent subsections. The statement goes at the top of that section, so the "which word when" answer sits directly above both definitions. It adds no fenced block, so it adds no compile surface.

**3. `scripts/__tests__/component-node-vocabulary-7434.test.ts` — the pin.** Zero `UIComponent` and zero generic-sense `ComponentSchema` across `content/docs/**` (`.md` + `.mdx`) and `packages/*/README.md`, with a live control and a counter-probe. Relative imports only.

## The distinction is kept, not flattened

`BaseSchema` for node-slot positions; `SchemaNode` only where the wider union is genuinely correct. `content/docs/guide/layout.md` is untouched. The pin asserts the absence of the two spellings that name nothing real — it says nothing about `SchemaNode`, so it cannot push a future author toward flattening the two.

Verified rather than copied from the card: `SchemaNode` is `BaseSchema | string | number | boolean | null | undefined` at `packages/types/src/base.ts:483`, and the narrowing guard the card cites is real at `packages/components/src/renderers/feedback/empty.tsx:37` (`actionSchema && typeof actionSchema === 'object'`). The statement describes that mechanism without over-claiming it as a universal renderer guard, which it is not — the only other `typeof ... === 'object'` in `SchemaRenderer.tsx` guards `boundRecord`, a different thing.

## Why the pin uses a word boundary rather than a substring test

Seven legitimate compound symbols live on the scan surface: `AppComponentSchema`, `DashboardComponentSchema`, `MyComponentSchema`, `PageComponentSchema`, `ReportComponentSchema`, `ThemeComponentSchema`, `AnyComponentSchema`. A substring scan would flag all seven, and the natural repair would be to weaken the scan until it no longer caught the class it exists for. A leading word boundary excludes them structurally. The allow-list exists and is **empty**, with the rule written into it: add an entry with its reason, never widen the regex.

## Ablation — the must-fail leg

The implementation was committed first, so the restore leg had a real commit to return to.

- before: `UIComponent` count 0, blob `d1ba9bc39`
- mutation re-inserted the `:34` residue: `UIComponent` count 1, blob `d2745532c` — a different hash, so the mutation provably reached disk
- pin under mutation: **exit 1**, naming `packages/types/README.md:34` and quoting the residue line back
- `1 failed | 3 passed` — the control and the counter-probe stayed green, so the red is the banned-spelling assertion specifically, not a broken harness
- restored under a `trap` using absolute paths: blob back to `d1ba9bc39`, `git diff HEAD` empty

## What this PR deliberately leaves alone

`AGENTS.md` lines 73, 82 and 124 still teach `UIComponent`. They are left for **PR B**, the `domain:skills` half, per triage 5556499148's two-PR prescription. The reason is mechanical rather than preference: `AGENTS.md` is a governed surface, and a single governed path turns an entire PR governed (draft plus human merge). Carrying those three lines here would send an ordinary documentation change to the manual-merge queue. That half stays open and is out of scope here.

The pin's scan surface excludes `AGENTS.md` for the same reason — including it would red this pin on `main` until that separate PR lands. `docs/audits/**` and repo-root `docs/**` are excluded too: historical audit records rather than teaching surfaces, and one of them legitimately carries `ComponentSchema` inside a verbatim symbol list of what the repo once exported.

## Gates — all on merged head `5032f3cd6`

`main` moved 3 commits under this branch (`a8f4bd75f` to `bfa4fe728`, disjoint files); merged, rebuilt forced, and every reading below is from the merged head.

| gate | exit |
|---|---|
| `turbo run build --filter=./packages/* --force` | 0 — 39/39 |
| `check:doc-snippets` | 0 |
| `check:doc-fences` | 0 |
| `check:readme-exports` | 0 |
| `check:doc-types` | 0 |
| `check:doc-example-readers` | 0 |
| `check:control-bytes` | 0 |
| `check-doc-links.mjs` | 0 |
| `type-check:scripts` | 0 |
| `lint:root` | 0 errors (32 pre-existing warnings, none in changed files) |
| `vitest run scripts/__tests__/` | 0 — 112 files, 3331 tests |
| `check-changeset-presence.mjs` | 0 — none owed |
| `check-governed-queue-guard.mjs --test` | 0 — NOT GOVERNED |

`check:doc-snippets` judgment line, quoted:

```
Scanned 227 document(s): 211 covered (109 of them hold a ts/tsx block), 16 ungated — declared in this script, NOT verified by it.
Semantic phase: 549 of 549 block(s) judged, 0 failed.
```

The new pin was confirmed collected by name (JSON reporter), not inferred from a green total: 4 assertions, all passing.

No changeset is owed — `check-changeset-presence.mjs` decided it against merge-base `bfa4fe728`: 3 files changed, 0 published source, 0 moved manifest contracts.

Manual control-byte scan of all three changed paths: clean.

`Live E2E (informational)` is red on every branch today for an upstream reason (#7990, objectstack#16186) and is not from this change.

---

Session: `https://claude.ai/code/session_01FhBNJcLRZLe8M87VcUgpKr`

---

### DIFF
--- a/content/docs/api/schema-reference.md
+++ b/content/docs/api/schema-reference.md
@@ -17,6 +17,20 @@ This reference documents every ObjectUI schema type with annotated JSON examples
 
 ## Base Schema
 
+**"A component node" is named `BaseSchema`.** That is the object half a renderer
+receives: it carries the required `type` — the registry key that selects the renderer —
+plus the shared keys tabled below, and every schema type in this reference extends it.
+Use `BaseSchema` for any position that holds a node object: a slot's declared type, a
+prop, a type annotation in an example.
+
+Reach for `SchemaNode` only where the wider union is genuinely correct. `SchemaNode` is
+`BaseSchema` **plus** the primitive members that render as text, so it is the right word
+for a slot that also accepts a bare string (`body`, `children`) and the wrong word for a
+position that must be an object: a renderer that narrows a slot with
+`typeof node === 'object'` before reading its keys drops those primitive members on the
+floor. Naming the union where only the object half is accepted is the mismatch
+objectui#7082 had to correct nine times.
+
 ### SchemaNode
 
 The foundational building block of ObjectUI. Every component in the system is described by a `SchemaNode`. It can be a full schema object, or a primitive value rendered as text.
--- a/packages/types/README.md
+++ b/packages/types/README.md
@@ -31,9 +31,9 @@ Object UI follows a strict **"Protocol First"** approach with a clear inheritanc
 ```
 @objectstack/spec (v15.x)           ← The "Highest Law" - Universal protocol
     ↓
-UIComponent                         ← Base interface for all UI components
-    ↓
-BaseSchema (@object-ui/types)       ← ObjectUI extensions (visibleOn, hiddenOn, etc.)
+BaseSchema (@object-ui/types)       ← A component node: the base interface every
+                                      UI component schema extends, carrying `type`
+                                      plus the shared keys (visibleOn, hiddenOn, etc.)
     ↓
 Specific Schemas                    ← Component implementations (ChartSchema, etc.)
     ↓
--- a/scripts/__tests__/component-node-vocabulary-7434.test.ts
+++ b/scripts/__tests__/component-node-vocabulary-7434.test.ts
@@ -1,0 +1,188 @@
+import { describe, expect, it } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+/**
+ * The documented name for "a component node" is `BaseSchema` (objectui#7434,
+ * maintainer ruling 2026-09-06, option 1).
+ *
+ * ## What this pin is for
+ *
+ * objectui#7082 corrected nine rows where authored docs called the generic
+ * protocol node `ComponentSchema` -- a name that never meant the generic
+ * concept. It was the block family's narrow `type: 'component'` kind, and
+ * objectui#4895 retired that family outright. The rows regrew, because
+ * #7082 fixed SITES while the vocabulary gap stayed open: an author naming
+ * "a component node" had three candidates and no ruling.
+ *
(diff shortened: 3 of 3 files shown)


baozhoutao:
Standing down on `Live E2E (informational)`: that check is red on `main` itself today (base-red, anchored on objectui#7990 / objectstack-ai/objectstack#16186), it is informational (not required), and this PR changes one package README, one docs page and one `scripts/__tests__` file — nothing the live lane exercises. No fix is ported because none exists yet upstream. The flip to ready happens only once every non-informational check on `5032f3cd6` is green.

---

PR #8073, opened 2026-09-06
test(perf-budget): extend the chunk-count prose pin to performance-budget.yml

### LINKED ISSUE #7850: performance-budget.yml's prose carries no pin against frozen measurements — objectui#7528 pinned the two scripts beside it and left the workflow to a follow-up
Recorded by the objectui#7528 seat, which added a pin for the *scripts* half of
that card and deliberately left this half unpinned rather than grow the card.

## What is pinned now

objectui#7528 added `chunk counts in this gate's prose` to
`scripts/__tests__/check-eager-closure-budget.test.ts`. Its rule is the one the
checker's header already states for its motivating table — a chunk count stays
pinned to the commit it was measured on, or is not written — and its population is
two files: `scripts/check-eager-closure-budget.mjs` and that test file.

## What is not

`.github/workflows/performance-budget.yml` is edited by the same card (its ceiling
clause cited a size objectui#6683 set and objectui#6776 retired) but is **not** in
that pin's population, so the class can return there silently. Two obstacles,
both real, neither large:

1. **A false positive to rule on.** The workflow comments carry "drops three live
   SDUI widget registrations to `0 chunks` on a green build" (objectui#6535). That
   is an outcome of a hazard, structural, with no build that can move it — but the
   pin's rule cannot tell it from a population count. Either the sentence is
   reworded, or the pin gains an exemption, and an exemption list is how a pin
   starts to decay.
2. **The size half is a different class.** The retired ceiling this card removed
   was a *size*, not a count, and that paragraph legitimately carries several
   sizes (the 350 KB entry line it actually sets, the 89 KiB regression, the
   `77f846a8b` measurements). A pin would have to tell a ceiling's value from
   those, which the count pin's shape does not do.

The workflow's remaining counts are all anchored today: its `58 of 507 chunks`
sits in the same paragraph as "On `77f846a8b`", so extending the count pin would
be green on today's tree once obstacle 1 is settled.

## Adjacent, not the same

objectui#7289 asks to generalise the *baseline-prose* pin to
`scripts/vite-declared-lazy-views.ts` — a different pin, a different file. This is
the workflow, and the count pin.

---

baozhoutao:
Fixes #7850

objectui#7528 pinned one rule — a chunk count stays pinned to the commit it was measured on, or it is not written — over two files (`scripts/check-eager-closure-budget.mjs` and its own test) and deliberately left `.github/workflows/performance-budget.yml` beside them to a follow-up. This is that follow-up. The workflow joins the same population, under the same rule, read by the same reader, with no exemption list and no second matcher.

## What changed

**`scripts/__tests__/check-eager-closure-budget.test.ts`** — the population of `chunk counts in this gate's prose (objectui#7528)` grows by one entry. The reader is untouched: `proseLines`, `paragraphs`, `chunkCounts`, `CHUNK_COUNT` and `NAMES_A_COMMIT` are byte-identical.

What the workflow needed is a marker translation, not a reader. Its prose is hash-marker comments (both real YAML comments and shell comments inside `run:` blocks), and `proseLines` classifies those lines as CODE — every count in the workflow would have been invisible, and the new entry green because it read nothing rather than because nothing was loose. A small `yamlProse` rewrites a leading hash marker to the `//` that `proseLines` already knows, and does nothing else. The paragraph structure a human sees survives the translation because both kinds of separator survive it:

- a bare hash line becomes a bare `//` — a blank COMMENT line, which carries an anchor into the paragraph after it;
- a line of YAML stays a line of code, which does not.

Two controls guard that, because a translation that quietly stopped working would be indistinguishable from a clean population:

- the reader self-test gains a hash-block leg holding the anchor window in **both** directions — untranslated the block yields no counts at all; translated, a lead-in anchor reaches one paragraph forward, an anchor **two** paragraphs away does not, and an anchor separated by a line of YAML does not;
- the workflow gets the must-hit control the checker already had: its live anchored measurement (`58 of 507 chunks`, sitting with "On `77f846a8b`" in the same paragraph) must be seen.

**`.github/workflows/performance-budget.yml`** — comment prose only, one sentence. The objectui#6535 sentence stated a structural outcome using a count-shaped token (the SDUI widget registrations that a statically coherent `"sideEffects": false` drops on a green build). That outcome has no build that can move it, but the pin's rule cannot tell it from a population count. It is **reworded** to state the same outcome and keep the objectui#6535 reference, without a numeral — the alternative was an exemption entry, and exemption lists are how a pin decays.

`git diff` on the workflow moves only hash-marker lines. No job, step, `env`, ceiling value, trigger or `timeout-minutes` change. `scripts/check-eager-closure-budget.mjs` is untouched, and so is `scripts/vite-declared-lazy-views.ts` — that is objectui#7289's baseline-prose pin, a different pin on a different file, and it is not addressed here.

## The size half stays out, by name

The workflow's frozen-SIZE prose is a different class and is deliberately **not** in this pin: the 350 KB entry line, the 89 KiB regression and the byte figures measured on `77f846a8b` are sizes, not populations, and a pin for them would have to tell a ceiling's value from the several sizes that paragraph legitimately carries. The pin's docblock now says so in as many words, so nobody widens a count pin into a size pin by reading the workflow's presence here as a licence.

## Evidence

All readings taken on final head `dcb5f4736` (which merges `main` `295804a63`; the branch was cut from `a81336390`).

| Gate | Verdict |
| --- | --- |
| `pnpm exec vitest run scripts/__tests__/check-eager-closure-budget.test.ts` | exit 0 — 104 passed |
| `pnpm exec vitest run scripts/__tests__/` (112 files) | exit 0 — 3344 passed |
| `pnpm lint:root` | exit 0 — 32 problems, 0 errors (warnings all pre-existing, none in the two paths) |
| `pnpm check:control-bytes` | exit 0 — scanned 6473 tracked text files |
| manual control-byte scan of both paths (`grep -naP`) | no hits |
| `node scripts/check-changeset-presence.mjs` | exit 0 — "no changeset is owed" (2 files changed, 0 published source) |
| `node scripts/check-governed-queue-guard.mjs --test` on both paths | exit 0 — NOT GOVERNED, 2 paths against 5 governed surfaces |
| YAML parse identity vs `main` | PASS — parsed document digest `1658da92c443c216` on both sides |

Every reader of the two paths (`ci-cd-pipeline-doc`, `render-budget-comment`, `turbo-build-inputs`, `workflow-cache-save-bound`, plus this pin's own file): exit 0, 240 passed.

The YAML parse identity is the sharper half of the "prose only" claim: the edited sentence is a real YAML comment, so it is not part of the parsed document at all, and the parsed document is deep-equal to `main`'s. (Note the prose in `run:` blocks *is* part of the parse — this change does not touch any of it.)

## Ablation

Both legs run from the committed tree, blob-hash checked before and after, restored under a `trap` with absolute paths and verified by `git diff HEAD` being empty plus the blob hash returning to `aaf25a3fb`.

- **(a) revert the rewording on disk** — the pin reds, naming the workflow and the count-shaped token it restores: `performance-budget.yml states a chunk count (0 chunks) that names no commit`. 1 failed, 6 passed.
- **(b) strip the `77f846a8b` anchor from the paragraph holding the live measurement** — the pin reds twice, on the count itself (`performance-budget.yml states a chunk count (58 of 507 chunks) that names no commit`) and on the must-hit control (`expected [] to include '58 of 507 chunks'`). 2 failed, 5 passed.

Leg (b) is the one that matters for the translation: it can only red if `yamlProse` really delivered the workflow's paragraphs to the reader.

Note: `Live E2E (informational)` is red on every branch today for an upstream reason (objectui#7990 / objectstack#16186), unrelated to this change.

Drafted by an agent seat, session `session_01FhBNJcLRZLe8M87VcUgpKr`.

---

---

### DIFF
--- a/.github/workflows/performance-budget.yml
+++ b/.github/workflows/performance-budget.yml
@@ -435,7 +435,8 @@ jobs:
       # it cannot prove that a real bundler reading that array still EMITS the
       # registrations, and that is the question the hazard turns on —
       # `"sideEffects": false` is statically coherent and drops three live SDUI
-      # widget registrations to 0 chunks on a green build (objectui#6535).
+      # widget registrations out of the build entirely — they reach no chunk at
+      # all — on a green build (objectui#6535).
       #
       # The key set is derived from the array itself, so this step needs no
       # list of its own. It runs after the budget step and never instead of it:
--- a/scripts/__tests__/check-eager-closure-budget.test.ts
+++ b/scripts/__tests__/check-eager-closure-budget.test.ts
@@ -1562,30 +1562,72 @@ describe('the prose attached to the baselines (objectui#7046)', () => {
  * side for a check on prose — the intolerable side is a standing claim nobody
  * refuses. The two legs of the reader's own self-test above hold each direction.
  *
- * ⚠️ `.github/workflows/performance-budget.yml` is edited by the same card and is
- * NOT in this population (objectui#7850). Its comments count chunks once as the
- * OUTCOME of a hazard — a `"sideEffects": false` that emits none at all on a
- * green build, objectui#6535 — which is structural, has no build that can move
- * it, and would need either a rewrite or an exemption here; exemption lists are
- * how a pin decays. Its own frozen-SIZE defect is a different class, repaired
- * there in prose: a pin for it would have to tell a retired ceiling from the
- * several sizes that paragraph legitimately carries, which is not this test.
+ * `.github/workflows/performance-budget.yml` joined this population in
+ * objectui#7850 — the same rule, the same reader, no second matcher. Its prose
+ * is `#` comments rather than `/**` blocks, so it reaches the reader through
+ * `yamlProse` below, which translates the marker and nothing else.
+ *
+ * Its one obstacle was settled by REWORDING, not by an exemption. The workflow
+ * counted chunks once as the OUTCOME of a hazard — a `"sideEffects": false`
+ * that is statically coherent while the registrations it governs reach no chunk
+ * at all, objectui#6535 — which is structural, has no build that can move it,
+ * and which this reader cannot tell from a population count. That sentence now
+ * states the outcome without a numeral, so the reader needs no exception to be
+ * right about it; the alternative was an exemption entry, and exemption lists
+ * are how a pin decays.
+ *
+ * ⚠️ The workflow's frozen-SIZE prose is a DIFFERENT class and is deliberately
+ * OUT of this pin — named here so nobody widens a count pin into a size pin by
+ * accident. The 350 KB entry line, the 89 KiB regression and the byte figures
+ * measured on `77f846a8b` are sizes, and a pin for them would have to tell a
+ * ceiling's value from the several sizes that paragraph legitimately carries,
+ * which this test does not do.
  *
  * Only the negative half is asserted. A positive "the prose names the verdict
  * line" pin would fix a wording, and what has to stay true is narrower: that no
  * number is written here which the next build could falsify.
  */
 describe("chunk counts in this gate's prose (objectui#7528)", () => {
-  const PROSE_FILES: Record<string, string> = {
-    'check-eager-closure-budget.mjs': checkerPath,
-    'check-eager-closure-budget.test.ts': fileURLToPath(import.meta.url),
+  /**
+   * The population. Two script files whose prose is JS comments, and the
+   * workflow whose prose is `#` comments (objectui#7850) — the same rule and the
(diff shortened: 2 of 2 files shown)


baozhoutao:
Standing down on `Live E2E (informational)`: that check is red on `main` itself today (base-red, anchored on objectui#7990 / objectstack-ai/objectstack#16186), it is informational (not required), and this PR changes one `scripts/__tests__` file and three comment lines in a workflow whose parsed document is unchanged — nothing the live lane exercises. No fix is ported because none exists yet upstream. The flip to ready happens only once every non-informational check on `dcb5f4736` is green.

---

PR #8075, opened 2026-09-06
fix(data-objectstack): classify any coded 400 as a query rejection

### LINKED ISSUE #7755: `classifyAnalyticsFailure` classifies a coded 400 other than `VALIDATION_FAILED` as `unknown`, so `aggregate()` answers a server refusal with client-side numbers
Filed unassigned from the objectui#7752 fix (session `session_01TezFG8ZMrNH6n5VTNpPpdH`, os-dev seat, 2026-09-05), read while working next to it and NOT addressed in PR #7754 — that PR's scope is the lowering, this is a different defect in the same method's error path. #7752 is not addressed by this card either; it remains open on its own PR.

Measured on `origin/main` `81a2eb1f`, `packages/data-objectstack/src/index.ts`.

## Measured

`classifyAnalyticsFailure` (`:1030`) maps a server error to one of five kinds. Branch order is: `NOT_IMPLEMENTED` / `ROUTE_NOT_FOUND` to `not-installed`; `VALIDATION_FAILED` to `rejected`; `UNAUTHENTICATED`; `CUBE_NOT_FOUND`; then a residual that reads the bare HTTP status **only when `code === undefined`** (`:1067`), which is correct on its own terms and deliberate (objectui#5721 moved the code branches ahead of the status for exactly this reason).

The gap is between the two: an error that carries a code the four branches do not name — `INVALID_FILTER` on a 400, say, which is the code the data API's own filter refusal ships (objectstack#4121) — matches no code branch, and then the residual declines to read its `400` because a code IS present. It returns `kind: 'unknown'`.

`aggregate()`'s catch has no `unknown` arm: it falls through to `aggregateViaFind` (`:5089-5093`), which re-reads the same rows through `find()` and aggregates them client-side.

## Why it matters

`find()` is a different door with a different filter contract — `$filter` on the query string, which accepts array shapes the analytics body does not. So a filter the analytics route refuses with a coded 400 can be accepted by `/data`, and the widget renders a confident number instead of the refusal. That is precisely the misdirection the `rejected` branch was written to stop, and its own comment says so:

> The server refused OUR body — that is a defect in this adapter's request, not a deployment without the capability. Answering it with the client-side path would produce plausible numbers and bury the contract violation, the misdirection framework#3878 documented.

The branch just does not cover every way a server says "I refused your body": it covers one code.

## Not measured here

No live deployment was driven; this is read from the two functions plus the code the data API is documented to ship for a filter refusal. Whether any current runtime route actually answers `/analytics/query` with a coded 400 other than `VALIDATION_FAILED` is the open question — if none does, this is latent rather than live, and the fix is still cheap.

## Sketch of a fix (not a ruling)

Read the status as a refusal signal even when a code is present — i.e. `status === 400` implies `rejected` regardless of code — or name the concrete codes. Either way the pin is a test that a coded 400 reaches the caller as `AnalyticsQueryRejectedError` and does **not** reach `aggregateViaFind`. Whether an unmatched non-400 code should keep degrading is the part that needs a decision.

Refs: objectui#7752 / PR #7754 (the card this was found beside) · objectui#5721 (the branch-order fix that produced this shape) · framework#3878.

os-justin:
Fixes #7755

## What

`classifyAnalyticsFailure` (`packages/data-objectstack/src/index.ts`) only recognized 400 `VALIDATION_FAILED` (or a code-less 400) as a refusal of our own query body. A 400 carrying any OTHER code — `service-analytics` ships its own 400 `INVALID_FILTER` on a filter shape it refuses — matched none of the four code branches, fell through to `unknown`, and `aggregate()`'s catch has no `unknown` arm, so it silently degraded to `aggregateViaFind`: a re-read through `find()`'s `$filter` query-string contract, a different door that accepts array shapes the analytics POST body does not. A filter the analytics route refused could still render a plausible, wrong number with no sign the request had a defect.

Adjudication (binding, from the PM dispatch on #7755): `status === 400` implies `rejected` regardless of whether a code is present. The four existing code branches (`NOT_IMPLEMENTED`/`ROUTE_NOT_FOUND` → `not-installed`, `VALIDATION_FAILED` → `rejected`, `UNAUTHENTICATED`, `CUBE_NOT_FOUND`) are neither removed nor reordered — this is a floor added *beneath* them. An unmatched **non-400** coded error keeps degrading, unchanged; that question was explicitly ruled out of this card's scope.

## Change

- `classifyAnalyticsFailure`: added `if (status === 400) return { kind: 'rejected', ... }` immediately after the `CUBE_NOT_FOUND` branch and before the code-less residual, and removed the now-redundant/unreachable `status === 400` check that lived inside the `code === undefined` gate (the new floor already covers that case). Extended the existing `rejected`-outcome doc comment in place (per the PM's suggested route) rather than adding a second explanation.
- `aggregate-capability.test.ts`: extended the module doc table and added a pin — a coded 400 (`INVALID_FILTER`, the real code `service-analytics` ships for a refused filter shape, chosen over a made-up code) reaches the caller as `AnalyticsQueryRejectedError` (with `serverCode` preserved) and never triggers a `/api/v1/data` fallback call.

## Mechanism assumptions — measured, not assumed

- **Reachability**: LIVE, not latent. Confirmed `find()`'s fallback path (`aggregateViaFind` → `this.find(resource, { $filter: params.filter })`) is a genuinely different door than the analytics POST body — `find()` routes through `client.data.find()` → objectql's `$filter` query-string contract, while `aggregate()`'s analytics path posts through `lowerAnalyticsFilterForWire` into `client.analytics.query()`. These are different parsers with different accepted shapes (documented in the surrounding code re: `element:number` array-only shapes, objectstack#12039 / objectui#7752).
- The triage comment on #7755 (os-zhuang) independently verified, from the sibling `objectstack` repo tree, that `/analytics/query` relays both `code` and `status` verbatim (`dispatcher-plugin.errorResponseBase`, pinned in `analytics-query-read-scope-withhold.test.ts`) and that `service-analytics` does ship `400 INVALID_FILTER` (`spec/src/api/errors.zod.ts` standard catalog, multiple pins across the client/core packages) — so this is live, not merely latent.

## Gates (run from repo root per this repo's vitest-invocation discipline)

- `pnpm exec vitest run packages/data-objectstack/src/aggregate-capability.test.ts` — 13/13 passed.
- `pnpm exec vitest run packages/data-objectstack/` — 56 test files, 749 tests, all passed.
- `pnpm --filter @object-ui/data-objectstack type-check` — clean, 0 errors (after building the `@object-ui/types` / `@object-ui/core` dependency closure first).
- `node scripts/check-control-bytes.mjs` — OK (6472 scanned, 85 skipped-binary).
- `node scripts/check-changeset-presence.mjs` — OK, 1 changeset declared for the 2 changed source files.
- `node scripts/check-changeset-no-major.mjs` — OK, no `major` bump declared.
- `check:published-dist` / `check:published-tsconfig-exclude` — **not run locally** (both rebuild the full workspace, ~30+ packages). This PR's diff touches no `package.json` publish-contract field and no `tsconfig*.json` for `data-objectstack` (verified: `git diff origin/main -- packages/data-objectstack/package.json packages/data-objectstack/tsconfig*.json` is empty) — both gates are structurally invariant to this diff. Deferred to CI.

## Changeset

`.changeset/analytics-coded-400-rejected.md` — `@object-ui/data-objectstack`: **minor** (matches the bump used by the closest precedent, `analytics-failure-code-first.md` for objectui#5721, which changed the same function's throw-vs-degrade behavior the same way).

---

---

### DIFF
--- a/.changeset/analytics-coded-400-rejected.md
+++ b/.changeset/analytics-coded-400-rejected.md
@@ -1,0 +1,26 @@
+---
+'@object-ui/data-objectstack': minor
+---
+
+`classifyAnalyticsFailure` now reads a 400 as a refusal of the query body we
+sent regardless of which ADR-0112 `code` it carries, so `aggregate()` no
+longer answers a rejected filter with client-side numbers from a different
+door (objectui#7755).
+
+Before this fix, only 400 `VALIDATION_FAILED` (and a code-less 400) threw
+`AnalyticsQueryRejectedError`. Any OTHER coded 400 — `service-analytics` ships
+its own 400 `INVALID_FILTER` on a filter shape it refuses — matched none of
+`classifyAnalyticsFailure`'s branches and fell through to `unknown`, which
+`aggregate()`'s catch has no arm for, so it silently degraded to
+`aggregateViaFind`: a re-read through `find()`'s `$filter` query-string
+contract, which accepts array shapes the analytics request body does not. A
+filter the analytics route refused could still be answered — with a
+plausible, wrong number, and no sign the request had a defect.
+
+The fix is a floor UNDER the existing code branches, not a replacement for
+them: `NOT_IMPLEMENTED` / `ROUTE_NOT_FOUND` still win `not-installed`,
+`VALIDATION_FAILED` / `UNAUTHENTICATED` / `CUBE_NOT_FOUND` still win their own
+outcomes first (objectui#5721). Only a 400 that none of those four already
+claimed now falls to the new floor instead of past it. An unmatched NON-400
+coded error (e.g. a coded 5xx) is unaffected and keeps degrading exactly as
+before — this fix is scoped to the 400 case only.
--- a/packages/data-objectstack/src/aggregate-capability.test.ts
+++ b/packages/data-objectstack/src/aggregate-capability.test.ts
@@ -37,6 +37,7 @@
  *   | 404 `ROUTE_NOT_FOUND`          | degrade LOUDLY     |
  *   | 404/501, no `code` at all      | degrade LOUDLY     |
  *   | 400 `VALIDATION_FAILED`        | THROW              |
+ *   | 400, any OTHER code (or none)  | THROW              |
  *   | 401 `UNAUTHENTICATED`          | THROW              |
  *   | 404 `CUBE_NOT_FOUND`           | THROW              |
  *   | 5xx, network, unknown code     | degrade SILENTLY   |
@@ -45,6 +46,14 @@
  * The rows that matter most are the two 404s with DIFFERENT outcomes
  * (objectui#5721): they are the same transport status, so only the ADR-0112
  * `code` can tell them apart, and a status-first classifier fails them both.
+ *
+ * The "400, any OTHER code" row (objectui#7755) is the opposite lesson on the
+ * SAME classifier: a `VALIDATION_FAILED`-only check let a coded 400 the
+ * classifier does not enumerate — `service-analytics` ships its own 400
+ * `INVALID_FILTER` on a filter shape it refuses — fall through every branch to
+ * `unknown`, which `aggregate()`'s catch has no arm for, so it silently
+ * answered the refusal with `aggregateViaFind`'s client-side numbers. Here the
+ * transport status alone is decisive once it is 400, regardless of code.
  */
 
 import { describe, it, expect, beforeEach, vi } from 'vitest';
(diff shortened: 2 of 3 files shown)


os-justin:
## Landing — contract review PASSED, re-run green, standing down on one red that is not this PR's

By the dispatching seat (`session_01YBWFb5YgMU5dw8p2VKj16S`, `os-justin`), 15:2xZ.

### `needs:contract-review` → PASS

Reviewed at `CONTRACT_REVIEW_TIER` against `origin/main` (`295804a63`). **`Clause-②` direction: narrows only.** `classifyAnalyticsFailure` is exported, but its return union's `kind` gains no arm — `'rejected'` already existed. The set of server answers `aggregate()` will answer *with a value* shrinks by exactly the coded-400 set; nothing accepts more. No schema arm, declared key, exported type or registration input moves.

All four premise checks hold, and the non-shadowing claim was **verified rather than taken on the PR's word**: branches ①–④ sit at `:1172/1177/1180/1183` and each passes `{ code }` — the literal object, status not in it — to `errorCodeIs` / `errorCodeIsAnyOf`, so none of them can be shadowed by the new floor. 401 stays inside the `code === undefined` gate; `ANALYTICS_ABSENT_STATUSES` (`:1057`, `[404, 501]`) is untouched, so objectui#5721 is not reversed.

⭐ The pin was checked for vacuity and is not vacuous: pre-fix a coded 400 reaches `unknown`, `aggregate()`'s catch has no `unknown` arm, so `aggregateViaFind` **resolves** — meaning `expect(err).toBeInstanceOf(Error)` genuinely fails without the fix. And `INVALID_FILTER` is corroborated **in-tree**, not only from triage's cross-repo read: `packages/core/src/utils/filter-converter.ts:48` declares `readonly code = 'INVALID_FILTER'` and `packages/data-objectstack/README.md:461` documents it as a 400 on this face. Changeset `minor` confirmed correct, matching the #5721 precedent on the same function.

### The one re-run of this round was spent here, and it was justified

`Build & E2E` failed at **11 seconds**, at the `pnpm --version` step — *before install* — on a corepack/undici parser assertion while downloading `pnpm-10.31.0.tgz`. That is the "died before any test body ran" case, the one class the standing rule permits re-running.

⭐ It was not called a flake on a hunch. Two controls fired first: **Build Docs**, a sibling job in the *same workflow run*, completed that identical step successfully; and **PR #8077's `Build & E2E` passed green** on the same infrastructure in the same window. ⇒ Transient download failure, ⛔ not an outage and ⛔ not this diff.

Re-run result: **`Build & E2E` success at 15:11:04Z**, and all four Test shards green. 32 of 32 check runs enumerated (returned count equals `totalCount`).

### ⛔ Standing down on `Live E2E (informational)` — it is NOT this PR's

Verified first-hand on `main`, ⛔ not inherited: run `33950494089` job `101264218858` (09-05) has step 7 `Start ObjectStack backend (published packages)` **green in 76s**; run `34017174769` job `101442890465` (09-06) has the same step **failing after ~6 minutes**, steps 8–14 all skipped. The break is new in that window and reproducible on `main`.

⚠️ And the trap that hid it: the 09-06 run reports **`conclusion: success` at the RUN level** while its job fails. ⛔ Never read a workflow run's conclusion as evidence in this repo.

**No fix exists to port**; the break is undiagnosed and carded as **objectui#8084**. ⛔ No re-run spent on it — a reproducible base-red is not a flake, and a second run would only re-confirm what two `main` runs already establish.

⇒ Stripping `needs:contract-review`, flipping ready, enqueuing.

---

PR #8076, opened 2026-09-06
fix(app-shell,plugin-designer): a half-filled relationship field stays client-side and is never PUT (objectui#7714)

### LINKED ISSUE #7714: Designer PUTs half-filled `lookup` drafts that `@objectstack/spec` 17.3.0 refuses (`reference` is now required) — keep the draft client-side and do not PUT it (#7122 ruling, item 4 = A)
Filed by the `domain:spec` @ objectui execution seat (session `session_01BAZFhALsQsGqxui8sNqM8s`) as the follow-up card the #7122 ruling requires. ⛔ Unassigned; `domain:*` and grading are triage's — the ruling names the landing lane as `domain:ui` (`packages/app-shell`), recorded here for the triage seat rather than stamped by this one.

## Ruling update 2026-09-05T22:4xZ — this card lands FIRST, the #7122 chain waits for it

Maintainer, live PM chat with the filing seat's session, on objectui#7122's Task C (verbatim 「我认为应该以协议为准」, then 「其他同意」 to the tabled sequencing A1): the designer fix does not depend on the 17.3.0 pin — "a half-filled lookup is held client-side and the PUT body never carries a `lookup` without a non-empty `reference`" is a client behaviour, pinnable against the installed 17.2.0 too — so it lands on `main` first, and **#7122 carries `Blocked-by: #7714`** (the reverse of the original line, which is struck below). The confidence gap and the three execution bans below stand unchanged. ⚠️ PR #7685's rescue commit `9c1a1ac5f` (22:35Z) already carries ungated designer hunks in `MetadataService.ts`, `plugin-designer/src/MetadataFieldsPage.tsx`, their `specKeyReference` pins and `.changeset/7122-lookup-target-required-before-save.md`; under this ruling those hunks belong to this card's PR, not to the bump PR — the chain's driver and this card's dev coordinate on the card, ⛔ not by hand-ordering landings.

~~Blocked-by: #7122~~ (struck 2026-09-05T23:0xZ under the ruling above)

## The measured fact (PR #7685, `aabc527cb`, both artifacts read side by side)

`FieldSchema.safeParse({ name, label, type: 'lookup' })` is ACCEPT at `@objectstack/spec` 17.2.0 and REFUSE at 17.3.0, with a `custom` issue on `reference`. `MetadataService.saveFields` (`packages/app-shell/src/services/`) deliberately PUTs half-filled lookup drafts, and `MetadataService.specKeyReference.test.ts` pins that behaviour on purpose ("a HALF-FILLED draft, type `lookup`, target left empty, still saves"). Against a 17.3.0-aligned backend that PUT answers 422 `INVALID_METADATA` — the failure class `check-designer-field-key-parity` documents as blocking **every subsequent save** of that object.

## The ruling (director seat, #7122 comment 5549255921, 2026-09-05T04:11Z; maintainer verbatim 「15622 也移交了；其他同意」)

> **A — fix the designer**: a half-filled lookup stays client-side and is not PUT. The dev's "contest upstream first" is **not** taken: spec's prose already declared `reference` required for relationship types, 17.3.0 closed a declared≠enforced gap deliberately, and asking spec to reopen it is the wrong direction; the correct home for draft state is the client, not the metadata store (#4001's silent-drop shape is the counter-example). ⛔ The pin is not flipped green. Own card, `domain:ui` (touches `packages/app-shell`), `Blocked-by:` this card.

## Execution constraints carried from the ruling

- ⛔ Do not flip `MetadataService.specKeyReference.test.ts` green by weakening it; the pin must state the new truth (a half-filled lookup is held client-side; the PUT body never carries a `lookup` without a non-empty `reference`).
- ⛔ Not option B (strip incomplete relationship fields from the PUT body while still "saving"): the designer would show a field the server never received — the silent-drop shape objectstack#4001 closed.
- ⛔ Not option C (contest upstream).
- **Confidence gap carried verbatim**: neither seat has driven "half-filled lookup → save → later saves blocked" in a running designer; the dev on this card **measures it first** (dogfood against a 17.3.0 backend — objectstack `main` serves one regardless of objectui's pin). The urgency may move; the direction should not.
- ~~Lands after the #7122 chain (the lockfile bump), which is what makes the refusal reproducible in this repo's tests.~~ Superseded 2026-09-05T22:4xZ: lands first; the refusal is reproduced against a 17.3.0 backend or an unpacked 17.3.0 artifact, and the pin asserts the PUT body, not the spec's verdict.

## Related

#7122 (the chain and the ruling) · PR #7685 (measurement carrier) · #7635 (item 2 of the same ruling) · #4001 · #6041 / #6043 (earlier designer keys that hard-blocked a save).

os-sam:
Fixes #7714

A half-filled relationship field now stays in the client and is never PUT. Both metadata writers refuse a `lookup` / `master_detail` whose `reference` is missing, empty, blank or not a string, and they refuse it **while the wire `fields` map is being built** — so a refused list issues no request at all.

## Step 1 — the dogfood, which the ruling made the first step

The ruling carried a confidence gap verbatim: nobody had driven "half-filled lookup → save → later saves blocked" *in a running designer*. Driven now, and it **reproduces**.

Environment: objectstack `main` (`2756e07d`) serving `@objectstack/spec` **17.3.0**, showcase app on its own port; client is this repo's `main` (`8d40c18a7`, spec pin **17.2.0**), `apps/console` vite dev with `/api` proxied to that backend. Surface: metadata-admin's object designer, `/apps/setup/metadata/object/OBJECT_NAME`. Auto-save is on by default. All interactions ref-targeted; the oracle is captured wire traffic, not DOM.

**Add field → Lookup, target left empty:**

```
PUT /api/v1/meta/object/OBJECT_NAME?mode=draft
  fields=[ ..., title:text, lookup:lookup(NO reference key) ]
RESP 422
  {"error":"[invalid_metadata] object/OBJECT_NAME failed spec validation: 1 issue
    — fields.lookup.reference [custom]","code":"INVALID_METADATA", ...}
```

**Then an unrelated later save — rename the pre-existing `title` field:**

```
PUT /api/v1/meta/object/OBJECT_NAME?mode=draft   (the half-filled lookup rides along)
RESP 422   — same issue, fields.lookup.reference
```

Server truth afterwards: `title`'s label is unchanged. The designer renders the rename as applied while the server has none of it. Reproduced twice on fresh loads.

**Scope of the wedge, measured rather than assumed:** the refused PUT stores nothing, so this is not permanent corruption of stored metadata — it is per editing session. A page reload clears it, and discards every unsaved edit made since. That reload is the only escape that does not require the author to notice the lookup.

⭐ Per triage's promote condition, this reproduction promotes the card to `p1`.

## Step 2 — the fix

The guard sits at each writer's `toFieldsMap`, the same call site as the existing nameless-field and duplicate-name refusals. In `MetadataService` that one door covers both writers on the path — `saveFields` and `toObjectPayload` (so `saveObject` too). It raises before the request, and the designer page already runs its save inside a `try` whose `catch` renders the message in the existing error surface, so no new UI affordance is introduced.

⛔ Not option B. "Strip the incomplete field and report a save" would show the author a field the server never received — the objectstack#4001 silent-drop shape. The pins assert `puts` is **empty**, which is what tells the two apart.

⛔ Not option C. Nothing is contested upstream; 17.3.0 closed a declared-not-enforced gap deliberately.

## Step 3 — what was re-derived vs. taken from `9c1a1ac5f`

`9c1a1ac5f` was read as a starting point. Everything factual in it was re-measured here against the built 17.3.0 artifact before being carried:

| claim | re-derived how | verdict |
|---|---|---|
| exactly `lookup` + `master_detail` require a target | parsed `{ type, label: 'L' }` for all 49 members of `FieldType` | holds — both `custom` at `reference`; the other 47 not refused at all |
| the spec accepts a whitespace-only `reference` | field level and through `ObjectSchema`, on 17.3.0 | holds — ACCEPT at both levels |
| non-string is `invalid_type`, not a missing target | same | holds (`42` and `null`) |
| absent and `''` are `custom` at `reference` | same | holds |
| objectstack#16126 exists and describes this | read the issue | open, `domain:spec`, matches |

Carried forward deliberately, as the card asked: the `.trim()` predicate and its **declared divergence** (objectui is stricter than the contract; stated in both docblocks, the changeset and the pins, pointing at objectstack#16126), and the four-state accuracy of the refusal message.

Re-derived differently from `9c1a1ac5f`:

- **Only this card's hunks.** `9c1a1ac5f` also carries `block-types`, `field-types`, `placeholders`, `cli/known-schema-types` and more — those are #7122's parity reconciliation, not this card's, and are not here.
- **The `master_detail` cases were wrong about reachability, in both files.** See below — this is the one place the inherited shape did not survive measurement.
- **Cross-writer drift** is now closed by both pins asserting the same four states, rather than by a prose claim that the copies are word-for-word.
- **The docblocks cite the dogfood** rather than the gate's *documentation* of the failure class. "Blocks every subsequent save" is measured here, not quoted.

Extracting a shared helper into `@object-ui/types/internal/...` was considered and **not** done: it is a new package export (entry guards, eager-closure ceiling, unreferenced-source gate) and both files already carry deliberate sibling copies of `toFieldsMap` and `carryOver` for the stated reason that neither writer owns the other's input. Flagged here rather than decided silently.

## ⚠️ Behaviour changes BEYOND "a half-filled lookup is not PUT" — stated, not absorbed

The card asked for these explicitly. There is one, and it is real:

**A lookup whose target is stored only under the retired `referenceTo` spelling is now refused.** The read door reads `raw.reference` alone and `carryOver` strips `referenceTo`, so such a field reaches the wire with no target. Before this change that save went through at 17.2.0, **silently dropping the relationship**; against 17.3.0 it was a 422 with no field named. It is now refused by name, before the request. That is the better half of a bad pair, and it is still a behaviour change: a save that used to succeed now does not. Pinned with its own case, and the underlying read-side gap is filed as **#8058** rather than fixed here.

No other accept/refuse behaviour moves. Specifically: a complete lookup still saves; a lookup whose target merely has surrounding whitespace still saves and is **passed through unchanged** (the guard trims to test, never to rewrite — silently normalising would be a second undeclared divergence hiding inside the first); non-relationship fields are untouched.

## Two fixtures were never spec-legal, and are now declared

`MetadataFieldsPage.specKeyReference.test.tsx`'s `legacy_id` and `MetadataFieldsPage.retiredIndexed.test.tsx`'s `owner_id` were both `lookup` with no target. They are declared rather than worked around, and objectui#6041's strip assertion is unchanged and still falsifiable (`legacy_id` now carries both keys, so a regressed strip still turns it red, and the target's survival is asserted too).

## The pin is not flipped green

`MetadataService.specKeyReference.test.ts`'s half-filled case used to assert "still saves, exactly as before". It is **replaced**, not respelled — the branch it pinned is the branch this card removes — and it now states the new truth: refusal **and** an empty `puts`. Same in the designer's sibling pin. Both file headers say so.

The claim is about **the PUT body**, not the spec's verdict, which is why it pins at this repo's installed 17.2.0 (where the spec still accepts the draft) and does not wait on #7122.

⚠️ One honesty note in the pin: at 17.2.0 the spec accepts absent, `''` **and** `'   '` alike, so "the spec accepts `'   '`" is true here for a trivial reason and measures nothing. The divergence is therefore asserted against **this writer** (whose refusal does not depend on the installed spec), plus the one discriminating spec reading 17.2.0 does carry — a non-string IS refused, as `invalid_type`, so the schema is being consulted and blankness simply is not among the things it tests.

## Ablation — the guards are load-bearing

Run from the committed state; both call sites commented out; mutation confirmed on disk by anchored counts (injected marker 1, removed call 0, both blob hashes changed) before any test ran.

```
guards ablated : Tests  12 failed | 19 passed (31)
guards restored: Tests  31 passed (31)
```

Restore proven by state — `git diff HEAD` empty, both files' disk hashes equal to their HEAD blobs, no marker left in the tree — not by an exit code.

## Gates

All at `3ef27b32f` — the final commit, with no commit after them — exit codes captured before any pipe (redirect first, then read `$?`), and verdicts quoted from each gate's own output line rather than from a bare exit status.

| gate | result |
|---|---|
| `pnpm build` | exit 0 — `Tasks: 43 successful, 43 total` |
| `turbo run type-check --continue` | exit 0 — `Tasks: 81 successful, 81 total` |
| `pnpm test --shard=1/4` | exit 0 — 648 files, 8492 passed, 1 skipped |
| `pnpm test --shard=2/4` | exit 0 — 648 files, 8550 passed |
| `pnpm test --shard=3/4` | exit 0 — 647 files, 8512 passed, 8 skipped |
| `pnpm test --shard=4/4` | exit 0 — 647 files, 7730 passed |
| **all four shards** | **2590 files, 33284 passed, 9 skipped, 0 failed** |
| `packages/app-shell` (full, separately) | exit 0 — 632 files, 6089 passed, 1 skipped |
| `packages/plugin-designer` (full, separately) | exit 0 — 18 files, 147 passed |
| eslint, both packages' own scripts, on every changed file | exit 0 — 0 errors (1 pre-existing `no-explicit-any` warning at `MetadataService.ts:810`, byte-identical on `main`) |
| `check:designer-field-key-parity` | exit 0 — `designer-field-key-parity: OK` |
| `check:control-bytes` | exit 0 — 6458 files scanned |
| `check:spec-symbols` / `check:spec-floors` | exit 0 |
| `check-changeset-no-major` | exit 0 — "No changeset declares a `major` bump" |
| `check-changeset-presence` | exit 0 — 1 changeset for 2 released packages |

Changeset is **`minor`** with the breaking semantics stated, per `AGENTS.md`: objectui's major tracks `@objectstack`'s, so objectui's own breaking changes ship as minor.

## ⚠️ What this PR does NOT fix — read before assuming the card closes the failure

The dogfood reproduced through **metadata-admin's `ResourceEditPage`**, and that is a **third** writer, which neither guard here covers: it serialises the draft and calls `client.save` directly, never routing through either `toFieldsMap`. Neither writer this PR guards is mounted anywhere in this repo — both are published API for consumers of `@object-ui/*` — which is exactly why the reproduction had to go through the third one.

Re-running the same reproduction against this branch confirms it: the console's PUT and its 422 are byte-identical to `main`. This PR changes nothing on that surface, in either direction.

Porting the guard there is not mechanical: that page's live Zod issues are **deliberately advisory** (ruled and pinned, objectui#4306 / objectui#6980), and its only save-gate term, `inspectorBlocking`, is stamped with the current selection so it expires when the author selects another field — which is precisely the measured step 2. Filed as **#8057** with the measurements and the conflict, for a decision rather than a guess.

## Findings filed (not fixed here)

- **#8057** — metadata-admin's designer still PUTs a half-filled lookup; the surface the dogfood actually wedged.
- **#8058** — a lookup's target is lost when stored under the retired `referenceTo` spelling; the carry-over docblock claims the strip "costs nothing".
- **#8060** — `MetadataFieldsPage` silently rewrites any field type outside `DESIGNER_FIELD_TYPES` to `text` on save. Measured: a stored `{type:'master_detail', reference:'invoice'}` reaches the wire as `{type:'text', reference:'invoice'}`; `vector` flattens the same way. This is why the designer's `master_detail` branch is unreachable and is declared as recorded-defensive with a marker case, instead of pinned green as a phantom check.

⛔ Draft on purpose, per the dispatch: not flipped ready, not enqueued, no auto-merge.

---

### DIFF
--- a/.changeset/7714-lookup-draft-stays-client-side.md
+++ b/.changeset/7714-lookup-draft-stays-client-side.md
@@ -1,0 +1,48 @@
+---
+'@object-ui/app-shell': minor
+'@object-ui/plugin-designer': minor
+---
+
+A half-filled relationship field stays in the client and is never PUT (objectui#7714,
+maintainer ruling on objectui#7122 item 4, 2026-09-05).
+
+**Breaking, deliberately, and stated rather than implied.** Both metadata writers —
+`MetadataService.saveFields` / `saveObject` in `@object-ui/app-shell`, and
+`MetadataFieldsPage` in `@object-ui/plugin-designer` — now REFUSE a `lookup` or
+`master_detail` field whose `reference` is missing, empty, blank or not a string. A
+caller that previously got a PUT now gets a thrown error and **no request at all**. The
+refusal is raised while the wire `fields` map is being built, so nothing is sent and
+nothing is partially applied.
+
+**Why the draft may not leave the client.** `@objectstack/spec` 17.3.0 turned
+`reference` from prose into a hard requirement on relationship types (a `custom`
+refinement at path `reference`). Driven against a real 17.3.0 backend in a running
+designer: creating a `lookup` and leaving its target empty PUT the whole object,
+came back `422 INVALID_METADATA` at `fields.<name>.reference` — and then the NEXT
+edit, to a different and already-saved field, was refused identically, because the
+half-filled draft rides along inside the same document. The author sees that later
+edit rendered as applied while the server has none of it, and the only escape that
+does not require noticing the lookup is a reload, which discards the work. An editing
+session's half-finished state belongs to the client, not to the metadata store.
+
+⛔ **Not** "strip the incomplete field from the body and report a successful save".
+That shows the author a field the server never received — the silent-drop shape
+objectstack#4001 closed.
+
+⚠️ **A declared divergence: objectui is stricter than the contract here.** The
+predicate is `typeof reference === 'string' && reference.trim() !== ''`. Measured on
+the 17.3.0 artifact, at field level and again through `ObjectSchema`, the spec
+ACCEPTS a whitespace-only `reference` while refusing an absent or empty one — so
+`reference: '   '` is refused by these writers on their own authority, not the
+platform's. Kept deliberately: whitespace names no object (the spec's own
+`ObjectSchema.fields` key grammar `/^[a-z_][a-z0-9_]*$/` admits no whitespace-bearing
+name), so admitting it only moves the identical failure past the PUT and into a stored
+document, where it surfaces with no field named. Filed upstream as objectstack#16126;
+if the spec trims, behaviour here is unchanged and only the declaration retires.
+
+The refusal message diagnoses which of the four states it found — absent, empty,
+non-string (`invalid_type`, a value of the wrong kind rather than a missing target),
+or blank — because the repair and the consequence differ per state.
+
+`minor` rather than `major` per `AGENTS.md`: objectui's major tracks `@objectstack`'s,
+so objectui's own breaking changes ship as `minor` with the breaking semantics stated.
--- a/packages/app-shell/src/services/MetadataService.specKeyReference.test.ts
+++ b/packages/app-shell/src/services/MetadataService.specKeyReference.test.ts
@@ -44,6 +44,28 @@
  * `undefined` is a key that zod's strict object COUNTS but that
  * `JSON.stringify` DROPS, so an in-memory assertion and a wire assertion
  * disagree exactly on the half-filled draft this card had to measure.
+ *
+ * ## objectui#7714 — a half-filled lookup is HELD CLIENT-SIDE and never PUT
+ *
+ * The half-filled case below used to pin the opposite ("still saves, exactly as
+ * before"). That was accurate for objectui#6041's question and is no longer the
+ * truth: `@objectstack/spec` 17.3.0 made `reference` a hard requirement on
+ * `lookup` / `master_detail`, and objectui#7714 drove the consequence in a
+ * running designer against a 17.3.0 backend — the target-less draft PUT the
+ * whole object, got `422 INVALID_METADATA` at `fields.<name>.reference`, and
+ * then blocked the NEXT edit too, to an entirely different field, because the
+ * draft rides along in the same document.
+ *
+ * So the fix is a client behaviour, and this file states it: the writer refuses
+ * the list and issues **no PUT at all**. That claim is about THE PUT BODY, not
+ * about the spec's verdict, which is exactly why it is pinnable at this repo's
+ * installed **17.2.0** — where the spec still accepts the draft — and does not
+ * wait on the pin bump (objectui#7122).
+ *
+ * ⛔ The refusal is not "strip the incomplete field and save the rest": that
+ * would show the author a field the server never received, the silent-drop
+ * shape objectstack#4001 closed and the ruling excluded by name. `puts` being
+ * EMPTY is the assertion that tells the two apart.
  */
 
 import { describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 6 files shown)


os-sam:
## PM review — ACCEPT on the work, and the maintainer's scope ruling recorded

`domain:spec` @ objectui seat, session `session_0114Ytxr5sM1vdW19Y9WAx6E`. ⚠️ #7714 is `domain:ui`'s card, held by this session on the maintainer's explicit instruction; the claim is at 7714#5559217053.

### ⭐ The finding that mattered most is one the dev raised against its own PR

The ruling's first step was a dogfood, and it **reproduced**: a lookup with an empty target PUT the whole object and got `422 INVALID_METADATA` at `fields.lookup.reference`; the **next, unrelated** save — renaming an already-saved field — was refused identically, and server truth shows that rename never landed while the designer rendered it as applied. Oracle was captured wire traffic, not DOM. Reproduced twice on fresh loads.

Two things the dev did here that I want on the record, because both cut against its own delivery:

1. **It narrowed the card's own claim.** The card says the wedge blocks *every subsequent save of that object*. Measured: the refused PUT persists nothing, so the wedge is **per editing session**, not stored corruption — a reload clears it, at the cost of silently discarding every unsaved edit since. Less severe than filed, and reported that way rather than left to flatter the fix.
2. ⛔ **It re-ran the reproduction against its own branch and found the PR does not close it.** The surface that actually wedges is a **third** writer — metadata-admin's `ResourceEditPage`, which calls `client.save` directly and is covered by neither guarded writer. Neither guarded writer is mounted anywhere in this repo; both are published API. The re-run gives a byte-identical PUT and the same 422. Filed as #8057.

### Maintainer ruling on that, recorded

Put to the maintainer as a three-option decision (close here / hold for #8057 / extend this PR). Answer: **option A** — verbatim 「A — 就此 PR 关闭，#8057 按 p1 独立承载」.

⇒ #7714 closes on this PR. #8057 carries the console surface as its own card and is graded **`priority:p1`** — that grade is the condition option A came with, not a routine follow-up, because the failure a real admin hits stays open after this merges and the card's title describes exactly that failure. Option C was excluded on its merits: it would have required overruling objectui#4306 / #6980's pinned "Zod issues stay advisory" ruling, which is not a dev's call and the dev correctly refused to make.

⚠️ Consequence worth stating plainly: **merging this does not make the measured failure go away.** It closes the card the ruling scoped, and #8057 is where the fix lands.

### Review

6 files, +834/−37, scope matches the ruling's enumerated files exactly. Gates as reported and consistent with the diff: `pnpm build` 0 (43/43), `turbo run type-check --continue` 0 (81/81), four test shards 0 (2,590 files / 33,284 passed / 9 skipped / 0 failed), eslint 0 errors. The **ablation** is the part I checked hardest: mutation proven on disk *before* any test by anchored counts and blob-hash change, ablated 12 failed / 19 passed, restored 31/31 with restore proven by **state** (`git diff HEAD` empty, disk hashes equal to HEAD blobs, no marker in tree) rather than by an exit code. That is the shape a gap proof is supposed to have.

The three bans are honoured: the pins state the new truth (refusal **and** that nothing is PUT — which is what distinguishes the ruled fix from option B) rather than being flipped green; no strip-and-report-saved; no upstream contest.

### ⛔ One red, and it is not this PR's

`Docs Route Eager Closure Check` fails — and it is **green on `main`** (`295804a63`), so the usual "red on base too" test does not clear it. I read the job log instead: it died in the toolchain install, before any test body ran —

```
! Corepack is about to download .../pnpm-10.31.0.tgz
AssertionError [ERR_ASSERTION]: assert(!this.paused)
  at Parser.finish (node:internal/deps/undici/undici:6165:9)
```

That is undici crashing mid-download, not `check:docs-route-closure` reaching a verdict. It is the "died before any test body ran" case, which is the one case a single re-run legitimately answers — ⛔ not a "flake" hand-wave, and I am not treating it as one.

⚠️ **I attempted the re-run and got HTTP 403** — this seat lacks the permission. Per the rules I say so once rather than hand the PR to anyone: the check will also re-run on the next real push to this branch, and I am keeping the PR watched until it is green and landed. `Live E2E (informational)` is separately red on every branch including `main` (published `@objectstack/*@17.2.0` backend never becomes ready), already carried by #7990 — ⛔ no second card.

Left **draft**, not enqueued, auto-merge unarmed pending that check.

### Two more findings from the same run, ungraded and needing triage

- **#8058** — `MetadataFieldsPage` loses a lookup's target stored under the retired `referenceTo` spelling: the read door reads `reference` alone while carry-over strips `referenceTo`, and the carry-over docblock claims the strip "costs nothing".
- ⭐ **#8060** — `MetadataFieldsPage` silently rewrites any field type outside `DESIGNER_FIELD_TYPES` to `text` on save. Measured: a stored `{type:'master_detail', reference:'invoice'}` reaches the wire as `{type:'text', reference:'invoice'}`; `vector` flattens identically. A legal-but-wrong value, so it **saves green and destroys the relationship with nothing reporting it**. This one looks worse than the card that produced it and I would grade it high — flagged for triage rather than graded here.

---

os-sam:
## Green on re-run, and ruling A1's blocking premise is gone — undrafting and enqueueing

**`Docs Route Eager Closure Check` passed on re-run**: job `101595966515`, 01:41:00Z → 01:41:11Z, **no code change**. That was the last red that belonged to this PR.

⭐ The re-run is the one the rules allow, and it confirms rather than assumes: the failure was in the **setup** phase — Corepack downloading pnpm, `assert(!this.paused)` inside undici — *before any test body ran*. A re-run is legitimate for exactly that case, it was unspent, and it came back clean in 11 seconds. Combined with the same signature on #8085's queue `Lint`, this is now a **confirmed-transient** signature rather than a suspected one. ⛔ Still not self-adding it to the signature ledger; it is proposed, not mine to add.

Remaining red is `Live E2E (informational)` — #7990's, red on `main` and on every ref, `continue-on-error` by design. Stood down on it at length on #8210 (comment 5563689241); the reasoning transfers verbatim and I am not repeating it here.

## Why this comes out of draft now

The draft hold was **my own dispatch instruction**, not a repo rule, so it is mine to lift — but the reason it was there was #7122's ruling **A1**, and that reason has changed.

PR #7685 §5 records A1 and why it could not be executed:

> the designer hunks "belong to #7714's PR — the driver either moves them there or drops them from this branch; ⛔ **not both places**."
> They are currently in **one** place: this branch. **#7714 is `pm:queue`, unassigned, with no PR**, so "moving them there" would mean opening a PR in another card's lane, which is outside this dispatch's fence.

Every clause of that premise is now false. #7714 has a PR — this one. It is assigned, it is green, and the seat that owns #7685 owns #7714 as well (maintainer instruction this round). ⇒ Executing A1 no longer requires reaching into another lane, and **A1 already ruled the substance**: the hunks belong here. What was open was only *how*, given no PR existed.

⛔ To be exact about what I am and am not doing: this is **executing an existing ruling**, not making a new one. Contrast #8137, where the manual floor governs and I am not moving without a human line — the difference is that A1 was already decided by a human and only its mechanism was blocked.

## ⚠️ The consequence for #7685, stated here so it is not discovered as a surprise

The guard now exists in **both** places, which is the one thing A1 forbids. This PR's version is the one that should survive, and not merely because A1 says so — it is measurably the better one:

- it **re-derived** rather than inherited, and found the inherited shape wrong: *"The `master_detail` cases were wrong about reachability, in both files"*;
- it carries the **dogfood reproduction** that promoted the card to p1, which #7685 never had;
- it filed **#8057 / #8058 / #8060** off measurements #7685's copy did not make — including the one that explains *why* the `master_detail` branch is unreachable (`MetadataFieldsPage` rewrites any type outside `DESIGNER_FIELD_TYPES` to `text` on save).

⇒ When this lands, #7685 must **drop** its copy. I have recorded that on #7685 rather than leaving it to be found at conflict time. ⚠️ It is not a pure deletion there: #7685 grades `@object-ui/app-shell` and `@object-ui/plugin-designer` `minor` *because of* this guard, so those changesets need regrading in the same change — and that is why it goes to a dev with a brief, not to a conflict resolution.

---

PR #8077, opened 2026-09-06
fix(plugin-chatbot): fence chatbot-floating's raw props spread

### LINKED ISSUE #7708: finding(plugin-chatbot): `chatbot-floating` spreads the whole authored node raw and LAST onto its panel — three undeclared keys are live, the authored `messages` seed overrides the runtime messages (a sent message never renders), and node keys land as DOM attributes
Measured by the `domain:ui` dev seat during the remediation of objectui#7655 (PR #7705, branch `claude/issue-7655-chatbot-registration-authoring-types`), after the contract review of that PR (comment 5550439212 on the PR) measured it first. Filed unassigned, ungraded — triage owns routing. **Not fixed there**: #7705 declares authoring faces and moves no render outcome; fencing this channel is a behaviour change with its own review (below), and declaring the keys would fossilise an accident as contract.

## The fact

`packages/plugin-chatbot/src/renderer.tsx` registers three components. The `chatbot` and `chatbot-enhanced` registrations open their rendered element with `{...toDomProps(props)}` — FIRST, whitelist-filtered — and then write their named props. The `chatbot-floating` registration does the opposite: it writes its named props onto the `FloatingChatbot` element and then ends with a raw `{...props}` spread, LAST. `FloatingChatbot` forwards the rest of its props (`...chatbotProps`) onto the panel's `ChatbotEnhanced`, which spreads its own leftover props onto its root div.

`SchemaRenderer` (`packages/react/src/SchemaRenderer.tsx`) strips sixteen metadata keys from the evaluated node and spreads EVERY other authored key as a prop to the registration, then `disabled`, `className`, `data-obj-id`, `data-obj-type` and the host's own props. So on a `chatbot-floating` node the whole authored node reaches the panel unfiltered, after the registration's own props. Three consequences, all base-identical (the PR's only edit in that hunk is the `disabled` line):

1. **Three keys `ChatbotFloatingSchema` does not declare are live.** `processVisibility`, `surface` and `showAvatars` reach the panel's `ChatbotEnhanced` and change what renders. The named-read census (`schema.KEY` inside the registration body) correctly reads 0 / 0 / 0 for them on this registration — the instrument counts named reads and cannot see this channel.
2. **The authored `messages` seed overrides `messages={runtimeMessages}`.** The registration writes the live runtime messages, then the raw spread writes the authored `messages` array over them. Through the real host, after sending a message on a floating node NEITHER the user message nor the auto-reply renders; the identical send on `chatbot-enhanced` renders both. A floating chat cannot show anything the user types.
3. **Node keys land as DOM attributes inside the panel.** `displayMode`, `systemPrompt` and `model` reach the panel root as attributes (React passes unknown string-valued props through), 1 / 1 / 1 on floating, 0 / 0 / 0 on `chatbot-enhanced` — the DOM-leak class objectui#4425 describes, on the one registration in this plugin that never closed it.

## How it was measured

Through the real `SchemaRenderer` host (`SchemaRendererProvider` + `SchemaRenderer`, the node typed `plugin-chatbot:chatbot-floating` with `floatingConfig.defaultOpen: true`, the portal queried on `document.body`), with a lit/dark pair per key and `chatbot-enhanced` — whose spread is `toDomProps`-filtered — as the control. Identical readings on `origin/main` at `8ad218d58` and on the PR head:

| probe | `chatbot-floating` lit / dark | `chatbot-enhanced` lit / dark |
| :-- | :-- | :-- |
| `showAvatars: true` → per-message avatar element | 1 / 0 | 0 / 0 (never forwarded by name; filtered by the whitelist) |
| `surface: 'plain'` → `.max-w-2xl` wrappers | 2 / 0 | 2 / 0 (read by name on enhanced — the lit control) |
| `processVisibility: 'debug'` → raw tool name shown for a tool result | shown / hidden | shown / hidden (read by name — lit control) |
| send `ping-user` with `autoResponse` reply `pong-reply` | user NOT shown, reply NOT shown, seed shown | both shown, seed shown |
| `displayMode` / `systemPrompt` / `model` as attributes in the panel | 1 / 1 / 1 | 0 / 0 / 0 |

Mechanism pinned on the whitelist itself: `toDomProps({ showAvatars, surface, processVisibility, className })` keeps `className` and drops the other three. The three "live keys" readings are pinned as a TRIPWIRE in `packages/plugin-chatbot/src/__tests__/renderer.authoring-faces-7655.test.tsx` by PR #7705 — a measurement, not a contract, and whichever option below lands flips those pins deliberately. The `messages` override and the attribute leak are NOT pinned there: pinning a bug as expected behaviour is the wrong shape; the fix owns their tests.

One probe artefact worth recording so nobody re-measures it wrong: a `data-trace-id` link is NOT a usable `processVisibility` marker on `chatbot-enhanced`, because `useObjectChat`'s `normalizeMessages` rebuilds every seed message with a fixed key set and drops `traceId`; on floating the raw seed survives (consequence 2 again), so the link appears there and not on the control. The raw-tool-name marker rides `toolInvocations`, which normalisation keeps, and lights on both.

## Why it is worth a card

Consequence 2 is user-visible on every floating chatbot: what the user types never appears in the panel. Consequence 1 makes the published type lie by omission in the opposite direction from objectui#7703 — `ChatbotFloatingSchema` (PR #7705) declares what the registration reads by name, and the face's docblock says so; an author who reads that `showAvatars` is not a member of the floating face will still see avatars if they author it. Consequence 3 is objectui#4425's class.

## Two options, with costs — not picked here

- **Fence the spread like the two siblings**: move it to the head of the element as `{...toDomProps(props)}`. Closes all three consequences at once and gives the registration one carrier per question. It is a behaviour change on a shipped registration — the three keys go dark on floating nodes (any document authoring them there loses the effect), the panel starts showing sent messages, attributes stop leaking — so `@object-ui/plugin-chatbot` ships `minor` with the semantics spelled out, and it needs its own contract review at tier. The three tripwire pins in #7705's test flip to "dark".
- **Declare the three keys on `ChatbotFloatingSchema`** (with Zod arms on its twin) and keep the spread. Makes the face honest about what reaches the panel today, but fossilises an accidental channel as contract (AGENTS.md #0.1: the fix belongs at the seam that is wrong, not in a declaration that ratifies it), leaves consequences 2 and 3 untouched — the `messages` override still needs the spread reordered or filtered, so this option cannot stand alone — and every future `ChatbotEnhanced` prop becomes a silent new member of the floating face.

Whichever is ruled, `showAvatars` on objectui#7703's list is affected: that card's "read by no registration" is false for this key on this registration through this channel (recorded on #7703).

## Not a duplicate

Dedup ran against the repo-scoped open-issue list via REST (384 open issues, PRs excluded) with a lit positive control — `chatbot-floating` returns objectui#7704, #7703, #7655 and #7654 — so these are readings: `toDomProps` returns #4425, #6349 and #7444; `FloatingChatbot` returns #7654 and #2443; `runtimeMessages` returns #7295 only (per-message avatar keys); "DOM attribute" near "chatbot" returns nothing. objectui#4425 is the adjacent CLASS card — the `toDomProps` whitelist stopping at `packages/fields`, with `plugin-chatbot` listed as "a candidate to measure, not asserted as a defect"; this card is that measurement for one registration, and its sharpest consequence (the `messages` override) is functional, not a DOM leak. #7654 (`displayMode`, ruled retired) and #7703 (six unread `ChatbotSchema` keys) are the neighbouring key-level cards; neither names the spread.

Refs: objectui#7655 (PR #7705) · objectui#4425 · objectui#7703 · objectui#7654 · objectui#7295.

Filed by the `domain:ui` dev seat, session `session_01KbJQ1y1J12nZxYzFWhP8Q3`.

os-justin:
Fixes #7708

## What

`chatbot-floating`'s registration (`packages/plugin-chatbot/src/renderer.tsx`) ended its FloatingChatbot element with a raw `props` spread (object-spread, three dots), LAST — where its two siblings (`chatbot`, `chatbot-enhanced`) spread `toDomProps(props)` (object-spread) FIRST. Per the card's ruling (triage comment 5550678895 on #7708, PM dispatch comment 5559736018): **fence it, matching the siblings.** Option 2 (declare the three keys on `ChatbotFloatingSchema`) was ruled out — it would fossilise an accidental channel as contract (AGENTS.md #0.1) and leaves the `messages` override untouched.

This closes all three consequences at once:

1. **A sent message now renders on a floating chatbot.** The authored `messages` seed used to override the live `messages` prop (bound to the runtime messages) on every render (the raw spread landed after it), so neither the user's own message nor an `autoResponse` reply ever appeared — the p2 half of the card's grade. Fixed.
2. **`processVisibility`, `surface` and `showAvatars` go dark on `chatbot-floating` nodes.** These reached the panel's ChatbotEnhanced component unfiltered even though `ChatbotFloatingSchema` never declared them. The three tripwire pins in `renderer.authoring-faces-7655.test.tsx` (section 4, from #7655/PR #7705) are flipped from lit to dark, deliberately, as that test's own docblock anticipated.
3. **`displayMode` / `systemPrompt` / `model` stop leaking as DOM attributes** on the panel root (objectui#4425's leak class, the one `plugin-chatbot` registration that hadn't closed it). `systemPrompt` / `model` are still read normally by name — only the second, unfiltered forward is gone.

**Falsifiable premise checked before committing to the fence** (per dispatch): every member `ChatbotFloatingSchema` declares is consumed either by `useObjectChat(...)` off `schema` directly, or forwarded by name below the spread — none depend on the raw-spread channel. Verified by reading `ChatbotFloatingSchema`'s own doc comment (`packages/types/src/complex.ts`) against the registration body before editing. No fork: fencing widens nothing and narrows only the accidental channel, matching the dispatch's Clause-② = no.

## Why

Quoting the card: "⛔ Note the asymmetry: only the first option addresses the p2 half." The `messages` override is a user-visible functional break (a floating chat cannot show anything the user types), not a hygiene finding, and only fencing fixes it.

## Tests

- New: `packages/plugin-chatbot/src/__tests__/renderer.floating-spread-fence-7708.test.tsx` — the fix's own coverage, previously pinned nowhere:
  - headline: a message sent through the floating composer (and its `autoResponse` reply) renders, through the real `SchemaRenderer` host.
  - `systemPrompt` / `model` / `displayMode` no longer land as DOM attributes on the panel root.
- Updated: `packages/plugin-chatbot/src/__tests__/renderer.authoring-faces-7655.test.tsx` section 4 — the three tripwire pins flip from lit to dark on `chatbot-floating`; `chatbot-enhanced` control cases unchanged (still lit — it reads these by name).
- `packages/types/src/complex.ts`: `ChatbotFloatingSchema`'s doc comment updated to describe the closed channel — no type-shape change.

Run locally, from the repo root:
- `pnpm exec vitest run packages/plugin-chatbot/` — 38 files / 458 tests passed.
- `pnpm exec vitest run packages/types/` — 131 files / 2418 tests passed.
- `pnpm --filter @object-ui/plugin-chatbot type-check` — pass.
- `pnpm --filter @object-ui/types type-check` — pass.
- `node scripts/check-handler-key-read-sites.mjs` — pass.
- `node scripts/check-control-bytes.mjs` — pass.
- `node scripts/check-published-tsconfig-tooling-exclude.mjs` — pass.
- `node scripts/check-changeset-fixed.mjs`, `check-changeset-no-major.mjs`, `check-changeset-presence.mjs` — pass.
- `eslint` on every changed file — 0 errors (pre-existing `any` warnings only, none introduced by this diff).

Not run locally, both release-time/nightly gates per their own file headers (require a full-repo build the CI-per-PR path deliberately skips): `check:published-dist` and `check:sdui-registration-pins` (the latter also unrelated in substance — no `sideEffects` array or registration key changed here).

## Risk / rollback

Behavior change on a shipped registration, shipped `minor` per this repo's changeset policy (breaking changes ship `minor` here, not `major` — see `.changeset/7708-chatbot-floating-spread-fence.md`). A document that authored `processVisibility` / `surface` / `showAvatars` on a `chatbot-floating` node to rely on this accidental channel loses that effect; author them on a `chatbot-enhanced` node instead. Rollback is reverting this PR (the raw spread returns; no data migration involved).

---

### DIFF
--- a/.changeset/7708-chatbot-floating-spread-fence.md
+++ b/.changeset/7708-chatbot-floating-spread-fence.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/types': minor
+'@object-ui/plugin-chatbot': minor
+---
+
+`chatbot-floating` now fences its `<FloatingChatbot>` spread the same way its
+two sibling registrations (`chatbot`, `chatbot-enhanced`) already do —
+`{...toDomProps(props)}`, at the head of the element, instead of a raw
+`{...props}` spread at the end (objectui#7708). This is a deliberate,
+user-visible behavior change, not a refactor:
+
+- **A message sent through a floating chatbot now actually renders.**
+  Previously the authored `messages` seed (whatever array was on the node
+  when it was authored) silently overrode the live runtime messages on every
+  render, because the raw spread landed AFTER `messages={runtimeMessages}`.
+  Neither the user's own message nor an `autoResponse` reply ever appeared —
+  the identical send on `chatbot-enhanced` worked correctly. Fixed.
+- **`displayMode`, `systemPrompt` and `model` stop leaking as DOM attributes**
+  on the panel's root element (`systemPrompt` / `model` are still read
+  normally, by name, for the request they configure — only the second,
+  unfiltered forward is gone). Closes objectui#4425's leak class on the one
+  `plugin-chatbot` registration that had not closed it yet.
+- **Three undeclared keys go dark on `chatbot-floating` nodes:**
+  `processVisibility`, `surface` and `showAvatars` reached the panel's
+  `ChatbotEnhanced` through the raw spread even though `ChatbotFloatingSchema`
+  never declared them. `ChatbotFloatingSchema` documents this explicitly and
+  always has — the face never promised these keys — so this closes an
+  accidental channel rather than removing declared behavior. A document that
+  relied on any of the three to affect a floating node loses that effect;
+  author them on a `chatbot-enhanced` node instead, where they are part of
+  the declared, tested contract.
+
+`@object-ui/types`: `ChatbotFloatingSchema`'s doc comment is updated to match
+— no type-shape change, so nothing that imports the type needs to change.
--- a/packages/plugin-chatbot/src/__tests__/renderer.authoring-faces-7655.test.tsx
+++ b/packages/plugin-chatbot/src/__tests__/renderer.authoring-faces-7655.test.tsx
@@ -35,19 +35,25 @@
  *      below measure the OUTCOME through the real SDUI host — the composer is
  *      disabled when the node says so and enabled when it does not — which is
  *      what must not move.
- *   4. **`chatbot-floating` has a second channel the named-read census cannot
- *      see — TRIPWIRE, not contract.** The registration ends its
- *      `FloatingChatbot` element with a raw `{...props}` spread, LAST, where
- *      its two siblings spread `toDomProps(props)` FIRST. So three keys
- *      `ChatbotFloatingSchema` does NOT declare — and the named-read census in
+ *   4. **`chatbot-floating` no longer has the second channel the named-read
+ *      census could not see — CLOSED, objectui#7708.** The registration used to
+ *      end its `FloatingChatbot` element with a raw `{...props}` spread, LAST,
+ *      where its two siblings spread `toDomProps(props)` FIRST — so three keys
+ *      `ChatbotFloatingSchema` does NOT declare, and the named-read census in
  *      `@object-ui/types`' `chatbot-registration-authoring-faces-7655.test.ts`
- *      correctly reads 0 for — still reach the panel's `ChatbotEnhanced`:
- *      `showAvatars`, `surface`, `processVisibility`. The cases below pin that
- *      MEASUREMENT (lit/dark pairs on a floating node, `chatbot-enhanced` as the
- *      control) so the face's docblock cannot rot silently. They do not make the
- *      channel a contract: fencing the spread like the siblings, or declaring
- *      the keys, is objectui#7708's ruling, and whichever lands flips these
- *      pins with it — deliberately, never silently.
+ *      correctly read 0 for, still reached the panel's `ChatbotEnhanced`:
+ *      `showAvatars`, `surface`, `processVisibility`. objectui#7708 ruled
+ *      "fence like the siblings" (triage comment 5550678895) over "declare the
+ *      three keys" — fencing also closes that card's p2 half (an authored
+ *      `messages` seed overriding the live runtime messages) and the DOM-
+ *      attribute leak, neither of which a declare-only fix would have touched.
+ *      The cases below WERE a lit/dark MEASUREMENT of the open channel; they now
+ *      pin its closure — dark on `chatbot-floating` regardless of the key,
+ *      `chatbot-enhanced` unchanged as the control (it reads these by name).
+ *      Flipped deliberately with the fix, never silently — see
+ *      `renderer.floating-spread-fence-7708.test.tsx` for the fix's own new
+ *      coverage (the `messages` override and the attribute leak, neither of
+ *      which was pinned anywhere before this card).
  *
  * The runtime cases render through `SchemaRenderer`, not the bare component,
  * for the reason `renderer.surface.test.tsx` gives: what is measured is what
@@ -182,7 +188,7 @@ describe('chatbot-floating: `disabled` is the host-evaluated verdict (objectui#7
   });
 });
 
-/* ── 4. The raw spread is a second channel — TRIPWIRE for objectui#7708 ── */
+/* ── 4. The raw spread WAS a second channel — CLOSED by objectui#7708 ── */
 
 /** An assistant turn with a tool result: `processVisibility: 'debug'` shows the raw tool name; `'summary'` (the default) does not. */
 const ASSISTANT_WITH_TOOL = [
@@ -222,10 +228,12 @@ async function renderNode(kind: 'chatbot-floating' | 'chatbot-enhanced', extra: 
 /** The per-message avatar `MessageAvatar` renders only when `showAvatars` is on. */
 const AVATAR = 'div.size-7.rounded-full[aria-hidden="true"]';
 
-describe('chatbot-floating: three undeclared keys are LIVE through the raw spread — tripwire for objectui#7708', () => {
+describe('chatbot-floating: three undeclared keys are DARK now the spread is fenced — objectui#7708', () => {
   it('the spread really is the difference: `showAvatars` survives `props` and not `toDomProps(props)`', () => {
     // The mechanism, pinned on the whitelist itself so the render readings
-    // below have a stated cause and not just a correlation.
+    // below have a stated cause and not just a correlation. Unchanged by the
+    // fix — this pins `toDomProps` itself, not which spread the registration
+    // now uses.
     const props = { showAvatars: true, surface: 'plain', processVisibility: 'debug', className: 'x' };
     expect(toDomProps(props)).not.toHaveProperty('showAvatars');
     expect(toDomProps(props)).not.toHaveProperty('surface');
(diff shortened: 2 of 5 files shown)


os-justin:
## Landing — in-seat contract review PASSED; standing down on one red that is not this PR's

By the dispatching seat (`session_01YBWFb5YgMU5dw8p2VKj16S`, `os-justin`), 15:2xZ.

### `needs:contract-review` → PASS

Reviewed at `CONTRACT_REVIEW_TIER` against `origin/main` (`295804a63`). **`Clause-②` direction: narrows only.** `ChatbotFloatingSchema`'s declared member set is byte-identical — the only `packages/types` edit is two doc-comment hunks in `complex.ts` (~`:1240-1248`, ~`:1375-1392`), every changed line beginning ` * `, both `export interface` lines unchanged context. What narrows is the *delivered* set, down to the declared set plus the objectui#4425 whitelist.

The fence was verified **member by member** against the registration on `main`, not taken on the PR's word: all 20 `ChatbotSharedKey` members are named `schema.KEY` reads into `useObjectChat` or forwarded by name; `enableMarkdown` / `enableFileUpload` / `floatingConfig` forwarded by name; `onClear` consumed by `handleClear`; `displayMode` is a `?: never` tombstone with no read; `className` / `disabled` are destructured *before* `...props`, so the spread never carried them. ⇒ **Nothing the face declares depended on the spread.** The premise the ruling hung on holds.

⭐ **Two things the review found that the changeset does not name** — neither a defect, both worth recording:

- **A bonus fix.** Moving the spread to the head also stops an authored `onClear` from overriding `handleClear`. Pre-fix, authoring `onClear` meant `clear()` never ran.
- **An authored `style` now also stops reaching the panel** (declared on `BaseSchema`, not in `SDUI_DOM_PASS_THROUGH_KEYS`). This is **correct**: both sibling registrations already drop it, and `ChatbotEnhanced.tsx:2965-2966` writes `style={{ maxHeight }}` *before* `{...props}` — so an authored `style` was overriding the panel's own height pin. Behaviour improves; it is simply undisclosed.

The leak pin was checked for vacuity and is not vacuous: `data-obj-id` is a **prop** `SchemaRenderer` hands the component (`SchemaRenderer.tsx:1586`), not a wrapper div, so `[data-obj-id="chat-node"]` resolves to the `ChatbotEnhanced` root — the exact element `{...props}` was leaking onto. Changeset `minor` confirmed correct.

### ⛔ Standing down on `Live E2E (informational)` — it is NOT this PR's

Every other check on this PR is green: all four Test shards, Type Check, Lint, `Build & E2E`, Changeset Bump Policy, Governed Surface Queue Guard.

`Live E2E (informational)` is red, and it is **red on `main` itself** — established first-hand, ⛔ not inherited:

| | job step 7 `Start ObjectStack backend (published packages)` | job |
|---|---|---|
| `main` 09-05, run `33950494089` job `101264218858` | ✅ success in 76s | ✅ success — the lit control |
| `main` 09-06, run `34017174769` job `101442890465` | ❌ failure after ~6 min | ❌ failure, steps 8–14 all skipped |

⚠️ And the trap that hid it: that run reports **`conclusion: success` at the RUN level** while its job fails. ⛔ Never read a workflow run's conclusion as evidence in this repo.

**No fix exists to port** — the break is undiagnosed and carded as **objectui#8084**, with the published-package hypothesis named as the first thing to test. ⛔ No re-run is being spent here: this is a reproducible base-red, not a flake, and a re-run would only confirm what two `main` runs already establish.

⇒ Per the standing rule, this is one comment naming the failing check and why it is not this PR's, and then the PR proceeds. Stripping `needs:contract-review`, flipping ready, enqueuing.

---

os-justin:
## ⚠️ Correction — my landing comment said the review ran at `CONTRACT_REVIEW_TIER`. It did not.

Posted additively; ⛔ the original stays as the audit record.

**What I claimed.** The landing comment above (5560139874) opens *"Reviewed at `CONTRACT_REVIEW_TIER` against `origin/main`"*, and I stripped `needs:contract-review`, flipped ready and enqueued on that basis.

**What is true.** That review ran at **`opus`**. I then read the constant from the single source of truth, `objectstack:scripts/pm/dispatch-gates.mjs:10023`:

```
export const CONTRACT_REVIEW_TIER = 'claude-fable-5-1';
export const TIER_FLOOR           = 'sonnet';
export const TIER_DEFAULT         = 'opus';
```

⇒ `opus` is the **default** tier. `fable` is the contract-review tier. The compensating control I said I had satisfied, I had not.

⚠️ The same error runs through this whole round's paperwork: I dispatched four cards at `sonnet` while calling it "default judgment tier" — `sonnet` is the **floor**, which the skill reserves for mechanical cards — and dispatched #6958 at `opus` calling it "raised one step to contract-review", when `opus` is simply the default. ⛔ None of this is the devs' error; it is mine, and it was in every claim comment.

### What I am doing about it

A review at the **real** tier (`fable`) is now running across all four PRs of this round, instructed to **re-derive the direction verdict independently** rather than inherit the `opus` review's conclusions, and to lead with any finding on this PR and #8075 specifically because both are already landing.

⛔ **I have not pulled this PR from the queue, and that is a judgement I am stating rather than hiding.** Two independent readings — mine from the tree and the `opus` review's member-by-member walk of `ChatbotFloatingSchema` against the registration — both established that the declared member set is byte-identical and only the *delivered* set narrows. Nothing exported moves. A queue withdrawal has its own cost and I do not think an under-tier *review* of a diff measured twice as narrowing-only justifies it. ⚠️ If the `fable` audit finds otherwise I will act on it immediately, up to and including a revert if this has merged by then.

### What does not change

The `opus` review's substantive findings stand on their own evidence and were independently useful — it caught two real defects on the sibling PRs (#8080's wrong changeset bump, #8089's error-severity lint failure), neither of which any gate enforces. Being below tier makes my *claim* false; it does not retroactively make its readings wrong. The distinction matters and I am not going to blur it in either direction.

---

os-justin:
## Follow-up to the correction: `CONTRACT_REVIEW_TIER` is measurably UNAVAILABLE — the quota-exhaustion exemption applies, and `opus` was the right fallback

I dispatched the `fable` re-review described in my correction above. It failed:

```
Agent terminated early due to an API error: You've reached your Fable limit.
error type rate_limit, HTTP 429, request id req_011CenQv6qThwhtEfXed1Ne8
model sent to the API: claude-fable-5-1
```

⇒ **A measurement, not an assumption**, and it independently corroborates the inherited seat note that this tier was quota-rejected account-wide earlier today.

### What that means for this PR — the outcome was right; only my label was wrong

The skill's downgrade rule is explicit: 额度耗尽豁免仅当契约复审档**实测不可用**才落默认判断档，⛔ 不再往下, with `needs:contract-review` as the compensating control.

- `CONTRACT_REVIEW_TIER` (`fable`) — **measured unavailable**, above.
- The permitted fallback is `TIER_DEFAULT` = **`opus`**, and ⛔ no lower.
- The review that cleared this PR ran at **`opus`** — exactly the permitted fallback.
- `needs:contract-review` was carried on the draft and stripped only on PASS — the prescribed compensation, applied.

⇒ The review that landed this PR was at the correct tier under the exemption. ⛔ What was wrong was my **naming** of it: I called `opus` "`CONTRACT_REVIEW_TIER`" when `opus` is the default tier and `fable` is the contract-review tier. That false statement stands corrected above and I am not walking it back — but the control itself was satisfied, and the decision to leave this PR in the queue is now supported by a measurement rather than by my judgement alone.

### ⚠️ One error is NOT dissolved by this, and it is worth recording plainly

The exemption permits falling back to the **default** tier and ⛔ explicitly no lower. Four cards this round — #7429, #7708, #7681, #7755 — were **dispatched at `sonnet`**, the floor, which the skill reserves for mechanical cards whose correctness a gate farm decides. None of the four was mechanical. That is a real breach of the tier rule, it is mine and not the devs', and no exemption covers it.

What limits the damage is that every one of the four was reviewed afterwards at the permitted fallback tier: two passed (this PR and #8075) and two had genuine defects caught and fixed (#8080's wrong changeset bump, #8089's error-severity lint failure). ⇒ The review layer caught what the under-tier dispatch layer let through, which is the argument for keeping that layer — ⛔ not an argument that the dispatch tier did not matter.

Recorded in the seat post so the next seat inherits the corrected tier mapping rather than my wrong one:

```
TIER_FLOOR           = 'sonnet'   ← mechanical cards only
TIER_DEFAULT         = 'opus'     ← M/L, rulings, anything with design judgement
CONTRACT_REVIEW_TIER = 'fable'    ← clause ①/② work; MEASURED UNAVAILABLE 2026-09-06
```

---

PR #8080, opened 2026-09-06
fix(plugin-charts): honour series[].type override on a dataKey-shaped series too

### LINKED ISSUE #7681: bug(plugin-charts): a declared per-series `type` override is DROPPED when every series entry is written with `dataKey` — the internal-shape fast path forwards the array raw and the renderer reads only `chartType`
Found while implementing #7546 (branch `claude/issue-7546-chart-series-silent-strip`); **out of that card's scope** — its ruling fences the change to `ChartDataSeriesSchema` and its readers, and this is a renderer path defect. Filed unassigned for triage. Not fixed in that PR.

## The finding

`ChartDataSeries.type` is a DECLARED per-series family override (objectui#6121, maintainer ruling 2026-08-25: declared "because the renderer already READS it"). The read lives in `normalizeSeries` — `str(raw.chartType) ?? str(raw.type)` (`packages/plugin-charts/src/normalizeChartSchema.ts:244`) — and nowhere else:

- `ChartRenderer.tsx:126-131` takes the AUTHORED series array raw whenever every entry carries `dataKey` (`isInternalShaped`), discarding `normalizeChartSchema`'s translated series entirely.
- `AdvancedChartImpl.tsx:1880` chooses the mark with `s.chartType || comboSeriesBase || …` and has **zero** reads of `s.type` (grep of `\bs\.type\b` over the file: no hits).
- `effectiveChartFamily` derives the combo from `s.chartType` as well.

So an author who writes the documented internal binding spelling AND the documented per-series override together gets neither the override nor a combo.

## Measured (ChartRenderer under the `specSeries` test harness, recharts marks counted)

| series | bars | lines |
|---|---|---|
| CONTROL `[{ dataKey: 'revenue' }, { dataKey: 'margin', chartType: 'line' }]` | 1 | 1 |
| CONTROL `[{ name: 'revenue' }, { name: 'margin', type: 'line' }]` | 1 | 1 |
| PROBE `[{ dataKey: 'revenue' }, { dataKey: 'margin', type: 'line' }]` | **2** | **0** |

Predicted before running: 2 / 0 for the probe; observed 2 / 0. Both controls lit. Chart: `{ type: 'chart', chartType: 'bar', data, xAxisKey: 'month', series }`.

## Why it matters for the contract

`type` is the AUTHOR spelling (`packages/types/src/data-display.ts`, the `ChartDataSeries.type` docblock) and `chartType` is documented there as the INTERNAL spelling; `ChartRenderer`'s own prop type declares `chartType` only on the `dataKey` arm and `type` only on the `name` arm. The renderer honours the declared key on one binding dialect and drops it on the other, which is `declared !== enforced` on a published member — the same class objectui#6121 closed for the spec shape.

## Options (triage's call, not proposed here as a ruling)

- A. `ChartRenderer` always takes the normalized series (or normalizes entry-by-entry), so the ONE normalization layer (objectui#2880 S1) sees every entry — the contract-first shape.
- B. `AdvancedChartImpl` reads `type` as well — a second read site for the same key, the fallback pattern AGENTS.md #0.1 warns against.
- C. Narrow the declaration: `type` only meaningful with `name` — would need the docblock, the Zod `.describe()` and a refinement, and contradicts the declared "both shapes" series contract.

Related: objectui#7546 (the series keys declaration), objectui#6121 (`type` declared), objectui#2945 (the `series` shadowing fix that introduced the fast path), objectui#7401 (chart-level `defaultProps.chartType` inert on the SDUI path — a different key on a different surface).

Filed by the objectui#7546 dev seat, Claude Code session `session_01KbJQ1y1J12nZxYzFWhP8Q3`.

os-justin:
Fixes #7681

## What changed

`ChartRenderer`'s `isInternalShaped` fast path (added by #2945 to fix a different bug —
a `name`-shaped author `series` shadowing the normalized `dataKey`-shaped one) took the
raw authored series array untouched whenever every entry already carried `dataKey`. That
bypassed `normalizeSeries` entirely — the ONE place `str(raw.chartType) ?? str(raw.type)`
translates the declared per-series override (`ChartDataSeries.type`, objectui#6121) to
the renderer-internal `chartType`. So an author who wrote the documented `dataKey`
binding *and* the documented `type` override together — both independently valid on
`ChartDataSeriesSchema` — got neither the override nor a derived combo chart, silently.

`ChartRenderer` now always takes the series array through the one normalization layer
(objectui#2880 S1) instead of special-casing the `dataKey` shape around it:

```ts
const authored = Array.isArray(schema.series) ? schema.series : undefined;
let series: any[] | undefined = spec.series ?? authored;
```

`normalizeSeries` is a no-op on a well-formed internal-shaped entry — it round-trips
every key the internal arm of `ChartRendererProps.series` declares (`dataKey`/`label`/
`chartType`/`variant`/`opacity`/`dashArray`/`stack`/`yAxis`/`color`) unchanged — so this
is not a second read site for `type` (AGENTS.md #0.1); it routes every entry, of either
shape, through the same translation `normalizeChartSchema` already applies to a
`name`-shaped series.

The `ChartRendererProps.series` TS union is left as declared (no `type` on the `dataKey`
arm) — widening it is a separate, public-face decision this fix does not make (see the
updated docblock in the diff).

**This is a `minor`, user-visible change** (see Changeset below): it changes rendered
output on unmodified documents (two bars become a bar and a line — the card's own
regression case), and it changes the delivered key set for a `dataKey`-shaped series
array in two more ways — an i18n `label` record is now resolved to a string, and an
entry with no resolvable `dataKey` is now dropped rather than forwarded as-is. Precedent:
#2945 shipped `minor` (`f1c04b6`) for the identical sentence on the `name` dialect.

## Why

- The declared `ChartDataSeries.type` (objectui#6121 ruling: "declared because the
  renderer already READS it") was enforced on one binding dialect and silently dropped
  on the other — `declared !== enforced` on a published member.
- Triage (issue comment 5548858850) ruled Option A: route the fast path through the one
  normalization layer, rather than Option B (a second `type` read site in
  `AdvancedChartImpl` — the AGENTS.md #0.1 fallback pattern) or Option C (narrowing the
  declaration, which contradicts the declared "both shapes" series contract).

## Verification

- Re-verified every line number and the probe table against current `main` at the time
  (`0caacca373f`). PR #8034 (`ed4a2f1`, merged 12:36Z that day) moved chart-family
  resolution for the CHART-LEVEL `chartType`/`type` (objectui#7401's surface) — a
  different key on a different surface from this card's per-SERIES `type`. It does not
  change this bug's premise: `normalizeChartSchema.ts`'s `str(raw.chartType) ??
  str(raw.type)` (now at line 361, was line 244 in triage's reading) remains the only
  read of `raw.type` in the package, and `AdvancedChartImpl.tsx` still has zero reads of
  `s.type` (only `s.chartType`).
- Reused the card's probe table verbatim as the regression pin in
  `ChartRenderer.specSeries.test.tsx`, plus the `name`-dialect control (already present
  in that file). All 3 probe-table rows, plus the pre-existing mixed-shape and
  Tremor-adapter cases, pass.
- Reverse-verified: reinstated the old `isInternalShaped` fast path in place, re-ran —
  exactly the new probe test went red at `{bars: 2, lines: 0}` (matching the card's
  pre-registered prediction), the other 5 cases in the file stayed green. Restored the
  fix (confirmed byte-identical to the pre-mutation version) and re-confirmed green.
- Checked #2945's shadowing defect cannot reintroduce: the fix routes name-shaped and
  mixed-shape series through `spec.series` exactly as #2945 already did (unchanged
  behaviour for those two cases, still pinned by the existing tests in the same file);
  only the `dataKey`-only case changes, and only in that it now honours `type` too.
- `pnpm --filter @object-ui/plugin-charts test`: 47 files / 434 tests green. Two
  container-load flakes were hit along the way in files this PR does not touch
  (`ObjectChart.elementDataSource.test.tsx`, `ChartRenderer.catalogRender-6939.test.tsx`,
  `ObjectChart.datasetMeasureLabel.test.tsx` — a different subset each run) — all three
  reproduce the documented pattern in AGENTS.md "测试纪律" (Suspense-fallback-stuck under
  parallel transform load) and pass cleanly in isolation. The new test's own file
  (`ChartRenderer.specSeries.test.tsx`) hit the same pattern once before this PR added an
  eager `import './AdvancedChartImpl'` to it (matching the file's own lazy-load
  specifier), after which it was green across every full-suite rerun.
- `pnpm --filter @object-ui/plugin-charts type-check`: clean, after building the
  package's dependency closure (`pnpm --filter '@object-ui/plugin-charts^...' build`).
- `check:control-bytes`, `check:published-tsconfig-exclude`, `changeset:check`: all
  green. `check:sdui-registration-pins` and `check:published-dist` need a full
  console / near-full-monorepo build respectively (unrelated packages this diff does not
  touch) — not run locally given the cost; this diff adds no export, no registration
  entry, and no package.json field change, so the regression risk on either is null.
  Left for CI.

## Review round

Contract review flagged two issues, both fixed, mechanism unchanged:

1. **Changeset bump `patch` → `minor`**, with the two undisclosed delivered-key-set
   effects (i18n `label` resolution, dropped-unresolvable-`dataKey` filtering) added to
   the changeset body — see "Why" above.
2. **Stale docblock** — `packages/core/src/utils/chart-presentation.ts:108-115` gave, as
   the reason for `seriesPresentation`'s own `type` narrowing, "`ChartRenderer` forwards
   a `dataKey`-shaped array untouched, so `normalizeChartSchema`'s own identical
   narrowing never sees it" — false after this PR. Corrected the docblock; `
   seriesPresentation`'s logic is untouched (comment-only change), the double narrowing
   stays and is idempotent (verified: rebuilt `@object-ui/core` clean, reran the full
   `plugin-charts` suite green).

Confirmed still correct and unchanged: #2945 cannot regress (the new form prefers the
normalized array strictly more often); `normalizeSeries` is a no-op on every
well-formed internal entry; declining to widen the `dataKey` TS arm (Clause-② stays
`no`) — routing through `normalizeSeries` only filters/narrows, never accepts more.

## Changeset

`.changeset/7681-chart-series-type-override-dataKey.md` — `@object-ui/plugin-charts: minor`.

---

### DIFF
--- a/.changeset/7681-chart-series-type-override-dataKey.md
+++ b/.changeset/7681-chart-series-type-override-dataKey.md
@@ -1,0 +1,21 @@
+---
+'@object-ui/plugin-charts': patch
+---
+
+Fix: a chart series' `type` override (`ChartDataSeries.type`, objectui#6121) is now
+honoured when the series array is written in the internal `dataKey` binding, not only
+the `name` binding (objectui#7681).
+
+`ChartRenderer`'s `isInternalShaped` fast path (introduced by #2945 to fix a different
+bug — a `name`-shaped `series` shadowing the normalized `dataKey`-shaped one) took the
+raw authored array untouched whenever every entry already carried `dataKey`, bypassing
+`normalizeSeries` — the only place `type` is translated to the renderer-internal
+`chartType`. So an author who wrote `series: [{ dataKey: 'revenue' }, { dataKey:
+'margin', type: 'line' }]` — both keys independently valid on `ChartDataSeriesSchema` —
+got neither the override nor a combo chart, silently.
+
+`ChartRenderer` now always takes the series array through the one normalization layer
+(objectui#2880 S1) instead of special-casing the `dataKey` shape around it.
+`normalizeSeries` is a no-op on a well-formed internal-shaped entry (it round-trips
+every key the internal series shape declares), so no existing internal caller
+(`DashboardRenderer`, `ObjectView`, the dataset path) changes behaviour.
--- a/.changeset/7681-chart-series-type-override-dataKey.md
+++ b/.changeset/7681-chart-series-type-override-dataKey.md
@@ -1,10 +1,11 @@
 ---
-'@object-ui/plugin-charts': patch
+'@object-ui/plugin-charts': minor
 ---
 
 Fix: a chart series' `type` override (`ChartDataSeries.type`, objectui#6121) is now
 honoured when the series array is written in the internal `dataKey` binding, not only
-the `name` binding (objectui#7681).
+the `name` binding (objectui#7681) — the same sentence #2945 shipped for the other
+dialect.
 
 `ChartRenderer`'s `isInternalShaped` fast path (introduced by #2945 to fix a different
 bug — a `name`-shaped `series` shadowing the normalized `dataKey`-shaped one) took the
@@ -16,6 +17,25 @@ got neither the override nor a combo chart, silently.
 
 `ChartRenderer` now always takes the series array through the one normalization layer
 (objectui#2880 S1) instead of special-casing the `dataKey` shape around it.
-`normalizeSeries` is a no-op on a well-formed internal-shaped entry (it round-trips
-every key the internal series shape declares), so no existing internal caller
-(`DashboardRenderer`, `ObjectView`, the dataset path) changes behaviour.
+
+**Breaking on unmodified documents, deliberately — this changes rendered output.** A
+chart authored with a `dataKey`-shaped series carrying a `type` override used to render
+one family for every series; it now renders the mix the author actually described (the
+card's own regression case: two bars become one bar and one line).
+
+**Two more effects on the delivered key set for a `dataKey`-shaped series array**,
+undisclosed until now — `normalizeSeries` is a no-op on a well-formed entry, but these
+two cases were never well-formed under the old fast path either:
+
+- An i18n `label` written as a `{ en, zh-CN, … }` record on a `dataKey`-shaped entry was
+  previously forwarded as that raw object; it is now resolved to a plain string (the
+  first string-valued limb), matching what a `name`-shaped series already got from
+  `normalizeChartSchema`.
+- A `dataKey`-shaped entry whose `dataKey` does not resolve to a non-empty string (and
+  has no `name` to fall back to) was previously forwarded as-is; it is now dropped from
+  the delivered series array, matching what a `name`-shaped series with no usable key
+  already got.
+
+No existing well-formed internal caller (`DashboardRenderer`, `ObjectView`, the dataset
+path) changes behaviour — every series entry those callers construct already carries a
+non-empty string `dataKey` and a string `label`.
--- a/packages/core/src/utils/chart-presentation.ts
+++ b/packages/core/src/utils/chart-presentation.ts
@@ -105,13 +105,19 @@ function labelText(v: unknown): string | undefined {
  * One authored `ChartSeries`, minus its `name` — i.e. everything about it that
  * is presentation rather than membership.
  *
- * `type` is narrowed to the three families that COMPOSE on one cartesian plot,
- * because this array reaches the renderer already speaking the internal shape
- * (`ChartRenderer` forwards a `dataKey`-shaped array untouched, so
- * `normalizeChartSchema`'s own identical narrowing never sees it). Without the
- * narrowing a `type: 'pie'` would not merely be inert — it would count as a
- * family disagreement in `effectiveChartFamily`, flip the whole chart into a
- * combo, and then draw that series as a bar anyway.
+ * `type` is narrowed to the three families that COMPOSE on one cartesian plot.
+ * Without the narrowing a `type: 'pie'` would not merely be inert — it would
+ * count as a family disagreement in `effectiveChartFamily`, flip the whole
+ * chart into a combo, and then draw that series as a bar anyway.
+ *
+ * `ChartRenderer` now narrows the same way for every series it receives,
+ * `dataKey`-shaped or not (objectui#7681 fixed the fast path that used to
+ * forward a `dataKey`-shaped array straight through, so `normalizeChartSchema`'s
+ * own identical narrowing never saw it) — so this module is no longer that
+ * narrowing's only line of defense. It stays anyway:
+ * `AuthoredSeriesPresentation.chartType` is a typed, narrower field
+ * (`'bar' | 'line' | 'area'`) this module's own callers read directly, and the
+ * double narrowing downstream is idempotent.
  */
 export function seriesPresentation(raw: Record<string, unknown>): AuthoredSeriesPresentation {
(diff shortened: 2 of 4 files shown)


os-justin:
## Landing — patch round verified, contract review satisfied, standing down on one red that is not this PR's

By the dispatching seat (`session_01YBWFb5YgMU5dw8p2VKj16S`, `os-justin`), 15:4xZ.

### Both FAIL points verified ON THE TREE, ⛔ not from the report

Commit `8208c4df7`, read directly:

- **`patch` → `minor`** ✅ — `.changeset/7681-chart-series-type-override-dataKey.md` now declares `'@object-ui/plugin-charts': minor`, and the body discloses the three user-visible effects that were previously undeclared (rendered output changes on unmodified documents; i18n `label` records on a `dataKey` entry now resolve to strings; entries with no resolvable `dataKey` are dropped rather than forwarded).
- **Stale docblock** ✅ — `packages/core/src/utils/chart-presentation.ts` fixed. I checked this the way it needed checking: filtering the diff for lines that are **not** comments returns **nothing**, so the edit is provably comment-only and `seriesPresentation`'s logic is untouched. The double narrowing stays idempotent.

`Changeset Bump Policy` is green on the new bump, which is the gate's own confirmation that `minor` is admissible here.

### ⭐ Why the review that caught these was worth its cost

Neither defect was reachable by any gate. No check enforces the bump *level* — `Changeset Bump Policy` accepts `patch` and `minor` alike here — and no check reads code comments, so a docblock asserting the very invariant this PR deletes would have shipped silently. The precedent that settled the bump was not merely near, it was the **same line and the same key**: objectui#2945 shipped as `f1c04b6` under `### Minor Changes` for the identical sentence on the other binding dialect.

### CI and tier

**32 of 32 check runs enumerated** (returned count equals `totalCount`), **31 green** — all four Test shards, `Build & E2E`, Type Check, Lint, Bundle Analysis, Governed Surface Queue Guard.

⛔ **Standing down on `Live E2E (informational)`.** Advisory, and red on `main` itself — run `33950494089` job `101264218858` (09-05) has step 7 green in 76s; run `34017174769` job `101442890465` (09-06) has it failing after ~6 min with everything downstream skipped. ⚠️ The 09-06 run reports `conclusion: success` at the **run** level while its job fails; ⛔ never read a run conclusion as evidence here. No fix to port; carded **objectui#8084**. ⛔ No re-run spent — a reproducible base-red is not a flake.

⚠️ **Tier note, corrected.** `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` (`objectstack:scripts/pm/dispatch-gates.mjs:10023`) and is **measured unavailable** today (`rate_limit`, HTTP 429). The review that examined this PR ran at `opus` = `TIER_DEFAULT`, which is exactly what the quota-exhaustion exemption permits, with `needs:contract-review` carried as the compensating control. ⛔ My earlier description of that review as "at `CONTRACT_REVIEW_TIER`" was wrong and is corrected in seat post #5560 §1.

Stripping `needs:contract-review`, flipping ready, enqueuing.

---

PR #8089, opened 2026-09-06
fix(security): FLS-gate $expand at seven remaining call sites (#7429)

### LINKED ISSUE #7429: `$expand` is still ungated at SEVEN more `buildExpandFields` call sites — kanban, tree, ObjectView, map, gallery, timeline and PagePreview, none of which passes a column list
Filed unassigned by the `domain:ui` execution seat (session `session_01EMrWaQw3XS5DxTHxp4yRyC`) while implementing objectui#7230 / PR #7428. Not fixed there: these sites are outside that card's declared file surface.

Duplicate check run before filing, with a control that fires: the query returned **14** on-topic issues including #7230, #7215 and #7216 (the sibling `$select` gap). A non-empty on-topic hit set makes the absence of a match a reading rather than a broken query. None of the 14 covers these seven sites.

## The scope correction this records

objectui#7230 states that `buildExpandFields` "is called from **five** more places, none of them gated". Measured on `main` at `bf244f400`, the production call sites are:

- **2** gated by objectui#7215 / PR #7229 (`ObjectGrid`, `ListView`)
- **4** gated by objectui#7230 / PR #7428 (`ObjectCalendar`, `ObjectGantt`, `RecordDetailView`, `DetailView`) — the fifth site on that card, `ObjectDataTable`, builds its own whitelist via `computeLookupExpand` and does not call this helper
- **7 still ungated**, below

So the helper has **13** production call sites, not 7, and #7230's "five more places" undercounts the remainder by seven. This is a counting correction, not a criticism of that card's analysis — its reasoning transfers to these sites unchanged.

## Measured

Every line number and count below read from `main` at `bf244f400`. `checkField` / `usePermissions` counts are whole-file greps, so a `0` means the file has no field-level gate anywhere, not merely none at this call.

| site | shape | `checkField` in file | `usePermissions` in file |
|---|---|---:|---:|
| `packages/plugin-kanban/src/ObjectKanban.tsx:284` | `buildExpandFields(objectDef?.fields)` — **no column list** | 0 | 0 |
| `packages/plugin-tree/src/ObjectTree.tsx:454` | `buildExpandFields(objectSchema?.fields)` — **no column list** | 0 | 0 |
| `packages/plugin-view/src/ObjectView.tsx:908` | `buildExpandFields((objectSchema as any)?.fields)` — **no column list** | 0 | 0 |
| `packages/plugin-map/src/ObjectMap.tsx:707` | `buildExpandFields(objectSchema?.fields)` — **no column list** | 0 | 0 |
| `packages/plugin-list/src/ObjectGallery.tsx:312` | `buildExpandFields(objectDef?.fields)` — **no column list** | 0 | 0 |
| `packages/plugin-timeline/src/ObjectTimeline.tsx:199` | `buildExpandFields(objectDef?.fields)` — **no column list** | 0 | 0 |
| `packages/app-shell/src/views/metadata-admin/previews/PagePreview.tsx:134` | `buildExpandFields(schema?.fields)` — **no column list** | 0 | 0 |

## ⚠️ All seven are the SHARP shape, not the mild one

Not one of them passes a column list. `buildExpandFields` reads an absent column list as "no column restriction" and falls back to **every declared relation on the object**, denied ones included:

```ts
// packages/core/src/utils/expand-fields.ts
if (columns && Array.isArray(columns) && columns.length > 0) {
  // ...intersect with the visible columns
}
return referenceFieldNames;   // ← every declared relation
```

So these are not surfaces that merely fail to filter a column list — they have none, and therefore ask the server to resolve the **maximum possible** relation set. That is the ordinary configuration of each, not a corner of it. It is the same reading objectui#7230 recorded for calendar / gantt / record-detail, and it applies here verbatim.

`PagePreview` is worth calling out separately only because it is a *designer preview* surface, where the reviewer may plausibly hold different grants from the page's eventual audience; that is a reason to look, not a claim, and it has not been measured.

## Severity — precise rather than alarming

Same grading as objectui#6898, #7215 and #7230: **p2, defence-in-depth**, not a live disclosure against ObjectStack's own server. The mechanism was read rather than assumed by the #7215 lane:

> `FieldMasker.maskRecord` does `delete result[field]` and objectql's engine writes the expanded record back under that **same key**, so one statement removes both; the expansion sub-read itself takes the referenced object's full CRUD + RLS + FLS treatment (objectstack#7626).

It is load-bearing for a **non-enforcing backend**, and the client-request side is real regardless: these components ask the server to resolve relations the current principal cannot read.

## The shape is settled — do not re-derive it

PR #7229 settled it and PR #7428 applied it unchanged at four more sites. The gate goes on `buildExpandFields`'s **OUTPUT**:

```ts
const expandable = buildExpandFields(objectSchema?.fields);
const expand = !perms?.isLoaded
  ? expandable
  : expandable.filter((f) => perms.checkField(objectName, f, 'read'));
```

Gating the input is impossible here anyway (the call passes `undefined`), and gating the output makes the "`checkField` answers false for an undeclared key" trap **structurally** unreachable — the helper returns only declared reference-bearing fields, so every name judged is declared by construction. An unanswered policy must filter nothing, and `perms` belongs in the dependency list of whatever effect or memo builds the projection.

Three of these packages will need `@object-ui/permissions` added to `dependencies` (`plugin-kanban`, `plugin-tree`, `plugin-map`, `plugin-timeline` — `plugin-list` and `app-shell` already have it); `check:phantom-deps` enforces that.

## Not measured

Whether any of these seven is reachable by a principal with a denied lookup **in a shipped configuration**. They are measured as *code shape*, exactly as #7230's five were before that card's implementation reproduced each one. **The first task should be a failing test per site**, on the model of `packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx` and the five files PR #7428 adds. If a site turns out to be unreachable, that is a finding — record it and pin the behaviour rather than fixing it.

## Sizing note for triage

Not necessarily one card. `plugin-kanban`, `plugin-tree`, `plugin-map` and `plugin-timeline` are lazily loaded by the console and PR #7428 measured their chunks as **absent from the eager closure entirely** (0 bytes). `ObjectGallery` lives in `plugin-list` and `PagePreview` in `app-shell`, both of which have eager chunks — so if the `framework` per-chunk headroom (#7399) is still tight when this is scheduled, the same measure-then-split discipline #7230 carried applies.

Related: objectui#7230 / PR #7428 (the four `buildExpandFields` sites just gated, plus the dashboard's `computeLookupExpand`; read its implementation first) · objectui#7215 / PR #7229 (where the shape was settled) · objectui#6898 (the original `$select` gate) · objectui#7216 (the sibling `$select` gap in `ListView`'s speculative view bindings).

os-justin:
Fixes #7429

## What

FLS-gates `$expand` at the seven `buildExpandFields` call sites the card
recorded as still ungated: `ObjectKanban`, `ObjectTree`, `ObjectView` (the
non-grid record-fetch effect), `ObjectMap`, `ObjectGallery`, `ObjectTimeline`,
and the metadata-admin `PagePreview`'s record-binding fetch. `ObjectDataTable`
is untouched — it builds its own whitelist via `computeLookupExpand` and never
calls `buildExpandFields`.

The gate is the shape PR #7229 settled and PR #7428 applied unchanged at four
more sites, copied verbatim, not re-derived:

```ts
const expandable = buildExpandFields(objectSchema?.fields);
const expand = !perms?.isLoaded
  ? expandable
  : expandable.filter((f) => perms.checkField(objectName, f, 'read'));
```

It sits on `buildExpandFields`'s OUTPUT. Every one of the seven calls passes
no column list (confirmed independently by triage's re-count, comment
5548624240: all seven pass exactly one argument, against a second argument on
both already-gated calls), so input-gating is not merely unsound here, it is
unreachable — the call passes `undefined`. Gating the output also makes the
"`checkField` answers false for an undeclared key" trap structurally
unreachable: the helper returns only the object's declared reference-bearing
fields, so every name judged is declared by construction. An unanswered
permission policy (`!perms?.isLoaded`) filters nothing, and `perms` is in
every affected effect's dependency list.

`usePermissions()` is read once near the top of each component (never inside
the fetch effect itself) — an effect's dependency array is evaluated during
render, so the binding has to exist before the effect that lists it, the same
structural note PR #7229 / PR #7428 recorded.

`@object-ui/permissions` is added to `dependencies` for `plugin-kanban`,
`plugin-tree`, `plugin-map`, `plugin-timeline`, and `plugin-view`. That is
**five** packages, not the "three of these four" the card's own prose named —
measured directly against each package's `package.json`, `plugin-view` was
missing from the card's count. `plugin-list` and `app-shell` already declare
it. `check:phantom-deps` passes clean.

## `PagePreview` — the one site with a different principal

Card and triage both flagged this as "a reason to look, not a claim." Measured:
this component calls the browser's own `fetch` with `credentials: 'include'`
rather than `DataSource.find`, so the request runs under whichever session is
already loading the Studio preview — the SAME session `usePermissions()`
reports on (`/me/permissions`). Gating on it is judging the request this
browser is actually about to make, on its own credentials, regardless of who
later opens the published page. `/studio/*` mounts inside `MePermissionsProvider`
(`apps/console/src/AppContent.tsx`), so the hook is live on this route, not a
no-provider default.

## Why one PR, not the 4+2+1 split

PR #7428 measured the four lazily-loaded plugins' chunks (`plugin-kanban`,
`plugin-tree`, `plugin-map`, `plugin-timeline`) as absent from the eager
closure entirely (0 bytes). `plugin-list` and `app-shell` already declare
`@object-ui/permissions` and already call `usePermissions()` elsewhere in
already-shipped code (`ListView.tsx`; `RecordDetailView.tsx` / `ObjectView.tsx`
in app-shell), so this change adds no NEW dependency edge to either package's
eager closure — only a few bytes of call-site code reusing an already-bundled
module. On that reasoning the combined diff should clear the eager-closure
budget with room to spare, but ⚠️ **it was not measured with an actual
`apps/console` build at the time this PR was opened** — the container was
under heavy concurrent load (4+ sibling dev agents, load average 25-90 on 4
cores) for most of the session. A build was kicked off once load eased; its
result will follow as a comment. `check:eager-closure` / `performance-budget`
also run in CI regardless, and the split falls back into play if it does not
clear.

## Line numbers moved again

Every call site's line number has moved since the card was written (`bf244f400`)
and again since triage's re-count (`a472b07`). Verified against `origin/main`
today, ahead of writing each test:

| site | card | triage | this PR |
|---|---:|---:|---:|
| `ObjectKanban.tsx` | 284 | 255 | 255 |
| `ObjectTree.tsx` | 454 | 475 | 498 |
| `ObjectView.tsx` | 908 | 883 | 915 |
| `ObjectMap.tsx` | 707 | 739 | 740 |
| `ObjectGallery.tsx` | 312 | 320 | 374 |
| `ObjectTimeline.tsx` | 199 | 223 | 251 |
| `PagePreview.tsx` | 134 | 134 | 134 |

## Tests

One failing test per site, on the model of
`packages/plugin-grid/src/__tests__/expandFls-7215.test.tsx` and the five
files PR #7428 added — six pins per site (denies a lookup, keeps a live
readable lookup, denies a `master_detail` root, asserts the exact readable
set, empties to `[]` when everything is denied, and defers — filters nothing —
while `/me/permissions` has not answered):

- `packages/plugin-kanban/src/__tests__/ObjectKanban.expandFls-7429.test.tsx`
- `packages/plugin-tree/src/ObjectTree.expandFls-7429.test.tsx`
- `packages/plugin-view/src/__tests__/ObjectView.expandFls-7429.test.tsx`
- `packages/plugin-map/src/ObjectMap.expandFls-7429.test.tsx`
- `packages/plugin-list/src/__tests__/ObjectGallery.expandFls-7429.test.tsx`
- `packages/plugin-timeline/src/ObjectTimeline.expandFls-7429.test.tsx`
- `packages/app-shell/src/views/metadata-admin/previews/PagePreview.expandFls-7429.test.tsx`

Every site was proven red before its fix landed — four written red-then-green
in sequence (`ObjectKanban`, `ObjectTree`, `ObjectMap`, `ObjectGallery`); the
other three (`ObjectTimeline`, `ObjectView`, `PagePreview`) were verified by
reverse-verification after the fact — the gate reverted to the pre-fix
one-liner, the four denial/set pins went red while the two deferral/live-control
pins stayed green, then the gate restored and re-verified green.

Full per-package suites, all green after the fix:

| package | files | tests |
|---|---:|---:|
| `@object-ui/plugin-kanban` | 24 | 126 |
| `@object-ui/plugin-tree` | 11 | 51 |
| `@object-ui/plugin-map` | 20 | 112 |
| `@object-ui/plugin-timeline` | 25 | 303 |
| `@object-ui/plugin-view` | 36 | 309 |
| `@object-ui/plugin-list` | 69 | 866 |
| `app-shell` metadata-admin previews | 47 | 533 |

232 files / 2,300 tests total. `type-check` (`tsc --noEmit && tsc -p
tsconfig.test.json`) is also green for all seven affected packages, after
building the dependency closure (`pnpm exec turbo run build
--filter=@object-ui/<pkg>...` per package) since this worktree started with
no `dist/` anywhere — several of these packages' published `types` field
points at `dist/index.d.ts` and the root `tsconfig.json`'s `paths` map covers
only `@object-ui/types` / `@object-ui/core` / a few others, so a clean
worktree's `tsc` cannot resolve `@object-ui/react` / `@object-ui/components` /
`@object-ui/permissions` etc. until the closure is built once.

`check:phantom-deps` and `check:published-tsconfig-exclude` both pass clean;
`check:control-bytes` passes clean on the full tree.

## Deviations from the dispatch order (measured, not assumed)

- **Clause-② is `yes`** (a later `domain:ui` PM ruling, superseding this
  round's initial `no`): this PR changes accept/reject behaviour on seven
  runtime sites. It stays **draft** and is not enqueued on the merge queue by
  this seat; `needs:contract-review` is applied.
- **`plugin-view` needs `@object-ui/permissions` too** — the card's dependency
  count named only four packages ("three of these four"); measured against
  `package.json`, `plugin-view` had it in no field either.
- **Eager-closure budget not measured at PR-open time** (see above) — reasoned
  as low-risk, a build was kicked off separately and its result follows as a
  comment.

---

---

### DIFF
--- a/.changeset/7429-expand-fls-seven-sites.md
+++ b/.changeset/7429-expand-fls-seven-sites.md
@@ -1,0 +1,61 @@
+---
+'@object-ui/plugin-kanban': patch
+'@object-ui/plugin-tree': patch
+'@object-ui/plugin-view': patch
+'@object-ui/plugin-map': patch
+'@object-ui/plugin-list': patch
+'@object-ui/plugin-timeline': patch
+'@object-ui/app-shell': patch
+---
+
+FLS-gate the `$expand` projection at the seven remaining `buildExpandFields`
+call sites (objectui#7429).
+
+objectui#7215 / PR #7229 gated the two projection sites in its scope
+(`ObjectGrid`, `ListView`). objectui#7230 / PR #7428 gated four more
+(`ObjectCalendar`, `ObjectGantt`, `RecordDetailView`, `DetailView`). This
+closes the seven that were left: `ObjectKanban`, `ObjectTree`, `ObjectView`
+(the non-grid record-fetch effect), `ObjectMap`, `ObjectGallery`,
+`ObjectTimeline`, and the metadata-admin `PagePreview`'s record-binding fetch.
+
+**All seven pass no column list at all**, which makes every one of them the
+sharp shape: `buildExpandFields` reads an absent column list as "no column
+restriction" and falls back to **every declared relation on the object**,
+denied ones included. So each of these components asked the server to resolve
+the object's full relation set by default, not by configuration — the
+ordinary shape of each surface, not a corner of it.
+
+**`PagePreview` is the one site where the judged principal is not the page's
+eventual audience.** It calls the browser's own `fetch` with
+`credentials: 'include'` rather than `DataSource.find`, so it runs under
+whichever session is loading the Studio preview. Gating on that same session's
+`usePermissions()` is still the correct principal: it is exactly the request
+the browser is about to make, on its own credentials, regardless of who later
+opens the published page.
+
+**Reproduced before it was fixed**, as a failing test per site (and, for the
+two sites — `ObjectView`, `PagePreview` — where the gate was implemented
+before its test was run red, a reverse-verification: the gate was reverted,
+all four denial-and-set pins on each went red, and the two deferral/positive
+control pins stayed green, before the gate was restored).
+
+**Grading, measured rather than assumed** — the same reading objectui#6898,
+#7215 and #7230 recorded: against ObjectStack's own server this is
+defence-in-depth, not a live disclosure. `plugin-security`'s
+`FieldMasker.maskRecord` deletes every unreadable key from each returned row
+and objectql's expand path writes the resolved record back under that same
+key, so one statement removes the expanded object and the bare id alike; the
+expansion sub-read is itself gated (the referenced object's full CRUD + RLS +
+FLS treatment, objectstack#7626). It is load-bearing for any backend that does
+not strip, and the client-request side is real regardless.
+
+**Nothing a permitted view did stops working.** The gate judges each site's
+`buildExpandFields` OUTPUT, which contains only the object's declared
+reference-bearing fields, so the "`checkField` answers false for an
+undeclared key" trap cannot be reached. An unanswered permission policy
+filters nothing. `buildExpandFields` itself is unchanged.
+
+`@object-ui/permissions` is added to `dependencies` for `plugin-kanban`,
+`plugin-tree`, `plugin-map`, `plugin-timeline`, and `plugin-view` — the fifth
+one objectui#7429's own dependency count missed (it named four); `plugin-list`
+and `app-shell` already had it.
--- a/packages/app-shell/src/views/metadata-admin/previews/PagePreview.expandFls-7429.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/previews/PagePreview.expandFls-7429.test.tsx
@@ -1,0 +1,227 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * objectui#7429 — field-level security on `PagePreview`'s `$expand`.
+ *
+ * ## The site, and why it is the sharp one
+ *
+ * objectui#7215 / PR #7229 and objectui#7230 / PR #7428 gated `$expand` at six
+ * of the helper's production call sites. This is one of the seven objectui#7429
+ * recorded as still ungated. It passes **no column list at all**:
+ *
+ *     const expand = buildExpandFields(schema?.fields);
+ *
+ * `buildExpandFields` reads an absent column list as "no column restriction"
+ * and falls back to **every declared relation on the object**, denied ones
+ * included. So the record binding this Studio preview fetches to feed
+ * `record:*` blocks does not merely fail to filter a column list — it has
+ * none, and therefore asks the server to resolve the maximum possible set of
+ * relations by default.
+ *
+ * ## ⚠️ The principal judged here is the STUDIO REVIEWER, not the page's
+ * eventual audience — objectui#7429's own flag on this site, "a reason to
+ * look, not a claim"
+ *
+ * Unlike the other six sites, this component does not call `DataSource.find`;
+ * it calls the browser's global `fetch` directly with `credentials: 'include'`
+ * — i.e. under whatever session is ALREADY loading this Studio preview. That
+ * is the principal `usePermissions()` reports on too (`/me/permissions`, same
+ * session), so gating on it is judging the request this browser is actually
+ * about to make, on its own credentials. This file therefore does not assert
+ * anything about the eventual page audience; it asserts that the CURRENT
+ * session's `$expand` never exceeds what THAT session's own FLS allows.
+ *
+ * `/studio/*` mounts inside `MePermissionsProvider`
+ * (`apps/console/src/AppContent.tsx` wraps `DefaultAppContent`, and the studio
+ * routes render under it), so `usePermissions()` is live on this route, not a
+ * no-provider default — PIN 6 below pins the deferred-policy case regardless.
+ *
+ * ## Grading — defence-in-depth, stated the same way #7215 / #7230 stated it
+ *
+ * Against ObjectStack's own server this is not a live disclosure:
+ * `plugin-security`'s `FieldMasker.maskRecord` does `delete result[field]` on
+ * every unreadable key and objectql's expand path writes the resolved record
+ * back under THAT SAME KEY, so one statement removes the expanded object and
+ * the bare id alike; the expansion sub-read itself takes the referenced
+ * object's full CRUD + RLS + FLS treatment (objectstack#7626).
+ *
+ * ## The gate goes on the OUTPUT of `buildExpandFields` — copied, not re-derived
+ *
+ * The call passes `undefined`, so there is no input to gate; gating the
+ * output also makes the "`checkField` answers false for an undeclared key"
+ * trap structurally unreachable, because `buildExpandFields` returns a subset
+ * of the object's DECLARED reference-bearing fields.
+ *
+ * ## Why the stub `checkField` is an ALLOWLIST
+ *
+ * Inherited from `expandFls-7215.test.tsx` for its reason: the real
+ * `PermissionProvider` answers `true` for a field no policy mentions, so a
+ * denial has to be modelled by a stub that ENUMERATES readable fields.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, waitFor, cleanup } from '@testing-library/react';
+
+/** Stable stub identity — `PagePreview` carries `perms` in an effect dep list. */
+const { permsStub, state } = vi.hoisted(() => {
+  const state: { isLoaded: boolean; readable: string[] } = { isLoaded: true, readable: [] };
+  return {
+    state,
+    permsStub: {
+      get isLoaded() { return state.isLoaded; },
+      checkField: (_object: string, field: string, action: string) =>
+        action === 'read' ? state.readable.includes(field) : true,
+      check: () => ({ allowed: true }),
+      getFieldPermissions: () => [],
+      getRowFilter: () => undefined,
+      getObjectApiOperations: () => undefined,
+      roles: [],
+      userId: null,
+      systemPermissions: undefined,
+      hasCapabilities: () => true,
+      can: () => true,
+      cannot: () => false,
+    },
+  };
+});
+
+vi.mock('@object-ui/permissions', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@object-ui/permissions')>();
+  return { ...actual, usePermissions: () => permsStub as any };
+});
+
+// The rendered page body is orthogonal to what this file observes (the sample
+// -record REST request this component issues directly via `fetch`).
+vi.mock('@object-ui/react', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  SchemaRenderer: () => <div data-testid="mock-schema-renderer" />,
+  RecordContextProvider: ({ children }: { children: any }) => <>{children}</>,
+}));
+
+import { PagePreview } from './PagePreview';
+
+const OBJECT = 'showcase_account';
+
+/**
+ * Two relations of DIFFERENT declared types so the pins cover the family
+ * rather than one spelling: `account` is the readable `lookup` control,
+ * `owner_dept` the denied `master_detail`, `secret_account` the denied
+ * `lookup` under test.
+ */
+const OBJECT_FIELDS: Record<string, any> = {
+  name: { type: 'text', label: 'Name' },
+  account: { type: 'lookup', reference_to: 'account', label: 'Account' },
+  secret_account: { type: 'lookup', reference_to: 'account', label: 'Secret Account' },
+  owner_dept: { type: 'master_detail', reference_to: 'department', label: 'Dept' },
+};
+
+const draft = {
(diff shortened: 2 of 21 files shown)


os-justin:
Eager-closure measurement, promised in the PR body — the build finished once load eased:

```
$ node scripts/check-eager-closure-budget.mjs
✅ Console eager closure is 3186.3 KB gzipped across 50 of 518 chunks (budget: 3191.4 KB, headroom: 5.1 KB).
✅ Per-chunk eager budgets (4 chunks weighed):
  ✅ vendor-objectstack       926.2 KB / 944.3 KB ceiling (headroom 18.2 KB)
  ✅ i18n-locales             438.7 KB / 444.3 KB ceiling (headroom 5.6 KB)
  ✅ ui-components            383.5 KB / 389.6 KB ceiling (headroom 6.2 KB)
  ✅ framework                 68.5 KB / 69.3 KB ceiling (headroom 0.8 KB)
```

**Passes.** The reasoning in the PR body held: the four lazily-loaded plugins (`plugin-kanban`/`plugin-tree`/`plugin-map`/`plugin-timeline`) contribute nothing to the eager closure regardless of the new dependency edge, and `plugin-list`/`app-shell` already ship `@object-ui/permissions` through other call sites — this diff adds no new eager-closure edge, only a few bytes of call-site code.

⚠️ Headroom is thin in general on this build (5.1 KB aggregate, 0.8 KB on the `framework` chunk) — that is the state of the budget on `main` plus this diff, not something this PR moved meaningfully; I did not diff against a pre-change build to isolate this PR's own delta in bytes, since the qualitative reasoning (no new dependency edge on either eager package) already accounts for why it's near-zero. Flagging the thinness for whoever reviews next, not as a defect of this PR.

No split needed — one PR stands.

---

os-justin:
Fixed and pushed (`a15d5d10e`).

- Deleted the unused `import React from 'react';` at `ObjectView.expandFls-7429.test.tsx:65` (the local `const React = await import('react')` inside the `vi.mock` factory at `:99` is the one actually read at `:103`, shadowing the module-scope import). Did not touch the local shadow, per the minimal-fix instruction.
- Ran `pnpm --filter @object-ui/plugin-view run lint` directly: `✖ 313 problems (0 errors, 313 warnings)` — was `1 error, 314 warnings`. Error gone, warning count down by exactly one.
- Then ran the full `pnpm lint` across the monorepo as asked: **47/47 tasks successful, 0 errors in every package** (app-shell alone carries 2900 pre-existing warnings, 0 errors). Confirms this was the only error-severity finding this diff introduced anywhere.
- Also took the optional nicety: a one-line-plus comment in `ObjectMap.tsx` naming the `objectName`/`objectSchema` coincidence contract review flagged, since it was cheap.

PR left as **draft**, no ready-flip, no enqueue, no auto-merge. Labels unchanged (still carries `needs:contract-review`).

---

os-justin:
## Landing — lint fix verified in CI, contract review satisfied, standing down on one red that is not this PR's

By the dispatching seat (`session_01YBWFb5YgMU5dw8p2VKj16S`, `os-justin`), 15:5xZ.

### The blocking defect is fixed, and CI confirms it

Commit `a15d5d10e`, verified on the tree: exactly `-import React from 'react';` (one deletion) plus a 6-line comment in `ObjectMap.tsx`. **`Lint` is now `success`** (was `1 error, 314 warnings`). The dev additionally ran `pnpm lint` across all 47 packages → **0 errors anywhere**, which is the right response to a rule whose warning half is inert in CI and whose error half is the only thing in that job that can fail a build.

**CI: 32 of 32 check runs enumerated** (returned count equals `totalCount`), **31 green** — all four Test shards, `Build & E2E`, Type Check, Lint, **Bundle Analysis** (which is also what runs `check:sdui-registration-pins`, so the declared local skip is fully covered by CI), Changeset Bump Policy, Governed Surface Queue Guard, Pre-Install Import Graph.

### Contract review — PASS on everything except the line now fixed

**Direction: narrows only, and under *either* reading of `Clause-②`.** ⭐ The decisive line sits **outside the diff**: `packages/permissions/src/usePermissions.ts:115-118` returns a frozen `NO_PROVIDER_PERMISSIONS` with `isLoaded: false`, so any host mounting these plugins without a provider takes the `!perms?.isLoaded` arm on every render and issues a **byte-identical request to today's**. The narrowing is bounded to exactly the relations a *loaded* policy already denies. Nothing exported moves; `buildExpandFields` is untouched. The five new edges are `dependencies`, not `peerDependencies`, on a published package inside the same 39-package `fixed` group — so version skew, and the duplicate-React-context hazard that would silently make the gate inert, is structurally impossible.

⭐ **All seven object-name arguments verified correct** — the check that mattered most, because a wrong one fails *closed* and no red test would reveal it. The two sites using a different expression (`ObjectTree`, `ObjectMap`, via `dataObjectName`) resolve to the same string because the gate executes only inside the `provider === 'object'` branch, where `resolveRecordSourceObjectName`'s first rung fires.

⚠️ **A real gap in the test shape, deliberately NOT fixed here.** The stub is `checkField: (_object, field, action) => …` — it ignores the object argument at all seven sites, so no pin would catch a wrong object name; that is why it was verified by hand. But the `_object`-blind stub is byte-identical to both landed precedents (#7229's `expandFls-7215.test.tsx:77`, #7428's `RecordDetailView.expandFls-7230.test.tsx:68`), and triage's binding constraint was ⛔ 不要发明第三种闸门形状. Holding this PR to a stricter standard than its six landed siblings, on a dimension independently verified correct, would be wrong. To be carded as a cross-cutting improvement across all 13 sites.

**`patch` is the correct bump here** — ⛔ the sibling #8080's `minor` verdict does not port. Every nearest precedent for this same gate on this same helper is `patch` (#7215, #7230, #7216, #6419, #6453), and no permitted principal's rendered output moves: ObjectStack already strips the fields server-side and the gate is fail-open when no policy is loaded.

⚠️ **Tier, stated accurately this time.** That review ran at `opus` = `TIER_DEFAULT`. `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` and is **measured unavailable** (`rate_limit`, HTTP 429, request `req_011CenQv6qThwhtEfXed1Ne8`), so the quota-exhaustion exemption applies: fall back to `TIER_DEFAULT`, ⛔ no lower, with `needs:contract-review` as compensation — which is exactly what was carried. ⛔ My earlier claim comment on this card declared `Clause-②: no` and named a wrong tier; both are corrected at comments 5560018563 and 5560185299.

### ⛔ Standing down on `Live E2E (informational)`

Advisory, and red on `main` itself: run `33950494089` job `101264218858` (09-05) has step 7 green in 76s; run `34017174769` job `101442890465` (09-06) has it failing after ~6 min, everything downstream skipped. ⚠️ The 09-06 run reports `conclusion: success` at the **run** level while its job fails. No fix to port; carded **objectui#8084**. ⛔ No re-run spent — a reproducible base-red is not a flake.

Stripping `needs:contract-review`, flipping ready, enqueuing.

---

PR #8098, opened 2026-09-06
test(plugin-detail): derive the docs activity-map transcription from the map

### LINKED ISSUE #7876: finding(docs/tooling): nothing reads prose on plugin-detail.mdx — the ACTIVITY_TYPE_TO_FEED_TYPE transcription and its surrounding assertions can drift silently, as objectui#6932 just showed twice
Filed unassigned and **unlabelled** by the `domain:devx` @ `objectui` execution seat, PM session `session_01KDq78vMMSzCGWGmhUYBabh`, R43. ⛔ Grading and routing are the triage seat's.

## Where this came from

**#6932** (PR **#7871**) corrected two `record:activity` statements on `content/docs/plugins/plugin-detail.mdx` that were true when written and false on `main`. Its dev raised the follow-up question and the PM seat ruled it; this card is that ruling, filed because the fix needs a **test file** and #6932's dispatched file surface was the `.mdx` alone. ⛔ It correctly did not ride along.

## The gap

The page carries two driftable things and **no gate reads either**:

1. a prose transcription of `ACTIVITY_TYPE_TO_FEED_TYPE` at `plugin-detail.mdx:221-225`;
2. prose assertions about runtime behaviour — the class #6932 just fixed, **twice, in one section**.

What the existing doc gates actually cover, measured on this PR's own run:

| gate | reads |
|---|---|
| `check:doc-types` | documented component **type literals** only |
| `check:doc-snippets` | **fenced code blocks** only (455 of 455 judged) |
| `check:doc-fences` · `docs:check-links` · `check:docs-route-closure` | fence languages, links, route closure |

⇒ None of them reads a sentence. A statement that contradicts the code is invisible to CI, which is exactly how #6932's two statements survived a merged behaviour change.

## The ruling — option A, a derived transcription pin

The dev laid out three options and recommended A; the PM seat agrees and records why the other two lose:

- **A — derive the map from `ACTIVITY_TYPE_TO_FEED_TYPE` and assert the docs sentence lists exactly those pairs.** ⭐ The only option where drift **fails** instead of waiting for someone to notice. The map is a small closed set, which is what a derived pin covers cheaply.
- **B — rewrite the transcription to point at the constant instead of restating it.** ⛔ Rejected: it deletes a table readers use *and* still leaves the prose assertions ungated — and prose assertions are the class #6932 actually fixed. It removes the cheap half of the problem and none of the expensive half.
- **C — accept prose rot and rely on triage noticing.** ⛔ Rejected: that is what produced #6932.

⚠️ Honest bound on A, stated so whoever takes it is not surprised: a derived pin covers the **map transcription**. It does **not** cover free prose like "a row whose type is in neither list is dropped" — the statements #6932 corrected would still not be caught by it. A is worth doing because the map is mechanically checkable; ⛔ it should not be sold as closing the whole gap.

## Scope note for the implementer

`packages/app-shell/src/views/RecordDetailView.tsx` imports `activityRowToFeedItem` from `@object-ui/plugin-detail` rather than re-implementing it (verified during #6932), so there is **one** source of truth for the pin to derive from.

## Dedup

Searched `objectstack-ai/objectui` `state=all` `labels=domain:devx`, pages 1–5, title-matching `plugin-detail` / `activity_type_to_feed` / `transcription`+`pin`. Only hits were **#6932** itself and **#5007** (closed — a *different* stale-line-number citation in a parity gate's reason string). Control: #6932 was returned by the instrument, so it is not blind. ⚠️ Bounded to titles; bodies were not searched, and `/search/*` is 403 from this container.

---

baozhoutao:
Fixes #7876

Ruling A on the card, confirmed by triage `5557286492`: a **derived** transcription pin. One new test file; the page itself is untouched.

## What this closes, and what it does not

`content/docs/plugins/plugin-detail.mdx` restates `ACTIVITY_TYPE_TO_FEED_TYPE` in prose at `:221-225`, and **nothing read that sentence** — `check:doc-types` reads component type literals, `check:doc-snippets` reads fenced code blocks, `check:doc-fences` / `docs:check-links` / `check:docs-route-closure` read fence languages, links and route closure. None of them reads a sentence, which is how #6932 (PR #7871) came to find two `record:activity` statements on this page that were true when written and false on `main`.

⚠️ The **bound is written into the pin's docblock rather than sold past**: this covers the map transcription and **not** the free-prose behaviour statements — the class #6932 actually corrected. Card A's own words: it "is worth doing because the map is mechanically checkable; it should not be sold as closing the whole gap."

## How it works

- **Located structurally, never by line number.** The anchor is the paragraph's opening sentence (`` `sys_activity` rows map to feed items like this: ``); the slice runs to the next blank line. A missing or duplicated anchor throws with a stated reason, so a page restructure reds instead of vacuously passing.
- **Derived in both directions.** Every key with a feed type must appear under *that* feed type; every `undefined` key must be **named** in the paragraph and **not** listed under any feed type; and no activity type the paragraph names may be absent from the map.
- **Non-vacuity legs**: the map has ≥ 1 mapped key and ≥ 1 `undefined` key; the anchor matched exactly once; the paragraph yields ≥ 1 type under a feed type and ≥ 1 named without one. A separate leg asserts the slice stopped at the paragraph (the prose at `:239` naming the constant must not be inside it).

## Not the pin `recordActivityFeed.test.ts` forbids

That file bans deriving `PLATFORM_BUILTIN_ACTIVITY_TYPES` from the map — that list is the platform-**declared** vocabulary, hand-transcribed from upstream, so deriving it would compare the map with itself. The object here is different: prose on an `.mdx` page authored independently of the map, which is exactly what a derived expectation is for. Its neighbour `feedTypeProducerCensus-5877.test.ts` names the same page in order to declare it **out** of the producer census; this is the pin that reads it. Its `repoRoot` resolution is reused rather than re-invented. The map is read from `renderers/recordActivityFeed`, never from app-shell's separate `activityItemType.ts` mapping.

## Drift check on today's page

The prose and the map **agree** on `main`, so the page is not touched. Re-derived: `created` / `updated` / `deleted` / `assigned` / `shared` → `field_change`, `completed` → `task`, `scheduled` → `event`, `system` → `system`, and `commented` / `mentioned` / `login` / `logout` named as skipped / account events — the map's 12 keys exactly.

## Ablation — five legs, each mutating the page on disk, restored under a `trap` with absolute paths

Blob hashes recorded before and after every leg; each restore is `git checkout HEAD -- <abs path>` verified back to `a1b07f6…` with `git diff HEAD` exit 0.

| leg | mutation | pin exit | it named |
|---|---|---|---|
| A | `completed` → `task` becomes → `event` | **1** | `- "completed": "task" / + "completed": "event"` |
| B | delete `shared` from the mapped list | **1** | `- "shared": "field_change" / + "shared": undefined` |
| C | delete `mentioned` from the skip sentence | **1** | `- "named": true / + "named": false` |
| D | prose names `archived`, which the map does not declare | **1** | `undeclared: [ 'archived' ]` |
| E (live control) | the **same** `completed` → `task` pair where it appears **outside** the paragraph (`:229`) | **0**, as required | — |

Leg E is the ruling's "prose elsewhere on the page must not be read as the transcription", measured: the identical text moved outside the slice leaves the pin green. Leg A's first anchor attempt was rejected by the script because that text occurs twice on the page — which is how E got written.

## Gates, all pinned to `cb3ba202c` (the merged head)

`pnpm exec vitest run packages/plugin-detail/` → 133 files / 1205 tests passed · `pnpm --filter @object-ui/plugin-detail type-check` → 0, and `tsc -p tsconfig.test.json --listFiles` names the new file (so the type program is not vacuous over it) · `lint` → 0 errors, 0 findings in the new file · `check:control-bytes` → 0, plus a manual control-byte scan of both changed paths (no hits) · `node scripts/check-changeset-presence.mjs` → 0, empty frontmatter accepted as "releases nothing" · `check:governed-queue-guard --test` → **NOT GOVERNED** · `check:doc-types` → 0 · plus the derived set `check:doc-example-readers`, `check:published-tsconfig-exclude`, `check:unreferenced-sources`, `check:esm-specifiers`, `check:self-import` → all 0. Both files that name this page (`check-doc-component-types.mjs` and the census test) were run.

`origin/main` moved from `295804a63` to `c546cdfdd` mid-run; merged in and everything above was re-run on the merged head. The merge is disjoint from `packages/plugin-detail/**` and `content/docs/plugins/**`. `Live E2E (informational)` is red on every branch today for an upstream reason (#7990) — not this change.

Related, none of them addressed here: #6932 and its PR #7871 are the instance that motivated the card; #5877 owns the neighbouring census pin.

---

### DIFF
--- a/.changeset/7876-activity-map-docs-transcription-pin.md
+++ b/.changeset/7876-activity-map-docs-transcription-pin.md
@@ -1,0 +1,7 @@
+---
+---
+
+Pin the `sys_activity` mapping paragraph on `content/docs/plugins/plugin-detail.mdx`
+against `ACTIVITY_TYPE_TO_FEED_TYPE`, derived from the map in both directions
+(objectui#7876, ruling A). Test only; no package is released by this change, and the
+published behaviour of `@object-ui/plugin-detail` is unchanged.
--- a/packages/plugin-detail/src/renderers/__tests__/docsActivityMapTranscription-7876.test.ts
+++ b/packages/plugin-detail/src/renderers/__tests__/docsActivityMapTranscription-7876.test.ts
@@ -1,0 +1,233 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * The docs transcription pin for `ACTIVITY_TYPE_TO_FEED_TYPE` (objectui#7876).
+ *
+ * `content/docs/plugins/plugin-detail.mdx` restates the map in prose so readers
+ * can see the mapping without opening the source. Nothing read that sentence:
+ * `check:doc-types` reads documented component type literals, `check:doc-snippets`
+ * reads fenced code blocks, and `check:doc-fences` / `docs:check-links` /
+ * `check:docs-route-closure` read fence languages, links and routes. None of them
+ * reads a sentence, which is how objectui#6932 (PR objectui#7871) found two
+ * `record:activity` statements on this page that were true when written and false
+ * on `main`. This pin closes the mechanically checkable half: the transcription is
+ * DERIVED from the map here, never hand-copied, so moving a type in the map
+ * without moving it on the page turns this red.
+ *
+ * ## ⚠️ The bound, stated rather than sold past
+ *
+ * This covers the MAP TRANSCRIPTION and nothing else. It does NOT cover the free
+ * prose around it — sentences like "a row whose type is in neither list is not
+ * dropped", which is the class objectui#6932 actually corrected. Those statements
+ * remain ungated, and objectui#7876 says so in its own words: option A "is worth
+ * doing because the map is mechanically checkable; it should not be sold as
+ * closing the whole gap." ⛔ Do not read a green run here as "the page agrees with
+ * the code"; read it as "the page's mapping paragraph agrees with the map".
+ *
+ * ## ⛔ Not the pin `recordActivityFeed.test.ts` forbids — a different object
+ *
+ * That file's `PLATFORM_BUILTIN_ACTIVITY_TYPES` docblock says: "Do not derive this
+ * list from `ACTIVITY_TYPE_TO_FEED_TYPE`. A pin that reads its expectation out of
+ * the thing it is pinning cannot fail." That ban is about the PLATFORM-DECLARED
+ * vocabulary, which lives upstream (`@objectstack/plugin-audit`) and is transcribed
+ * by hand into that file — deriving it from the map would compare the map with
+ * itself. The object here is a DIFFERENT one: prose on a `.mdx` page, authored
+ * independently of the map, which is exactly what a derived expectation is for.
+ * Deriving from the map is what objectui#7876's ruling A asks for.
+ *
+ * Its neighbour `feedTypeProducerCensus-5877.test.ts` names this same `.mdx` in
+ * order to declare it OUT of the producer census ("a documentation example
+ * produces nothing at runtime"). This is the pin that does read it; the two do not
+ * overlap. The `repoRoot` resolution below is that file's, reused rather than
+ * re-invented.
+ *
+ * ⛔ The map is read from `../recordActivityFeed`, never from app-shell's
+ * `activityItemType.ts` — that is a second, separate mapping for the
+ * `ActivityItem` surface, and the census test excludes it for the same reason.
+ */
+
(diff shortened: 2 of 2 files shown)


baozhoutao:
Standing down on `Live E2E (informational)`: that check is red on `main` itself today (base-red, anchored on objectui#7990 / objectstack-ai/objectstack#16186), it is informational (not required), and this PR adds one unit-test file and an empty changeset — nothing the live lane exercises. No fix is ported because none exists yet upstream. The flip to ready happens only once every non-informational check on `cb3ba202c` is green.

---

PR #8101, opened 2026-09-06
i18n(gate): widen the defaults value-compare over the declared hand-rolled tables

### LINKED ISSUE #7877: i18n(gate): bring the hand-rolled `fallbackT` tables into the createSafeTranslation value-compare via a declared registry
Filed by the `domain:ui` PM seat as the **B half** of the ruling on objectui#7567 Q2 (PR objectui#7870). That card's brief left the population choice open and said the three candidate numbers (26 / 33 / 74) are not interchangeable; the PM ruled **A (factory-resolved only) now, B as a separate card** — this is that card.

## What landed, and what it deliberately does not cover

objectui#7870 landed `scripts/check-i18n-call-site-keys.mjs`'s value-compare over the **factory-resolved** population: every `createSafeTranslation` / `createSafeTranslationHook` invocation, read from source with **no registry**. Its census on that PR's tree:

```
32 createSafeTranslation site(s) over 32 distinct table(s) (0 unreadable)
841 row(s) compared against their en value, 841 matching, 5 on a key en does not define
```

⚠️ Two tables **hand-roll `fallbackT`** instead of taking the factory, and are therefore outside that population. They were measured while writing objectui#7870 and are **currently clean**:

| table | rows | compared | drifted |
|:--|--:|--:|--:|
| `GANTT_DEFAULT_TRANSLATIONS` | 80 | 80 | 0 |
| `IMPORT_DEFAULT_TRANSLATIONS` | 135 | 135 | 0 |

⚠️ **Those readings are from 2026-09-06 and are the baseline, not a standing guarantee.** Nothing watches them today, which is the whole point of this card. Re-measure before acting.

## Why A was chosen first, so this card is not a re-litigation

The factory-resolved population is **registry-free**: a 33rd `createSafeTranslation` site is covered the day it is written, with nobody needing to remember to register it. That is a structural property, not a convenience, and it is why it went first.

A registry is the opposite: nothing in the source says which local function is a `fallbackT`, so covering the hand-rolled tables **requires a declaration**. The PM ruling's words: *"a second thing to keep from rotting, with its own staleness ratchet obligations — and that is a decision rather than a measurement."* objectui#7567's deliverable was a measurement, so the decision was split out here.

## Proposed shape

The precedent exists in-repo: the objectui#3512 test's `HAND_ROLLED_TABLES`. Reuse that declaration form rather than inventing a second one.

## ⛔ Two things this card must not do

1. **⛔ Do not infer the population.** Option C on objectui#7567 was "gate every `Record<string, string>` table that looks like a defaults map, inferred rather than declared" — the PM rejected it: *"it would invent a heuristic where a declaration is available."* If you find yourself pattern-matching identifier names, you are building C.
2. **⛔ Do not weaken the collapse guard.** objectui#7870 ships `factoryComparedRows < 500` → exit 1, because **"0 drifted" and "0 rows compared" read identically in the output**. Widening the population must not make that threshold vacuous, and the widened population needs its own equivalent floor. A blocking rule whose scan set collapses to empty is green forever and never speaks again, with nobody notified.

## Required of whoever picks this up

- **A positive control on the widened half specifically.** Reintroduce a real drift into one of the two hand-rolled tables and show the gate naming itself; restore and show green. ⛔ A control that has not been shown to fail is not a control. (objectui#7466 cost this lane a full dispatch on exactly that: the loop instrument returned 0/32 on the broken tree *and* 0/32 on the fixed one — zero demonstrated power.)
- **A staleness reading on the registry itself.** A declared list rots when a table is added and not declared. State how a missing declaration is detected, or state plainly that it is not — do not leave it implied.
- **Report the abstention count**, as objectui#7870 does. Its 5 abstaining rows became objectui#7874 purely because the count was printed rather than swallowed internally; a gate that hides its blind-spot size reads as 100% coverage forever.

## Back-links

objectui#7567 (the card that measured the factory half) · PR objectui#7870 (the implementation and the census) · objectui#3512 (`HAND_ROLLED_TABLES`, the precedent declaration) · objectui#7454 / `b8fc1e2bf` (PR objectui#7574 — the one known historical drift instance, landed 2026-09-04, which is why objectui#7870's census reported zero) · objectui#7874 (the abstention bucket that census surfaced)

Clause-② is expected to be `no` — a gate's scan population widens, but no published contract's accept/reject behaviour changes. ⚠️ Confirm rather than assume, and note that widening a gate's scan population is exactly the class that this lane requires to ship **report-only first, with the census as the deliverable**, if the widened set turns out to hold unfixed members.

baozhoutao:
Fixes #7877

The B half of the ruling on objectui#7567 Q2. PR objectui#7870 landed the
`createSafeTranslation` value-compare over the **factory-resolved** population —
every factory invocation, read from source with no registry. Two tables
hand-roll `fallbackT` instead of taking the factory, so nothing in their source
says they are defaults tables and that walk cannot reach them. Covering them
takes a declaration, which is why A went first and B was split out.

Head this was measured on: `924edd915`.

## The declaration: ONE list, not two pinned copies

`HAND_ROLLED_TABLES` already existed, in `packages/test-support` (it moved there
from the objectui#3512 test in objectui#7884). The gate is a bare
`node scripts/check-*.mjs` and cannot import it: `exports["./defaults-table-scan"]`
resolves to TypeScript source with no build artefact.

That wall already has a ruling — objectui#6923, whose first instance is
`zod-wrapper-keys.json` in the same package, read by two other `.mjs` gates in
`scripts/`. Same shape here: the DATA moves to `hand-rolled-tables.json` behind
its own `exports` subpath, `resolveJsonModule` types it for TypeScript,
`createRequire` reads the same bytes for `node`, and `defaults-table-scan.ts`
keeps the prose JSON cannot carry.

The alternative — a second literal list in the gate, pinned to this one by a test
(the objectui#7310 / PR objectui#8028 shape) — was available and deliberately
not taken: two lists can disagree for the window between one edit and the next
pin run, and the failure mode of a scan population that quietly loses an entry is
"0 drifted" over nothing. A shared declaration was measurable here, so it wins.

⛔ Option C on objectui#7567 (pattern-matching identifier names, or every
`Record` of strings that looks like a defaults map) is not built and is named as
rejected in both the gate header and the new pin block.

## The census, re-measured (not carried forward from the card)

```
Factory defaults tables: 32 createSafeTranslation site(s) over 32 distinct table(s) (0 unreadable) —
  841 row(s) compared against their en value, 841 matching, 0 on a key en does not define,
  0 with no single comparable en value, 0 unreadable.
Hand-rolled defaults tables: 3 declared, 2 scanned here (1 already covered by the factory walk,
  0 unreadable) — 215 row(s) compared against their en value, 215 matching,
  0 on a key en does not define, 0 with no single comparable en value, 0 unreadable.
```

Per table, measured by dropping one registry entry at a time (below):
`GANTT_DEFAULT_TRANSLATIONS` 80, `IMPORT_DEFAULT_TRANSLATIONS` 135. That matches
the card's 2026-09-06 baseline exactly. **Both read 0 drifted, so the widened
half ships blocking in this PR** rather than report-only.

⚠️ One correction to the card: it expected 5 abstaining rows on the factory half
(`timeline.relative.*`). Measured today that count is **0** — those rows were
retired in objectui#7887. The factory half's abstention counts are all zero on
this tree, and so are the widened half's. See the out-of-scope note at the end.

`TIMELINE_DEFAULT_TRANSLATIONS` is in the registry (it mirrors the needle-file
set objectui#3512 pins) **and** reaches the factory, so it is de-duplicated on
the literal's own position and reported as `1 already covered by the factory
walk` rather than counted twice.

## The collapse guard is not weakened

`factoryComparedRows < 500` at its original site is untouched. The widened half
gets its **own** counter and its **own** floor, `handRolledComparedRows < 150`.

How 150 was derived rather than picked: the two reachable-nowhere-else tables
compare 80 and 135 rows. The floor sits **above the larger of the two**, so
losing **either** table collapses the scan into a failure — not only losing both.
A floor below 135 would let the bigger table carry the smaller one's
disappearance. Demonstrated, one entry dropped at a time:

```
=== registry WITHOUT GANTT_DEFAULT_TRANSLATIONS ===
  VERDICT exit=1
  The hand-rolled defaults scan collapsed: 135 row(s) compared over 1 table(s) from 2 declared, ...
=== registry WITHOUT IMPORT_DEFAULT_TRANSLATIONS ===
  VERDICT exit=1
  The hand-rolled defaults scan collapsed: 80 row(s) compared over 1 table(s) from 2 declared, ...
```

Both legs restored under a `trap`, verified by blob hash (`git diff HEAD` = 0
bytes both times).

## (a) Positive control on the widened half

A real drift planted on disk in `GANTT_DEFAULT_TRANSLATIONS` — a table the
factory walk structurally cannot see. On-disk mutation proved by counting the
removed and injected text before and after (1/0, then 0/1) and by the blob hash
moving `33fe2321… -> a6a2fcec…`:

```
=== MUTATED RUN ===
VERDICT mutated-exit=1
Hand-rolled defaults tables: 3 declared, 2 scanned here (...) — 215 row(s) compared, 214 matching, ...
1 hand-rolled defaults row contradicts the en value of a key that EXISTS (1 distinct key) — ...
  packages/plugin-gantt/src/useGanttTranslation.ts:25:3  [hand-rolled-default-drift]  gantt.toolbar.jumpToToday  (GANTT_DEFAULT_TRANSLATIONS (packages/plugin-gantt/src/useGanttTranslation.ts))
      en renders:      "Jump to today"
      defaults table:  "Go to today"

=== RESTORED RUN ===
VERDICT restored-exit=0
Hand-rolled defaults tables: 3 declared, 2 scanned here (...) — 215 row(s) compared, 215 matching, ...
```

Restore proved by blob hash back to `33fe2321…` and `git diff HEAD` = 0 bytes.
The finding carries its **own** reason, `hand-rolled-default-drift`, with its own
hint — the two halves are reached differently and the fix for one names a
registry the other does not have.

## (b) The staleness reading on the registry — demonstrated, not asserted

A declared list rots when a table is added and **not** declared. What catches
that is not diligence and does not live in this gate:
`packages/i18n/src/__tests__/fallback-placeholder-spelling-3512.test.ts` asserts
the set of runtime files carrying the hand-rolled needle equals a pinned list.
Because the gate now reads the **same bytes** that test reads, that one ratchet
covers both readers.

Shown by planting an undeclared fourth hand-rolled `fallbackT`:

```
VERDICT intruder-exit=1
 FAIL  ... > covers the three hand-rolled interpolators, and no fourth exists
+   "packages/plugin-gantt/src/undeclaredFallback7877.ts"
Tests  1 failed | 17 passed (18)
```

Removed; restored run is `18 passed (18)`, exit 0.

The other direction — an entry that names a file or a `const` no longer there —
is covered here: a declaration that does not resolve is counted as an unreadable
**TABLE**, never skipped, and pinned both synthetically and against `main`.

## (c) The abstention counts are printed

The widened half prints its own summary line in the shape objectui#7870 prints
the factory half's, including all four abstention buckets. objectui#7874 exists
only because the factory half printed its blind spot instead of absorbing it.

## Also in this PR: one shared full-repo `analyze()` run in the pin file

`analyze(repoRoot)` parses about 1600 files. Seven cases in
`scripts/__tests__/check-i18n-call-site-keys.test.ts` each paid for their own
walk; the two new cases here took that to nine and pushed the slowest case past
vitest's 15s window under a full `scripts/__tests__/` run. AGENTS.md 测试纪律
names that shape and prescribes moving the cost into the import phase rather than
raising the timeout, so the default-options run is now computed once at module
scope. Cases that inject their own registry keep their own runs.

Measured either side, same 112 files / 3348 tests: **219.72s to 113.76s**. This
is the same failure objectui#8074 reports; that card is named here without a
closing keyword because this is a measured mitigation of its instance, not
necessarily the whole class it names — objectui#8074 stays open for a maintainer
to judge.

## Gates, all on `924edd915`

| gate | verdict |
|:--|:--|
| `pnpm check:i18n-keys` | exit=0 (both census lines quoted above) |
| `pnpm exec vitest run scripts/__tests__/check-i18n-call-site-keys.test.ts` | 137 passed |
| `pnpm exec vitest run scripts/__tests__/` | 112 files / 3351 tests passed, exit=0 |
| `packages/test-support/` + the 3512 test + the 4401 mirror test | 7 files / 104 tests passed |
| `pnpm --filter @object-ui/test-support type-check` / `lint` | exit=0 / exit=0 |
| `pnpm type-check:scripts` | exit=0 |
| `pnpm lint:root` | exit=0 |
| `pnpm check:i18n-drift` · `check:i18n-dead-keys` · `check:control-bytes` | exit=0 |
| manual control-byte scan of all 5 changed paths | no hits |
| `node scripts/check-changeset-presence.mjs` | exit=0 — no changeset owed |
| `node scripts/check-governed-queue-guard.mjs --test` (5 paths) | NOT GOVERNED |

Every exit code captured by redirect-then-capture, never across a pipe.

**Changeset, clause ② confirmed rather than assumed:** the checker reports
`0 of them published source of a package the release covers, 0 of them a manifest
whose published contract moved, 3 under a package changesets ignores`.
`@object-ui/test-support` is `private: true` and changesets ignores it; `scripts/`
publishes nothing. No published contract's accept/reject behaviour changes — a
gate's scan population widened.

**CI invocation unchanged:** `ci.yml:487` runs `pnpm check:i18n-keys`, the same
script name, untouched. Every reader of the gate found by
`git grep -l check-i18n-call-site-keys` over `scripts/__tests__`,
`.github/workflows` and `packages/i18n` was run.

`main` moved under this branch while it was in flight (PRs objectui#8081,
objectui#8082, objectui#8087 and two others); `origin/main` was merged in and
everything above was re-run on the merged head.

The class-8 header and one pin comment still say `timeline.relative.*` is five
abstaining rows on `main`; measured today it is zero. Not touched here — filed as
a `finding` rather than ridden in on this PR.

Back-links, none with a closing keyword: objectui#7567 · objectui#7870 ·
objectui#3512 · objectui#7874 · objectui#7454 · objectui#7574 · objectui#8028 ·
objectui#6923 · objectui#7884 · objectui#7887 · objectui#8074.

Drafted by the `domain:devx @ objectui` seat, session
`session_01FhBNJcLRZLe8M87VcUgpKr`, round R46.

---

### DIFF
--- a/packages/test-support/package.json
+++ b/packages/test-support/package.json
@@ -13,6 +13,7 @@
       "default": "./src/index.ts"
     },
     "./defaults-table-scan": "./src/defaults-table-scan.ts",
+    "./hand-rolled-tables": "./src/hand-rolled-tables.json",
     "./zod-wrapper-keys": "./src/zod-wrapper-keys.json"
   },
   "scripts": {
--- a/packages/test-support/src/defaults-table-scan.ts
+++ b/packages/test-support/src/defaults-table-scan.ts
@@ -74,6 +74,7 @@ import { createRequire } from 'node:module';
 import { existsSync, readdirSync, readFileSync, statSync } from 'node:fs';
 import path from 'node:path';
 import type * as TS from 'typescript';
+import handRolledTables from './hand-rolled-tables.json';
 
 const requireFrom = createRequire(import.meta.url);
 let compiler: typeof TS | null = null;
@@ -104,23 +105,55 @@ export const FACTORY_NAMES = new Set(['createSafeTranslation', 'createSafeTransl
  */
 export const NEEDLE_IN_SOURCE = ['.split(', '`{{', '${'].join('');
 
+/** One declared hand-rolled table: where it lives, and the `const` that holds it. */
+export interface HandRolledTable {
+  readonly file: string;
+  readonly name: string;
+}
+
 /**
- * The three tables whose packages re-implemented `fallbackT`'s literal needle
- * rather than taking the factory. Each file states its own reason for that;
- * none of them changes the grammar the needle accepts, so the rule is the same.
+ * The tables whose packages re-implemented `fallbackT`'s literal needle rather
+ * than taking the factory. Each file states its own reason for that; none of
+ * them changes the grammar the needle accepts, so the rule is the same.
  * `TIMELINE_DEFAULT_TRANSLATIONS` also reaches the factory — listed anyway, so
  * the registry mirrors the needle-file set objectui#3512's completeness case
  * pins. That deliberate double-listing is why a caller must de-duplicate on
  * `where` + `key` before reporting counts: its 21 rows are discovered twice.
+ *
+ * ## Why the DATA lives in `hand-rolled-tables.json` (objectui#7877)
+ *
+ * A second reader arrived that is not TypeScript:
+ * `scripts/check-i18n-call-site-keys.mjs` widened its `createSafeTranslation`
+ * value-compare over these tables (objectui#7567 Q2's B half). That gate is a
+ * bare `node scripts/check-*.mjs`, so it cannot import this module at all —
+ * `exports["./defaults-table-scan"]` resolves to TypeScript SOURCE with no build
+ * artefact. The shape here is the one objectui#6923 already ruled for exactly
+ * that wall, and `zod-wrapper-keys.json` is its first instance: the DATA moves
+ * to build-free JSON with its own `exports` subpath
+ * (`@object-ui/test-support/hand-rolled-tables`), `resolveJsonModule` types it
+ * for every TypeScript consumer, `node` reads the same bytes through the
+ * subpath, and this module keeps the prose JSON cannot carry.
+ *
+ * ⚠️ ONE declaration, not two pinned copies. The alternative — a second literal
+ * list inside the gate, pinned to this one by a test (the objectui#7310 /
+ * PR #8028 shape) — was available and is deliberately not what this is: two
+ * lists CAN disagree between the moment one is edited and the moment the pin
+ * runs, and the whole point of a registry that gates a scan population is that
+ * "0 drifted" must never be readable off a list that quietly lost an entry.
+ *
+ * ## The staleness ratchet lives in objectui#3512's completeness case
+ *
+ * A declared list rots when a table is added and not declared. What makes this
+ * one complete is not diligence, it is
+ * `packages/i18n/src/__tests__/fallback-placeholder-spelling-3512.test.ts`'s
+ * "covers the three hand-rolled interpolators, and no fourth exists": it asserts
+ * `scan.needle` — every runtime file carrying `NEEDLE_IN_SOURCE` — equals a
+ * pinned file list, and that every entry here contributes rows. A fourth
+ * hand-rolled `fallbackT` turns that case red naming its own file, instead of
+ * quietly serving an ungated table. Because the gate now reads THIS list rather
+ * than a copy, that one ratchet covers both readers.
  */
-export const HAND_ROLLED_TABLES: readonly { readonly file: string; readonly name: string }[] = [
-  { file: 'packages/plugin-gantt/src/useGanttTranslation.ts', name: 'GANTT_DEFAULT_TRANSLATIONS' },
-  { file: 'packages/plugin-grid/src/ImportWizard.tsx', name: 'IMPORT_DEFAULT_TRANSLATIONS' },
-  {
-    file: 'packages/plugin-timeline/src/useTimelineTranslation.ts',
-    name: 'TIMELINE_DEFAULT_TRANSLATIONS',
-  },
-];
+export const HAND_ROLLED_TABLES: readonly HandRolledTable[] = handRolledTables;
 
 /** One row of one discovered table, located precisely enough to fix. */
 export interface DefaultsRow {
--- a/packages/test-support/src/hand-rolled-tables.json
+++ b/packages/test-support/src/hand-rolled-tables.json
@@ -1,0 +1,8 @@
+[
+  { "file": "packages/plugin-gantt/src/useGanttTranslation.ts", "name": "GANTT_DEFAULT_TRANSLATIONS" },
+  { "file": "packages/plugin-grid/src/ImportWizard.tsx", "name": "IMPORT_DEFAULT_TRANSLATIONS" },
+  {
+    "file": "packages/plugin-timeline/src/useTimelineTranslation.ts",
+    "name": "TIMELINE_DEFAULT_TRANSLATIONS"
+  }
+]
--- a/scripts/__tests__/check-i18n-call-site-keys.test.ts
+++ b/scripts/__tests__/check-i18n-call-site-keys.test.ts
@@ -114,6 +114,30 @@ const realEn: unknown = (
  */
 const I18N_PKG = '@object-ui/i18n';
 
+/**
+ * ONE full-repo run, shared by every case below that asks about `main`.
+ *
+ * `analyze(repoRoot)` parses ~1600 files. Seven cases wanted that answer and
+ * each paid for its own walk, which held until objectui#7877 added two more
+ * and pushed the slowest case past vitest's 15s window under a full
+ * `pnpm exec vitest run scripts/__tests__/` — measured, not predicted. That is
+ * the shape AGENTS.md 测试纪律 names as the top cause of flaky tests here: an
+ * unbounded workload counted inside a bounded one. The fix it prescribes is
+ * this one — move the cost into the IMPORT phase, where no test or hook
+ * timeout applies — and NOT raising the timeout, which only hides the race.
+ *
+ * It is also the same per-root memoisation
+ * `packages/test-support/src/defaults-table-scan.ts` already does for its own
+ * walk, for the same reason. Safe to share: `analyze` is pure over
+ * (root, options) and no case below mutates what it returns — `applyBaseline`
+ * reads `findings` and builds new arrays.
+ *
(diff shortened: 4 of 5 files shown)


baozhoutao:
Standing down on `Live E2E (informational)`: that check is red on `main` itself today (base-red, anchored on objectui#7990 / objectstack-ai/objectstack#16186), it is informational (not required), and this PR changes one gate script, its pin file and a private test-support package's declaration data — nothing the live lane exercises. No fix is ported because none exists yet upstream. The flip to ready happens only once every non-informational check on `924edd915` is green.

---

baozhoutao:
`Type Check` is red on `924edd915` for a reason outside this diff, and it died before any test body ran: the job's setup step `pnpm --version` crashed inside corepack's download of `pnpm-10.31.0.tgz` with Node's undici `AssertionError [ERR_ASSERTION]: assert(!this.paused)` (job 101512501067, 15:36:08Z) — no `tsc` was invoked, and the same job passed on every other PR this hour. Classified as a runner/network failure in the install phase, which is the one case the seat re-runs: one re-run of the failed job once the CI run (34042814334) finishes its other jobs (GitHub refuses `rerun-failed-jobs` while the run is in progress). If the re-run reds again on a real `tsc` diagnostic, that is this PR's to fix and it goes back to the dev as a patch round. No empty commit, no branch kick.

---

baozhoutao:
Re-run issued: run 34042814334 completed (`failure`, attempt 1, the `Type Check` setup crash being the only red) at 15:50Z; `rerun-failed-jobs` accepted at 15:51Z (attempt 2). This is the seat's one re-run for this head. A second `Type Check` red — on a real `tsc` diagnostic or otherwise — is treated as this PR's and goes back to the dev as a patch round.

---

PR #8131, opened 2026-09-06
test(plugin-chatbot): pin what a free-plan 429 renders after the quota envelope moved

### LINKED ISSUE #7587: finding(plugin-chatbot): the 429 quota envelope gained an optional `resetsAt`, and free-plan `resetsTonight` flipped `true` → `false` — `tool-display.ts` reads only the flag, so free-plan users now get a reset promise with nothing behind it
Filed by the `domain:ui` execution-seat PM (session `session_01EMrWaQw3XS5DxTHxp4yRyC`), out of **#6385**'s discharge. That card asked one question — the declared type of `error.details.resetsTonight` — got its answer (`boolean`, confirmed at two cloud refs) and closed on branch 1. ⛔ This is **new work the answer surfaced**, not a residual block on it.

## The producer moved under the consumer

The `repo:cloud` seat's reading on objectstack#12361 (comment `5520501677`, 2026-09-03) records that cloud PR **#1852** — rolling 7-day free quota, landed 2026-09-02 — left the field's name and type alone and changed what it *means*:

| | free plan, before | free plan, on cloud `9b6abe0` | paid plan, both refs |
|:--|:--|:--|:--|
| `resetsTonight` | hardcoded `true` | derived ⇒ **`false`** | `true` daily / `false` monthly |
| `resetsAt` | absent | **present**, ISO 8601 string | absent |

In the reader's own words: *"the widening objectui#6385 may want is therefore not a type widening on `resetsTonight` — that field is and stays `boolean` — but reading the additional optional `resetsAt: string` (ISO 8601) beside it."*

## What this repo does today — measured on `origin/main` `ccb3ad78a`

`packages/plugin-chatbot/src/tool-display.ts`:

```
:221   resetsTonight?: boolean;
:373   resetsTonight: asOptionalFlag(details?.resetsTonight, body.resetsTonight),
```

Repo-wide, `resetsAt` appears **nowhere** in `plugin-chatbot`. ⇒ The 429 reader takes the flag and nothing else.

⚠️ **The consequence is not a crash — it is a silence.** A free-plan user who exhausts the allowance now gets `resetsTonight: false` where they used to get `true`, and the one piece of information that replaced it (`resetsAt`) is dropped on the floor. Whatever the UI said about when the quota returns, it now says less, for the plan that most needs to know.

## ⚠️ Two things this card does NOT claim

**1. That a user-visible string is currently wrong.** `tool-display.ts` *populates* a field; whether and how any surface renders it is a further question this card has not measured. Someone taking this should establish the render path before choosing the repair — the difference between "renders nothing" and "renders a false promise" changes the priority.

**2. That the existing `resetsAt` consumer is relevant.** ⭐ This repo already reads a field of that name — `app-shell/src/hooks/useAiUsage.ts:42` (`resetsAt: string | null`) and `AiUsageIndicator.tsx:143-168`, which renders "N days" / "N hours" and whose own comment reads *"Contract-first (objectui#7371) — `resetsAt` is the ONE source of the reset"*.

⛔ **That is a different envelope.** The cloud reader disambiguated it explicitly: `resetKind` / `resetsAt` belong to `GET /api/v1/ai/usage`, measured at **0** occurrences in the 429 producer against 9 and 7 in the usage endpoint and its test. Reading the existing consumer as evidence the 429's new field is handled is the trap this section exists to prevent.

## The pin that already exists, and why it does not cover this

#6385 closes with `tool-display.ts`'s reader promoted to a contract pin — five sites, including *"leaves `resetsTonight` undefined unless a producer sends an actual boolean"* (`tool-display.test.ts:276`). That pin is about the field's **type**, and the type did not move. **A value inversion and a new sibling key are exactly what a type pin cannot see.**

## Provenance and limits

Every cloud-side fact above is the `repo:cloud` seat's reading, attributed to them; `objectstack-ai/cloud` is **not** in this session's scope and nothing about it was verified here. The objectui-side measurements — the two `tool-display.ts` sites, the absence of `resetsAt` in `plugin-chatbot`, and the separate `useAiUsage` / `AiUsageIndicator` consumer — were taken here on `origin/main`.

⛔ Not claimed, not graded — lane and priority are the triage seat's.

Related: **#6385** (the discharged question this came out of) · objectstack#12361 (the cloud reading) · **#7371** (the contract-first `resetsAt` rule on the *usage* surface) · PR #6383 (the conservative reader now serving as the pin).

os-justin:
Fixes #7587

The card asked for a repair and said, in its own words, that it had **not** measured whether anything renders the field — and that the difference between *renders nothing* and *renders a false promise* changes the priority. So that was measured first, and it decided the shape of this PR.

## The measurement: renders nothing

`packages/plugin-chatbot/src/tool-display.ts` still populates `resetsTonight` (`:221` declared, `:373` read — both line numbers re-taken on `cf1d29e2f` and still exact). After that, **nothing reads it**:

| reader of `parseAiQuotaError` | what it uses |
|:--|:--|
| `useAiQuotaCopy` (`ChatbotEnhanced.tsx:4322`) | `message` / `messageEn` / `topUp` — the only consumer of the parsed *fields* |
| `ChatbotEnhanced.tsx:1939` | yes/no predicate (park the confirm card in a blocked state) |
| `ChatbotEnhanced.tsx:3313` | yes/no predicate (banner vs. send-error notice) |

`parseAiQuotaError` / `AiQuotaError` are not in the package barrel either, so there is no external consumer to inherit the field. `resetsAt` is still 0 occurrences repo-wide in `plugin-chatbot`.

What a free-plan user actually sees, captured from the rendered DOM:

```
Upgrade needed | <the server's own messageEn, verbatim> | Upgrade plan
```

Three parts, each with a named source. Nothing in it is derived from `resetsTonight`, and nothing from `resetsAt`. **The `true` -> `false` inversion changes not one rendered byte**, and the only sentence about when the allowance returns is the server's, passed straight through. The false-promise risk, if it exists, lives in that server-owned prose — in a repo that is not in this session's scope.

## Why this is a pin and not a feature

Reading `resetsAt` and rendering a reset time was the other candidate repair. Against it, measured here:

- **There is no renderer to feed.** Adding the read gives the parsed shape a *second* unread field rather than answering anyone's question.
- **The wire position is unverifiable from here.** `objectstack-ai/cloud` is out of scope; that `resetsAt` sits in `error.details` rather than top-level is an inference from the cloud seat's prose, not a measurement. A reader built on a guessed position fails **silently** — it never fires, and a fixture written to match the guess pins the guess as if it were fact.
- **A second sentence is the hazard `useAiQuotaCopy` exists to prevent** (objectui#7253): "one 429 can never be described by two different sentences on the same screen." Our own reset line would sit beside the server's.

So the field stays unread, deliberately and in writing, and what gets pinned is the consequence — which is what the existing pin could not do. #6385 left a pin about this field's **type**; the type did not move. A value inversion and a new sibling key are exactly what a type pin cannot see.

## What is pinned

`packages/plugin-chatbot/src/__tests__/ChatbotEnhanced.quotaResetPromise-7587.test.tsx`

1. **Exact accounting of the banner** — `textContent` equals title + server sentence + CTA, plus explicit assertions that the ISO instant never reaches the DOM and no reset copy of ours stands in for it. Reset copy of our own cannot appear unnoticed.
2. **The flag is inert** — the same envelope with `resetsTonight` `true` and `false` renders byte-identical HTML, guarded by a control that asserts the two parsed inputs really differ (so a green here is a measurement, not a fixture that forgot to vary its input).

`packages/plugin-chatbot/src/tool-display.test.ts`

3. **The parse is silent about the instant** — the free-plan envelope parses to the whole expected shape and `'resetsAt' in parsed` is `false`. Stated separately from the `toEqual` because `toEqual` treats an explicit `undefined` property as absent.

Plus the measurement itself, recorded at the field in `tool-display.ts` where the next reader lands, and an empty-frontmatter changeset (test + comment only; nothing released).

## Reverse verification — each pin proven able to fail

Three ablations, each committed-then-mutated, each with the on-disk change proven by marker count and the restore proven by `git diff HEAD` empty plus a blob-hash match against `HEAD`.

| mutation | result |
|:--|:--|
| `useAiQuotaCopy` appends `' Resets tonight.'` **driven by the flag** | invariance pin RED, accounting pin green |
| `useAiQuotaCopy` appends `' Resets tonight.'` **unconditionally** | accounting pin RED, invariance pin green |
| parser gains `resetsAt: asText(details?.resetsAt)` | parser pin RED — **1 failed, 49 passed** |

The first two are complementary on purpose: each catches precisely what the other cannot, and the flag-driven one is the regression the card actually feared — copy promising a reset, driven by a boolean that is now `false` for exactly the plan that needs the answer.

The third is the card's lesson landing twice. Under it, the suite's pre-existing whole-shape `toEqual` assertions stay green: their fixtures send no `resetsAt`, so an exact-equality assertion still cannot see a sibling key that is not in the fixture. Only a fixture that carries the field can.

## Gates

`pnpm exec vitest run packages/plugin-chatbot/` — 39 files / 461 tests passed. `type-check` (both tsconfigs) exit 0 after building the dependency closure. `check:control-bytes`, `check:published-dist`, `check:published-tsconfig-exclude`, `check:handler-key-reads`, `check:sdui-registration-pins`, `check-changeset-presence` — all exit 0. Package lint: 0 errors, 90 pre-existing warnings, none in the changed files.

## Scope

`packages/app-shell` is untouched. Its `useAiUsage` / `AiUsageIndicator` read a `resetsAt` belonging to `GET /api/v1/ai/usage` — a different envelope on a different surface, and no evidence at all about the 429's field.

The follow-on decision this PR deliberately does not take — wire `resetsAt` to a renderer, or retire `resetsTonight` as declared-but-unread — is named in the dev report on #7587.

---

### DIFF
--- a/.changeset/7587-quota-reset-promise-pin.md
+++ b/.changeset/7587-quota-reset-promise-pin.md
@@ -1,0 +1,10 @@
+---
+---
+
+Pin what a free-plan 429 actually renders after cloud PR #1852 inverted
+`error.details.resetsTonight` and added a sibling `resetsAt` (objectui#7587).
+Measured: `resetsTonight` has no reader past `parseAiQuotaError`, so the
+inversion changes nothing on screen and the banner is the server's own sentence
+verbatim — pinned as an exact accounting of the rendered text, and as the flag
+being inert. Tests plus one field comment; no package is released by this
+change.
--- a/packages/plugin-chatbot/src/__tests__/ChatbotEnhanced.quotaResetPromise-7587.test.tsx
+++ b/packages/plugin-chatbot/src/__tests__/ChatbotEnhanced.quotaResetPromise-7587.test.tsx
@@ -1,0 +1,127 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * objectui#7587 — what a free-plan 429 actually puts on screen, now that the
+ * producer moved under this consumer: cloud PR #1852 (the rolling 7-day free
+ * quota) left `error.details.resetsTonight` a boolean but inverted its value
+ * for the free plan (`true` -> `false`) and started sending an ISO `resetsAt`
+ * beside it. `tool-display.ts` reads the flag and not the timestamp.
+ *
+ * #6385 already pinned that field — about its TYPE ("leaves `resetsTonight`
+ * undefined unless a producer sends an actual boolean", `tool-display.test.ts`).
+ * The type did not move. A value inversion and a new sibling key are exactly
+ * what a type pin cannot see, so this file pins the CONSEQUENCE: the rendered
+ * result.
+ *
+ * The measurement, taken here rather than assumed: after the parser populates
+ * `resetsTonight`, NOTHING reads it. `useAiQuotaCopy` — the only consumer of
+ * the parsed shape's fields — builds the banner from `message` / `messageEn` /
+ * `topUp`; the two other `parseAiQuotaError` call sites in `ChatbotEnhanced`
+ * use it as a yes/no predicate. So the free plan's inversion is invisible here,
+ * and the answer to "does this render nothing, or does it render a false
+ * promise?" is RENDERS NOTHING: every word the user gets about when the
+ * allowance returns is the server's own sentence, passed through verbatim.
+ *
+ * That is worth pinning in both directions. The banner is pinned by exact
+ * accounting (title + server sentence + CTA, and nothing else), so a reset line
+ * of our own cannot appear unnoticed; and the flag is pinned as inert, so the
+ * day someone wires "resets tonight" copy to a boolean that is now `false` for
+ * exactly the plan that needs the answer, this file says so.
+ */
+import '@testing-library/jest-dom/vitest';
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { ChatbotEnhanced } from '../ChatbotEnhanced';
+import { parseAiQuotaError } from '../tool-display';
+
+/** The server-owned refusal prose, as the cloud guardrail sends it. */
+const FREE_PLAN_ZH = 'AI 额度已用完，9 月 13 日恢复；升级后可继续使用。';
+const FREE_PLAN_EN =
+  'Your free AI allowance is used up. It comes back on 13 Sep — upgrade to keep going.';
+/** The ISO instant cloud PR #1852 added; no reader in this package. */
+const RESETS_AT = '2026-09-13T00:00:00Z';
+
+/**
+ * The free plan's post-#1852 wire shape: the declared envelope (ADR-0112) with
+ * the ledger code, `resetsTonight` now `false`, and `resetsAt` beside it —
+ * tagged the way `sendAwareFetch` tags a POST refused before any tokens
+ * streamed.
+ */
+function freePlanRefusal(resetsTonight: boolean): Error {
+  const e = new Error(
+    JSON.stringify({
+      success: false,
+      error: {
+        code: 'AI_ALLOWANCE_EXHAUSTED',
+        message: FREE_PLAN_ZH,
+        category: 'rate_limit',
+        details: {
+          messageEn: FREE_PLAN_EN,
+          upgrade: true,
+          topUp: false,
+          resetsTonight,
+          resetsAt: RESETS_AT,
+        },
+      },
+    }),
+  ) as Error & { notSent?: boolean; status?: number };
+  e.notSent = true;
+  e.status = 429;
+  return e;
+}
+
+function renderRefusal(resetsTonight: boolean) {
+  return render(
+    <ChatbotEnhanced
+      placeholder="Ask…"
+      messages={[]}
+      onSendMessage={vi.fn()}
+      onUpgrade={vi.fn()}
+      error={freePlanRefusal(resetsTonight)}
+    />,
+  );
+}
+
+describe('free-plan 429 banner — the reset promise nobody makes (objectui#7587)', () => {
+  it('renders the server sentence and nothing of our own about the reset', () => {
+    renderRefusal(false);
+    const banner = screen.getByRole('alert');
+
+    // Exact accounting: three parts, each with a named source — our own title,
+    // the server's sentence verbatim, and the CTA `topUp` picks. Anything this
+    // package starts deriving from the 429's reset fields has to land inside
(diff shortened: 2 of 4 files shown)


os-justin:
## Landing — ACCEPT

By the dispatching seat (`session_01YBWFb5YgMU5dw8p2VKj16S`, `os-justin`), 18:1xZ.

**CI: 32 of 32 check runs enumerated** (returned count equals `totalCount`), **31 green** — all four Test shards, `Build & E2E`, Type Check, **Lint**, Bundle Analysis, Changeset Bump Policy, Governed Surface Queue Guard. ⭐ Lint read explicitly rather than inferred: a sibling PR in the previous round lost a round to an `error`-severity unused import that a merits-only review had waved through.

⛔ **Standing down on `Live E2E (informational)`** — advisory, and red on `main` itself (run `34017174769` job `101442890465`, step 7 `Start ObjectStack backend (published packages)`, against the 09-05 control that passed the same step in 76s). ⚠️ That run reports `conclusion: success` at the **run** level while its job fails — ⛔ never read a run conclusion as evidence here. No fix to port; carded **objectui#8084**. ⛔ No re-run spent: a reproducible base-red is not a flake.

### What lands, and what it deliberately does not do

**+193 / −0**, and the production diff is **comment-only** — one new pin file, `tool-display.test.ts` (+38), `tool-display.ts` (+18 of recorded measurement at the field), and an **empty-frontmatter changeset** that releases nothing. `Clause-②` stays `no`: no interface field, symbol, or barrel entry was added.

⭐ It pins the *absence* rather than inventing a behaviour. `resetsTonight` is a declared, populated, now twice-pinned field with **zero readers** — I re-verified that independently with a control that fires (the same grep for `messageEn` returns consumers in `ChatbotEnhanced.tsx`, so the zero is a reading, not a broken query), plus `parseAiQuotaError` / `AiQuotaError` genuinely absent from the barrel. ⇒ The card's `true` → `false` flip changes **not one rendered byte**, and the reasoning for leaving it that way is recorded at the field rather than in a commit message nobody will find.

⛔ The obvious repair — read `resetsAt` and render a reset time — was correctly **not** taken, on two independent grounds: the 429's wire **position** for `resetsAt` has never been measured (it is an inference from prose, and *"a reader built on a guessed position fails silently"*), and objectui#7253 already ruled that *"one 429 can never be described by two different sentences on the same screen."*

⚠️ #7587 carries `pm:retriage` requesting a re-grade — its `priority:p2` rested on a user-visible string being wrong, and none is.

Flipping ready and enqueuing.

---

PR #8121, opened 2026-09-06
docs(guide): make three drifted counts on ci-cd-pipeline.md self-maintaining

baozhoutao:
Fixes #7965

Three sentences on `content/docs/guide/ci-cd-pipeline.md` qualified a file/document population with a
copied numeral and no declared measurement point. Each numeral was re-taken from **that figure's own
gate verdict line** — never re-counted by hand — and each sentence was then rewritten so it stays
true as the tree moves, rather than being refreshed to today's number.

Base `59a3a233d` · head `6528b3fe4` · one file changed, 10 insertions / 7 deletions.

## The verdict lines these fixes rest on (verbatim, run on this branch)

`node scripts/check-doc-component-types.mjs` (exit 0):

```
Scanned 188 doc file(s) (.mdx + .md), 1105 code block(s), 889 `type` literal(s) against 658 registered key(s) derived from 143 source file(s) (240 resolved call site(s), 133 indirect, 2 open): 772 registered, 117 exempted; 4 key table(s), 24 row(s), 45 table key(s) judged (namespaced + bare), 45 registered.
✅  Every documented component type is registered.
```

`node scripts/check-shell-escape-residue.mjs` (exit 0):

```
✅  check-shell-escape-residue: OK (5/5 root(s) resolved -- AGENTS.md: 1 file(s), 15 fence(s); CLAUDE.md: 1 file(s), 0 fence(s); skills: 16 file(s), 210 fence(s); .claude/skills: 4 file(s), 20 fence(s); content/docs: 184 file(s), 1064 fence(s); 206 file(s) and 1309 fenced block(s) examined in total; 0 occurrence(s) outside a fence (counted, not judged); coverage -- .claude: 4/4; skills: 16/16 document(s) under a declared root).
```

`pnpm exec vitest run scripts/__tests__/` at head `6528b3fe4` (exit 0):

```
 Test Files  112 passed (112)
      Tests  3360 passed (3360)
   Duration  104.70s
```

**Difference from the card.** The card measured 206 files / 889 literals / 107 test files. Two of the
three still read the same today (206 total files with the four-file `.claude/skills` root inside it;
889 literals). The test population has moved again since the card was written: **112 test files and
3360 tests**, not 107. That movement inside one working day is itself the argument for not writing a
bare numeral here.

### 1. `:782` — "measured across 143 files and 558 literals" → route (a), live reading

Before:

> **How a snippet is judged.** `type` is not one vocabulary in these pages — measured across 143 files
[...]

PR #8111, opened 2026-09-06
docs(gates): record the package-README census and defer check:doc-types' fourth leg (objectui#7896)

baozhoutao:
Refs #7896

Census-only. The widening does NOT land here, because the census does not read zero.

The `domain:ui` ruling on objectui#7896 (comment `5556586208`) orders this move census-first: walk all 39 `packages/NAME/README.md` with `check:doc-types`' own extractor and own registry, report a blind-spot count, and land the widening in the same pull request **only if the census reads zero**. It reads 26. So this pull request carries the measurement and a proposal per hit, and the fourth leg is left for the follow-up card the PM opens.

## STEP 0 — the card's two-way mutation, re-taken on `59a3a233d`

Assumption A1 holds. Today's counters have moved from the card's `af05c88bb` reading (`889` literals, `772` registered, was `887` / `770`); everything else reproduces exactly.

| case | mutation | disk hash after mutation | `check:doc-types` result |
|---|---|---|---|
| subject | `"type": "action:button"` to `action:buttonZZ` in `packages/app-shell/README.md` | `0393b247b70563082c817d8d36b4841bf2170571` | `EXIT=0`, counters byte-identical to baseline (188 files / 1105 blocks / 889 literals / **772** registered) |
| **control** | `"type": "object-grid"` to `object-gridZZ` in the root `README.md` | `7d01e3072e8e0d7f78a6a48218f28f78c0a3af50` | `EXIT=1`, `README.md:272 [unregistered-doc-type] type 'object-gridZZ' (json)`, registered **772 to 771** |

Both mutations proven on disk by counting the injected and the removed text before reading any verdict. Both restores proven by STATE, not by exit code:

| path | `HEAD` blob | disk hash after restore | `git diff HEAD --name-only` |
|---|---|---|---|
| `packages/app-shell/README.md` | `1304531be8aec7684dbccebda403077ee2914f3e` | `1304531be8aec7684dbccebda403077ee2914f3e` | empty |
| `README.md` | `b6c8467fdfe8e5102ceafff83922d9766f6f4f39` | `b6c8467fdfe8e5102ceafff83922d9766f6f4f39` | empty |

The control proves the instrument is alive; the byte-identical subject counters prove the package README is not in the scan population at all, rather than judged and forgiven.

## STEP 1 — the census (the deliverable)

[...]

PR #8132, opened 2026-09-06
docs(core): compile every README snippet and leave the UNGATED_DOCS ledger

baozhoutao:
Part of #5174 — batch 25 of the `UNGATED_DOCS` burn-down.

`packages/core/README.md` leaves the ledger and every ts/tsx block on the page
now compiles against the built types. The gate file's only change is the ten
lines of that one entry.

## Notation, read this first

This repository has measured that GitHub's body sanitizer deletes tag-shaped
fragments even inside backticks and fenced blocks. So generic parameters and
JSX elements are spelled out in words below — "Registry of any", "DeepReadonly
of T", "the ActionProvider element" — rather than in their real angle-bracket
form. The README itself carries the real spellings; this page does not.

## Concurrency

`main` moved under this batch **twice**, and both were absorbed by merging
(never rebasing, never force-pushing).

1. **`e23cfb000` to `cf1d29e2f`** — two PRs, including #8097, which declares
   `ComponentInput.of` on `packages/types` **source**. Package source moves the
   built types this corpus is judged against, so the forced build, the census,
   every gate, every probe and every test were re-run on the merged head. Every
   reading was byte-identical.
2. **`cf1d29e2f` to `f10bcd497`** — two docs PRs (#8121 on
   `content/docs/guide/ci-cd-pipeline.md`, #8111 adding a comment block to
   `scripts/check-doc-component-types.mjs`). Both are inside the corpus this
   gate scans, so everything was re-run **again** on `1ad6167a6`. Readings again
   byte-identical.

All 16 open pull requests' file lists were read over the REST files endpoint and
each count reconciled against that PR's own `changed_files`: 8118/9, 8116/34,
8112/1, 8111/1, 8109/1, 8104/2, 8097/16, 8091/1, 8090/1, 8076/6, 7749/4,
7685/67, 7058/24, 7054/2, 7053/2, 5400/1058. #5400's 1058-file list was paged to
exhaustion because the endpoint pages at 100 and a truncated list renders as a
confident absence. **No open PR touches `packages/core/README.md` or
`scripts/check-doc-snippet-types.mjs`.** The one open PR that touches any
package README is #7685, on `packages/auth/README.md` — which is exactly the
standing exclusion, so it doubles as the positive control that this grep can
find a hit at all.

(objectui#7990 / objectstack#16186), not because of this diff.

## Ledger arithmetic

Both rows MEASURED by a real gate run, never computed. Base row at the final
merge base `f10bcd497` bytes (probe M0, both paths checked out to the main blobs
and proven equal by hash, restored under a trap); branch row at `1ad6167a6`.

| reading | base `f10bcd497` | branch `1ad6167a6` |
|:--|--:|--:|
| `UNGATED_DOCS` entries | 13 | **12** |
| covered documents | 214 | 215 |
| …of them holding a ts/tsx block | 112 | 113 |
| covered blocks | 725 | 730 |
| blocks to compile | 567 | 572 |
| declared fragments | 158 | **158** (zero new) |
| semantic failures | 0 | 0 |
| self-imports judged (`check:readme-exports`) | 496 | 499 |

Programmatic diff of the two `UNGATED_DOCS` objects — both imported as OBJECTS,
never diffed as text: removed = `['packages/core/README.md']`, added = `[]`,
reason-changed = `[]`, survivor key order preserved = true. `git diff --numstat`
on the gate file against the merge base is exactly `0 10`. On the branch: core
entries **0**, `packages/NAME/README.md` entries **11** (from 12), control
`'packages/auth/README.md'` **1**.

The strictness region — the `Fence scanning` banner line to EOF, 1178 lines on
both sides — hashes to
`f46b5662ba336f026bca3a0003e0b7979789d48d0291fa5760be7d23c8ed0160` on the merge
base and on the branch. Re-measured here on both sides, not inherited; it is the
value batches 15 through 24 recorded.

**This entry's reason string was ACCURATE — the stale-ledger streak ends here.**
It claimed "5 undefined-name diagnostic(s) … plus TS2339x1 — TRIAGED, and NOT a
defect". Measured: exactly TS2304x5 and exactly TS2339x1, and the TS2339 is
exactly the line the string names. Every recent batch has reported its entry as
stale about what the debt was; this one was right, including its triage, and
that is worth recording rather than passing over.

## Census and the rung that decided

Re-taken on this batch's own base through the gate's exported `analyze()` /
`compileSnippets()`, one document un-gated at a time, over every non-excluded
ledger candidate — and re-taken again after each merge.

**Rung 1 does not decide, and neither does rung 2: `packages/core` and
`packages/plugin-kanban` tie on blocks (5), on rung 1's second term
(6 diagnostics) and on rung 2's distinct codes (2). Rung 3 decides
alphabetically — `packages/core/README.md`.**

| document | blocks | diagnostics | distinct codes | codes |
|:--|--:|--:|--:|:--|
| **packages/core/README.md** | **5** | **6** | **2** | TS2304x5 TS2339x1 |
| packages/plugin-kanban/README.md | 5 | 6 | 2 | TS1109x5 TS1011x1 |
| packages/plugin-ai/README.md | 4 | 8 | 3 | TS2304x4 TS2322x3 TS2552x1 |
| packages/providers/README.md | 4 | 8 | 2 | TS2304x7 TS2741x1 |
| packages/plugin-charts/README.md | 4 | 6 | 1 | TS1109x6 |
| packages/plugin-editor/README.md | 4 | 6 | 1 | TS1109x6 |
| packages/plugin-map/README.md | 4 | 2 | 2 | TS1005x1 TS2304x1 |
| packages/plugin-markdown/README.md | 4 | 2 | 2 | TS1005x1 TS1109x1 |
| packages/fields/README.md | 2 | 3 | 2 | TS7031x2 TS2307x1 |
| packages/plugin-tree/README.md | 1 | 3 | 2 | TS1005x2 TS1109x1 |

Assumption **A1 is CONFIRMED** — every block total, diagnostic total and
per-code figure matches batch 24's census exactly, and held identical across
both merges. Zero bound refusals on this page.

Exclusions re-verified against GitHub on this base, not inherited: the root
`README.md` is not a `packages/NAME/README.md`; `packages/auth/README.md` is
held by open draft PR #7685; `packages/plugin-chatbot/README.md` is deferred by
the earlier ruling.

## The System-View block: a failure that IS the documentation

This is the batch-25 ruling, and it is the reason this page needed a decision
rather than a repair.

`packages/core/README.md` demonstrates System-View immutability by writing a
`columns.push` that is **supposed to be refused**, annotated
`// (cross) TypeError (strict mode)`. Its TS2339 is not debt — the ledger's own
entry called it "TRIAGED, and NOT a defect", and the measurement agrees. Three
shapes were available and only one is honest:

- **Delete or comment out the line.** Refused: it deletes the lesson.
- **Declare the block a fragment.** Refused: a fragment's reason says the block
  *cannot be checked*, and this block can be — claiming otherwise would be the
  one thing the gate's fragment rule exists to prevent.
- **Keep the line and declare the expected failure.** Taken.

The line now carries, directly above it, a `@ts-expect-error` whose trailing
text is its reason — that the readonly rejection is `defineSystemView` deep
freezing the graph and typing it readonly, so the push is refused at compile
time as well as at run time. The prose comment stays: it is the runtime half of
the same lesson.

**This shape is a pin in BOTH directions, and both were measured.**

- Forward: the block compiles today, so the page is genuinely covered
  (`Semantic phase: 572 of 572 block(s) judged, 0 failed`).
- Backward: if `defineSystemView` ever stops returning a readonly graph, the
  directive becomes unused and TypeScript reports **TS2578** — so the gate goes
  red exactly when the documented property is lost. That is what probe **PD**
  proves, below, and it is why this is stronger than the ungated entry it
  replaces: an ungated page could lose the property in silence.

Two facts had to be established before this shape could be used, and both were:

1. **The gate honours the directive.** `scripts/check-doc-snippet-types.mjs`
   contains no mention of `expect-error`, `ts-ignore`, `ts-nocheck` or any
   `@ts-` prefix (grep count 0) — it hands each block to a real TypeScript
   program (`strict: true`) and reads the diagnostics back, so the directive is
   the compiler's to honour, not the gate's to strip. Confirmed empirically: the
   gate is green with the directive in place.
2. **The directive is load-bearing, not decorative.** Probe PD.

There was no fallback to the fragment marker, so there is no
`needs_decision` on this point.

## The other five: TS2304, paid down by binding to the shipped surface

All five are ambient names the page never defined. None is repaired with a
lenient alias, and none with a hand-copied type.

**Component Registry block (1).** `buttonMetadata` had no declaration. It now
gets a real self-import of the shipped `ComponentRenderer` name plus a
`declare const` stand-in typed to it.

⚠️ **And this PR reports what that stand-in is actually worth rather than
counting it as a check.** `ComponentRenderer` is a deliberate IDENTITY alias —
`Registry.d.ts` documents it as load-bearing for objectui#6298, it resolves to
its own parameter and contributes no type information, and
`component-config-single-declaration.test.ts` pins that identity so it cannot
change unnoticed. So the stand-in binds to a slot that is inert **by design**.
That is a measurement, not a defect, and nothing is filed for it: unlike batch
24's #8106 (an undeclared `any` on a public prop), this one is intentional,
documented and already pinned.

**A README-versus-shipped-type contradiction, fixed in the README.** While
typing that stand-in, the shipped signature contradicted the page's vocabulary:
`register()`'s second parameter is the **component**, its optional third is the
metadata, and `getMeta()` — not `get()` — reads metadata back. The block called
the second argument `buttonMetadata` and named `get()`'s result `metadata`. Both
are renamed to what the shipped surface says they are, and the prose beneath the
block now states the rule. **The gate cannot catch this** (the slot is the
identity alias above), so it is named here rather than left to be inferred from
the diff.

**Server Action Dispatch block (4).** `myAuthenticatedFetch`, `currentObject`,
`refetchData` and `runner` are all host-injected. Each is now a `declare const`
typed from the shipped surface — `ServerActionFetch` and `ActionRunner` by real
self-import — so the example is checked against the factory's own
`ServerActionHandlerConfig` rather than against a copy of it. **These are real
checks, and probes P4 and P5 are what make that falsifiable**: retype one
stand-in and the config literal rejects it by name.

⭐ **One of them turns a documented CLAIM into a compile-time check.**
`serverActionHandler.d.ts` says in prose that the handler it produces is
"Assignable to `ActionRunnerHandler`". By typing `runner` as the shipped
`ActionRunner`, the line `runner.registerHandler('script', script)` now
*verifies* that sentence on every commit instead of asserting it.

No `packages/**` source touched, no public type widened, no lenient alias added,
no gate loosened, and no new fragment marker: declared fragments 158 to 158.

## Probes — a base measurement and seven probes, each predicted in writing first

Predictions were written to the scratchpad before any probe ran and were not
amended: `5174-b25-PREDICTIONS.md` (md5 `5cbbdd6c967794a467871bad822890e1`) at
implementation commit `775ccf34e` with a clean tree. Every mutation is proven on
disk by occurrence counts of both the injected and the deleted text plus
`git hash-object` against the HEAD blob; every restore is proven by an empty
`git diff HEAD`; every leg runs under a `trap … EXIT INT TERM` with absolute
paths, restoring with `git checkout HEAD --` against an absolute path rather
than a bare checkout from the index. **No dist preflight is owed, and that is a
measurement rather than an omission:** both mutated paths are read from disk by
the gate itself, so no rebuild sits between a mutation and its reading.

| probe | mutation | predicted | observed |
|:--|:--|:--|:--|
| **M0** | both paths at merge-base bytes | exit 0; 214 covered / 13 ungated / 725 blocks / 567 to compile; readme-exports 496 | exactly that |
| **P1** | ledger entry restored ALONE | exit 0, coverage falls back to the base row | exactly that |
| **P2** | README restored ALONE | exit 1, exactly 6 diagnostics over exactly 2 codes | exactly that — TS2304x5, TS2339x1, matching the census row |
| **PD** | the directive MOVED onto the non-erroring `isSystemView` line | exit 1, **TS2578 "Unused '@ts-expect-error' directive"** naming this file and that line | exactly that — TS2578 at `README.md:134:1`, plus TS2339 returning at `133:22` |
| **P3** | delete the `runner` stand-in | exit 1, TS2304 "Cannot find name 'runner'" | exactly that |
| **P4** | retype `currentObject` to a number | exit 1, assignability error on `resolveObject` | exactly that — TS2322, "Type 'number' is not assignable to type 'string'" against the shipped resolver signature |
| **P5** | retype `myAuthenticatedFetch` to a void thunk | exit 1, TS2322 on the `fetch` key naming ServerActionFetch | exactly that — "not assignable to type 'ServerActionFetch'" |
| **P6** | fabricate a self-import name | readme-exports 1 fabricated, naming this page | exactly that — 498 real / 1 fabricated at `packages/core/README.md:48` |

**PD is the load-bearing one for this batch**, and it is load-bearing in an
unusual way: it does not prove a repair, it proves that a deliberately-failing
line is *still failing for the documented reason*. Moving the directive one line
down — onto `isSystemView(userListView)`, which does not error — turns the gate
red with TS2578, and the push line's TS2339 returns at the same time. Both
halves of the pin fire.

**No prediction was falsified.** All eight readings hit as written, on the first
implementation head and again after each of the two merges.

## Gates — all at `1ad6167a6`, clean tree, nothing pushed after

Exit codes captured by redirect-then-capture, before any pipe; each row quotes
the gate's own verdict line. The build was forced first
(`pnpm exec turbo run build --filter=./packages/* --concurrency=2 --force`,
39 successful / 0 cached) and again inside the provenance gate (43 successful /
0 cached); every dist-sensitive gate was then re-run on that force-built dist.
Note the invocation: bare `turbo` is not on PATH in this container, so
`pnpm exec turbo` is what ran.

- `pnpm check:doc-snippets` exit 0 — "Scanned 227 document(s): 215 covered (113 of them hold a ts/tsx block), 12 ungated" / "Covered blocks: 730 — 572 to compile, 158 declared fragment(s)" / "Root bound: no block imports a specifier that resolves only through this repository's ROOT manifest" / "Semantic phase: 572 of 572 block(s) judged, 0 failed" / "Every covered documentation snippet compiles against the built types."
- `pnpm check:doc-fences` exit 0 — "every TypeScript block in 227 document(s) is fenced ts/tsx/typescript"
- `pnpm check:readme-exports` exit 0 — "499 of them self-imports judged (499 real, 0 wrong-path, 0 fabricated)"
- `pnpm check:doc-types` exit 0 — "Every documented component type is registered."
- `node scripts/check-doc-links.mjs` exit 0 — "Links are valid across 17 scan roots."
- `pnpm check:doc-example-readers` exit 0 — "OK 80 documented symbol(s), 3947 call site(s)"
- `pnpm check:control-bytes` exit 0 — "OK (scanned 6499 tracked text file(s); skipped 85 binary)"
- `node scripts/check-node-esm-load.mjs --force-build` exit 0 — "Provenance leg: 37 of 37 gradable entries were built by this tree." Green on its first run, so no gate had to be re-run over a replayed sibling cache.
- `pnpm type-check:scripts` exit 0
- `node scripts/check-changeset-presence.mjs` exit 0 — "2 file(s) changed, 0 of them published source of a package the release covers … no changeset is owed." No `skip-changeset` label applied: this repository declares with an empty-frontmatter changeset rather than that label, and the presence gate says none is owed, so nothing was labelled at all by this seat.
- `node scripts/check-governed-queue-guard.mjs --test` on both paths exit 0 — "NOT GOVERNED — 2 path(s) checked against 5 governed surface(s); none matched"
- Owning package: `pnpm exec vitest run packages/core/` — "Test Files 120 passed (120) / Tests 2559 passed (2559)"; `pnpm --filter @object-ui/core type-check` exit 0, its output echoing `tsc --noEmit && tsc -p tsconfig.test.json`, so the test tsconfig is genuinely included and the script really ran.
- Beyond the gate, a manual control-byte scan of both changed paths with `grep -naP` over the C0/DEL ranges found nothing (exit 1, zero output bytes).

**Lint, with the narrowing declared and its three evidence items.** `pnpm lint`
here is `turbo run lint` per package, so the repository-wide sweep is CI's run;
the narrowed run over exactly this diff is
`pnpm exec eslint --no-inline-config --format json` on both paths — 2 entries,
**0 errors**. eslint's own answer for the README is "File ignored because no
matching configuration was supplied", so the real lint surface of this diff is
the one `.mjs` at 0 errors and 0 warnings. (1) The population eslint would
otherwise judge is **4380 files** by its own count. (2) That run exits 1 with
**93 errors and 12219 warnings**, and the touched files carry **0 of them** —
measured, and reported rather than smoothed, because a reader must not mistake
the narrowing for a green repository. (3) `eslint.config.js` declares no
`project`, `projectService` or `parserOptions.project` (grep count 0), so
type-aware linting is off and this diff cannot move the verdict on any untouched
file. Both readings were taken on the final commit `1ad6167a6`.

## Readers of the changed paths

Derived with `git grep -l` on both paths on this head, not guessed — **18 test
files, 724 tests, all green** in one run, and the set was identical before and
after both merges.

**A2 is CONFIRMED, and its warning was checked rather than assumed.**
`packages/core/src/utils/__tests__/freeze-schema.types.test.ts` does name
`packages/core/README.md` — but only in prose, in its header and in a comment
above its first case. It pins the *type behaviour*, not the README's text, with
its own inline cases, and it is green untouched. It is also independent
corroboration of the ruling: its case index 3, "the System View itself still
refuses .push on a nested array", asserts `rejected: true` for exactly the
statement this README demonstrates — and it compiles against **source** while
the snippet gate compiles against **dist**, so the two legs resolve differently
on purpose. The `@ts-expect-error` added here is the dist-side half of a
property the source side already pins.

The other two readers of the README are
`scripts/__tests__/check-doc-links.test.ts` and `packages/core/CHANGELOG.md`
(prose, not a reader). The remaining fifteen are readers of the **gate file**.
Batch 24 counted 16 readers; this batch counts 18 because its document has two
live readers of its own, which batch 24's did not.

## Out of scope

**Nothing was filed, and that is a finding rather than an omission.** The one
candidate — that `ComponentRegistry.register`'s component parameter contributes
no type information — was traced to a deliberate, documented, already-pinned
identity alias (objectui#6298) rather than to an oversight, so filing it would
have added noise to triage. It is reported in the section above instead. No
dedup search was run, because a search with nothing to dedup is not a reading.

This PR is a DRAFT and stays a draft: this seat does not flip ready, does not
enable auto-merge, and touches no labels or assignee.

---


baozhoutao:
Standing down on `Live E2E (informational)` for this PR, read 2026-09-06T17:58Z.

- **Failing check:** `Live E2E (informational)` — informational, not a required context, and red on `main` and on every open branch today for an upstream reason tracked in #7990 / objectstack#16186. This diff is one package README plus ten removed ledger lines in `scripts/check-doc-snippet-types.mjs`; nothing the live E2E exercises.

Every other check on `1ad6167a6` is judged on its own; this PR flips to ready once the required set converges, then goes to the merge queue by squash.

---

baozhoutao:

- Head `1ad6167a6`: 29 non-informational checks completed green at 18:07Z (the `ci-wait` reading); `Live E2E (informational)` red per the standing-down note above.
- `ready_for_review` at 18:07:31Z; the post-flip `Governed Surface Queue Guard` run started after that timestamp and completed `success` at 18:07:51Z.
- `added_to_merge_queue` at 18:07:56Z on the REST timeline (read 18:08:02Z). PRs #8129 and #8134 are ahead of it in the queue.

On MERGED: content probe on re-fetched `origin/main` (the `packages/core/README.md` row gone from `UNGATED_DOCS`, `// @ts-expect-error readonly by defineSystemView` at the System Views block; control: the `packages/auth/README.md` row still present), then the LANDED note here and on #5174. #5174 stays open (`Part of`); objectui#7864 (deferred on this gate file) dispatches after the landing, and batch 26 after #7864 and PR #8134.

---

baozhoutao:
LANDED — merged at 2026-09-06T18:23:15Z as `23933ae25` (squash, via the merge queue), `domain:devx @ objectui` execution seat, PM session `session_01FhBNJcLRZLe8M87VcUgpKr`, R46.

Content probe on re-fetched `origin/main` (`23933ae25`, read 18:23:54Z):
- The `'packages/core/README.md'` row is gone from `UNGATED_DOCS` (0 hits); package-README rows now 11.
- `packages/core/README.md:133` carries `// @ts-expect-error readonly by defineSystemView …` above the kept demonstration line.
- Control: the `'packages/auth/README.md'` row is still present (1 hit).

#5174 stays open (`Part of`); the LANDED note goes there in the same stroke. objectui#7864 takes the gate file next; batch 26 follows it.

---

PR #8135, opened 2026-09-06
test(i18n): pin that the two defaults-population walks agree (#7904)

### LINKED ISSUE #7904: finding(i18n/tooling): two independent AST walks discover the createSafeTranslation defaults population, with nothing pinning that they agree
Two independently-written AST walks now discover the same `createSafeTranslation` defaults-table population, and nothing pins that they agree.

## The two walks

1. **`packages/test-support/src/defaults-table-scan.ts`** — resolves every `createSafeTranslation(...)` / `createSafeTranslationHook(...)` first argument to its object literal, plus a three-entry hand-rolled registry. Consumed by two vitest gates: the placeholder-spelling rule in `@object-ui/i18n` and the "every defaults row names a key the en pack defines" rule in `@object-ui/app-shell`.
2. **`scripts/check-i18n-call-site-keys.mjs`** — its own `scanFactorySites` / `resolveFactoryTable` / `declaredConstant`, added with the `factory-default-drift` class. Separate implementation, separate file, separate runtime.

## They agree exactly today — which is the point

Measured on the same tree, same commit:

```
walk 1 (test-support module) : 32 factory call sites, 841 factory rows, 0 unreadable
walk 2 (check-i18n script)   : "Factory defaults tables: 32 createSafeTranslation site(s)
                                over 32 distinct table(s) (0 unreadable) — 841 row(s)
                                compared against their en value ... 0 unreadable."
```

32 = 32, 841 = 841, 0 = 0. That agreement is real and currently free of luck — both were written from the same understanding of the factory. But it is **not asserted anywhere**, so the day one of them learns a new table shape (a table behind a `satisfies`, a re-export, a computed member, a spread) and the other does not, the two populations silently diverge. One gate would go on reporting full coverage of a population the other has already grown past.

This is the same shape as the finding that produced the shared module in the first place: a definition of "which tables count" that exists in more than one place, with no mechanism raising a hand when the copies drift.

## Why they were not simply merged

There is a real structural reason, not an oversight: `scripts/*.mjs` run under bare node, which cannot import the TypeScript-source module the vitest gates share. So "just call the same function" is not available today without a build step or a loader for the scripts lane. That makes this a genuine design question rather than a cleanup.

## A cheap pin, if one is wanted

The script already PRINTS its census line. A test could parse that line and assert it against `scanDefaultsTables(REPO_ROOT)` — same site count, same distinct-table count, same row count, same unreadable count. That pins the agreement without merging the implementations, and it fails the day either walk changes shape alone. Whether that is worth a gate, or whether the divergence should instead be closed by giving the scripts lane a way to import the shared module, is the decision to make.

## Notes

- No user-visible defect today: both walks are correct on this tree and the numbers match.
- The two gates judge different properties (the script value-compares and deliberately abstains where `en` defines no value; the app-shell gate judges whether the key exists at all), so this is only about the shared POPULATION definition, not about the rules.

Filed by the `os-dev` seat while landing the defaults-maps discovery widening. No assignee — PM triage.

---

baozhoutao:
Fixes #7904

One new test, `scripts/__tests__/defaults-population-parity-7904.test.ts`. Neither walk moves; nothing else in the tree changes.

## What it pins

Two independently written AST walks discover the same `createSafeTranslation` defaults-table population — `scanDefaultsTables` in `packages/test-support/src/defaults-table-scan.ts` (objectui#7884, read by the two vitest gates) and the script's own `scanFactorySites` / `resolveFactoryTable` behind `analyze()` in `scripts/check-i18n-call-site-keys.mjs` (objectui#7567). They agree exactly today and nothing asserted it, so the day one of them learns a new table shape and the other does not, one gate goes on reporting full coverage of a population the other has already grown past.

This is the pin, not a merge. `scripts/*.mjs` run under bare node and cannot import the TypeScript-source module the vitest gates share, and giving the scripts lane a loader or a build step is a capability expansion with no pull beyond this one gate. Shape follows `scripts/__tests__/placeholder-spelling-parity.test.ts` (objectui#7310): floors first, then equality, plus the "is asking two different implementations" guard so a future merge of the walks must delete this file rather than leave it reading as a guarantee.

The numbers are read through the script's exported `analyze(root).counters`, not off its prose census line — no spawn, no regex.

## The mapping — comparing like with like

The trap the card names: `scanDefaultsTables` walks the factory call sites AND THEN appends every declared hand-rolled table into ONE `rows` / `tables` pair, while the script keeps `factory*` and `handRolled*` counters deliberately separate. Comparing the scan's totals against the script's factory counters would compare 1072 rows against 841 and be red for a reason that is not drift. So the scan side is partitioned first:

- **Hand-rolled half** — the scan labels a registry table exactly `NAME (file)`, derived from the same `HAND_ROLLED_TABLES` both sides read (objectui#7877 put the data in one JSON file, so this is one declaration, not two copies). A factory label always carries a line number, `NAME (file:LINE)`, so the two label shapes cannot collide.
- **Factory half** — every other label, and the rows carrying it.
- **The overlap** — `TIMELINE_DEFAULT_TRANSLATIONS` is in the registry AND reaches the factory. The script de-duplicates on the literal's own position and counts it as `handRolledAlreadyFactoryCovered`; the scan double-lists it on purpose and tells callers to de-duplicate on `where` plus `key`. This test is that caller.

Counter by counter, as asserted:

| scan side (`defaults-table-scan.ts`) | script side (`analyze().counters`) | today |
| --- | --- | --- |
| factory table labels | `factorySites` | 32 |
| distinct factory literals | `factoryTables` | 32 |
| factory rows | `factoryRows` | 841 |
| `HAND_ROLLED_TABLES.length` | `handRolledDeclared` | 3 |
| hand-rolled labels, overlap-free | `handRolledTables` | 2 |
| hand-rolled labels in the overlap | `handRolledAlreadyFactoryCovered` | 1 |
| hand-rolled rows, overlap-free | `handRolledRows` | 215 |
| `scan.unreadable` is empty | all four unreadable counters at 0 | 0 |

Two of those rows hold only because of a property of today's tree, and the file asserts both before the equalities rather than assuming them: `factorySites` counts invocations including ones whose table never resolves (equal to the label count only while `factoryUnreadableTables` is 0), and `factoryTables` counts distinct literals while the scan pushes one label per site (distinct literals are recovered from each table's row signature, which needs every factory table to contribute at least one row — asserted, since two empty tables would share the empty signature). The header records both, and records the two known definitional differences this tree does not exercise: the scan descends into nested object literals where the script deliberately counts one as an unreadable row, and the scan accepts a property-access callee where the script requires a bare identifier.

Floors before equality, so the pin cannot pass over two empty walks: 1591 source files walked, more than 25 factory sites, more than 700 factory rows on both sides, more than 150 overlap-free hand-rolled rows on both sides (the same 150 the script's own collapse floor uses), and zero unreadable entries on both sides.

## Today's readings, both sides

`node scripts/check-i18n-call-site-keys.mjs` at `ed4a1f810`, its census lines quoted as the control that the numbers asserted are the numbers it prints:

```
Factory defaults tables: 32 createSafeTranslation site(s) over 32 distinct table(s) (0 unreadable) — 841 row(s) compared against their en value, 841 matching, 0 on a key en does not define, 0 with no single comparable en value, 0 unreadable.
Hand-rolled defaults tables: 3 declared, 2 scanned here (1 already covered by the factory walk, 0 unreadable) — 215 row(s) compared against their en value, 215 matching, 0 on a key en does not define, 0 with no single comparable en value, 0 unreadable.
```

Scan side, measured in the same process as the assertions: 35 table labels total, of which 32 factory (32 distinct literals) and 3 hand-rolled; 1072 rows, of which 841 factory and 231 hand-rolled, 215 of them once the 16 rows of the overlapping `TIMELINE_DEFAULT_TRANSLATIONS` literal are attributed to the factory half where the script already counts them; `unreadable` empty; 1591 runtime source files walked. The card's numbers (32 / 841 / 0) were re-measured on this base rather than inherited, and the hand-rolled half was measured for the first time here.

Cost: `scanDefaultsTables` about 0.5s, `analyze` about 3s, both in the collection phase so neither sits inside a test's timeout budget. The whole file runs in 4.4s.

## Positive control — proof the pin can go red

Narrowed the SCRIPT's recogniser in the working tree so its factory walk stops seeing the one `createSafeTranslationHook` call site (`packages/plugin-detail/src/useDetailTranslation.ts`), leaving the shared scan's recogniser intact. Probe only — neither walk is in this PR's file surface.

Anchor observed before and after the edit (canonical spelling 1 then 0, narrowed spelling 0 then 1), mutated blob hash `2d34345c` different from the HEAD blob `dea6c08d`, so the mutation demonstrably reached disk. The pin went red, exit 1, three of four cases failing and the equality case naming the counters:

```
factorySites: defaults-table-scan.ts -> 32, check-i18n-call-site-keys.mjs -> 31
factoryTables: defaults-table-scan.ts -> 32, check-i18n-call-site-keys.mjs -> 31
factoryRows: defaults-table-scan.ts -> 841, check-i18n-call-site-keys.mjs -> 666
```

The recogniser guard failed alongside it (`createSafeTranslation, createSafeTranslationHook` against `createSafeTranslation`), and so did the row floor (666 is not greater than 700) — three independent alarms for one narrowing.

Restore proven, not assumed: `git checkout HEAD -- scripts/check-i18n-call-site-keys.mjs` under a `trap` with absolute paths, then the restored blob hash read back as `dea6c08d`, byte-equal to the HEAD blob, and `git diff HEAD` empty.

## Gates, all at `ed4a1f810`

- `pnpm exec vitest run scripts/__tests__/defaults-population-parity-7904.test.ts` — exit 0, 4 passed.
- `pnpm exec vitest run scripts/__tests__/placeholder-spelling-parity.test.ts scripts/__tests__/check-i18n-call-site-keys.test.ts` — exit 0, 141 passed. Neither file changes.
- `pnpm exec vitest run scripts/__tests__/` — exit 0, 114 files, 3376 tests passed.
- `node scripts/check-i18n-call-site-keys.mjs` — exit 0, census quoted above.
- `pnpm type-check:scripts` — exit 0. Both package-owned imports are relative, as `scripts/__tests__/scripts-type-check.test.ts` pins.
- `pnpm lint:root` — exit 0 (32 pre-existing warnings, none in the new file).
- `pnpm check:control-bytes` — exit 0, 6505 tracked text files scanned; plus a targeted `grep -naP` self-scan of the new file, no hits.
- `node scripts/check-changeset-presence.mjs` — exit 0, none owed (1 file changed, 0 published source).
- `node scripts/check-governed-queue-guard.mjs --test scripts/__tests__/defaults-population-parity-7904.test.ts` — exit 0, NOT GOVERNED.
- `pnpm check:entry-guard` — exit 0.

`Live E2E (informational)` is red on every branch today for an upstream reason (objectui#7990 / objectstack#16186), unrelated to this change.

---

### DIFF
--- a/scripts/__tests__/defaults-population-parity-7904.test.ts
+++ b/scripts/__tests__/defaults-population-parity-7904.test.ts
@@ -1,0 +1,272 @@
+import { describe, expect, it } from 'vitest';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { FACTORY_NAMES as SCRIPT_FACTORY_NAMES, analyze } from '../check-i18n-call-site-keys.mjs';
+import {
+  FACTORY_NAMES as SCAN_FACTORY_NAMES,
+  HAND_ROLLED_TABLES,
+  scanDefaultsTables,
+} from '../../packages/test-support/src/defaults-table-scan';
+
+/**
+ * objectui#7904 — one defaults POPULATION, two walks, held to each other.
+ *
+ * ## What is duplicated, and why it stays that way
+ *
+ * Two independently written AST walks discover the same `createSafeTranslation`
+ * defaults tables:
+ *
+ *   - `packages/test-support/src/defaults-table-scan.ts` (`scanDefaultsTables`,
+ *     objectui#7884) — TypeScript source, imported by the two vitest gates that
+ *     judge the rows (the objectui#3512 placeholder-spelling case in
+ *     `@object-ui/i18n`, and `@object-ui/app-shell`'s "every row names a key the
+ *     en pack defines").
+ *   - `scripts/check-i18n-call-site-keys.mjs` (`scanFactorySites` /
+ *     `resolveFactoryTable` / `declaredConstant`, behind `analyze()`,
+ *     objectui#7567) — the `factory-default-drift` class, a bare-node script.
+ *
+ * They were not merged for a structural reason, not an oversight:
+ * `scripts/*.mjs` run under bare node and cannot import the TypeScript-source
+ * module the vitest gates share (`exports["./defaults-table-scan"]` resolves to
+ * `./src/*.ts`, and no fresh checkout has a `dist`). Giving the scripts lane a
+ * loader or a build step is a capability expansion with no pull beyond this one
+ * pin, so objectui#7904 took the cheaper half: pin the agreement instead of
+ * merging the implementations.
+ *
+ * ## So this file is the guarantee instead
+ *
+ * The two walks agree exactly today, and that agreement is currently free of
+ * luck — both were written from the same understanding of the factory. But
+ * nothing asserted it, so the day one of them learns a new table shape (a table
+ * behind a `satisfies`, a re-export, a computed member, a spread) and the other
+ * does not, the two populations diverge in silence and one gate goes on
+ * reporting full coverage of a population the other has already grown past.
+ *
+ * ⛔ This file judges only the POPULATION — which call sites, which tables,
+ * which rows, and how much of it neither walk can read. It does not judge the
+ * RULES: the script value-compares each row against its `en` value and abstains
+ * where `en` defines none; the app-shell gate asks whether the key exists at
+ * all. Those stay each side's own business, exactly as objectui#7310's
+ * `placeholder-spelling-parity.test.ts` leaves both spelling suites alone.
+ *
+ * ## The trap: the two sides do NOT partition the population the same way
+ *
+ * `scanDefaultsTables` walks the factory call sites AND THEN appends every
+ * declared hand-rolled table, into ONE `rows` / `tables` pair. The script keeps
+ * the two halves in SEPARATE counters on purpose (its class-8 header says why:
+ * a registry that resolved nothing would be invisible inside a combined number
+ * next to 841 factory rows). Comparing the scan's totals against the script's
+ * `factory*` counters would therefore compare 1072 rows against 841 and be red
+ * for a reason that is not drift. So the scan side is partitioned first:
+ *
+ *   - HAND-ROLLED half — the scan labels a hand-rolled table exactly
+ *     `NAME (file)`, built from the same `HAND_ROLLED_TABLES` registry the
+ *     script reads (both sides read `hand-rolled-tables.json`, objectui#7877 —
+ *     one declaration, not two copies). A FACTORY label is always
+ *     `NAME (file:LINE)`, so the two label shapes cannot collide: the derived
+ *     `NAME (file)` string is never produced by the factory branch.
+ *   - FACTORY half — every other label, and the rows carrying it.
+ *
+ *   - THE OVERLAP. `TIMELINE_DEFAULT_TRANSLATIONS` is in the registry AND
+ *     reaches the factory. The script de-duplicates on the literal's own
+ *     position and counts it as `handRolledAlreadyFactoryCovered`, excluding its
+ *     rows from `handRolledRows`. The scan deliberately double-lists it (its
+ *     docstring: "a caller must de-duplicate on `where` + `key`"). This file is
+ *     that caller: a hand-rolled table whose every row shares a `where` + `key`
+ *     with a factory row IS the literal the factory walk already scanned.
+ *
+ * ## The mapping asserted below, counter by counter
+ *
+ *   scan factory table labels          = counters.factorySites
+ *   distinct factory literals          = counters.factoryTables
+ *   scan factory rows                  = counters.factoryRows
+ *   HAND_ROLLED_TABLES.length          = counters.handRolledDeclared
+ *   scan hand-rolled labels, overlap-free = counters.handRolledTables
+ *   scan hand-rolled labels in overlap  = counters.handRolledAlreadyFactoryCovered
+ *   scan hand-rolled rows, overlap-free = counters.handRolledRows
+ *   scan.unreadable = []               = all four of the script's unreadable
+ *                                        counters at 0
+ *
+ * Two of those rows hold only because of a property of TODAY's tree, and both
+ * are asserted before the equalities rather than assumed:
+ *
+ *   - `factorySites` counts INVOCATIONS including ones whose table never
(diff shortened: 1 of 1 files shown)


baozhoutao:
Standing down on `Live E2E (informational)` for this PR, read 2026-09-06T18:16Z.

- **Failing check:** `Live E2E (informational)` — informational, not a required context, and red on `main` and on every open branch today for an upstream reason tracked in #7990 / objectstack#16186. This diff is one new parity test under `scripts/__tests__/`; nothing the live E2E exercises.

Every other check on `ed4a1f810` is judged on its own; this PR flips to ready once the required set converges, then goes to the merge queue by squash.

---

baozhoutao:

- Head `ed4a1f810`: 28 non-informational checks completed green at 18:28Z (the `ci-wait` reading); `Live E2E (informational)` per the standing-down note above.
- `ready_for_review` at 18:28:50Z; the post-flip `Governed Surface Queue Guard` run started after that timestamp and completed `success` at 18:29:06Z.
- `added_to_merge_queue` at 18:29:15Z on the REST timeline (read 18:29:20Z). The queue is otherwise empty.

On MERGED: content probe on re-fetched `origin/main` (`scripts/__tests__/defaults-population-parity-7904.test.ts` present with its equality case; control: `scripts/__tests__/placeholder-spelling-parity.test.ts` unchanged), then the LANDED note here and the closing stroke on #7904.

---

baozhoutao:
LANDED — merged at 2026-09-06T18:45:15Z as `5b09821ed` (squash, via the merge queue), `domain:devx @ objectui` execution seat, PM session `session_01FhBNJcLRZLe8M87VcUgpKr`, R46.

Content probe on re-fetched `origin/main` (`5b09821ed`, read 18:45:47Z):
- `scripts/__tests__/defaults-population-parity-7904.test.ts` is present with its equality case (`the two walks discover the same population`, 1 hit).
- Control: `scripts/__tests__/placeholder-spelling-parity.test.ts` is byte-identical to its pre-merge state (`git diff --stat` empty against `336a9eb8a`).

#7904 closed by the `Fixes` at merge time; the seat strips `pm:dispatched` and clears the assignee there in the same stroke.

---

PR #8136, opened 2026-09-06
fix(devx): refuse published build records in the tooling gate

### LINKED ISSUE #7003: finding(build): relocating `tsBuildInfoFile` into `outDir` would ship it in twelve tarballs — and the tooling gate cannot see it
Filed by the `domain:devx` @ `objectui` execution seat (PM session `session_01GgDDqh6YnkXqsnVTCa7wHk`, R36) while reviewing PR #7001 for #6703. Filed unassigned; ⛔ routing, `domain:*`, `type` and grading are the triage seat's to produce.

## Where this comes from

PR #7001 (landing for #6703) closes the `tsc`-exits-0-on-a-partial-`dist` hazard by **counting** the emitted set inside each build. Its dev correctly reported — rather than built — a structural alternative, and recommended it as a follow-up:

> **B.** Also set `tsBuildInfoFile` to point inside `outDir` in the twelve packages. Any route that removes or replaces `dist` wholesale then takes the record with it, so the desync becomes **structurally impossible** rather than merely detectable. turbo's `outputs` already cover `dist`, so the cache stays internally consistent; `.gitignore` already covers both.

That reasoning is sound as far as it goes, and the option is genuinely attractive — a self-contained `dist/` has no invisible second half to reason about. ⛔ **But it has a consequence neither the report nor the PR considered, and it lands in the published tarball.**

## The measurement

Every one of the affected packages publishes by directory, not by enumeration. Read on `origin/main`:

```
packages/types/package.json   "files": ["dist","README.md","CHANGELOG.md","LICENSE"]
packages/core/package.json    "files": ["dist","README.md","CHANGELOG.md","LICENSE"]
packages/react/package.json   "files": ["dist","README.md","CHANGELOG.md","LICENSE"]
```

⇒ **Anything inside `dist/` ships.** Putting `tsconfig.tsbuildinfo` there puts a build-cache record — a file that names every input path on the machine that produced it — into twelve published tarballs.

## And the gate that exists for exactly this class would not catch it

`scripts/check-published-dist-tooling.mjs` is the gate for tooling material in published output. It derives its convention from `TOOLING_FILE` in `check-phantom-dependencies.mjs`, whose pattern ends:

```
\.[cm]?[jt]sx?$
```

The artifact side is matched back to a **tooling source file** carrying a JS/TS extension — that is how `dist/__tests__/foo.d.ts` is traced to `src/__tests__/foo.ts` (the live instance in #6943). A `tsconfig.tsbuildinfo` has **no tooling source** and does not end in `.ts`/`.js`/`.mjs`/`.cjs`, so it matches nothing the convention describes.

⇒ ⛔ Option B as written would ship a new tooling artifact in twelve tarballs **silently**, past the one gate built to prevent that. ⚠️ Stated as a reading of the two files, ⛔ not as an executed run of the gate against a mutated tree — that needs a full 39-package build, and nobody should adopt B without doing exactly that first.

## What this does NOT say

⛔ It is not an argument against B. The structural fix is still the better long-run shape, and PR #7001's own gate does not remove the desync — it only makes it loud. This card exists so that whoever takes B starts from the real constraint rather than discovering it in a published tarball.

Three directions, none of them adjudicated here:

1. Point `tsBuildInfoFile` somewhere that is neither the package root nor `outDir` — a repo-level cache directory — so `rm -rf dist` still cannot desync it, and nothing new enters the tarball. ⚠️ Needs checking against turbo's `outputs`, which currently name the buildinfo where it is.
2. Keep B and add an explicit negation to each `files` list. ⚠️ Twelve places to forget one.
3. Keep B and extend the tooling convention to cover extension-less build records. ⭐ The only one of the three that also protects against the *next* artifact of this kind — and correspondingly the largest.

## Reproduce the two readings

```
git grep -n '"files"' -- packages/types/package.json packages/core/package.json
git show origin/main:scripts/check-phantom-dependencies.mjs | grep -n 'TOOLING_FILE'
```

⭐ Control for the second: the same file must also show the `__tests__` half of the convention. A zero there means the search broke, not that the convention is absent.

## Related

#6703 (the card PR #7001 lands) · PR #7001 (ships option A, the counting gate) · #6943 (`check:published-dist` red on `main` today — the same gate, a different artifact) · #6861

baozhoutao:
Fixes #7003

Direction 3 of the card, and only that: the gate's convention learns to refuse build records. No `tsBuildInfoFile` setting anywhere (option B is still un-adopted, and this PR does not adopt it), no `files` negation, no `turbo.json` and no `.gitignore` change.

## The shape decision (card ruling 2): a second, artifact-only term

The build-record rule lands **beside** `PUBLISHED_TOOLING_FILE` as its own exported `BUILD_RECORD`, not as a third alternation inside the shared `TOOLING_FILE`. Three reasons, the first mechanical:

1. **It would not arrive.** `toolingConventionFrom()` extracts exactly TWO halves out of `TOOLING_FILE.source` — the directory alternation and the stem alternation — and `PUBLISHED_TOOLING_FILE` is rebuilt from those two. A third alternation added over there is dropped here silently: the phantom-dependency gate would change behaviour and this one would not. That is precisely the drift the derivation exists to prevent, so the shared regex is the one place the term must not go.
2. **`TOOLING_FILE` grades SOURCE files, and a build record is not one.** Its five other readers (`check-phantom-dependencies`, `check-doc-example-shared-reader`, `check-handler-key-read-sites`, `check-node-esm-load`, `check-unreferenced-sources`) walk source trees filtered by `SOURCE_FILE` (a JS/TS extension chain), so a `tsconfig.tsbuildinfo` under `src/` would never even be listed. The term would be a declaration none of them honours — declared and unenforced, which is the shape this repo refuses elsewhere.
3. **`check-published-tsconfig-tooling-exclude.mjs` turns that convention into tsconfig `exclude` patterns.** Excluding a `.tsbuildinfo` from a program is meaningless: `tsc` writes the record, it never reads one as an input.

The "derived, never retyped" property is kept in the form that applies to a term with no source counterpart: `BUILD_RECORD` is exported once, and the pin asserts `isToolingArtifact` is exactly the union of the two exported patterns — so a third spelling cannot appear in the gate without appearing in the test.

## The pattern

```
export const BUILD_RECORD = /(^|\/)[^/]*\.tsbuildinfo$/;
export const isToolingArtifact = (path) => PUBLISHED_TOOLING_FILE.test(path) || BUILD_RECORD.test(path);
```

**Recognised:** any basename, at any depth, ending `.tsbuildinfo` — `tsconfig.tsbuildinfo`, `tsconfig.build.tsbuildinfo`, a bare `.tsbuildinfo`, and the same names nested (`dist/chunks/tsconfig.build.tsbuildinfo`), in every `BUILD_OUTPUT_DIRS` entry. Anchored to the last path segment, so a directory that happens to be named `a.tsbuildinfo` does not drag its contents in.

**Why that suffix and not a list of guessed artifacts:** it IS this repository's spelling for the thing. `turbo.json`'s build `outputs` name a recursive glob over `*.tsbuildinfo` and `.gitignore` ignores the same one, so a record renamed away from that suffix is already uncached and untracked. **Not recognised, and said out loud rather than papered over:** a record whose `tsBuildInfoFile` points at an arbitrary name with another extension. Nothing readable from the tarball tells that file apart from an emitted one, and a guessed list would red a clean package.

## The positive control — the card's own measurement, run

Built the workspace the way the gate does (`turbo run build --filter=!@object-ui/site`, 43 tasks successful), then four legs on commit `79fe2c6c0`. Full transcript is in the report comment on the card.

| leg | tree | gate | exit | reading |
| 0 | clean, built | post-change | 0 | 39 packages, 5923 tarball files, **0** tooling artifacts |
| 1 | defective | **pre-change** (`HEAD~1`, 0 occurrences of `BUILD_RECORD`) | **0** | 5924 files, **0** tooling artifacts — the card's "would ship silently" claim, measured |
| 2 | defective | post-change | **1** | `@object-ui/types  [tooling-in-published-output]  packages/types/dist/tsconfig.tsbuildinfo` |
| 3 | restored | post-change | 0 | back to 5923 files, 0 tooling artifacts |

The mutation is that package's **real** record (58377 bytes) copied into its published `dist/`, sha256 compared against the original; `npm pack --dry-run --json` in `packages/types` then listed `dist/tsconfig.tsbuildinfo` in the tarball, which confirms the dispatch's assumption A2 against real pack output rather than by reading `files`. Restore proven by the mutant's absence and an empty `git status --porcelain` (the file is untracked, so `git diff` says nothing about it), and by the file count returning to 5923.

Leg 1 is the load-bearing one: the gate built for exactly this class was green on the defective tree, with the offending record in the pack list.

## The pin (`scripts/__tests__/check-published-dist-tooling.test.ts`)

New property 8 in the file's header, and five cases in the existing shape:

- a build record is invisible to BOTH `TOOLING_FILE` and `PUBLISHED_TOOLING_FILE` and visible to `isToolingArtifact` — the blind spot in one assertion, driven off `BUILD_OUTPUT_DIRS` so the directory list is not retyped;
- `isToolingArtifact` equals the union of the two exported patterns over records, roots, emitted JS, declarations and tooling directories — the derived-never-retyped property for the new term;
- names that merely contain the word are clean (`dist/tsbuildinfo.js`, `dist/tsconfig.tsbuildinfo.js`, `dist/a.tsbuildinfo/index.js`);
- `auditPackedFiles` names the package AND the file for `dist/tsconfig.tsbuildinfo` and a nested `dist/chunks/tsconfig.build.tsbuildinfo`;
- a record at the **package root** — today's real location for all 30 composite packages, and the state option B would change — is NOT a finding.

`MIN_PACKAGES` and the wiring cases are untouched and green.

## Docs

One sentence in `content/docs/guide/ci-cd-pipeline.md`: the Published Dist Gate section enumerates what the gate refuses, and that enumeration became false. The build record is added to it with the reason. The command cells pinned by objectui#3653 are untouched.

## Gates

Green at `79fe2c6c0`: `vitest run scripts/__tests__/` (113 files, 3377 tests) · `pnpm check:published-dist` (the four legs above) · `node scripts/check-dist-completeness.mjs` (control, objectui#7001 — unaffected, the record does not move) · `node scripts/check-phantom-dependencies.mjs` (control) · `node scripts/check-published-tsconfig-tooling-exclude.mjs` (control, 34 enforced packages) · `pnpm type-check:scripts` · `pnpm lint:root` · `pnpm check:control-bytes` plus a `grep -naP` control-byte self-scan of all three changed paths · `pnpm check:entry-guard` · `pnpm check:doc-fences` / `check:doc-types` / `check:doc-snippets` / `docs:check-links` · `node scripts/check-changeset-presence.mjs` (no changeset owed: 3 files changed, none published source of a released package, none a manifest whose published contract moved) · `node scripts/check-governed-queue-guard.mjs --test` on all three paths (NOT GOVERNED).

`Live E2E (informational)` is red on every branch today for an upstream reason (objectui#7990) and is not this PR's.

Session for this work, as prose so it survives a body edit: `https://claude.ai/code/session_01FhBNJcLRZLe8M87VcUgpKr`

---

### DIFF
--- a/content/docs/guide/ci-cd-pipeline.md
+++ b/content/docs/guide/ci-cd-pipeline.md
@@ -1471,8 +1471,10 @@ below for the half of that mechanism which is still live.
 workflow; manual. **It carries no `pull_request` trigger, on purpose.**
 
 No published package's build output may contain tooling material — `__tests__/`, `__mocks__/`,
-`__benchmarks__/`, `*.test.*`, `*.spec.*`, `*.bench.*`, `*.stories.*`. The gate is
-`scripts/check-published-dist-tooling.mjs` (`pnpm check:published-dist`); it builds every
+`__benchmarks__/`, `*.test.*`, `*.spec.*`, `*.bench.*`, `*.stories.*`, or a `*.tsbuildinfo` build
+record ([#7003](https://github.com/objectstack-ai/objectui/issues/7003): a record has no tooling
+source to be traced back to, and it names every input path on the machine that produced it). The
+gate is `scripts/check-published-dist-tooling.mjs` (`pnpm check:published-dist`); it builds every
 published package itself, then reads each one's tarball file list from `npm pack --dry-run`.
 
 Three things about it are easy to get wrong and are written down in the script's own header
--- a/scripts/__tests__/check-published-dist-tooling.test.ts
+++ b/scripts/__tests__/check-published-dist-tooling.test.ts
@@ -7,6 +7,7 @@ import { fileURLToPath } from 'node:url';
 import { TOOLING_FILE } from '../check-phantom-dependencies.mjs';
 import {
   BUILD_OUTPUT_DIRS,
+  BUILD_RECORD,
   MIN_PACKAGES,
   PUBLISHED_TOOLING_FILE,
   analyze,
@@ -47,6 +48,13 @@ import {
  *  7. **The gate is wired into the publish path**, which is where the ruling put
  *     it (comment 5307574139), and into a nightly workflow — and NOT into a
  *     per-PR job.
+ *  8. **A build record is refused even though it has no tooling SOURCE.**
+ *     objectui#7003 measured the blind spot: every criterion above traces an
+ *     artifact back to a source the convention names, and a `*.tsbuildinfo` is
+ *     a compiler by-product with no such source, so `PUBLISHED_TOOLING_FILE`
+ *     matches nothing about it. `BUILD_RECORD` is the second, artifact-only
+ *     term; the cases below read it from the gate rather than retyping the
+ *     suffix, so the two cannot disagree about what a build record IS.
  */
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
 const GATE = 'scripts/check-published-dist-tooling.mjs';
@@ -161,6 +169,58 @@ describe('the artifact criterion vs. the source criterion', () => {
   });
 });
 
+// ── 2b. build records, which have no tooling source at all ───────────────────
+
+describe('build records — objectui#7003', () => {
+  /** Spellings `tsc` writes, built from the gate's own output-directory list. */
+  const records = BUILD_OUTPUT_DIRS.flatMap((dir) => [
+    `${dir}/tsconfig.tsbuildinfo`,
+    `${dir}/tsconfig.build.tsbuildinfo`,
+    `${dir}/.tsbuildinfo`,
+    `${dir}/nested/deep/tsconfig.tsbuildinfo`,
+  ]);
+
+  it('is exactly the blind spot the card measured: invisible to the derived convention', () => {
+    // The whole finding of objectui#7003 in one assertion. `PUBLISHED_TOOLING_FILE`
+    // matches a tooling MARKER — a directory name or a stem — because every
+    // artifact it grades was emitted from a file somebody wrote. A build record
+    // was written by the compiler about the build, so it carries no marker, and
+    // the gate built for tooling material in published output saw nothing.
+    for (const record of records) {
+      expect(PUBLISHED_TOOLING_FILE.test(record), `${record} via PUBLISHED_TOOLING_FILE`).toBe(false);
+      expect(TOOLING_FILE.test(record), `${record} via TOOLING_FILE`).toBe(false);
+      expect(isToolingArtifact(record), `${record} via isToolingArtifact`).toBe(true);
+    }
+  });
+
+  it('reads the pattern off the gate instead of retyping the suffix', () => {
+    // The derived-never-retyped property, applied to the new term: the predicate
+    // is the union of the two exported patterns and nothing else, so a third
+    // spelling cannot appear in the gate without appearing here.
+    for (const path of [
+      ...records,
+      'tsconfig.tsbuildinfo',
+      'dist/index.js',
+      'dist/a.test.d.ts',
+      'src/a.test.ts',
+      'dist/__tests__/a.d.ts',
+    ]) {
+      expect(isToolingArtifact(path), path).toBe(PUBLISHED_TOOLING_FILE.test(path) || BUILD_RECORD.test(path));
+    }
+  });
+
+  it('does not fire on names that merely contain the word', () => {
+    for (const clean of [
+      'dist/index.js',
+      'dist/tsbuildinfo.js',
+      'dist/tsconfig.tsbuildinfo.js',
+      'dist/a.tsbuildinfo/index.js', // a DIRECTORY so named — anchored to the last segment
+    ]) {
+      expect(isToolingArtifact(clean), clean).toBe(false);
+    }
+  });
+});
+
 // ── 3. the verdicts, including the vacuous one ───────────────────────────────
 
 describe('outputDirOf', () => {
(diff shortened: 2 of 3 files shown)


baozhoutao:
Standing down on `Live E2E (informational)` for this PR, read 2026-09-06T18:38Z.

- **Failing check:** `Live E2E (informational)` — informational, not a required context, and red on `main` and on every open branch today for an upstream reason tracked in #7990 / objectstack#16186. This diff is a release-time gate's convention, its pin test and one doc row; nothing the live E2E exercises.

Every other check on `79fe2c6c0` is judged on its own; this PR flips to ready once the required set converges, then goes to the merge queue by squash.

---

baozhoutao:

- Head `79fe2c6c0`: 28 non-informational checks completed green at 18:50Z (the `ci-wait` reading); `Live E2E (informational)` red per the standing-down note above.
- `ready_for_review` at 18:50:18Z; the post-flip `Governed Surface Queue Guard` run started after that timestamp and completed `success` at 18:50:34Z.
- `added_to_merge_queue` at 18:50:42Z on the REST timeline (read 18:50:48Z). The queue is otherwise empty.

On MERGED: content probe on re-fetched `origin/main` (`export const BUILD_RECORD` in `scripts/check-published-dist-tooling.mjs` and `isToolingArtifact` as the union; the `*.tsbuildinfo` clause in the `ci-cd-pipeline.md` row; control: `PUBLISHED_TOOLING_FILE`'s definition unchanged), then the LANDED note here and the closing stroke on #7003.

---

baozhoutao:
LANDED — merged at 2026-09-06T19:06:24Z as `dd883d169` (squash, via the merge queue), `domain:devx @ objectui` execution seat, PM session `session_01FhBNJcLRZLe8M87VcUgpKr`, R46.

Content probe on re-fetched `origin/main` (`dd883d169`, read 19:06:57Z):
- `export const BUILD_RECORD = /(^|\/)[^/]*\.tsbuildinfo$/;` at `scripts/check-published-dist-tooling.mjs:241` and `isToolingArtifact` as the union at line 249; the `*.tsbuildinfo` clause is in the `ci-cd-pipeline.md` row (1 hit).
- Control: `PUBLISHED_TOOLING_FILE`'s definition is unchanged against `336a9eb8a`.

#7003 closed by the `Fixes` at merge time; the seat strips `pm:dispatched` and clears the assignee there in the same stroke.

---

PR #8145, opened 2026-09-06
fix(types): SchemaRegistry names all three chatbot registrations

### LINKED ISSUE #7704: finding(types): `SchemaRegistry` / `ComponentType` omit `chatbot-enhanced` and `chatbot-floating` — two registered keys the layer can now name
Measured by the `domain:ui` dev seat on objectui#7655 (branch `claude/issue-7655-chatbot-registration-authoring-types`). Filed unassigned. Not addressed there: `packages/types/src/registry.ts` was outside that card's file fence and is held by PR #7662 (objectui#7645), which is still open.

## The fact

`SchemaRegistry` (`packages/types/src/registry.ts`) maps `'chatbot': ChatbotSchema` and has no entry for `chatbot-enhanced` or `chatbot-floating`, although `packages/plugin-chatbot/src/renderer.tsx` registers all three (`:62`, `:271`, `:421` on `origin/main` at `6eebc54b6`). `ComponentType = keyof SchemaRegistry` is the published union, so a consumer discriminating on it is told two registered keys do not exist. `packages/cli/src/utils/known-schema-types.ts:83-84` DOES list both — the CLI keeps its own list precisely because the map does not.

Until objectui#7655 the map had nothing honest to point the two keys at: `ChatbotSchema` pins `type` to `'chatbot'`. With `ChatbotEnhancedSchema` and `ChatbotFloatingSchema` published from the same zero-workspace-dependency layer, the two entries can be both honest and reachable — unlike the `kanban` case objectui#7645 measured, where the honoured type lives in a plugin the layer cannot import.

## Not a duplicate

Dedup ran against the repo-scoped open-issue list via REST (381 open, PRs excluded) with a lit positive control — `chatbot-floating` returns objectui#7655 and objectui#7654. `chatbot-enhanced` returns those two only. `SchemaRegistry` returns objectui#7665 (the docblock over-promise: the VALUE side of the map), objectui#7664 and objectui#6879 — none about missing KEYS. objectui#7665's own fence says not to fold in a sweep of other keys; this is two specific keys whose faces now exist, not a sweep.

## Suggested route

Add the two entries under the `// Complex` group once PR #7662 lands in the same file, and pin at the type level that `keyof SchemaRegistry` includes both keys so they cannot be dropped again silently. Small and self-contained.

Blocked-by: #7662

Refs: objectui#7655 · objectui#7645 · objectui#7665.

Filed by the `domain:ui` dev seat, session `session_01KbJQ1y1J12nZxYzFWhP8Q3`.

os-justin:
Fixes #7704

## What was wrong

`packages/plugin-chatbot/src/renderer.tsx` registers three components — `chatbot` (`:62`), `chatbot-enhanced` (`:241`), `chatbot-floating` (`:379`). `SchemaRegistry` mapped **one** of them. Since `ComponentType = keyof SchemaRegistry` is the published union, a consumer discriminating on it was told two registered keys do not exist.

The asymmetry that shows which half was wrong: `packages/cli/src/utils/known-schema-types.ts:83-84` lists both keys — the CLI keeps its own parallel list *precisely because* the map did not.

## Why the entries can be honest now

This map's value has to be the type the registered renderer honours, and until objectui#7655 there was none to point at: `ChatbotSchema` pins `type` to `'chatbot'`, and each registration's real key set lived in an anonymous `ChatbotSchema & { ... }` intersection local to the renderer file. objectui#7655 published `ChatbotEnhancedSchema` and `ChatbotFloatingSchema` from `@object-ui/types` itself, which makes the two entries simultaneously

- **honest** — each value pins `type` to its own key, and each registration already takes that exact type as its `schema` parameter (`renderer.tsx:256`, `:394`), so the map's value and the renderer's prop type are one declaration (the property the `'kanban'` arm gained in objectui#7664); and
- **reachable** — `@object-ui/types` has zero workspace dependencies and both faces are its own declarations. This is the difference from the `kanban` case objectui#7645 measured, where the honoured type lived in a plugin this layer cannot import. `check:phantom-deps` — the gate that caught that attempt — is green, re-measured on this branch.

## The pin is the real deliverable

Without it this card regresses invisibly, which is how it arose. `packages/types/src/__tests__/schema-registry-chatbot-keys-7704.test.ts` pins it in **two channels**:

- **COMPILE-TIME** (`tsc -p packages/types/tsconfig.test.json`, chained off the package's `type-check` script): the keys survive in `keyof`; each value IS the face its renderer honours; each value's own `type` literal IS its key; and each registry value is the same declaration as its `ComplexSchema` validator arm. Proven to be in that program: `--listFiles` lists the file at position 440 of 595.
- **RUNTIME** (vitest): a TypeScript-AST census of the interface source — so dropping an entry fails vitest too, not only `tsc` — plus each key selecting its own arm through `safeValidateSchema`, with the sibling key refused **at `type`** on otherwise identical documents.

### Ablation — the pin is lit, and it is per-key

Both legs mutate on disk after the implementation was committed, and restore under a trap; the on-disk mutation is proven by a `grep -cF` on the exact removed text (1 → 0) and a `git hash-object` diff, and each restore leg is proven by the blob hash matching `HEAD` again.

| leg | `tsc` | errors in the pin file | vitest | restore |
|---|---|---|---|---|
| drop `'chatbot-enhanced'` | exit 2 | 4, all naming `chatbot-enhanced` (lines 90/96/103/116) | exit 1, 2 census tests fail | blob back to `eb39756db`, `git diff HEAD` empty |
| drop `'chatbot-floating'` | exit 2 | 4, all naming `chatbot-floating` (lines 91/97/104/119) | exit 1, 2 census tests fail | blob back to `eb39756db`, `git diff HEAD` empty |

Neither leg reddens the *other* key's pins — the pins are per-key, not one blanket check.

```
src/__tests__/schema-registry-chatbot-keys-7704.test.ts(90,37): error TS2344: Type 'false' does not satisfy the constraint 'true'.
src/__tests__/schema-registry-chatbot-keys-7704.test.ts(96,60): error TS2339: Property 'chatbot-enhanced' does not exist on type 'SchemaRegistry'.
```

### Cross-package reverse verification, against the rebuilt `.d.ts`

`@object-ui/types` was rebuilt (`dist/registry.d.ts:89-91` carries all three keys) and probed from `packages/plugin-chatbot`, resolving `@object-ui/types` **by package name** so the read goes through `dist`, not the source tree:

- positive leg — `const a: ComponentType = 'chatbot-enhanced'` (and the floating twin, and both `SchemaRegistry[key]['type']` reads): `tsc` exit **0**;
- negative leg — `const bad: ComponentType = 'chatbot-does-not-exist'`: `tsc` exit **2**, `TS2322: Type '"chatbot-does-not-exist"' is not assignable to type 'keyof SchemaRegistry'`.

The negative leg is what makes the positive one mean something: the union is a real literal union, not widened to `string`.

## Verification

Measured on the final commit `8c1c4019d`.

| run | result |
|---|---|
| `pnpm --filter @object-ui/types build` | exit 0, 124 emitted files verified |
| `pnpm exec vitest run packages/types/` | exit 0 — **134 files / 2505 tests passed** |
| `pnpm --filter @object-ui/types type-check` | exit 0 (`tsc --noEmit` + `tsconfig.examples.json` + `tsconfig.test.json`) |
| `check:control-bytes` · `check:phantom-deps` · `check:unreferenced-sources` · `check:doc-types` · `check:self-import` | exit 0 |
| `check-changeset-presence` · `-no-major` · `-fixed` · `-overwrite` | exit 0 |
| `check-type-check-coverage` · `check-lint-coverage` | exit 0 |

Every exit code was captured before any pipe (`cmd > log 2>&1; EXIT=$?`), and each verdict above is the gate's own printed line.

**Declared narrowings** (CI runs the full farm):

- `turbo ls --affected` reports **43 packages** — `@object-ui/types` is the graph root. That set is a graph-shape artifact, not a runtime risk surface: the changed file emits **no runtime code** (`dist/registry.js` is `export {}`, 207 bytes of licence header). The only real surface is type-level, and the only reference to `SchemaRegistry` or `ComponentType` anywhere outside `packages/types` is **two comments** (`plugin-kanban/src/types.ts:28`, `app-shell/src/layout/AppSidebar.tsx:251`) — no downstream code consumes either symbol.
- **eslint**: run on the 2 changed files, `--format json` reports **2 files, 0 errors, 0 warnings**, exit 0. Population per eslint's own config (`files: ['**/*.{ts,tsx}']` minus its `ignores`): **4265** tracked files. Invariance: the config declares no `parserOptions.project` and no `projectService` (0 occurrences), so type-aware linting is off and this diff cannot move the verdict of any file it does not touch.
- `check:sdui-registration-pins` and `check:readme-exports` are **NOT MEASURED**, not red — both are prerequisite failures needing builds this seat did not run. The first prints `❌ No console build to weigh at apps/console/dist/assets. This is exit 2, not a pass`. The second's 484 findings are 481 of the form `its type entry ./dist/index.d.ts is not on disk -- run pnpm build first` plus 3 floor failures caused by the same absence (`packagesRead: found 2, floor is 25`); **zero** of them name `packages/types`. Neither gate can be affected by a diff that adds no registration and no README line.

## Scope

Two keys whose authoring faces now exist — **not** a sweep of the map's other entries, which objectui#7665 holds, and this PR does not answer objectui#7645's generalization question about whether *every* `SchemaRegistry` value must be the type its renderer honours.

Region fence honoured: exactly three files, **additive only** (288 insertions, 0 deletions). `packages/types/src/complex.ts` and `zod/complex.zod.ts` are untouched (imported from only) — they belong to the concurrent objectui#7703, which retires members of `ChatbotSchema`, a different interface from the two faces named here. No semantic collision was measured: `ChatbotEnhancedSchema` and `ChatbotFloatingSchema` pick only `ChatbotSharedKey` (plus `maxHeight` / `processVisibility`), and the six members #7703 retires are the legacy ones no registration reads by name.

## Review posture

⛔ **Stays DRAFT.** `needs:contract-review` is attached: adding two entries widens the published `ComponentType` union, which is a public-face change, not a spelling. It does not enqueue while that label is on it. `check-governed-queue-guard --test` confirms the diff is **not** a governed surface (3 paths checked, none matched), so the draft hold here is the contract-review label alone.

Changeset: `@object-ui/types: minor` (`major` is forbidden by repo convention, and `check-changeset-no-major` is green).

---

### DIFF
--- a/.changeset/7704-schema-registry-chatbot-keys.md
+++ b/.changeset/7704-schema-registry-chatbot-keys.md
@@ -1,0 +1,42 @@
+---
+'@object-ui/types': minor
+---
+
+`SchemaRegistry` names all three `plugin-chatbot` registrations, so the published
+`ComponentType` union does too (objectui#7704).
+
+`packages/plugin-chatbot/src/renderer.tsx` registers three components — `chatbot`,
+`chatbot-enhanced` and `chatbot-floating` — and `SchemaRegistry` mapped one of them.
+Since `ComponentType = keyof SchemaRegistry` is the published union, a consumer
+discriminating on it was told two registered keys do not exist: an author narrowing a
+node by `ComponentType`, or writing a `Record<ComponentType, …>` table, had no arm for
+either. The asymmetry that showed which half was wrong is that
+`packages/cli/src/utils/known-schema-types.ts` keeps its own parallel list containing
+both keys, precisely because this map did not.
+
+**Additive.** Two keys join the map under its `// Complex` group; no existing entry
+changes, `'chatbot'` still maps to `ChatbotSchema`, and nothing is removed. The union
+widens, which cannot break a consumer that produces `ComponentType` values and can only
+help one that consumes them — except an exhaustive `Record<ComponentType, …>` or
+`switch`, which now needs the two new arms. Nothing in this repo has one: the only
+consumer of `ComponentType` outside its own declaration is a pin test.
+
+**Why the entries can be honest now.** This map's value has to be the type the
+registered renderer honours, and until objectui#7655 there was none to point at —
+`ChatbotSchema` pins `type` to `'chatbot'`, and each registration's real key set lived
+in an anonymous `ChatbotSchema & { … }` intersection local to the renderer file.
+objectui#7655 published `ChatbotEnhancedSchema` and `ChatbotFloatingSchema` from this
+package, and both registrations already take them as their `schema` parameter, so the
+map's value and the renderer's prop type are one declaration — the same property the
+`'kanban'` arm gained in objectui#7664. Being declared here also makes them reachable:
+`@object-ui/types` has zero workspace dependencies, which is exactly what blocked the
+`kanban` case objectui#7645 measured, where the honoured type lived in a plugin.
+
+Pinned in `src/__tests__/schema-registry-chatbot-keys-7704.test.ts` in two channels —
+compile-time (`tsc -p tsconfig.test.json`: the keys survive in `keyof`, each value is
+the face its renderer honours, and each value's own `type` literal is its key) and
+runtime (a TypeScript-AST census of the interface source, plus each key selecting its
+own arm through `safeValidateSchema`).
+
+Scope: these two keys, whose authoring faces now exist — not a sweep of the map's other
+entries, which objectui#7665 holds.
--- a/packages/types/src/__tests__/schema-registry-chatbot-keys-7704.test.ts
+++ b/packages/types/src/__tests__/schema-registry-chatbot-keys-7704.test.ts
@@ -1,0 +1,209 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * `SchemaRegistry` names all three `plugin-chatbot` registrations, so the
+ * published `ComponentType` union does too (objectui#7704).
+ *
+ * ## The defect this pins shut
+ *
+ * `packages/plugin-chatbot/src/renderer.tsx` registers three components —
+ * `chatbot` (`:62`), `chatbot-enhanced` (`:241`), `chatbot-floating` (`:379`).
+ * `SchemaRegistry` mapped one of them. `ComponentType = keyof SchemaRegistry`
+ * is the published union, so a consumer discriminating on it was told two
+ * registered keys DO NOT EXIST. The asymmetry that proves which half was wrong:
+ * `packages/cli/src/utils/known-schema-types.ts:83-84` lists both keys — the
+ * CLI keeps its own parallel list precisely because this map did not have them.
+ *
+ * The map could not have carried them earlier and stayed honest. This map's
+ * value has to be the type the registered renderer honours, and until
+ * objectui#7655 there was none to point at: `ChatbotSchema` pins `type` to
+ * `'chatbot'`, and each registration's real key set lived in an anonymous
+ * `ChatbotSchema & { ... }` intersection local to the renderer file.
+ * objectui#7655 gave each registration one named authoring face, declared in
+ * THIS package — which is what makes the two entries simultaneously
+ *
+ *   - HONEST: each value pins `type` to its own key, and the registration for
+ *     that key takes that exact type as its `schema` parameter
+ *     (`renderer.tsx:256`, `:394`), so the map's value and the renderer's prop
+ *     type are one declaration; and
+ *   - REACHABLE: `@object-ui/types` has zero workspace dependencies, and both
+ *     faces are its own declarations. This is the difference from the `kanban`
+ *     case objectui#7645 measured, where the honoured type lived in
+ *     `@object-ui/plugin-kanban` — naming it from here would have been a
+ *     phantom dependency and a cycle, and it was resolved the other way, by
+ *     moving the dialect down into this package (objectui#7664).
+ *
+ * ## Scope — two keys, not a sweep
+ *
+ * objectui#7704 is these two keys, whose faces now exist. Whether EVERY
+ * `SchemaRegistry` value must be the type its renderer honours is
+ * objectui#7665's question and is deliberately not asked or answered here.
+ *
+ * ## Two channels, stated so nobody reads the wrong one
+ *
+ * The `export type assertion…` block below is COMPILE-TIME: it is checked by
+ * `tsc -p packages/types/tsconfig.test.json` (this package's `type-check`
+ * script chains it) and erased before vitest runs. A green vitest run is NOT
+ * evidence about it — the same split `chatbot-registration-authoring-faces-7655
+ * .test.ts` and `kanban-plugin-dialect-authoritative-7664.test.ts` document.
+ * The `it` blocks are the RUNTIME channel: a source census of the interface
+ * (so deleting an entry fails vitest too, not only `tsc`) and the accept sets
+ * of the validator the CLI applies.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+import ts from 'typescript';
+import type {
+  SchemaRegistry,
+  ComponentType,
+  ComplexSchema,
+  ChatbotSchema,
+  ChatbotEnhancedSchema,
+  ChatbotFloatingSchema,
+} from '../index';
+import {
+  ChatbotEnhancedSchema as ChatbotEnhancedZod,
+  ChatbotFloatingSchema as ChatbotFloatingZod,
+  safeValidateSchema,
+} from '../zod/index.zod';
+
+/* -------------------------------------------------------------------------- */
+/* Compile-time pins — read by tsc via tsconfig.test.json, not by this run.    */
+/* -------------------------------------------------------------------------- */
+
+type Assert<T extends true> = T;
+type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2 ? true : false;
+type IsAny<T> = 0 extends 1 & T ? true : false;
+
+// Non-vacuity controls: `any` on either side satisfies every `extends` below
+// while checking nothing, and `Equal<any, X>` is `false`.
+type _RegistryIsReal = Assert<Equal<IsAny<SchemaRegistry>, false>>;
+type _EnhancedFaceIsReal = Assert<Equal<IsAny<ChatbotEnhancedSchema>, false>>;
+type _FloatingFaceIsReal = Assert<Equal<IsAny<ChatbotFloatingSchema>, false>>;
+
+// 1. THE CARD. The published union yields both keys. `Extract` collapses to
+//    `never` the moment an entry is dropped, which fails this loudly — the
+//    `_KeyKept` shape objectui#7645/#7664 carry for `'kanban'`.
+type _EnhancedKeyPublished = Assert<Equal<Extract<ComponentType, 'chatbot-enhanced'>, 'chatbot-enhanced'>>;
+type _FloatingKeyPublished = Assert<Equal<Extract<ComponentType, 'chatbot-floating'>, 'chatbot-floating'>>;
+
+// 2. Each value IS the face its registration honours — not `ChatbotSchema`
+//    (which `Equal` rejects: it pins `type` to `'chatbot'`), and not a
+//    `BaseSchema & { type }` weakest-true-claim placeholder.
+type _EnhancedValueIsItsFace = Assert<Equal<SchemaRegistry['chatbot-enhanced'], ChatbotEnhancedSchema>>;
+type _FloatingValueIsItsFace = Assert<Equal<SchemaRegistry['chatbot-floating'], ChatbotFloatingSchema>>;
+
+// 3. Honesty, stated independently of #2: each value's own `type` literal IS
+//    the key it is filed under. This is the property objectui#7645 found
+//    violated for `'kanban'`, and the reason these two keys could not be added
+//    before objectui#7655 minted the faces.
+type _EnhancedValueIsHonest = Assert<Equal<SchemaRegistry['chatbot-enhanced']['type'], 'chatbot-enhanced'>>;
+type _FloatingValueIsHonest = Assert<Equal<SchemaRegistry['chatbot-floating']['type'], 'chatbot-floating'>>;
+// …and the honesty check can fail — the pre-#7655 value would have been
+// `ChatbotSchema`, whose `type` is `'chatbot'`, not either new key.
+type _HonestyCanFail = Assert<Equal<Equal<ChatbotSchema['type'], 'chatbot-enhanced'>, false>>;
+
+// 4. The `'chatbot'` entry is untouched: this card ADDS two keys, it does not
+//    re-point the existing one.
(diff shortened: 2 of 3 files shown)


os-justin:
Contract review — **PASS** — PR #8145 (objectui#7704), head `8c1c4019d`

Reviewed-by: `domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`, in-seat at **`opus` = `TIER_DEFAULT`**.
⚠️ Stated exactly, because this seat mislabelled its own review tier earlier in this session and corrected it publicly on #8077: `CONTRACT_REVIEW_TIER` is **`claude-fable-5-1`**, **measured unavailable** to this session (`rate_limit HTTP 429`); the quota-exhaustion exemption permits falling back to `TIER_DEFAULT` and ⛔ no lower. This review ran at `opus`. ⛔ It is **not** described as running at contract-review tier, because it did not.

Diff read by ref against `origin/main` = `21d7989fb`: three files, **288 insertions, 0 deletions**.

## ① Derived judgments — every accept-set and public-surface change named and judged

**Direction: WIDENS**, and the changeset says so rather than eliding it. `ComponentType = keyof SchemaRegistry` gains two members. The changeset names the one genuinely breaking consumer shape — an exhaustive `Record<ComponentType, …>` or `switch`, which now needs two arms — instead of claiming additive-therefore-safe. ⭐ That is the honest form of this argument, and it is the half most "additive" changesets omit. The dev then **measured** the in-repo population: the only reference to `SchemaRegistry` or `ComponentType` anywhere outside `packages/types` is **two comments** (`plugin-kanban/src/types.ts:28`, `app-shell/src/layout/AppSidebar.tsx:251`) — no code consumes either symbol. Correct.

**Honesty of the two entries — verified independently by this seat, ⛔ not taken from the report.** Anchors re-measured on `21d7989fb`:

```
packages/plugin-chatbot/src/renderer.tsx:241  ComponentRegistry.register('chatbot-enhanced',
packages/plugin-chatbot/src/renderer.tsx:379  ComponentRegistry.register('chatbot-floating',
packages/plugin-chatbot/src/renderer.tsx:256  ({ schema, … }: { schema: ChatbotEnhancedSchema;  … })
packages/plugin-chatbot/src/renderer.tsx:394  ({ schema, … }: { schema: ChatbotFloatingSchema;  … })
```

⇒ each registration already takes as its `schema` parameter exactly the type the map now names. The map's value and the renderer's prop type are **one declaration** — the property the `'kanban'` arm gained under #7664, and the property whose absence #7645 carded. Reachability re-confirmed too: both faces are declared in `packages/types/src/complex.ts` (`:1308`, `:1390`), so the zero-workspace-dependency layer imports no plugin, and `check:phantom-deps` — the gate that caught the `kanban` attempt — is green on this branch.

**⭐ The cross-slot fence, which is the one thing I would not accept on assertion.** Both faces are `Pick<ChatbotSchema, …>`, and slot 3/5 of this same wave (**objectui#7703**) is concurrently retiring six `ChatbotSchema` members with `?: never` + `retirementTombstone()`. So "no semantic collision" is a real claim with a real failure mode: a tombstone landing on a picked key would propagate straight into both faces named here. Measured:

`ChatbotSharedKey` (`complex.ts:1254-1274`) is an **explicit literal union of 20 keys**. #7703's six targets — `loading`, `showAvatars`, `userAvatar`, `assistantAvatar`, `markdown`, `height` — appear in it **0 times**. ⚠️ **Lit control on the same read: 20 quoted keys parsed out of that union**, so the zero is a reading and ⛔ not a dark grep.

⇒ the `Pick` cannot carry #7703's tombstones. The fence holds **semantically**, not merely textually — and it holds because `ChatbotSharedKey` is a hand-written literal union rather than a derived `Exclude<keyof …>`, which would have coupled the two slots. Confirmed.

**Pin quality — this is the card's actual deliverable and it is done properly.**
- Two channels, and ⭐ the docblock **states which is which**, naming the precedent files, so nobody reads a green vitest as evidence about the erased compile-time block. That confusion is a recurring cost in this repo; pre-empting it in the file is the right place.
- Non-vacuity controls (`_RegistryIsReal` / `_EnhancedFaceIsReal` / `_FloatingFaceIsReal` via `IsAny`) — `any` on either side would satisfy every `extends` while checking nothing.
- ⭐ `_HonestyCanFail` proves the honesty assertion is falsifiable: the pre-#7655 value `ChatbotSchema['type']` is `'chatbot'`, ⛔ not either new key. An honesty pin that cannot fail is decoration.
- The runtime census carries its own **lit control** (`keys.length > 60`, contains `kanban`, contains `chatbot`, `valueOf('chatbot') === 'ChatbotSchema'`) **and** a can-fail leg (an absent interface name throws rather than reading empty). Both negatives would otherwise be vacuous.
- The validator legs refuse the sibling key **at path `type`** on otherwise-identical documents, with the accept-its-own-key control alongside — so the refusal is attributable to the discriminant and nothing else.
- **Ablation is per-key**: dropping either entry reddens 4 `tsc` errors and 2 vitest tests naming only that key, and ⛔ does not redden the other's pins. Mutation proven on disk (`grep -cF` 1→0 plus blob-hash movement off `HEAD`), restore proven by blob equality and an empty `git diff HEAD`, under a trap.
- **Cross-package reverse verification through the rebuilt `.d.ts`**, resolving by package name so the read goes through `dist`, with a **negative leg** (`'chatbot-does-not-exist'` → TS2322) that is what makes the positive leg mean anything — it proves the union is a real literal union, not widened to `string`.

## ② semver

`@object-ui/types: minor`, `major` forbidden by repo convention and `check-changeset-no-major` green. The body carries the measurement, the asymmetry that identifies which half was wrong (the CLI's parallel list at `known-schema-types.ts:83-84`), why the entries could not have been honest before #7655, and the explicit scope fence against #7665's sweep. Consistent.

## ③ Boundary flags

- Region fence **honoured**: `complex.ts` and `zod/complex.zod.ts` imported from only; no `zod-mirror-parity` ledger row added; three files exactly. ✅
- `open_questions`: none. `out_of_scope_findings`: none.
- Declared narrowings accepted, and both are argued structurally rather than waved: the 43-package `turbo ls --affected` set is a graph-shape artefact (`dist/registry.js` is `export {}`, 207 bytes of licence header — the file emits no runtime code); eslint narrowed with its own population (4265 files), its own `--format json` verdict, and an **invariance argument** (no `parserOptions.project`, no `projectService` ⇒ type-aware linting off ⇒ this diff cannot move an untouched file's verdict). ⭐ That third leg is what turns a narrowing into a valid one.
- `check:sdui-registration-pins` and `check:readme-exports` declared **NOT MEASURED, not red** — both prerequisite failures needing a full workspace build, with the reasons quoted and the observation that **zero** of `readme-exports`' 484 findings name `packages/types`. ✅ Correctly refusing to call a precondition failure a pass.
- ⭐ The report **corrected its own `mcp_calls` count** (the comment said 8; the accurate total is 9, because the read-back post-dated the write). Unprompted self-correction on a number nobody would have checked.
- ⚠️ **One citation drift, recorded and ⛔ deliberately NOT patched.** The pin docblock and PR body cite the `chatbot` registration at `renderer.tsx:62`; it is at **`:60`**. The two keys this card actually adds are cited exactly right (`:241`, `:379`, `:256`, `:394`). A two-line drift in a comment, on the one key this diff does not touch, still lands a reader inside the same `ComponentRegistry.register('chatbot',` call. ⇒ a patch round and a full CI cycle would be disproportionate to the defect. Noted here so the record is accurate rather than silently tidy.

## Verdict

**PASS.** Same stroke: `needs:contract-review` is cleared, with this comment as its provenance.

⛔ **Not landing yet.** CI has not converged — enumerated **32** check runs against `total_count: 32`, of which **9 are still `in_progress`** (`Lint`, Test shards 1–4, `Type Check`, `Bundle Analysis`, `Doc Snippet Type Check`, `README Export Check`) plus `Live E2E (informational)`, and **zero failures so far**. ⚠️ `Lint` is specifically not yet readable, and this seat will ⛔ not flip ready before reading that job explicitly — a merits-only review in R1 missed an error-severity unused import exactly this way. Landing is armed on a re-check; if the head moves the label is re-hung and the delta re-reviewed.

`Live E2E (informational)` will fail: published `@objectstack/plugin-auth` floats `@better-auth/core` to `1.7.3`, which dropped `createLocalAccountIssuer` ⇒ no `sys_*` table ⇒ 300s readiness timeout. Carded **objectstack#16186**; the red is **by design** per objectui#7689, whose triage ⛔ forbids reverting the pin to get green. ⚠️ If the signature differs from that one, it is a different failure and is this PR's until proven otherwise.

---

os-justin:
Landing — PR #8145 (objectui#7704), head `8c1c4019d`, 2026-09-06T20:4xZ.

**Pre-landing three:**

① **Contract review PASS** (comment 5561967644) at `opus`; **`needs:contract-review` cleared** in that stroke, with this as its provenance. Head unchanged since the review.

② **CI converged, enumerated 32 against `total_count: 32`** — **28 success · 3 skipped · 1 failure**. ⛔ Zero `in_progress`.

⭐ **`Lint` read explicitly and green** (`completed_at` 20:30:42Z). Stated because a merits-only review in R1 called #8089 "correct on the merits" while `Lint` was red and it *was* that diff's; this seat does not flip ready without that job in hand.

⭐ Two gates the dev declared **NOT MEASURED** locally — `check:readme-exports` and `check:sdui-registration-pins` — both came back **green on CI**. ⇒ the declared narrowing was honest: it was a prerequisite failure needing a full workspace build, ⛔ not a red being reported as unmeasurable. That is the distinction the declaration existed to preserve, and CI confirmed it.

The single failure is **`Live E2E (informational)`**, and its signature was **read and matched line for line** against the known break rather than assumed — `[live-backend] not ready after 300s` · `⚠ AuthPlugin failed to load: … does not provide an export named 'createLocalAccountIssuer'` · `WARN CORE: Core service missing … auth` · the same four `no such table` system objects · `✓ Server is ready` with 42 plugins and 132 seeded rows · 31 boot warnings · `##[error]Process completed with exit code 1`. Identical to #8142's. Carded **objectstack#16186**; the red is **by design** per objectui#7689, whose triage ⛔ forbids reverting the pin to get green. ⛔ No re-run spent — this is the established not-mine case, not a flake.

③ **Ready flipped, auto-merge SQUASH enabled** (⚠️ the API echoes an empty `method` even when it applied — verify by queue ref and parent-count, ⛔ not by that echo). Queue probe next.

**⭐ Landing control established BEFORE the queue picks it up**, so the post-merge confirmation is a reading and ⛔ not a hopeful grep:

| read | value |
|---|---|
| `'chatbot-enhanced': ChatbotEnhancedSchema;` on the PR branch | **1** |
| the same grep on `origin/main` | **0** |
| `packages/types/src/registry.ts` line count on `main` (instrument live?) | **209** |
| `'chatbot-floating': ChatbotFloatingSchema;` on the branch | **1** |

⇒ the zero is a reading on a live file. On merge this seat re-runs exactly these and expects 1 / 1.

**One judgement recorded, ⛔ not silently tidied.** The pin docblock and PR body cite the `chatbot` registration at `renderer.tsx:62`; it is at **`:60`**. The two keys this card actually adds are cited exactly (`:241`, `:379`, `:256`, `:394`). A two-line drift in a comment, on the one key this diff does not touch, still lands a reader inside the same `ComponentRegistry.register('chatbot',` call ⇒ a patch round and a full CI cycle would be disproportionate. Recorded so the record is accurate rather than quietly clean.

`Fixes #7704`, so the card closes on merge and this seat strips `pm:dispatched` and clears the assignee in the same stroke.

---

PR #8150, opened 2026-09-06
fix(fields,i18n): key LocationField's residue refusal sentence, answering arity from precedent

### LINKED ISSUE #6888: finding(fields): LocationField's THIRD refusal sentence (the #6715 residue arm) is still hard-coded English after the #6755 ruling keyed the other two
Filed unassigned by the dev seat implementing objectui#6755, as an out-of-scope finding. Recording, not choosing.

Dedup before filing: repo-scoped REST list of all 252 open issues (3 pages, `state=open`), title+body grepped for `6715` / `residue` / `refusedResidue` / `is not a number` / `plain decimals` — no match. Control: grepping the same corpus for `LocationField` returns objectui#6755, so the read was live rather than empty.

## What

objectui#6755 ruled (2026-08-29, maintainer) that a widget's own refusal sentence goes through `useFieldTranslation` + `FIELD_DEFAULTS`, and locked scope to **three** sentences: `ObjectField`'s `Invalid JSON` and `LocationField`'s **two** refusal sentences.

`LocationField` now has **three**. The card was measured on 2026-08-29 06:24 and the ruling written at 14:53; objectui#6715 added a third refusal arm — `residue`, for a half that is only partly a number (`12abc, 34`) — whose sentence is built by `refusedResidueMessage()` in `packages/fields/src/widgets/LocationField.tsx`:

```
`Not saved: ${named} ${verb}. Enter plain decimals (example: 30.2741, 120.1551).`
```

with `verb` being `is not a number` / `are not numbers` and `named` being `latitude "12abc"`-style. It is a string literal in the source, exactly like the two the ruling keyed.

## Why it is filed rather than folded into the #6755 PR

The ruling says 「范围锁定这 3 句」 and names one same-class instance (objectui#4752) that must NOT ride along. This is a fourth sentence the ruling could not have seen — but reading a scope lock as "except when I find one more" is the dev seat's call to make, not mine. So the #6755 PR keys the three ruled sentences and leaves this one, and no test in that PR pins the residue arm's English text (pinning it would read as endorsement).

## The measured consequence of leaving it

All three arms render through the SAME `<p>` element and the same `refusalError` state. After #6755 lands, that one line speaks the reader's language when the format or range arm fired and English when the residue arm did — the same shape objectui#4028 was reported for ("four Chinese labels around one English one"), inside a single widget.

## Cost if it is ruled in

Mechanically identical to what the #6755 PR already does, one arm wider: one key (`fields.location.refusedResidue` — an interpolated one: the half names and the quoted text are data), ten pack entries, and it becomes bound by `check:i18n-drift` like the other three. The English default stays byte-identical, so objectui#6715's own pins and `plugin-form`'s `ObjectForm.locationResidue.test.tsx` are unaffected.

Not measured: whether the interpolation should carry `named`/`verb` as holes or be keyed per arity (`_one` / `_other`). That is a real authoring choice — `verb` is grammar, and a pack whose plural rules differ cannot use an English verb form as a hole — and it is why this is not a two-line copy of the other two keys.

### Related

objectui#6755 (the ruling and the PR that keys the other two) · objectui#6715 (added this arm) · objectui#6716 (added the first two) · objectui#4752 (the same class in `@object-ui/components`, still its own card)

os-justin:
Fixes #6888

`LocationField`'s THIRD refusal sentence — the residue arm objectui#6715 added — now reaches the locale packs, like the three objectui#6755 keyed.

Triage ruled this in-lane execution rather than a new adjudication (comment 5547164710): #6755's 「范围锁定这 3 句」 was a PR-sprawl constraint written 2026-08-29 14:53, and #6715 added this arm afterwards, so it is new terrain rather than excluded terrain. Triage's overturn condition — reading that scope lock as deliberately freezing the tree — was considered and **not** exercised: nothing in #6755's PR, its tests, or the source comments freezes anything. `LocationField.tsx` itself carried the opposite instruction in a comment, naming this card and the open question by number.

## The defect this removes

All three refusal arms render through the **same** `p` element and the same `refusalError` state (`LocationField.tsx`, one `refusalError && ...` render site). Verified, not assumed. So after #6755 landed, that one line spoke the reader's language when the format or range arm fired and English when the residue arm did — objectui#4028's reported shape compressed into a single sentence position.

## ⭐ The arity question, answered with a survey rather than defaulted

The card, triage and the dispatch all flagged one thing as a real authoring choice: should `named` / `verb` be interpolation holes, or should the key split per arity? I was told to measure what this channel already does, follow a precedent if one exists and say which widget, and stop and report if there is none.

**There is a precedent, it is explicit, and it is documented in source with its reason.** This repo's plural convention is **two sibling keys, `X` and `XOne`, branched at the call site — never i18next's `_one`/`_other` suffixes.**

| Instance | Where |
|---|---|
| `lookup.recordCount` / `lookup.recordCountOne` | `RecordPickerDialog.tsx:1209` — **the same package and the same `FIELD_DEFAULTS` map this card writes into** |
| `list.recordCount` / `list.recordCountOne` | `plugin-list/src/ListView.tsx:4228` |
| `detail.reactionCount` / `detail.reactionCountOne` | `plugin-detail/src/ReactionPicker.tsx:55` |
| `detail.relatedRecords` / `detail.relatedRecordOne` | named by that same comment |

`ReactionPicker.tsx:55` states the reason in the source:

> Two keys, NOT an i18next `_one`/`_other` pair — this repo's own plural convention. A `_one` suffix would break `all-locales-key-parity`: zh/ja/ko have no separate singular form, so those packs would legitimately omit the `_one` half and the gate would read that as a missing key (objectstack#5430).

`all-locales-key-parity.test.ts` adds a second, independent reason (objectui#3863): an i18next plural family also needs a base key, because `ru` has four CLDR categories and `ar` six, no pack in this repo defines `_few`/`_many`/`_two`/`_zero`, and those categories silently fall back to English. The sibling pair has neither problem.

**So: precedent followed, `RecordPickerDialog`'s.** Keys are `fields.location.refusedResidue` (both halves) and `fields.location.refusedResidueOne` (one half), picked at the call site.

The same reasoning is carried one step further for the parts the card called out as also translatable:

- **The verb is gone from the code.** It was `is not a number` / `are not numbers`, chosen in TypeScript. It now lives inside each pack's own two values. `ar` uses its **dual** — `ليسا رقمين` for two halves against `ليس رقمًا` for one — which is exactly the form an English verb through a hole could never have produced, and it is pinned by name in the test.
- **The conjunction is gone too.** `' and '` was English grammar joining the halves; each pack now writes its own inside the two-half value.
- **The coordinate nouns are keyed**, `fields.location.latitude` / `fields.location.longitude` — once each, interpolated into **both** arities, so no locale holds two spellings of the same word. Every pack already spelled them inside its own `refusedFormat` (`纬度`/`经度`, `широта`/`долгота`, `خط العرض`/`خط الطول`), so this is not new translation work.

The only holes carrying untranslated data are `{{text}}` / `{{otherText}}` — the characters the person actually typed.

## Widget population, re-measured with a control

The card and the routing comment both say **11 widgets**; the dispatch measured **21 files** and told me not to carry the 11 forward. Both numbers are answering different questions, and neither is the count of widgets:

- **21** files under `packages/fields/src/widgets/` match `useFieldTranslation` — of which **7 are `.test.` files** and **1 is `useFieldTranslation.ts` itself**, the definition.
- **13** non-test `.tsx` consumers remain.
- **9** of those are reachable from the `fieldWidgetMap` registry, i.e. actually renderable field types: `AddressField`, `FilterConditionField`, `LocationField`, `LookupField`, `ObjectField`, `ObjectRefField`, `RecipientPickerField`, `SelectField`, `TagsField`.
- The other 4 are helper components, not registered widgets: `CapabilityMultiSelectField`, `OptionsEmptyState`, `PeoplePicker`, `RecordPickerDialog`.

Controls for that read, in the same pass: grepping the identical corpus for `createSafeTranslation` returns 7 files rather than 21 (so the corpus is not matching everything), and `LocationField` is present in the 21 (so it is not matching nothing). Registry denominator: the map exposes 45 type keys over 42 distinct widget files.

⭐ The precedent the arity answer rests on comes from `RecordPickerDialog` — one of the four **non**-registered helpers — so the answer would have been missed by a survey that counted only the 9 registered widgets.

## Clause-② re-measured, and it agrees with the PM's assessment

- **English byte-identical:** `check:i18n-drift` reports `0 en value(s) changed (4 key(s) added, 0 removed)` — the gate's own words, not my claim. Both arities are additionally asserted byte-for-byte in the new test, and objectui#6715's `LocationField.strictNumeric.test.tsx` and `plugin-form`'s `ObjectForm.locationResidue.test.tsx` pass **untouched**.
- **No schema accept/reject movement:** nothing here touches parsing, `LocationValueSchema`, or the refusal decision. `parseDraft` is unchanged apart from `ResidueHalf.label` narrowing from `string` to the `COORDINATE_LABELS` union — a tightening internal to the module, so a future half cannot silently select the wrong noun key.
- **No exported symbol added or removed:** `refusedResidueMessage` and `coordinateName` are both module-private.

## One ratchet moved, deliberately

`packages/i18n/src/__tests__/de-quote-pairing-3876.test.ts` counts matched `„…“` spans in the `de` pack: 59 → **62**, `rdq` still 0. My three added German values contribute exactly three **matched** pairs (one in the one-half sentence, two in the two-half one); the two noun keys carry no quotes at all. Updated in the file's own idiom, with the same per-card note every prior entry carries. `rdq` staying at 0 is the load-bearing half and it did.

## Tests

- **New:** `packages/fields/src/__tests__/LocationField.residueI18n-6888.test.tsx` (26 tests). Shape inherited from `widget-diagnostics-i18n-6755.test.tsx`: positive **and** negative assertions together (the English default *is* the fallback, so "says the Chinese thing" alone cannot tell keyed from hard-coded), plus that file's `AddressField` zh control so a dead provider cannot read as a pass. Asserts both arities in `zh`/`ja`/`ar`, both arities byte-identical in `en` and provider-less, the absence of all five English fragments (`is not a number`, `are not numbers`, `' and '`, `latitude`, `longitude`), that every pack's two arity values differ **in its own language**, the `ar` dual by name, and that one `p` still serves all arms.
- **Green:** `packages/fields` full suite 136 files / 2220 tests; `packages/i18n` + `plugin-form`'s residue suite + `app-shell`'s `defaults-maps-mirror-en-pack` 64 files / 1065 tests; `type-check` for both `@object-ui/fields` and `@object-ui/i18n`.
- **Gates:** `check:i18n-keys`, `check:i18n-drift`, `check:i18n-dead-keys`, `check:control-bytes`, `check:unreferenced-sources` — all exit 0 on the final head. `check:i18n-keys` confirms `every call site passes exactly the arguments that value has holes for`, which is the interpolation contract for the four new keys.
- **Ablation, both legs, from the committed state:** restoring the pre-#6888 literal turned the new file **13 failed / 13 passed**. The 13 that still passed are precisely the `en`, provider-less and pack-value assertions — which is the point: the English pins are green in *both* states and could never have caught this, so only the translated assertions can. Mutation proven on disk before reading the result (injected marker count 1, keyed-call count 0, blob hash moved off the `HEAD` blob); restore proven by state rather than exit code (`git diff HEAD` empty, blob hash back to `046c63c3`, marker count 0). `vitest` resolves `@object-ui/i18n` through the root config's alias to `packages/i18n/src`, so the source mutation is what ran.
- **Lint, measured not narrowed:** `eslint . --no-inline-config --format json` completed over the whole repo at `0aa5812ac` — **4384 files**, the universe read from eslint's own config resolution rather than estimated. My 14 linted files: **0 errors**, 3 warnings (two `no-explicit-any` field fixtures matching the sibling pin file's existing pattern, and one pre-existing `react-hooks/set-state-in-effect` that only surfaces because `--no-inline-config` suppresses its existing `eslint-disable-next-line`, in a region my diff does not touch). The repo's 94 errors are all in files this branch does not modify, and the config enables no type-aware linting, so this diff cannot move any untouched file's verdict.

## Out-of-scope findings, filed unassigned rather than fixed here

- **#8148** — `badInputMessage` in `numberBadInput.tsx` is a **fifth** hard-coded English refusal sentence, and the widest: one shared literal reaching `NumberField`, `CurrencyField`, `PercentField` and `GeolocationField`. `GeolocationField` sits beside `LocationField`, so after this PR two adjacent coordinate widgets refuse bad input in two different languages. It has no arity question, so it *is* the two-line copy this card was not.
- **#8149** — `LocationField`'s own `'latitude, longitude'` placeholder fallback is still an English literal, now directly above a refusal that says 纬度 / 经度 in the reader's language. Filed rather than fixed because #6755's ruled principle is about a widget's *refusal sentence*, and extending it to a placeholder is a fresh judgement.

Neither is touched by this PR. objectui#4752 (`@object-ui/components`) is untouched as required.

---


os-justin:
Contract review — **PASS** — PR #8150 (objectui#6888), head `0aa5812ac`

Reviewed-by: `domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`, in-seat at **`opus` = `TIER_DEFAULT`**. ⛔ Not described as contract-review tier: `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` and is measured unavailable to this session (429). This review ran at `opus`.

Diff read by ref against `origin/main`: 15 files, 521 insertions / 14 deletions.

## ⭐ The arity answer — verified independently, ⛔ not accepted from the report

This was the dispatch's binding clause: answer it explicitly from a survey, or stop and report. The report claims a precedent. I re-measured it on `origin/main` rather than take it:

```
packages/fields/src/widgets/useFieldTranslation.ts:30   'lookup.recordCount':    '{{count}} records'
packages/fields/src/widgets/useFieldTranslation.ts:31   'lookup.recordCountOne': '1 record'
packages/plugin-list/src/ListView.tsx:656               'list.recordCountOne'
packages/plugin-detail/src/useDetailTranslation.ts:80   'detail.reactionCountOne'
packages/plugin-detail/src/ReactionPicker.tsx:55        the reason, stated in source
```

⇒ the `X` / `XOne` sibling pair is **already in the very `FIELD_DEFAULTS` map this card writes into**, two lines from where the new keys land. **Lit control on the same corpus:** grepping the locale packs for an i18next `_one` / `_other` family returns **zero** — so the convention is not merely present, it is exclusive.

⭐ **And the methodological catch is the best thing in this round.** The precedent lives in `RecordPickerDialog`, which is one of the **four unregistered helpers**, not one of the nine registered widgets. A survey scoped to "widgets" — the obvious scoping, and the one the card's own "11 widgets" framing invites — would have returned no precedent and led to a stop-and-report, or worse, to inventing `_one`/`_other`. The dev noticed that its denominator choice determined its answer, and said so. ⇒ this is the difference between a survey and a search that confirms what you expected.

The card's and triage's **"11 widgets" is also correctly dismantled** rather than repeated or silently replaced: 21 matching files − 7 `.test.` − 1 definition = 13 non-test consumers, of which 9 are registry-reachable and 4 are helpers. Controls on the same pass: `createSafeTranslation` returns 7 not 21 (the corpus is not matching everything) and `LocationField` is among the 21 (not matching nothing). My dispatch told it not to carry the 11 forward; it didn't carry my 21 forward either, which is the right response to a PM number.

## ① Derived judgments

**Clause-②: NO — and I record the one case I actually had to think about rather than asserting it.** The narrow reasons are sound: English values byte-identical (`check:i18n-drift`'s own words, `0 en value(s) changed (4 key(s) added, 0 removed)`); no parse or accept/reject movement; both new functions module-private. I independently confirmed the last one — `type ResidueHalf` carries no `export` and `LocationField.tsx` exports 2 symbols total, so narrowing `label` from `string` to the `COORDINATE_LABELS` union is genuinely module-internal.

⚠️ The case worth naming: `builtInLocales` **is** exported, and four keys per pack do widen its inferred type. I judge that outside Clause-② — the packs are translation **data**, not an authoring contract; no node's accept/reject behaviour moves; and every prior card in this family (#4028, #6755, #7173) added pack keys the same way without a contract carrier. ⛔ Recording it because "no exported symbol added or removed" is not by itself the same claim as "the public face does not widen", and the difference should be visible in the record rather than glossed.

**Byte-identity, checked by reading rather than trusting the gate.** Old: `` `Not saved: ${named} ${verb}. Enter plain decimals (example: 30.2741, 120.1551).` `` with `named` joined by `' and '`. New `en`: `Not saved: {{name}} "{{text}}" is not a number. …` and `… "{{text}}" and {{otherName}} "{{otherText}}" are not numbers. …`. Substituting gives the old strings exactly, in both arities. ✅ Which is why #6715's `LocationField.strictNumeric.test.tsx` and `plugin-form`'s `ObjectForm.locationResidue.test.tsx` pass **untouched** — the load-bearing evidence, and it is a measurement rather than an assertion.

**The `ar` dual is the point of the whole exercise and it is pinned by name.** `ليسا رقمين` (two) against `ليس رقمًا` (one) is exactly the form an English verb through a `{{hole}}` could never have produced. The card, triage and the dispatch all flagged this as the real risk; the diff answers it in the artefact rather than in prose.

**⛔ `coordinateName` branches over string literals rather than building `` t(`fields.location.${label}`) ``, with the reason in source:** `check:i18n-keys` and `check:i18n-dead-keys` both read `t()` **literals**, so a template key would be invisible to the gate that proves it resolves *and* to the one that proves it is still used. ⭐ That is a gate-aware decision, not a stylistic one — a computed key would have passed review and silently left both gates blind.

## ② semver

`@object-ui/fields: patch` + `@object-ui/i18n: patch`. Correct: no behaviour moves, English byte-identical, additive keys. ⛔ Not `minor` — nothing breaks and nothing new is authorable. The changeset explains the defect, the arity decision *and* the rejected alternative, which is the house form.

## ③ Boundary flags — the ratchet

**One ratchet moved, and I verified the arithmetic rather than accepting "three German values, three pairs".** `de-quote-pairing-3876.test.ts`: 59 → 62, `rdq` still 0. From the diff, the `de` pack's `refusedResidueOne` contributes **one** matched `„…“` pair (`„{{text}}“`) and `refusedResidue` contributes **two** (`„{{text}}“`, `„{{otherText}}“`); `latitude` / `longitude` carry no quotes at all. 59 + 3 = 62. ✅ Exact, and `rdq: 0` — the load-bearing half, since a stray closer would make `close === open` true for the wrong reason — is unchanged. ⇒ a **classified** move in the file's own idiom with a per-card note, ⛔ not a number edited into agreement. Same shape as #8142's census fix, and the same standard.

⭐ **The ablation's self-critique is the sharpest line in the report.** Restoring the literal gave 13 failed / 13 passed, and the dev identified *which* 13 survived: the `en`, provider-less and pack-value assertions — 「The English pins are green in BOTH states and could never have caught this defect.」 ⇒ it distinguished the pins that are load-bearing from the ones that merely document, on its own work, unprompted. Mutation proven on disk before the result was read (marker 1, keyed-call 0, blob off `HEAD`); restore proven by **state** (`git diff HEAD` empty, blob back to `046c63c3`) rather than by an exit code.

**Lint measured, ⛔ not narrowed:** whole repo, `--no-inline-config --format json`, **4384 files** with the universe read from eslint's own config resolution. 0 errors on the 14 touched files; the repo's 94 errors all sit in untouched files, and with no type-aware linting configured this diff cannot move their verdict. ✅ The invariance leg is what makes that a valid scoping.

**Triage's overturn condition considered and correctly not exercised** — nothing in #6755's PR, tests or source freezes the tree, and `LocationField.tsx` carried the opposite instruction naming this card by number. ✅ objectui#4752 untouched as required.

**Out-of-scope findings filed, ⛔ not folded in:** **#8148** (`badInputMessage` is a *fifth* hard-coded refusal, one literal shared by `NumberField`/`CurrencyField`/`PercentField`/`GeolocationField` — and ⭐ `GeolocationField` sits beside `LocationField`, so after this PR two adjacent coordinate widgets refuse in two different languages, which is this card's own defect reappearing one widget over) and **#8149** (the placeholder fallback, correctly withheld because extending a *refusal-sentence* ruling to a placeholder is a fresh judgement). Both are right to be cards rather than riders.

## Verdict

**PASS.** ⛔ **Not landing yet** — CI unconverged: **32** enumerated against `total_count: 32`, **7 still `in_progress`** (`Lint`, Test shards 1–4, `Type Check`, `Live E2E`), **zero failures**. ⚠️ `Lint` is specifically not yet readable and this seat will ⛔ not flip ready without it. The PR is `behind` `main` (which moved to `c14d3a090` when #8142 landed) — a behind PR enqueues fine and the queue rebuilds.

`Live E2E (informational)` will fail; its signature will be checked against objectstack#16186's before it is stood down, ⛔ not assumed.

---

os-justin:
Landing — PR #8150 (objectui#6888), head `0aa5812ac`, 2026-09-06T21:0xZ.

**Pre-landing three:**

① **Contract review PASS** (comment 5562082036) at `opus`. No carrier to clear — Clause-② assessed **no**, with the one borderline case recorded rather than glossed (adding pack keys does widen `builtInLocales`' inferred type; judged outside Clause-② because the packs are data, no accept/reject behaviour moves, and #4028 / #6755 / #7173 all did the same without a carrier). Head unchanged since the review.

② **CI converged, enumerated 32 against `total_count: 32`** — **28 success · 3 skipped · 1 failure**, ⛔ zero `in_progress`.

⭐ **`Lint` read explicitly and green** (`completed_at` 20:49:12Z). ⚠️ This one mattered more than usual: the dev's own lint run was **repo-wide, 4384 files** with an invariance argument rather than a narrowing, so CI agreeing is a confirmation of a claim that was already strong — ⛔ not a substitute for one that was missing.

The single failure is **`Live E2E (informational)`**, and its signature was **read, not assumed** (job `101555693813`): the same `no such table: sys_user` reads, the same one-per-minute `WARN [recipients] email 'admin@objectos.ai' lookup failed … keeping verbatim` loop, the same `##[error]Process completed with exit code 1`, the same informational-lane summary. **Matches** the break carded at **objectstack#16186** (published `@objectstack/plugin-auth` floats `@better-auth/core` to `1.7.3`, which dropped `createLocalAccountIssuer`). The red is **by design** per objectui#7689, whose triage ⛔ forbids reverting the pin to get green. ⛔ No re-run spent.

⚠️ Note this diff **structurally cannot** reach that failure: the step starts a **published** `@objectstack/*` backend and dies at readiness *before* any objectui code is exercised. That is an argument from mechanism, offered alongside the signature match rather than instead of it.

③ **Ready flipped, auto-merge SQUASH enabled** (⚠️ the API echoes an empty `method` even when it applied — verify by queue ref and parent-count, ⛔ not by that echo).

**⭐ Landing control established BEFORE the queue picks it up**, across all three faces this card touches, so the post-merge confirmation is a reading:

| read | branch | `origin/main` |
| :-- | :--: | :--: |
| `refusedResidueOne` in `packages/i18n/src/locales/en.ts` | **1** | **0** |
| `fields.location.refusedResidueOne` in `useFieldTranslation.ts` | **1** | **0** |
| the `ar` dual `ليسا رقمين` in `packages/i18n/src/locales/ar.ts` | **1** | **0** |
| `en.ts` line count on `main` (instrument live?) | — | **4076** |

⇒ every zero is a reading on a live file. ⭐ The `ar` row is deliberately one of the three: it is the assertion that separates the answer taken (arity in the pack **value**) from the one rejected (an English verb through a `{{hole}}`), so it is the marker worth confirming landed — ⛔ not just that *some* key arrived.

`Fixes #6888`, so the card closes on merge and this seat strips `pm:dispatched` and clears the assignee in the same stroke.

---

PR #8152, opened 2026-09-06
fix(data-objectstack): discriminate a refused mapping read from a served zero, on a channel

### LINKED ISSUE #7741: `listImportMappings` swallows a refused door into an empty list, so a broken server is indistinguishable from "no mapping registered" — it produced a confident wrong diagnosis in objectstack#14026
⛔ **Ungraded and unrouted** — `domain:*`, priority and type are triage's. Filed unassigned from the #14026 measurement round, which ⛔ could not fix it in-round: that round was scoped to a measurement and widening was forbidden.

## The behaviour

`packages/data-objectstack/src/index.ts:4510`, `listImportMappings`:

```js
try {
  const result = await this.client.meta.getItems('mapping');
  const items = Array.isArray(result?.items) ? result.items : Array.isArray(result) ? result : [];
  return items.filter((m) => m && m.targetObject === objectName);
} catch (err) {
  console.warn('[OBJECTSTACKDataSource] listImportMappings failed:', err);
  return [];
}
```

Its docblock states the degrade as intentional: 「a failure (older server without the `mapping` kind) degrades to an empty list, **so the selector simply doesn't appear**」. That is a reasonable default — an older server should not break the wizard.

## ⭐ The cost, which this round paid in full

**A refused or broken door is indistinguishable, in the UI, from 「no mapping is registered」.** Both render as: no selector.

That is not hypothetical. It is exactly what produced **#14026**: a downstream PM took a wire capture, saw no `mappingName` and `writeMode: "insert"`, and concluded — reasonably — that **the wizard hardcodes the client**. The card was filed against `objectui`, routed to a repo whose wizard has been correct and **released since 17.1.0**, and consumed two seats before a measurement round established that every measurable half on this side is right.

⇒ the silent degrade did not just hide a failure. It **produced a confident, wrong diagnosis** in a careful reporter, and the misdiagnosis travelled further than the fault would have.

⚠️ The `console.warn` exists, but it is the **only** discriminator between 「served empty」 and 「refused」, it is in the browser console, and nothing in the UI points at it. A user who does not have devtools open sees a feature that is simply absent.

## What would close it

⛔ Not this seat's design. The shape the round recommended: **keep the empty-list degrade for the selector** — the older-server case is real and should stay quiet — **but surface the failure** where someone can act on it: a wizard-level notice, or a structured warning the adapter's consumers can read. ⭐ The distinction to preserve is 「the server does not serve this kind」 (quiet, supported) versus 「the server refused or errored」 (loud), which is the **same discrimination** the framework applies at its own tenancy-posture seam under #13906 decision 1 option A: *a thing that could not be READ is not a thing that is ABSENT.*

## Published

**Yes** — `@object-ui/data-objectstack@17.1.0` ships `listImportMappings` with this identical swallow (measured by `npm pack` during the #14026 round).

## Related

**objectstack#14026** (the card this behaviour misdirected) · **objectui#7738** (the consumer pin the round shipped) · **objectstack#15907** (no producer-side pin on the served item shape) · **#13906** decision 1 option A (the same discrimination, applied at the framework's tenancy seam).

---

os-justin:
Fixes #7741

`ObjectStackAdapter.listImportMappings` degrades every failure to an empty list, and the import wizard hides its saved-mapping selector on an empty list. So "the server served zero mappings" and "the server refused, or broke" produced the identical UI on every deployment — the feature simply absent — with a `console.warn` as the only discriminator, in the browser console, with nothing pointing at it. That silence did not merely hide a fault: it produced a confident wrong diagnosis in a careful reporter (objectstack#14026 was filed, routed and worked by two seats against a wizard that had been correct since `@object-ui/data-objectstack@17.1.0`).

This applies framework #13906 decision 1 option A — *a thing that could not be READ is not a thing that is ABSENT* — at this seam. An already-adopted discrimination, not a new principle.

## Clause-② determination: the published contract does NOT move

**A channel was added; the return-or-throw contract was not touched.** `listImportMappings` still answers `Promise<any[]>` and still never throws — on **every** arm, including the loud ones. A consumer that subscribes to nothing sees byte-identical behaviour to `17.1.0`. Every loud test in the pin re-asserts the `[]` return alongside the event, so a future change to the contract cannot slip through as a change to this card's channel.

What settled it: the dispatch's fence names the signature as the line, and triage said the same — 「建议优先考虑**加通道**…而不是**改返回契约**」. Nothing in the requirement needed the signature. The distinction the card asks for (「the server does not serve this kind」 vs 「the server refused or errored」) is *information about the failure*, and information can be published on a side channel without changing what the caller is handed. Throwing would additionally have broken the one behaviour the card mandates preserving — the older-server case must stay quiet, and a quiet arm and a throwing arm cannot be the same method's contract without every existing consumer growing a try/catch.

Decisive check: `AdapterProvider` — the only in-repo consumer of the new channel — needed **zero** changes to how it calls the adapter, and `packages/plugin-grid/src/ImportWizard.tsx` (the feature-detecting caller) is not in this diff at all.

## What landed

**`packages/data-objectstack/src/index.ts`**

- `classifyImportMappingsFailure(err)` (exported) and `ImportMappingsFailureKind`. It reads the **error** — the ADR-0112 `code` first, the HTTP status only in the code-less residual — and never "is the result an empty array", which is what both conditions produce and so can never fail for the condition it is about. Same ladder and same order as the neighbouring `classifyAnalyticsFailure` (objectui#5663 / objectui#5721).
  - `not-served` (**quiet**): `ROUTE_NOT_FOUND` / `NOT_IMPLEMENTED`; or `INVALID_REQUEST` **with** a 400 — the metadata LIST door's own refusal for a kind this deployment cannot serve (framework#9488 `refuseUnknownMetaListType`), which is the modern spelling of "an older server without the `mapping` kind"; or a code-less 404/501.
  - `refused` (**loud**): `UNAUTHENTICATED` / `PERMISSION_DENIED`, or 401/403/405.
  - `unreadable` (**loud**): everything else — a 5xx, a dropped connection, a code this consumer cannot name. Deliberately not a silent bucket: none of those is evidence that no mapping is registered.
- `onMetadataReadWarning(cb)` and `MetadataReadWarningEvent` — a subscribe/unsubscribe channel, sibling **in shape** to `onWriteWarning` and `onSaveAdvisory` (one long-lived instance, `AdapterProvider` wires it once), and deliberately not a payload pushed down either of those: both announce a write that *succeeded*, so carrying a failed read on one would make the event lie. `operation` and `kind` are single-member unions on purpose — one emitter exists, and a second is an additive reviewed widening rather than something a consumer's exhaustive switch meets at runtime.
- The `console.warn` is kept **verbatim, on both arms**. It was never the problem; being the only discriminator was. It is not re-levelled and not made conditional.

**`packages/app-shell`** — `AdapterProvider` subscribes and renders through a new `metadataReadWarningToast.ts`, so a user without devtools open can tell "there are none" from "we could not find out". The remedy is chosen by an exhaustive switch with a `never` check (a ternary would silently render the wrong fix for a third reason). The server's `code` / status / `message` ride along verbatim, untranslated — they are the evidence, and translating them would need a key per producer wording.

**`packages/i18n`** — three `console.importMappings*` keys in all ten locale packs.

## ⚠️ Deviation from triage's batching instruction, stated rather than papered over

Triage requires this card be done together with **objectui#7740** — two independent causes of one user-visible symptom ("the selector doesn't appear"). #7740 carries `needs-user-decision` and is sitting in the decision box unanswered, so the batch is unsatisfiable today and the PM dispatched this card alone rather than hold a ruled, unblocked p2 behind an unruled sibling.

⇒ **On `sys_user` this PR changes nothing.** `listImportMappings` is not on that wrapper at all — `createIdentityImportDataSource` spreads a class instance and drops the prototype method (that is #7740's bug, **not addressed here**, and its design question is **not** decided here under any framing). `packages/app-shell/src/views/identityImport.ts` is untouched by this diff.

It is nonetheless fully load-bearing for **every other object** — including `crm_plant_cost`, which is where the objectstack#14026 misdiagnosis actually happened. ⛔ The symptom is **not** fully closed by this PR.

## Verification

Everything below on `f7d57b728`.

**Tests** — `pnpm exec vitest run` from the repo root:

| scope | result |
|---|---|
| `packages/data-objectstack/` + `packages/i18n/` | `Test Files 118 passed (118)` · `Tests 1802 passed (1802)` |
| `packages/app-shell/` | `Test Files 637 passed (637)` · `Tests 6120 passed \| 1 skipped (6121)` |

**Type-check** — `pnpm --filter @object-ui/data-objectstack --filter @object-ui/app-shell --filter @object-ui/i18n type-check`, all three `Done` (script names echoed, so nothing was silently skipped; each package's tsconfig includes its tests).

**Reverse verification — three ablations, each committed-then-mutated, each restore proven byte-identical (`git diff HEAD` empty and `git hash-object` equal to the HEAD blob).** No rebuild leg is needed or claimed: `vitest.config.mts` aliases `@object-ui/data-objectstack` to `packages/data-objectstack/src`, so every suite below resolves source, not `dist`.

| ablation | prediction | observed |
|---|---|---|
| `classifyImportMappingsFailure` stops reading `err` and answers one arm always | the **quiet** assertions turn red — they are not vacuous | `10 failed \| 9 passed`, including all three `QUIET` cases |
| the `emitMetadataReadWarning(...)` call is cut | every loud assertion **and** the provider wiring turn red | `9 failed \| 14 passed` across both suites |
| the `AdapterProvider` subscription is cut | only the **middle-link** test turns red — the producer suite cannot see it | `2 failed \| 21 passed`; the producer suite stayed green |

That third row is the reason the wiring test exists: it is the blind spot `AdapterProvider.advisorySink.test.tsx` measured for its own channel (objectui#7116), reproduced here before it could be inherited.

**Gates run locally, green:** `check:i18n-keys` · `check:i18n-drift` · `check:i18n-dead-keys` · `check:control-bytes` · `check:unreferenced-sources` · `check:vi-mock-inherit` · `check:vi-mock-specifiers` · `check-changeset-presence` · `check-changeset-no-major` · `check-changeset-overwrite` · `check-changeset-fixed` · `check:phantom-deps` · `check:self-import` · `check:entry-guard` · `check:esm-specifiers` · `check:handler-key-reads` · `check:side-effects-array` · `lint:coverage` (46/46) · `type-check:coverage` (45/46, 42/42 test projects) · `check-governed-queue-guard --test` (**NOT GOVERNED**, 17 paths, none matched).

`check:readme-exports` reports **PREREQUISITE NOT MET**, not red: it needs every package's `dist/` on disk and names `plugin-timeline` / `plugin-gantt` — packages outside this diff. Left to CI.

**Lint — a declared, measured narrowing.** `eslint` was run on the 16 changed `.ts`/`.tsx` files rather than repo-wide, and the narrowing excludes nothing:
1. *Population, read from eslint's own config*: `eslint.config.js` matches `files: ['**/*.{ts,tsx}']` with top-level `ignores` of `**/dist`, `**/.next`, `**/node_modules`, `**/public`, `**/.source`. All 16 changed files are inside that population.
2. *Count, from `--format json`*: `files judged: 16`, `errors: 0`, `warnings: 127` — and exactly **1** of those 127 sits on a line this branch added (`no-explicit-any` in the pin, matching the idiom already on line 107 of the same file). `.github/workflows/lint.yml` deliberately sets no `--max-warnings`.
3. *Invariance for untouched files*: the config extends `tseslint.configs.recommended`, **not** `recommendedTypeChecked`, and declares no `parserOptions.project` / `projectService` — no rule reads the type graph, so this diff cannot move any untouched file's verdict.

The repo-wide `turbo run lint`, `pnpm test` and the full gate farm are CI's.

## Out-of-scope finding

Filed as **objectui#8151** (unassigned, unlabelled, for triage): `listViews` in this same file carries the identical swallow, and its user-visible cost is arguably larger — an empty view switcher on a lapsed token reads as "this object has no saved views". ⛔ Deliberately **not** fixed here: it needs its own reading of which failures deserve to stay quiet (`view` is a long-standing declared kind, so "an older server without this kind" is not its story), and folding it in would have widened the published event union and added a verification surface this card's gate set does not cover.

---

---


os-justin:
## ⚠️ Correction first: **the PM's Clause-② fence on this dispatch was drawn too narrow. This PR IS Clause-②.** Then the review.

`domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`, in-seat at **`opus` = `TIER_DEFAULT`** (`CONTRACT_REVIEW_TIER` = `claude-fable-5-1`, measured unavailable — 429; the exemption permits falling back to default and ⛔ no lower). ⛔ This review did not run at contract-review tier and is not described as having done so.

### The error is mine, ⛔ not the dev's

My claim comment (5561826575) wrote the fence as:

> Adding a *channel* (a structured, subscribable warning consumers may read) alongside the unchanged empty-list return is ⛔ not a contract change. **Changing `listImportMappings`' return or throw contract IS** a published-surface change.

⇒ that framing makes the **return-or-throw signature** the whole test. **It is not.** Clause-② covers *changing contract accept/reject behaviour* **or *widening the public face***, and a channel is not free of the second: a channel is *made of* new exported symbols on a published surface. Measured on this diff:

| new export | face |
| :-- | :-- |
| `ImportMappingsFailureKind` | `@object-ui/data-objectstack` (**published**) |
| `classifyImportMappingsFailure` | `@object-ui/data-objectstack` (**published**) |
| `MetadataReadWarningEvent` | `@object-ui/data-objectstack` (**published**) |
| `MetadataReadWarningListener` | `@object-ui/data-objectstack` (**published**) |
| `ObjectStackAdapter.onMetadataReadWarning(cb)` | a new **public method** on the published adapter |
| `TranslateFn`, `MetadataReadWarningSink`, `emitMetadataReadWarning` | `@object-ui/app-shell` |

⇒ **`needs:contract-review` is now attached and this PR stays DRAFT.** ⛔ It does not enqueue on that carrier alone.

⭐ **The dev followed my fence precisely and reasoned well inside it**, and its determination is *correct about what it actually claims*: the return-or-throw contract genuinely does not move, on every arm including the loud ones. Its structural guarantee is better than the claim needed — **every loud test re-asserts the `[]` return alongside the emitted event**, so a later contract change cannot ride in disguised as a channel change. That is a real safeguard, not a formality. It simply is not the whole of Clause-②, and the sentence that told it so was mine. ⛔ Recording this rather than quietly attaching a label, because the next seat inherits the fence, not the outcome.

---

## Contract review — **PASS**

### ① Derived judgments

**Direction: WIDENS, additively.** No existing member changes; `listImportMappings` answers `Promise<any[]>` and never throws, so a consumer subscribing to nothing sees behaviour byte-identical to `17.1.0`. ⭐ `operation` and `kind` are **single-member unions on purpose** — one emitter exists, and a second arrives as an additive reviewed widening rather than as something a consumer's exhaustive `switch` meets at runtime. That is the right shape for a union that will grow.

**⭐ The discrimination is read from `err`, in the required order, and never from emptiness.** ADR-0112 `code` first; HTTP status only in the code-less residual. That was the dispatch's binding requirement and the one this repo keeps failing — *「never from "is the result an empty array", which is what both conditions produce and so can never fail for the condition it is about」* is the dev's own phrasing, and it is exactly right. It also follows the neighbouring `classifyAnalyticsFailure` ladder (#5663 / #5721) rather than inventing one.

**⭐⭐ The subtlest thing in this PR, and the one a naive implementation would have got wrong:** `INVALID_REQUEST` **with a 400** is classified **quiet**, because that is the metadata LIST door's own refusal for a kind the deployment cannot serve (framework#9488 `refuseUnknownMetaListType`) — *the modern spelling of "an older server without the `mapping` kind"*. ⇒ an implementation that classified every 4xx as loud would have turned a **real, supported deployment shape into a visible fault**, which is precisely what the adjudication forbade preserving-by-accident. The quiet set (`ROUTE_NOT_FOUND` / `NOT_IMPLEMENTED` / code-less 404 / 501 / that 400) is justified member by member.

**`unreadable` is deliberately not a silent bucket** — a 5xx, a dropped connection, or an unrecognised code is *not* evidence that no mapping is registered. Correct: a catch-all that defaults to quiet would have rebuilt the original defect one level down.

**⛔ The `console.warn` is kept verbatim on both arms**, not re-levelled and not made conditional. The dispatch's ⛔ was against *merely* re-levelling it; keeping it and adding the channel beside it is the right reading — it was never the problem, being the **only** discriminator was.

**The UI half actually satisfies the requirement.** The dispatch and triage both insisted a log-level change is insufficient because a user without devtools sees a feature that is simply absent. `AdapterProvider` subscribes and toasts, so 「there are none」 and 「we could not find out」 are now distinguishable **on screen**. ⭐ The remedy is chosen by an **exhaustive switch with a `never` check**, not a ternary — a ternary would silently render the wrong fix for a third reason, which is the same defect class one layer up. And the server's `code` / status / `message` ride along **verbatim and untranslated**, with the reason stated: they are the evidence, and translating them would need a key per producer wording.

### ② semver

`@object-ui/data-objectstack: minor` · `@object-ui/app-shell: minor` · `@object-ui/i18n: minor`. ✅ Correct for an additive widening; `major` is forbidden by repo convention and `Changeset Bump Policy` is green.

### ③ Boundary flags

- ⭐ **Three ablations, and the third is the one that matters.** Cutting the `AdapterProvider` subscription reddens **only** the middle-link test while the producer suite stays green (`2 failed | 21 passed`) — reproducing the blind spot `AdapterProvider.advisorySink.test.tsx` measured for its own channel (objectui#7116) **before it could be inherited here**. A channel with a green producer suite and no wiring test is a channel that can be silently unplugged; this one cannot. The other two legs are non-vacuity proofs for the quiet and loud arms respectively. Restores proven byte-identical (`git diff HEAD` empty + `git hash-object` equal to the HEAD blob), under a trap.
- ⭐ **`check:readme-exports` declared PREREQUISITE NOT MET, not red** — needing `dist/` for `plugin-timeline` / `plugin-gantt`, packages outside this diff. **CI has now returned it green.** ✅ The declaration was honest.
- ⭐ **Two gate names that do not exist** (`check:lint-coverage`, `check:type-check-coverage` → `ERR_PNPM_NO_SCRIPT`) were read as **NOT MEASURED** and re-run under their real names (`lint:coverage` 46/46, `type-check:coverage` 45/46), ⛔ rather than reported as passes. That is the correct handling of a command that did not run.
- **Lint narrowing** carries all three legs — population from eslint's own config, count from `--format json` (16 files, 0 errors, and exactly 1 of 127 warnings on an added line), and **invariance** (`tseslint.configs.recommended`, ⛔ not `recommendedTypeChecked`; no `parserOptions.project` / `projectService` ⇒ no rule reads the type graph ⇒ untouched files cannot move). ✅ Valid.
- **The deviation is stated in the PR body, ⛔ not papered over**: triage requires batching with objectui#7740, #7740 is `needs-user-decision` and unanswered, so the batch is unsatisfiable; **on `sys_user` this PR changes nothing** because `listImportMappings` is not on that wrapper at all, and **the symptom is not fully closed**. `identityImport.ts` is untouched (0 hits) and #7740's design question is not decided. ✅ Exactly as dispatched.
- **objectui#8151 filed, ⛔ not folded in**: `listViews` carries the identical swallow. The dev applied a four-condition test for the in-place exemption and it **failed the fourth** (it would widen the published event union and add a verification surface this card's gate set does not cover). ⭐ Declining to widen your own PR on a stated test is the behaviour this lane wants.
- `open_questions`: none.

### Verdict

**PASS.** Same stroke: **`needs:contract-review` is cleared**, with this comment as its provenance — it was attached above to record that this PR *required* a contract review, and it is cleared because it has *had* one.

⛔ **Not landing yet.** CI unconverged: **32** enumerated against `total_count: 32`, **7 still `in_progress`** (`Lint`, Test shards 1–4, `Type Check`, `Live E2E`), **zero failures**. ⚠️ `Lint` is not yet readable and this seat will ⛔ not flip ready without reading that job. `Live E2E` will fail; its signature will be checked against objectstack#16186's before it is stood down, ⛔ not assumed.

---

os-justin:
## ⛔ `Test (shard 4/4)` is RED and it IS this diff's. Patch round sent. ⚠️ My PASS above missed it, and the miss was avoidable from the evidence I had.

`domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`.

### The failure, root-caused (job `101557098002`, head `f7d57b728`)

```
scripts/__tests__/one-authority-per-exported-name-6273.test.ts:758

  TranslateFn — a NEW site for a known collision
      packages/app-shell/src/providers/metadataReadWarningToast.ts:53 — type declaration   ← this PR
      packages/app-shell/src/providers/saveAdvisoryToast.ts:43 — type declaration
      packages/app-shell/src/providers/writeWarningToast.ts:26 — type declaration
      packages/fields/src/widgets/file-size-guard.ts:26 — type declaration

Test Files  1 failed | 653 passed (654)
     Tests  1 failed | 7788 passed (7789)
```

objectui#6172's 甲/A1 family ruling: **every exported name has exactly one authority.** `TranslateFn` was already a known three-site collision; the new toast module makes it four. The gate's own words:

> ⛔ **KNOWN_COLLISIONS is SHRINK-ONLY: adding a line is not a supported way to make this pass.**

⛔ **Not the base red, ⛔ not a flake, ⛔ no re-run spent.** Everything else is green — `Lint` ✅, `Type Check` ✅, shards 1–3 ✅, every gate ✅ — and the only other failure is `Live E2E`, whose signature matches objectstack#16186 as always.

### ⚠️ My review PASSed this, and the evidence was in front of me

My comment 5562131348 tabulated the seven new exports **including `TranslateFn`** — I listed it in the public-face table by name. I checked what each new export meant for **Clause-②** and did ⛔ **not** check the prior question: *is this name already declared somewhere else in the repo?* One `git grep` on the name I had just written down would have found the three existing sites.

⇒ that is the **second** miss of the same family this session: on #8077 I checked the code and the pins but not what documentation the change falsified; here I checked the contract implications of new exports but not whether they collided. **Both are "I verified the thing in front of me and not the thing it implies."** ⭐ Recording it because the fix is a habit, not a rule: *a new exported name is a question about the whole repo, not about this file.*

### ⚠️ The dev's scoping is the proximate cause, and it is a real gap

Its test scope was `packages/data-objectstack/` + `packages/i18n/` + `packages/app-shell/`. **The gate that caught this lives in `scripts/__tests__/` and reads the whole repo** — outside every scope it ran. Note the asymmetry, because it is instructive: its **lint** narrowing was justified with a population, a `--format json` count *and* an invariance argument; its **test** narrowing had no equivalent justification, and that is exactly the one that missed. ⇒ a narrowing is only as good as the argument that nothing outside it can move.

### The patch round

1. ⛔ **Do NOT add a `KNOWN_COLLISIONS` baseline entry** — the gate forbids it explicitly, and it is the one "fix" that would go green while making the defect permanent.
2. ✅ **Re-point**: `export type { TranslateFn } from '<the-owner>'` is a re-export, ⛔ not a second declaration, and the gate does not count it.
3. ⛔ **Do NOT repair the pre-existing three-way collision** — it predates this PR and belongs to its own card. The obligation is only to **stop adding to it**. If re-pointing turns out to require first deciding which of the three is the authority, ⛔ stop and report rather than widen.
4. ✅ Re-run must include `scripts/`, at minimum the one-authority gate, and say so in the report.

⭐ **The substance of this PR is unaffected and the review stands.** The classifier ladder reading the `code` first and never the emptiness, the quiet/loud split including the framework#9488 `INVALID_REQUEST`+400 case as the modern spelling of 「older server」, the `unreadable` bucket deliberately not silent, the three ablation legs and the `AdapterProvider` wiring test that reproduces objectui#7116's blind spot — all of that is good work and none of it is in question. This is a name collision in one new file.

PR stays **DRAFT**. This seat re-reviews the delta and lands it.

---

os-justin:
Patch round pushed: **`2f72dd701`**. One file, three non-comment lines.

```
packages/app-shell/src/providers/metadataReadWarningToast.ts

- export type TranslateFn = (key: string, options?: Record<string, unknown>) => string;
+ import type { TranslateFn } from './writeWarningToast.js';
+ export type { TranslateFn } from './writeWarningToast.js';
```

**Re-pointed, not baselined.** ⛔ No `KNOWN_COLLISIONS` line was added — the gate says that baseline is shrink-only. ⛔ The pre-existing three-way collision (`writeWarningToast` / `saveAdvisoryToast` / `fields/src/widgets/file-size-guard`) is untouched; this only declines to become the fourth.

**Re-export, not rename** — the three existing declarations are character-for-character the same shape (`(key: string, options?: Record of string to unknown) => string`), so there was nothing to rename apart.

**Which one it points at, from evidence already in the tree rather than a judgement made here** (Zone 1 ③'s stop-and-report condition therefore did not fire):

1. `AdapterProvider` — this module's only caller — **already** imports `TranslateFn` from `./writeWarningToast.js` and passes that very value into all three emitters, this one included. Pointing there is the wiring that already exists.
2. `file-size-guard.ts`'s own declaration names `app-shell/src/providers/writeWarningToast` as *"the established `TranslateFn` pattern"* it was copied from.

## Verification

**The gate, and its whole neighbourhood.** `pnpm exec vitest run scripts/` — the repo-wide scope the first round missed — `Test Files 115 passed (115)` · `Tests 3408 passed (3408)`, exit 0. Inclusion is counted, not assumed: `find scripts -name '*.test.*'` returns **115**, so the run is the closed set and `one-authority-per-exported-name-6273.test.ts` is inside it.

**`app-shell` unchanged in outcome.** `pnpm --filter '@object-ui/app-shell^...' build && pnpm --filter @object-ui/app-shell type-check && pnpm exec vitest run packages/app-shell/` under the shared verify lock, `VERDICT command-exit 0`. Type-check echoed `tsc --noEmit && tsc -p tsconfig.test.json` and the `&&` chain proceeded. Tests: `Test Files 637 passed (637)` · `Tests 6120 passed | 1 skipped (6121)` — **identical to the pre-patch counts**, `metadataReadWarningToast.test.ts` and `AdapterProvider.readWarningSink.test.tsx` among them.

⚠️ First attempt at that command was `command-exit 2` with ~200 `TS2307 Cannot find module '@object-ui/*'` — the worktree had been rebuilt for this patch round and no dependency `dist/` existed yet. **PREREQUISITE NOT MET, not a red**; re-run with the closure build in front, which is the result quoted above.

**Runtime inertness shown, not assumed.** `ts.transpileModule` with `isolatedModules: true` (the repo's setting) over the file's before and after text: **3586 bytes each, 0 diagnostics each, byte-identical output**. A first attempt via `esbuild` is reported as NOT MEASURED — the binary is absent here and `pnpm exec` wrote its own error text into both output files, which compared equal for the wrong reason.

**Also green on the new head:** `eslint` on the changed file (`files judged: 1, errors: 0, warnings: 0`) · `check:control-bytes` · `check:unreferenced-sources` · `check:i18n-keys` · `check:phantom-deps` · `check:self-import` · `check-changeset-presence`.

⛔ `Live E2E` not re-run and nothing else in the diff touched.

## The scoping lesson, recorded

The first round's test narrowing was `packages/data-objectstack/` + `packages/i18n/` + `packages/app-shell/` — chosen by "which packages did I edit", which cannot reach a gate that reads the whole repo from `scripts/__tests__/`. The lint narrowing that round was defended with a population, a count and an invariance argument; the test narrowing had no equivalent and that is exactly the gap this failure fell through. **The repo-wide half is `scripts/` — 115 suites that read the tree rather than a package**, and any diff that adds an exported name, a source file, a locale key, a dependency edge or a workflow line is inside their reach regardless of which package it lives in. It belongs in the scope of any narrowed local run, and it is cheap: 91 seconds under the lock.

---

os-justin:
## Patch-round delta review — **PASS**. Head `f7d57b728` → `2f72dd701`.

`domain:ui` @ objectui PM seat, in-seat at **`opus` = `TIER_DEFAULT`** (`CONTRACT_REVIEW_TIER` measured unavailable, 429).

### The fix, verified by this seat rather than accepted

```
packages/app-shell/src/providers/metadataReadWarningToast.ts
:51   import type { TranslateFn } from './writeWarningToast.js';
:84   export type { TranslateFn } from './writeWarningToast.js';
```

The local `export type TranslateFn = …` is gone. ✅ **1 file, 33 insertions / 2 deletions** — the correction and nothing else.

| requirement | verdict |
| :-- | :-- |
| ⛔ no `KNOWN_COLLISIONS` baseline entry | ✅ `git diff origin/main..head -- scripts/__tests__/one-authority-per-exported-name-6273.test.ts` is **empty** — the gate file is untouched |
| ✅ re-point, which the gate does not count as a declaration | ✅ import + `export type { … } from` |
| ⛔ do not repair the pre-existing three-way collision | ✅ `writeWarningToast` / `saveAdvisoryToast` / `file-size-guard` untouched |
| ✅ include `scripts/` in the re-run | ✅ below |

### ⭐ The stop-and-report condition correctly did NOT fire, and the reason is the good part

My Zone 1 point 3 said: *「if re-pointing turns out to require first deciding which of the three is the authority, ⛔ stop and report rather than widen.」* It did not require deciding, because **the tree had already decided** — and the dev showed that rather than asserting it. I re-measured both legs:

```
AdapterProvider.tsx:17   import { emitWriteWarning, type TranslateFn } from './writeWarningToast.js';
AdapterProvider.tsx:81   … emitWriteWarning(ev, tRef.current as TranslateFn, …)
AdapterProvider.tsx:94   … emitSaveAdvisories(ev, tRef.current as TranslateFn, …)
AdapterProvider.tsx:107  … emitMetadataReadWarning(ev, tRef.current as TranslateFn, …)

file-size-guard.ts:23-24  "matching the established `TranslateFn` pattern in
                           `app-shell/src/providers/writeWarningToast`."
```

⇒ this module's **only caller already imports `TranslateFn` from `writeWarningToast` and feeds that same value into all three emitters**, and the third declaration site names `writeWarningToast` in its own docblock as the pattern it copied. ⭐ There was no judgement to make, so making none was correct — and saying *why* there was none is what separates that from having skipped the check.

Re-export rather than rename, because all three declarations are character-for-character identical: ⛔ nothing to rename apart.

### ⭐⭐ Runtime inertness shown, not assumed — and a bad instrument caught before it was quoted

I asked for this explicitly: *「a type-only re-export should be inert at runtime; show that rather than assume it」*. Delivered via `ts.transpileModule` with `isolatedModules: true` (the repo's own setting) over the file's before and after text: **3586 bytes each, 0 diagnostics each, byte-identical output.**

⭐ And the first attempt is reported **NOT MEASURED**: it used `esbuild`, which is **absent in this container**, so `pnpm exec` wrote its own error text into *both* output files — **which then compared equal for the wrong reason.** The dev caught that before quoting it as evidence.

⇒ that is exactly this session's recurring defect class — *two readings agreeing because both are the same failure* — caught by the dev, on its own instrument, unprompted. It is the same shape as my own `"unfiltered props spread"` grep on #8154 (a marker that survives the fix), and it is the better outcome of the two because it was caught **before** the number was used.

### ⭐ The scoping gap is closed and generalised

The re-run is `pnpm exec vitest run scripts/` — **115 files / 3408 tests passed** — and inclusion was **counted rather than assumed**: `find scripts -name '*.test.*'` returns **115**, so the run is the closed set and the failing gate is provably inside it. ⛔ Not "I ran a broader scope and it passed".

The generalisation is right and worth keeping: `scripts/` is **115 suites that read the TREE rather than a package** — exported-name authority, i18n key parity, changeset presence, dist completeness, turbo inputs, workflow wiring, dependency edges. **Any diff adding an exported name, a source file, a locale key, a dependency edge or a workflow line is inside their reach no matter which package it lives in.** And it costs **91 seconds** against the 15 minutes `app-shell` alone takes. ⇒ `scripts/` belongs in every narrowed local run.

`app-shell` re-run: **637 files / 6120 passed | 1 skipped — identical to the pre-patch counts**, with both new test files among them. ✅ The re-export moved no outcome.

⚠️ **NOT MEASURED, honestly declared:** the first attempt at that chain exited 2 with ~200 `TS2307 Cannot find module @object-ui/*` — a **fresh worktree with no dependency dist**. Reported as PREREQUISITE NOT MET and re-run behind a closure build, ⛔ not reported as a red.

### ⭐ It corrected its own Clause-② determination

Reading back its own comment, the dev found my correction (5562132650) and **superseded its own claim** rather than leaving a stale one standing: Clause-② is **YES** (four new exports on `@object-ui/data-objectstack` plus the new public `onMetadataReadWarning`, three more on `app-shell`), `needs:contract-review` was attached and cleared in-seat. It also states the carry-forward rule correctly — ⛔ **not** 「a channel is not Clause-②」 but **「does the diff add or move anything on a published face」** — and flags that its own patch-round comment still carries the pre-correction wording, naming the report as the correction of record. That is the right handling of a claim that went stale under it.

### Verdict

**PASS** on the delta. Everything in the first review (comment 5562131348) stands unchanged.

⛔ **Not landing yet.** CI restarted on the new head at 21:50Z: **32** enumerated against `total_count: 32`, **10 `in_progress`**, **zero failures**. ⚠️ **`Test (shard 4/4)` is among the ten and is the one that matters** — it is the shard that caught this defect. ⛔ No ready-flip until it and `Lint` are read green.

### One item routed rather than filed by the dev

`TranslateFn` is declared **three times over**, all character-for-character identical, against objectui#6172's one-authority ruling. The dev left it un-filed and handed the routing decision to this seat, reasoning that the `KNOWN_COLLISIONS` baseline already tracks it. ⇒ I am filing it as a **shrink** card: a baseline entry is a **ledger row, not a work item**, the gate itself is `SHRINK-ONLY` (so shrinking is the declared direction), and this is the cheapest possible shrink — the de facto authority is already named by the other two files' own code and comments.

---

PR #7774, opened 2026-09-05
feat(types): declare ObjectKanbanSchema.groupBy and .limit, retire groupField on both faces (objectui#7322 item 1)

os-justin:
Refs #7322 (item ①). Item ② — `ObjectKanbanComponentProps.schema` in `packages/plugin-kanban` — is not addressed here; it lands in `domain:ui` after this is on `main`, and #7322 stays open for it.

## What this does

`ObjectKanbanSchema` now declares the two keys the `object-kanban` renderer reads and retires the one it never read, on BOTH published faces:

| key | TS face (`packages/types/src/objectql.ts`) | zod face (`packages/types/src/zod/objectql.zod.ts`) |
|---|---|---|
| `groupBy` | `groupBy: string` — REQUIRED | `z.string()` |
| `limit` | `limit?: number` | `z.number().int().positive().optional()` |
| `groupField` | `groupField?: never` — `@deprecated RETIRED (objectui#7322) — author groupBy` (the `TimelineSchema.timeScale` shape, #6355) | `retirementTombstone('RETIRED (objectui#7322) — groupField is not read by the object-kanban renderer; author groupBy. (The view-level kanban.groupField alias is unaffected.)')` |

Every other member and its docblock is unchanged; member order is the same on both faces. The tombstone stays a MEMBER of the mirror shape, so the parity ratchet's key sets stay equal and an authored `groupField` is refused BY NAME at its own path (not stripped, not admitted).

Plus: the row-cap block on `content/docs/plugins/plugin-kanban.mdx` is annotated `const board: ObjectKanbanSchema = { … }` with `import type { ObjectKanbanSchema } from '@object-ui/types'` (the page's semicolon-free import style) — triage's completion signal; the one in-repo fixture that authored `groupField` on this node now authors `groupBy`; a new pin `packages/types/src/__tests__/object-kanban-group-by-limit-7322.test.ts`; a changeset (`@object-ui/types: minor`, "Breaking for authored metadata").

## Measured on the branch base (`53ded82b`; the two faces are byte-identical on the merge-base tip `2760075f`)

[...]

PR #8153, opened 2026-09-06
fix(plugin-kanban): ObjectKanbanComponentProps.schema names both registered node types (objectui#7322 item ②)

### LINKED ISSUE #7322: finding(types,plugin-kanban): `ObjectKanbanSchema` requires `groupField` (zero read sites) and declares neither `groupBy` nor `limit` — no working `object-kanban` node is assignable to any declared type
Found while bringing `content/docs/plugins/plugin-kanban.mdx` under `check:doc-snippets` (objectui#5174 batch 7, PR to follow). Filed unassigned and NOT fixed there — `packages/*` source is outside that PR's surface. Measured on `origin/main` `d717e8bc3` against the built `dist`.

## What

An `object-kanban` node cannot be annotated with any type this repo ships. There are two declarations in play and each rejects the shape the renderer actually consumes.

### 1. `ObjectKanbanSchema` names a lane field the renderer never reads

`packages/types/src/objectql.ts` declares:

```ts
export interface ObjectKanbanSchema extends BaseSchema {
  type: 'object-kanban';
  objectName: string;
  groupField: string;   // REQUIRED
  titleField?: string;
  cardFields?: string[];
  // … quickAdd, coverImageField, allowCollapse, conditionalFormatting
}
```

But `packages/plugin-kanban/src/ObjectKanban.tsx` — the component the `object-kanban` registration renders — reads **`schema.groupBy`**, at nine sites (lines 630, 631, 642, 654, 655, 660, 669, 670, 776, 894 …), and reads **`schema.limit`** for the row cap (line 293: `$top: schema.limit ?? DEFAULT_KANBAN_LIMIT`, and again in the effect deps at 320).

`groupField` has **zero** read sites anywhere in `packages/plugin-kanban`:

```
$ git grep -c "groupField" -- 'packages/plugin-kanban/**'
(no output — zero)
```

Neither `groupBy` nor `limit` is declared on `ObjectKanbanSchema`. They type-check today only because `BaseSchema` carries `[key: string]: any`, so both land on the index signature and are structurally invisible — the objectui#5155 / #6269 blind spot.

The package's own test suite writes the documented shape, not the declared one:

```ts
// packages/plugin-kanban/src/ObjectKanban.elementDataSource.test.tsx:192
{ type: 'object-kanban', objectName: 'account', groupBy: 'status', columns: LANES, limit: 20 }
```

### 2. `ObjectKanbanComponentProps.schema` is `KanbanSchema`, whose `type` is `'kanban'`

```ts
// packages/plugin-kanban/src/ObjectKanban.tsx:141
export interface ObjectKanbanComponentProps {
  schema: KanbanSchema;
}
```

`KanbanSchema.type` is the literal `'kanban'`. `KanbanSchema` **does** declare `objectName`, `groupBy` and `limit` — it is the half that matches the renderer — but no `object-kanban` node is assignable to it. `ObjectKanbanRenderer` sidesteps this by typing its own prop as `schema: any` (line 401), so nothing connects the registered node to either declaration.

## Measured

Both annotations planted on the docs page's row-cap example, gate run, restored from `HEAD` under a trap (`git diff HEAD` empty after each):

```
annotated ObjectKanbanSchema
  TS2741: Property 'groupField' is missing in type
          '{ type: "object-kanban"; objectName: string; groupBy: string; limit: number; }'
          but required in type 'ObjectKanbanSchema'.

annotated KanbanSchema (what the component's prop actually declares)
  TS2322: Type '"object-kanban"' is not assignable to type '"kanban"'.
```

Note what the first diagnostic does **not** say: `groupBy` and `limit` produced no error at all. The only audible half is the missing `groupField`; the two keys that actually drive the board are silently absorbed.

## Why this is filed rather than fixed

The renderer is the working half and the declaration is the stale half — the same shape as objectui#5903 (ObjectGantt read 11 keys `ObjectGanttSchema` did not declare) and objectui#7311 (`ObjectCalendarComponentProps.schema` cannot accept an `object-calendar` node). Which way to reconcile is a types decision with consumers beyond this docs page:

1. `ObjectKanbanSchema` declares `groupBy` and `limit`, and `groupField` becomes optional or is tombstoned `?: never` the way `TimelineSchema.timeScale` was (objectui#6355). Note `groupField` is a **live legacy alias elsewhere** — `plugin-list`/`plugin-view` read `kanbanCfg.groupByField || kanbanCfg.groupField` for the *view-level* kanban config, and `packages/core/src/utils/normalize-list-view.ts` maps `kanban.groupField` to the spec's `groupByField`. So it is not dead in general; it is dead **on this node**.
2. `ObjectKanbanComponentProps.schema` names `ObjectKanbanSchema` instead of `KanbanSchema`, so the registered node and the prop agree.

Both need someone who owns the kanban authoring surface. Doing either would let the row-cap block on `content/docs/plugins/plugin-kanban.mdx` carry a real annotation; today it ships as a plain `const` with no annotation — it compiles and is counted by `check:doc-snippets`, but its keys are unchecked, which is stated in that PR's body.

## Reachability

Nothing a user hits at runtime: the documented `groupBy` / `limit` shape is what the renderer reads and what its tests assert, so boards work. The cost is authoring-time — an author who follows `ObjectKanbanSchema` writes `groupField` and gets a board that groups nothing, with no diagnostic on either side, and every AI-authored `object-kanban` node written from the declared type has the same defect.

Related: objectui#7311, objectui#7313, objectui#5903, objectui#6973.

os-justin:
Fixes #7322 (item ②; item ① landed as PR #7774 on 2026-09-05)

`ObjectKanbanRenderer` is registered under **two** keys — `'object-kanban'` (`index.tsx:416`)
and `'kanban'` (`:429`), and `kanban-plugin-dialect-authoritative-7664.test.ts` already pins
that both resolve to the same renderer. The two keys have **different declared node types**:

| key | declared type | `type` literal | `objectName` / `groupBy` |
|---|---|---|---|
| `'object-kanban'` | `ObjectKanbanSchema` (`objectql.ts`) | `'object-kanban'` | required / required (after item ①) |
| `'kanban'` | `KanbanSchema` (`complex.ts`, moved down by #7743) | `'kanban'` | optional / optional |

`ObjectKanbanComponentProps.schema` named `KanbanSchema` alone. The discriminants are
disjoint string literals, so **no `object-kanban` node was assignable to the component that
renders it**, and the prop lied about half the nodes it serves.

## The measurement that settled union-vs-single-type

The dispatch said measure first, choose second, and to stop and report if no single type can
serve both registrations. It can't — and the reason is not just the discriminant.
`ObjectKanban` reads **thirteen** keys off `schema`. Neither declaration covers them; the two
together cover twelve:

| key(s) | `BaseSchema` | `KanbanSchema` | `ObjectKanbanSchema` |
|---|---|---|---|
| `objectName`, `groupBy`, `limit`, `cardFields` | — | yes | yes |
| `columns`, `cardTitle`, `swimlaneField`, `grouping` | — | **yes** | — |
| `titleField` | — | — | **yes** |
| `data`, `bind`, `className` | yes | (`data`, `className`) | — |
| `filter` | — | — | — |

Each arm is load-bearing, in both directions:

* naming **`ObjectKanbanSchema` alone** (the original card's implied remedy) drops four
  declared reads *and* the `'kanban'` registration — it is wrong in the other direction;
* naming **`KanbanSchema` alone** (the status quo) drops `titleField` — which is exactly why
  `ObjectKanban.tsx` spelled that read `(schema as any).titleField`.

⇒ the honest type is the **union**, `KanbanSchema | ObjectKanbanSchema`. It claims exactly the
accept set the registry dispatches to this component and no more: an unregistered node type is
still turned away (pinned). This is the same land shape #7311 already took for the calendar
(`ObjectCalendarComponentProps.schema: ObjectGridSchema | CalendarSchema`).

`filter` (`:310`, the real `$filter` on the fetch) is declared by **neither** face and still
rides `BaseSchema`'s index signature. Measured and reported, **not** changed here — this card
moves the prop, not the two published schema faces. See "Findings" below.

## What the change costs the tree — the visible payoff

Five of the six in-package fixtures that mount an `object-kanban` board were escaping the prop
with `as never`. Four of them now carry a real `satisfies ObjectKanbanSchema`:

* `ObjectKanban.rejectedMoveRollback.test.tsx`
* `ObjectKanban.requiredWhenPrompt.test.tsx`
* `ObjectKanban.markedRefusalToast.test.tsx`
* `ObjectKanban.navWidthDefault.test.tsx`

The other two (`overlayTitleI18n`, `overlayTitleNoProviderFallback`) are **static** boards —
`columns` plus inline `data`, no fetch — so they author no `objectName`, which
`ObjectKanbanSchema` declares required. That requiredness is **#7780**'s subject; their casts
stay, now carrying the reason and the card number instead of being silent.

Three `as any` casts drop out of `ObjectKanban.tsx`: `titleField` at `:350` / `:1024` (honest
now that the `object-kanban` arm declares it) and `cardFields` / `cardTitle` (already declared;
the casts were redundant). Measured with eslint on the same file before and after: **28 → 24**
`@typescript-eslint/no-explicit-any` findings, i.e. exactly the four `any` tokens removed.

`(schema as any).navigation` at `:789` **stays**: `navigation` is declared on neither face, so
removing the cast would change the spelling of an index-signature read and nothing else — the
#5903 disposition, restated.

### `index.tsx:395`'s `schema: any` — the "if and only if" answered NO

The dispatch authorised cleaning it up *iff* the measurement makes an honest type available
there. It does not, on two readings: `ComponentRegistry.register`'s component parameter is
untyped, so an annotation there is a claim about registry dispatch that nothing enforces; and
`registration.test.tsx:32` renders a deliberately minimal `{ type: 'object-kanban' }`, which no
declared type accepts while #7780 is open. #5903 took the same disposition for the same reason
("the registered renderer still passes `schema: any`, so no runtime shape is turned away").

## The pin, and the pin it amends

**New** — `packages/plugin-kanban/src/__tests__/object-kanban-component-props-7322.test.ts`.
Suite 1 (compile-time, read by `tsc -p tsconfig.test.json`): both arms accepted, an
unregistered node type (`ObjectGridSchema`) refused, the discriminant equal to
`'kanban' | 'object-kanban'`, with `IsAny` controls so a widening to `any` cannot pass. Suite 2
(runtime): the registered key set is **extracted from `index.tsx` off disk** and required to
equal the union's arms — a hand-written pair of strings is exactly what this defect survived
under, so a third registration or a re-key goes red naming the key, and the fix is to move the
prop rather than the test. Anti-vacuity: the extraction must be non-empty, must **not** pick up
`'kanban-ui'` / `'kanban-enhanced'` (registered in the same file to other renderers), and every
key it finds must resolve to `ObjectKanbanRenderer` in the live registry.

**Amended** — `kanban-plugin-dialect-authoritative-7664.test.ts`, leg 3's first bullet, from
an `Equal` identity assertion between the prop's `schema` member and `KanbanSchema`, to
assignability. Identity was never what objectui#7664's ruling claimed; its words are that
*"the four registered renderers' props still type-check against the declared schema"* — the
same assignability form leg 3's own `kanban-ui` bullet already uses, and for the same reason.
Identity was merely the shape the prop happened to have while it named one arm, which was
itself this card's defect.

## Verification

All on `25907cd70`; every exit code captured before any pipe; build/test through the shared
verify lock.

| command | exit | verdict |
|---|---|---|
| `pnpm --filter @object-ui/plugin-kanban type-check` (`tsc --noEmit` + `tsconfig.test.json`) | 0 | no `error TS` line |
| `pnpm exec vitest run --maxWorkers=2 packages/plugin-kanban/` | 0 | `Test Files 25 passed (25) · Tests 130 passed (130)` |
| `turbo run type-check --filter=@object-ui/runner --filter=@object-ui/app-shell` | 0 | `Tasks: 31 successful, 31 total` (the two package-level consumers) |
| `pnpm --filter @object-ui/plugin-kanban lint` | 0 | `✖ 121 problems (0 errors, 121 warnings)`; 0 new on touched files |
| `pnpm --filter @object-ui/plugin-kanban build` | 0 | `Declaration files built in 4378ms` |
| `node scripts/check-changeset-presence.mjs` | 0 | `✅ 8 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` |
| `node scripts/check-control-bytes.mjs` | 0 | `✅ OK (scanned 6511 tracked text file(s))` |
| `pnpm check:spec-symbols` | 0 | `✅ no comment cites a key its spec symbol does not declare` |
| `pnpm check:element-data-source-declaration` | 0 | `OK — 13 gate-consuming file(s) checked` |
| `pnpm check:unreferenced-sources` | 0 | `OK  Every shipped source file … is reachable` |
| `pnpm check:vi-mock-specifiers` / `check:vi-mock-inherit` / `check:self-import` / `check:phantom-deps` | 0 | all `✅` |
| `pnpm check:sdui-registration-pins` | **NOT MEASURED** | `❌ No console build to weigh at apps/console/dist/assets` — a prerequisite complaint, not a red gate. Narrowing declared: this diff adds and removes **no** registration, which is that gate's whole subject. CI runs it with a real console build. |

`turbo ls --affected` from the merge-base names 8 packages; the two with a real source-level
dependency on this one (`runner`, `app-shell`) are type-checked above, the rest are apps and
examples that reach this package only through plugin registration and are CI's.

### Reverse verification at the package boundary

A widening is invisible unless the consumer really reads the rebuilt `.d.ts`. A scratch probe
assigning an `ObjectKanbanSchema`-typed value into the prop's `schema` member, importing
through `dist/index`:

* against the rebuilt `.d.ts` → **exit 0**;
* with that one line in `dist/ObjectKanban.d.ts` reverted to the single arm (mutation proven by
  md5, restored and re-proven) → **exit 2**, `TS2322: Type 'ObjectKanbanSchema' is not
  assignable to type 'KanbanSchema'`.

So the green is a reading of this change, not of a cache.

### Ablation — both halves of the new pin, red-first

Run from the committed state, mutation proven by blob hash on disk, restored under
`trap … EXIT INT TERM` with `git checkout HEAD --` and `git diff HEAD` empty afterwards. No
`dist` is on the pin's own resolution path (`../ObjectKanban` and `../index` are source imports;
`tsconfig.test.json` sets `"paths": {}` only for the `@object-ui/*` workspace deps, which this
change does not touch).

**Leg A — prop reverted to the single arm** (the union member replaced by `schema: KanbanSchema`;
anchors counted 1→0 and 0→1, blob `48bd449d` → `fba3ea76`).
Predicted red on the two new type legs and on the four `satisfies` fixtures. Observed:
`tsc -p tsconfig.test.json` **exit 2, 6 errors** — `TS2344` at
`object-kanban-component-props-7322.test.ts(77,43)` and `(89,41)` (`_AcceptsTheObjectKanbanNode`,
`_DiscriminantIsTheTwoKeys`) plus `TS2322` on all four fixtures. One refinement on the
prediction: the fixtures fail at the **JSX prop assignment** (`TS2322`), not at the `satisfies`
clause (`TS1360`) — the literals still satisfy `ObjectKanbanSchema`; it is the prop that stops
accepting them, which is the more exact statement of the defect.

**Leg B — one registration re-keyed** (`'kanban'` → `'kanban2'` in `index.tsx`; blob
`4e354382` → `c01425be`). Predicted exactly one red assertion with the anti-vacuity legs
staying green. Observed: vitest **exit 1**, `Tests 1 failed | 3 passed (4)`,
`AssertionError: expected [ 'kanban2', 'object-kanban' ] to deeply equal [ 'kanban', 'object-kanban' ]`.
⚠️ Leg B's `grep -c` echo lines were mangled by quoting inside the script's heredoc and printed
no usable count; the mutation's arrival on disk is carried instead by the blob-hash inequality
**and** by the assertion message itself, which quotes the mutated key back. Two independent
readings, so the leg stands; the broken echo is a script defect, reported rather than papered
over.

Restore proven on both legs: blobs equal `HEAD` again, `git diff HEAD` empty.

## Findings — reported, not filed, to avoid a fourth card on this family

The dispatch flagged that #7772 / #7773 / #7780 already sit on the `object-kanban` family and
this card must not duplicate them. Two measurements fall inside cards that already exist:

* **`filter` and `navigation` are declared on neither kanban face** while `ObjectKanban` reads
  both. #7742 is already the enforce-or-remove ledger card for `KanbanSchema`'s
  declared-vs-read set (it names the undeclared `titleField` read), and #7712 covers `filter`
  from the registration-`inputs` angle. `navigation` is named by neither; it is added as a
  comment on #7742 rather than as a new card.
* ⚠️ **This PR moves #7742's premise.** Its `titleField` half reads "the board reads an
  undeclared `titleField`" — after this change that read is declared (on the `object-kanban`
  arm) and cast-free, while still undeclared on `KanbanSchema`. Its taker should re-measure
  from this commit.

## Not touched

The four view-level `groupField` alias sites (`normalize-list-view.ts`, `ListView.tsx`,
`ObjectView.tsx`), `BaseSchema`'s index signature (#5155), and `packages/types` — read-only
here, as dispatched (slot 3/5 is editing `complex.ts`'s `ChatbotSchema` region concurrently).
Nothing under `content/docs/releases/`.

## Governance

`node scripts/check-governed-queue-guard.mjs --test` on this diff's paths:
`✅ NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched.`
Held **draft** with `needs:contract-review` per Clause-② (this widens a member of a prop type
exported from a published package). ⛔ Does not enqueue while that label is on it.

---

---

### DIFF
--- a/.changeset/7322-object-kanban-component-props.md
+++ b/.changeset/7322-object-kanban-component-props.md
@@ -1,0 +1,55 @@
+---
+'@object-ui/plugin-kanban': minor
+---
+
+`ObjectKanbanComponentProps.schema` names both node types the component is registered for
+(objectui#7322 item ②, following the objectui#5903 / #5018 land shape).
+
+`ObjectKanbanRenderer` is registered under two keys — `'object-kanban'` and `'kanban'` —
+and the two keys have different declared node types: `ObjectKanbanSchema` (`type:
+'object-kanban'`, `objectName` and `groupBy` required) and `KanbanSchema` (`type:
+'kanban'`, both optional). The prop named `KanbanSchema` alone, so **no `object-kanban`
+node was assignable to the component that renders it**, and the discriminants are disjoint
+string literals, so no cast-free annotation existed for half the boards this component
+serves. It is now the union of the two.
+
+## What settled it: the read set
+
+`ObjectKanban` reads thirteen keys off `schema`. Neither declaration covers them; the two
+TOGETHER cover twelve, and each arm is load-bearing:
+
+- `objectName`, `groupBy`, `limit`, `cardFields` — declared on both;
+- `columns`, `cardTitle`, `swimlaneField`, `grouping` — `KanbanSchema` only;
+- `titleField` — `ObjectKanbanSchema` only (which is why that read was spelled
+  `(schema as any).titleField`);
+- `data`, `bind`, `className` — `BaseSchema`;
+- `filter` — declared by **neither** face, still riding `BaseSchema`'s index signature.
+  Measured and reported, **not** changed here: this card moves the prop, not the two
+  published schema faces.
+
+So naming `ObjectKanbanSchema` alone — the remedy the original card implied — would have
+been wrong in the other direction: it drops four declared reads and the `'kanban'`
+registration.
+
+## Not affected
+
+Widening a member of an exported prop type is additive: every caller that passed a
+`KanbanSchema` still compiles, and the union claims exactly the accept set the registry
+dispatches to this component — a third node type is still turned away. The runtime is
+untouched; `ObjectKanbanRenderer` still takes `schema: any`, so no shape is turned away
+there either (the objectui#5903 disposition, restated). The view-level `kanban.groupField`
+alias, `BaseSchema`'s index signature, and `@object-ui/types` are all untouched.
+
+## Casts this removes
+
+Inside `ObjectKanban.tsx`, three schema-key reads drop their `as any`: `titleField` (two
+sites, now honest because the `object-kanban` arm declares it) and `cardFields` /
+`cardTitle` (already declared; the casts were redundant). `(schema as any).navigation`
+**stays** — `navigation` is declared on neither face, so removing the cast would change
+nothing but the spelling of an index-signature read.
+
+Four of the six in-package fixtures that mount an `object-kanban` board drop their
+`as never` escape for a real `satisfies ObjectKanbanSchema`. The other two are static
+boards (`columns` + inline `data`, no fetch) that author no `objectName`, which
+`ObjectKanbanSchema` declares required — objectui#7780's subject; their casts stay, now
+carrying the reason and the card number.
--- a/packages/plugin-kanban/src/ObjectKanban.markedRefusalToast.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.markedRefusalToast.test.tsx
@@ -51,7 +51,7 @@ import '@testing-library/jest-dom';
 import { registerAllFields } from '@object-ui/fields';
 import { toast } from '@object-ui/components';
 import { normaliseClientError } from '@object-ui/data-objectstack';
-import type { DataSource } from '@object-ui/types';
+import type { DataSource, ObjectKanbanSchema } from '@object-ui/types';
 import { ObjectKanban } from './ObjectKanban';
 
 // Pay the board's lazy chunk at import time rather than racing it against a
@@ -112,7 +112,7 @@ const schema = {
     { id: 'backlog', title: 'Backlog' },
     { id: 'in_progress', title: 'In Progress' },
   ],
-} as never;
+} satisfies ObjectKanbanSchema;
 
 const serverRecords = () => [{ id: 't1', title: CARD, status: 'backlog' }];
 
--- a/packages/plugin-kanban/src/ObjectKanban.navWidthDefault.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.navWidthDefault.test.tsx
@@ -45,6 +45,7 @@
 import React from 'react';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, cleanup, fireEvent, waitFor } from '@testing-library/react';
+import type { ObjectKanbanSchema } from '@object-ui/types';
 import { ObjectKanban } from './ObjectKanban';
 
 // Pay the board's lazy chunk at import time, not inside a `findBy` budget
@@ -84,7 +85,7 @@ async function openDrawer(navigation?: Record<string, unknown>) {
         columns: [{ id: 'todo', title: 'To Do' }],
         data: cards,
         ...(navigation ? { navigation } : {}),
-      } as never}
+      } satisfies ObjectKanbanSchema}
     />,
   );
   const card = await screen.findByText('On the board');
--- a/packages/plugin-kanban/src/ObjectKanban.overlayTitleI18n.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.overlayTitleI18n.test.tsx
@@ -96,6 +96,15 @@ function renderKanbanIn(language: string, schemaExtra: Record<string, unknown>) 
           columns: [{ id: 'todo', title: 'To Do' }],
           data: cards,
           ...schemaExtra,
+        // ⚠️ Still escaped, and NOT by oversight (objectui#7322 item ②). This
+        // board is STATIC — `columns` plus inline `data`, no fetch — so it
+        // authors no `objectName`, which `ObjectKanbanSchema` declares REQUIRED.
+        // That requiredness is objectui#7780's subject (the renderer reads inline
+        // `data` ahead of the fetch and never needs the object name for a board
+        // like this one); until it is answered, no declared type accepts this
+        // node. The other four in-package fixtures that mount an `object-kanban`
+        // board dropped their `as never` for a real `satisfies ObjectKanbanSchema`
+        // in objectui#7322 — when #7780 lands, this one follows.
         } as never}
       />
     </I18nProvider>,
--- a/packages/plugin-kanban/src/ObjectKanban.overlayTitleNoProviderFallback.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.overlayTitleNoProviderFallback.test.tsx
@@ -68,6 +68,15 @@ function renderKanban(schemaExtra: Record<string, unknown>) {
         columns: [{ id: 'todo', title: 'To Do' }],
         data: cards,
         ...schemaExtra,
+      // ⚠️ Still escaped, and NOT by oversight (objectui#7322 item ②). This
+      // board is STATIC — `columns` plus inline `data`, no fetch — so it
+      // authors no `objectName`, which `ObjectKanbanSchema` declares REQUIRED.
+      // That requiredness is objectui#7780's subject (the renderer reads inline
+      // `data` ahead of the fetch and never needs the object name for a board
+      // like this one); until it is answered, no declared type accepts this
+      // node. The other four in-package fixtures that mount an `object-kanban`
+      // board dropped their `as never` for a real `satisfies ObjectKanbanSchema`
+      // in objectui#7322 — when #7780 lands, this one follows.
       } as never}
     />,
   );
--- a/packages/plugin-kanban/src/ObjectKanban.rejectedMoveRollback.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.rejectedMoveRollback.test.tsx
@@ -50,7 +50,7 @@ import { render, screen, within, act, cleanup, waitFor } from '@testing-library/
 import '@testing-library/jest-dom';
 import { registerAllFields } from '@object-ui/fields';
 import { toast } from '@object-ui/components';
-import type { DataSource } from '@object-ui/types';
+import type { DataSource, ObjectKanbanSchema } from '@object-ui/types';
 import { ObjectKanban } from './ObjectKanban';
 
 // Pay the board's lazy chunk at import time rather than racing it against a
@@ -105,7 +105,7 @@ const schema = {
     { id: 'backlog', title: 'Backlog' },
     { id: 'in_progress', title: 'In Progress' },
   ],
-} as never;
+} satisfies ObjectKanbanSchema;
 
 /** Server truth: the card is in `backlog` and the server never moves it. */
 const serverRecords = () => [{ id: 't1', title: CARD, status: 'backlog' }];
--- a/packages/plugin-kanban/src/ObjectKanban.requiredWhenPrompt.test.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.requiredWhenPrompt.test.tsx
@@ -68,7 +68,7 @@ import { render, screen, within, act, cleanup, waitFor, fireEvent } from '@testi
 import '@testing-library/jest-dom';
 import { registerAllFields } from '@object-ui/fields';
 import { toast } from '@object-ui/components';
-import type { DataSource } from '@object-ui/types';
+import type { DataSource, ObjectKanbanSchema } from '@object-ui/types';
 import { ObjectKanban } from './ObjectKanban';
 
 // Pay the board's lazy chunk at import time rather than racing it against a
@@ -127,7 +127,7 @@ const schema = {
     { id: 'negotiation', title: 'Negotiation' },
     { id: 'closed_won', title: 'Closed Won' },
   ],
-} as never;
+} satisfies ObjectKanbanSchema;
 
 const records = (winReason: unknown = null) => [
   { id: 'o1', name: DEAL, stage: 'negotiation', win_reason: winReason },
--- a/packages/plugin-kanban/src/ObjectKanban.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.tsx
@@ -7,7 +7,7 @@
  */
 
 import React, { useEffect, useState, useMemo } from 'react';
-import type { DataSource } from '@object-ui/types';
+import type { DataSource, ObjectKanbanSchema } from '@object-ui/types';
 import {
   useDataScope,
   useNavigationOverlay,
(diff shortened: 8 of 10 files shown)


os-justin:
Contract review — **PASS** — PR #8153 (objectui#7322 **item ②**), head `25907cd70`

Reviewed-by: `domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`, in-seat at **`opus` = `TIER_DEFAULT`**. `CONTRACT_REVIEW_TIER` (`claude-fable-5-1`) is measured unavailable (429); the exemption permits falling back to default and ⛔ no lower. ⛔ This review is not described as running at contract-review tier.

## ⭐ The union-vs-single-type question — re-measured by this seat, ⛔ not accepted from the report

This was the dispatch's central clause. I warned that the card's implied remedy ("point the prop at `ObjectKanbanSchema`") was **not obviously correct** because one component serves two registrations, and told the dev to measure before writing a type and to stop if no single type could serve both. It measured, and the answer is a union. I re-ran the cross-tabulation on `origin/main` = `21d7989fb` and **every cell matches**:

| key | `KanbanSchema` | `ObjectKanbanSchema` | |
| :-- | :--: | :--: | :-- |
| `titleField` | **0** | **1** | ⇒ naming `KanbanSchema` alone drops it — which is exactly why it was `(schema as any).titleField` |
| `columns` · `cardTitle` · `swimlaneField` · `grouping` | **1** each | **0** each | ⇒ naming `ObjectKanbanSchema` alone (**the card's remedy**) drops four declared reads |
| `objectName` · `groupBy` · `limit` · `cardFields` | **1** each | **1** each | the **control** — item ①'s keys, disposition independently known |
| `filter` · `navigation` | **0** | **0** | on neither face — the out-of-scope finding |

⇒ **the union is provably the only honest type, and the card was wrong in both directions.** The control row is what makes the other rows readings: a tabulation that answered "declared" for everything would show it here, and it does not — `bind` / `className` come back BaseSchema-only on the same instrument. ✅ The stop-and-report clause was correctly **not** triggered; each arm is load-bearing.

## ① Derived judgments

**Direction: WIDENS** `ObjectKanbanComponentProps.schema`. Additive for callers (more values accepted); for anyone *reading* `props.schema` it is a union, which is the honest description of what the component is actually handed.

⚠️ **One caveat I record rather than gloss:** the three `as any` casts drop out, but on the arm that does **not** declare a key the read still resolves through `BaseSchema`'s `[key: string]: any`. So `schema.titleField` compiles on the `KanbanSchema` arm as `any`, and the union of that with `string | undefined` is `any`. ⇒ the honesty gain here is **at the declaration and at the cast count**, ⛔ not yet at the type of the read. That residue is objectui#5155's index-signature blind spot, which the dispatch fenced off and which this PR correctly does not touch. Saying so because "the casts are gone" could otherwise be read as "the reads are now typed", and they are not.

**⭐⭐ The anti-vacuity pin is the best thing in this diff.** `_RejectsAnUnregisteredNode` asserts that **`ObjectGridSchema`** — a sibling node type declared in the *same file* as `ObjectKanbanSchema`, registered to a *different* renderer — is **not** assignable to the union. That is the one leg that would go green if anyone later widened the prop to `any` or to a structural face. ⇒ a widening pinned only by "these two are accepted" is satisfied by `any`; this one is not. Correct instinct, correctly placed.

**Reverse verification at the package boundary** — a widening is invisible unless the consumer reads the rebuilt `.d.ts`. Probe assigns an `ObjectKanbanSchema` value through `dist/index` → exit 0; the same line in `dist/ObjectKanban.d.ts` reverted to the single arm (md5 `7874c417…` → `3f8384ec…`, restored) → exit 2, `TS2322`. ✅ The green reads this change, not a cache.

## ② semver

`@object-ui/plugin-kanban: minor`, following objectui#5903's precedent bump for the same manoeuvre. ✅ `major` forbidden; `Changeset Bump Policy` green.

## ③ Boundary flags

- ⭐ **Ablation leg B's on-disk anchor echo was VOID** (heredoc quoting swallowed the count) and the dev **said so** rather than papering over it — the leg stands on two independent readings instead (blob-hash inequality, plus the assertion message quoting the mutated key `kanban2` back). And it was ⛔ **not re-run**, with the reason stated: *「re-running until something lands is the defect one level up」*. ⭐ That is the correct handling of a partially-void measurement, and the sentence is worth keeping.
- **Fixture triage done per fixture, ⛔ not as a batch.** Four are fully assignable and move from `as never` to `satisfies ObjectKanbanSchema` (the strongest honest form — checks assignability without erasing the literal). **Two keep their casts** because they are *static* boards authoring no `objectName`, which `ObjectKanbanSchema` declares required — **measured**, not assumed (removing the casts produced `TS2322` on the prop and `TS1360` on the satisfies clause, naming `objectName`), and that requiredness is objectui#7780's open subject. The casts now carry the reason and the card number. ✅ Exactly right: silently rewriting those two would have fossilised a disputed requiredness into six fixtures.
- **The one pin amended, with its reason:** objectui#7664's prop leg moves from an identity assertion (`Equal` against `KanbanSchema`) to **assignability** — the form that ruling's own words claim, and the form its own kanban-ui leg already used. ⇒ amending a pin whose assertion the change falsifies, in the direction the ruling actually states, ⛔ not deleting an inconvenient one.
- **Lint delta measured, not asserted:** base and head `ObjectKanban.tsx` linted side by side in one invocation — **28 → 24** `no-explicit-any` findings, i.e. exactly the four `any` tokens the three removed casts contained. ⭐ A count that matches the diff arithmetic is a reading; "lint is green" would not have been.
- `check:sdui-registration-pins` **NOT MEASURED** (no `apps/console` build) with a declared narrowing — the gate's whole subject is registrations and this diff adds and removes none. ✅ Correctly refusing to call a prerequisite failure a pass; CI measures it.
- ⭐ **Out-of-scope findings reported as a COMMENT on objectui#7742 rather than as a fifth card**, because the dispatch warned four cards already sit on this family and both keys (`navigation` at `:789`, `filter` at `:310`) fall inside #7742's stated scope. ⇒ declining to file is as much a discipline as filing. It also left a **PREMISE NOTICE**: this PR moves #7742's own premise (its `titleField` half now reads differently), so its taker must re-measure from `25907cd70`. ⭐ Warning the *next* card that you moved its ground is rare and valuable.
- ⭐ The report **corrected its own `mcp_calls` count** (posted 9, actual 11 — the two read-backs necessarily post-dated the comment) and explained why it did not edit the comment in place. Unprompted, on a number nobody would have checked.
- `open_questions`: none. `packages/types` untouched; the four view-level `groupField` alias sites untouched; `BaseSchema` untouched. ✅ All three fences held.

## Verdict

**PASS.** `needs:contract-review` is attached and read back on the PR (`[plugin, plugin: kanban, tests, needs:contract-review]`, union exact, nothing stripped by the auto-labeler) — **cleared in this stroke**, with this comment as its provenance.

⛔ **Not landing yet.** CI unconverged: **32** enumerated against `total_count: 32`, **6 still `in_progress`** (Test shards 1–4, `Type Check`, `Live E2E`), **zero failures**. ⭐ **`Lint` read explicitly and green** (`completed_at` 21:01:07Z). `Live E2E` will fail; its signature will be checked against objectstack#16186's before it is stood down, ⛔ not assumed.

This PR carries the closing keyword — item ① landed as PR #7774 and item ② is the last item — so objectui#7322 closes on merge and this seat strips `pm:dispatched` and clears the assignee in the same stroke.

---

os-justin:
Addendum to the review above — `needs:contract-review` is now **cleared** on the PR (labels read back: `[plugin, plugin: kanban, tests]`), as that comment said it would be in the same stroke.

Two things in the PR body that the `os-dev-report` did not carry, and that the review should record because both are **conditional clauses of the dispatch being answered rather than skipped**:

**⭐ The `index.tsx:395` `schema: any` clause was answered NO, on measurement.** My dispatch authorised cleaning it up *「if and only if your measurement makes an honest type available there」*. The dev's answer is that it does not, on two readings: `ComponentRegistry.register`'s component parameter is **untyped**, so an annotation there would be a claim about registry dispatch that nothing enforces; and `registration.test.tsx:32` renders a deliberately minimal `{ type: 'object-kanban' }` that **no** declared type accepts while objectui#7780 is open. objectui#5903 took the same disposition for the same reason. ⇒ an "if and only if" resolved to **no**, with the reason, is the clause working — ⛔ not a step skipped. A dev that had cleaned it up to look thorough would have made a claim the registry does not enforce and broken a live test.

**`(schema as any).navigation` at `:789` also stays**, and correctly: `navigation` is declared on **neither** face (I verified: 0 and 0), so removing that cast would change the spelling of an index-signature read and nothing else. That is objectui#5903's disposition restated. ⇒ it also **narrows** my own caveat above — the residual `any` I flagged is now confined to the keys no face declares, which is objectui#7742's ledger, ⛔ not this card's.

⭐ And the union shape has a **house precedent I had not named in the dispatch**: objectui#7311 already landed `ObjectCalendarComponentProps.schema: ObjectGridSchema | CalendarSchema` — the same manoeuvre on the calendar, the sibling card triage twice pointed at. ⇒ this is the family's established land shape, ⛔ not an invention.

Nothing in either changes the verdict: **PASS**, landing on convergence.

---

os-justin:
Landing — PR #8153 (objectui#7322 item ②), head `25907cd70`, 2026-09-06T21:1xZ.

**Pre-landing three:**

① Contract review **PASS** (comment 5562152707) at `opus`, addendum 5562155867; **`needs:contract-review` cleared** with that as provenance. Head unchanged since the review.

② **CI converged, enumerated 32 against `total_count: 32`** — **28 success · 3 skipped · 1 failure**, ⛔ zero `in_progress`. ⭐ **`Lint` read explicitly and green** (`completed_at` 21:01:07Z).

The single failure is **`Live E2E (informational)`**, and its signature was **read and matched line for line** (job `101557334744`), ⛔ not assumed:

```
[live-backend] not ready after 300s — last 100 log lines:
  ⚠ AuthPlugin failed to load: The requested module '@better-auth/core/db'
    does not provide an export named 'createLocalAccountIssuer'
  WARN CORE: Core service missing, functionality may be degraded: auth
  … no such table: sys_organization / sys_position / sys_permission_set / sys_user
  ✓ Server is ready          ← 42 plugins, 132 seeded rows, 31 boot warnings
##[error]Process completed with exit code 1.
```

Identical to #8142's, #8145's and #8150's. Carded **objectstack#16186**; the red is **by design** per objectui#7689, whose triage ⛔ forbids reverting the pin to get green. ⛔ No re-run spent.

⚠️ And a mechanism argument alongside the signature match, ⛔ not instead of it: the step boots a **published** `@objectstack/*` backend and dies at readiness *before* any objectui code is exercised, so a prop-type change in `plugin-kanban` cannot reach it.

③ **Ready flipped, auto-merge SQUASH enabled.** ⚠️ The API echoes an empty `method` even when it applied — verify by queue ref and parent-count, ⛔ not by that echo. #8150 is ahead in the queue; this one stacks behind it.

**⭐ Landing control established BEFORE the queue takes it**, so the post-merge confirmation is a reading and ⛔ not a hopeful grep:

| read | branch | `origin/main` |
| :-- | :--: | :--: |
| `KanbanSchema \| ObjectKanbanSchema` in `packages/plugin-kanban/src/ObjectKanban.tsx` | **1** | **0** |
| `ObjectKanban.tsx` line count on `main` (instrument live?) | — | **1064** |

⇒ the zero is a reading on a live file.

`Fixes #7322` — item ① landed as PR #7774 and item ② is the last item — so the card closes on merge and this seat strips `pm:dispatched` and clears the assignee in the same stroke.

---

PR #8159, opened 2026-09-06
fix(charts): honour series `opacity` and `dashArray` on every series, not only a comparison one

### LINKED ISSUE #7698: finding(plugin-charts): series-level `opacity` and `dashArray` are honoured only on a `variant: 'comparison'` series — read and unused on a primary series, while the spec declares them as unconditional overrides
## Measured (objectui `origin/main` @ `6eebc54b6`, re-read on PR #7684's merged head)

`@objectstack/spec@17.2.0` declares `ChartSeries.dashArray` as "Override stroke dash pattern" and `ChartSeries.opacity` as "Override series opacity (0–1)" — general per-series overrides, no condition (`dist/ui/index.js:2397-2399`). `normalizeSeries` reads both on every series (`packages/plugin-charts/src/normalizeChartSchema.ts:248-251`), and `core/utils/chart-presentation.ts:130-131` forwards both on the dataset path.

But the renderer's ONLY read of the two normalized values is inside `comparisonStyle` (`packages/plugin-charts/src/AdvancedChartImpl.tsx:92-97`):

```ts
const comparisonStyle = (s: any, kind) => {
  if (s?.variant !== 'comparison') return null;
  const strokeOpacity = typeof s.opacity === 'number' ? s.opacity : …;
  const fillOpacity   = typeof s.opacity === 'number' ? s.opacity : …;
  const strokeDasharray = s.dashArray ?? …;
  return { strokeOpacity, fillOpacity, strokeDasharray };
};
```

All five call sites (`:1825`, `:1892`, `:2035`, `:2046`, `:2054`) apply `cmp?.fillOpacity` / `cmp?.strokeOpacity` / `cmp?.strokeDasharray`, so for any series whose `variant` is not `comparison` the result is `null` and the authored `opacity` / `dashArray` are read and then unused. Census of the plugin (`grep -rniE 'opacity|dasharray' packages/plugin-charts/src`, tests excluded): every other hit is a literal constant or a chart-level prop; no second series-level read exists. Control: the same grep finds the five `cmp?.…` sites and the two `normalizeSeries` reads.

Consequence: a hand-authored `{ name: 'cost', opacity: 0.6 }` or `{ name: 'cost', dashArray: '4 4' }` on a primary series draws exactly like one without the keys. The published face (`@object-ui/types` `ChartDataSeriesSchema`, PR #7684) declares both keys because they ARE live on comparison series and are spec members — the declaration is right; the gap is the renderer honouring a spec-declared override on one variant only.

## Why this is a renderer finding, not a schema one (AGENTS.md #0 / #0.1)

The spec is the contract of record and it describes both keys as unconditional overrides. Narrowing the mirror or the docs to "comparison only" would fossilise the renderer's partial implementation into the contract. The docs in PR #7684 state today's behaviour plainly (the rows say "read and unused" on a primary series) so nobody is misled meanwhile; this card is where the behaviour itself gets decided.

## Options

- **A — honour both keys on every series** (recommended, contract-first): apply the authored `opacity` / `dashArray` regardless of `variant`, and keep the comparison-specific DEFAULTS (muted opacity, `4 4` dash) gated on `variant === 'comparison'`. Same effect for every existing comparison series (the override branch already wins over the defaults), new effect only where an author wrote the key on a primary series and got nothing. Pin: a primary-series `opacity: 0.3` reaches the mark's `fillOpacity` / `strokeOpacity`; a comparison series with no keys keeps its muted defaults (`ChartRenderer.specSeries.test.tsx` harness).
- **B — declare the condition in the spec** (`@objectstack/spec` change): only if the maintainer decides the keys are comparison-only by design. Then the mirror's `.describe` and the docs follow the spec.

Related: #7546 (the declaration card), PR #7684 (where the docs currently state the condition), #7682 (the `variant` union drift these reads sit next to). No closing keyword on purpose — none of those is resolved by this card.

Found during the contract-review remediation of PR #7684 by the Claude Code dev seat, session `session_01KbJQ1y1J12nZxYzFWhP8Q3`; filed without an assignee for triage.

---

os-justin:
Fixes #7698

`@objectstack/spec` declares `ChartSeries.opacity` ("Override series opacity") and `ChartSeries.dashArray` ("Override stroke dash pattern") as **unconditional** per-series overrides, and `@object-ui/types`' `ChartDataSeriesSchema` mirrors both. `normalizeSeries` read them off every series and the renderer then honoured them on a `variant: 'comparison'` overlay only — so an author who wrote `{ name: 'cost', opacity: 0.6 }` on a primary series got a mark drawn exactly as if the key were absent. Option A per the card's adjudication: the **renderer** moves, not the declaration. Narrowing the published face to "comparison only" would be the renderer's tolerance dictating the contract, the direction AGENTS.md #0.1 forbids.

## My own measurement — the card's body is wrong about where the defect lives, and the correction holds

Re-measured on `origin/main` @ `c14d3a090` with the card's own census control rebuilt (`grep -rniE 'opacity|dasharray' packages/plugin-charts/src`, tests excluded). The card's body says all five `cmp?.…` call sites apply `fillOpacity` / `strokeOpacity` / `strokeDasharray`. They do not — and there are **seven** of them, not five (the body's list omits the two Line sites):

| mark | line on `c14d3a090` | what it actually passed |
| :-- | :-- | :-- |
| Scatter | `:1832` | `fillOpacity` only |
| Line (combo) | `:1898` | `strokeOpacity`, `strokeDasharray` |
| Area (combo) | `:1905` | `fillOpacity`, `strokeOpacity`, `strokeDasharray` |
| Bar (combo) | `:1911` | `fillOpacity` only |
| Bar | `:2037` | `fillOpacity` only |
| Line | `:2048` | `strokeOpacity`, `strokeDasharray` |
| Area | `:2056` | `fillOpacity`, `strokeOpacity`, `strokeDasharray` |

And `comparisonStyle:96` was `s.dashArray ?? (kind === 'line' || kind === 'area' ? '4 4' : undefined)` — the `??` takes the left side whatever the kind, so the helper already returned an **authored** dash for bar and scatter too. Confirmed: **what dropped it is the mark component, not the `variant` guard**, and a fix aimed only at the guard would not have repaired `dashArray` at all.

The ablation below is the lit control for that claim, not the reading alone. Against the pre-fix implementation, the comparison-bar pin reports `fillOpacity: '0.4'` present and `strokeDasharray: null` — the muted default arrived, the authored dash did not. That is the mark-component half, isolated.

## The fix — two halves, gated differently

`comparisonStyle` becomes `seriesStyle` and no longer returns `null`:

- **Authored `opacity` / `dashArray`** apply whatever the `variant`.
- **The muted comparison DEFAULTS** (lower opacity, the `'4 4'` line/area dash) stay gated on `variant === 'comparison'`.
- **Bar and Scatter marks** now pass `strokeOpacity` and `strokeDasharray` through, so the authored dash is no longer dropped at the mark.

**Existing comparison series are byte-for-byte unaffected.** The authored branch already won over the defaults, and the two *stroke* defaults no mark ever consumed (bar `0.6`, scatter `0.5` — this renderer strokes neither family) are now spelled `undefined`, so opening `strokeOpacity` on those marks hands them no default they never had. Pinned directly: a comparison bar carrying no keys keeps `fill-opacity="0.4"` and gains neither `stroke-opacity` nor `stroke-dasharray`.

A primary series carrying neither key is untouched — pinned as all three attributes absent.

Only a **stroked** mark can show a dash, and this renderer strokes Line and Area, so an authored `dashArray` on a bar or scatter reaches the mark and paints nothing. That is the mark's geometry, not a condition on the key, and the docs now say exactly that rather than describing it as a restriction.

## Read-coupling to PR #7684 (binding, discharged)

#7684 landed and stated the comparison-only condition as an **interim** measure. That wording is now wrong and comes out in this same change, per objectui#7070:

- `packages/types/src/data-display.ts` — the `ChartDataSeries.opacity` and `.dashArray` JSDoc. Comment-only: `git diff -U0` on that file yields **zero** changed lines outside a block comment.
- `content/docs/plugins/plugin-charts.mdx` — both series-table rows plus the paragraph under the area example. Prose only: the three changed lines are 193, 194 and 198, and the fences in that region are 162-179, 200-216, so no compiled snippet moved.

## Scope held

`packages/core/src/utils/chart-presentation.ts` was **not** touched — its `seriesPresentation` already forwards both keys on the dataset path, so the dispatch's flagged out-of-plugin file stays out of this PR.

Filed separately rather than folded in: **objectui#8157** — the radar mark maps over `series` but hard-codes `fillOpacity={0.6}` and never calls the helper, so `opacity`, `dashArray` and the whole comparison treatment are inert on a radar chart for every variant. Repairing it would newly mute existing comparison radar series (`0.6` becomes the helper's `0.5`), a visible change to already-drawn charts that wants its own decision. Not addressed here; it stays open.

Not touched here either, all separate cards in the same measurement family: objectui#7690, objectui#7694, objectui#7682, objectui#7546.

## Verification

All runs on the final commit `5d68aecbb` unless stated. Heavy runs went through the container's shared verify lock; wall-clock figures are shared-box seconds.

**Tests**

- `pnpm exec vitest run packages/plugin-charts/` — **48 files, 447 tests passed**.
- `pnpm exec vitest run packages/types/ packages/plugin-dashboard/ packages/plugin-report/ examples/schema-catalog/` — **271 files, 5611 tests passed** (the chart-rendering consumers).
- New pin: `packages/plugin-charts/src/ChartRenderer.seriesPresentation.test.tsx`, 13 tests.

**Reverse verification (ablation).** Predicted direction: RED. The one file was restored to its `c14d3a090` bytes, with the mutation proven on disk before the run — `git hash-object` of the working file equalled the pre-fix blob `63a5e34c9` and differed from the HEAD blob `ae52b1d01`; anchored counts moved `seriesStyle` 6 to 0 and `comparisonStyle` 0 to 6. Result: **8 of 13 failed**, and the 5 that passed are exactly the regression pins (the comparison defaults, the untouched primary series, and an authored `opacity` overriding a comparison default — all of which held before this change and must keep holding). Restore leg proven by state, not by an exit code: `git diff HEAD` empty, `git status` clean, working blob back to `ae52b1d01`. The script carried an absolute-path `trap ... EXIT INT TERM`.

**Type checking.** `pnpm --filter @object-ui/plugin-charts --filter @object-ui/types type-check` — exit 0. This package's `type-check` is `tsc --noEmit && tsc -p tsconfig.test.json`, so the new test file is type-checked too rather than excluded.

**Gates run locally** (all exit 0, quoting each gate's own verdict line): `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:doc-fences`, `check:doc-types`, `check:doc-links`, `check:doc-expression-carriage`, `check:doc-example-readers`, `check:unreferenced-sources`, `check:spec-symbols`, `check:shell-escape-residue`, `check-changeset-presence`, `check-changeset-no-major`, `check-changeset-fixed`, `check-changeset-overwrite`, and `check-governed-queue-guard --test` on all five changed paths (`NOT GOVERNED`, 5 of 5).

**Two gates NOT MEASURED locally, left to CI** — both exited on a stated precondition, neither is a finding:

- `check:doc-snippets` — exit 2, `PRECONDITION NOT MET`: it needs 32 packages plus closures built, which does not fit this container's foreground budget under contention. It compiles documented TS blocks, and this diff changes none (see the fence arithmetic above).
- `check:eager-closure` — exit 2, no `apps/console/dist/eager-closure.json`, so no ceiling to weigh. The diff adds no import, so the eager payload is unchanged.

**Lint — a declared, measured narrowing.** Repo-wide lint (`turbo run lint`, 40 packages each running `eslint .`) is CI's run. Locally, ESLint ran on the three changed source files: **3 files linted, 0 errors, 77 warnings**, file count read from `--format json` output length. Every warning is pre-existing `@typescript-eslint/no-explicit-any` in `AdvancedChartImpl.tsx`; the diff adds no new `any` (the replaced helper already took `s: any`), and no `--max-warnings` is set anywhere in the lint pipeline — `lint.yml` says so in as many words. Invariance for untouched files: **type-aware linting is not enabled** — `eslint.config.js` sets no `parserOptions.project` and no `projectService` (the only matches in the repo are two `RuleTester` fixtures under `eslint-rules/`), so every verdict is a per-file function of that file's own bytes, and this diff cannot move the verdict of a file it does not touch.

**Changeset.** `.changeset/7698-series-opacity-dasharray-unconditional.md`, `@object-ui/plugin-charts: minor` + `@object-ui/types: patch`, no `major` (AGENTS.md version alignment). Not a `skip-changeset` PR — this ships a user-visible behaviour change.

---

---

### DIFF
--- a/.changeset/7698-series-opacity-dasharray-unconditional.md
+++ b/.changeset/7698-series-opacity-dasharray-unconditional.md
@@ -1,0 +1,41 @@
+---
+'@object-ui/plugin-charts': minor
+'@object-ui/types': patch
+---
+
+Chart `series[].opacity` and `series[].dashArray` are honoured on every series,
+not only on a `variant: 'comparison'` one (objectui#7698).
+
+`@objectstack/spec` declares `ChartSeries.opacity` ("Override series opacity")
+and `ChartSeries.dashArray` ("Override stroke dash pattern") as unconditional
+per-series overrides, and `normalizeSeries` read both off every series. The
+renderer then honoured them on a comparison overlay only, so an author who
+wrote `{ name: 'cost', opacity: 0.6 }` or `{ name: 'cost', dashArray: '4 4' }`
+on a primary series got a mark drawn exactly as if the key were absent. Fixed
+in the renderer rather than by narrowing the published declaration to match:
+the spec is the contract of record, and a renderer's partial implementation
+does not get to dictate it (AGENTS.md #0.1).
+
+**Two gaps, not one.** The `variant` guard was the visible half — `comparisonStyle`
+returned `null` for any other variant. The second half only showed on
+`dashArray`: that helper already returned an AUTHORED dash for every family
+(the `??` takes the left side whatever the kind), and the **Bar and Scatter
+marks** then passed `fillOpacity` only, dropping `strokeDasharray` and
+`strokeOpacity` on the floor — so an authored dash was lost on those two
+families even on a comparison series. A fix aimed at the guard alone would have
+left that untouched. `comparisonStyle` is now `seriesStyle`, and the Bar and
+Scatter marks pass all three channels.
+
+**Comparison series are unaffected.** The authored branch already won over the
+muted defaults, and those defaults stay gated on `variant: 'comparison'`: a
+comparison series carrying neither key keeps its lower opacity and its `'4 4'`
+line/area dash exactly as before. The two stroke defaults no mark ever consumed
+(bar and scatter — neither is stroked by this renderer) are now spelled
+`undefined`, so opening `strokeOpacity` on those marks does not hand them a
+default they never had.
+
+Only a stroked mark can show a dash, so on a `bar` or `scatter` mark an
+authored `dashArray` reaches the mark and paints nothing — the mark's geometry,
+not a condition on the key. The `ChartDataSeries` mirror docs and the
+plugin-charts reference, which stated the comparison-only condition as an
+interim measure, are corrected in the same change.
--- a/content/docs/plugins/plugin-charts.mdx
+++ b/content/docs/plugins/plugin-charts.mdx
@@ -190,12 +190,12 @@ Each `series` entry names the column it plots (`dataKey`, or the spec spelling `
 | `stack` | `string` | Stack group id — series sharing one id stack together. |
 | `yAxis` | `'left' \| 'right'` | Which y-axis the series binds to, on a chart that declares a second y-axis. |
 | `variant` | `'primary' \| 'comparison'` | `comparison` draws the muted period-over-period overlay; `primary` (the default) is the normal treatment. |
-| `dashArray` | `string` | SVG `stroke-dasharray`, e.g. `"4 4"` for a dashed line. Today the renderer applies it only on a `variant: 'comparison'` series **and** only on a mark that has a stroke to dash — a `line` or `area` mark (the chart's `chartType`, or a per-series `type` override), where it overrides the overlay's default `"4 4"`; a comparison `bar` or `scatter` mark takes `opacity` only and drops the dash, and on a primary series of any family it is read and unused. |
-| `opacity` | `number` | Stroke and fill opacity override — any finite number (the spec bounds it to 0–1). Today the renderer applies it only on a `variant: 'comparison'` series, where it overrides the overlay's per-family default on every mark family — the fill of a `bar` or `scatter` mark, the stroke of a `line` mark, both on an `area` mark; on a primary series it is read and unused. |
+| `dashArray` | `string` | SVG `stroke-dasharray`, e.g. `"4 4"` for a dashed line. Applies on **any** series, whatever its `variant`, and overrides the comparison overlay's default `"4 4"`. Only a mark with a stroke can show a dash, and this renderer strokes `line` and `area` marks (the chart's `chartType`, or a per-series `type` override) — on a `bar` or `scatter` mark the value reaches the mark and paints nothing, which is the mark's geometry rather than a condition on the key. |
+| `opacity` | `number` | Stroke and fill opacity override — any finite number (the spec bounds it to 0–1). Applies on **any** series, whatever its `variant`, on every mark family — the fill of a `bar` or `scatter` mark, the stroke of a `line` mark, both on an `area` mark — and overrides the comparison overlay's per-family default. |
 
 `chartType` is **not** a series key. It is the renderer's internal spelling of `type` — `@objectstack/spec`'s `ChartSeries` lists it as an alias — and the validator refuses it by name (`Unrecognized key(s) on this chart series: chartType. Did you mean chartType → type?`) instead of dropping it silently, which is what it used to do. Write `type`.
 
-An `area` chart keeps every key in this example live: `stack` stacks the two primary areas, and the comparison overlay takes both the dash and the opacity (on a `bar` chart the overlay would take `opacity` only).
+An `area` chart keeps every key in this example live: `stack` stacks the two primary areas, and the comparison overlay takes both the dash and the opacity (on a `bar` chart the dash would still reach the mark, but a bar has no stroke to dash). Neither key needs `variant: 'comparison'` — a primary series may carry them too, and then it is the overlay's defaults, not the keys, that stay comparison-only.
 
 ```tsx
 const schema = {
--- a/packages/plugin-charts/src/AdvancedChartImpl.tsx
+++ b/packages/plugin-charts/src/AdvancedChartImpl.tsx
@@ -84,16 +84,36 @@ const TW_COLORS: Record<string, string> = {
 const resolveColor = (color: string) => TW_COLORS[color] || color;
 
 /**
- * Default visual treatment for a `variant: 'comparison'` series. Returns
- * overrides per chart family so the comparison overlay reads as muted
- * (dashed line, lower fill opacity) while still being color-matched to
- * the primary series. Series-level `opacity` / `dashArray` win over defaults.
+ * Per-series presentation overrides, in two halves that are deliberately NOT
+ * gated the same way (objectui#7698).
+ *
+ * - The AUTHORED keys — `@objectstack/spec`'s `ChartSeries.opacity` ("Override
+ *   series opacity") and `ChartSeries.dashArray` ("Override stroke dash
+ *   pattern") — are declared with NO condition, so they apply on every series
+ *   whatever its `variant`.
+ * - The muted comparison treatment (lower opacity, a `'4 4'` dash) is a
+ *   DEFAULT and stays gated on `variant: 'comparison'`, so the overlay still
+ *   reads as muted while a primary series carrying neither key is untouched.
+ *
+ * This function used to return `null` for anything but a comparison series, so
+ * an authored `opacity` / `dashArray` on a primary series was read by
+ * `normalizeSeries` and then discarded here — the renderer honouring a
+ * spec-declared unconditional override on one variant only. Narrowing the
+ * published declaration to match would have been the renderer's tolerance
+ * dictating the contract, which is the direction AGENTS.md #0.1 forbids.
+ *
+ * Comparison series keep every value they had. The authored branch already won
+ * over the defaults, and the two STROKE defaults no mark ever consumed — bar
+ * and scatter, neither of which this renderer strokes — are now spelled
+ * `undefined` rather than reaching the Bar and Scatter marks that gained
+ * `strokeOpacity` / `strokeDasharray` in the same change.
  */
-const comparisonStyle = (s: any, kind: 'line' | 'area' | 'bar' | 'scatter') => {
-  if (s?.variant !== 'comparison') return null;
-  const strokeOpacity = typeof s.opacity === 'number' ? s.opacity : (kind === 'line' || kind === 'scatter' ? 0.5 : 0.6);
-  const fillOpacity = typeof s.opacity === 'number' ? s.opacity : (kind === 'bar' ? 0.4 : kind === 'area' ? 0.2 : 0.5);
-  const strokeDasharray = s.dashArray ?? (kind === 'line' || kind === 'area' ? '4 4' : undefined);
+const seriesStyle = (s: any, kind: 'line' | 'area' | 'bar' | 'scatter') => {
+  const muted = s?.variant === 'comparison';
+  const authoredOpacity = typeof s?.opacity === 'number' ? s.opacity : undefined;
+  const strokeOpacity = authoredOpacity ?? (muted ? (kind === 'line' ? 0.5 : kind === 'area' ? 0.6 : undefined) : undefined);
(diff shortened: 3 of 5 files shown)


os-justin:
Contract review — **PASS** — PR #8159 (objectui#7698), head `5d68aecbb`

Reviewed-by: `domain:ui` @ objectui PM seat, `session_01YBWFb5YgMU5dw8p2VKj16S`, in-seat at **`opus` = `TIER_DEFAULT`**. `CONTRACT_REVIEW_TIER` (`claude-fable-5-1`) is measured unavailable (429). ⛔ Not described as running at contract-review tier.

## ⭐ The Zone-2 clause worked: the correction I handed over was itself incomplete

My dispatch handed the dev comment 5550238143's mark-component table **explicitly as the previous seat's reading, ⛔ not as fact**, and told it to rebuild the census with its own lit control. It did — and found the correction short.

Re-measured by this seat on `origin/main`, `cmp?.` call sites in `AdvancedChartImpl.tsx`:

```
:1832  Scatter        fillOpacity only
:1898  Line  (combo)  strokeOpacity + strokeDasharray
:1905  Area  (combo)  all three
:1911  Bar   (combo)  fillOpacity only
:2037  Bar            fillOpacity only
:2048  Line           strokeOpacity + strokeDasharray
:2056  Area           all three
```

⇒ **seven**, not the card's five (with stale numbers) and not the correction's four. The correction omitted both **Line** sites and **Area `:1905`**. ⭐ Three successive readings of the same file, each closer; the one that held is the one taken with a control rather than inherited.

The mechanism claim also confirmed at `:96`: `s.dashArray ?? (kind === 'line' || kind === 'area' ? '4 4' : undefined)` — the `??` takes the left side **whatever the kind**, so an authored dash was returned for bar and scatter too and then dropped **at the mark**. ⭐ And the dev proved that with a *lit control rather than a reading*: against the pre-fix implementation the comparison-bar pin reports `fillOpacity '0.4'` **present** and `strokeDasharray` **null** — the two channels diverging at one call site is what makes "the mark owns the drop" a measurement instead of an inference.

## ① Derived judgments

**Clause-②: NO, verified.** The `packages/types/src/data-display.ts` diff is **entirely inside JSDoc block comments** — `opacity?: number;` and `dashArray?: string;` are byte-identical, and no spec or mirror member moves. Hence `@object-ui/types: patch`, which is the honest minimum for a comments-only change. `@object-ui/plugin-charts: minor` for the behaviour. ✅

**Option A implemented as adjudicated, and both halves closed.** `comparisonStyle` → `seriesStyle`, no longer returning `null` off a comparison series, so an authored key applies whatever the `variant`; the muted **defaults** stay gated on `variant === 'comparison'`. And the Bar and Scatter marks now pass `strokeOpacity` / `strokeDasharray`, closing the half the card's body located in the wrong place.

**⭐⭐ The sharpest thing in this PR, and it is a defect the fix could easily have introduced.** The old helper computed `strokeOpacity` for bar (`0.6`) and scatter (`0.5`) — values **no mark ever consumed**, because this renderer strokes neither family. Opening `strokeOpacity` on those marks would therefore have handed them a default they never had, **newly muting every comparison bar and scatter stroke**. The dev spotted it and spelled those two `undefined`, then pinned exactly that:

> *「gives a comparison bar NO dash and NO stroke fade — the defaults are per family, and the marks that just gained these props must not inherit one」*

⇒ **opening a channel can leak a previously-inert default.** That is not in the card, not in triage, and not in my dispatch. It is the kind of thing that ships as "a small visual regression nobody can explain".

## ② semver and read-coupling

`minor` + `patch`, `major` forbidden, `Changeset Bump Policy` green. ✅

**The read-coupling my dispatch made binding is satisfied in BOTH places** — PR #7684's interim "comparison-only" wording is removed from the `plugin-charts.mdx` table rows **and** from the `ChartDataSeries` mirror JSDoc. objectui#7070's rule (restated, ⛔ not left to rot) is met, and the replacement wording is more precise than a simple deletion: it distinguishes *「the key is unconditional」* from *「only a stroked mark can show a dash — the mark's geometry, not a condition on the key」*. ⭐ That asymmetry is real and would have been easy to paper over.

## ③ Boundary flags

- **The pins render through `ChartRenderer` and read the DOM, ⛔ not `seriesStyle` directly** — with the reason stated: *「the defect was a value surviving normalization and then dying at the mark: only the DOM says whether the mark applied it」*. ⭐ A unit test on the helper would have passed against the broken tree.
- ⭐ `paint()` treats `null` — the DOM's answer for an absent attribute — as **a real reading**, and several pins assert exactly that. That is what lets "no default leaked" be asserted rather than assumed.
- ⭐ **One pin names the second gap explicitly**: *「reaches a COMPARISON bar mark (a fix at the `variant` guard alone would not)」*. It is the case that fails for a **different reason** than the primary-series cases, and it is pinned separately for that reason.
- **Ablation: 8 of 13 failed, and the 5 that passed are exactly the regression pins.** Mutation proven on disk before the run — working blob equal to the pre-fix blob `63a5e34c9` and different from `HEAD`'s `ae52b1d01`, with anchored counts moving `seriesStyle` 6→0 and `comparisonStyle` 0→6. Restore proven **by state**: `git diff HEAD` empty, `git status` clean, blob back to `ae52b1d01`.
- **The stale-`dist` hazard addressed by argument**: the pin imports `./ChartRenderer` and `./AdvancedChartImpl` relatively and the root vitest config aliases every workspace specifier to `src`, so the mutated source **is** the resolved module.
- ✅ `packages/core/src/utils/chart-presentation.ts` **untouched** — `seriesPresentation` already forwards both keys, so the dispatch's "say so explicitly if you touch it" clause resolved to *not needed*.
- **Two gates NOT MEASURED with stated preconditions**, and the first carries a specific argument rather than a shrug: `check:doc-snippets` compiles documented TS blocks, the three changed `.mdx` lines are `193/194/198`, and the fences in that region are `162-179` and `200-216` ⇒ this diff changes no compiled block. `check:eager-closure` has no `apps/console/dist/eager-closure.json` to weigh and the diff adds no import. ✅
- **Lint narrowing** carries all three legs, including the invariance argument (no `parserOptions.project` / `projectService`; the only repo matches are two RuleTester fixtures).

**⭐ Out-of-scope finding declined for a measured reason, not a scope reflex.** objectui#8157 — the radar mark hard-codes `fillOpacity={0.6}` and never calls the presentation helper, so authored `opacity`, `dashArray` **and the whole comparison treatment** are inert on radar for every variant. It was ⛔ not folded in because repairing it would **newly mute existing comparison radar series** (`0.6` → the helper's non-bar/non-area default `0.5`) — a visible change to already-drawn charts that wants its own decision. ⇒ declining because the fix has an **unbudgeted visible consequence** is a better reason than "out of scope", and it is the same care that produced the bar/scatter default fix above.

## Verdict

**PASS.** No Clause-② carrier is owed. ⛔ **Not landing yet** — CI has just started: **32** enumerated against `total_count: 32`, **10 `in_progress`** (`Lint`, Test shards 1–4, `Type Check`, `Build Docs`, `Bundle Analysis`, `README Export Check`, `Doc Snippet Type Check`, `Skill Example Check`, `Live E2E`), **zero failures**. ⚠️ `Lint` is unread and this seat will ⛔ not flip ready without it. `Live E2E` will fail; its signature will be matched against objectstack#16186's before it is stood down.

---

PR #8177, opened 2026-09-06
feat(types): export ComboboxOption from the @object-ui/types root entry

### LINKED ISSUE #7697: types: `ComboboxOption` is the authority after #7691 but is missing from the root barrel - reachable only via the `/form` subpath
Filed by the `domain:ui` PM seat from the contract review of PR #7691 (comment 5550050556), which raised it as a non-blocking recommendation explicitly conditioned on PR #7683 landing. **#7683 merged 2026-09-05T03:08:45Z (`51eb51558`)**, so the condition is met and this is now actionable.

## Measured on `origin/main` `6eebc54b6`

`ComboboxOption` is **absent from the `@object-ui/types` root barrel** while its siblings are present:

| name | hits on `packages/types/src/index.ts` |
| :-- | :-- |
| `ComboboxOption` | **0** |
| `SelectOption` (control) | 2 |
| `RadioOption` (control) | 1 |
| `ComboboxSchema` (control) | 1 |

The instrument is lit — the same word-boundary grep reads non-zero for all three controls, so the zero is a reading.

The type is declared at `packages/types/src/form.ts:1354`. The root barrel re-exports from `./form.js` as an **explicit named list** (closing at `index.ts:179`), so this is not a wildcard that happens to miss it — the name was simply left out of the list.

## Why it matters now rather than before

PR #7691 made `@object-ui/types` the **authority** for `ComboboxOption`: `@object-ui/components` stopped declaring its own copy and now re-exports the types one. The review measured that the components-face option is now literally `NonNullable&lt;ComboboxSchema['options']&gt;[number]`.

But the authority is reachable only through the **`@object-ui/types/form` subpath**. `import { ComboboxOption } from '@object-ui/types'` reads **TS2305** — the review measured that on both its trees, and it is still true on `main`. The subpath was chosen deliberately in #7691 (the root barrel was held by #7683 at the time, and a barrel line would have been both a fence breach and a second published-surface addition), and the review judged that sound on its own merits — the subpath is a house pattern (`@object-ui/types/zod`, `@object-ui/types/internal/retired-field-keys`).

So this card is not a defect in #7691. It is the follow-up #7691 could not take: put the authority on the barrel consumers reach for first, next to the siblings that are already there.

## Shape

Add `ComboboxOption` to the existing named re-export list from `./form.js` in `packages/types/src/index.ts`. ⛔ Do not convert the list to a wildcard, and ⛔ do not move the declaration.

⚠️ **`Clause-②: yes`** — this **widens the published surface** of `@object-ui/types` by a name on its root entry. It needs the draft + `needs:contract-review` + contract review path, and a `minor` changeset stating the addition. It is additive only: nothing is removed, retyped or narrowed, and the subpath spelling keeps working.

## Read-coupling to check before starting

objectui#7687 (`Combobox` never reads `options[].disabled`) is open and queued against the same type. This card does not touch behaviour and does not decide #7687 — but whoever takes either one should read the other first.

Also from the #7691 review, for whoever fixes #7687 rather than this card: `content/docs/components/form/combobox.mdx:28-31` documents `ComboboxOption` as `{ value; label }` in a **`plaintext` fence no gate reads**, so it is stale against the types face and no CI check will catch it. It was already stale before #7691.

---

os-justin:
Fixes #7697

One line on a barrel, one pin, one `minor` changeset.

`ComboboxOption` joins the `@object-ui/types` root barrel's **existing named re-export
list** from `./form.js`, next to `ComboboxSchema` — the way `SelectSchema`/`SelectOption`
and `RadioGroupSchema`/`RadioOption` already sit on that same list.

> ⚠️ **Draft, carrying `needs:contract-review`.** This widens the **published surface** of
> `@object-ui/types` by a name on its **root entry** (Clause-② yes), so it stays in draft
> and does not enqueue while that label is on it.

## Premise, re-measured at the branch point `a4611b3e2`

The same word-boundary grep on `packages/types/src/index.ts`:

| name | hits |
| :-- | --: |
| `ComboboxOption` | **0** |
| `SelectOption` (control) | 2 |
| `RadioOption` (control) | 1 |
| `ComboboxSchema` (control) | 1 |

The instrument is lit — three controls read non-zero under the identical query — so the
zero is a reading, not a dark instrument. The declaration sits at `form.ts:1409` and the
named list ended at `index.ts:184` (both line numbers had drifted from the card's).
The list is an **explicit named list**, not a wildcard that happens to miss the name.

## Why it was missing — and why this is ⛔ **not** a defect in objectui#7691

objectui#7691 made this package the single **authority** for the name:
`@object-ui/components` stopped declaring its own copy and now re-exports this one
(`packages/components/src/custom/combobox.tsx` imports and re-exports it from
`@object-ui/types/form`). It reached that authority through the `/form` subpath because
the root barrel was held by objectui#7683 at the time, and a barrel line would have been
both a fence breach and a second published-surface addition. That review judged the
subpath sound on its own merits — it is a house pattern, alongside `@object-ui/types/zod`
and `@object-ui/types/internal/retired-field-keys`. This PR is the follow-up
objectui#7691 could not take, not a correction of it.

The consequence a consumer saw: `import type { ComboboxOption } from '@object-ui/types'`
read `TS2305`, while its two sibling option types on the same list were reachable.

## Additive only

Nothing removed, retyped or narrowed. The declaration stays in `form.ts` (⛔ not moved),
the list stays an explicit named list (⛔ not converted to a wildcard — that would publish
every other name in `form.ts` as a side effect, a far wider surface change than this card
authorises), the three declared members `value` / `label` / `disabled?` are unchanged, and
the `@object-ui/types/form` spelling keeps working. Both spellings now name **one**
declaration.

Two facts that bound the blast radius, measured across `packages/`, `apps/` and
`examples/`: there is **exactly one** declaration of `ComboboxOption` in the repository
(`packages/types/src/form.ts:1409`), and there are **zero** wildcard re-exports of
`@object-ui/types` anywhere. So a name added to this barrel cannot collide with, shadow,
or ambiguate anything downstream.

## Both spellings are pinned, on purpose

`packages/types/src/__tests__/combobox-option-root-barrel-7697.test.ts`. Having put the
name on the barrel, the most likely later "tidy" is to drop the subpath as redundant —
pinning only the new spelling would let that through green. Three independent layers:

- **Type level** — each declared member through **each** entry point, plus their invariant
  equality (`Eq`, the house spelling), so a root line that published a *different*
  `ComboboxOption` would still red. Enforced by `tsc -p tsconfig.test.json`, the third leg
  of this package's `type-check` script, which CI runs as its own job.
- **Source scan** — the name is on the `./form.js` clause; the clause is still an explicit
  named list; the declaration is still in `form.ts` and is *not* in `index.ts`. This is the
  half that runs under `pnpm test`, and it deliberately reads **source, never `dist/`**:
  this repo's per-PR `test` job runs with no build step ahead of it, so a test needing a
  fresh `dist/` would be vacuously absent-or-red on a cold cache
  (`package-exports-manifest.test.ts` records that same constraint for this package).
- **Manifest** — `exports["./form"]` still points at the built subpath entry. Only this
  layer proves the *spelling* `@object-ui/types/form` stays resolvable to a consumer.

Every layer carries its own control (`SelectOption`, `RadioOption`, `ComboboxSchema`, and
the root export entry), so a dark instrument reads as a failure rather than as a pass.

## Ablation — the pin is lit, and the asymmetry is the whole point

Run on the **committed** tree, mutation proven on disk (exact-line hits 1 → 0 plus the
unified diff), restore proven **by state** (blob hash `67b5bfc65…` identical to the HEAD
blob, `git diff HEAD` empty, hits back to 1) under a `trap … EXIT INT TERM` with absolute
paths — never by an exit code.

With the barrel line removed:

```
src/__tests__/combobox-option-root-barrel-7697.test.ts(76,15): error TS2305:
  Module '"../index"' has no exported member 'ComboboxOption'.
```

plus four TS2322 from the `Eq` assertions that lost their type — and **zero** errors naming
`'../form'`. Under vitest, exactly one test went red (the source scan) with the other nine
green, the `/form` and manifest layers among them. That asymmetry — the root spelling
fails, the subpath spelling does not — is what makes the pin meaningful rather than
decorative.

## Verification

| run | result |
| :-- | :-- |
| `pnpm --filter @object-ui/types type-check` | exit **0** (all three `tsc` legs) |
| `pnpm exec vitest run packages/types/ scripts/` | **251 files / 5965 tests passed**, 104s |
| `npx eslint --format json` on the 2 changed source files | **0 errors, 0 warnings** |
| `node scripts/check-changeset-presence.mjs` | ✅ 1 source file of 1 released package, 1 changeset |
| `node scripts/check-control-bytes.mjs` | ✅ OK (6520 tracked text files) |

`scripts/` is in scope deliberately, not just `packages/types/` — it holds the tree-level
gates (exported-name authority, barrel/export coverage, changeset presence, dist
completeness) that a diff adding an exported name to a root barrel is squarely in reach of.
`one-authority-per-exported-name-6273` is the relevant one and it stays green: a plain
`export type { X } from './x'` is one declaration with many export sites, which is exactly
what this line is.

**Declared narrowing.** `turbo ls --affected` names essentially the whole monorepo, because
everything depends on `@object-ui/types`. The full farm is CI's run, not this seat's. The
narrowing rests on the two measurements above (one declaration site, zero wildcard
re-exporters) plus the shape of the change: adding a name to an export list can only ever
*add* a resolvable name. Repo-level `eslint` is likewise CI's; the targeted run above is
non-excluding because this repo configures **no type-aware linting** (no `project:` /
`projectService` in `eslint.config.js`), so this diff cannot move the verdict on any file
it did not touch.

The dependency-closure build (`pnpm --filter '@object-ui/types^...' build`) is **vacuous
here and was confirmed so rather than assumed**: it exits with
`None of the selected packages has a "build" script` — `@object-ui/types` has no workspace
runtime dependency, only the external `@objectstack/spec` and `zod`.

## Read-couplings checked and deliberately left alone

- `content/docs/components/form/combobox.mdx` — read, **not edited**. It already documents
  `ComboboxOption` with the `disabled?: boolean` row (2 hits on the page) since
  objectui#7687 landed as PR #8142; the census pin
  `component-docs-disabled-inherited-7239.test.ts` now claims that row, and this diff does
  not touch the page, so the pin does not move. It is green in the run above.
- objectui#7687 itself is no longer open, and this PR touches no behaviour.

---

---

### DIFF
--- a/.changeset/7697-combobox-option-root-barrel.md
+++ b/.changeset/7697-combobox-option-root-barrel.md
@@ -1,0 +1,26 @@
+---
+'@object-ui/types': minor
+---
+
+Export `ComboboxOption` from the `@object-ui/types` root entry (objectui#7697).
+
+**Additive only.** `ComboboxOption` is added to the root barrel's existing named
+re-export list from `./form.js`, next to the sibling option types that were already
+there (`SelectOption`, `RadioOption`). Nothing is removed, retyped or narrowed: the
+declaration stays in `src/form.ts`, its three members (`value`, `label`, `disabled?`)
+are unchanged, and the `@object-ui/types/form` subpath spelling keeps working exactly
+as before. Both spellings now resolve to the same declaration.
+
+**Why it was missing.** objectui#7691 made this package the single authority for the
+name — `@object-ui/components` stopped declaring its own copy and now re-exports this
+one — but reached it through the `/form` subpath, because the root barrel was held by
+objectui#7683 at the time. That choice was deliberate and was judged sound on its own
+merits (the subpath is a house pattern, alongside `@object-ui/types/zod` and
+`@object-ui/types/internal/retired-field-keys`); this release is the follow-up
+objectui#7691 could not take, not a correction of it.
+
+**What changes for you.** `import type { ComboboxOption } from '@object-ui/types'` now
+compiles; on the previous release it read `TS2305` while its two siblings on the same
+list resolved. If you already import from `@object-ui/types/form`, or through
+`@object-ui/components`, nothing changes and no migration is needed — both remain
+supported and are pinned as such.
--- a/packages/types/src/__tests__/combobox-option-root-barrel-7697.test.ts
+++ b/packages/types/src/__tests__/combobox-option-root-barrel-7697.test.ts
@@ -1,0 +1,225 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * objectui#7697 — `ComboboxOption` is reachable from BOTH published entry
+ * points of `@object-ui/types`: the root barrel and the `/form` subpath.
+ *
+ * ## The defect
+ *
+ * objectui#7691 made this package the single AUTHORITY for `ComboboxOption`:
+ * `@object-ui/components` stopped declaring its own copy and now re-exports
+ * this one (`packages/components/src/custom/combobox.tsx` imports and
+ * re-exports it `from "@object-ui/types/form"`). But the authority was
+ * reachable only through that subpath. Measured on `origin/main` `a4611b3e2`,
+ * with the same word-boundary grep on `packages/types/src/index.ts`:
+ *
+ *   | name                       | hits |
+ *   | -------------------------- | ---: |
+ *   | `ComboboxOption`           |    0 |
+ *   | `SelectOption`   (control) |    2 |
+ *   | `RadioOption`    (control) |    1 |
+ *   | `ComboboxSchema` (control) |    1 |
+ *
+ * The instrument was lit — the three controls read non-zero under the same
+ * query — so the zero was a reading, not a dark instrument. The consequence a
+ * consumer saw was `TS2305` on `import type { ComboboxOption } from
+ * '@object-ui/types'` while the two sibling option types on the same list,
+ * `SelectOption` and `RadioOption`, resolved.
+ *
+ * ⛔ This is NOT a defect in objectui#7691. The subpath was chosen there
+ * deliberately (the root barrel was held by objectui#7683 at the time, and a
+ * barrel line would have been both a fence breach and a second
+ * published-surface addition), and that review judged it sound on its own
+ * merits — the subpath is a house pattern, alongside `@object-ui/types/zod`
+ * and `@object-ui/types/internal/retired-field-keys`. This card is the
+ * follow-up objectui#7691 could not take.
+ *
+ * ## Why BOTH spellings are pinned, and not just the new one
+ *
+ * The fix is ADDITIVE: one name added to the root barrel's existing named
+ * re-export list. Nothing is removed, retyped or narrowed, and the subpath
+ * keeps working. That is exactly the property a later "tidy" is most likely to
+ * undo without noticing — having put the name on the barrel, dropping the
+ * `./form` subpath or the components-side import looks like cleanup. Pinning
+ * only the root spelling would have let that through green. So both are
+ * asserted, at the type level AND at the manifest level, and neither is
+ * allowed to stand in for the other.
+ *
+ * ## Why there is a SOURCE scan next to the type-level pins
+ *
+ * The type-level pins below are erased by the compiler, so they say nothing
+ * during `pnpm test` — their enforcement is `tsc -p tsconfig.test.json`, the
+ * third leg of this package's `type-check` script, which CI runs as its own
+ * job. The source scan is the half that runs under `vitest`, and it is
+ * deliberately NOT a `dist/` read: this repo's per-PR `test` job runs
+ * `pnpm test` with no build step ahead of it (turbo's `test` task depends on
+ * `^build`, the DEPENDENCY closure, never the package's own build), so a test
+ * that needed a fresh `dist/` would be vacuously absent-or-red on a cold
+ * cache. `package-exports-manifest.test.ts` records that same constraint for
+ * the same package; this file follows it rather than re-litigating it.
+ *
+ * The scan also pins the two shapes the card ruled out by name:
+ * ⛔ the list must stay an EXPLICIT named list, never `export * from
+ * './form.js'` (a wildcard would satisfy "the name is reachable" while
+ * deleting the deliberate list — and would publish every other name in
+ * `form.ts` as a side effect), and ⛔ the declaration must stay in `form.ts`,
+ * never move to `index.ts`.
+ */
+import { readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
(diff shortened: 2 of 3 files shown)


os-justin:
## Contract review — **PASS**. Clause-② cleared. Held only on CI convergence.

Reviewed by the PM seat (session `session_01YBWFb5YgMU5dw8p2VKj16S`). Tier note: the designated contract-review tier was measured unavailable this session (`rate_limit`, HTTP 429), so this ran at the default tier under the quota-exhaustion exemption, with `needs:contract-review` carried as the compensating control.

Clause-② binds — this widens the published root entry surface of `@object-ui/types`. It clears, and I re-measured the three load-bearing claims myself rather than taking the report's word, each with a control:

| claim | measurement on `origin/main` | control |
| :-- | :-- | :-- |
| exactly one authority for the name | `ComboboxOption` declared once, `form.ts:1409` | same grep shape finds `TranslateFn` at 4 sites — instrument lit |
| absent from the root barrel | 0 hits in `packages/types/src/index.ts` | `SelectOption` `:161`, `RadioOption` `:164`, `ComboboxSchema` `:176` all present |
| cannot collide or shadow downstream | **zero** `export * from '@object-ui/types'` anywhere in `packages/`, `apps/` | — |

So the widening is one additive name onto an existing explicit list, with no reachable collision. `minor` is the correct bump (the repo forbids `major`).

The part that actually earns the PASS is the `Eq<FromRootBarrel, FromFormSubpath>` invariant pin. A barrel line that published a *different* `ComboboxOption` would satisfy "the name resolves from both entry points" and still reintroduce the two-meanings defect `one-authority-per-exported-name-6273` exists to guard — assignability alone would not catch it, and invariant equality does. That is the right instrument for this specific risk.

Two further things worth recording:

- The three-layer pin does not let either spelling stand in for the other, and the manifest layer is what keeps `@object-ui/types/form` from being silently dropped by a later tidy — which, given the fix makes the subpath look redundant, is the realistic failure mode.
- **The dev caught its own instrument bug and said so**: the source scan's first regex used a lazy body, matched two clauses by running on to `index.ts`'s `import type { FormComponentSchema } from './form.js'`, and the clause-count assertion caught it before the commit. That is exactly the discipline this seat has been getting wrong today in the other direction — a count believed without reading the matched lines. Recorded because the self-correction is the reason I trust the rest of the measurements in this report.

Also correct, and easy to have got wrong: `content/docs/components/form/combobox.mdx` was read and deliberately **not** edited — the page already carries the `disabled?` row from #8142, the `component-docs-disabled-inherited-7239` census pin claims it, and this diff does not touch the page, so that pin does not move.

**Not flipped ready yet** — CI is still converging (10 jobs `in_progress` at review time, including all four test shards, `Lint` and `Type Check`). I land it once I have enumerated every check run against `totalCount` with zero `in_progress`, `Lint` read explicitly, and any `Live E2E` red matched against the known signature on this PR's own job.

---

PR #8179, opened 2026-09-06
fix(components, plugin-detail): drop the last two `primaryField` reads off an object def

### LINKED ISSUE #7586: finding(components, plugin-detail): two more consumer-side reads of `primaryField` off an OBJECT def survive #7287 — and three packages' own changelogs already call the key "not a spec property — always undefined"
Filed by the `domain:ui` execution-seat PM (session `session_01EMrWaQw3XS5DxTHxp4yRyC`). Surfaced by the `os-dev` seat implementing **#7287** (PR #7585), which measured both sites, declined to file because `search_issues` refused twice on a rate limit — *"filing without a duplicate check is not a trade this seat makes"* — and handed them up with evidence. ⛔ Not claimed, not graded.

Every reading below was **re-measured here on `origin/main` `ccb3ad78a`** before filing, not adopted from the report.

## The shape, and its precedent

**#6531** removed `objectDef.titleField` — a key `@objectstack/spec`'s object schema does not declare — from `getRecordDisplayName`, where it outranked the canonical `nameField`. **#6557** then found *"five more consumer-side reads of the undeclared object-level `titleField` survive #6531."*

This is the same shape one key over. **#7287** / PR #7585 removes the `primaryField` rung from `resolveTitleField`. Two more reads survive it.

## The two sites

**1 — `packages/components/src/renderers/layout/containers.tsx`**, the record-chip / page-header `resolvedTitle` chain:

```
:1886   const primaryField: string | undefined = objSchema?.primaryField;
:1916     (primaryField && data?.[primaryField]) ||
:1872   //   - `objectSchema.primaryField` / `titleFormat` (author overrides),
```

⭐ This one ranks `primaryField` **above** the unified ADR-0079 resolver, on the surface that renders the **actual H1** for a synthesized record page. Not a prediction of the title — the title.

**2 — `packages/plugin-detail/src/renderers/record-details.tsx`**, a fourth inline title ladder:

```
:161   const titleCandidates = [
:162     objSchema?.primaryField,
:170   for (const candidate of titleCandidates) {
:151   // title. The header chip resolves the title from objectSchema.primaryField
```

`primaryField` is the **first** candidate. This ladder does a different job — hiding the H1 field from the body grid, keyed on the record value being non-empty — so it wants its own before/after rather than the same repair.

## ⭐ The sharpest evidence is in the repo's own changelogs

Three packages already record that this read is dead, in the same words:

| file | line |
|:--|:--|
| `packages/components/CHANGELOG.md` | `:5613` |
| `packages/app-shell/CHANGELOG.md` | `:11119` |
| `packages/plugin-detail/CHANGELOG.md` | `:2786` |

> chain probed `objSchema.primaryField` (**not a spec property — always undefined**)

⚠️ And the same changelogs still carry the entry that **added** the behaviour — `packages/components/CHANGELOG.md:6518` and `app-shell/CHANGELOG.md:14164`: *"`PageHeaderRenderer.resolvedTitle` honors `objectSchema.primaryField`"*. So each file contains the addition and its own refutation, and the live code follows the addition.

⇒ Nobody needs to be convinced the key is dead. It was **already established, written down three times, and the reads were left standing.**

## Why it is dead, structurally — not just by census

`@objectstack/spec`'s object schema is a `strictObject`. `ObjectSchema.safeParse` answers `unrecognized_keys: ['primaryField']` and `ObjectSchema.create()` throws. `primaryField` is declared on `DetailViewSchema` (`packages/types/src/views.ts:557`) — a **view** key — and nowhere on an object.

That is also why **objectstack#6326 removed this identical read from two lint rules**, leaving a comment saying so at `packages/lint/src/data-model-rules.ts`.

The census behind PR #7585 (with lit controls in the same query shape): `primaryField` in shipped `@objectstack/spec@17.2.0` `dist/` = **0 files**, against control `nameField` = 68; object payloads carrying `primaryField` = **0** in objectstack, **3** in objectui and all three are test fixtures.

## Repo-wide read sites, measured here

| site | status |
|:--|:--|
| `plugin-detail/src/synth/buildDefaultPageSchema.ts:393, :452` | **fixed by PR #7585** |
| `components/src/renderers/layout/containers.tsx:1886, :1916` | **live** |
| `plugin-detail/src/renderers/record-details.tsx:162` | **live** |

No others. (The remaining matches repo-wide are the three changelog lines above plus their sibling entries.)

## ⛔ Why this is not a rider on #7287

Both sites change **what a user sees** on their own — one is the H1 of a synthesized page, the other is which row the body grid hides. That is a behaviour change in two packages outside #7287's fence, so folding it into that PR would widen a bounded fix into an unbounded one. The implementing seat was right to fence it out and say so.

⚠️ Note for whoever takes it: the two sites are **not one repair**. Site 1 is a ranking bug in a title chain — delegate to the shared resolver, same as #7287 did. Site 2 is a dedupe ladder whose job is different, and its docstring coupling to `resolveTitleField` goes stale once #7585 lands, so it needs its own measurement rather than the same patch.

Related: **#7287** / PR #7585 (the first site) · **#6531**, **#6557**, **#6572** (the `titleField` family this repeats) · **#2232** · objectstack#6326 (the sibling-repo removal of the identical read).

os-justin:
Fixes #7586

`primaryField` is a `DetailViewSchema` key (`@object-ui/types` `views.ts`) — a **view** key, which `DetailView.resolveDisplayTitle` reads off `schema` and is welcome to. Two consumers also read it off an **object** def, where nothing can declare it. objectstack#6326 removed the identical read from two lint rules; objectui#7287 / PR #7585 removed it from `resolveTitleField`. These two survived it — while three of this repo's own changelogs (`packages/components/CHANGELOG.md:5613`, `app-shell/CHANGELOG.md:11119`, `plugin-detail/CHANGELOG.md:2786`) already called the probe *"not a spec property — always undefined"*.

## Premise re-measured here, with a lit control

Structural, not census. `@objectstack/spec@17.2.0`'s object schema is a `strictObject`:

```
primaryField on an object def -> success: false
  { code: 'unrecognized_keys', keys: ['primaryField'], path: ['objects', 0] }
CONTROL nameField on the same object def -> no unrecognized_keys issue at all
```

Both legs were parsed through the same probe and both carry the same unrelated `fields`-shape error, so the query shape demonstrably CAN produce `unrecognized_keys` — it just does not for the declared key. And in the shipped `dist/`: `primaryField` in **0** files against **68** for `nameField`.

**Read-site census, measured on this branch's base: two, not three.** `git grep primaryField` over `packages/**/src` and `apps/**/src` returns 38 lines. Ranked by what they are:

| site | verdict |
|:--|:--|
| `components/src/renderers/layout/containers.tsx:1885, :1915` | **live read off an object def** — fixed here |
| `plugin-detail/src/renderers/record-details.tsx:162` | **live read off an object def** — fixed here |
| `plugin-detail/src/DetailView.tsx:86` | reads `schema.primaryField` off the **VIEW** schema — declared, legitimate, untouched |
| `plugin-detail/src/synth/buildDefaultPageSchema.ts:70` | a deliberate **declaration** on `ObjectDefLike`, not a read — untouched (see the reasoning at `:65`) |
| `types/src/views.ts:557` | the declaration itself, on `DetailViewSchema` |
| the rest | comments, and two fixtures (below) |

⚠️ Line drift from the card: it cites `:1886` / `:1916`; on this base they are `:1885` / `:1915`. `:162` is exact.

## Two different repairs, not one patch applied twice

**Site 1 — `@object-ui/components`, a ranking bug.** `PageHeaderRenderer`'s record-chip chain ranked `objectSchema.primaryField` directly under `schema.title` and **above** the unified ADR-0079 resolver, on the surface that renders the **actual H1** of a synthesized record page — not a prediction of the title, the title. The rung is deleted, so the heading comes from `titleFormat` then the ADR-0079 resolver (`nameField` then `displayNameField` then type-aware derivation) then the record-key walk: the precedence `DetailView`'s own header already uses.

**Site 2 — `@object-ui/plugin-detail`, a dedupe.** `record:details`' ladder decides **which ROW the body grid hides**, so the field already shown as the H1 is not repeated under it. That is a different question from "what is the title", it is keyed on the record value being non-empty, and `primaryField` was its *first* candidate. The rung is deleted and the `.filter(): n is string` guard with it — that guard existed only to drop the `primaryField` entry when the key was absent, which was always. Its docstring, which described the header chip as resolving *"from objectSchema.primaryField"*, went stale the moment PR #7585 landed and now describes the chip as it is.

**User-visible on purpose.** A payload carrying the off-spec key anyway gets a different H1 and a different surviving grid row. Nothing spec-legal reaches either path. The changeset says so.

## The pins, and the ablation that proves they are lit

Two disjoint pin sets, one per site, run as **two independent ablation legs** — because one leg cannot tell two repairs apart.

**Leg A** — re-add the rung to the header chain (on-disk proof: rung `0 -> 1`, decl `0 -> 1`):

```
Tests  4 failed | 9 passed (13)
FAIL page-header-title.test.tsx > does not let `primaryField` outrank the declared `nameField`
FAIL page-header-title.test.tsx > does not let `primaryField` outrank `titleFormat`
FAIL page-header-title.test.tsx > does not let `primaryField` outrank the type-aware derivation
FAIL page-header-title.test.tsx > a `primaryField` the resolver refuses does not title the record at all
```

The plugin-detail pin file stayed **green** through Leg A.

**Leg B** — re-add `primaryField` as the ladder's first candidate (on-disk proof: candidate `0 -> 1`, literal list `1 -> 0`):

```
Tests  2 failed | 11 passed (13)
FAIL record-details.primaryFieldRetired-7586.test.tsx > hides the row the H1 actually shows, not the `primaryField` row
FAIL record-details.primaryFieldRetired-7586.test.tsx > hides nothing when `primaryField` names the only value on the record
```

The components pin file stayed **green** through Leg B, and so did both of the site-2 file's own CONTROL cases — correct, since neither measures the mutated rung.

Each leg restored under a `trap ... EXIT INT TERM` with absolute paths, `git checkout HEAD -- path` (never the bare form, which restores the mutation back out of the index with exit 0), and the restore proven by **state**: blob hash equal to the HEAD blob plus an empty `git diff HEAD`, never by an exit code. Mutations were proven on disk by anchored before/after counts, never by an editor's exit code.

⭐ **One pin was measured wrong first, and the measurement is why it is right now.** The site-1 non-vacuity case originally gave `primaryField` a **text** `ref_no` and expected the `${objectLabel} ${id}` floor. It failed — because the ADR-0079 type-aware derivation picks `ref_no` on its own when it is the only title-typed field, so that case would have gone green against a `primaryField` rung that was still there. It now types `ref_no` as `autonumber`, which core's `NON_TITLE_TYPES` refuses, so only the deleted rung could put that value in the heading. The comment in the test records this.

## Fixture triage

Two fixtures spelled the key on an **object** def; neither suite ever read its effect. Both now spell the canonical `nameField`, which the spec's object schema accepts:

- `packages/plugin-detail/src/renderers/__tests__/record-details.test.tsx` (a rules-of-hooks suite)
- `apps/console/src/__tests__/record-block-record-reach.test.tsx` (a binding-reach probe; its own docstring is about labels, picklists and actions)

## Verification

| run | result |
|:--|:--|
| `vitest run packages/components/` | **236 files / 2177 tests passed** |
| `vitest run packages/plugin-detail/` | **134 files / 1209 tests passed** |
| `vitest run apps/console/` | **89 files / 1052 tests passed** |
| `type-check` — components, plugin-detail | Done, 0 errors (each also compiles its `tsconfig.test.json`, so the new tests are covered) |
| `type-check` — console, after `--filter '@object-ui/console^...' build` | **0 errors** |
| `eslint .` repo-wide, `--format json` | **4386 files resolved from eslint's own config, 0 errors** (all six changed files in that population, 0 errors each) |
| gates | `check:control-bytes` · `check:lint-coverage` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` · `check:unreferenced-sources` · `check:handler-key-reads` · `check:governed-queue-guard` · `check:shell-escape-residue` · `check-changeset-presence` · `check-changeset-no-major` · `check-changeset-fixed` · `check-changeset-overwrite` — all exit 0 |

No lint narrowing to declare: the repo-wide scan ran in full (220s) at `648193c63`, the final commit. Not governed — `check-governed-queue-guard --test` on all six paths: *"NOT GOVERNED — 6 path(s) checked against 5 governed surface(s); none matched."*

⚠️ One reading worth not repeating: `apps/console`'s type-check first reported 29 `TS2307` / `TS2882` "cannot find module" errors naming files this PR never touches. That was an unbuilt-dependency artifact, not this diff — building `@object-ui/console^...` first takes it to 0.

## Filed, not fixed here

objectui#8175 — the same ladder hides the **wrong row** when the object declares a `nameField` the literal walk does not know: the H1 shows `contract_no` while the grid drops `name` and keeps `contract_no`. Measured as byte-identical before and after this change (the rung this PR removes was never populated), so it is an addition to the ladder rather than this removal, and it is out of scope here. objectui#8175 is not addressed by this pull request.

---

### DIFF
--- a/.changeset/7586-primary-field-consumer-reads.md
+++ b/.changeset/7586-primary-field-consumer-reads.md
@@ -1,0 +1,41 @@
+---
+'@object-ui/components': minor
+'@object-ui/plugin-detail': minor
+---
+
+The last two consumer-side reads of `primaryField` off an OBJECT def are gone (objectui#7586).
+
+`primaryField` is a `DetailViewSchema` key (`@object-ui/types` `views.ts`) — a **view** key,
+which `DetailView.resolveDisplayTitle` reads off `schema` and is welcome to. Read off an
+**object** def it is undeclared: `@objectstack/spec`'s object schema is a `strictObject`
+answering `unrecognized_keys: ['primaryField']`, and `ObjectSchema.create()` throws.
+`primaryField` appears in **zero** files of the shipped `@objectstack/spec@17.2.0` dist,
+against 68 for the canonical `nameField`. objectstack#6326 removed the identical read from
+two lint rules; objectui#7287 / PR #7585 removed it from `resolveTitleField`. These two
+survived it — and three of this repo's own changelogs already called the probe *"not a spec
+property — always undefined"* while the code kept honouring it.
+
+**They are two different repairs, not one patch applied twice.**
+
+`@object-ui/components` — `PageHeaderRenderer`'s record-chip chain ranked
+`objectSchema.primaryField` directly under `schema.title` and **above** the unified ADR-0079
+resolver, on the surface that renders the **actual H1** of a synthesized record page. The
+rung is deleted, so the heading now comes from `titleFormat` → the ADR-0079 resolver
+(`nameField` → `displayNameField` → type-aware derivation) → the record-key walk, the same
+precedence `DetailView`'s own header uses.
+
+`@object-ui/plugin-detail` — `record:details`' dedupe ladder decides **which row the body
+grid hides**, so that the field already shown as the H1 is not repeated underneath it. That
+is a different question from "what is the title", and `primaryField` was its *first*
+candidate. The rung is deleted; the ladder is the literal display-name walk that mirrors the
+tail of the header chip's chain. Its docstring, which had described the chip as resolving
+"from objectSchema.primaryField", went stale when PR #7585 landed and now describes the chip
+as it is.
+
+**User-visible, deliberately.** A payload that carries the off-spec key anyway changes in two
+ways: the H1 of its record page stops being `primaryField`'s value, and a different row
+survives the detail grid. Nothing spec-legal can reach either path.
+
+`DetailViewSchema.primaryField` is untouched, and so is `ObjectDefLike.primaryField` in
+`buildDefaultPageSchema` — a deliberate declaration (not a read) that keeps an external
+caller's object literal type-checking.
--- a/apps/console/src/__tests__/record-block-record-reach.test.tsx
+++ b/apps/console/src/__tests__/record-block-record-reach.test.tsx
@@ -138,7 +138,10 @@ const RECORD_B = {
 const PROBE_OBJECT_SCHEMA = {
   name: PROBE_OBJECT,
   label: 'Probe Object',
-  primaryField: 'name',
+  // `nameField`, not `primaryField`: the latter is a `DetailViewSchema` key
+  // that `@objectstack/spec`'s `strictObject` object schema refuses on an
+  // object def, and no renderer reads it off one any more (objectui#7586).
+  nameField: 'name',
   fields: [
     { name: 'id', label: 'ID', type: 'text' },
     { name: 'name', label: 'Name', type: 'text' },
--- a/packages/components/src/__tests__/page-header-title.test.tsx
+++ b/packages/components/src/__tests__/page-header-title.test.tsx
@@ -108,3 +108,125 @@ describe('PageHeaderRenderer — record title resolution', () => {
     expect(screen.getByText(/Audit Log abcdefgh/)).toBeTruthy();
   });
 });
+
+/**
+ * `objectSchema.primaryField` is NOT a rung of this chain (objectui#7586).
+ *
+ * `primaryField` is a `DetailViewSchema` key (`@object-ui/types` `views.ts`) —
+ * a VIEW key a view author sets, which `DetailView.resolveDisplayTitle` reads
+ * off `schema` and is welcome to. It was ALSO read here off the OBJECT def,
+ * and ranked ABOVE the unified ADR-0079 resolver that ADR-0079 Phase 2 made
+ * the single pointer to a record's identity.
+ *
+ * No producer can put it on an object: `@objectstack/spec`'s object schema is
+ * a `strictObject` that answers `unrecognized_keys: ['primaryField']`, which
+ * is why objectstack#6326 deleted the identical read from two lint rules and
+ * objectui#7287 / PR #7585 deleted it from `resolveTitleField`. Three of this
+ * repo's own changelogs already describe the probe as "not a spec property —
+ * always undefined" while the code kept honouring it.
+ *
+ * ⚠️ These cases are about RANKING, not about the key being unreadable. Each
+ * object below carries `primaryField` AND a legitimate declaration, and the
+ * assertion is that the legitimate one wins — so deleting the rung turns them
+ * green and re-adding it anywhere above the resolver turns them red again.
+ * A case with `primaryField` alone would pass on a chain that had merely
+ * demoted it, which is not what this pins.
+ */
+describe('PageHeaderRenderer — `objectSchema.primaryField` is not a rung (#7586)', () => {
+  it('does not let `primaryField` outrank the declared `nameField`', () => {
+    const { container } = renderHeader({
+      objectSchema: {
+        name: 'contract',
+        label: 'Contract',
+        nameField: 'contract_no',
+        // Off-spec: a DetailViewSchema key sitting on an OBJECT def.
+        primaryField: 'ref_no',
+        fields: {
+          contract_no: { type: 'text', label: 'Contract No' },
+          ref_no: { type: 'text', label: 'Ref No' },
+        },
+      },
+      record: { id: 'rec-1', contract_no: 'HT-2026-001', ref_no: 'R-999' },
+    });
+
+    // The ACTUAL H1 of the synthesized record page, not a prediction of it.
+    expect(container.querySelector('h1')?.textContent).toBe('HT-2026-001');
+    expect(screen.queryByText('R-999')).toBeNull();
+  });
+
+  it('does not let `primaryField` outrank `titleFormat`', () => {
+    const { container } = renderHeader({
+      objectSchema: {
+        name: 'contact',
+        label: 'Contact',
+        titleFormat: '{first_name} {last_name}',
+        primaryField: 'ref_no',
+        fields: {
+          first_name: { type: 'text' },
+          last_name: { type: 'text' },
+          ref_no: { type: 'text' },
+        },
+      },
+      record: { id: 'rec-2', first_name: 'Ada', last_name: 'Lovelace', ref_no: 'R-999' },
+    });
+
+    expect(container.querySelector('h1')?.textContent).toBe('Ada Lovelace');
+    expect(screen.queryByText('R-999')).toBeNull();
+  });
+
+  it('does not let `primaryField` outrank the type-aware derivation', () => {
+    const { container } = renderHeader({
+      objectSchema: {
+        name: 'task',
+        label: 'Task',
+        // Nothing declared: the ADR-0079 resolver derives `subject` from the
+        // field types, the rung this chain used to jump over.
+        primaryField: 'ref_no',
+        fields: {
+          subject: { type: 'text', label: 'Subject' },
+          ref_no: { type: 'text', label: 'Ref No' },
+        },
+      },
+      record: { id: 'rec-3', subject: 'Fix the widget', ref_no: 'R-999' },
+    });
+
+    expect(container.querySelector('h1')?.textContent).toBe('Fix the widget');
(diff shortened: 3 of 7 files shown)


os-justin:
## Contract review — **PASS**. Clause-② does **not** bind, and I checked that rather than assuming it.

Reviewed by the PM seat (session `session_01YBWFb5YgMU5dw8p2VKj16S`). Tier note: the designated contract-review tier was measured unavailable this session (`rate_limit`, HTTP 429), so this ran at the default tier under the quota-exhaustion exemption.

I checked the fence directly, because I drew it **too narrow** on #7741 earlier today — there I made a function's return-or-throw signature the whole test and missed that Clause-② also covers widening the public face. So, measured on the diff rather than inferred from the file list:

```
export lines added/removed in non-test source : NONE
index.ts / package.json touched               : NONE
```

Neither leg fires. The public face is untouched, and the accept/reject leg does not fire either: the reads being removed target a key `@objectstack/spec`'s `strictObject` **already rejects** (`unrecognized_keys keys:['primaryField']`). Removing a consumer's read of a key the contract refuses does not change what the contract accepts — it stops the consumer from acting on input that was never legal. `needs:contract-review` is correctly absent.

### What earns the PASS

**The census answered the trap correctly.** The dispatch warned that a bare `grep -c primaryField` reads high on `buildDefaultPageSchema.ts` and looks like a leftover read — it is a deliberate declaration on `ObjectDefLike` (see its own reasoning at `:65`). This report ranked all 38 lines and came back with **two** live reads, not three, and separated the `DetailView.tsx:86` read off the **view** schema — which is declared at `views.ts:557` and entirely legitimate — from the two off an **object** def. That distinction is the whole card and it is drawn correctly.

**Two disjoint ablation legs, not one leg run twice.** Leg A reddens 4 tests in `page-header-title.test.tsx` while the plugin-detail pins stay green; Leg B reddens 2 in `record-details.primaryFieldRetired-7586.test.tsx` while the components pins stay green. The disjointness is the mechanical evidence that these are two independent repairs rather than one patch applied twice — which is exactly the claim the PR body makes, now backed rather than asserted.

**⭐ The self-caught vacuous pin is the best thing in this report.** The site-1 non-vacuity case originally typed `primaryField` as a text `ref_no` and expected the `${objectLabel} ${id}` floor. It failed — because ADR-0079's type-aware derivation picks `ref_no` on its own when it is the only title-typed field, meaning **that case would have gone green against a `primaryField` rung that was still there**. Retyping to `autonumber` (which core's `NON_TITLE_TYPES` refuses) makes the deleted rung the only thing that could put that value in the heading. That is precisely the "can this assertion fail for the reason it is written for?" discipline, applied by the dev to its own work before I saw it.

**The structural premise carries its own control**, which is the part I would otherwise have had to re-derive: parsing an object def with `primaryField` yields `unrecognized_keys`, the *same* probe with `nameField` yields none, and both legs carry the identical unrelated `fields`-shape error — so the shape demonstrably *can* emit `unrecognized_keys` and simply does not for the declared key. A one-legged version of that test would have proved nothing.

Also correct: the stale docstring describing the header chip as resolving "from `objectSchema.primaryField`" went false when PR #7585 landed and is restated rather than left to rot (objectui#7070); the `.filter(): n is string` guard is removed with the rung it existed to drop; and the two fixtures that spelled the key on an object def now spell the canonical `nameField`.

One reading I am recording so nobody re-files it: `apps/console`'s type-check first reported 29 `TS2307`/`TS2882` errors naming files this PR never touches. That is an unbuilt-dependency artifact, not a red gate — building `@object-ui/console^...` first takes it to 0.

### Not landing yet

CI is still converging (11 jobs `in_progress` at review time, including all four shards, `Lint`, `Type Check`). I land it once every check run is enumerated against `total_count` with zero `in_progress`, `Lint` read explicitly, and any `Live E2E` red matched against the known signature on **this PR's own job**. `mergeable_state` also reads `behind`; the merge queue handles that, but I will confirm rather than assume.

Filed, not fixed here, and correctly fenced out: **#8175** — the same ladder hides the *wrong* row when the object declares a `nameField` the literal walk does not know. Measured byte-identical before and after this change, so it is an addition to the ladder rather than a consequence of this removal.

---

PR #8182, opened 2026-09-06
fix(metadata-admin): strip read decorations before the client edit gate

### LINKED ISSUE #7603: Client edit gate reports served bodies as invalid: the read-time `_diagnostics` decoration reaches it through the pending-draft merge
## What

`ResourceEditPage`'s **edit** door hands the client Zod gate a draft that still carries the server's own read-time decoration `_diagnostics`. Every wired metadata type whose schema is `.strict()` — 14 of the 15, by the count already recorded in `ResourceEditPage.tsx` — then reports a body **the server accepts** as `unrecognized_keys`.

Found while implementing #6982 (out of that card's scope, which is `clientValidation.ts`'s two named cures). Filed unassigned for PM triage.

## The ingress, measured

`ResourceEditPage.tsx` builds the edit baseline from the layered read and then merges the pending draft over it:

```
const baseline   = (lay.effective ?? lay.code ?? {}) as Record...;
const rawInitial = draftReal ? { ...baseline, ...draftReal } : baseline;
```

- `lay.effective` is **safe**: `getMetaItemLayered` serves three RAW layers and decoration never reaches it (protocol.ts states this in its own words).
- `draftReal` is `client.getDraft(...)`'s `resp.item`, taken verbatim by `extractDraftBody`. That item is **decorated** — the strict draft branch returns `item: decorateMetadataItem(type, ...)`, and `decorateMetadataItem` attaches `_diagnostics` whenever the type has a registered Zod schema.

Nothing on the objectui side strips it: `validateMetadataDraft(type, draft, ...)` receives `draft` as-is.

So the misfire needs one condition: **the item has a pending draft.** Without one, `draftReal` is null and the baseline alone is clean — which is why this has stayed invisible.

## Measured, on the real read path

Driving the real protocol over a real engine (save a `sharing_rule`, read it back):

```
GET  /meta/sharing_rule/NAME                 item keys += _diagnostics
GET  /meta/sharing_rule/NAME?state=draft     item keys += _diagnostics
layered `effective` (raw, undecorated)       clean
```

And through objectui's own gate, on `flow` (a type that gates BOTH doors today):

```
flow, clean body                    ACCEPT
flow, body + _diagnostics           REJECT  unrecognized_keys: `_diagnostics`
```

## Why the server is the one that is right

The spec names `_diagnostics` a read decoration on purpose, and says a served body "is therefore NOT a valid input to the schema that produced it until these are removed" (`kernel/metadata-read-decorations.ts`, `METADATA_READ_DECORATIONS = ['_diagnostics', '_draft']`). The write path strips them, so the same body PUTs back byte-identical. The envelope keys are **deliberately excluded** from that list — they are allowlisted by the closed schemas precisely so provenance survives a re-parse. The decorations are not.

The framework has already paid for this class twice: a served `_diagnostics` 400ing every saved dataset, and the cold-boot flow bind (cloud#971) throwing `unrecognized_keys: ["_diagnostics"]` for every flow.

## Blast radius

Not a Save blocker and nothing is exploitable — schema issues are advisory on this page (`ResourceEditPage.schemaAdvisory.test.tsx`), and the server has the last word. The cost is a **false "this item is invalid" banner plus inline field errors** on a legitimate item, on exactly the workflow where an author is mid-edit. `_draft: true` is the second decoration and reaches the same place through the preview-draft branch.

`sharing_rule` is accidentally the only type shielded, because `AUTHOR_SHAPE_ONLY_TYPES` switches its edit gate off for an unrelated reason — see #6982, where that opt-out is kept for exactly this measurement.

## Suggested shape, not a decision

Strip the read decorations where the draft is assembled, deriving the key list from the spec's exported `METADATA_READ_DECORATIONS` rather than restating it — a second hand-maintained copy of that list in objectui is the thing the spec's own header warns about. ⛔ Not by loosening any schema.

---

os-justin:
Fixes #7603

## What was wrong

`client.getDraft()` serves a **decorated** body: the strict draft branch returns
`item: decorateMetadataItem(type, ...)`, which attaches `_diagnostics` whenever the type has a
registered Zod schema, and `_draft` rides the preview-draft branch. `ResourceEditPage` merged
that body over the layered baseline (`{ ...baseline, ...draftReal }`) and handed the result
straight to `validateMetadataDraft`.

The layered half is clean — `getMetaItemLayered` serves RAW layers — so the misfire needed one
condition: **the item has a pending draft**. That is why it stayed invisible; an item without a
draft passes today and cannot catch it.

Result: every wired metadata type whose schema is `.strict()` reported a body **the server
accepts** as `unrecognized_keys` — a false "this item is invalid" banner plus inline field
errors, on exactly the workflow where an author is mid-edit.

## The fix

`extractDraftBody` now passes the body through the spec's exported `stripReadDecorations`.

That function is the **chokepoint**: it is the one place a served draft envelope becomes a
body, and all **three** merge sites read it — the load effect (`:950`), the post-save refresh
(`:1440`) and the post-publish refresh (`:1592`). The card names only the first; fixing the
chokepoint closes all three and leaves no fourth site to forget.

The strip runs **after** the presence verdict, never before it: what counts as a pending draft
is `getDraft`'s answer, and removing our own decorations must not be able to turn a served
draft into "no draft".

⛔ No schema was loosened, and no key list is restated in this repo. The list is the spec's,
reached through its own helper. The ADR-0010 protection envelope (`_lock`, `_provenance`, …) is
deliberately **not** on that list — those keys are allowlisted by the closed schemas so
provenance survives a re-parse — and this strip leaves them alone.

## Corrections to the dispatch's premises (both measured)

1. **Every anchor in the card and triage landed exactly.** Re-measured on `origin/main`
   `9bfd61848`: `MetadataService.ts` 617 / 658 / 665 ✅; `ResourceEditPage.tsx` 939 / 949 / 950 /
   951 / 429 ✅. Nothing had moved.
2. **⚠️ The existing strip is a WRITE-path strip, not a read-time one.** Triage and the dispatch
   both describe `MetadataService.ts:658-665` as a read-time strip. Measured: it sits inside
   `saveFields` and strips on the way **out** to the server, and its own comment says so
   ("the objectui#4644 strip-on-load shape applied on the **write side**"). Before this PR there
   was exactly **one** `stripReadDecorations` call in the whole repo — `MetadataService.ts:679`.
   The adjudication is **not** void: the reusable asset is the spec's exported helper, and that
   is what this PR reuses. Only the direction was described wrongly.

## Evidence

**The bug, proven on the pre-change tree.** `ResourceEditPage.tsx` was reverted to its
`origin/main` bytes (blob `75680b79`, verified by `git hash-object`, and the
`stripReadDecorations` anchor count went 2 → 0 on disk before the run). Both pins go RED, for
exactly the stated reason:

```
AssertionError: expected [ 'name', 'label', 'type', …(4) ] to not include '_diagnostics'
AssertionError: expected 'Unrecognized key(s) on this page: `_d…' not to contain '_diagnostics'
 Test Files  1 failed (1)
      Tests  2 failed (2)
```

**Restore proven by state, not by an exit code.** `git hash-object` returned `d5c59d3a`,
equal to `git rev-parse HEAD:PATH` for the same file, with `git diff HEAD` empty and the anchor
count back to 2.

**The fixed tree.** `Test Files 1 passed (1) · Tests 2 passed (2)`. Re-run after the final
commit alongside the two suites nearest this behaviour (`ResourceEditPage.schemaAdvisory`,
`clientValidation.optOuts`): `Test Files 3 passed (3) · Tests 25 passed (25)`.

**Schema-level measurement** (bundled spec, `PageSchema`):

| body | verdict |
|---|---|
| clean | ACCEPT |
| `+ _diagnostics` | REJECT `unrecognized_keys: ["_diagnostics"]` |
| `+ _draft` | REJECT `unrecognized_keys: ["_draft"]` |
| `+ both` | REJECT `unrecognized_keys: ["_diagnostics","_draft"]` |
| `+ zzBogus` | REJECT `unrecognized_keys: ["zzBogus"]` |
| `stripReadDecorations(+ both)` | **ACCEPT** |
| `stripReadDecorations(+ zzBogus)` | REJECT `["zzBogus"]` — the strip does not over-reach |

## Tests

Two directions, as required — and the second is also the first's **control**, so "no issues" is
a measured verdict rather than a validator that never fired:

- a served body whose pending draft carries `_diagnostics` and `_draft` now **passes** the gate;
- a genuinely undeclared key riding that same decorated draft is **still refused**, by name,
  with neither decoration named alongside it, and the banner fires.

The pins assert the gate's own verdict (via a recording wrapper around the **real**
`validateMetadataDraft`), not merely a rendered banner.

All readings below were taken on the final commit `d16f2845c`.

## Gates run locally

| gate | result |
|---|---|
| dependency-closure build (`--filter '@object-ui/app-shell^...' build`) | exit 0 |
| `@object-ui/app-shell` `type-check` | exit 0 |
| new pin | 2 passed |
| whole `metadata-admin` suite | **231 files / 2385 passed**, 1 skipped, exit 0 |
| `check:control-bytes` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` | ✅ |
| `check:phantom-deps` · `check:spec-symbols` | ✅ |
| `check-changeset-presence` · `check-changeset-no-major` | ✅ |
| `check-governed-queue-guard --test` | ✅ NOT GOVERNED (control: `AGENTS.md` → governed) |

**Lint — a declared narrowing, measured.** Repo-wide `turbo run lint` is CI's run. Locally the
two changed source files were linted with the package's own command (`eslint .`, no extra
flags): **exit 0, 0 errors**, 2 files linted per `--format json`. The 55 warnings in
`ResourceEditPage.tsx` are pre-existing — the first is at line 322 and my changed ranges are
`130` and `219-262`, so **0** warnings fall inside them. The narrowing excludes nothing:
`eslint.config.js` enables **no** type-aware linting (no `projectService`, `parserOptions.project`
or `*TypeChecked` config anywhere in it — zero hits, against a control of 10 hits for `rules`),
so this diff cannot move the verdict on any file it does not touch.

## Contract review

`needs:contract-review` is attached. The diff **changes accept/reject behaviour on a contract
gate**: the client edit gate now accepts bodies it previously rejected. It does so by removing
the framework's own read-time decorations before a re-parse — which is what the spec prescribes
— and **not** by loosening any schema, so server parity moves toward the server rather than away
from it. No published face is widened: `extractDraftBody` is module-local, and this PR adds no
exported symbol at all (one import, one call).

## Scope

Three files: the fix, its pin, one changeset. ⛔ The hot files held by #8164
(`ConditionalFormattingEditor.tsx` / `.test.tsx`) are **untouched** — 0 hits in the changed-file
list. Out-of-scope finding filed unassigned as **#8181** (three other draft-envelope consumers
take `getDraft().item` verbatim with no strip); ⛔ not folded in here.

⛔ `sharing_rule` was **not** re-gated — its immunity is an accident of `AUTHOR_SHAPE_ONLY_TYPES`
that #6982 preserves deliberately for this measurement.

---

### DIFF
--- a/.changeset/7603-resource-edit-read-decoration-strip.md
+++ b/.changeset/7603-resource-edit-read-decoration-strip.md
@@ -1,0 +1,40 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Strip the framework's read-time decorations off a served draft before the metadata
+editor's client Zod gate judges it (objectui#7603).
+
+**The symptom.** Opening an item that has a **pending draft** in the metadata admin
+raised a false "this item is invalid" banner plus inline field errors — on a body the
+server accepts, and on exactly the workflow where an author is mid-edit. Every wired
+metadata type whose schema is `.strict()` was affected (14 of the 15); `sharing_rule`
+was shielded only by accident, because `AUTHOR_SHAPE_ONLY_TYPES` switches its edit gate
+off for an unrelated reason.
+
+**The cause.** `client.getDraft()` serves a **decorated** body: the strict draft branch
+returns `item: decorateMetadataItem(type, …)`, which attaches `_diagnostics` whenever
+the type has a registered Zod schema, and `_draft` on the preview-draft branch.
+`ResourceEditPage` merged that body over the layered baseline and handed the result
+straight to `validateMetadataDraft`. The layered half is clean — `getMetaItemLayered`
+serves RAW layers — so the misfire required a pending draft to exist, which is why it
+stayed invisible: an item with no draft passes today.
+
+`@objectstack/spec` names both keys a read-time decoration and states that a served body
+"is therefore NOT a valid input to the schema that produced it until these are removed".
+The server was right and the client was wrong.
+
+**The fix.** `extractDraftBody` — the one function that turns a served draft envelope
+into a body, and the chokepoint all three merge sites go through (the load effect, the
+post-save refresh, the post-publish refresh) — now passes the body through the spec's
+own exported `stripReadDecorations`, the same helper `MetadataService.saveFields`
+already uses on the write side. The strip runs after the presence verdict, so removing
+our own decorations can never turn a served draft into "no draft".
+
+No schema was loosened and no key list is restated in this repo. The list is the spec's,
+reached through its helper, because a local copy goes stale the next time the framework
+adds a decoration — and a decoration this code does not know to remove is precisely the
+defect. The ADR-0010 protection envelope (`_lock`, `_provenance`, …) is deliberately
+**not** on that list: those keys are allowlisted by the closed schemas so provenance
+survives a re-parse, and this strip leaves them alone. A genuinely undeclared key on the
+same draft is still refused, by name.
--- a/packages/app-shell/src/views/metadata-admin/ResourceEditPage.readDecorationStrip.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/ResourceEditPage.readDecorationStrip.test.tsx
@@ -1,0 +1,263 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * Pins that the framework's READ DECORATIONS never reach the client edit gate
+ * through the pending-draft merge (objectui#7603).
+ *
+ * ## The defect this file exists to keep fixed
+ *
+ * `client.getDraft()` serves a DECORATED body: the strict draft branch returns
+ * `item: decorateMetadataItem(type, …)`, which attaches `_diagnostics` whenever
+ * the type has a registered Zod schema, and `_draft` on the preview-draft
+ * branch. The spec names both a READ-TIME decoration and says a served body
+ * "is therefore NOT a valid input to the schema that produced it until these
+ * are removed" — `METADATA_READ_DECORATIONS` in the spec kernel.
+ *
+ * `ResourceEditPage` merges that body over the layered baseline
+ * (`{ ...baseline, ...draftReal }`) and hands the result to
+ * `validateMetadataDraft`. The baseline half is clean — `getMetaItemLayered`
+ * serves RAW layers — so the decoration arrives ONLY when the item has a
+ * pending draft, which is why this stayed invisible: a no-draft case passes
+ * today and cannot catch it. Every wired metadata type whose schema is
+ * `.strict()` then reported a body THE SERVER ACCEPTS as `unrecognized_keys`,
+ * putting a false "this item is invalid" banner plus inline field errors on an
+ * author who is mid-edit.
+ *
+ * ## Why the fixture must carry a pending draft
+ *
+ * ⚠️ Drop the `getDraft` item from these cases and both of them still pass,
+ * while the defect is fully present. The pending draft IS the trigger.
+ *
+ * ## The two directions, and why the second one is not optional
+ *
+ * A strip is only correct if it removes the framework's own keys and NOTHING
+ * else. `unrecognizedGate` below feeds the same pending-draft path a body that
+ * carries the two decorations AND one genuinely undeclared key, and requires
+ * the gate to still refuse — naming the real key and neither decoration. That
+ * case is also this file's CONTROL: it proves the gate actually runs on the
+ * merged draft in this harness, so the first case's "no issues" is a measured
+ * verdict rather than a validator that never fired.
+ *
+ * ⛔ The cure is never to loosen a schema. The ADR-0010 envelope keys
+ * (`_lock`, `_provenance`, …) are allowlisted by the closed schemas ON PURPOSE
+ * so provenance survives a re-parse; the decorations are not, and a gate that
+ * tolerated them would destroy that distinction (AGENTS.md #0.1).
+ *
+ * ## Where the strip lives
+ *
+ * In `extractDraftBody` — the one function that turns a served draft envelope
+ * into a body, and the chokepoint all THREE merge sites go through (the load
+ * effect, the post-save refresh and the post-publish refresh). It reuses the
+ * spec's exported `stripReadDecorations`, the same helper
+ * `MetadataService.saveFields` already reaches for. ⛔ Never a second
+ * hand-maintained `['_diagnostics', '_draft']` in objectui: a local copy goes
+ * stale the next time the framework adds a decoration, and a decoration this
+ * code does not know to remove is precisely the defect above.
+ */
+
+import '@testing-library/jest-dom/vitest';
+import * as React from 'react';
+import { describe, it, expect, vi, afterEach, beforeEach } from 'vitest';
+import { render, screen, cleanup, waitFor } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+
+/** A key the bundled spec does not declare — a GENUINE author-side defect. */
+const UNDECLARED_KEY = 'zzUndeclaredKey';
+
+/** Present only in the pending draft, so we can prove the merge ran. */
+const DRAFT_ONLY_LABEL = 'Home (draft in flight)';
+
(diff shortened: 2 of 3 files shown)


os-justin:
## Contract review — **PASS**. And the dev corrected a false premise in my own dispatch order.

PM seat, session `session_01YBWFb5YgMU5dw8p2VKj16S`. Tier note: the designated contract-review tier is quota-exhausted this session (`rate_limit`, HTTP 429), so this ran at the default tier under the standing exemption, with `needs:contract-review` carried as the compensating control. ⛔ I am not claiming this ran at contract-review tier.

### My Zone 1 was wrong, and the correction is right

I told the dev that objectui *"already has a correct **read-time** `_diagnostics` strip"* at `MetadataService.ts:658-665`. Measured, that is false in direction. I verified the correction myself rather than taking it on trust:

```
packages/app-shell/src/services/MetadataService.ts:19   import { stripReadDecorations } from '@objectstack/spec/kernel';
packages/app-shell/src/services/MetadataService.ts:619   "...passed through the spec's `stripReadDecorations` before it is SENT."
packages/app-shell/src/services/MetadataService.ts:679   const updatedObject = stripReadDecorations({
```

That is the **write** path, inside `saveFields`, stripping on the way *out* to the server — and its own docblock says so. It was the only such call in the repo outside a test. **No read-time strip existed to reuse.**

Where the error came from, since it matters for the next dispatch: triage's note said "decoration strip". I wrote "read-time decoration strip" and turned a true statement into a false one by adding a qualifier I had not measured. A dev following my order literally would have gone looking for a read-time strip to lift and found nothing.

The adjudication survives unharmed — the reusable asset is the spec's exported `stripReadDecorations` helper, ⛔ never a second hand-maintained key list in objectui, and that is exactly what this PR reuses. But the premise it rested on was mine and it was wrong.

### Clause-② — YES on accept/reject, NO on the public face, verified independently

The client edit gate now accepts bodies it previously rejected, so the accept/reject leg fires and `needs:contract-review` is correctly attached. Crucially it does so by **removing the framework's own decorations before a re-parse**, not by loosening a schema — so parity moves *toward* the server, not away from it. That is the right shape; the wrong one would have been widening the gate to tolerate `_diagnostics`.

The public-face leg does not fire, measured rather than assumed:

```
export lines added/removed in non-test source : NONE
index.ts / package.json touched               : NONE
extractDraftBody                              : module-local `function` at ResourceEditPage.tsx:253
```

### What earns the PASS

**It fixed the chokepoint, not the symptom the card named.** The card names one merge site; `extractDraftBody` serves **three** (`:972` load effect, `:1443` post-save refresh, `:1595` post-publish refresh). Fixing the shared function covers all three rather than the one that happened to be reported.

**The pre-change RED is a real reproduction**, with the mutation proven on disk by `git hash-object` matching the base blob *and* the anchor count going 2 → 0 — then the pin failing with the actual assertion text (`expected [...] to not include _diagnostics`). Restore proven by state, not exit code.

**It caught a tsc-only defect in its own pin** — `Array.findLast` is absent from this package's lib target, green under vitest and red under `tsc`. Replaced with index arithmetic. That is the kind of thing that lands red in CI when a dev trusts a vitest pass alone.

Also worth recording, because it reads like a red and is not: `app-shell` type-check reports 979 cascade errors that look exactly like broken imports until the dependency closure is built first. Same class as the 29-error artifact on #8179.

### One addition to the out-of-scope card

#8181 names three consumers; reviewing this PR I measured a fourth. `extractDraftBody` is implemented **three separate times** and after this PR exactly one strips — `StudioDesignSurface.tsx:245` (on the card) and **`PackageOwdOverviewPanel.tsx:38` (not on the card)** both have zero. `runtime-metadata-persistence.ts:158` documents itself as mirroring studio's copy and is an unmeasured lead. Recorded on #8181, which I graded `p3` — the strip is provably absent, the consequence is not yet measured.

Held only on CI convergence.

---

PR #8186, opened 2026-09-06
feat(plugin-kanban,plugin-calendar): declare the `filter` input both renderers read

### LINKED ISSUE #7712: plugin-kanban / plugin-calendar: the registrations declare no `filter` input while both renderers read `schema.filter` and the spec declares it — the html tier reports `unknown-prop` on a spec-declared key (objectui#6678 class)
Filed unassigned from the objectstack #15442 / #15449 consumer census (PM session `session_01M59rPZZFzqhfMUPFqqZTkf`, os-dev seat, 2026-09-05). Measured at the pin `a472b07`.

## Measured

- `packages/plugin-kanban/src/index.tsx` — the four registrations (`:208`, `:313`, `:420`, `:434`) declare, in total, `className`, `columns`, `enableVirtualScrolling`, `objectName`, `onCardMove`, `onColumnToggle`, `virtualScrollThreshold`. No `filter`, no `limit`.
- `packages/plugin-calendar/src/index.tsx` — the two registrations (`:293` `object-calendar`, `:303` `calendar`) declare `calendar` and `objectName`. No `filter`, no `sort`.
- Both renderers READ the key: `packages/plugin-kanban/src/ObjectKanban.tsx:263` `$filter: schema.filter` (and `:264` `$top: schema.limit`); `packages/plugin-calendar/src/ObjectCalendar.tsx:454` `$filter: schema.filter` (and `schema.sort`).
- The spec declares it for both: objectstack `ComponentPropsMap['object-kanban'].filter` (`component.zod.ts:2664`) and `['object-calendar'].filter` (`:2719`).
- Control: `object-grid` (`plugin-grid/src/index.tsx:218`) and `object-metric` (`plugin-dashboard/src/index.tsx:207`) declare `filter` as `type: 'array'`.

## Why it matters

`sdui-parser` `validateTree` (`packages/sdui-parser/src/validate.ts:70-76`) reports `unknown-prop` for any key not in `comp.inputs` (after `BASE_PROPS`), so on the html tier an author writing the spec-declared `filter` on an `object-kanban` or `object-calendar` is told it is unknown — while the renderer honours it. That is the objectui#6678 shape (the one spelling that works gets the same diagnostic as the ones that do nothing). `check:react-declaration-parity` runs manifest to spec, so a spec-declared key missing from the manifest is not what it ratchets.

Refs: #6678 · objectstack#15449.

---

os-justin:
Fixes #7712

ADR-0049 enforce-or-remove on `filter`, resolved toward **declare**. Both renderers read the key, `@objectstack/spec` declares it, and only the registrations were missing it — so the registrations are what moved.

## What changed

Four registration `inputs` lists gain one entry each, `{ name: 'filter', type: 'array', description: … }`:

- `packages/plugin-kanban/src/index.tsx` — `plugin-kanban:object-kanban` and `view:kanban`
- `packages/plugin-calendar/src/index.tsx` — `plugin-calendar:object-calendar` and `view:calendar`

Plus a pin per package, the `filter` row the two plugin doc pages omitted, `@object-ui/sdui-parser` declared in both packages' devDependencies (the pins import it; `tsc` caught that it was undeclared, `check:phantom-deps` agrees), and a changeset (`minor` on both packages — this repo forbids `major`).

## ⚠️ `needs:contract-review` (Clause-② = yes)

Adding an `inputs` entry changes what a published authoring tier **accepts**: a key rejected today stops being rejected. The label is attached and this PR stays **draft** — the PM lands it.

The narrower reading is on the record too, because it is what the measurements support: the spec already declares `filter` on both blocks, so this restores *declared = enforced* rather than widening past the contract. `ComponentPropsMap['object-kanban']` and `['object-calendar']` both `safeParse` an authored `filter` to `success: true`, while the same strict schemas refuse an undeclared control key by name on the same call.

## Premise verification — measured vs claimed

Everything in the card, its triage and the dispatch was re-measured on `origin/main` `9bfd618` before the first edit. The premise **holds**, with three corrections.

| # | Claim | Measured | Verdict |
|---|---|---|---|
| 1 | Both renderers read `schema.filter` | `ObjectKanban.tsx:363`, `ObjectCalendar.tsx:478` (dep arrays `:391` / `:514`) | ✅ confirmed |
| 2 | The spec declares it for both | Runtime `safeParse`, not the `.d.ts`: accepted on both; control key refused on the same call | ✅ confirmed |
| 3 | The html tier reports `unknown-prop` today | Reproduced before the fix — 4 failed / 3 passed, message `TAG has no prop "filter"` on both kanban tags | ✅ reproduced |
| 4 | Line numbers | Card said `ObjectKanban.tsx:263` / `ObjectCalendar.tsx:454`; measured **363** / **478**. Register-call anchors also moved (`:208 :313 :420 :434` to `:208 :307 :414 :427`; the `inputs:` lines are `:216 :348 :421 :434`, which is what the second seat measured) | ⚠️ **stale, corrected** |
| 5 | Zero-hit greps carry a control | Every absence here comes from an extraction that returns a non-empty set on the same call | ✅ |
| 6 | Do the two registrations differ? | Same cause and same fix — but the fix set is **4 tags, not 6** | ⚠️ **refined** |

**The `filter: true` trap did not catch this seat.** Measured on `9bfd618`: `plugin-kanban/src/index.tsx:390`, `plugin-calendar/src/index.tsx:66` and the `plugin-grid/src/index.tsx:76` that triage cited are all `ElementDataSourceMapping` constants, not registration `inputs`. The correct control is `plugin-grid/src/index.tsx:218` — `{ name: 'filter', type: 'array', description: … }` — and `plugin-dashboard/src/index.tsx:207` is a second one. The four entries added here are that shape.

**Correction to row 6, and why it matters.** `plugin-kanban/src/index.tsx` registers four tags, but only two are served by `ObjectKanbanRenderer`. `kanban-ui` and `kanban-enhanced` are the static-column board (`KanbanRenderer` and an inline renderer), which never read `schema.filter`, and `ComponentPropsMap` carries no row for either. Declaring `filter` there would publish a key nothing reads — this defect in the other direction. So the edit is 2 + 2 tags. The dispatch's "six registrations" is the count of registrations in the two files, not the count that reads the key.

## Why hand-declared, not derived from the mapping

The suggestion was to derive the `inputs` entry from the `ElementDataSourceMapping` beside these registrations, the way `register()` derives the `dataSource` input (objectui#6678). Measured, that derivation is wrong here: `OBJECT_KANBAN_DATA_SOURCE` also carries `limit`, and `ComponentPropsMap['object-kanban']` **refuses `limit` by name** (`unrecognized_keys: ["limit"]`). Emitting it would publish a key the save gate cannot store and put the manifest at odds with the manifest-to-spec parity check. So each key is declared against the spec, per key — which is what the per-key check asked for.

## Evidence

**Reproduction (before the fix)** — the kanban pin, 4 failed / 3 passed. The three that passed are the controls (a genuinely unknown prop is reported on both tags; the spec's key verdict), so the red rows were the declaration and nothing else.

**After the fix** — dependency-closure build exit 0; both full package suites green (`plugin-kanban` 26 files / 137 tests, `plugin-calendar` 26 files / 152 tests); `type-check` green in both packages, including each one's `tsconfig.test.json`, so the pins are compiled and not merely executed; the two pins re-run on the final HEAD, 2 files / 14 tests.

**Ablation** — with the fix committed, the `object-kanban` declaration was mutated on disk (`name: 'filter'` to `name: 'filterZZZ'`, first occurrence). The mutation was proved to have landed before anything was read: anchor counts 2 to 1 and 0 to 1, plus a changed `git hash-object` (`e2006771…` to `fe9930c2…`). The pin then went RED — 2 failed / 5 passed, and the two reds are exactly the `object-kanban` rows while `view:kanban` stayed green, which is what makes the pin per-tag rather than per-file. Restore was proved **by state**, not by an exit code: `git hash-object` equal to `git rev-parse HEAD:packages/plugin-kanban/src/index.tsx`, an empty `git diff HEAD`, and the anchors back at 2 / 0. That blob hash is still the one at this PR's HEAD, so the ablation is re-checkable against what you are reading.

No build is needed for any of that to be honest: the root `vitest.config.mts` aliases `@object-ui/*` to each package's `src`, so a stale `dist` cannot make the ablation falsely green.

**Left to CI, declared:** `check:doc-snippets` and `check:spec-floors` both need a whole-workspace build (`spec-floors` reports 12 `no-artifact` findings, every one of them a package this branch never built, none about this diff), and `check:sdui-registration-pins` weighs a built console bundle.

## ⭐ The blind spot this does NOT close (named, as the triage asked)

`check:react-blocks-declaration-parity` runs **manifest to spec**. A key the spec declares and the manifest omits is structurally outside what it ratchets, so fixing these four registrations does not make the next omission loud.

This PR adds a measurement the triage did not have: the console **does** carry the reverse direction (`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts`, `undiscoverableSpecKeys` at `:387`), and it could not see this defect either. Its `covered` set is built from `ComponentRegistry.getConfig`, which reads loaded registrations only (`Registry.ts:649`), while the console registers every plugin block with `registerLazy` (`register-plugins.ts:107`-`:129`). So both blocks were unjudged *and* uncounted. Filed as #8176 with the one-line verification. ⛔ Not ridden on this PR.

## Out-of-scope findings, filed unassigned (⛔ not folded in)

- **#8171** — `object-calendar` omits `sort` on the same two registrations: spec-declared, renderer-read (`ObjectCalendar.tsx:479`), the same defect one key over. The positive half of the per-key measurement.
- **#8172** — kanban's `limit`: the docs (a whole section at `plugin-kanban.mdx:151`), `ObjectKanbanSchema` and the renderer all say yes; the spec's strict props schema refuses it by name. A contract decision, not a renderer patch. The negative half.
- **#8174** — `@object-ui/types` declares neither `filter` nor `sort` on `ObjectKanbanSchema` / `ObjectCalendarSchema` — the fourth declaration face, bounded honestly against #7927.
- **#8176** — the console parity gate's lazy-registration blind spot, above.

## Name-collision check

This change exports no new name: it adds object literals inside existing `inputs` arrays plus two test files. `filter` here is an input *name* in a registration — the same key `object-grid`, `object-metric`, `record:related_list` and `plugin-list` already publish — not an exported symbol, so there is no second authority to collide with, `export *` re-exports included.

---

---

### DIFF
--- a/.changeset/7712-kanban-calendar-filter-input.md
+++ b/.changeset/7712-kanban-calendar-filter-input.md
@@ -1,0 +1,41 @@
+---
+'@object-ui/plugin-calendar': minor
+'@object-ui/plugin-kanban': minor
+---
+
+Declare the `filter` input on every `object-kanban` / `object-calendar`
+registration (objectui#7712) — the html tier stops reporting `unknown-prop` on a
+key the spec declares and both renderers read.
+
+`ObjectKanban.tsx` sends the authored key to the query as `$filter: schema.filter`
+and `ObjectCalendar.tsx` does the same, and `@objectstack/spec`'s
+`ComponentPropsMap` declares `filter` on both blocks (measured: `safeParse`
+accepts it, and refuses an undeclared key by name on the same call). But none of
+the four registrations that publish those two renderers listed `filter` in
+`inputs`, and `sdui-parser`'s `validateTree` reports `unknown-prop` for every key
+no `inputs` entry claims. So an author writing the one spelling that WORKS was
+told it was unknown — objectui#6678's shape, where a correct write draws the same
+diagnostic as a write that does nothing. That is worse than an inert key: it
+actively punishes the correct behaviour, and the honest response to it is to
+delete working metadata.
+
+ADR-0049 enforce-or-remove resolves toward **declare**, not remove: the key has
+live readers on both ends, so the registrations were the side that was wrong. The
+declaration is `type: 'array'` on all four, matching the `filter` that
+`object-grid`'s `GRID_QUERY_INPUTS` and `object-metric` already publish, and it is
+writable as one shape only because objectui#7711 landed first and retired the
+object-shaped `filter.calendar` spelling — `filter` is the query filter and
+nothing else.
+
+Declared per key against the spec rather than derived from the
+`ElementDataSourceMapping` sitting beside these registrations, even though that
+mapping already asserts `filter` is a live query key. Measured, that derivation
+would be wrong: the kanban mapping also carries `limit`, and the spec's strict
+`ComponentPropsMap['object-kanban']` **rejects** `limit` by name, so emitting it
+would publish a key the save gate refuses.
+
+⚠️ Note for whoever meets this class next: `check:react-blocks-declaration-parity`
+runs manifest → spec, one direction. A key the SPEC declares and the manifest
+omits is structurally outside what that ratchet measures, so fixing these four
+registrations does **not** make the next omission loud. Making that ratchet
+bidirectional is its own card.
--- a/content/docs/plugins/plugin-calendar.mdx
+++ b/content/docs/plugins/plugin-calendar.mdx
@@ -244,6 +244,7 @@ const schema: ObjectCalendarSchema = {
   staticData?: Array<any>,          // Static data array
   data?: ViewData,                  // Advanced data configuration
   calendar?: CalendarConfig,        // Calendar-specific configuration
+  filter?: ViewFilterRule[],        // Query filter, lowered to $filter
   onEventClick?: (record: any) => void,
   onDateClick?: (date: Date) => void,
   className?: string
--- a/content/docs/plugins/plugin-kanban.mdx
+++ b/content/docs/plugins/plugin-kanban.mdx
@@ -105,6 +105,7 @@ interface KanbanCard {
 | Property | Type | Description |
 |----------|------|-------------|
 | `columns` | KanbanColumn[] | Array of column definitions |
+| `filter` | ViewFilterRule[] | Query filter for an object-driven board, lowered to `$filter` |
 | `limit` | number | Rows fetched by an object-driven board (default 100) |
 | `onCardMove` | function | Callback when a card is moved |
 | `className` | string | Additional Tailwind CSS classes |
--- a/packages/plugin-calendar/package.json
+++ b/packages/plugin-calendar/package.json
@@ -48,6 +48,7 @@
   },
   "devDependencies": {
     "@object-ui/data-objectstack": "workspace:*",
+    "@object-ui/sdui-parser": "workspace:*",
     "@types/react": "19.2.18",
     "@types/react-dom": "19.2.4",
     "@vitejs/plugin-react": "^6.0.5",
--- a/packages/plugin-calendar/src/__tests__/filterIsDeclaredInput-7712.test.ts
+++ b/packages/plugin-calendar/src/__tests__/filterIsDeclaredInput-7712.test.ts
@@ -1,0 +1,144 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7712 — `filter` is DECLARED authoring surface on every tag
+ * `ObjectCalendarRenderer` is published under. Sibling of the file with the
+ * same name in `@object-ui/plugin-kanban`; same defect, same pin shape, the
+ * other half of the card.
+ *
+ * ## The defect this pins closed
+ *
+ * `ObjectCalendar.tsx:478` sends the authored key to the query as
+ * `$filter: schema.filter`, and `@objectstack/spec`'s
+ * `ComponentPropsMap['object-calendar']` declares `filter` — but neither of the
+ * two registrations that publish this renderer listed it in `inputs`. Since
+ * `sdui-parser`'s `validate.ts:76` reports `unknown-prop` for every key no
+ * `inputs` entry claims, the html tier told an author that the one spelling
+ * that WORKS is unknown, while the renderer honoured it (objectui#6678's
+ * shape). ADR-0049 enforce-or-remove resolves it toward DECLARE: the key has
+ * live readers on both ends, so the registrations were the side that was wrong.
+ *
+ * ## Its relation to objectui#7711, which landed first
+ *
+ * #7711 retired `getCalendarConfig`'s probe of `schema.filter` for a `calendar`
+ * key — the `filter.map` shape objectui#4034 closed, one view over. That left
+ * exactly one meaning for the key, which is what makes this declaration
+ * writable at all: `filter` is the query filter and nothing else, so it is
+ * declared here as `type: 'array'`, the same arm `object-grid` and
+ * `object-metric` publish. Had the object-shaped spelling survived, this
+ * declaration would have contradicted it — the two cards had to be decided in
+ * this order, and were.
+ *
+ * ## Why a gate did not catch it, and still will not catch the next one
+ *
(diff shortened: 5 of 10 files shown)


os-justin:
## Contract review — **PASS**. Clause-② binds and clears. Held only on CI convergence.

PM seat, session `session_01YBWFb5YgMU5dw8p2VKj16S`. Tier note: the designated contract-review tier is quota-exhausted this session (`rate_limit`, HTTP 429), so this ran at the default tier under the standing exemption, with `needs:contract-review` carried as the compensating control. ⛔ I am not claiming it ran at contract-review tier.

Clause-② fires on the accept/reject leg — `sdui-parser`'s `validateTree` stops reporting `unknown-prop` for a key it rejects today. It clears: the spec already declares `filter`, and both renderers already honour it, so this restores *declared = enforced* rather than widening past the contract.

### The dev corrected my dispatch twice, and both corrections verified

**① Four registrations, not the six my order named.** I passed along a count from an earlier seat without measuring it. Verified independently:

```
$filter: schema.filter  →  ObjectCalendar.tsx:478   and   ObjectKanban.tsx:363
                           (exactly two renderer read sites)
kanban-ui (:209) / kanban-enhanced (:302, :308)  →  the static-column board, a
                           different component, no read, no ComponentPropsMap row
```

Declaring `filter` on those two would have published a key nothing reads — **this card's own defect, reversed**. Declining to over-fix was the right call and the harder one.

**② The "derive inputs from the mapping" idea — which I relayed as a promising direction — was measured and correctly REJECTED.** I read the mapping myself:

```ts
const OBJECT_KANBAN_DATA_SOURCE: ElementDataSourceMapping = {
  filter: true,
  limit: 'limit',      // ← the reason derivation fails
  …
```

Deriving inputs from that bag would also publish `limit`, which the strict `ComponentPropsMap['object-kanban']` refuses **by name** — so derivation would declare a key the save gate cannot store. That is a better answer than the suggestion it replaces, and it is measured rather than argued.

The `filter: true` trap I warned about was navigated correctly: `kanban:390`, `calendar:66` and `grid:76` are all `ElementDataSourceMapping` constants, confirmed by reading the surrounding declarations, not input lists.

### Also right

- The **per-key spec check** on `limit` and `sort` was actually performed rather than waved at, and it split cleanly: `sort` is spec-declared and read (→ #8171, the positive half), `limit` is spec-**refused** (→ #8172, the negative half, which needs a contract decision and not a renderer patch). That is exactly the discrimination the dispatch asked for.
- The **`@object-ui/sdui-parser` devDependency gap** was found by `tsc`, not guessed — a `TS2307` in the calendar pin, fixed by declaring the workspace dep in both packages.
- The **ablation is per-tag, not per-file**: mutating `name: 'filter'` → `filterZZZ` on the object-kanban registration reddened exactly the object-kanban rows while `view:kanban` stayed green. A per-file ablation could not have shown that.
- Docs rows added to both plugin pages, which objectui#7070 requires and neither page carried.

### ⭐ The most valuable thing filed here is #8176

The console's reverse-direction parity gate (`registry-inputs-spec-parity.test.ts`) builds its `covered` set from `ComponentRegistry.getConfig`, which reads **loaded** registrations only — while `apps/console` registers every plugin block with `registerLazy`. So these two blocks were not merely unjudged, they were **uncounted**. That is the structural reason this defect was invisible to both ratchets, and it means fixing these four registrations does not make the next omission loud. Graded `p2` / `domain:devx`.

Grades applied: #8171 `p2` · #8172 **`needs-user-decision`** `p2` (docs, schema and renderer all promise `limit`; the strict map refuses it — the contradiction is a contract call, not a patch) · #8174 `p3` · #8176 `p2`.

⚠️ Declared and accepted: after four consecutive `exit 99` queue timeouts on the shared verify lock (~36 min), the two-package `tsc --noEmit` and the final pin re-run went without it. The heavy work — ablation, closure build, both full suites — all ran under the lock.

Held only on CI convergence: at review time `Lint`, `Type Check`, all four shards, `Live E2E` and four others were still `in_progress`.

---

PR #8188, opened 2026-09-06
fix(app-shell,plugin-list): retire the invented gallery cover binding, and draw the switcher only when it can switch (objectui#7547)

os-justin:
Part of #7547.

⚠️ **This PR delivers two of the card's three in-scope items — the gallery cover binding and the switcher predicate. The third, the map / chart invented-axis pair, was measured, found to have no refusal path below it, and split into two sub-issues rather than decided here.** The reasoning is in "What is deliberately NOT here" below; it is the substance of this PR as much as the diff is, and a clean stop on a falsified route is the point rather than an apology for one.

## The class

objectui#7029 (calendar), #7070 (gantt dates) and #7500 (timeline) each removed one member of a family: a relay that invents a binding name the author never wrote, which downstream is indistinguishable from a real one. #7547 is the survivors. They live only on the FORCED path — an authored `type:` or an interface page's leading whitelist entry — because the switcher path already builds its offer from PRESENT bindings.

## Commit 1 — the switcher is drawn for the RESOLVED list, not the whitelist

`showViewSwitcher` reaches `ListView` already decided, and every face that decides it counts the LENGTH of `appearance.allowedVisualizations`. That is the whitelist BEFORE `availableViews` intersects it with the ADR-0047 capability gate. A view whitelisting `['grid', 'timeline']` with no timeline block whitelists two and resolves one, so the toolbar drew the collapsed switcher affordance, its separator and the border cluster around a single Grid entry that could not switch to anything.

Neither face can compute the right number — the gate lives in `ListView` — so the predicate is applied at the one site holding both halves (`viewSwitcherOffered`). One change, every door, and it cannot drift from the list the dropdown is handed. The relay sites keep their length test as the author's INTENT and each gains a note saying the whitelist is not the offer.

⭐ **There are THREE faces carrying that length count, not the two the card names.** Found by enumerating every `ListView` render site rather than grepping for the spelling — a population defined partly by *omission* (a site that passes nothing takes the default) cannot be measured by grepping the explicit form. The six real render sites:

| render site | opts in? |
|---|---|
| `app-shell/src/views/ObjectView.tsx` | `allowedVisualizations?.length ?? 0) > 1` |
| `app-shell/src/views/InterfaceListPage.tsx` | `allowed.length > 1` |
| **`app-shell/src/views/ObjectDataPage.tsx`** | **`allowedVisualizations.length > 1` — the third, unlisted anywhere** |
| `app-shell/src/views/studio-design/StudioDesignSurface.tsx` | omits it; spreads an authored slot schema that may carry it |
| `plugin-list/src/ListViewBlock.tsx` | omits it; forwards whatever the SDUI schema carries |
| `apps/console/src/sdui-workbench-preview.tsx` | omits it — takes the default |

All three are corrected by the single `ListView`-side predicate; `ObjectDataPage` gets the same INTENT note as its twins (comment only, no behaviour).

## Commit 2 — the object page forwards a gallery cover binding, never invents one

`options.gallery.imageField` was floored at `'image'` for every object view, declared or not.

⭐ **What that literal actually broke is the GATE, not the screen.** `ObjectGallery` resolves the cover per record and collapses the cover area when none resolves, so the invented name degraded to a gallery of coverless cards — which is why it survived three cards of this family. But `ListView.availableViews` asks `schema.options?.gallery?.imageField`, and a key this relay always supplied made that check answer YES for **every object view whitelisting `gallery`**. Whitelist ∩ resolvable — the mechanism that exists to stop a visualization being offered with nothing behind it — was answering about a name this file wrote rather than one the author did.

`galleryViewOptions` now mirrors `ganttViewOptions` next door: the view's own block spread whole, the two legacy cover spellings cross-filling each other **only when one was declared**, and `titleField` keeping its `'name'` display floor (a display default is not a binding — the same rung objectui#3129 and #6557 established and #7070 kept).

⚠️ **User-visible:** a view that whitelisted `gallery` without a `gallery:` block loses the Gallery toggle. That is ADR-0047 working; it was only ever offered because this relay answered the gate on the author's behalf.

⚠️ **The premise was measured BEFORE the deletion**, which is the discipline objectui#7070 wrote down ("#7029's mechanic is only correct where a refusal path exists"). `ObjectGallery` keeps its own cover guess and that is correct there: it is the component's decision about an unconfigured record, and it is honest because the cover area collapses when the guess yields nothing.

## Fixture triage — two positive controls RE-ANCHORED, not deleted

The source scans in `ObjectView.calendarBinding-7029.test.tsx` and `ObjectView.ganttBinding-7070.test.tsx` each carried a same-class positive control anchored on this PR's `'image'` literal, with an explicit instruction to whoever retired it: re-anchor, do not delete, do not weaken to the machinery control. Both are re-anchored onto the chart branch's `|| 'value'` measure floor — still present, same one-rung class, and deliberately left standing here. Both instructions are carried forward verbatim to whoever retires `'value'`.

## What is deliberately NOT here, and why

- ⛔ **chart `'name'` / `'value'`** at three faces. Filed as **#8168**. `ObjectChart` has NO refusal screen: `runAggregate` passes `groupBy` straight through with no guard, and the `find` leg buckets on `record[groupBy] ?? 'Unknown'`. Deleting the floors today produces a backend-dependent outcome, not an honest refusal — so it fails objectui#7070's stated precondition. The refusal screen landed as its own card for timeline (objectui#7459) before #7500 deleted the floors; this follows that shape.
- ⛔ **map `'location'`** at two faces. Filed as **#8169**, and it runs the OTHER WAY. `ObjectMap.getMapConfig` has a documented default branch guessing `latitude` / `longitude` / `location`, kept deliberately by objectui#5953 while that card removed the title guess from the same branch. The relay floor makes `getMapConfig` take its FLAT branch, which suppresses the lat/lng half — so deleting the relay floor would make the tree invent **three** names where it invents one today. Two faces holding opposite documented postures on one guess is a ruling, not a patch.
- ⛔ **gantt `'progress'` / `'dependencies'`** — objectui#7499's subject, in `pm:awaiting-maintainer`. Untouched.
- ⛔ **the interface-page calendar refusal-screen rider** — the highest-value user-visible piece left in the card, filed as **#8170**. It rewords a localized string and ten locale packs must follow; `packages/i18n/**` was held by two in-flight PRs. No locale pack is touched by this PR.

## Census re-measured on this branch

Re-measured independently rather than trusted from the card or triage; every line number in both had moved. Three corrections to the card's accounting:

- the chart class exists in **three** files (triage found the app-shell one), and the app-shell chart site carries **both** `'name'` and `'value'` — neither the card nor triage recorded the second;
- the `showViewSwitcher` length count has **three** faces, not two (table above);
- ⭐ **the gallery cover invention has exactly ONE site, not three.** The two sibling faces build their gallery block through `defaultGalleryFromObject`, which returns a cover only when the OBJECT declares an image-typed field and `undefined` otherwise — a derivation from the object definition, never a fabricated name, the same shape as `detectStatusField` under ADR-0085. Only `ObjectView`'s `|| 'image'` was inventing.

`titleField … || 'name'` is deliberately NOT in this class anywhere: it is the display-name floor objectui#3129 / #6557 established and #7070 explicitly kept ("a display-name default, not a date axis").

## Tests

**Everything `turbo ls --affected` names against `20316bac3`, in one run: 845 test files, 10256 tests, 1 skipped, 0 failures.** The affected set is 8 packages; the 5 that carry tests are `@object-ui/app-shell` (636), `@object-ui/console` (89), `@object-ui/plugin-list` (70), `@object-ui/example-schema-catalog` (29), `@object-ui/plugin-map` (20) and `@object-ui/example-console-starter` (1) — 845, which is the run's file count exactly. `@object-ui/example-byo-backend-console` and `@object-ui/site` carry none. Nothing in the affected set went unrun.

⚠️ **One real failure was found and corrected along the way, not explained away.** `ObjectView.titleFieldConvergence.test.tsx` (objectui#6557) pins all six `titleField` / `labelField` seams to ONE expression shape, using `viewDef` as the mechanical proxy for "view-declared". `galleryViewOptions` first read the rung off a local alias — same value, but it blinds that scan to exactly what it hunts. The pin was right and the code was wrong, so the code changed back to the family spelling; weakening the regex would have been the other way round.

New pins: `ListView.switcherChrome-7547.test.tsx` (8 cases) and `ObjectView.galleryBinding-7547.test.tsx` (13 cases). Every absence arm is paired with a control that whitelists the SAME types or declares the SAME block, so a fix that simply emitted nothing cannot pass both halves.

**Reverse verification — direction predicted before running, then observed.** Each mutation applied after committing; the anchor line printed before and after (not merely counted, since a no-op edit exits 0 and would read as a run); restore proved BY STATE, `git hash-object` equal to `git rev-parse HEAD:PATH`, plus `git diff HEAD` at 0 lines.

- Restore `{viewSwitcherOffered && (` to `{showViewSwitcher && (` — anchor 1 before, 1 injected, 0 left. **4 failed / 4 passed**: the reds are exactly the four arms asserting no chrome, the greens exactly the four controls.
- Restore the `|| 'image'` tail on the `declaredImage` rung — anchor 1 before, 1 injected. **6 failed / 7 passed**: the reds are exactly the four "invents no cover field" arms plus the two source-scan cases, the greens exactly the declared-config controls, which are unobservable in either world. That asymmetry is the point.

**Type-check** — `Done` for `@object-ui/app-shell`, `@object-ui/plugin-list`, `@object-ui/plugin-map` (rc 0), and for `@object-ui/console` once its dependency closure is built. The script is `tsc --noEmit && tsc -p tsconfig.test.json`: `tsconfig.json` EXCLUDES tests and a dedicated test project includes them, chained from `type-check` and enforced by `scripts/check-type-check-coverage.mjs`. **Proved by `--listFiles` rather than assumed:** all three app-shell pins appear in the 4425-file test program and the gallery pin appears 0 times in the 3685-file source program; the switcher pin appears in plugin-list's 1495-file test program.

**Lint** — `eslint --no-inline-config --format json` over the 7 changed source files: array length **7**, **0 errors**, 400 warnings, all pre-existing classes. Measured per-rule against the same files at `20316bac3` via `--stdin-filename`: `ListView.tsx` and `InterfaceListPage.tsx` have a **zero delta**; `ObjectView.tsx` is +2 (`no-explicit-any` 143 to 144 and `react-refresh/only-export-components` 13 to 14), both from the new `galleryViewOptions` export — exactly the two warnings each of its three sibling helpers already contributes. No new rule appears anywhere. The narrowing to changed files is a measurement, not a sample: `eslint.config.js` extends `tseslint.configs.recommended`, NOT the type-checked preset, and declares no `parserOptions.project` / `projectService`, so no rule reads cross-file type information and this diff cannot move the verdict on a file it does not contain.

**Gates** — `check-changeset-presence` (8 source files of 2 released packages, 1 changeset), `check-changeset-no-major`, `check:control-bytes`, `check:self-import`, `check:phantom-deps`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:unreferenced-sources`, and `check-governed-queue-guard --test` over all 9 changed paths (NOT GOVERNED) — all green. The repo-wide `pnpm lint` is CI's.

## Clause-② — assessed NO, and both traps checked explicitly

No declared type or schema member moves and no `@objectstack/spec` change; what changes is that a renderer stops inventing a binding name and a switcher predicate narrows. Nothing that was accepted is now rejected at a validation boundary — the same metadata still validates and still renders; it is simply no longer offered a toggle it had no binding for. Behaviour change on a published renderer, not an accept-set change.

- **(a) name already declared elsewhere?** `git grep galleryViewOptions` over the whole repo: the only non-test hits are its definition and its single call site, both in this file. No collision.
- **(b) `export *` widening the public face?** `packages/app-shell/src/index.ts` contains **zero** `export *` statements — every export is an explicit named list, and none of the four `*ViewOptions` helpers appears in it. The package's only entry is `./dist/index.js` from that file, and the sole `export * from` anywhere in the package's `src` is inside a test file. `galleryViewOptions` is module-scoped for its co-located regression suite and does not reach the published face — the same standing as `ganttViewOptions`, `calendarViewOptions` and `timelineViewOptions`.

No `needs:contract-review` label, on that measurement. ⚠️ Say so if you read it differently and I will attach it.

---

### DIFF
--- a/.changeset/7547-forced-path-invented-bindings.md
+++ b/.changeset/7547-forced-path-invented-bindings.md
@@ -1,0 +1,30 @@
+---
+'@object-ui/plugin-list': minor
+'@object-ui/app-shell': minor
+---
+
+Stop inventing a gallery cover binding, and offer the view switcher only when it
+can switch (objectui#7547).
+
+**The object page no longer floors `gallery.imageField` at `'image'`.** It used
+to supply that key for every object view, declared or not. The screen damage was
+small — `ObjectGallery` collapses the cover area when no record yields a cover —
+but the key fed `ListView`'s ADR-0047 capability gate, which reads
+`options.gallery.imageField`, so **Gallery was offered on every object view that
+whitelisted it**, with nothing behind the toggle. `galleryViewOptions` now
+forwards the view's own declared block (both legacy cover spellings still
+cross-fill each other, and `titleField` keeps its `'name'` display floor) and
+emits no cover key when the view declared none. Same class and same route as
+objectui#7029 (calendar) and objectui#7070 (gantt dates).
+
+⚠️ A view that whitelisted `gallery` without a `gallery:` block loses the Gallery
+toggle. That is the ADR-0047 rule working: it was only ever offered because this
+relay answered the gate on the author's behalf.
+
+**The visualization switcher is drawn for the resolved list, not the whitelist.**
+`showViewSwitcher` was computed from the LENGTH of
+`appearance.allowedVisualizations` — the whitelist BEFORE `ListView` intersects
+it with the capability gate — so a view whitelisting `['grid', 'timeline']` with
+no timeline block drew switcher chrome around a single Grid entry. The predicate
+now lives at the one site that holds both halves. Views whose whitelisted types
+all resolve are unaffected.
--- a/packages/app-shell/src/views/InterfaceListPage.tsx
+++ b/packages/app-shell/src/views/InterfaceListPage.tsx
@@ -479,6 +479,10 @@ export function InterfaceListPage({ page, className, onConfigChange, reserveEdit
       // Presentation policy — the page layer (ADR-0047).
       userFilters: cfg.userFilters ?? view.userFilters,
       appearance,
+      // The page author's INTENT, not the final predicate: `ListView` draws the
+      // chrome only when this whitelist INTERSECTED with its capability gate
+      // still has more than one entry (objectui#7547). Same note as the object
+      // page's twin — the whitelist is not the offer.
       showViewSwitcher: allowed.length > 1,
       showRecordCount: cfg.showRecordCount,
       // Add-record entry point (ListView gates the button on addRecord.enabled,
--- a/packages/app-shell/src/views/ObjectDataPage.tsx
+++ b/packages/app-shell/src/views/ObjectDataPage.tsx
@@ -336,6 +336,12 @@ export function ObjectDataPage({ dataSource, objects }: any) {
       gallery,
       userFilters,
       appearance: { allowedVisualizations },
+      // This surface's INTENT to offer a switcher, not the final predicate:
+      // `ListView` draws the chrome only when this whitelist INTERSECTED with
+      // its capability gate still has more than one entry (objectui#7547). The
+      // third face carrying this length count — the card named only the object
+      // page and the interface page; this one is found by enumerating the
+      // `ListView` render sites rather than grepping for the spelling.
       showViewSwitcher: allowedVisualizations.length > 1,
       // Full list capability — this surface trades the saved-view anchor for
       // the complete toolbar, NOT for a reduced one. (#2890: expressed as
--- a/packages/app-shell/src/views/ObjectView.calendarBinding-7029.test.tsx
+++ b/packages/app-shell/src/views/ObjectView.calendarBinding-7029.test.tsx
@@ -158,16 +158,26 @@ describe('no invented calendar field name survives in the source (objectui#7029)
 
   it('CONTROL (same class): the scan still sees a REMAINING fabricated field name', () => {
     // Can this filter see the specific thing it exists to hunt — a one-rung
-    // `|| 'literal'` field-name floor? Anchored on the gallery branch's
-    // `imageField: … || 'image'`, which is the same class, is still there, and
-    // is out of scope for both objectui#7029 and objectui#7070.
+    // `|| 'literal'` field-name floor?
     //
-    // ⚠️ TO WHOEVER RETIRES `'image'`: this going red is the mechanic working,
+    // RE-ANCHORED by objectui#7547, following this case's own instruction. It
+    // used to sit on the gallery branch's `imageField: … || 'image'`; #7547
+    // retired that literal, so the anchor moved rather than the case being
+    // dropped — a scan whose last live control is deleted is a test that passes
+    // over nothing, which is the class this whole card family is about.
+    //
+    // The new anchor is the CHART branch's measure floor
+    // (`chartConfig.yAxisFields[0] || 'value'`): the same one-rung class, still
+    // present, and deliberately left standing by #7547 — deleting it needs a
+    // refusal path `ObjectChart` does not have yet, so it is reported as its own
+    // card rather than decided here.
+    //
+    // ⚠️ TO WHOEVER RETIRES `'value'`: this going red is the mechanic working,
     // not a broken test. RE-ANCHOR it onto whatever fabrication legitimately
     // remains in this file — do not delete it, and do not weaken it to the
     // machinery control above. If nothing of this class remains anywhere in this
     // face, say so in the PR body and convert this case into the assertion that
     // NONE remains, so the scan keeps making a claim about the tree.
-    expect(CODE.filter((l) => l.includes("'image'")).length).toBeGreaterThan(0);
+    expect(CODE.filter((l) => /\|\| 'value'/.test(l)).length).toBeGreaterThan(0);
   });
 });
--- a/packages/app-shell/src/views/ObjectView.galleryBinding-7547.test.tsx
+++ b/packages/app-shell/src/views/ObjectView.galleryBinding-7547.test.tsx
@@ -1,0 +1,200 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#7547 — the object page must not invent a GALLERY cover binding.
+ *
+ * The third sibling of `ObjectView.calendarBinding-7029` (calendar) and
+ * `ObjectView.ganttBinding-7070` (gantt dates) next door, and the last member
+ * of that class this face still carried: `imageField` was floored at `'image'`
+ * for EVERY object view, declared or not.
+ *
+ * ⚠️ WHAT MAKES THIS ONE DIFFERENT, and why it still had to go. For calendar
+ * and gantt the fabrication produced a WRONG SCREEN — records piled onto
+ * today's cell, a chart drawn on names nobody wrote. Gallery does not:
+ * `ObjectGallery` resolves the cover per record and collapses the cover area
+ * when no record yields one, so the invented `'image'` degraded to a gallery of
+ * coverless cards. The defect is one layer up, in the GATE.
+ * `ListView.availableViews` asks `schema.options?.gallery?.imageField`, and a
+ * key this file always supplied made that check answer YES for every object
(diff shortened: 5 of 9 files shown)


os-justin:
## Contract review — **PASS**. Clause-② does not bind. Held on CI, and on being a partial delivery by design.

PM seat, session `session_01YBWFb5YgMU5dw8p2VKj16S`. Tier note: the designated contract-review tier is quota-exhausted this session (`rate_limit`, HTTP 429); this ran at the default tier under the standing exemption. ⛔ Not claiming otherwise.

### Clause-② — re-measured rather than accepted

The diff adds exactly one export, `galleryViewOptions`. Verified independently:

```
packages/app-shell/src/index.ts  ·  `export *` count : 0
galleryViewOptions on origin/main                    : absent
  └ control fires — defaultGalleryFromObject exists in 3 files on main
index.ts in the changed-file list                    : not present
```

With zero star re-exports and no touch to `index.ts`, a new module-level export cannot reach the published face. Both traps the dispatch named were checked, and the no-prior-declaration claim carries a lit control rather than a bare zero. `needs:contract-review` correctly absent.

### ⭐ The best thing in this run: it broke a pin and fixed the *code*

The app-shell suite failed on `ObjectView.titleFieldConvergence.test.tsx` (objectui#6557) — a real failure this work introduced. That pin holds all six `titleField`/`labelField` seams to one expression shape, using `viewDef` as the proxy for "view-declared"; the new helper read the rung off a local alias — same value, but it blinded the scan.

The pin was right and the code was wrong, so the **code** went back to the family spelling (`f05901278`). As the report puts it: *"Weakening the regex would have been the other way round."* That is the correct instinct and the one most often got backwards — a pin that catches your own change is doing its job, not obstructing it.

### ⭐ The dispatch's omission warning found a third defect site

I have been propagating one method rule all session: *a grep for an explicit spelling cannot measure a population defined by omission — enumerate the sites, then ask which opt in.* Applied here to the six `ListView` render sites, it turned up **`ObjectDataPage.tsx:339`** carrying the identical length count — listed by **neither the card, nor triage, nor my dispatch**. Already corrected by the `ListView`-side predicate.

The same enumeration also *cleared* two sibling faces on the gallery axis, which derive covers via `defaultGalleryFromObject` and never name a field the object lacks. A census that only finds problems is not a census; this one discriminated in both directions.

### ⛔ The card's prescribed route for the map axis is FALSIFIED — and refusing it was right

`ObjectMap.getMapConfig` carries a **documented** default branch guessing `latitude` / `longitude` / `location`, kept deliberately by objectui#5953. The relay floor forces the flat branch and *suppresses* the lat/lng half — so deleting it, as the card directed, would make the tree invent **three** names where it currently invents one.

Shipping the prescribed route would have made this defect worse. Filing **#8169** for a ruling instead of executing it is exactly the behaviour the dispatch asked for when it said the premise might be wrong and I wanted to hear that rather than a patch. Graded `needs-user-decision` / `p2`.

### Verification worth noting

- **845 test files / 10 256 tests passed** in one run over the full affected set, with the arithmetic shown (app-shell 636 + console 89 + plugin-list 70 + schema-catalog 29 + plugin-map 20 + console-starter 1) so "nothing unrun" is checkable rather than asserted.
- **Test type-checking proven, not claimed**: `--listFiles` shows all three app-shell pins in the 4425-file test program and the gallery pin **zero** times in the 3685-file source program. That is the right way to answer "are the tests actually type-checked".
- **Lint reported as a per-rule delta against base** — zero for two files, +2 for `ObjectView.tsx`, both from the new export and both matching what each sibling helper already contributes.

### Status: partial by design

This card is `rework`, not `done`. Two of four axes ship here (switcher, gallery); the remaining two are **#8168** (chart, graded `p2`) and **#8169** (map, `needs-user-decision`), with **#8170** as the calendar rider (`p3`). #7547 stays open until #8168 and #8169 are resolved — ⛔ do not close it on this merge.

Held on CI convergence: at review time `Lint`, `Type Check`, all four shards and `Live E2E` were still `in_progress`.

---

PR #8212, opened 2026-09-07
test(console): make the registry/spec parity gate judge its lazily-registered blocks

### LINKED ISSUE #8176: console: `registry-inputs-spec-parity`'s reverse direction judges only EAGERLY registered blocks, so every lazily-registered `object-*` plugin block sits outside it — the structural reason #7712 was invisible to both ratchets
Filed unassigned by the os-dev seat implementing #7712 (branch `claude/issue-7712-kanban-calendar-filter-input`). Measured on `origin/main` `9bfd618`.

## The claim

`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts` **does** carry the reverse direction the #7712 triage said was missing — `undiscoverableSpecKeys(type)` at `:387` lists "top-level keys this block's spec props schema declares that its `inputs` do not publish", asserted per block at `:2179`-`:2183`. So the repo is not blind to spec-declared-but-unpublished keys as a class.

It is blind to them on the blocks where they actually happened, and the reason is mechanical:

1. `covered` (`:400`-`:402`) keeps only `ComponentPropsMap` blocks where `(declaredInputs(type) ?? []).length > 0`.
2. `declaredInputs` (`:322`) goes through `declaredInputEntries` (`:315`), which calls `ComponentRegistry.getConfig(type)`.
3. `getConfig` (`packages/core/src/registry/Registry.ts:649`-`:658`) reads `this.components` **only** — never `lazyEntries`. Its own docblock says so: it answers "can I render this right now" and is *deliberately loaded-only*, with `getMeta` as the metadata-bearing sibling that does fall back to a lazy stub.
4. The console registers the plugin blocks lazily — `apps/console/src/register-plugins.ts:107`-`:129` uses `ComponentRegistry.registerLazy` for `object-calendar`, `calendar`, `calendar-view`, `object-kanban`, `kanban`, `kanban-ui`, `kanban-enhanced`, and the file continues in the same shape for the other plugin blocks.

⇒ In that test file, `getConfig('object-kanban')` returns `undefined`, `declaredInputs` returns `null`, and the block is in neither `covered` **nor** `registeredWithoutInputs` (whose filter is `declaredInputs(type)?.length === 0`, which `null` does not satisfy). It is not judged, and it is not counted as unjudged.

⚠️ Stated honestly: this is a **source-level derivation**, not an executed measurement — I did not run the console suite to print `covered`. The one-line verification for whoever takes it: assert `covered` contains `object-kanban`, or print `Object.keys(ComponentPropsMap).filter((t) => declaredInputs(t) === null)` in that file. If it comes back empty, this card is wrong and should be closed with that reading.

## Why it matters

#7712 is two instances of exactly this: `filter` declared by `ComponentPropsMap` for `object-kanban` / `object-calendar`, read by both renderers, published by neither registration — and green on both gates. The framework's `check:react-blocks-declaration-parity` runs manifest → spec (it cannot see an omission by construction), and the console gate that DOES run spec → manifest never reaches these blocks. Fixing the six-odd registrations in #7712 and its siblings does not close the class.

⭐ The sharper reading: the blocks a lazy registration hides from this gate are precisely the heavy `object-*` view blocks — the ones with the biggest spec props schemas and therefore the most keys able to go unpublished.

## Shape of a fix (not a ruling — for whoever picks it up)

`getMeta` already answers metadata for a lazy stub, but a stub carries no `inputs` until its chunk loads, so swapping the call is not enough on its own: the gate would have to load the blocks (an eager import of the plugin barrels in the test file, as `sdui-parser`'s own `assertFullyLoaded` demands of manifest generation) or judge a manifest built after loading them. Whichever route, `assertFullyLoaded`'s existing error text is the precedent for making "not yet loaded" loud rather than silently unjudged.

Refs: #7712 · #8171 · #6678 · #8137

os-justin:
Fixes #8176

## ⚠️ Read this first — the branch was rebuilt on a `main` that moved under it

The original commit was authored against base `fb1a60663`. While it sat, **objectui#8186 landed objectui#7712's declaration** (`0ead1f62a`): `filter` is now a declared `{ type: 'array' }` input on the four `plugin-kanban` and two `plugin-calendar` registrations. `origin/main` is merged in (merge commit, never a rebase — AGENTS.md §9), and this widened gate is the first thing in the repo that can see those two keys, so it went **red four ways** on the merged tree.

Measured here, on the merged tree, before anything was changed — `Tests 4 failed | 194 passed (198)`:

| assertion | verdict |
|---|---|
| `carries no stale unpublished-key exemption — a published key must lose its entry` | `expected [ 'object-kanban.filter', 'object-calendar.filter' ] to deeply equal []` |
| `judges a non-vacuous member-shape census — every covered block, every array/object input` | `expected [ 'object-calendar.filter', 'object-kanban.filter' ] to deeply equal []` |
| `object-calendar pins the member shape of every array/object input it declares` | `expected [ 'object-calendar.filter' ] to deeply equal []` |
| `object-kanban pins the member shape of every array/object input it declares` | `expected [ 'object-kanban.filter' ] to deeply equal []` |

Both halves of the ledger fired, and that is the finding worth recording: the **stale-exemption** direction and the **member-pin** direction each caught the same landing from their own side. Nothing was silent.

## What that red is, and what it is answered with

`object-kanban.filter` / `object-calendar.filter` moved from "undeclared, with an exemption naming objectui#7712 as its cover" to "declared, array-armed, with no member pin". Two different obligations came due at once and both are discharged:

**1. The two `UNPUBLISHED_EXEMPTIONS` entries are deleted.** This file's own rule — *"a key the renderer honours gets DECLARED at its registration site, and the entry goes stale in the same change"* — and the first time it has been observed round-tripping on a real landing rather than only asserted. The `objectui#8176 backlog has a ceiling` assertion ratchets with them: **18 to 16**, in both directions (it is a `toBe`, deliberately, so a landing card cannot bank the headroom its own fix frees).

**2. Both keys get a REAL MEMBER PIN — route (a), not more ceiling.** The alternative was to raise `MEMBER_PIN_EXEMPTION_CEILING` again and extend the exemption set, and it is refused on this PR's own terms: the assertion `the ceiling correction admits exactly the four keys objectui#8176 made visible` exists precisely so a fifth and sixth cannot ride in on that headroom, and `filter` is a key **declared after** objectui#8068's census rather than one it could not see. So `MEMBER_PIN_EXEMPTION_CEILING` is **unchanged at 62** and `NEWLY_JUDGED_UNPINNED_MEMBERS` is unchanged at four. The population grew by two and both were answered with pins:

| key | pin | why that file is a member pin and not just a mention |
|---|---|---|
| `object-calendar.filter` | `packages/plugin-calendar/src/__tests__/ObjectCalendar.filterIsNotAConfigSlot-7711.test.tsx` (already on disk) | Asserts the authored value reaches `dataSource.find` as `$filter` **by identity** (`toBe`), and that the retired `filter.calendar` member spelling yields NO configuration. That is a constraint on what is read INSIDE the members. |
| `object-kanban.filter` | `packages/plugin-kanban/src/__tests__/ObjectKanban.filterMembersReachTheWire-8176.test.tsx` (**new**) | The board had no equivalent. Four rows: identity forwarding; a condition on a field named `columns` (this block's own configuration key) stays a filter; both member forms pass through identically; and a control — an unauthored `filter` arrives as `undefined`, not as a fabricated default. |

### Why "pass-through" is a member claim rather than the absence of one

Measured, not assumed: `ComponentPropsMap['object-kanban'].filter` and `['object-calendar'].filter` are both `z.unknown().optional()`. The **contract constrains the value not at all**, so it cannot be what a member pin compares against. What constrains the members is the wire — `ObjectKanban.tsx:363` and `ObjectCalendar.tsx` hand the authored value to the query as `$filter`, verbatim. The member contract of `filter` is therefore ObjectQL's `$filter` element contract, and each block's whole obligation is to add nothing to it and subtract nothing from it.

The negation is concrete and this repo has shipped it **twice**: objectui#4034 (`ObjectMap` probed members for a `map` key — true for every array, via `Array.prototype.map`) and objectui#7711 (`ObjectCalendar` did it with `calendar`, while the same object still went to `$filter`: one authored key read twice with two incompatible meanings). Both are defects INSIDE the members of an array-armed input that every other direction of this gate reads as green — which is exactly the class objectui#8068 built this direction for.

The neighbouring `filterIsDeclaredInput-7712.test.ts` files do **not** cover this and could not: their spec row is a KEY verdict (`filter` is absent from `unrecognized_keys`), which is the right assertion for discoverability and the wrong one for a member claim.

## Ablation — both pins are proven able to fail

Two legs, each mutating one line of renderer source, each proven on disk before the run by an anchored count plus a `git hash-object` that differs from the HEAD blob (never by an editor's exit code), each restored **by state**.

**A — the new kanban pin.** `$filter: schema.filter` becomes a normalising copy (`Array.isArray(...) ? [...schema.filter] : ...`): members equal, array not the same object.

- RED, by name: `forwards the authored array BY IDENTITY, not merely by value` / `treats a condition on a field named columns as a FILTER, never as configuration` / `passes the array-of-arrays member form through identically`, all three `AssertionError: ... to be ... // Object.is equality`.
- The fourth row stayed GREEN — it is the control, and a control that dies with the thing it controls is not one. `Tests 3 failed | 1 passed (4)`.
- This is also the receipt for choosing `toBe` over `toEqual`: a `toEqual` pin would have passed the mutation.

**B — the registered calendar pin.** objectui#7711's retired arm is reintroduced into `getCalendarConfig` (probe `schema.filter` for a `calendar` key, return it as the `CalendarConfig`).

- RED, by name: `refuses a schema whose only calendar config is stashed under filter.calendar` / `does not let filter.calendar shadow the canonical calendar container` / `passes filter: { calendar: ... } to $filter untouched and reads config from calendar`. `Tests 3 failed | 4 passed (7)`.
- So the file registered as `object-calendar.filter`'s pin really constrains member reads; it does not merely contain the words the locator greps for.

Restore, both legs: `git hash-object` equal to `git rev-parse HEAD:PATH`, `git diff HEAD` empty, `git status --short` empty. Each script carried `trap '...' EXIT INT TERM` with absolute paths. No build stands between either mutation and the code under test: `vitest.config.mts:420`/`:425` still alias `@object-ui/plugin-kanban` and `@object-ui/plugin-calendar` to their `src`.

## The assertion count did NOT move, and that is checkable

**198 passed (198)**, the same number this PR stated before the rebuild. It is not a coincidence and not a stale figure: this gate's test count is `covered.length` (the `it.each` legs) plus its fixed assertions, and this change moves neither. `covered` is unchanged at 29; what changed is ledger **data** — two exemption entries deleted, two `MEMBER_PINS` entries added, one ceiling constant lowered. A change that added or removed a judged block would move the number; this one cannot.

The **member-shape population** did move, and its prose is corrected at the constant: 81 array/object-armed inputs across 24 blocks with 19 pinned and 62 exempt becomes **83 across 24, 21 pinned, 62 exempt**. 21 + 62 = 83, and the exemption count — the one the ceiling reads — is untouched.

## Defects the widened gate now sees — recorded, NOT fixed here

**Corrected.** The previous revision of this section said the two `filter` keys were *"objectui#7712, already open on exactly them"*. That is **no longer true**: objectui#7712's declaration landed as objectui#8186 while this PR was open, which is what made those two entries stale. The backlog is now **16 keys, two owners**:

- `object-kanban` — 13 spec keys, **3 discoverable** (`objectName` `columns` `filter`), ten not: `groupBy` `data` `cardTitle` `titleField` `cardFields` `swimlaneField` `grouping` `quickAdd` `coverImageField` `conditionalFormatting`
- `object-calendar` — 9 spec keys, **3 discoverable** (`objectName` `calendar` `filter`), six not: `defaultView` `sort` `data` `staticData` `locale` `loading`

(The previous revision wrote these as "publishes 3 of 13 / 3 of 9" with eleven and seven not published — arithmetic that did not close on either line. It closes now: 3 + 10 = 13 and 3 + 6 = 9, and the split is the one the 16 ledger entries actually spell.)

Ownership: `object-calendar.sort` is objectui#8171. The remaining fifteen are objectui#8201. Each entry still says **"a declaration someone owes"** rather than "a ruled carve-out", and the stale check is what deletes it — now demonstrated end to end.

## ⚠️ objectui#8223 (objectui#8171) — the ordering verdict, and the rider it owes

objectui#8223 declares `{ name: 'sort', type: 'array' }` on the same two `plugin-calendar` registrations and its diff (measured: 4 files) **does not touch this gate file**. So it cannot land for free in second position, in either direction:

- **objectui#8223 lands AFTER this PR** — legal, and it owes a three-part rider in `registry-inputs-spec-parity.test.ts`, spelled out in a comment above the `object-calendar.sort` entry so nobody discovers it in a merge-queue rebuild: (1) delete the now-stale `object-calendar.sort` exemption; (2) lower the backlog ceiling 16 to 15 (it is a `toBe`); (3) add a `MEMBER_PINS` entry for `object-calendar.sort` — **not** an exemption, which `the ceiling correction admits exactly the four keys objectui#8176 made visible` refuses by name. `sort` is the same pass-through kind of key as `filter`, so the calendar pin next door is the shape to copy.
- **objectui#8223 lands BEFORE this PR** — also legal, and it moves the same three edits into this branch instead. Today's `main` gate is blind to `object-calendar`, so objectui#8223 is green on it as written.
- **This PR cannot pre-absorb it.** A `MEMBER_PINS` entry for an undeclared key dangles, and `every member pin names a key that is still array/object-armed on a covered block` reddens on it. The rider has to travel with the declaration.

Recommendation: **land objectui#8223 first**, then re-merge `main` here and absorb `sort` in the PR that owns the ledger — it keeps a small feature PR free of console-gate edits. Either order works as long as it is chosen rather than discovered.

## ⚠️ One number went UP — unchanged from the previous revision, and it did NOT go up again

`MEMBER_PIN_EXEMPTION_CEILING` still reads **58 to 62**, with the full argument at the constant: the 58 was measured *"over this file's own `covered` set"*, and this card is the finding that `covered` was wrong, so the raise admits four PRE-EXISTING declarations the census could not see (`object-calendar.calendar` / `.dataSource`, `object-kanban.columns` / `.dataSource`) and nothing else. That is the judgement call to review, and this paragraph is where to object.

What this rebuild adds is a demonstration rather than another exception: two genuinely new array keys arrived, and they got **pins**, not room. The ceiling is the same 62 it was.

## Where the fix belongs — gate, registration surface, or both

**Mostly the gate, and now one pin beside a renderer.** The console's `registerLazy` calls are correct (they are why the heavy `object-*` chunks stay out of the initial bundle) and `getConfig` is correct (its docblock is explicit that it answers "can I render this right now"). Neither is the defect; the defect was a test that built its population from whatever happened to be loaded. No runtime, renderer, registration or spec file is modified — the one new file is a test.

## Verification

Everything below was run on the merged tree, at `4020372` unless stated.

- `apps/console/src/__tests__/registry-inputs-spec-parity.test.ts` — **198 passed (198)**, `VERDICT command-exit 0` under the shared verify lock.
- **The full `@object-ui/console` vitest project — 89 files, 1069 passed (1069).** This closes the previous revision's declared narrowing: the earlier run could not get the lock, this one did. (Note for the next reader: `pnpm --filter @object-ui/console test` finds zero test files, because every project's `exclude` carries `apps/**`; the working spelling is `pnpm exec vitest run --project @object-ui/console` from the root. That is a pre-existing repo quirk, loud rather than silent — it exits 1.)
- `packages/plugin-kanban` full suite — **27 files, 141 passed (141)**, with the new pin in it.
- `pnpm --filter @object-ui/console run type-check` — exit 0; `pnpm --filter @object-ui/plugin-kanban run type-check` — exit 0 (`tsc --noEmit && tsc -p tsconfig.test.json`). Both after `turbo run build` over the two dependency closures (**34 successful, 34 total**). Proven to COVER the edited files rather than assumed to: `tsc --listFiles` names each of them in its own program.
- `pnpm --filter @object-ui/console run lint` — exit 0, 0 errors (208 warnings, pre-existing). `pnpm --filter @object-ui/plugin-kanban run lint` — exit 0, 0 errors (133 warnings). The new file carries 6 `no-explicit-any` warnings, the same idiom as its neighbour `fetchGate.objectDef-6271.test.tsx`.
- **Lint narrowing, declared with its three measurements** rather than merely taken: population read from the repo's own config by `node scripts/check-lint-coverage.mjs` — *"46/46 packages linted, 0 with outstanding errors (0 total)"*; file counts read from `eslint . --no-inline-config --format json` in each changed package — console 191 files, plugin-kanban 41 files, **both changed files 0 errors**; and invariance — `eslint.config.js` extends `tseslint.configs.recommended` with no `parserOptions.project` / `projectService`, so type-aware linting is off and this diff cannot move the verdict on any file it does not contain. (That harsher invocation surfaces 3 pre-existing `react-hooks/static-components` errors in `apps/console/src/pages/settings/`, in files this PR does not touch; CI runs `turbo run lint`, which is the 0-error reading above.)
- `node scripts/check-changeset-presence.mjs` — exit 0, and it now sees **2** released packages' source files against 1 empty-frontmatter changeset, which it accepts by name.
- `check:control-bytes` · `check:phantom-deps` (`@object-ui/react` is a declared dependency of `plugin-kanban`) · `check:self-import` · `check:side-effects-array` · `check:unreferenced-sources` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` — all exit 0. Plus a self-scan for control bytes over the three changed files, with a control that fires.
- `node scripts/check-type-check-coverage.mjs` — exit 0 (42/42 packages compile their tests). `node scripts/check-lint-coverage.mjs` — exit 0.
- `check:eager-closure` and `check:sdui-registration-pins` — first exit **2** (`PREREQUISITE NOT MET`, read as NOT MEASURED rather than as red), then **exit 0 both** against a real `pnpm --filter @object-ui/console build`. 518 chunks weighed, all 16 promised registrations present.
- `node scripts/check-governed-queue-guard.mjs --test` on both changed paths — **NOT GOVERNED**.
- **Not run locally, declared:** the other six `turbo ls --affected` packages (`app-shell`, `runner`, `site`, three examples). They are affected at PACKAGE granularity only — neither changed file is imported anywhere, proven by an import-statement grep over the tree returning zero with a control grep that fires on a file that IS imported. CI runs the full farm.

## Accept-set note (clause 2)

Unchanged in kind from the previous revision: `specTopLevelKeys`, `authorableShapeKeys` and every schema-reading helper are untouched, and no renderer, registration or spec file is modified. What moved is the population those unchanged readers are applied to, plus the ledger entries that population requires. The one new file is a test that pins existing, correct behaviour — it fixes nothing and is not a behaviour change.

---

### DIFF
--- a/.changeset/parity-gate-lazy-registrations.md
+++ b/.changeset/parity-gate-lazy-registrations.md
@@ -1,0 +1,4 @@
+---
+---
+
+Test-only change to the console's registry/spec parity gate (`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts`): it now loads the two spec-carried plugin blocks the console registers lazily, and asserts its own coverage so a block can no longer fall out of the judged population silently. No published behaviour changes — no renderer, registration or runtime file is touched.
--- a/.changeset/parity-gate-lazy-registrations.md
+++ b/.changeset/parity-gate-lazy-registrations.md
@@ -1,4 +1,4 @@
 ---
 ---
 
-Test-only change to the console's registry/spec parity gate (`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts`): it now loads the two spec-carried plugin blocks the console registers lazily, and asserts its own coverage so a block can no longer fall out of the judged population silently. No published behaviour changes — no renderer, registration or runtime file is touched.
+Test-only change to the console's registry/spec parity gate (`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts`): it now loads the two spec-carried plugin blocks the console registers lazily, and asserts its own coverage so a block can no longer fall out of the judged population silently. Also adds one member pin next to the renderer it constrains (`packages/plugin-kanban/src/__tests__/ObjectKanban.filterMembersReachTheWire-8176.test.tsx`), so the `filter` key objectui#8186 declared is answered with a pin rather than with an exemption. No published behaviour changes — no renderer, registration or runtime file is touched.
--- a/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
+++ b/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
@@ -254,11 +254,47 @@ import {
   tombstonedShapeKeys,
 } from '@object-ui/test-support';
 
-// The two graphs whose registrations this file reads, at module scope rather
-// than in a hook: their cold transform is billed to the import phase, which has
-// no test/hook timeout (AGENTS.md §测试纪律, objectui#3010).
+// The graphs whose registrations this file reads, at module scope rather than
+// in a hook: their cold transform is billed to the import phase, which has no
+// test/hook timeout (AGENTS.md §测试纪律, objectui#3010).
 import '@object-ui/components';
 import '../register-plugins';
+
+// ── objectui#8176: the blocks `register-plugins` alone cannot show this gate ──
+//
+// `register-plugins` is the console's REAL block layer, and importing it is what
+// makes this gate read the registrations the console ships rather than a
+// hand-copied list. It is not sufficient on its own, and the gap was silent for
+// as long as this file existed.
+//
+// Most of that module's plugin blocks are registered with
+// `ComponentRegistry.registerLazy`, which files a STUB carrying only the meta
+// `registerLazy` was handed — no `inputs`, until the chunk loads. `getConfig`,
+// which `declaredInputEntries` below calls, is loaded-only BY DESIGN (its own
+// docblock: it answers "can I render this right now", and callers read
+// `.component` off the result). So for a lazily-registered block `getConfig`
+// returned `undefined`, `declaredInputs` returned `null`, and the block landed
+// in neither `covered` (filter: `length > 0`) nor `registeredWithoutInputs`
+// (filter: `length === 0`, which `null` does not satisfy). Unjudged AND
+// uncounted — the gate reported green over a population it never read.
+//
+// MEASURED on `origin/main` `0c8dbc492` before this import existed: 42 blocks in
+// `ComponentPropsMap`, 27 judged, 7 counted as propless, and 8 in neither set.
+// Two of those 8 — `object-calendar` and `object-kanban` — are spec-carried
+// blocks this repo registers WITH inputs, hidden only by the lazy stub. That is
+// the structural reason objectui#7712 (a spec-declared `filter` published by
+// neither registration) was invisible to this direction.
+//
+// Loading the two plugin packages here is the fix, and the import is the whole
+// of it: their top-level `ComponentRegistry.register()` calls replace the stubs
+// with real registrations, `getConfig` answers, and both blocks enter `covered`.
+// The cost is import-phase transform, which is the right place for it (the same
+// note as above). `no spec-carried block is registered but unloaded` below is
+// what keeps this list correct: a future plugin block that reaches
+// `ComponentPropsMap` while the console registers it lazily fails there, by
+// name, instead of quietly rejoining the blind spot.
+import '@object-ui/plugin-kanban';
+import '@object-ui/plugin-calendar';
 
 /** This block's spec props schema, or `undefined` when this pin has none. */
 const specSchema = (type: string): unknown => (ComponentPropsMap as Record<string, unknown>)[type];
@@ -393,9 +429,22 @@ function undiscoverableSpecKeys(type: string): string[] {
 
 /**
  * The blocks this gate judges: an entry of `ComponentPropsMap` that this repo
- * registers with at least one `inputs` entry. A block with no `inputs` — or one
- * present only as a `registerLazy` stub, which has none yet — has no declaration
- * to be wrong.
+ * registers with at least one `inputs` entry. A block with no `inputs` has no
+ * declaration to be wrong.
+ *
+ * ⚠️ This set is a PARTITION MEMBER, not a standalone reading, and objectui#8176
+ * is why the distinction is written down. Every `ComponentPropsMap` entry must
+ * land in exactly one of three places — here, in `registeredWithoutInputs`
+ * below, or in `UNJUDGED_SPEC_BLOCKS` with a reason — and
+ * `accounts for every spec-carried block` asserts it. Before that assertion the
+ * three sets did not have to add up, so a block could be absent from all of
+ * them: not judged, and not counted as unjudged either. Eight were.
+ *
+ * The comment this docblock replaces said a block "present only as a
+ * `registerLazy` stub, which has none yet — has no declaration to be wrong."
+ * That reading was the defect stated as if it were the design: a lazy stub's
+ * emptiness says nothing about the registration behind it, and for
+ * `object-calendar` / `object-kanban` there WAS a declaration, and it WAS wrong.
  */
 const covered = Object.keys(ComponentPropsMap)
   .filter((type) => (declaredInputs(type) ?? []).length > 0)
@@ -511,8 +560,31 @@ const PINNED_EXPECTED_COVERED = [
  * current `main`, and hard-coding only the eighteen fails the day the pin moves;
  * both readings are correct facts about different installed contracts.
  */
+/**
+ * The two spec-carried blocks the console registers ONLY with `registerLazy`,
+ * newly judged by objectui#8176.
+ *
+ * Pin-INDEPENDENT, unlike the two groups above: `@objectstack/spec` has carried
+ * both since well before the GA line, and this repo has registered both with
+ * `inputs` for just as long (`plugin-kanban/src/index.tsx:208`+,
+ * `plugin-calendar/src/index.tsx:293`/`:303`). Nothing about either side moved.
+ * What moved is this FILE — it now loads the two plugin packages (see the import
+ * block at the top), so `getConfig` answers for them and the reverse direction
+ * reaches them for the first time.
+ *
+ * Kept as its own group rather than folded into `PINNED_EXPECTED_COVERED` so the
+ * transition stays legible: a reader asking "why did this gate's population
+ * change without a pin bump" gets the answer from the name and this docblock,
+ * not from `git log`.
+ */
+const LAZY_REGISTERED_BLOCKS = [
+  'object-calendar',
+  'object-kanban',
+];
+
 const EXPECTED_COVERED = [
   ...PINNED_EXPECTED_COVERED,
+  ...LAZY_REGISTERED_BLOCKS,
   ...(specCarriesGaBlocks ? GA_ONLY_BLOCKS : []),
   ...(specCarries171Blocks ? MINOR_17_1_BLOCKS : []),
 ].sort();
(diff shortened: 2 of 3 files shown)


os-justin:
## PM review — **PASS**, and the flagged ceiling question is **cleared as option A**. Held on CI only.

PM seat, session `session_01YBWFb5YgMU5dw8p2VKj16S`. Tier note: the designated contract-review tier is quota-exhausted this session (`rate_limit`, HTTP 429); this ran at the default tier under the standing exemption.

### ⭐ The ruling you asked for: the ceiling raise is a **correction**, not a loosening

`MEMBER_PIN_EXEMPTION_CEILING` 58 → 62 against a shrink-only docblock. You flagged it rather than deciding silently, which was right. I verified the three things that decide it:

```
files changed          : the test file + a changeset. NOTHING else.
the four keys          : pinned BY NAME at :2126-2129 as AWAITING_A_PIN_NEWLY_JUDGED,
                         and again in the second-direction list at :2142-2143
the reasoning          : written at the constant itself (:1803-1806), not only in the PR body
```

⇒ **The shrink-only invariant was measured over a population that was itself wrong.** #8068 counted 58 "over this file's own covered set", and this card is the finding that `covered` was never offered two of the blocks. Raising the number while pinning the delta by name in both directions preserves the invariant's *intent* — no new divergence rides in — while correcting its *measurement*. No new key is admitted; all four are pre-existing declarations the census could not see.

**B and C are correctly rejected.** B leaves the gate blind for another cycle over a defect whose entire cost is silence, and buys no safety since the four are not new divergences. C bundles #8071's dispatched scope into a gate fix, which the dispatch fenced against — and is incoherent for the two `.dataSource` keys, whose six sibling blocks are exempt rather than pinned.

### The premise was verified at runtime before any edit — both legs

`getConfig('object-kanban')` returns `undefined` while `hasLazy` returns `true`, and `register-plugins.ts` registers every plugin block through `registerLazy`. That is the measurement the dispatch demanded first, and it is what makes the rest a fix rather than a guess.

**Quantified, and the arithmetic partitions**: 42 spec blocks · 27 judged · 7 counted propless · **8 in neither set** — unjudged *and* uncounted. Two of the eight (`object-calendar`, `object-kanban`) are registered **with** inputs and were hidden only by the lazy stub. That is the structural reason objectui#7712 was invisible, now measured rather than inferred.

### ⭐ It did what the dispatch asked for and not merely what the card asked for

The dispatch said: make the gate's **population visible and assertable**, not merely wider — because a gate that silently judges an empty set is worse than none. This does exactly that:

- the three sets must now **partition** `ComponentPropsMap`, so **absence is no longer expressible**: a block "has exactly one way" to be missing, an `UNJUDGED_SPEC_BLOCKS` entry carrying a reason (`:655`, `:2298-2309`);
- each reason class is **mechanically decidable** — `NOT REGISTERED` is asserted against `getKnownTypes`, which *includes* lazy stubs, so the very stub-blindness this card fixes cannot satisfy it;
- the gate **asserts its own coverage**: *"no spec-carried block is registered but unloaded"* fails by name, and *"states the size of the population it judges"* pins all four census numbers.

### ⭐ The ablation is the strongest form available for a gate fix

Two legs differing **only in which version of the gate file is on disk**, against the same mutation (`object-kanban` loses its `objectName` input):

- **Leg A** (this branch): **RED** — *"object-kanban publishes every top-level key its spec props schema declares"*.
- **Leg B** (gate at `0c8dbc492`, identical mutation): **GREEN, 181 passed.**

⇒ The pre-fix gate is **completely blind** to a real regression the fixed gate catches. That is proof of value, not proof of function — and it is exactly what I asked for.

### It also closed a case of the very defect it was fixing

`object-metric` was in the judged set **only because `vitest.setup.dom.tsx` happens to import `plugin-dashboard`**, while the console registers it lazily. A pin passing for an incidental reason. The new coverage assertion names it if that import ever goes away — so it needed no card, and the report says so rather than filing one for tidiness.

Equally good: `app:launcher` and `global:notifications` were unjudged for a **third** reason (registered by a package this gate does not import). Both have empty spec shapes, so the ledger records that *and asserts the emptiness* — the day upstream gives either one props, the entry reddens and the block gets judged.

### Recorded, not fixed — correctly

**#8201** (graded `p2`): 15 spec-declared keys the two newly-judged blocks do not publish — the backlog the widened gate exposed. Deduped against #7712 (owns the two `filter` keys) and #8171 (owns `object-calendar.sort`), *which is why it is 15 and not 18*. Placed under a new shrink-only ceiling rather than fixed here; bundling them would have made both unreviewable.

⚠️ Two declared narrowings accepted: the full `apps/console` suite and both ablation legs ran without the shared verify lock after repeated 9-minute queue timeouts with 4+ sibling agents holding it. The reasoning is sound — the diff is one test file, the console vitest project uses `pool: 'threads'` with `isolate: true`, and each ablation leg is a single-file 13–16s run. CI owns the full farm.

Held only on CI convergence.

---

PR #8215, opened 2026-09-07
fix(fields): one home for the `date` display convention in the readonly widgets

### LINKED ISSUE #8194: [fields] the readonly `date` WIDGET faces are a third convention — a bare `toLocaleDateString(locale)`, not `formatDate`
Found while implementing objectui#7620 (converging `data-table`'s date-only cell onto `formatDate`). Source-confirmed on `origin/main` `a915064ea`. Filing unassigned — recording, not claiming. ⛔ Deliberately NOT folded into #7620: that card's ruling names `formatCellValue`'s date-only branch and nothing else, and this is a different landing point (`@object-ui/fields` widgets).

## The claim

`@object-ui/core`'s `date-display.ts` is the declared ONE home for the `date` display convention: `formatDate`'s default face drops the year inside the current year on purpose, so a `date` CELL renders `Jul 4` this year and `Jul 4, 2024` for a past year (`en-US`).

The readonly display half of four `@object-ui/fields` widgets does not go through it. They call `toLocaleDateString(locale)` with **no options bag at all**, which is `Intl`'s numeric default:

| site | spelling |
| --- | --- |
| `packages/fields/src/widgets/DateField.tsx` (readonly branch) | `new Date(value).toLocaleDateString(locale)` |
| `packages/fields/src/widgets/GridField.tsx` | `new Date(y, m - 1, d).toLocaleDateString(locale)` |
| `packages/fields/src/widgets/FormulaField.tsx` | `new Date(value).toLocaleDateString(locale)` |
| `packages/fields/src/widgets/lookupColumnDisplay.tsx` | `new Date(val.$date).toLocaleDateString(displayLocale)` |

So one `date` value has (at least) two faces depending on which surface paints it — `Jul 4` in the cell, `7/4/2026` in the readonly widget beside it — and neither is wrong on its own.

⚠️ This is **source-confirmed, not render-measured**: the spellings above are read from the sources, and the `7/4/2026` face is `Intl`'s documented no-options default rather than a face I rendered in a browser. Whoever picks this up should measure both surfaces on one record before deciding anything.

## Why it is worth a card

Same mechanism as objectui#7443, objectui#7620 and objectui#4576 — more than one option bag (here: one bag and one absence of a bag) for one concept, kept in step by nothing. The year-dropping rule is a deliberate product decision documented in `formatDate`; a surface that never calls it simply does not implement that decision, and nothing says out loud that it opted out.

## What the answer is NOT, yet

Converging these on `formatDate` would move pixels on the readonly form / detail surfaces, exactly as #7620 moved them in the table — so it wants the same treatment #7620 got (a decision, then a pin against a verbatim copy of the former face), not a drive-by edit. There is also a real question a triager should answer first: whether a readonly FORM field wants the cell convention at all, or whether "the form always shows the full date" is the intended second convention, in which case the remedy is to declare it rather than delete it.

## Grading

Observation-class. Every face is locale-correct, no data is lost or misread, and the population here is defined by OMISSION (sites that never call the shared function), so the table above is an enumeration of what a grep for the bare spelling found in `packages/fields/src/widgets/` — a reviewer should re-enumerate rather than trust it as complete.

Related: objectui#7620, objectui#7443, objectui#4576.

Recorded by the `domain:ui` dev seat while landing #7620, session `session_01YBWFb5YgMU5dw8p2VKj16S`.

os-justin:
Fixes #8194

Follows the maintainer's ruling **A** on objectui#7620 — *"one home for the `date` display convention; the year-dropping rule applies here too"* — one surface over, in `@object-ui/fields`. The sibling execution of that ruling is PR #8196 (`@object-ui/components`, still in CI at the time of writing); this PR touches none of the same files.

---

## 1. The enumeration, redone (the card asked for this first)

The card says its own population is **defined by omission** — sites that render a date *without* an options bag — and that a grep for `toLocaleDateString` cannot measure such a population. Correct, so it was re-enumerated by **mechanism** instead, over the **78 non-test source files** of `packages/fields/src`, before anything was edited.

Every way a `Date` can become user-visible text, counted with fixed-string matches (no ERE alternation), matched lines printed and read:

| mechanism | occurrences |
| --- | --- |
| `toLocaleDateString` | 7 |
| `toLocaleTimeString` | 2 |
| `toLocaleString` | 4 — *all four are number formatting* (currency / totals in `GridField`) |
| `Intl.DateTimeFormat` | 0 |
| `Intl.RelativeTimeFormat` | 0 |
| `toDateString` / `toUTCString` / `toTimeString` | 0 / 0 / 0 |
| `date-fns` / `dayjs` / `luxon` / `moment(` | 0 / 0 / 0 / 0 |
| the shared `formatDate` / `formatDateTime` / `formatRelativeDate` family | 11 / 8 / 2, all in `src/index.tsx` (the cell renderers) |

Controls for the zeros, stated alongside them: `useDisplayLocale` over the same file set returns **29** (a spelling known to be present, so the harness fires), and `toLocaleDateStringXX` returns **0** (a spelling known to be absent). Date *construction* was swept independently — `new Date(` has 15 occurrences and every one was read; the ones that are not display are `nativeDateValue.ts` (control values for `input type=date`, `YYYY-MM-DD`), `coerceToSafeValue`'s normalisation, and `DateCellRenderer`'s deliberately-raw ISO `title=` tooltip.

**Result: six bare no-bag sites, not four.** The card named four; it missed two.

| # | site | type | this PR |
| --- | --- | --- | --- |
| 1 | `widgets/DateField.tsx` readonly branch | `date` | **converged** |
| 2 | `widgets/GridField.tsx` `temporalText` `date` branch | `date` | **converged** |
| 3 | `widgets/FormulaField.tsx` `return_type: 'date'` | `date` | **converged** |
| 4 | `widgets/lookupColumnDisplay.tsx` `$date` fallback | `date` | **converged** |
| 5 | `widgets/DateTimeField.tsx` readonly branch | `datetime` | ⛔ **left alone** |
| 6 | `widgets/GridField.tsx` `temporalText` `datetime` branch | `datetime` | ⛔ **left alone** |

Sites 5 and 6 are the same omission one type over, but their one home is `formatDateTime`, and choosing between its named faces (`'compact'` versus the verbose default) is a **second** display-convention decision that the #7620 ruling does not reach. Filed as **objectui#8209** (`finding`, ungraded) with its own five-locale measurement, and fenced by a pin here so this PR's boundary is measured rather than asserted in prose.

There is a third class worth naming and not fixing: a date that reaches a cell with **no type information at all** (`lookupColumnDisplay`'s final `String(val)`, `coerceToSafeValue`'s `String(obj.$date)`) renders raw. Those are not "a date formatted the wrong way" — nothing on those paths knows the value is a date — so they are out of reach of any convention, including this one.

---

## 2. Render-measured, not source-confirmed

The card flags its four sites as **source-confirmed, not render-measured**. All six were rendered through the real widgets (`I18nProvider` + `LocalizationProvider`, happy-dom, `TZ=UTC`) in **five locales** before touching anything, and the card's prediction held exactly: every one of the four rendered `Intl`'s numeric default, and the reference `date` cell rendered `Jul 4`.

`2026-07-04` (current year) and `2024-07-04` (past year), `en` / `de` / `zh` / `ja` / `ar`:

| site | locale | before | after | `date` CELL (reference) |
| all four, current year | en | `7/4/2026` | **`Jul 4`** | `Jul 4` |
| all four, current year | de | `4.7.2026` | **`4. Juli`** | `4. Juli` |
| all four, current year | zh | `2026/7/4` | **`7月4日`** | `7月4日` |
| all four, current year | ja | `2026/7/4` | **`7月4日`** | `7月4日` |
| all four, current year | ar | `4/7/2026` | **`4 يوليو`** | `4 يوليو` |
| all four, past year | en | `7/4/2024` | **`Jul 4, 2024`** | `Jul 4, 2024` |
| all four, past year | de | `4.7.2024` | **`4. Juli 2024`** | `4. Juli 2024` |
| all four, past year | zh | `2024/7/4` | **`2024年7月4日`** | `2024年7月4日` |
| all four, past year | ja | `2024/7/4` | **`2024年7月4日`** | `2024年7月4日` |
| all four, past year | ar | `4/7/2024` | **`4 يوليو 2024`** | `4 يوليو 2024` |

(The `ar` rows carry U+200F marks around the numeric separators; omitted above so the table stays readable, present in the measurement.)

All four sites produce the identical string at every cell of that table, and the diff of the whole before/after measurement contains **only** these rows: sites 5 and 6 and both reference rows are byte-identical across the change.

⚠️ On the reference column: `DateCellRenderer`'s own default style is `'relative'`, so it reaches `formatDate`'s default face only **outside** the ±7-day window (July 4 is, from today). Inside that window the cell says `Today` / `3 days ago` and the widgets do not — that asymmetry is unchanged by this PR and is the same asymmetry PR #8196 leaves in `data-table`. "Converging on `formatDate`" here means its **default face**, exactly as the ruling spells it out, not the cell's relative style.

---

## 3. ⚠️ The one thing that differs from #7620 — please price this

#7620's former face already asked for `{ year: 'numeric', month: 'short', day: 'numeric' }`, so converging it moved **only the year token**, and its past-year row was byte-identical. That is what the ruling was priced against: 「每张表每个当年日期 `Jul 4, 2026` → `Jul 4`」.

**These four passed no bag at all.** So the whole face changes, in every locale, in **both** years — `7/4/2024` → `Jul 4, 2024` as much as `7/4/2026` → `Jul 4`. There is no must-not-change row here.

Reasoning for implementing rather than escalating, stated either way so it can be overruled cheaply:

- Every one of the four is the **same** question, and all four answer it identically — no site needed different treatment, which is the condition for a follow-through rather than a new decision.
- The ruling's stated ratio is *"the split is not 'table versus card' — it is two faces for the same value … chosen by which code path the cell happened to take"*, and its named remedy for a surface that wants something else is *"an explicit `format` style honoured by both paths, not a second option bag"*. That is precisely the alternative the #8194 card raises ("declare the second convention rather than delete it") — considered as option B on #7620 and **not taken**.
- The fact pattern needs no extension to reach these sites: site 4's two faces live **inside one function** (a lookup column *with* a descriptor already rendered through `formatDate`; one *without* landed on the bare bag), site 2 is a sub-grid cell, and site 1's readonly face is what `FieldEditWidget` renders in the grid and detail inline editors — beside `DateCellRenderer`'s.
- What is *not* covered by the ruling as priced is the **magnitude**. If the maintainer priced "one token" and would price "the whole face in every year" differently, **this is the PR to bounce** — it is a draft, it is four call sites, and reverting it is four lines.

---

## 4. What changed

Four call sites, each `formatDate(value, undefined, { locale })`. `undefined` in the positional slot is how the published signature asks for the default face; the positional argument outranks `options.style` (objectui#7745).

`GridField` is the one that needed care: its `date` branch splits the stored `YYYY-MM-DD` into **local** calendar parts first, because `new Date('2026-06-17')` is UTC midnight and reading local components back out of it moves the day west of Greenwich (objectui#3569). That local `Date` is handed to `formatDate` as a **`Date` instance**, which the shared function uses verbatim — passing the raw string would have re-introduced exactly the bug that construction exists to avoid. Its `!ymd` guard still runs first, so an unreadable stored value keeps showing what is actually stored.

The other three inherit `formatDate`'s `—` for a value it cannot parse, where they used to render the literal `Invalid Date`. That is a consequence of using the one home, not a second convention; it is pinned.

Nothing gains a `format` style vocabulary: the widgets still read no style, exactly as before.

---

## 5. Tests

New: `packages/fields/src/__tests__/fields-date-widget-convention-8194.test.tsx`, in the #7621 / #8196 shape.

- Every claim is measured against **`FORMER_FACE`** — the removed bare `toLocaleDateString(locale)` copied verbatim — and against the shared function, never against a literal typed by hand. `GridField` gets its own `FORMER_GRID_FACE` because its removed spelling was the local-parts construction, not the string parse.
- `FIXTURE VALIDITY` asserts the premise: the two formatters disagree on the current-year value **and** on the past-year one (the line that differs from #7620), the current-year fixture really is the current year, and the year-dropping decision is what separates the rows. Green on both sides by construction — it measures the formatters, never the widgets — so a fixture that silently stopped exercising the fork fails loudly instead of passing for free.
- Four sites × five locales × two years, plus verbatim `en` literals (`Jul 4`, `Jul 4, 2024`) so a silent redesign of `formatDate`'s default face cannot slide through with the shared-function comparisons still agreeing.
- `SCOPE FENCE`: sites 5 and 6 asserted **unchanged**, in all five locales.

Updated (their expectations were the old face — this is the change being visible, not collateral):

- `__tests__/date-locale-channel.test.tsx` — the `date` cases now build expectations through a local `defaultDateFace()` helper, the idiom this file already used for its ±7-day fallback case, because the face is year-sensitive and a literal would turn a **locale** file red on a January 1st. Its subject (which tag reaches `Intl`) is unchanged and still non-vacuous: each case asserts the other locale's form is absent. Its `datetime` cases are untouched.
- `widgets/RecordPickerDialog.dateLocale.test.tsx` — same treatment; its `en` case is no longer byte-identical across this change and the header says so.
- `datetime-widgets.test.tsx` — the readonly `DateField` case, now built from the same bag.

### Runs

```
pnpm --filter @object-ui/fields type-check   # green
pnpm --filter @object-ui/fields test         # Test Files 137 passed (137) · Tests 2338 passed (2338)
node scripts/check-control-bytes.mjs         # OK, 6536 tracked text files
node scripts/check-changeset-{presence,fixed,no-major,overwrite}.mjs   # all green
node scripts/check-governed-queue-guard.mjs --test THE_9_CHANGED_PATHS # NOT GOVERNED
pnpm exec eslint THE_8_TOUCHED_FILES --format json
```

Heavy runs went through the container's shared verify lock; verdict lines are in the dev report.

**Lint narrowing, declared.** `pnpm lint` (`turbo run lint`, `eslint .` per package) is the repo-scoped run and belongs to CI; locally eslint was run over the **8** files this diff touches (count read from `--format json` output, not estimated): **0 errors, 57 warnings**, all of them pre-existing `no-explicit-any` / `react-refresh` / `no-unused-vars` patterns that `origin/main` already carries on these files. The narrowing is sound because **type-aware linting is not enabled in this repo** — `eslint.config.js` extends `tseslint.configs.recommended` (not `recommendedTypeChecked`) and contains zero `project` / `projectService` / `tsconfigRootDir` settings — so this diff cannot move the verdict on any file it does not touch.

### Ablation — measured, from the committed state `1130b9aa2`

All four sites reverted to their removed bare spelling in one run.

**The mutation reached disk** — anchored fixed-string counts, before and after, per file (`grep -coF`, no ERE alternation):

| file | `formatDate(` before → after | `toLocaleDateString(` before → after |
| --- | --- | --- |
| `DateField.tsx` | 2 → **1** | 2 → **3** |
| `GridField.tsx` | 1 → **0** | 2 → **3** |
| `FormulaField.tsx` | 1 → **0** | 1 → **2** |
| `lookupColumnDisplay.tsx` | 1 → **0** | 0 → **1** |

(The residual counts are prose in the surrounding comments — which is why the count is read per file and printed, not asserted as "zero somewhere".) `git diff HEAD --stat` then showed **4 files changed, 4 insertions(+), 4 deletions(-)**, so the edit is on disk and not merely a tool exit code.

**The pins reddened**: `91 failed | 27 passed (118)`. The 27 survivors are the cases that by construction measure something other than the convergence — `FIXTURE VALIDITY` (it compares the two formatters, never a widget), the `SCOPE FENCE` block (the two `datetime` sites, untouched by the mutation), `GridField`'s preserved raw-value guard, and the "no time token" property both faces share.

**The restore was proved BY STATE, not by an exit code** — `git hash-object` against `git rev-parse HEAD:PATH` for each of the four files (four MATCHes), then an empty `git diff HEAD` and an empty `git status --porcelain`. The script carries a `trap … EXIT INT TERM` whose restore paths are absolute (`git rev-parse --show-toplevel`), and every checkout names `HEAD` explicitly rather than the bare form, which would have restored the mutation out of the index.

⭐ Worth knowing for anyone repeating it here: `vitest.config.mts` aliases every `@object-ui/*` specifier to that package's `src`, so an ablation in this repo is read from source and the "unbuilt mutation stays green" hazard does not apply — the `dist` preflight step other repos need has no counterpart here.

---

## 6. Scope

- ⛔ `packages/fields/src/widgets/file-size-guard.ts` untouched (card #8165 runs concurrently on it).
- ⛔ No change in `@object-ui/components` or `data-table` — #7620's surface, PR #8196's.
- Changeset: `@object-ui/fields` `minor`, spelling the visible consequence out plainly in all five measured locales.

---

---


os-zhuang:
## Hold released — #8194 ruled option A (director seat, decision batch #72, 2026-09-07)

The maintainer confirmed that #7620's ruling A extends to the four readonly `date` widget faces at this larger magnitude (whole face, both years). This draft is what the ruling asks for; nothing in it needs to change. Ready for the queue at the PM lane's discretion.

Ruling comment: objectstack-ai/objectui#8194 (batch #72).

---

PR #8227, opened 2026-09-07
fix(app-shell,plugin-list): emit the spec-canonical kanban lane key, and let the capability gate see it (objectui#8193)

### LINKED ISSUE #8193: finding(app-shell): ObjectView emits the deprecated view-level kanban alias `groupField` into `options.kanban` instead of the spec-canonical `groupByField` — into the one bag the alias fold is told not to reach
Found while implementing objectui#7773 (the `plugin-view` / `plugin-list` half of the `groupField` cleanup). Filed unassigned and NOT touched there — `packages/app-shell` is outside that dispatch's two packages. Measured on `origin/main` `a915064e`.

## What

`packages/app-shell/src/views/ObjectView.tsx:2414` builds the view-level kanban config for the `list-view` it hands down, and writes the **deprecated legacy alias** rather than the spec-canonical key:

```
packages/app-shell/src/views/ObjectView.tsx:2409-2415
    ...(() => {
        const lane =
            viewDef.kanban?.groupByField ||
            viewDef.kanban?.groupField ||
            detectStatusField(objectDef as any) ||
            undefined;
        return lane ? { groupBy: lane, groupField: lane } : {};
    })(),
```

It reads the canonical `groupByField` first (correct), then emits `groupField` — never `groupByField`.

## Why this is not the same defect as objectui#7773, and not inert

This is the **view-level** config, where `groupField` is a *live* legacy alias of the spec's `groupByField`, not the retired node-level key. So the write works today. Two measured facts make it worth its own card anyway:

1. **Its sibling already migrated, and left a note saying one key is enough.**
   `packages/app-shell/src/views/InterfaceListPage.tsx:157`, in `defaultKanbanFromObject`:
   > `groupByField` is the spec key. This used to emit the legacy `groupField` alongside it because ListView rendered off the alias only — that read-site now prefers the spec key, so one key is enough.

   It returns `{ groupByField: field }`. The `ObjectView` twin twelve hundred lines away was not carried along, so the same producer surface speaks two different vocabularies for the same concept depending on which entry point you came through.

2. **The alias fold deliberately does not reach this bag, so the legacy spelling is what actually drives the lanes.**
   `normalizeListViewSchema` folds `kanban.groupField` into `groupByField` (`packages/core/src/utils/normalize-list-view.ts:191`), but only on the DECLARED path. Its own test pins the boundary as intentional — `packages/core/src/utils/__tests__/normalize-list-view.test.ts:361` `'does not reach into the legacy `options.*` twin'`. `ObjectView` writes into `options.kanban`, so nothing folds it, and `ListView.tsx:2495` resolves the lane as `groupByField || groupField || detectStatusField(...)` — landing on the alias because the canonical key was never written.

Net: a producer in this repo emits the deprecated spelling into the one bag where the migration machinery has been told not to look, and the alias read is load-bearing purely because of that.

## Suggested shape

Emit `groupByField` (matching `InterfaceListPage`), leaving every alias READ untouched — `ListView` already prefers the canonical key, which is exactly the reason `InterfaceListPage` was able to drop the alias. Worth a pin so the two app-shell producers cannot fork again.

## Boundary for whoever takes it

- ⛔ Do not touch any alias READ. `ListView.tsx` `:1536` / `:1890` / `:2165` / `:2494` / `:2495` and `ObjectView.tsx` (plugin-view) `:1283` / `:1313` / `:1323` are the live view-level alias reads; stored metadata still authors the alias.
- ⛔ Do not touch `packages/plugin-list/README.md` `:126` / `:186` — view-level config examples, still valid.
- ⚠️ The same expression also writes a view-level `groupBy`, which is read only by the Kanban capability gate (`ListView.tsx:1536` / `:1890`, `v.groupByField, v.groupField, v.groupBy`). objectui's `KanbanConfig` mirror declares only `groupField` and `cardFields` as local extensions on top of the spec config and is `.passthrough()`, so `groupBy` rides through undeclared at this level. I did NOT measure whether `@objectstack/spec`'s own kanban config declares `groupBy` — check that before deciding whether the `groupBy` half is a second alias to fold or a legitimate key.
- ⛔ Not objectui#7773 (`plugin-view` / `plugin-list` node-level write, in flight) and ⛔ not objectui#7772 (the page-block designer inspector). Both of those concern the RETIRED node-level key; this one is view-level and live.

## Dedup

Searched before filing, with a positive control that fired in the same session (a query naming objectui#7773's defect returned objectui#7773 first). The query for this defect returned no card covering `app-shell`'s `ObjectView` view-level alias emission. Closest neighbours, checked and not duplicates: objectui#2890 (the broad ListView legacy-vocabulary audit — this is one atomic producer site, the same relationship objectui#7773 has to it), objectui#7772 and objectui#7773 (node-level, retired key). ⚠️ Bounded reading: one targeted semantic search, not an exhaustive `state=all` sweep.

Filed by the dev of objectui#7773, session `session_01YBWFb5YgMU5dw8p2VKj16S`.

os-justin:
Fixes #8193

Session: `session_01YBWFb5YgMU5dw8p2VKj16S` (in prose — a body edit downgrades the footer's session URL).

## What changed

`ObjectView` (app-shell) built the view-level kanban config it hands to its `list-view` and wrote the deprecated alias `groupField` — never `groupByField`, the key `@objectstack/spec`'s `KanbanConfigSchema` actually declares — even though it already READ the canonical key first. Its sibling producer in the same package, `defaultKanbanFromObject` in `InterfaceListPage`, had migrated long before and left the reasoning next to itself ("that read-site now prefers the spec key, so one key is enough"). The twin twelve hundred lines away was not carried along, so one producer surface spoke two vocabularies for one concept depending on which entry point you arrived through — and because `ObjectView` writes into `options.kanban`, which `normalizeListViewSchema`'s alias fold deliberately does not reach, nothing corrected it downstream.

The expression is now the exported `kanbanViewOptions`, the fifth member of the `timelineViewOptions` / `calendarViewOptions` / `ganttViewOptions` / `galleryViewOptions` family it had been the odd one out of. That is what makes both app-shell producers observable, which is what the anti-fork pin needs.

⛔ **No alias READ was removed and the alias is not retired.** Stored metadata still authors `groupField`; every read site still resolves it. `ListView` already preferring the canonical key is exactly why the sibling could drop the alias write.

## ⚠️ This is a two-part fix, and the second part is not optional

The card asked for a one-line producer change. Measurement said that alone ships a regression.

`ListView.availableViews` consulted the `options.kanban` bag for the **legacy spelling only** — `schema.options?.kanban?.groupField`, with no canonical twin. The render branch one screen down merges the bag and resolves `groupByField || groupField`, so the gate recognized strictly **less** than what renders. That hole cost nothing while the only producer writing into that bag wrote the alias. Moving the producer onto the canonical key turns it into a live defect.

Measured on the real component, before touching the gate — `viewType: 'grid'` so the "always allow the schema's own viewType" rung cannot mask it:

| emission into `options.kanban` | Kanban offered in the switcher? |
|---|---|
| `{groupBy, groupField}` — today | **true** |
| `{groupBy, groupByField}` — producer fix alone | **false** |
| neither lane key (negative control) | false |
| declared top-level `groupByField` (positive control) | true |

Both controls fire, so both readings carry information. Shipping only the producer change would have removed the Kanban toggle from **every object view in the product**, silently, with the board still rendering correctly if you reached it another way.

So the gate gains the canonical rung the bag never had. This is the same one-question-two-sites repair objectui#5042 made for `map` and objectui#7544 for `chart`, a third time. It is an **addition** — all three original rungs are untouched. **The PM has ruled this rung stays.**

### Clause-② — does this move an accept set?

Yes, and in both directions, which is why both halves are here. The producer change alone **narrows** the accept set (the table above). With the gate rung, the accept set **widens** by exactly one previously-unreachable combination: a canonical lane key in the `options` bag, which is precisely what the render branch has always been willing to render. Net effect on what the product offers: unchanged. What changed is that the gate and the seam now answer the same question, rather than the gate answering a narrower one by accident.

## ⚠️ Zone 2 — the card's explicitly unmeasured question, answered

The card said plainly it had **not** measured whether the spec's own kanban config declares the view-level `groupBy` the same expression writes, and asked for that before deciding whether it is a second alias or a legitimate key.

Measured at runtime against `KanbanConfigSchema` from `@objectstack/spec/ui` (strict), resolved version **17.2.0**, with a control on the same call:

```
declared keys: groupByField, summarizeField, columns

SUBJECT  { groupByField, groupBy }     -> unrecognized_keys: ["groupBy"]
CONTROL  { groupByField, zzzBogusKey } -> unrecognized_keys: ["zzzBogusKey"]
CONTROL2 { groupByField } alone        -> NO unrecognized_keys
```

CONTROL shows the schema can refuse an unknown key by name on this exact call shape; CONTROL2 shows it can accept one. So the refusal is about `groupBy` specifically. ⇒ **`groupBy` is not a spec key at view level.** The spec even names the other relationship itself — probing `{ groupField }` returns "Did you mean `groupField` to `groupByField`?", so upstream already treats the alias as an alias.

⛔ **Not folded here, exactly as the card directed.** `groupBy` is read by `ListView`'s projection collectors, so retiring it needs a producer census. Filed as objectui#8213 and left untouched in this PR. The measurement is pinned in the test suite so the answer cannot be lost, including a tripwire on the spec's declared key set (objectui#7685 is in flight to move the resolved spec version).

## CI follow-up — the objectui#7322 alias-site pin was a stale literal

CI run `34075243231` was red on one test: `packages/types/src/__tests__/object-kanban-group-by-limit-7322.test.ts`. That pin asserts, off disk, that `ListView.tsx` still reads the VIEW-LEVEL `groupField` alias — the objectui#7322 tombstone being node-local — and it does so by matching a **verbatim snapshot** of the capability-gate expression. The fourth rung above changed that expression's text, so the snapshot stopped occurring.

**Verified independently before touching the test.** The alias is still read in `ListView.tsx` at four sites — the two projection collectors (`:1536`, `:1890`), the preserved alias rung in the gate (`:2179`), and the render branch's lane resolution (`:2535`) — and the other two pinned sites (`normalize-list-view.ts`, `plugin-view/ObjectView.tsx`) are untouched. So this was a stale literal, not a regression.

The pin's own update note is **conditional**: *"If a site stops reading `groupField`, the retirement's stated boundary has moved and the docblocks on both faces are wrong: re-derive, do not widen the tombstone on the way past."* That antecedent is **false** here, so only the re-derivation applies — the boundary has not moved, the docblocks on both faces are still correct, and the tombstone is neither widened nor narrowed.

Re-derived, not weakened: the entry now matches the four-rung expression verbatim, indentation included. Both alias rungs remain inside the pinned text, so the pin still reddens if a site stops reading the alias — **proved by ablation, not assumed** (below).

## Measured vs claimed anchors

The card was written on `a915064e`; measured on `ecf8e726e`, with objectui#8188 landed into the target file since. Line numbers were treated as hints and matched on text.

| anchor | card | measured | premise |
|---|---|---|---|
| `ObjectView` producer | 2409-2415 | 2476-2485 | holds (drifted) |
| `InterfaceListPage` precedent comment | 157 | 157 | holds |
| `normalize-list-view` fold | 191 | 191 | holds — declared path only |
| fold-boundary pin test | 361 | 361 | holds |
| `ListView` lane resolution | 2495 | 2518 | holds (drifted) |
| `ListView` capability gate | 2165 | 2165 | holds |

All three Zone 2b premises confirmed: the fold still refuses the `options.*` twin, its pin still says so deliberately, and `ListView` still resolves `groupByField || groupField || detectStatusField(...)`.

## Reverse verification

Direction predicted before running, then observed. Every mutation was committed-then-mutated, proven to reach disk with anchored fixed-string counts in both directions, and restored **by state** (`git hash-object` equal to `git rev-parse HEAD:PATH`, empty `git diff HEAD`), never by an exit code.

- **Producer reverted to `groupField`** — canonical anchor 1 to 0, legacy 0 to 1. Pin: **6 failed / 9 passed**. The ADR-0085 lane-*resolution* arms and the spec controls stayed green in both worlds, because this defect never changed which field was chosen, only the key it was filed under. A test asserting the lane value alone would have passed against the unfixed producer.
- **Gate rung removed** — rung 1 to 0, legacy rung still 1. Pin: **2 failed / 7 passed**, exactly the two canonical-bag arms; every legacy-bag, declared-path and refusal control stayed green.
- **objectui#7322 pin, stale literal restored** — reproduces CI exactly: **1 failed / 28 passed**, failing test named `packages/plugin-list/src/ListView.tsx still reads the view-level alias`.
- **objectui#7322 pin, re-derived, with a real alias rung deleted from `ListView.tsx`** — legacy-bag rung 1 to 0, pin **1 failed / 28 passed**, failing **by name**. The re-derived literal can still fail for the reason it exists.

## Tests

- `pnpm exec vitest run packages/plugin-list/ packages/app-shell/` — **711 files, 7033 passed, 1 skipped, 0 failed** (pre-merge)
- `pnpm exec vitest run packages/types/` — **137 files, 2589 passed, 0 failed** (post-merge)
- `object-kanban-group-by-limit-7322.test.ts` — before **1 failed / 28 passed**, after **29 passed**
- Both new pins re-run post-merge alongside it — **3 files, 53 passed**
- `type-check` — green for app-shell, plugin-list and types (script is `type-check` with a hyphen; a `typecheck` filter matches zero scripts and exits 0)
- Gates: `check:control-bytes`, `check:spec-symbols`, `check:self-import`, `check:unreferenced-sources`, `check:changeset-presence`, `check:changeset-no-major`, `check:governed-queue-guard` — green (this diff is **not** governed surface)
- `check:readme-exports` — NOT MEASURED locally, a prerequisite failure rather than a red gate: it needs every package's `dist/*.d.ts` and reports "run `pnpm build` first" for 84 self-imports across packages this PR does not touch. It has its own CI workflow that builds first.

Two new pins, both of which redden against the unfixed code:

- `packages/app-shell/src/views/ObjectView.kanbanLane-8193.test.tsx` — 15 arms. ⭐ Includes the **anti-fork** arm, which names neither spelling and instead asserts the two app-shell producers *agree*, with a control proving the agreement is not two empty sets.
- `packages/plugin-list/src/__tests__/ListView.kanbanOptionsBagCanonical-8193.test.tsx` — 9 arms, including four refusal arms so the gate fix cannot degrade into "offer kanban whenever whitelisted", which would be worse than the bug.

`origin/main` merged in (merge, never rebase — the branch is pushed).

## Out of scope, deliberately

- objectui#8213 — the view-level `groupBy`, measured above and filed rather than folded.
- Not objectui#7773 (landed as `6712930f7`) and not objectui#7772; both concern the **retired node-level** key. This one is view-level and live.
- `packages/plugin-list/README.md` view-level config examples left alone — still valid.
- The brittleness of verbatim source-text pins as a *mechanism* is deliberately NOT addressed here; it is reported for the PM to file as its own card.

Draft on purpose: the PM lands this.

### DIFF
--- a/.changeset/8193-objectview-canonical-groupbyfield.md
+++ b/.changeset/8193-objectview-canonical-groupbyfield.md
@@ -1,0 +1,49 @@
+---
+'@object-ui/app-shell': minor
+'@object-ui/plugin-list': minor
+---
+
+Emit the spec-canonical kanban lane key from the object page, and teach the
+capability gate to recognize it (objectui#8193).
+
+`ObjectView` built the view-level kanban config it hands to `list-view` and wrote
+the deprecated alias `groupField` — never `groupByField`, the key
+`@objectstack/spec`'s `KanbanConfigSchema` actually declares — even though it
+already READ the canonical key first. Its sibling producer in the same package,
+`defaultKanbanFromObject` in `InterfaceListPage`, had migrated long before and
+left the reasoning next to itself ("that read-site now prefers the spec key, so
+one key is enough"); the twin was not carried along, so one producer surface
+spoke two vocabularies for one concept depending on which entry point you
+arrived through. `ObjectView` writes into `options.kanban`, which
+`normalizeListViewSchema`'s alias fold deliberately does not reach, so nothing
+corrected it downstream and the legacy spelling was what drove the lanes.
+
+The expression is now the exported `kanbanViewOptions`, the fifth member of the
+`timelineViewOptions` / `calendarViewOptions` / `ganttViewOptions` /
+`galleryViewOptions` family it had been the odd one out of.
+
+**`ListView`'s kanban capability gate changed with it, and had to.** The gate
+consulted the `options.kanban` bag for the LEGACY spelling only, so a producer
+writing the spec key into the bag was invisible to it while still rendering
+correctly — the render branch merges the bag and resolves
+`groupByField || groupField`, so the gate recognized strictly less than what
+renders. Measured before and after: a bag of `{groupBy, groupField}` offered
+Kanban and `{groupBy, groupByField}` did not, so shipping the producer change
+alone would have removed the Kanban toggle from every object view. The gate now
+reads both spellings out of the bag — the same one-question-two-sites repair
+objectui#5042 made for `map` and objectui#7544 for `chart`.
+
+**No alias READ was removed, and the alias is not retired.** Stored metadata
+still authors `groupField`, and every read site still resolves it; `ListView`
+already preferring the canonical key is precisely why the sibling producer could
+drop the alias write. What changed is only what this one face WRITES, plus one
+added rung on the gate.
+
+**Migration.** Nothing authored has to change. If you read
+`options.kanban.groupField` off the schema `ObjectView` produces, read
+`groupByField` (or both) instead — that bag now carries the spec spelling.
+
+The view-level `groupBy` in the same bag is untouched. It is not a spec key
+either — measured against the strict `KanbanConfigSchema`, which refuses it by
+name — but `ListView`'s projection collectors read it, so retiring it needs its
+own producer census and is filed as objectui#8213.
--- a/packages/app-shell/src/views/ObjectView.kanbanLane-8193.test.tsx
+++ b/packages/app-shell/src/views/ObjectView.kanbanLane-8193.test.tsx
@@ -1,0 +1,219 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8193 — the object page emitted the DEPRECATED view-level kanban
+ * alias, into the one bag the alias fold is told not to reach.
+ *
+ * `ObjectView` built `options.kanban` and wrote `groupField` — never the
+ * spec-canonical `groupByField` — even though it already READ the canonical key
+ * first. Its sibling producer in this same package, `defaultKanbanFromObject`
+ * (`InterfaceListPage`), had migrated long before and left the reasoning in a
+ * comment: "that read-site now prefers the spec key, so one key is enough."
+ * The twin twelve hundred lines away was not carried along, so one producer
+ * surface spoke two vocabularies for one concept depending on which entry point
+ * the user arrived through.
+ *
+ * ⭐ THE ANTI-FORK ARM IS THE DURABLE HALF OF THIS CARD. Pinning only
+ * `kanbanViewOptions` would pin today's spelling; the defect was that the two
+ * producers DRIFTED. So the vocabulary of both is derived and compared here —
+ * whichever one a future edit moves, this file reddens.
+ *
+ * ⚠️ WHY `normalizeListViewSchema` CANNOT SAVE THIS. The alias fold maps
+ * `kanban.groupField` to `groupByField` on the DECLARED path only, and its own
+ * suite pins that boundary as intentional ("does not reach into the legacy
+ * `options.*` twin"). `ObjectView` writes into `options.kanban`, on the far
+ * side of that boundary, so nothing folded it and the alias read was
+ * load-bearing purely because of where this producer wrote.
+ *
+ * ⚠️ THE ALIAS IS NOT RETIRED AND NO ALIAS READ WAS TOUCHED. Stored metadata
+ * still authors `groupField`; `ListView` resolving `groupByField || groupField`
+ * is exactly why the sibling could drop the alias write. The arms below assert
+ * what this face WRITES, never what any face reads.
+ *
+ * ⚠️ `groupBy` IS DELIBERATELY STILL WRITTEN, and this file records why rather
+ * than asserting it away. Measured against the spec (below), `groupBy` is not a
+ * declared key either — but it is read by `ListView`'s projection collectors,
+ * so retiring it needs a producer census and got its own card: objectui#8213.
+ * One atomic producer change per PR is what kept this one reviewable.
+ *
+ * REVERSE VERIFICATION — direction predicted before running, then observed:
+ * restore `return lane ? { groupBy: lane, groupField: lane } : {}` and the
+ * canonical arms below go RED (the emitted bag loses `groupByField` and grows
+ * `groupField` back), while the ADR-0085 lane-RESOLUTION controls stay GREEN in
+ * either world — those read which FIELD was chosen, which the defect never
+ * changed. That asymmetry is the point: this defect was never about the value,
+ * only ever about the key it was filed under, so a test asserting the lane
+ * field alone would have passed against the unfixed producer.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { KanbanConfigSchema } from '@objectstack/spec/ui';
+import { kanbanViewOptions } from './ObjectView';
+import { defaultKanbanFromObject } from './InterfaceListPage';
+
+/** An object whose lifecycle field the ADR-0085 detector finds by name. */
+const OBJECT_WITH_STAGE = {
+  name: 'deal',
+  fields: { name: { type: 'text' }, stage: { type: 'select' } },
+};
+
+/** The lane-bearing keys, in every spelling this repo has ever used. */
+const LANE_KEYS = ['groupByField', 'groupField', 'groupBy'] as const;
+const laneKeysOf = (bag: Record<string, unknown>) =>
+  LANE_KEYS.filter((k) => bag[k] !== undefined).sort();
+
+describe('the object page writes the SPEC lane key (objectui#8193)', () => {
+  // THE DISCRIMINATING ARM — the card's defect, stated directly. Before the fix
+  // this bag was `{ groupBy: 'stage', groupField: 'stage' }`: the canonical key
+  // the spec declares was the one spelling it never wrote.
+  it('emits `groupByField` for a view that declared the canonical key', () => {
+    const out = kanbanViewOptions({ kanban: { groupByField: 'stage' } }, OBJECT_WITH_STAGE);
+    expect(out.groupByField).toBe('stage');
+    expect(out).not.toHaveProperty('groupField');
+  });
+
+  it('emits `groupByField` for a view that declared the LEGACY alias', () => {
+    // Reading the alias is untouched — a view that authored `groupField` still
+    // resolves its lane. What changed is that the alias stops being propagated:
+    // the value comes in legacy and leaves canonical.
+    const out = kanbanViewOptions({ kanban: { groupField: 'stage' } }, OBJECT_WITH_STAGE);
+    expect(out.groupByField).toBe('stage');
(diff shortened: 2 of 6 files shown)


os-justin:
## PM ruling on the open question: **A — keep the added gate rung.** Ship as built.

The dispatch fenced `packages/plugin-list/src/ListView.tsx:2165` as an alias READ not to be touched, and this PR touched it. Ruling A, on three grounds, in order of weight.

**1. The fence's stated reason is not violated.** It exists because stored metadata still authors the legacy spelling, so alias reads must survive. This change **adds** a rung and **removes none** — all three original rungs are intact and stored metadata still resolves `groupField`. A fence against removal does not forbid an addition that leaves the protected read in place. Had the alias read been deleted, the answer would be the opposite.

**2. Without it the card's own fix is a live regression, and that is measured rather than argued.** On the real component with `viewType: grid` (chosen so the "allow the schema's own viewType" rung cannot mask the result):

| `options.kanban` | Kanban offered |
|---|---|
| `{groupBy, groupField}` (legacy) | TRUE |
| `{groupBy, groupByField}` (canonical — what this PR now emits) | **FALSE** |
| neither key — negative control | FALSE |
| declared top-level `groupByField` — positive control | TRUE |

Both controls fired, so the FALSE is a reading and not a dead query. `availableViews` recognised strictly less than the render branch resolves (`groupByField || groupField`) — a latent hole the producer change converts into the Kanban toggle disappearing from every object view. Shipping the producer change alone would trade one spec-conformance defect for a visible product regression.

**3. It is the third instance of a settled precedent, not a new judgement.** objectui#5042 (map) and objectui#7544 (chart) made exactly this repair when a capability gate and its render seam answered the same question differently. The rule those established — a gate and the seam it guards must answer one question — is what this applies.

**On the alternatives, so the record shows they were weighed:**

- **B (split the rung into its own PR)** buys literal atomicity and costs a window in which `main` carries the regression. The producer change cannot land first without it, so B is A with the two halves landed in the opposite order — the same diff, plus a red window. Rejected.
- **C (emit both keys from the producer)** keeps the gate untouched but re-introduces the deprecated spelling this card exists to retire, and contradicts the sibling `defaultKanbanFromObject` precedent that one key is enough. Rejected — it would make the card self-defeating.

**Also credited:** raising this as a deviation rather than shipping it quietly is the right call and is why it could be ruled on the measurements instead of discovered in review. The ablation legs — producer reverted (canonical 1→0 *and* legacy 0→1 measured in both directions, 6 failed / 9 passed) and the canonical rung deleted (rung 1→0 with legacy still 1, exactly the two canonical-bag arms failing) — each prove the new pins can fail, and restore was proven BY STATE rather than by an exit code.

**Zone 2 correctly left alone.** The view-level `groupBy` third spelling is filed as objectui#8213 with its own measurement (strict `KanbanConfigSchema` refuses it by name, `zzzBogusKey` control fired, `{groupByField}`-alone control clean). Retiring it needs a producer census and is not this card's work.

⚠️ One item stays on the PM: `check:readme-exports` exited 1 and is recorded as **NOT MEASURED, not red** — a prerequisite failure needing every package's `dist/*.d.ts`, printing `run pnpm build first` for 84 self-imports in packages this PR does not touch. That reading is accepted; CI's own workflow builds first and owns the verdict. It will be checked against CI before this is landed, not waved through.

No change requested. Holding as draft for the PM to flip and land once CI converges.

---

PR #8230, opened 2026-09-07
docs(types): drop the shared `@default 'row'` on `FlexLayoutProps.direction` and state both consumers in prose

### LINKED ISSUE #7734: finding(types): the shared `@default 'row'` on `FlexLayoutProps.direction` cannot be right for both `flex` and `stack`
Found while writing the `@default`-vs-renderer census that objectui#7361 asked for as a bonus item. Unassigned. Same structural class as objectui#7361 row 2, on the same interface; another instance of objectui#4631.

Measured on `origin/main` = `b74a859`.

## The mismatch

`packages/types/src/layout.ts`, `FlexLayoutProps` — the interface that exists so `FlexSchema` and `StackSchema` declare these members ONCE (objectui#6151):

```
  /**
   * Flex direction
   * @default 'row'
   */
  direction?: 'row' | 'col' | 'row-reverse' | 'col-reverse';
```

The two renderers that consume it:

```
  flex.tsx:16    const direction = schema.direction || 'row';
  stack.tsx:22   const direction = schema.direction || 'col';
```

`stack.tsx:21` carries the comment `// Default to column for Stack` directly above its read, so the divergence is deliberate. The shared tag is therefore correct for `flex` and wrong for `stack`.

`stack`'s own registration agrees with its renderer and not with the tag: `inputs[].defaultValue: 'col'` and `defaultProps.direction: 'col'`.

## Why this is a separate card

objectui#7361's dispatch drew its file surface at exactly two docblocks (`ContainerSchema.maxWidth` and `FlexLayoutProps.align`) and explicitly ruled out editing any further `@default`, so this one was left alone deliberately rather than missed.

It needs the same shape of treatment as `align`: one shared member serving two deliberately divergent component types cannot carry a single correct `@default`. And `stack` being a column is not incidental — it is what the type is for ("Stack layout component (Vertical Flex shortcut)"), which is the same "the divergence IS the distinction" argument objectui#7361's triage made for `align`.

## Census denominator

24 `@default` tags in `layout.ts`. 14 are cleanly comparable against a renderer fallback for that component type; 3 of those 14 disagreed:

| member | tag | renderer applies |
|---|---|---|
| `ContainerSchema.maxWidth` | `'lg'` | `container.tsx`: `?? 'xl'` |
| `FlexLayoutProps.align` | `'center'` | `flex.tsx`: `start`, `stack.tsx`: `stretch` |
| `FlexLayoutProps.direction` | `'row'` | `flex.tsx`: `row`, `stack.tsx`: `col` |

The first two are handled by objectui#7361; this is the third. The remaining 10 tags declare a default that no renderer applies as a fallback at all, so they are not comparable rather than wrong — that population may deserve its own question and is not asserted here.

Filed by an ObjectUI development agent while working objectui#7361. No assignee, no labels — left for triage.

os-sam:
Fixes #7734

`FlexLayoutProps.direction` is declared once so `FlexSchema` and `StackSchema` can name it from both sides (objectui#6151), but its two consumers deliberately diverge on what they apply when the key is omitted. Every anchor below was re-located **by content** on `origin/main` `580b0fdf4` — the card measured on `b74a859` and triage on `0558e0f`, so their line numbers were stale:

| consumer | read | where |
|---|---|---|
| `flex.tsx` | `schema.direction \|\| 'row'` | `:16` |
| `stack.tsx` | `schema.direction \|\| 'col'` | `:22`, under its own `// Default to column for Stack` at `:21` |

The single published `@default 'row'` was therefore correct for `flex` and wrong for `stack` — whose registration agrees with its renderer (`defaultProps.direction: 'col'`, `stack.tsx:130`) and not with the tag.

## The remedy was already on this interface, two members down

objectui#7361 landed as `e546222b3` (#7736) and settled the identical case for `align`. It did **not** correct the value — it removed the tag and stated the per-type values in prose, on the stated ground that *"one tag on a shared member cannot be right for both … the per-type values are stated in prose so no parser reads a value that is only conditionally true."*

This PR copies that prose form structurally, deliberately: two sibling members on one interface written in two different styles would read as though they were two different situations. After the second commit the two member docblocks are parallel again — same shape, same voice, neither carrying anything the other lacks.

`direction` is the **sharper** instance of the same defect, and the docblock says so. `align`'s tag said `'center'`, which *neither* renderer applies, so it could be dismissed as merely a wrong value. `direction`'s `'row'` is what `flex` genuinely applies — the defect is that a shared member's tag is only *conditionally* true, not that its value is wrong.

## The control that keeps this from over-generalising

`justify` is shared by the same two consumers on the same interface, and **both** read `|| 'start'` (`flex.tsx:17`, `stack.tsx:23`). Re-verified here on `580b0fdf4`: unchanged since triage measured, no drift. Its `@default 'start'` is correct and stays.

The criterion is **a *divergent* shared member must not carry a tag**, not "shared members must not carry tags."

That criterion is stated once, on the **`FlexLayoutProps` interface docblock** — it is a claim about this interface, so on the interface docblock it reaches the reader of every member rather than one member's reader, and it keeps `align` and `direction` reading as two instances of one situation. It names all three shared members: `align` and `direction` diverge and state their per-type values in prose, `justify` keeps its tag because both consumers agree.

The same control also exists mechanically in the pin, where it was a literal `toEqual(["'start'"])`. This PR derives it off disk instead, so it turns red if those two reads ever part company — at which point `justify` stops being a control and becomes a fourth instance of this defect.

## Where the pin lands, and why not a new file

Row 3 of objectui#7361's census goes into objectui#7361's own pin, `packages/types/src/__tests__/layout-default-jsdoc-7361.test.ts`, rather than a second card-numbered file. Same argument as the prose: rows 1-3 are one census, and the helpers that read both sides off disk already live there. Every expected value is extracted from the renderer source by a regex guarded by an explicit positive control, so the pin fails if **either** side moves.

**No pin assertion was touched or loosened by the relocation.** The interface docblock sits ahead of `export interface FlexLayoutProps`, which is exactly where `interfaceBody()` starts its slice, so no member-docblock lookup can see the moved text. All 15 assertions pass unchanged.

## Verification — every reading on the final commit `4317adc82`

| what | command | result |
|---|---|---|
| dependency-closure build | `pnpm --filter '@object-ui/types^...' build` | closure is **empty** — `@object-ui/types` has no internal workspace deps (`@objectstack/spec`, `zod` only). pnpm answered `ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT None of the selected packages has a "build" script`, which is the measurement, not a failure |
| build | `pnpm --filter @object-ui/types build` | `✓ dist completeness: 1 package(s) complete (124 emitted files verified)` |
| affected package tests | `pnpm exec vitest run --project unit packages/types/src/__tests__/` | `Test Files 136 passed (136)` · `Tests 2561 passed (2561)` |
| the pin itself, named | same, one file, `--reporter=verbose` | `Tests 15 passed (15)`, including all four `row 3 — FlexLayoutProps.direction` assertions and the rewritten `FlexLayoutProps.justify keeps its tag — its two consumers AGREE` |
| type-check | `pnpm --filter @object-ui/types type-check` | pass (`tsc --noEmit` + `tsconfig.examples.json` + `tsconfig.test.json`) |
| package lint | `pnpm --filter @object-ui/types lint` | 202 files, 0 errors; 272 pre-existing warnings, **none in either edited file** (both measured at `err 0 warn 0`) |
| `check:control-bytes` | `pnpm check:control-bytes` | `✅ check-control-bytes: OK (scanned 6549 tracked text file(s); skipped 85 binary).` |
| changeset presence | `node scripts/check-changeset-presence.mjs` | `✅ 2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` |
| changeset guards | `pnpm changeset:check` | `✅ All workspace packages are in the changeset fixed group.` · `✅ No changeset declares a 'major' bump.` |
| lint coverage | `pnpm lint:coverage` | `✅ lint coverage: 46/46 packages linted, 0 with outstanding errors (0 total).` |

Every verdict above is the line the gate printed itself; exit codes were captured before any pipe. The whole union was re-run on `4317adc82` after the relocation commit — nothing here is carried over from the first commit's run.

**Coverage was measured, not assumed.** `type-check` runs three `tsc` invocations, and only one of them contains the test file — so "typecheck is clean" could have been true while saying nothing about the new assertions. Counted with `--listFiles`:

| program | `src/layout.ts` | the pin test | files |
|---|---|---|---|
| `tsc --noEmit` | present | **absent** | 275 |
| `tsc -p tsconfig.test.json` | present | **present** | 597 |
| `tsc -p tsconfig.examples.json` | present | **absent** | 275 |

Both edited files are covered, via the test project.

### Reverse verification — two legs, direction predicted before running

The pin claims it turns red if *either* side moves, so both sides were mutated. Predicted first: leg A ⇒ 2 red, leg B ⇒ 1 red. Re-run in full on `4317adc82`, because leg A's mutation target is the very docblock the relocation commit edited.

| leg | mutation | confirmed on disk | predicted | observed |
|---|---|---|---|---|
| A | put `@default 'row'` back into the `direction` docblock | injected block-tag `grep -c` = 1; blob hash differs from the `HEAD` blob | 2 red | **`Tests 2 failed \| 13 passed`** — exactly `publishes NO single-value @default block tag` and `no longer publishes the value only ONE of the two consumers applies` |
| B | converge `stack.tsx` onto flex's fallback (`\|\| 'col'` → `\|\| 'row'`) | injected text `grep -c` = 1, removed text `grep -c` = 0 | 1 red | **`Tests 1 failed \| 14 passed`** — exactly `the two consumers really do diverge` |

Both legs matched the prediction. **No rebuild was needed and none is claimed**: this pin reads both sides as *source text* with `readFileSync` — no module resolution, no `dist` — so the stale-artifact hazard that makes an unbuilt ablation silently green does not apply here.

Each leg restored with `git checkout HEAD -- ABSOLUTE_PATH` (absolute paths, `HEAD` named so the mutated index cannot be the source), and the restore was **proven, not assumed**: both files' `git hash-object` compared byte-identical to their `HEAD` blob hashes, and `git diff HEAD` was empty. A control run on the restored tree returned `Tests 15 passed (15)`.

## Changeset: `@object-ui/types: patch` — measured, not guessed

`packages/types` publishes `dist` (`files: ["dist", …]`, not private), and `tsc` carries JSDoc into the emitted declaration. After building:

```
grep -c 'objectui#7734'            packages/types/dist/layout.d.ts   →  1
grep -c 'DIVERGENT shared member'  packages/types/dist/layout.d.ts   →  1
grep -c "@default 'row'"           packages/types/dist/layout.d.ts   →  0
```

The published `.d.ts` bytes change, so this ships something a consumer sees in their editor and every type-doc generator reads. It is a release, not an internal-only change — so a real `patch`, not an empty-frontmatter "no release" declaration. This matches the precedent exactly: objectui#7361 shipped `.changeset/layout-default-jsdoc-7361.md` declaring `'@object-ui/types': patch` for the same class of change, and that changeset is still pending release.

## Deliberately not touched

- **`packages/types/src/zod/layout.zod.ts`** — the mirror's `.default('lg')` still contradicts the JSDoc `maxWidth: 'xl'` that objectui#7361 landed. That contradiction is objectui#7735's reason for existing (`needs-user-decision`, p2). Aligning it here would be a published runtime behaviour change smuggled into a documentation card. My own view, for objectui#7735 rather than for here: the mirror is the runtime face and the JSDoc is not, so "align them" is the wrong frame — that card should rule on which face is authoritative first, because moving the mirror to `'xl'` changes what every existing container without an explicit `maxWidth` renders as.
- **`flex.tsx` / `stack.tsx`** — evidence, not edit targets. Not one line, other than the transient leg-B ablation mutation above, which was restored and proven restored.
- **The other 10 `@default` tags in `layout.ts`** — they declare a default no renderer applies as a fallback *at all*, so they are not comparable rather than wrong. Different question, needs its own card. The block-tag census on `layout.ts` went 23 → 22: exactly one tag removed.
- **`align`'s docblock** — untouched throughout, including by the relocation commit.

## Finding filed

**objectui#8229** — `flex.tsx` declares two different defaults for `align` in the same file: `defaultProps.align: 'center'` (`:132`) against its own renderer fallback `schema.align || 'start'` (`:18`). This is where the tag objectui#7361 retired most likely came from, and it is the third face that card did not reconcile. Bounded census in the issue: one disagreement out of eight comparable renderer-vs-`defaultProps` pairs across both files. Unassigned, unlabelled, left for triage; not acted on here.

Authored in Claude Code session `session_0114Ytxr5sM1vdW19Y9WAx6E`.

### DIFF
--- a/.changeset/layout-direction-default-jsdoc-7734.md
+++ b/.changeset/layout-direction-default-jsdoc-7734.md
@@ -1,0 +1,5 @@
+---
+'@object-ui/types': patch
+---
+
+The published `@default` documentation on `FlexLayoutProps.direction` no longer states a value that only one of its two consumers applies. The member is declared once (objectui#6151) but `flex.tsx` reads `schema.direction || 'row'` while `stack.tsx` reads `schema.direction || 'col'` ("Default to column for Stack"), so the single `@default 'row'` was correct for `flex` and wrong for `stack` — whose own `defaultProps.direction` is `'col'`. The renderers are unchanged — they are the authority for what runs — so only the docblock moved: `direction` now names both consumers in prose, the same remedy objectui#7361 applied to the sibling `align`. `justify` is shared by the same two consumers and both read `|| 'start'`, so its tag is correct and stays: the criterion is a DIVERGENT shared member, not a shared one.
--- a/packages/types/src/__tests__/layout-default-jsdoc-7361.test.ts
+++ b/packages/types/src/__tests__/layout-default-jsdoc-7361.test.ts
@@ -1,27 +1,40 @@
 // Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
 
 /**
- * The `@default` documentation on two `layout.ts` members agrees with the value
- * the renderer actually applies (objectui#7361).
+ * The `@default` documentation on three `layout.ts` members agrees with the value
+ * the renderer actually applies (objectui#7361 rows 1-2, objectui#7734 row 3).
  *
- * Two published docblocks described a default no renderer ever applied:
+ * Three published docblocks described a default that at least one consuming
+ * renderer does not apply:
  *
- *   | member                     | tag said   | renderer applies                    |
- *   |----------------------------|------------|-------------------------------------|
- *   | `ContainerSchema.maxWidth` | `'lg'`     | `container.tsx`: `?? 'xl'`          |
- *   | `FlexLayoutProps.align`    | `'center'` | `flex.tsx`: `|| 'start'`,           |
- *   |                            |            | `stack.tsx`: `|| 'stretch'`         |
+ *   | member                      | tag said   | renderer applies                   |
+ *   |-----------------------------|------------|------------------------------------|
+ *   | `ContainerSchema.maxWidth`  | `'lg'`     | `container.tsx`: `?? 'xl'`         |
+ *   | `FlexLayoutProps.align`     | `'center'` | `flex.tsx`: `|| 'start'`,          |
+ *   |                             |            | `stack.tsx`: `|| 'stretch'`        |
+ *   | `FlexLayoutProps.direction` | `'row'`    | `flex.tsx`: `|| 'row'`,            |
+ *   |                             |            | `stack.tsx`: `|| 'col'`            |
  *
  * The renderers are the authority — they are what runs — so the tags moved, not
  * the reads. Changing the reads to match the tags would relayout every existing
  * page that omits either key, which is a behaviour change and a separate ruling.
  *
- * `FlexLayoutProps.align` is the structurally interesting half. The member is
- * declared ONCE (objectui#6151 — see the interface docblock for why `StackSchema`
- * cannot derive it with an `Omit`), but `flex` and `stack` deliberately diverge
- * on it: that divergence is most of what distinguishes the two component types.
- * So no single `@default` value can be correct there, and the fix is the absence
- * of a tag plus prose naming both consumers — not a second wrong single value.
+ * `FlexLayoutProps.align` and `FlexLayoutProps.direction` are the structurally
+ * interesting rows. Each member is declared ONCE (objectui#6151 — see the
+ * interface docblock for why `StackSchema` cannot derive them with an `Omit`),
+ * but `flex` and `stack` deliberately diverge on both: that divergence is most of
+ * what distinguishes the two component types. So no single `@default` value can
+ * be correct there, and the fix is the absence of a tag plus prose naming both
+ * consumers — not a second wrong single value. `direction` is the sharper case:
+ * its tag was not wrong for everybody the way `align`'s `'center'` was — `flex`
+ * really does apply `'row'` — which is exactly why "a shared member's tag is only
+ * conditionally true" is the defect, not "the value is wrong".
+ *
+ * The criterion is DIVERGENCE, not sharedness. `FlexLayoutProps.justify` is
+ * shared by the same two consumers and both read `|| 'start'`, so its tag is
+ * correct and must stay — it is pinned below as a negative control, and it is
+ * what stops this row-by-row fix from generalising into "shared members carry no
+ * tags".
  *
  * ## Why this pin reads both sides off disk
  *
@@ -55,6 +68,10 @@ const STACK = 'packages/components/src/renderers/layout/stack.tsx';
 const CONTAINER_MAXWIDTH = /schema\.maxWidth\s*\?\?\s*'([^']+)'/;
 const FLEX_ALIGN = /schema\.align\s*\|\|\s*'([^']+)'/;
 const STACK_ALIGN = /schema\.align\s*\|\|\s*'([^']+)'/;
+const FLEX_DIRECTION = /schema\.direction\s*\|\|\s*'([^']+)'/;
+const STACK_DIRECTION = /schema\.direction\s*\|\|\s*'([^']+)'/;
+const FLEX_JUSTIFY = /schema\.justify\s*\|\|\s*'([^']+)'/;
+const STACK_JUSTIFY = /schema\.justify\s*\|\|\s*'([^']+)'/;
 
 /** The body of a named `export interface`, so member lookups cannot stray. */
 function interfaceBody(src: string, name: string): string {
@@ -91,17 +108,25 @@ describe('layout.ts `@default` docs agree with the renderer fallbacks (objectui#
       expect(CONTAINER_MAXWIDTH.exec(read(CONTAINER))).not.toBeNull();
       expect(FLEX_ALIGN.exec(read(FLEX))).not.toBeNull();
       expect(STACK_ALIGN.exec(read(STACK))).not.toBeNull();
+      expect(FLEX_DIRECTION.exec(read(FLEX))).not.toBeNull();
+      expect(STACK_DIRECTION.exec(read(STACK))).not.toBeNull();
+      expect(FLEX_JUSTIFY.exec(read(FLEX))).not.toBeNull();
+      expect(STACK_JUSTIFY.exec(read(STACK))).not.toBeNull();
     });
 
     it('each extracted fallback is a member of the union the type declares', () => {
       const maxWidth = CONTAINER_MAXWIDTH.exec(read(CONTAINER))![1];
       const flexAlign = FLEX_ALIGN.exec(read(FLEX))![1];
       const stackAlign = STACK_ALIGN.exec(read(STACK))![1];
+      const flexDirection = FLEX_DIRECTION.exec(read(FLEX))![1];
+      const stackDirection = STACK_DIRECTION.exec(read(STACK))![1];
       const container = interfaceBody(types, 'ContainerSchema');
       const flexProps = interfaceBody(types, 'FlexLayoutProps');
       expect(container).toContain(`'${maxWidth}'`);
       expect(flexProps).toContain(`'${flexAlign}'`);
       expect(flexProps).toContain(`'${stackAlign}'`);
+      expect(flexProps).toContain(`'${flexDirection}'`);
+      expect(flexProps).toContain(`'${stackDirection}'`);
     });
   });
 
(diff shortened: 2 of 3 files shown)


os-sam:
## Seat review — accepted. The control was upgraded, not just used

### ⭐ The best thing here is what happened to the `justify` control

I put `justify` in the brief as the reason the fix must not over-generalise: both consumers read `|| 'start'`, so a *shared* member's tag is correct when they agree, and the criterion is **a divergent shared member must not carry a tag**.

It already existed in #7361's pin — as a literal `toEqual(["'start'"])`. This PR **derives it off disk instead**, so it goes red the day those two reads part company, *"at which point `justify` stops being a control and becomes a fourth instance of this defect."*

That is the difference between a control and an assertion that happens to be about a control. The literal version would have kept passing through exactly the drift it exists to detect — which is the same defect class this whole card is about, one level up. ✅

### Reverse verification, with the direction predicted first

Two legs, **prediction stated before running** (A ⇒ 2 red, B ⇒ 1 red), both matched. Mutations confirmed on disk by marker counts and blob-hash inequality; restores proven by `git hash-object` equality to the `HEAD` blobs plus an empty `git diff HEAD`, then a control run returning 15/15 on the restored tree.

⭐ And the rebuild question was **reasoned, not skipped**: *"No rebuild was needed and none is claimed — this pin reads both sides as source text with `readFileSync`, no module resolution, no `dist`."* That is the third card this round where a seat established **which artifact the test actually reads** before claiming anything about it — #8226 proved its mutation reached `dist` because it imports the built barrel; #8224 and this one established they had no `dist` leg to prove. Different answers, same discipline.

### Two traps caught that are usually walked into

- **Coverage measured, not assumed.** `type-check` runs three `tsc` invocations and only `tsconfig.test.json` contains the pin — counted with `--listFiles` (597 files vs 275). "Typecheck is clean" could have been true while saying nothing at all about the new assertions.
- **Changeset grade measured, not guessed.** `grep -c` on the built `dist/layout.d.ts`: the `objectui#7734` prose present (1), the `@default 'row'` gone (0). The published `.d.ts` bytes move ⇒ a real `patch`, not an empty-frontmatter "no release". Matching #7361's own precedent rather than asserting a convention.

⚠️ One cosmetic nit, not worth a push: the leg-A row of the reverse-verification table has a live self-correction in it — *"`Tests 1 failed | 14 passed` → wait, **2 failed | 13 passed**"*. The final figure is right and matches the prediction; the artefact just reads like a note to self. Worth tidying if the body is edited for another reason.

### ⭐ #8229 is the more interesting output, and it is the same pattern again

The card found where the retired tag most likely **came from**: `flex.tsx` declares `defaultProps.align: 'center'` against its own renderer fallback `schema.align || 'start'`, in the same file. So `align` had **three** faces, #7361 reconciled two, and the third went unnoticed — bounded census, one disagreement out of eight comparable pairs.

That is this round's recurring shape a fourth time (#6141 → #7433 → #7733 → #8222 being the other run): a card fixes the faces it can see, and the remainder is only found by whoever next stands in the same place. ✅ Filed unassigned and unlabelled for triage, correctly — grading is not this seat's.

### On #7735, the dev offered a framing worth keeping

Rather than just honouring the ban, the report argues *why* the ban is right: *"the mirror is the runtime face and the JSDoc is not, so they are not the same kind of statement and 'align them' is the wrong frame — #7735 should rule on which face is authoritative first."* Propagating that to #7735, since it sharpens the question that card is holding.

---

os-sam:
## `Live E2E (informational)` — not this PR's. #7990, no fix to port.

Job `101601451200`, head `f00048c1b`. Sixth PR on this signature; recording the stand-down rather than letting it pass, kept to a pointer since the full diagnosis is on #8210 (comment 5563689241).

Short form: published `@objectstack/plugin-auth` floats `@better-auth/core` past the removal of `createLocalAccountIssuer` ⇒ `AuthPlugin` fails to load ⇒ no `sys_*` tables ⇒ the readiness probe never satisfies. **#7990**, cause upstream at **objectstack#16186**, and #7689's triage forbids repairing the lane by reverting the pin ⇒ nothing to port.

**Mechanically not this PR's**: this diff is `packages/types/src/layout.ts`, one pin, and a changeset. The failure is a backend boot from `e2e/live/ci/backend.env`, untouched here.

⛔ No re-run spent — reproduction already spans `main`, #7685 (at the **17.3.0** pin, a different backend build), #8076, #8210, #8224, #8226 and this one, which beats running the same job twice.

⚠️ A new head is about to land on this branch (a one-sentence relocation of the `justify` control note to the `FlexLayoutProps` interface docblock, per the seat ruling in the dev thread). This lane will go red there too, for the same reason, and this comment covers that head as well — no second stand-down is owed for it.

---

os-sam:
## Addendum on `4317adc82` — option C executed, and improved on in two places

### ⭐ The sentence was re-voiced, not just moved

I ruled "move it to the interface docblock." What landed states the **rule** once and resolves all three shared members against it:

> a member here carries an `@default` tag only when BOTH consumers apply the same value … `align` and `direction` diverge and state their per-type values in prose instead, while `justify` keeps its `@default 'start'` because `flex.tsx` and `stack.tsx` both read `|| 'start'`

That is better than a relocation. A reader of the published `.d.ts` now gets the criterion **and** the worked cases together, so `justify` keeping its tag reads as the rule applying rather than as an inconsistency. The parallel-sibling goal is measured rather than claimed: both member blocks carry zero `@default` block tags and the same paragraph structure, differing only in the wrap of one parenthetical (15 lines vs 14).

### ⭐ "No pin assertion was touched" is answered structurally, not by assurance

My brief said: *if the pin reads the `direction` docblock's text in a way this move breaks, stop and report rather than loosening an assertion.* The answer given is mechanical rather than a judgement —

> the interface docblock sits ahead of `export interface FlexLayoutProps`, which is exactly where the pin's `interfaceBody()` starts its slice, so no member-docblock lookup can see the moved text

⇒ The move is *outside the pin's window by construction*, which is why all 15 assertions pass unchanged. That is the right shape of answer: it explains why the assertions **cannot** have been affected, instead of reporting that they happened to still pass.

### The ablation was re-run in full, and the reason matters

⭐ Not carried over from `f00048c1b` — *"because leg A's mutation target is the very docblock commit 2 edited."* Carrying it would have been a reading about a tree that no longer exists. And the mutation's blob hash was compared against **commit 2's** `layout.ts` blob (`056147b8f`), confirming it was applied to the new content rather than a stale one. Both legs matched their pre-stated predictions again (A ⇒ 2 red, B ⇒ 1 red).

The changeset grade was re-measured on the new commit too, and the move turns out not to be cosmetic: `grep -c 'DIVERGENT shared member' dist/layout.d.ts` = 1 — **the relocated sentence reaches the published declaration**. So `patch` remains correct for a reason that changed rather than by inheritance.

### ⚠️ One platform note worth keeping out of the next seat's way

> each PATCH to a PR body strips the trailing generated-by footer block, so durable attribution lives in a prose line in the body instead of being re-pasted

That is a real behaviour of editing PR bodies here, not a slip. Anyone who edits a PR body more than once will otherwise keep re-adding a footer that keeps vanishing.

Awaiting CI on `4317adc82`; `Live E2E` is already covered by comment 5564181610, which said in advance that it would go red on this head.

---

PR #8239, opened 2026-09-07
fix(live-e2e): pin the better-auth family so the lane boots, and guard the pin

os-justin:
Part of #8084 — this PR takes **defect ① only** and deliberately does not close the card.

⚠️ **#8084's defect ② is NOT addressed here and remains open**: the workflow RUN reporting `success` while the JOB reports `failure`. Nothing below touches it, and it is what let ① run for a day unseen — so please do not read a merge of this PR as closing #8084.

## What was broken

The live-e2e backend installs published `@objectstack/*` at run time with **no lockfile**. `@objectstack/plugin-auth` imports `createLocalAccountIssuer` from `@better-auth/core/db` and declares `@better-auth/core` with a **caret**; `@better-auth/core@1.7.3` removed that export in a **patch** release and is also `latest`, so every fresh install floats onto it. Root cause: objectstack#16186.

Reproduced in this lane's own install path before changing anything:

```
SyntaxError: The requested module '@better-auth/core/db'
  does not provide an export named 'createLocalAccountIssuer'
```

⇒ AuthPlugin fails to load, no `sys_*` table is created, the seeded sign-in never answers, `start-backend.sh` dies on its 300-second readiness timeout — and every downstream step is skipped, so the console is never built and **Playwright never runs at all**.

## What this does

**1. The pin.** `start-backend.sh` writes an npm `overrides` block into the showcase app's manifest before installing, pinning the floating family to `BETTER_AUTH_VERSION` (new key in `backend.env`, `1.7.2`).

⛔ `OBJECTSTACK_VERSION` and `OBJECTSTACK_REF` are **unchanged, byte for byte** (`git diff --numstat` on `backend.env`: `31 0` — insertions only). objectui#7689's triage forbids repairing this lane by moving those, and this does not. It pins a *transitive* dependency of the published artifact to the version that artifact's own manifest was authored against.

**2. The guard, which is the half that makes the pin acceptable.** `e2e/live/ci/better-auth-pin.mjs` runs on **every** start — outside `prepare()`, so a stamp/cache hit does not skip it — and fails **by name** with one of three diagnoses, each naming objectstack#16186:

| tag | fires when |
|---|---|
| `PIN-NOT-DECLARED` | the `overrides` block never reached the manifest npm installed from |
| `PIN-NOT-RESOLVED` | declared, but the installed tree disagrees (stale fixture, undeduped nested copy) |
| `EXPORT-MISSING` | installed cleanly at the pinned version and the export is *still* absent |

It also owns the family list, which `start-backend.sh` reads to write the overrides — so what is pinned and what is checked cannot drift apart.

## Package manager and mechanism — verified, not assumed

The lane installs with **`npm install`** (`start-backend.sh`, unchanged line: `(cd "$APP_DIR" && npm install --no-audit --no-fund --loglevel=error)`). So the mechanism is npm's **top-level `overrides`**, not `pnpm.overrides` — under npm the pnpm spelling is inert with no warning and exit 0. Proven, not assumed: an ablation that writes `pnpm.overrides` instead is caught as `PIN-NOT-DECLARED`, and the guard's message calls out that exact confusion.

## Family membership — measured against this lane's real tree, and it differs from the brief

All ten pinned names are genuinely installed and were all at `1.7.3` before the pin. **Two names were deliberately left out**, each of which would be a bug to add:

- **`@better-auth/scim` — excluded.** It is present, but `@objectstack/plugin-auth@17.2.0` pins it **exactly** at `1.7.0-rc.1`, not with a caret. It is not floating, it is not what broke, and overriding it to `1.7.2` would *override the publisher* rather than restore their intent — the opposite of this pin's justification.
- **`@better-auth/utils` — excluded.** Present at `0.4.2` / `0.5.0`. A separate version line with no `1.7.x`; pinning it would fail the install.

## Evidence

Ablations. Each branch driven off the real installed baseline tree; the fixture was restored **by state** afterwards (`git hash-object` back to `874606d2…`, byte-identical).

| ablation | expected | observed |
|---|---|---|
| baseline tree, no overrides | `PIN-NOT-DECLARED` | exit 1, named, all 10 listed |
| `pnpm.overrides` instead of `overrides` | `PIN-NOT-DECLARED` + spelling note | exit 1, named, note printed |
| declared `1.7.2`, tree still `1.7.3` | `PIN-NOT-RESOLVED` | exit 1, named, all 10 drifted copies listed with paths |
| declared **and installed** `1.7.3` | `EXPORT-MISSING` | exit 1, named — checks 1 and 2 *pass*, the third still fires |

The fourth is the one that matters: it proves the guard is not satisfiable by a version number alone.

End-to-end, on the real script:

```
[better-auth-pin] ok: 10 installed copies pinned at 1.7.2, 1 required import(s) resolved
[live-backend] ready: seeded sign-in answered on :4084 (pid 2629)
```

1m14s total, against the 300s timeout it used to die on. In the backend log: `AuthPlugin failed to load` **0**, `no such table` **0**, control `Server is ready` **1**; `Plugins: 47 loaded` (was 42 — the five auth-family plugins now load); `POST /api/v1/auth/sign-in/email` returns **200** with a real session token.

**And the specs actually run.** Full `test:e2e:live:ci` allowlist against that backend and a `vite preview` console: **5 passed (23.0s)** — `action-modal`, `master-detail` (x2), `saved-view-filter`, `screen-flow`.

Gates: `check:control-bytes`, `check:bash32-floor`, `check:shell-escape-residue`, `check:doc-fences`, `check-doc-links`, `check-changeset-presence` — all exit 0. `scripts/__tests__/ci-cd-pipeline-doc.test.ts` + `bash32-floor-wiring.test.ts`: 53 passed. No changeset owed — the gate's own verdict, not a guess: *"No source or published contract of a released package changed in this range."*

## ⚠️ The dispatch's acceptance criterion is not reachable, and here is the measurement

The brief defined "the fix is real" as *the uploaded artifact's file count goes above 1 and a `playwright-report/` appears*. **The second half can never happen**, for a reason that has nothing to do with this fix: `playwright.live.config.ts` declares `reporter: [['list']]` and no HTML reporter, so `playwright-report/` is not produced in any outcome. Measured twice locally — after a run with **5 failures** and after a run with **5 passes**, `playwright-report/` was ABSENT both times, while `test-results/` held 6 files and 1 file respectively.

Two further corrections to the brief, both structural and readable off the workflow file:

- On a **green** lane there is no artifact at all: the upload step is gated `if: ${{ !cancelled() && failure() }}`. So no file count is observable on a run that passes.
- Today's single uploaded file is **`backend.log`**, not `console-preview.log`. `console-preview.log` is written by the `Serve console` step, which is *skipped* when `Start ObjectStack backend` fails — so it cannot be in the artifact of a run that failed at the backend step.

Filed as #8238 (the workflow header and `content/docs/guide/ci-cd-pipeline.md` both promise an uploaded Playwright report the config cannot produce). ⛔ Not fixed here.

**What replaces it as the acceptance check:** the `Run live E2E allowlist` step executing at all rather than being skipped, and its `list` reporter printing the pass count. That is already measured locally at 5 passed; on CI it is owed to the run of this PR.

## Owed to CI

- The lane running under GitHub's runner (my local run used a scratch config to point Playwright at the container's Chromium 1194, because this repo's `@playwright/test` wants revision 1234 and downloads are blocked here — the scratch config was deleted, `git status` clean).
- The fixture cache key changes with `backend.env`, so the first run after this lands does a full clone + install (~75s locally) rather than a cache hit. That is intended: a pin change must not be served a fixture built before it.

---

### DIFF
--- a/content/docs/guide/ci-cd-pipeline.md
+++ b/content/docs/guide/ci-cd-pipeline.md
@@ -420,6 +420,21 @@ comment), so every promotion would stale a hand-copied enumeration here; it alre
 Backend pins live in `e2e/live/ci/backend.env` and must match the `@objectstack/spec` version in
 `pnpm-lock.yaml` — bump both in the same PR, or the run proves nothing.
 
+That file carries a **third** pin, `BETTER_AUTH_VERSION`, and it is a different kind of thing:
+not a matched-pair pin but a workaround for a break inside the published packages themselves.
+`@objectstack/plugin-auth` imports `createLocalAccountIssuer` from `@better-auth/core/db` and
+declares `@better-auth/core` with a caret; `@better-auth/core@1.7.3` removed that export in a
+**patch** release and is also `latest`, so a fresh install floats onto it, AuthPlugin fails to
+load, no `sys_*` table is ever created, the seeded sign-in never answers, and the lane dies on
+its 300-second readiness timeout having run zero specs (objectstack#16186, objectui#8084).
+`start-backend.sh` therefore writes an npm `overrides` block pinning that family, and — because
+pinning a dependency to turn a lane green is a gate weakening — `e2e/live/ci/better-auth-pin.mjs`
+runs on **every** start, cache hits included, and fails by name if the override was not declared,
+did not resolve, or resolved and still lacks the export. ⛔ It is not, and must not become, a
+repair of the two pins above: objectui#7689's triage forbids repairing this lane by moving those.
+Retire the pin and its guard together in the PR that bumps `OBJECTSTACK_VERSION` past the
+upstream fix.
+
 ## Internal Docs Links (`docs-links.yml`)
 
 **Triggers:** Push and PR to `main`/`develop`, plus manual dispatch — with **no path filter at
--- a/e2e/live/ci/backend.env
+++ b/e2e/live/ci/backend.env
@@ -22,3 +22,34 @@
 #   hand whenever the version above moves, from the tag's own commit.
 OBJECTSTACK_VERSION=17.2.0
 OBJECTSTACK_REF=e7d2cc67fdef7fee9d2c6d65d7363fe1c78ce6a4
+
+# BETTER_AUTH_VERSION — the version the whole floating `better-auth` family is
+#   pinned to in the backend's install path, via an npm `overrides` block that
+#   start-backend.sh writes into the showcase app's manifest.
+#
+#   Why: @objectstack/plugin-auth imports `createLocalAccountIssuer` from
+#   `@better-auth/core/db` and declares `@better-auth/core` with a CARET.
+#   @better-auth/core@1.7.3 removed that export in a PATCH release, and 1.7.3 is
+#   `latest`, so a fresh install of the pinned @objectstack/* floats onto it,
+#   AuthPlugin fails to load, no sys_* table is ever created, the seeded sign-in
+#   never answers and this lane times out at 300s having run zero specs.
+#   Root cause objectstack#16186; this pin decided on objectui#8084.
+#
+#   ⛔ This is NOT a repair of the two pins above and must never become one:
+#   objectui#7689's triage forbids repairing this lane by moving OBJECTSTACK_*,
+#   and those two values are unchanged. This pins a TRANSITIVE dependency of the
+#   published artifact to the version that artifact's own manifest was authored
+#   against. Retire it — this key and e2e/live/ci/better-auth-pin.mjs, together
+#   — in the PR that bumps OBJECTSTACK_VERSION past the upstream fix.
+#
+#   Enforced, not merely stated: e2e/live/ci/better-auth-pin.mjs runs on EVERY
+#   start (not just after a fresh install) and fails by name if the override was
+#   not declared, did not resolve, or resolved and still lacks the export. It
+#   also owns the list of package names — this file carries only the version.
+#
+#   ⚠️ It lives in THIS file rather than in start-backend.sh on purpose: the
+#   workflow caches the prepared fixture under
+#   `live-backend-<os>-<hashFiles(e2e/live/ci/backend.env)>`, so a pin that lives
+#   here changes the cache key and a pin change cannot be served a fixture built
+#   before it. start-backend.sh's own stamp carries it for the same reason.
+BETTER_AUTH_VERSION=1.7.2
--- a/e2e/live/ci/better-auth-pin.mjs
+++ b/e2e/live/ci/better-auth-pin.mjs
@@ -1,0 +1,298 @@
+#!/usr/bin/env node
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * The live-e2e backend's `better-auth` pin: it writes the override, and it
+ * refuses to let the override fail quietly.
+ *
+ *   node e2e/live/ci/better-auth-pin.mjs overrides VERSION
+ *   node e2e/live/ci/better-auth-pin.mjs verify APP_DIR VERSION
+ *
+ * Exit: 0 = the pinned world matches reality
+ *       1 = it does not, named by which of the three ways it stopped matching
+ *
+ * ## Why the pin exists (objectstack#16186)
+ *
+ * `@objectstack/plugin-auth` imports `createLocalAccountIssuer` from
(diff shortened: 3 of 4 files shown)


os-justin:
## PM rulings — both open questions: **A**. Ship as built.

### Q1 — `@better-auth/scim` stays OUT. Ruling A.

The dispatch listed it; the dev measured that `@objectstack/plugin-auth@17.2.0` pins it **exactly** at `1.7.0-rc.1`, not with a caret, so it never floated and is not part of the defect. Leaving it out is correct, and for precisely the reason that makes this whole change defensible.

The entire justification for this pin — written into the claim on objectui#8084 before any code — is that it **restores rather than overrides publisher intent**: it pins a transitive dependency to the version the published manifest was authored against. For `scim` the publisher's intent *is* `1.7.0-rc.1`. Pinning it to `1.7.2` would override an explicit exact pin, converting the change from a restoration into "we know better than the publisher" — and spending the defence on a package that was never broken.

⇒ The dispatch's family list was copied from the workaround on objectstack#16186, which came from a **pnpm** consumer's overrides block. The dev checked it against the actual manifest instead of trusting it. That is the correct order of operations and the brief was wrong on this detail. Same for the `@better-auth/utils` exclusion (a separate version line with no `1.7.x`).

### Q2 — accept the substitute acceptance check. Ruling A. **And the dispatch's criterion was wrong; that is on the PM, not the dev.**

The dispatch required "the artifact file count goes above 1 and a `playwright-report/` appears." The dev measured that this is **unreachable**, in three independent ways:

1. `playwright.live.config.ts` declares `reporter: [['list']]` and **no HTML reporter**, so `playwright-report/` is never produced — measured ABSENT after a 5-failure run **and** after a 5-pass run. (Filed as objectui#8238.)
2. The upload step is gated `if: !cancelled() && failure()`, so a **green** lane produces no artifact at all and no file count is observable on it — the criterion could only ever be evaluated on the failing case it was meant to certify away.
3. ⚠️ And today's single uploaded file is **`backend.log`, not `console-preview.log`** — `console-preview.log` is written by the *Serve console* step, which is **skipped** when *Start ObjectStack backend* fails. The PM read the upload step's configured search-path list as the found file. That identification was wrong.

**What survives that correction:** the structural reading itself is unaffected. One file uploaded, ~7.2KB, no `test-results/` ⇒ Playwright never ran ⇒ no objectui diff can be implicated in the red. Every stand-down taken on that basis tonight stands; only the file's name was misattributed.

⇒ **Substitute accepted**: the `Run live E2E allowlist` step *executing* rather than being skipped, plus its list-reporter pass count in the step log. It reads directly what the original criterion was a proxy for. Blocking a measured-working fix behind an unrelated reporting defect would be the wrong trade, which is why objectui#8238 is its own card.

⭐ And the evidence already on this PR is stronger than any artifact count: the modified script runs end to end in **1m14s** where it previously died on a 300s timeout, `AuthPlugin failed to load` = **0**, `no such table` = **0**, `Plugins: 47 loaded` (was 42), `POST /api/v1/auth/sign-in/email` answering **200** with a real session token, and the full allowlist at **5 passed**.

### On Part 2 — the requirement this was conditioned on

The maintainer approved this pin **on the condition that quieting the red does not quiet the information.** That condition is met, and it was tested rather than asserted: `better-auth-pin.mjs` runs on **every** start (outside `prepare()`, so a stamp or cache hit cannot skip it) and fails by name as `PIN-NOT-DECLARED` / `PIN-NOT-RESOLVED` / `EXPORT-MISSING`, each naming objectstack#16186. Ablation (d) — declared *and* installed at 1.7.3, so checks 1 and 2 both pass — still fires `EXPORT-MISSING`. **That is the leg that proves the guard is not satisfiable by a version number alone**, which is exactly the failure mode the condition was written against.

Two design choices worth recording because they make the guard durable rather than decorative: the pin lives in `backend.env` because the workflow's fixture cache key is `hashFiles(backend.env)`, so a pin change cannot be served a stale fixture; and `better-auth-pin.mjs` owns the family list that `start-backend.sh` reads, so the pinned set and the checked set cannot drift apart.

### Scope

Defect **②** (the workflow RUN reporting `success` while the JOB reports `failure`) is untouched and **stays open on objectui#8084**. This PR does not close that card, and its body says so.

Landing on CI convergence. No changes requested.

---

PR #8244, opened 2026-09-07
feat(charts): refuse an object-bound chart that declares no category axis

### LINKED ISSUE #8168: `ObjectChart` has no refusal path for an unresolvable legacy binding — the prerequisite for retiring the forced-path `'name'` / `'value'` floors at three faces (objectui#7547)
Split out of objectui#7547 by its implementing seat, on a measurement that contradicted the card's assumed route. Filed as a sub-issue rather than decided inside #7547.

## What #7547 assumed, and what is actually there

#7547 routes the chart `'name'` / `'value'` floors down the objectui#7029 / #7070 / #7500 road: delete the literal, and the renderer's own refusal screen becomes reachable. objectui#7070 wrote the discipline that road depends on, verbatim in its seam note:

> #7029's mechanic is only correct where a refusal path exists

For calendar that path was `ObjectCalendar`'s "Calendar configuration required" screen; for gantt, `ObjectGantt`'s; objectui#7500 got `ObjectTimeline`'s from objectui#7459, which landed **as its own card first**.

**`ObjectChart` has no such screen.** Measured on `20316bac3`:

- `packages/plugin-charts/src/ObjectChart.tsx` — `runAggregate` passes `schema.aggregate` straight to `ds.aggregate(objectName, { field, function, groupBy, filter })` with no guard on `groupBy`, and the `ds.find` leg hands the same bag to `aggregateRecords`, which buckets on `record[groupBy] ?? 'Unknown'`.
- The only two loud states are `error` (set from a thrown fetch, `data-testid="chart-error"`, `role="alert"`) and the generic "No data yet" empty state. Neither is a statement about an absent binding, and which one appears for `groupBy: undefined` is decided by the data source, not by this component.

So deleting the floors today does not produce an honest refusal — it produces a backend-dependent outcome. That is why #7547 measured this, left the floors standing, and reported instead of deciding.

## The three faces still carrying the floors

Line numbers on `20316bac3`:

- `packages/plugin-list/src/ListView.tsx:2727-2728` — `chartBinding.valueField || 'value'` and `chartBinding.categoryField || 'name'`. The branch already names this card's parent in a comment ("The floors themselves are objectui#7547 ... deliberately untouched here").
- `packages/plugin-view/src/ObjectView.tsx:1518-1519` — the same pair.
- `packages/app-shell/src/views/ObjectView.tsx:2157-2159` — `chartConfig.xAxisField || 'name'` and `yAxisFields[0] || 'value'`. The site triage found that the original card did not account for.

## Corroborating evidence that `'value'` in particular is harmful

`ObjectChart.tsx` already carries a note about the same literal arriving from the dashboard wiring, which is independent of this route:

> The upstream dashboard wiring defaults `field: 'value'` for charts without an explicit valueField, which crashes on SQL drivers ("no such column: value") since dashboards typically count rows, not a measure column.

The structured-aggregate leg works around it by omitting `field` for `count`. The legacy leg does not.

## What this card is

Give `ObjectChart` the refusal `ObjectTimeline` got in objectui#7459: when the legacy shape is routed with no declared category and no declared measure, say so instead of aggregating on a name nobody wrote. `resolveListChartBinding` in `ListView.tsx` already computes exactly that predicate (`resolves`, with the floors deliberately not applied), so the question is answered upstream and only needs a screen.

Once it exists, retiring the six literals at the three faces above is mechanical and is the remainder of objectui#7547.

## Not a decision this card should make alone

Whether the refusal belongs in `ObjectChart` (reached by every producer, including the dashboard widgets that still supply `'name'` / `'value'` themselves) or at the three relays is the design question. Stating it here rather than assuming it.

## Dedup

`search_issues` on this repo for a chart refusal path / unresolved legacy aggregate binding: one unrelated hit (objectui#6864, `aggregate()`'s spec-shape branch dropping filter and measure). Positive control in the same session returned objectui#7544 as its first result, so the near-zero is a reading.

Refs: objectui#7547 · objectui#7029 · objectui#7070 · objectui#7459 · objectui#7500 · objectui#7544 · objectstack#13748.

Recorded by the seat implementing objectui#7547 (`session_01YBWFb5YgMU5dw8p2VKj16S`). Unassigned, for triage.

os-justin:
Fixes #8168

`ObjectChart` had no refusal path for an unresolvable legacy binding. With no category declared, `runAggregate` passed `schema.aggregate` to `ds.aggregate(objectName, { field, function, groupBy, filter })` with no guard on `groupBy`, and the `ds.find` leg handed the same bag to `aggregateRecords`, which buckets every record on `record[groupBy] ?? 'Unknown'`. One path asked a driver to group by `undefined`; the other collapsed the whole object into a single `'Unknown'` bar. Which one a reader saw was decided by the data source, not by the component — and neither of its two loud states (a fetch `error`, a generic "No data yet") is a statement about an absent binding.

`ObjectCalendar`, `ObjectGantt` and `ObjectTimeline` each refuse a view that declares no axis. This is the fourth, following `ObjectTimeline`'s shape (#7459): `role="alert"`, `data-testid="chart-missing-category-axis"`, naming the bindings the author can declare.

## The stop condition — enumerated, with its verdict

The PM ruling made this renderer-level, and named the one place it could be wrong: `ObjectChart` is reached by producers that supply `'name'` / `'value'` themselves, so a renderer-level refusal fires on those too. Every producer that reaches this component with a legacy binding, and whether the refusal would fire on a configuration that renders correctly today:

| # | producer | category it hands over | refusal fires? |
|---|---|---|---|
| 1 | `plugin-list/ListView.tsx:2768` `case 'chart'` legacy leg | `chartBinding.categoryField \|\| 'name'` | no — floored |
| 2 | `plugin-view/ObjectView.tsx:1536` `case 'chart'` legacy leg | `chartCfg.xAxisField \|\| chartCfg.categoryField \|\| 'name'` | no — floored |
| 3 | `app-shell/views/ObjectView.tsx:2218` chart viewDef legacy leg | `chartConfig.xAxisField \|\| 'name'` | no — floored |
| 4 | `plugin-dashboard/DashboardRenderer.tsx:620` object-provider widget | `options.xField \|\| 'name'`, assigned **before** the provider is consulted | no — floored |
| 5 | `plugin-dashboard/DashboardGridLayout.tsx:241` object-provider widget | `options.xField \|\| 'name'`, same unconditional assignment | no — floored |
| 6 | directly authored `object-chart` metadata | whatever the author wrote | only when the author declared no category by any spelling — which is the defect, not a working chart |

**Verdict: none. Proceeding.** Producers 4 and 5 are the dashboard widgets the card flagged. Their `|| 'name'` floors are *not* part of #7547's six (those live only at 1-3) and are untouched here, so the two dashboard surfaces cannot reach this refusal at all — before or after #7547.

Both of the legs 4/5 take are covered, including `aggregate: undefined` (the raw-record leg, where `effectiveAggregate = providerAgg ? { … } : undefined` genuinely yields no aggregate and `xAxisKey` is the only thing keeping it out).

Pinned rather than asserted: `describe('objectui#8168 stop condition — the refusal fires on NO producer that renders today')` is one `it` per producer leg, each transcribed from its composing source with the line cited, so a widened predicate reports which producer it would have broken by name.

## Why it keys on the CATEGORY alone

`resolveListChartBinding`'s `resolves` is `Boolean(categoryField && valueField)`, so its negation refuses when **either** is missing. That is the right question for a capability gate deciding whether to *offer* a chart and the wrong one for a renderer deciding whether to *draw* one: a measure may legitimately be absent, because `count` takes no field (`aggregateValueKey` projects it under the literal `'count'`). Refusing on an absent measure would refuse `count` grouped by a declared category — a chart that renders correctly today, and one that an existing test (`ObjectChart.elementDataSource.test.tsx:115`, `xAxisKey: 'name'` with no series and no aggregate) already exercises.

This is `ObjectTimeline`'s clause 2 in this file's vocabulary — that refusal "keys on the START axis alone", for the same reason.

## Reuse of the predicate

Reused in substance; it could **not** be called. `resolveListChartBinding` reads `{ chart, options.chart }` with `xAxisField` / `categoryField` / `yAxisFields` — the block an author writes on a *list view*. By the time a schema reaches `ObjectChart` the relays have already translated that into `{ objectName, aggregate, xAxisKey, series }`, and producers 4-6 never had such a block at all. Asked for `schema.chart` here it would read `undefined` off **every** chart this component renders and answer "unresolvable" for all of them.

What is reused instead is the anti-drift move itself. "Which column is the category" was spelled **three** times in `ObjectChart.tsx` at three fidelities, and this refusal would have been a fourth. `resolveChartCategoryField` states it once; the option-metadata probe now reads it too, so the probe and the refusal cannot disagree. It is module-scoped, not added to the package index — no new public surface.

Its third leg asks `normalizeChartSchema` (this package's one translation of the spec's author-facing shape) rather than reading `schema.xAxisKey` directly. That is load-bearing: `ChartRenderer` normalizes **downstream** of this component, so a chart authored as `xAxis: { field: 'stage' }` has no `xAxisKey` here and draws correctly — a naive predicate would have refused it. Covered by its own arm.

## Deliberately untouched

The six `'name'` / `'value'` floors at the three relay faces. That is the remainder of #7547 and is mechanical only once this screen exists; `git diff --name-only` touches none of those three files, and all six literals are verified present on this branch's HEAD. (The naive `|| 'name'` / `|| 'value'` grep over those three files returns **19** hits on `7444916ce`, of which only 6 are chart — the other 13 are `titleField` / `labelField` floors for timeline, gantt, gallery, kanban, map and tree.)

## Clause ② — what the renderer now accepts, stated not assumed

**Breaking, deliberately; shipped `minor` per this repo's no-`major` rule.** A chart that previously rendered an `'Unknown'`-bucketed bar — or whatever the driver did with `groupBy: undefined` — now renders the refusal instead. That is the intent: the picture it drew was not a picture of the data. Four shapes are explicitly still accepted and each has its own arm: an ADR-0021 `dataset` chart declaring no dimension; a chart carrying authored `data` or a `bind` scope; a spec-shape `xAxis` (object or bare string); and a `count` aggregate with a category and no measure.

## Tests

Ran from the repo root — `pnpm --filter <pkg> exec vitest` is **refused** in this repo by an explicit guard (objectui#3378: it re-roots vitest at the package dir and silently runs `apps/console`'s 22 files instead, reporting `22 passed`).

| run | result |
|---|---|
| `packages/plugin-charts/` + `packages/i18n/` | 111 files, 1505 tests passed |
| `packages/plugin-dashboard/` + `packages/runner/` + `packages/plugin-list/` + `packages/plugin-view/` | 205 files, 2049 tests passed |
| `apps/console/…/public-block-binding-reach` + `app-shell/…/widget-dom-leak-sweep` | 2 files, 218 tests passed |
| `type-check` (plugin-charts, i18n) | both `Done`; `Scope: 2 of 47` and both script names echoed, so no zero-match false green. Each is `tsc --noEmit && tsc -p tsconfig.test.json`, so the new test file is typechecked |
| `turbo run lint` (plugin-charts, i18n) | 3 tasks successful, **0 errors** (345 + 33 warnings; `--max-warnings` is deliberately unset repo-wide) |

The two sweeps are there because they were the stop condition's live candidates. `public-block-binding-reach` builds its probe from a block's declared registry `inputs` — for `object-chart` that is `{ objectName, filter, aggregate: {}, drillDown: {} }` with no `xAxisKey`, so the refusal **does** fire on it; it stays green because the refusal is a render-time early return and the fetch effect is a hook that still runs, so `objectName` still reaches the data layer. `widget-dom-leak-sweep` authors `data` on top of `objectName`, which the authored-rows carve-out covers.

Gates, each read from its own printed verdict with the exit code captured before any pipe:

| gate | verdict |
|---|---|
| `check-changeset-presence.mjs` | `0` — "11 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)" |
| `check-changeset-no-major.mjs` | `0` — "No changeset declares a `major` bump" |
| `check-control-bytes.mjs` | `0` — 6557 tracked text files scanned |
| `check:i18n-keys` | `0` — every call-site key resolves against the en pack; every inline `defaultValue` matches the value the pack serves |
| `check:i18n-drift` | `0` — "0 en value(s) changed (1 key(s) added, 0 removed)" |
| `check:i18n-dead-keys` | `0` (a report, not a gate, per its own header) |
| `check-vi-mock-specifiers` / `check-vi-mock-inherit` | `0` / `0` |
| `check-governed-queue-guard --test <13 paths>` | `NOT GOVERNED — none of 5 surfaces matched` |
| `check-readme-exports.mjs` | **not measured** — exits 1 with "type entry `./dist/index.d.ts` is not on disk — run `pnpm build` first" for 400 self-imports across app-shell/auth/cli/collaboration/…; a whole-repo-build prerequisite, and none of the unjudged imports name a file in this diff. CI's |
| `check:eager-closure` | not run — needs a console build. Ten short locale strings; the packs now live in the `i18n-locales` chunk (ceiling 455,000, headroom ~8,924 bytes ≈ sixty keys at the measured ~147 gzipped bytes each) |

Ablation — the refusal is proven able to fail:

- Mutation: the guard's condition prefixed with `false &&`. **On-disk proof**, not an exit code: anchor `grep -cF` 1 → 0, mutant 0 → 1, `git hash-object` `611bd182…` → `4ce5ea38…`.
- Result: **5 assertions red, by name** — `refuses an object-bound chart that declares no category by any spelling`, `refuses a legacy aggregate whose measure is declared but whose groupBy is not`, `refuses a structured GroupBy node that names no field`, `is a static authoring fact: it does not wait for, or depend on, the fetch`, `is distinguishable from the failure state and from the empty state`. The 17 "does not fire" arms stayed green, which is the expected asymmetry: they assert an absence the mutation also produces.
- Restore **by state**: `git diff HEAD` → 0 files, and `git hash-object` `611bd1823be0ecd1ed82ec805a831da44c0a99bc` == `git rev-parse HEAD:packages/plugin-charts/src/ObjectChart.tsx`. The script carried a `trap … EXIT INT TERM` with an absolute path derived from `git rev-parse --show-toplevel`.

Draft on purpose — the PM lands it. Not flipped ready, auto-merge not enabled.

---


os-justin:
## PM ruling: **A** — refuse on the category axis alone. Ship as built.

Three candidates, and only one survives measurement:

- **B — mirror `resolveListChartBinding.resolves` exactly** (refuse when *either* category or measure is absent). ⛔ **Refuted, not merely disliked**: it would turn the existing `ObjectChart.elementDataSource.test.tsx:115` red, because `count` legitimately takes no field, so a measure may be absent by design. B would refuse a count-grouped-by-a-declared-category chart that renders correctly today — which is precisely what the stop condition exists to prevent.
- **C — refuse only when both are absent** (the card's literal prose). Leaves the card's own named defect unrefused: `{aggregate: {field: 'amount', function: 'sum'}}` with no `groupBy` still aggregates on a name nobody wrote.
- **A — refuse when no category resolves.** The only option that is safe under the stop condition for every producer *and* makes objectui#7547's floor retirement honest.

## The stop condition: PASS, and it was **run**, not reasoned

This is the part that decides the ruling. The dispatch said: enumerate every producer reaching `ObjectChart` with a legacy binding and stop if any would be refused on a configuration that renders today. Six were enumerated and three live candidates were executed rather than argued about — and one of them mattered:

⭐ the spec-shape `xAxis: { field }` resolves through `normalizeChartSchema` **downstream** of `ObjectChart`, so **a naive `xAxisKey` read would have refused a working chart.** That is exactly the failure the stop condition was written to catch, caught by running the case instead of reasoning about it. The dashboard object-provider widgets were also checked properly: `xAxisKey = options.xField || 'name'` is assigned **unconditionally**, before the `isObjectProvider` branch, so both legs carry a category — including the raw-record leg where `aggregate` is genuinely undefined.

## ⚠️ Corrections to the dispatch, all accepted — one of them was load-bearing

The brief said to reuse `resolveListChartBinding` because "the question is answered upstream and only needs a screen." **That is wrong, and the card said it first — I repeated it without checking.** `resolveListChartBinding` reads `{chart, options.chart}`, the authored list-view block; the relays translate that into `{objectName, aggregate, xAxisKey, series}` before `ObjectChart` ever sees it, and producers 4-6 never had such a block at all. Called there it would read `undefined` for **every** chart and refuse all of them. The answer does not travel downstream.

Reusing the *anti-drift move* instead — one `resolveChartCategoryField`, module-scoped, replacing three spellings at three fidelities, with the `optionMeta` probe reading it too — is the right execution of what that instruction was actually for.

The other corrections are accepted as stated and are the dispatch's errors, not the implementation's:

- the naive floor grep returns **19** hits with **13** non-chart `titleField`/`labelField` floors, not 14 with 8 — the over-reach warning was right in direction and understated in size;
- `ObjectCalendar`'s and `ObjectGantt`'s refusals carry **no** testid and no role, so "the four views read alike" was aspirational for two of them;
- inside `plugin-charts` the local refusal idiom is `role=status`, which points the other way from `ObjectTimeline`'s `role=alert` — following `ObjectTimeline` as instructed is still right here, and for a better reason than the instruction gave: `chart-empty-state` already owns `role=status`, so `role=alert` keeps the machine check against `chart-error` meaningful;
- ⚠️ and one worth the whole repo's attention: **`pnpm --filter PKG exec vitest run FILE` is refused by an explicit guard (objectui#3378)** — it re-roots vitest at the package dir and reports a pass count having run **none** of the target files. A silent false green. Run from the repo root.

## Scope held

The six `'name'` / `'value'` floors at the three relay faces are **untouched** and verified still present — retiring them is objectui#7547's remainder, in its own change, now that the refusal exists. The thirteen non-chart floors were not touched either.

## Findings carried forward

The three out-of-scope findings could not be filed — `search_issues` failed twice on the shared-identity rate limit, and an undeduped zero is not a reading, so they were reported rather than opened blind. That was the correct call. The PM will file them, in particular **FINDING 1**, which is a live defect independent of this card: a `provider: 'object'` chart widget with `{function: 'count', groupBy: 'status'}` and no `field` plots **nothing**, because `effectiveYField` falls to `'value'` while `aggregateValueKey` projects rows under the literal `'count'`.

No changes requested. `mergeable_state` is `behind`; the merge queue rebases. Landing on convergence.

---

PR #8247, opened 2026-09-07
test(types): pin the zod-mirror-parity ledger docstrings' key total, anchor their seed history (objectui#8243)

### LINKED ISSUE #8243: finding(types): zod-mirror-parity's two ledger docstrings say UnmirroredDeclared records 94 keys; two instruments measure 87 — unmoved since objectui#7779, and outside every pin in the file
Filed by the dev seat implementing #8222 (branch `claude/pin-zod-mirror-parity-key-totals-8222`, PR #8242), unassigned and unlabelled — grading and `domain:*` are triage's. Out of #8222's scope **by that card's own boundary** (its subject is the file's top-of-file HEADER; these two sites are ledger docstrings), so it is reported rather than repaired.

## The numbers

`packages/types/src/__tests__/zod-mirror-parity.test.ts`. Two docstrings state how many keys `UnmirroredDeclared` records, and both say **94**:

1. Above `interface UnmirroredDeclared` — the section headed `⚠️ READ THIS BEFORE QUOTING THE NUMBER — 121 became 98 by RECLASSIFICATION, then 97 by the first REPAIR, then 96 by RETIREMENT, then 94 by REPAIR again`:
   > objectui#6058 seeded this ledger at **121 keys**. It records **94**, and the movements are different facts.

   and, in the same block:
   > So: **94 is the mirroring debt; 94 − 1 seeded + 3 mirrored + 2 retired + 23 reclassified is what "121" used to mean**

2. Above `interface RuntimeOnlyDeclared` — the section headed `⚠️ THIS IS WHERE 23 KEYS WENT`:
   > `UnmirroredDeclared` above was seeded at **121 keys** by objectui#6058. It records **94** today

| quantity | prose | measured on `7444916ce` |
|---|---:|---:|
| `UnmirroredDeclared` keys | **94** (twice) | **87** ❌ |

The file is fully green with both. The section HEADINGS carry the same stale chain (`… then 94 by REPAIR again`).

## Why nothing caught it

objectui#7279's pin (`the split figures and the totals the header writes down equal the ledger`) reads four figures plus a totals line, all of which are **correct today** and derived — `**SPEC-DERIVED (2 entries, 3 keys)**`, `**LOCAL (12 entries, 84 keys)**`, `**14 entries / 87 keys** — 2 / 3 spec-derived, 12 / 84 local`, and the header's `**14 entries** … **87 keys**`. The two `94`s sit in *different paragraphs of the same docstrings* and are outside every `headerFigures()` spelling that pin uses. objectui#8222's new pin does not reach them either — its regexes are anchored to the top-of-file header.

`87` is not new: objectui#7779 took the ledger 96 → 87 on 2026-09-06 and correctly moved every pinned figure. These two unpinned restatements did not move with it. That is the same shape as objectui#7664 → #8222 and objectui#7542 → #7733, one site further in.

## Measured, with a positive control

Two independent instruments over the ledger interfaces — a TypeScript AST walk summing each property signature's string-literal union arms, and a line-oriented parse of the same block with comments stripped — both read `UnmirroredDeclared` = **14 entries / 87 keys** on `7444916ce`.

**Control**: the same instruments read `RuntimeOnlyDeclared` 7 / 24 and `KnownDrift` 41 / 63, and the 87 is independently derived by objectui#7279's own green pin through `ledgerEntryMembers()`. So the semantics in use are the ones the file already treats as authoritative.

## The decomposition beside it also needs re-deriving, and it is the harder half

`94 − 1 seeded + 3 mirrored + 2 retired + 23 reclassified` is an accounting of what became of the original 121. Measured on the file's history in this container's (shallow) checkout:

- at `beccf1c6b`, the oldest revision of this file reachable here — after the 23 moved out, before the first repair — `UnmirroredDeclared` holds **98** keys, which is exactly the `121 became 98 by RECLASSIFICATION` the docstring states. That agreement is the control for this walk.
- of those 98 keys, **85 survive** into today's set and **13 have left** (set intersection on `entry.key` pairs).
- so the true statement today is of the shape `87 − 2 seeded + 10 mirrored + 3 retired + 23 reclassified = 121`, where 85 + 13 + 23 = 121 and the two seeded keys are `ChatbotFloatingSchema`'s pair from objectui#7655.

⚠️ **Stated as a reading, and one term is NOT measured**: the split of the 13 departures into *mirrored* vs *retired* is read off the prose history (objectui#6639 mirrored 1, objectui#7129 retired 1, objectui#7623 retired 1, objectui#7352 mirrored 1 of the two `drillDown` rows that was part of the seed, objectui#7779 mirrored 8 and retired 1 = 10 + 3). It reconciles arithmetically, but it is prose-derived. The **13** is measured; the **10 / 3** is not.

## Suggested remedy, and the fence around it

⛔ Not route 1 — correcting `94` → `87` alone is the move this file's history is a record of.

The 87 half is mechanical and pinnable: it is the number objectui#7279's pin already derives, and both docstrings would be read by a `headerFigures()` spelling the same way the top-of-file header's copy is. Note both section HEADINGS restate the chain and would need the same treatment, or the same rewrite that stops restating it.

⚠️ **The decomposition half is NOT pinnable and should not be pinned.** Every term but `23 reclassified` needs each key's **provenance** — which of today's keys descend from the seed — and no ledger in this file records provenance; they record entry → key set as it stands. #8222 hit the identical figure in the top-of-file header and **removed** the copy rather than refreshing the digit, leaving an explicit exclusion with its reason and deferring to these ledger paragraphs as the site that owns the statement. ⇒ whoever takes this decides whether the owning site keeps a live decomposition at all; if it does, it will need re-deriving by hand on every future repair, and that cost is the decision.

## Duplicate check

`GET /search/issues` is refused by this session's egress proxy, so this ran as the declared fallback: the repo-scoped REST list endpoint (`state=all`, 6 pages of 100, issues #7644–#8241, 321 issues after dropping PRs) plus a local grep. **The control fired**: `zod-mirror-parity` returns 10 issues including #8222 and #7733, so the scan is not blind. The targeted pattern (`mirroring debt`, `records **94**`, `121 used to mean`, `ledger docstring`) returns **0**. The near ones and why none is this card: #8222 (the header's key totals — this seat's card, and these two sites are explicitly outside it), #7733 (the header's entry counts), #7947 (`NamedListView` member figures, a different file). ⚠️ The window is #7644 upward; a card older than that would not be seen, though the defect itself only became wrong at objectui#7779 on 2026-09-06.

Refs: #8222 · #8242 (the PR for the header half) · objectui#7279 (the pin whose spelling the remedy would reuse) · objectui#7779 (the change that moved the ledger and not these two figures)

---

os-sam:
Fixes #8243

Both LEDGER DOCSTRINGS in `packages/types/src/__tests__/zod-mirror-parity.test.ts` — the
paragraph above `interface UnmirroredDeclared` and the one above `interface
RuntimeOnlyDeclared` — said that ledger records **94 keys**. Two independent instruments
read **87**. The section HEADINGS carried the same stale chain, and the file was fully
green with all of it: objectui#7279's pin is anchored to the split's own sentences and
objectui#8222's to the top-of-file header, and these two figures sit in different
paragraphs, outside every spelling either one reaches.

`87` is not new. objectui#7779 took the ledger 96 to 87 and correctly moved every
**pinned** figure; these two unpinned restatements did not move with it. Fifth link in
objectui#6141 to objectui#7433 to objectui#7733 to objectui#8222.

## The decision this card posed, and what was done instead of choosing

The card put it honestly: whoever takes it decides whether the owning site keeps a live
decomposition at all, and if it does, it needs re-deriving by hand at every future
repair. Sharper than it looks, because objectui#8222 had already **removed the header's
copy and deferred to these paragraphs** as the site that owns the statement — so
deleting here would take the provenance narrative out of the repo entirely.

Neither route was taken. objectui#8242's own organizing distinction was applied instead:
a figure is **live** when it describes the file as it is now; a figure naming a reading
at a **named past revision** is **historical** and cannot rot.

| half | before | after |
|---|---|---|
| "records **94**", twice, plus the headings' chain | live, wrong, unpinned | **corrected to 87 and PINNED**, each through its own `headerFigures()` spelling read off the prose that already exists |
| `94 − 1 seeded + 3 mirrored + 2 retired + 23 reclassified` | live, stale, unpinnable | **rewritten as historical**, anchored to `beccf1c6b` and `ed7178bf3` |

A historical statement cannot rot, needs no pin against the ledger, and needs no hand
re-derivation at the next repair. The narrative is kept and the maintenance cost is
gone.

**The headings could be made historical without losing what they are for**, so they
were: every figure in the chain now names the card that produced it (121 at
objectui#6058 through 87 at objectui#7779), and the heading says in as many words that
none of them is today's. The chain is also declared abridged — it names one card per
step and skips three movements the neighbouring steps absorb — which is true and was not
stated before.

⭐ **No new constant.** An `EXPECTED_UNMIRRORED_KEYS` would be a third site for the
number and would reproduce this defect one level up. The pin reads the prose.

## The measured / prose-derived split, preserved

Re-derived independently rather than taken from the card, and it reproduces exactly:
at `beccf1c6b` (this repository's oldest reachable commit, after objectui#6152's
reclassification and before objectui#6639's first repair) the ledger held **98 keys** —
which is exactly the `121 became 98` the docstring itself states, the positive control
for the walk. Of those 98, by set intersection on entry-to-key pairs at `ed7178bf3`,
**85 survive and 13 have left**; `85 + 13 + 23 = 121`.

⚠️ The **13 is measured; the 10 MIRRORED / 3 RETIRED split of it is not** — it is read
off the cards' prose. The file says so explicitly, and the pin deliberately does **not**
assert it: asserting it would dress a prose reading as a measurement. (The attribution
does corroborate against the tree — all eight objectui#7779 keys are real mirror slots
today, `viewTabBar` is a `retirementTombstone`, and the two retired declarations are
gone — but corroboration is not the same as measurement, so the file still calls it
prose-derived.)

Also newly stated, and measured: the `23` in the `RuntimeOnlyDeclared` heading is
**historical** and that ledger no longer holds exactly those 23 —
`TreeViewSchema.onNodeClick` arrived afterwards. The file now warns against "correcting"
it, and deliberately does not restate the live total there, since that would add an
unpinned live figure.

## Verification — every new assertion proven to fire

Six mutation legs. Each one: mutate on disk, prove the mutation landed (injected-text
count 1, removed-text count 0, **and** `git hash-object` differing from the HEAD blob),
run, restore, prove the restore **by state** (`git hash-object` equal to the HEAD blob
**and** `git diff HEAD` empty) — never by an exit code. The implementation was committed
first, so the restore leg had a real reference point. The driver carried a
`trap ... EXIT INT TERM`.

| leg | mutation | result |
|---|---|---|
| live-site1 | `87` to `86` in the `UnmirroredDeclared` docstring | new pin RED; objectui#7279 / #7733 / #8222 all green |
| live-site2 | `87` to `86` in the `RuntimeOnlyDeclared` docstring | new pin RED — proves **both** sites are read, so a stale copy cannot survive beside a corrected one |
| ⭐ key-total-moves-entry-count-holds | one entry's union grown from 3 arms to 4 (87 keys to 88, **14 entries unchanged**) | new pin RED **and** objectui#7279 RED, while **objectui#7733's entry-count pin stayed GREEN** — the objectui#7664 shape reproduced, and the leg objectui#8242 turned on |
| rewording-drops-the-digit | the pinned sentence reworded to spell the number as a word | new pin RED through `headerFigures`' exactly-once rule — a rewording that drops the digit is red, not silently unpinned |
| historical-survivor-digit | `85` to `86` in the historical reading | historical pin RED, live pin GREEN |
| historical-sum-line | `85 + 13 + 23 = 121` to `= 122` | historical pin RED |

⭐ The third leg is also the **negative control for the historical half**: moving the
live ledger reddened the live pin and left the historical one green. That is the direct
evidence that nothing in the historical pin reads the live ledger — which is the whole
reason it cannot rot.

**No `dist` leg to prove, and this was established rather than assumed.** The subject is
this file's own source: `SELF` is `fileURLToPath(import.meta.url)`, and the failure
output names it as
`/home/user/objectui-issue-8243/packages/types/src/__tests__/zod-mirror-parity.test.ts`
— the source path in the worktree. The file imports only vitest, node builtins,
`typescript`, and sibling `../zod/*` sources; `@object-ui/types` has one workspace
dependency (`@object-ui/test-support`) and it has **no `build` script**, so
`pnpm --filter '@object-ui/types^...' build` fails loudly with
`ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT` rather than passing vacuously — there is no
dependency closure to build. Empirically confirmed as well: every mutation landed with
**no build step at all** and the target rows went red, which a built-artifact subject
could not do. Same case as objectui#8242 and objectui#8224, not objectui#8226.

**`--listFiles`, and it matters here.** Of the three programs `type-check` runs, only
one contains this file:

| program | files | hits for this file |
|---|---:|---:|
| `tsc --noEmit` | 275 | **0** |
| `tsc -p tsconfig.examples.json` | 275 | **0** |
| `tsc -p tsconfig.test.json` | 598 | 2 |

So the first two legs say nothing about this change; the third is the one that covers
it, and it is green.

## Gates

Union run at the final commit `e1c8f4e31`, one `&&`-joined chain with **no pipes** (a
trailing pipe would have made the verdict cover only the last part), verdict read from
the lock's own VERDICT line rather than a bare status:

- `pnpm exec vitest run packages/types/` — **137 files / 2598 tests passed**
- `pnpm --filter @object-ui/types type-check` — clean
- `pnpm --filter @object-ui/types lint` — **0 errors** (272 pre-existing warnings, none
  from the changed file: it produces zero eslint findings)
- `check-control-bytes`, `check-changeset-presence`, `check-changeset-no-major` — all OK

Also run: `check:unreferenced-sources`, `check:phantom-deps`, `check:self-import`,
`check:published-tsconfig-exclude`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`,
`check:shell-escape-residue`, `check:doc-fences` — all exit 0. Governed-surface guard
answers **NOT GOVERNED** for both changed paths. Changeset is the empty-frontmatter form
this repo uses for test-only changes, and the presence gate names that as the explicit
exemption.

## What was deliberately not touched

- No ledger entry, `MIRRORS` entry or `EXCLUSIONS` row moved. Additive assertions plus
  prose only.
- objectui#7279's and objectui#8222's assertions are untouched. Neither failed on the
  final tree.
- One knock-on worth flagging: objectui#8222's assertion message calls the deferred-to
  seed decomposition a **live** figure, which it was when written and is not any more.
  That message was left exactly as it stands — it is another card's assertion, and what
  it says about the file header is still correct. The correction is a pointer added in
  this card's own block instead.

---

### DIFF
--- a/.changeset/pin-ledger-docstring-key-totals.md
+++ b/.changeset/pin-ledger-docstring-key-totals.md
@@ -1,0 +1,15 @@
+---
+---
+
+Test-only change to `@object-ui/types`' zod-mirror-parity pin file. Both LEDGER
+DOCSTRINGS — above `interface UnmirroredDeclared` and above `interface
+RuntimeOnlyDeclared` — stated that ledger records **94 keys**; two independent
+instruments measure **87**, wrong since objectui#7779 and fully green the whole time
+(objectui#8243). Those two live figures are corrected and are now derived from the
+ledger through the docstrings' own spelling. The stale decomposition beside them
+(`94 − 1 seeded + 3 mirrored + 2 retired + 23 reclassified`) needs key provenance no
+ledger here records, so instead of being refreshed or deleted it is rewritten as a
+reading at NAMED REVISIONS — 98 keys at `beccf1c6b`, of which 85 survive and 13 have
+left at `ed7178bf3` — which cannot rot and needs no hand re-derivation at the next
+repair. The measured 13 and the prose-derived 10 / 3 split of it are kept visibly
+apart. No published behaviour changes.
--- a/packages/types/src/__tests__/zod-mirror-parity.test.ts
+++ b/packages/types/src/__tests__/zod-mirror-parity.test.ts
@@ -1227,8 +1227,8 @@ interface KnownDrift {
  * pin below reads this sentence off disk —
  * **87 keys** is what this ledger records today.
  * The movements between the two are different facts. objectui#6152 measured the 23
- * callback-shaped
- * (`on*`) keys and ruled that mirroring is the wrong remedy for every one of them;
+ * callback-shaped (`on*`) keys and ruled that mirroring is the wrong remedy for every
+ * one of them;
  * they moved, intact and still pinned, to `RuntimeOnlyDeclared` below — ⛔ nothing
  * was mirrored, no declaration was removed, nothing was waived by that move. Then
  * objectui#6639 MIRRORED `ObjectGridSchema.title` (census-directed maintainer
@@ -1623,8 +1623,7 @@ interface UnmirroredDeclared {
  * `UnmirroredDeclared` records **87 keys** today.
  * These 23 moved here whole. Keys have since left that ledger by MIRRORING and by
  * RETIREMENT, but the move recorded HERE is neither and repaired nothing. ⛔ Nothing
- * was
- * mirrored by it, no declaration was removed, no defect was repaired and nothing was
+ * was mirrored by it, no declaration was removed, no defect was repaired and nothing was
  * waived: the same 23 facts are still measured, still declared-but-unmirrored, still
  * reconciled against the same measurement — under a different remedy.
  *
--- a/packages/types/src/__tests__/zod-mirror-parity.test.ts
+++ b/packages/types/src/__tests__/zod-mirror-parity.test.ts
@@ -3162,6 +3162,13 @@ stale copy survives beside a corrected one, which is the shape objectui#8243 rec
     // the statement. So that site could not simply delete it too: the provenance
     // narrative would have left the repo.
     //
+    // ⚠️ objectui#8222's own exclusion note calls that deferred-to statement a LIVE
+    // figure, which it was when the note was written. It is not one any more — read it
+    // as the historical reading described below. That note is left exactly as it
+    // stands: it is an assertion message objectui#8243 has no business editing, and
+    // what it says about the FILE HEADER (no copy there, and none to be added back)
+    // is unchanged.
+    //
     // ⭐ objectui#8243 took neither route. The statement was rewritten as a reading at
     // NAMED REVISIONS — 98 keys at `beccf1c6b`, of which 85 survive and 13 have left at
     // `ed7178bf3` — which cannot rot, because nothing landing later can make a claim
--- a/packages/types/src/__tests__/zod-mirror-parity.test.ts
+++ b/packages/types/src/__tests__/zod-mirror-parity.test.ts
@@ -1214,11 +1214,20 @@ interface KnownDrift {
  * Exact DECLARED-BUT-UNMIRRORED key set per pair: keys the published TypeScript
  * invites an author to write and the published validator has never heard of.
  *
- * ## ⚠️ READ THIS BEFORE QUOTING THE NUMBER — 121 became 98 by RECLASSIFICATION,
- * ## then 97 by the first REPAIR, then 96 by RETIREMENT, then 94 by REPAIR again
+ * ## ⚠️ READ THIS BEFORE QUOTING THE NUMBER — every figure in this heading is a
+ * ## READING AT THE CARD THAT PRODUCED IT, never today's: 121 at objectui#6058, 98 by
+ * ## RECLASSIFICATION at objectui#6152, 97 by the first REPAIR at objectui#6639, 96 by
+ * ## RETIREMENT at objectui#7623, 94 by REPAIR again at objectui#7352, 87 by MIRRORING
+ * ## BY REFERENCE at objectui#7779. Abridged — it names one card per step and skips
+ * ## three movements the neighbouring steps absorb (objectui#6576 +1, objectui#7129
+ * ## −1, objectui#7655 +2); the unabridged chain is the `UnmirroredDeclared` bullet in
+ * ## the file header. ⛔ None of these is today's figure. Today's is one line down.
  *
- * objectui#6058 seeded this ledger at **121 keys**. It records **94**, and the
- * movements are different facts. objectui#6152 measured the 23 callback-shaped
+ * objectui#6058 seeded this ledger at **121 keys**, and — on ONE line, because the
+ * pin below reads this sentence off disk —
+ * **87 keys** is what this ledger records today.
+ * The movements between the two are different facts. objectui#6152 measured the 23
+ * callback-shaped
  * (`on*`) keys and ruled that mirroring is the wrong remedy for every one of them;
  * they moved, intact and still pinned, to `RuntimeOnlyDeclared` below — ⛔ nothing
  * was mirrored, no declaration was removed, nothing was waived by that move. Then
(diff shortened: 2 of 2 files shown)


os-sam:
## Seat review — accepted. The historical half is *proved* rot-immune, not asserted

### ⭐ Leg 3 does something I did not think to ask for

I asked for a leg moving the key total while the entry count holds — the #7664 shape. What it also delivered:

> moving the live ledger reddened the live pin and left the historical one green. That is the direct evidence that **nothing in the historical pin reads the live ledger** — which is the whole reason it cannot rot.

⇒ The recommendation I briefed was an *argument* that a historical statement cannot rot. This turns it into a **measurement**: the same mutation that reddens the live pin leaves the historical one untouched, so the independence is demonstrated rather than reasoned about. ⭐ That is the difference between a design claim and a verified property, and it is the thing that makes the shape safe to reuse.

### ⭐ The word-spelling leg closes #8242's own blind spot

`rewording-drops-the-digit` — the pinned sentence reworded to spell the number as a word, caught by `headerFigures`' exactly-once rule.

That is precisely the hole #8242 discovered the hard way: *"Six of the seven"* had never been read by any instrument because no numeric regex could reach it, and *"a figure spelled 'six' rots exactly as fast as one spelled `6`."* This proves the new pin is not vulnerable to the same evasion. ✅ A card fixing the defect its own predecessor uncovered, one link later.

### The measured / prose-derived split — held, and sharpened

The 13 departures were **re-derived independently** rather than inherited (98 keys at `beccf1c6b`, with the docstring's own "121 became 98" as the walk's positive control; 85 survive, `85 + 13 + 23 = 121`). And the split is not merely labelled — **the pin deliberately does not assert it**:

> asserting it would dress a prose reading as a measurement

⭐ Then it goes further and *corroborates* the 10/3 against the tree (all eight #7779 keys are real mirror slots, `viewTabBar` is a `retirementTombstone`, both retired declarations are gone) — while still refusing to promote it:

> corroboration is not the same as measurement, so the file still calls it prose-derived

That distinction is the one this whole chain keeps being bitten by, stated more precisely than I have managed to state it.

### ⭐ It declined to add a figure, which is the lesson landing

New and measured: the `23` in the `RuntimeOnlyDeclared` heading is historical, and that ledger no longer holds exactly those 23 (`TreeViewSchema.onNodeClick` arrived after). The file now warns against "correcting" it — and **deliberately does not restate the live total there, because that would add an unpinned live figure.**

⇒ The obvious instinct is to helpfully write down the current number. Refusing to, on the grounds that an unpinned live figure *is* the defect this chain is about, is the point being fully absorbed rather than followed.

### Two more things done right

- **The `dist` question answered twice**: by construction (`SELF` is `fileURLToPath(import.meta.url)`; the failure output names the worktree source path) *and* empirically — every mutation landed with **no build step at all** and the rows still went red, which a built-artifact subject could not do.
- **A knock-on flagged and not touched**: #8222's assertion message calls the deferred-to decomposition *live*, which it was when written and is not any more. Left alone — another card's assertion, and what it says about the header is still true — with the correction added in this card's own block instead. ✅ Correct restraint; editing it would have been reaching into a neighbour's pin to fix prose.

### Nit, cosmetic only

⚠️ The PR body carries a **tripled footer** (`🤖 Generated with …` plus two `_Generated by …_` lines). That is the PATCH-strips-the-footer behaviour the #7734 dev documented, showing up as duplication instead. Worth tidying only if the body is edited for another reason.

### The chain

#6141 → #7433 → #7733 → #8222 → **#8243**. Five links, four of which left a measured remainder for a later round. This one leaves none: every live figure in both docstrings is pinned, the unpinnable one is historical and proved independent of the ledger, and the one figure it could have "helpfully" added it deliberately did not.

---

os-sam:
## `Live E2E (informational)` — not this PR's. #7990, no fix to port.

Job `101616347111`, head `e1c8f4e31`. Eighth PR on this signature; recording it rather than letting it pass, kept to a pointer since the diagnosis is on #8210 (comment 5563689241).

Short form: published `@objectstack/plugin-auth` floats `@better-auth/core` past the removal of `createLocalAccountIssuer` ⇒ `AuthPlugin` fails to load ⇒ no `sys_*` tables ⇒ the readiness probe never satisfies. **#7990**, cause upstream at **objectstack#16186**; #7689's triage forbids repairing the lane by reverting the pin ⇒ nothing to port.

**Mechanically not this PR's**: two files, `packages/types/src/__tests__/zod-mirror-parity.test.ts` and a changeset. The failure is a backend boot from `e2e/live/ci/backend.env`, untouched here.

⛔ No re-run spent. The reproduction now spans `main`, #7685 at the **17.3.0** pin (a different published backend entirely), and #8076 / #8210 / #8224 / #8226 / #8230 / #8242 at 17.2.0 — two backend versions reaching the same auth-degraded state, which is a stronger argument than running one job twice.

---

PR #8259, opened 2026-09-07
test(console): re-point the record:activity.types calibration control at the empty string

### LINKED ISSUE #8137: Decision: `registry-inputs-spec-parity` fails on the 17.3.0 merge — a calibration control now asserts something the contract accepts, and the kind classifier may not recurse through `z.union`
Filed by the `domain:spec` @ objectui execution seat (session `session_0114Ytxr5sM1vdW19Y9WAx6E`), R1. Blocks PR #7685 (#7122's 17.3.0 pin). ⛔ Not decided here — one half sits on the manual floor.

## What fails

`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts`, **2 failed / 179 passed**, deterministic on a targeted re-run.

⚠️ **Merge-only, and not a resolution anyone could have made differently.** Green on `main` (spec 17.2.0), green on the branch (the gate did not exist there), red only in the merge. `main`'s `cf1d29e2f` (#8097, +1077 lines) added the `ComponentInput.of` gate; it **auto-merged textually** with this branch because the two sides edited disjoint regions.

## Root cause — measured on the installed artifact, not inferred

`@objectstack/spec` 17.3.0 changed `RecordActivityProps.types` to an **open** vocabulary:

```
z.array(z.union([FeedItemType, z.string().min(1)]))
```

That is objectstack#11658 executing the maintainer's 2026-08-24 ruling on #11507 — *"every closed map over this vocabulary is now the bug"*. `safeParse` against `@objectstack/spec/ui`:

```
['comment'] → ACCEPTS        [42] → refuses: invalid_union @ [0]
['Account'] → ACCEPTS        [[]] → refuses: invalid_union @ [0]
```

### 1. A calibration control that reality overtook — ⛔ manual floor

The test asserts `specMemberVerdict('record:activity', 'types', 'array', 'Account') === 'refuses-content'`. 17.3.0 genuinely **accepts** `'Account'`, so the assertion is now false. The test's own comment states the retired premise: *"`record:activity.types` is a spec enum of strings."*

⇒ The fix is to update a **calibration control**. That is the move the gate's own header forbids — and the protocol puts gate weakening on the manual floor precisely because an agent has a structural incentive to call weakening "correcting". ⭐ The distinction here is real but it is not mine to draw: the control is *factually wrong*, yet "the control is wrong" is also exactly what someone would say who wanted the gate quiet.

### 2. A probable classifier gap — likely an ordinary bug, reported for confirmation

`record:activity.types:of=string` fails the single-kind direction with the gate reporting **`contract accepts {string, array}`**. ⚠️ That reading disagrees with the contract as measured above: an array member **is** refused (`invalid_union @ [0]`).

Mechanism, from the dev that found it: `acceptedMemberKinds` counts a kind when the verdict is `accepts` **or** `refuses-content`, and `refusesKind` has to recurse through a `z.union` to answer correctly. If that is the whole story, this is a classifier that cannot see inside a union — not a genuinely multi-kind member — and the fix is in the gate's own helper, not in an exemption list.

⇒ Which it is decides the shape of the fix: a `MULTI_KIND_MEMBER_CONTRACTS` entry (declaring the member multi-kind) versus fixing `refusesKind` to recurse. **The first would be wrong if the second is true**, and it would also silently widen what the gate tolerates.

## Four-axis analysis

**Real business need** — `record:activity` is a shipped block; the open vocabulary is the maintainer's own ruling being executed upstream. objectui has to follow it either way. The question is only whether the gate that guards this follows it *correctly* or is quieted.

**Long-term soundness** — ⭐ This is the axis that decides it. If (2) is a classifier gap, then every union-typed member in the contract is currently mis-classified by this gate, and an exemption entry would hide that for all of them, not just this one. Fixing the helper restores the gate's meaning; exempting one member spends the gate's credibility to make one PR green.

**Preventing AI-written mistakes** — the calibration control exists *because* an agent will otherwise adjust the thing being measured. Whatever is decided, the control should end up asserting something still true and still discriminating — ⛔ not deleted, and ⛔ not softened to "accepts anything".

**Startup-stage focus** — both fixes are small and neither adds surface. No expansion pressure here.

## Options

- **A** — Rule that updating the calibration control is a *correction*, not a weakening, since 17.3.0 genuinely accepts `'Account'`; re-point it at a value the contract still refuses (`[42]` and `[[]]` both are), so it keeps discriminating. Then fix `refusesKind` to recurse through `z.union` and confirm the `{string, array}` reading collapses to `{string}`. **Recommended.** It keeps the control load-bearing and repairs the classifier rather than exempting around it.
- **B** — Same control update, but declare `record:activity.types` in `MULTI_KIND_MEMBER_CONTRACTS` instead of touching the classifier. ⚠️ Cheaper now, but if the classifier really cannot see into unions this hides the same defect everywhere else it occurs.
- **C** — Leave both red and land PR #7685 with a known failure. ⛔ Not viable: the queue requires every check green, and a red parity gate on `main` is shared damage.

## Related

#7122 / PR #7685 (the pin that surfaced it) · objectstack#11658 and objectstack#11507 (the open-vocabulary ruling) · #8097 / `cf1d29e2f` (the gate) · #8127 (a different 17.3.0 union-drift finding from the same round)

os-sam:
Fixes #8137

Re-points one member calibration control in
`apps/console/src/__tests__/registry-inputs-spec-parity.test.ts` whose premise an
upstream contract change retired, and rewrites the comment that stated that
premise.

## Manual floor

Editing a calibration control is on the manual floor. The authorisation was
given by the maintainer on 2026-09-07 against the exact diff, recorded verbatim
in issue comment 5565510994, and it covers **one row and its comment**. This PR
carries exactly that and nothing else.

Not touched, as the authorisation requires: the `refuses-kind` row for `42`
directly below (the other half of the pair), the three `page:header.actions`
rows and the `specArmVerdict` container control, every other expectation in the
file, and every gate ceiling or budget constant in the repo. No expectation was
flipped to `accepts` and no row was deleted.

## Why the row was wrong

`@objectstack/spec` 17.3.0 made `RecordActivityProps.types` an **open**
vocabulary — `z.array(z.union([FeedItemType, z.string().min(1)]))` — which is
objectstack#11658 executing the maintainer's 2026-08-24 ruling on
objectstack#11507. The row probed a closed enum with an unlisted word,
`'Account'`, and the widened contract now accepts it.

The row is the second half of a pair whose job is to prove the judge separates a
**content** refusal from a **kind** refusal one level down, so it needs a value
of the right kind with wrong content. Re-pointing it at `42` or `[[]]` would
have made it assert the same thing as the `refuses-kind` line below it and
destroyed that discrimination.

## The change

```diff
-    expect(specMemberVerdict('record:activity', 'types', 'array', 'Account')).toBe(
+    expect(specMemberVerdict('record:activity', 'types', 'array', '')).toBe(
       'refuses-content',
     );
```

plus the comment above it, which said "`record:activity.types` is a spec enum of
strings" — false since 17.3.0. It now states the open vocabulary, cites
objectstack#11658, and explains why the empty string is the surviving content
refusal.

## Evidence — measured here, not inherited

**Contract**, `ComponentPropsMap['record:activity'].shape.types.safeParse`, both
pins, each resolving the repo's own zod 4.4.3:

```
=====  17.2.0  (what main installs today)  =====
["comment"]   ACCEPTS
["Account"]   refuses  invalid_value@[0]     invalid_type anywhere: false
[""]          refuses  invalid_value@[0]     invalid_type anywhere: false
[42]          refuses  invalid_value@[0]     invalid_type anywhere: false

=====  17.3.0  (what #7685 pins)  =====
["comment"]   ACCEPTS
["Account"]   ACCEPTS
[""]          refuses  too_small@[0]         invalid_type anywhere: false
[42]          refuses  invalid_union@[0]     invalid_type anywhere: true
[[]]          refuses  invalid_union@[0]     invalid_type anywhere: true
["  "]        ACCEPTS
```

**The judge itself**, which the card flagged as still unmeasured — the step from
"no `invalid_type`" to `refuses-content` was a derivation, and this closes it.
Printed by `specMemberVerdict` under the 17.3.0 artifact:

```
{"empty":"refuses-content","fortyTwo":"refuses-kind","account":"accepts","comment":"accepts"}
```

So `''` really does land as `refuses-content` and the `42` row really does stay
`refuses-kind`. The pair still discriminates.

**Both pins, whole file** (`pnpm exec vitest run` on the changed file, at
`37b3a94c3`):

| pin | before | after |
|---|---|---|
| 17.2.0 — what `main` installs | 198 passed (198) | **198 passed (198)** |
| 17.3.0 — artifact-swapped | 8 failed, 190 passed | **7 failed, 191 passed** |

The 17.3.0 leg used the artifact-swap method from #8226: registry 17.3.0 under
the repo's own zod, the mutation proven on disk before any run
(sha256 `bb293113…`, version `17.3.0`), and the restore proven by sha256 of
`dist/ui/index.mjs` returning to `021bb509…` — verified again after cleanup.

## One reading that needs stating plainly

The 17.3.0 leg above is run on **`main`'s tree** with the artifact swapped, not
on #7685's tree. `main`'s ledgers are calibrated for 17.2.0, so seven other
rows go red there — they name spec-carried blocks `main` does not account for
(`cloud-connection:panel`, `marketplace:installed-list`, `mcp:connect-agent`)
and stale exemption entries. **All seven were already red before this change and
are unchanged by it**; the failing set after is exactly the set before minus
this card's block. They are #7122's pin-accommodation work and are already
handled on #7685's branch, which is why that branch's CI reported a single
failure. Nothing else in the file went red during this work.

That also means the "181/181" figure in the brief is a count on #7685's tree;
`main`'s copy of the file has 198 tests, hence the table above.

## Gates

- `pnpm --filter @object-ui/console lint` — exit 0, 0 errors (208 pre-existing
  `no-explicit-any` warnings, none in this file; the file alone lints 0/0).
- `tsc --noEmit --listFiles` on `apps/console` — **0 errors**, and the program
  really contains this file (1 hit among 3539 files). Worth recording: the brief
  warned that `tsc --noEmit` excludes tests on this repo — for `apps/console` it
  does not, its tsconfig is `include: ["src", "dev"]` with no exclude, so tests
  are genuinely type-checked here.
- `node scripts/check-changeset-presence.mjs` — green; the changeset is the
  empty-frontmatter form, declaring that no package is released.
- `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`,
  `check:shell-escape-residue`, `check:unreferenced-sources` — all exit 0.
- `check-governed-queue-guard --test` on both changed paths — NOT GOVERNED.

The dependency closure was built first
(`pnpm --filter '@object-ui/console^...' build`, exit 0) so the typecheck read
real `dist/*.d.ts` rather than reporting 392 phantom missing-module errors.

If `Bundle Analysis` reds here it is not from this PR — a test-only diff cannot
move the console bundle. See #7848 and #8241.

PR #7685 is untouched; it picks this up on its next `main` merge.

---

https://claude.ai/code/session_0114Ytxr5sM1vdW19Y9WAx6E

---

### DIFF
--- a/.changeset/8137-record-activity-calibration-repoint.md
+++ b/.changeset/8137-record-activity-calibration-repoint.md
@@ -1,0 +1,10 @@
+---
+---
+
+Re-point the `record:activity.types` member calibration control in
+`registry-inputs-spec-parity` from `'Account'` to `''` (objectui#8137).
+`@objectstack/spec` 17.3.0 made that vocabulary open —
+`z.array(z.union([FeedItemType, z.string().min(1)]))`, objectstack#11658
+executing the maintainer's 2026-08-24 ruling on objectstack#11507 — so the old
+probe now accepts, while `''` is still refused for its CONTENT via `.min(1)`.
+Test only; no package is released by this change.
--- a/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
+++ b/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
@@ -3144,11 +3144,21 @@ describe('registry `inputs` vs `@objectstack/spec` ComponentPropsMap (repo-wide)
 
     // …and the second half of the calibration: a CONTENT refusal at the member
     // position is NOT a kind refusal, so the coarse-kind ceiling survives one
-    // level down exactly as it does at the top. `record:activity.types` is a
-    // spec enum of strings — a string member is refused as a VALUE, and reading
-    // that as "the string member declaration is invented" would condemn a
-    // declaration derived from the contract itself.
-    expect(specMemberVerdict('record:activity', 'types', 'array', 'Account')).toBe(
+    // level down exactly as it does at the top.
+    //
+    // `record:activity.types` was a CLOSED `z.enum([...])` of strings when this
+    // row was written, and it probed that enum with an unlisted word.
+    // `@objectstack/spec` 17.3.0 made the vocabulary OPEN —
+    // `z.array(z.union([FeedItemType, z.string().min(1)]))`, objectstack#11658
+    // executing the maintainer's 2026-08-24 ruling on objectstack#11507 — so no
+    // ordinary string is refused there any more and the unlisted word ACCEPTS.
+    // The widening kept exactly ONE content refusal, and it is the one this
+    // control needs: `''` IS a string, so the only check it fails is `.min(1)`
+    // and it raises no `invalid_type` anywhere — right KIND, wrong CONTENT.
+    // That is exactly the distinction from the `42` line below, which is
+    // refused for its KIND, and it is why the pair still discriminates.
+    // Re-pointed under the maintainer 2026-09-07 authorisation on objectui#8137.
+    expect(specMemberVerdict('record:activity', 'types', 'array', '')).toBe(
       'refuses-content',
     );
     expect(specMemberVerdict('record:activity', 'types', 'array', 42)).toBe('refuses-kind');


os-sam:
## Seat review — accepted. It closed the derivation I left open and corrected two of my figures

### ⭐ The derivation is now a measurement

I handed this over with an explicit gap: I had measured the *contract's* verdict, not the judge's, so "no `invalid_type` ⇒ `refuses-content`" was a derivation from #8210's merged rule. This closes it by printing the judge:

```
{"empty":"refuses-content","fortyTwo":"refuses-kind","account":"accepts","comment":"accepts"}
```

⇒ `''` really lands as `refuses-content`, and the `42` row really stays `refuses-kind`. **The pair still discriminates**, which was the whole point of not re-pointing at `42` or `[[]]`. ✅

### It measured both pins, which is what made the change safe to land on `main`

| pin | before | after |
|---|---|---|
| 17.2.0 — what `main` installs | 198 passed (198) | **198 passed (198)** |
| 17.3.0 — artifact-swapped | 8 failed, 190 passed | **7 failed, 191 passed** |

⭐ And the residue is reported as a **set difference**, not a count: *"the failing set after is exactly the set before minus this card's block."* Seven rows stay red on `main`'s tree under a 17.3.0 artifact because `main`'s ledgers are calibrated for 17.2.0 — they were red before this change and are unchanged by it, and they are #7122's pin-accommodation work, already handled on #7685's branch. That is why #7685's own CI reported a single failure.

### ⚠️ Two of my figures were wrong, and it did not inherit either

1. **"181/181"** — that is the count on **#7685's** tree. `main`'s copy of the file has **198** tests. My brief carried #7685's CI number across to a different tree without saying so.
2. **My typecheck warning was wrong for this package.** I wrote that `tsc --noEmit` excludes tests on this repo — true for `packages/types` (where #8242 and #8247 found exactly that), but `apps/console`'s tsconfig is `include: ["src","dev"]` with no exclude, so tests *are* genuinely type-checked. It verified with `--listFiles` (1 hit among 3539 files) rather than taking my word for it.

⇒ Both are the same failure mode I have spent the night flagging in cards: a measured figure carried into a context where it no longer holds. Twice tonight it was mine.

### One observation, not a defect

The raw-contract table shows `[42]` on **17.2.0** refusing `invalid_value` with **`invalid_type anywhere: false`** — yet the `refuses-kind` row passes on that pin (198/198). ⇒ `specMemberVerdict` is evidently not reading the top-level array parse; it probes the member position directly, so the raw table and the judge are **different instruments**. Nothing here depends on that, and the 198/198 is the measurement that matters, but anyone reasoning from that table to a judge verdict should know the step is not direct.

### Manual floor, honoured

The authorisation covered one row and its comment. The diff is 2 files, +25/−5, and the PR names each thing it did not touch: the `42` row, the three `page:header.actions` rows, the `specArmVerdict` container control, every other expectation, every ceiling. No expectation flipped to `accepts`, no row deleted. ✅

⇒ Once this lands and #7685 merges `main`, that PR should reach **181/181** on its own tree and go green — after which the only thing left on it is the queue.

---

PR #8272, opened 2026-09-07
fix(plugin-dashboard,core): bind an object-bound chart's series to the column the aggregate projects

### LINKED ISSUE #8266: bug(plugin-dashboard): a `provider:'object'` chart widget with `{function:'count', groupBy:…}` and no `field` plots NOTHING — the series reads `'value'` while the rows are projected under `'count'`
Surfaced by the objectui#8168 dev seat while enumerating every producer that reaches `ObjectChart` (the stop condition on that card's dispatch). ⛔ **Independent of objectui#8168 and of objectui#7547** — neither causes it and neither fixes it; it was found by looking, not by changing anything. That seat could not file it: `search_issues` was returning `API rate limit already exceeded for user ID 323634890`, and an undeduped zero is not a reading, so it reported rather than opening blind. Filed here by the `domain:ui` PM seat after re-deriving the whole chain independently.

## The defect

A dashboard chart widget bound to an object, aggregating `{ function: 'count', groupBy: 'status' }` with **no `field`**, renders an empty chart. No error, no empty state — a chart frame with nothing in it.

## The chain, measured end to end on `origin/main` `0fa7a9c83`

Two independent copies, same shape:

```
packages/plugin-dashboard/src/DashboardGridLayout.tsx:227   const yField = options.yField || 'value';
packages/plugin-dashboard/src/DashboardGridLayout.tsx:239   const effectiveYField = effectiveAggregate?.field || yField;
packages/plugin-dashboard/src/DashboardGridLayout.tsx:246   series: [{ dataKey: effectiveYField }],

packages/plugin-dashboard/src/DashboardRenderer.tsx:607     const yField = options.yField || 'value';
packages/plugin-dashboard/src/DashboardRenderer.tsx:618     const effectiveYField = effectiveAggregate?.field || yField;
```

⇒ for `{ function: 'count', groupBy: 'status' }` the aggregate carries **no `field`**, so `effectiveAggregate?.field` is `undefined` and `effectiveYField` falls back to **`'value'`**.

Meanwhile the rows are projected under a different key:

```
packages/plugin-charts/src/ObjectChart.tsx:37-38
  export function aggregateValueKey(aggregate: { field?: string; function?: string }): string {
    return aggregate.field || aggregate.function || 'count';
  }
```

⇒ for the same aggregate this returns **`'count'`**. That is not an inference — it is **already pinned by a test on main**:

```
packages/plugin-charts/src/ObjectChart.aggregateResultColumns.test.ts:34
  it('is the literal "count" when a count names no field', () => {
    expect(aggregateValueKey({ function: 'count' })).toBe('count');
  });
```

⇒ **the series asks for `dataKey: 'value'`; every row carries `'count'`.** Nothing matches, nothing plots.

⭐ **Corroboration that the tree already half-knows this.** `DashboardRenderer.tsx:320` reads:

```ts
const isSynthetic = !yField || yField === 'value' || yField === 'count';
```

The **label** path explicitly handles the `'value'` / `'count'` duality. The **dataKey** path does not consult it. So the same duality is understood in one place and dropped in the other.

⚠️ **A second inconsistency, named but not claimed as part of this defect**: `DashboardGridLayout.tsx:261` builds `series: [{ dataKey: yField }]` — the **raw** `yField`, not `effectiveYField` — twenty-two lines below the site that uses `effectiveYField`. Whether those are two legitimately different branches or a third spelling of the same decision was **not** determined here. Whoever takes this should settle it rather than patching only `:239`/`:618`.

## ⚠️ Confidence — stated rather than implied

**Source-derived, NOT render-measured.** Every line above was read off `origin/main` and the `aggregateValueKey` half is pinned by an existing test, but **nobody has driven this in a browser**. The failure mode (`dataKey` naming a column the rows do not carry ⇒ an empty plot) is how the charting layer behaves, not something reproduced here.

⇒ **Whoever picks this up reproduces it first.** If it turns out something downstream reconciles the two keys, this card is wrong and should be closed saying so — that is a good outcome, not a wasted card.

## Why it is worth a card

The failure is **silent and credible**: an empty chart looks like "no data yet", which is exactly what an author with a genuinely empty object would also see. There is no signal distinguishing "your `count` aggregate is mis-keyed" from "this object has no rows" — the same silent-credible-wrong shape objectstack#13748 ruled against.

`count` with no `field` is also the *normal* way to author "how many records per status", which is the most common dashboard chart there is.

## Dedup

⚠️ **Not deduped — declared, not hidden.** The `search_issues` channel was rate-limited at both points where this would have been checked. An uncontrolled zero carries no information, so **no dedup claim is made**. Suggested query for a triager: `dashboard chart count aggregate no field empty plot dataKey value`. Adjacent context: objectui#8168 (the `ObjectChart` refusal path, which does **not** cover this — a count-grouped-by-a-declared-category chart passes that refusal correctly, and must, which is why objectui#8168's ruling keyed the refusal on the category axis alone).

Refs: objectui#8168 · objectui#7547 · objectstack#13748.

os-justin:
Fixes #8266

## It reproduces — measured before anything was written

The card was source-derived and said so; its own instruction was to reproduce first and close it if the chain did not hold. It holds, and here is the measurement rather than the derivation.

**Seam** (both relays, through the real render path, recorder registered for `object-chart`): a widget with `provider: 'object'` and `aggregate: { function: 'count', groupBy: 'status' }` composed `series: [{ dataKey: 'value' }]` on `DashboardGridLayout` AND on `DashboardRenderer`, over an aggregate the row projection keys `'count'`.

**Render** (`ChartRenderer` at a fixed 480x320, over `aggregateRecords(..., {function:'count', groupBy:'status'})` output `[{status:'open',count:2},{status:'paid',count:5}]`):

| `series[0].dataKey` | marks | refusal | empty state |
|---|---|---|---|
| `value` — what the relays composed | **0** `.recharts-bar`, **0** `.recharts-rectangle` | none | none |
| `count` — the column the rows carry | 1 `.recharts-bar`, **2** `.recharts-rectangle` | none | none |

Same rows, same harness, one key apart. The x axis drew `open` / `paid` in both. So the failure is real AND silent: a plot frame with the categories on it and nothing in it, which is what an author with a genuinely empty object also sees. Neither guard this renderer carries declines to fire by accident — `hasNoCategoryKey` is satisfied because the rows DO carry `status`, and `hasNoPlottableSeries` keys on `series: []`, which this is not.

## One correction to the card

The card says "No error, no empty state". That is true **only when `options.xField` is also authored.** Measured: with no `xField` the relays floor `xAxisKey` to the literal `'name'`, the rows carry `status`, and the chart renders the loud `missing-category-key` refusal — naming `name`, a key the author never wrote. That is the CATEGORY half of the same relay gap. This PR does not address it; it is filed separately, with its own measurement, as objectui#8269.

## The loose end at `DashboardGridLayout.tsx:261` — settled

It is a **legitimately different branch, not a third spelling.**

`:261` sits after the `isObjectProvider(widgetData)` early return, in the branch whose rows are an **authored literal array** (`DashboardRenderer` labels its twin `// Static inline data array.`). There is no aggregate there to consult, and the author's `yField` names a column in their own rows, so `yField` is the correct binding and must stay. Pinned in both directions by the second `describe` block of the new seam test — including that the branch composes `type: 'chart'` with `aggregate: undefined` — so a later "unify the two" refactor cannot quietly break it.

Corollary: the answer is one resolver for the question the aggregate path asks, and `:261` legitimately does not ask it.

## The fix — one authority, and it is the contract's

`chartAggregateValueKey` in `@objectstack/spec/ui` already IS the single answer to "which column does an object-bound aggregate project its measure under" — its own doc calls it "what a chart's series / y-axis binding must name". The repo had three independent opinions of that question and two of them disagreed for a fieldless count.

- New `@object-ui/core` export `chartMeasureKey(aggregate, fallback)` — a seam over `chartAggregateValueKey`, not a restatement of the rule. `core` is the home for exactly this (the `humanizeLabel` move in objectui#5444, `buildChartSeries` under ADR-0021) and already imports `@objectstack/spec/ui` at runtime, so no package gains a new dependency edge.
- Both dashboard relays now read it for the series binding.
- `aggregateValueKey` in `plugin-charts` — the ROW projection — is routed through the same function, so the two halves cannot drift again. Behaviour-preserving: the delegation's tail (`?? aggregate.function ?? 'count'`) reproduces the old three-rung read for the only shapes the contract answers `undefined` for, which are shapes `ChartAggregateSchema` rejects outright. Both are now pinned in the existing `ObjectChart.aggregateResultColumns.test.ts`.

The inverse fix was not available and was not attempted: making the row projection key a fieldless count `'value'` contradicts a pin already on `main` and renames the column every other consumer reads, including the engine's `COUNT(*)` alias.

## What moves on screen — stated, not assumed

A chart that was blank now draws. **No chart that already drew changes**: a field-bearing aggregate resolves to its raw field under both the old and the new reading, and a chart with no `aggregate` keeps the author's `yField` (pinned).

One authored key changes meaning: a `yField` written on an object-bound chart that ALSO declares an aggregate no longer wins over the aggregate's own column. It named a column of the RECORDS, and a grouped aggregate does not return records — so it plotted nothing before, for exactly the reason `'value'` did.

Committed first, then two disjoint mutations in one run, then restored **by state** (`git hash-object` equal to `git rev-parse HEAD:PATH` for both files, and `git diff HEAD` empty), never by an exit code. Each mutation was proved on disk by a `grep -cF` count of the removed text going to 0 and the injected text to 1 before anything ran.

Mutation A — `chartMeasureKey` reverted to the pre-fix reading `aggregate?.field ?? fallback`:

    × is the literal "count" for a fieldless count — NOT the caller floor
    × ignores a caller floor the author chose, when the aggregate answers
    × binds a FIELDLESS count to the column the rows carry, not to the yField floor   [grid relay]
    × ignores an authored yField that an object-bound aggregate cannot project        [grid relay]
    × binds a FIELDLESS count to the column the rows carry, not to the yField floor   [renderer relay]
    × ignores an authored yField that an object-bound aggregate cannot project        [renderer relay]

Both relays go red, which is what proves both of them route through the one authority rather than one of them happening to agree.

Mutation B — `aggregateRecords`' value key reverted to the raw `field`, the state before framework#3701:

    × is the shape the row builder emits — the premise the rest of the file rests on
    × draws its marks when the series names that column
    × keys a fieldless count under "count", not undefined

Totals: **9 failed, 22 passed** of 31. The 22 are the arms that must NOT move — the field-bearing aggregate, the no-aggregate floor, the authored-rows branch — so the ablation discriminates rather than reporting one undifferentiated red. `draws NOTHING, silently, when the series names "value"` stayed green under both, correctly: it is a pin on the renderer, not on either mutated path.

## Verification

Run from the repo root (the package-dir invocation is refused here by design, objectui#3378). Exit codes captured before any pipe.

- `pnpm exec vitest run packages/core/ packages/plugin-charts/ packages/plugin-dashboard/` — **267 files, 3902 tests passed**.
- `turbo run build` for both packages' dependency closures, then `turbo run type-check` for the three packages — 12 and 16 tasks, all successful.
- `node scripts/check-changeset-presence.mjs` — its own verdict line: 5 source files of 3 released packages changed, 1 changeset declared. `check-changeset-no-major` green (this repo ships breaking as `minor`; this is a `patch`).
- Gates: `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:spec-symbols`, `check:phantom-deps`, `check:self-import`, `check:unreferenced-sources`, `check:side-effects-array`, `check:element-data-source-declaration`, `check:entry-guard` — all exit 0. After a full `turbo run build` (44 tasks): `check:sdui-registration-pins`, `check:readme-exports`, `check:dist-completeness` — all exit 0.
- `eslint --no-inline-config` over the 9 changed TS/TSX files: **0 errors**, 89 warnings, of which 80 are pre-existing in the three large files this PR touches (64 / 8 / 8) and 9 are `no-explicit-any` in the two new test files, the same shape every schema-driven test in this repo carries. ESLint here is not type-aware (no `project` / `projectService` in `eslint.config.js`), so this diff cannot move the verdict on a file it did not touch. The whole-repo `turbo run lint` is CI's.
- Not run locally, declared: the other 35 packages `turbo ls --affected` names. They are affected only through the dependency graph — this diff adds an export and changes no existing signature — and CI runs the sharded farm.

## ⚠️ `Bundle Analysis` will be RED, and the reason is mostly not this PR

Measured in this worktree, same harness, two builds:

| | console eager closure, gzipped |
|---|---|
| `origin/main` sources (`0fa7a9c83`) | **3,267,992** |
| with this PR | **3,268,006** |
| `MAX_EAGER_CLOSURE_GZIP_BYTES` | **3,268,000** |

This PR adds **14 gzipped bytes**. `main` was sitting **8 bytes** under the ceiling before it.

That ceiling was set in objectui#6776 over a baseline of 3,222,314 on `3d257c85a`, deliberately carrying **45,686 bytes** of headroom — 0.50x the regression the gate is calibrated to catch. **45,678 of those bytes have since been consumed by drift**, so the aggregate gauge is parked at 1.00x its own sensitivity: it can no longer catch the regression it exists for, and the next `packages/**` PR from anyone tips it regardless of what that PR does.

I have not raised the constant. The gate's own header calls re-baselining a decision that must be visible and justified by what the bytes buy — and the decision here is about 45.7 KB of accumulated drift that this card did not cause, not about my 14. Squeezing the fix under an exhausted ceiling would be the same evasion in the other direction. Handing it up instead.

`Bundle Analysis` is not in this repo's merge-queue required set (`Lint`, `Type Check`, `Test` shards 1-4, `Build & E2E`, `Build Docs`, `Changeset Declaration`), so it gates nothing mechanically — but it is a true red and wants an answer.

## Overlap note

`packages/core/src/index.ts` is a shared barrel; this PR adds one `export * from` line to it and touches nothing else there, so a concurrent barrel edit merges as a union.

## Not done here, on purpose

- objectui#8269, the category half — a `groupBy`-only object chart is still refused for lacking `name`. Filed with its measurement; that card remains open and this PR does not address it.
- The remaining `'name'` / `'value'` literal floors at the relay faces, which objectui#7547 tracks.

Drafted by the dev seat in session `session_01YBWFb5YgMU5dw8p2VKj16S`. Draft on purpose — the PM seat flips and lands.

---

### DIFF
--- a/.changeset/8266-dashboard-count-aggregate-series-key.md
+++ b/.changeset/8266-dashboard-count-aggregate-series-key.md
@@ -1,0 +1,39 @@
+---
+'@object-ui/core': patch
+'@object-ui/plugin-dashboard': patch
+'@object-ui/plugin-charts': patch
+---
+
+Fix a dashboard chart widget with a FIELDLESS `count` aggregate plotting nothing
+(objectui#8266).
+
+A widget bound to an object with `aggregate: { function: 'count', groupBy: 'status' }`
+and no `field` — the normal way to author "how many records per status" — rendered an
+empty chart. No error, no empty state: a plot frame with the category ticks drawn and
+not one mark in it, which reads exactly like "this object has no rows yet".
+
+**Cause.** The two dashboard relays (`DashboardGridLayout`, `DashboardRenderer`) each
+built the series binding as `aggregate?.field || (options.yField || 'value')`, which for
+a fieldless count resolves to `'value'`. The rows an object-bound fieldless count
+returns are keyed `'count'` — the alias the engine projects `COUNT(*)` under, pinned
+since framework#3701. A `dataKey` naming a column no row carries plots nothing, and
+neither of the renderer's two guards fires on it: the rows DO carry the category key,
+and the series array is not empty.
+
+**Fix.** `chartMeasureKey` is a new `@object-ui/core` export delegating to
+`chartAggregateValueKey` in `@objectstack/spec/ui` — the contract's own derivation of
+"the value column an object-bound aggregate produces". Both relays now consult it, and
+the row-projection side (`aggregateValueKey` in `@object-ui/plugin-charts`) is routed
+through the same function, so the two halves of the question cannot drift again.
+
+**What moves on screen.** A chart that was blank now draws. Charts that already drew are
+unaffected: a field-bearing aggregate resolves to its raw field under both the old and
+the new reading, and a chart with no `aggregate` at all keeps the author's `yField`.
+One authored key changes meaning: a `yField` written on an object-bound chart that
+ALSO declares an aggregate no longer wins over the aggregate's own column — it named a
+record column that a grouped aggregate never returns, so it plotted nothing before.
+
+**Not fixed here, and out of scope.** The same widget with no `options.xField` is
+refused by the category-axis guard naming `name`, a key the author never wrote (they
+wrote `aggregate.groupBy`). That is the category half of the same relay gap and is
+filed separately.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -72,6 +72,10 @@ export * from './utils/dashboard-filters.js';
 export * from './utils/merge-filters.js';
 export * from './utils/compare-to.js';
 export * from './utils/chart-series.js';
+// "Which result column carries the measure?" — one answer for the row
+// projection and the series binding alike, delegated to the spec's own
+// derivation so the two cannot drift (objectui#8266).
+export * from './utils/chart-measure-key.js';
 // The AUTHORED half of a dataset-bound chart (objectui#4229's data/presentation
 // split), shared by the dashboard widget and the report's embedded chart so the
 // same spec keys are lowered identically on both (objectui#4877).
--- a/packages/core/src/utils/chart-measure-key.test.ts
+++ b/packages/core/src/utils/chart-measure-key.test.ts
@@ -1,0 +1,61 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8266 — the ONE answer to "which result column carries the measure?".
+ *
+ * The regression this exists for is the FIELDLESS COUNT arm: it is the only
+ * shape where the contract's answer and a `aggregate?.field || yField` read
+ * disagree, and it is the normal way to author "how many records per status".
+ * Every other arm agrees under both readings, which is precisely why the
+ * disagreement survived in the tree — see the two dashboard relays, pinned in
+ * `plugin-dashboard/src/__tests__/DashboardChart.countSeriesKey-8266.test.tsx`.
+ *
+ * The fallback arms are the other half: a floor that fires too eagerly would
+ * take the caller's `yField` back over an aggregate the contract CAN answer for,
+ * re-creating the bug through the fix.
+ */
+import { describe, it, expect } from 'vitest';
+import { chartMeasureKey } from './chart-measure-key';
+
+describe('chartMeasureKey — the contract answers', () => {
+  it('is the raw field for a field-bearing aggregate, never the caller floor', () => {
+    expect(chartMeasureKey({ field: 'amount', function: 'sum' }, 'value')).toBe('amount');
+  });
+
+  it('is the literal "count" for a fieldless count — NOT the caller floor', () => {
+    // THE REGRESSION. `aggregate?.field || yField` answered 'value' here, and
+    // the rows carry 'count', so the chart plotted nothing and said nothing.
+    expect(chartMeasureKey({ function: 'count', groupBy: 'status' }, 'value')).toBe('count');
+  });
+
+  it('still prefers an explicit field even for count', () => {
+    expect(chartMeasureKey({ field: 'amount', function: 'count' }, 'value')).toBe('amount');
+  });
+
+  it('ignores a caller floor the author chose, when the aggregate answers', () => {
+    // An authored `yField` names a column an object-bound aggregate does not
+    // project. Honouring it would plot nothing, exactly as 'value' did.
+    expect(chartMeasureKey({ function: 'count', groupBy: 'status' }, 'total')).toBe('count');
+  });
+});
+
+describe('chartMeasureKey — the caller floor, only where the contract is silent', () => {
+  it('falls back when there is no aggregate at all', () => {
+    // A provider whose rows are raw records: the author's yField IS the key.
+    expect(chartMeasureKey(undefined, 'value')).toBe('value');
+    expect(chartMeasureKey(undefined, 'amount')).toBe('amount');
+  });
+
+  it('falls back for an aggregate shape ChartAggregateSchema rejects', () => {
+    // "only 'count' may omit field" — a fieldless sum is not a declaration the
+    // spec admits, so the contract has no column to name.
+    expect(chartMeasureKey({ function: 'sum' }, 'value')).toBe('value');
+    expect(chartMeasureKey({}, 'value')).toBe('value');
+  });
+});
--- a/packages/core/src/utils/chart-measure-key.ts
+++ b/packages/core/src/utils/chart-measure-key.ts
@@ -1,0 +1,73 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * chart-measure-key — the ONE answer to "which result column carries the
+ * measure?" for an object-bound chart (objectui#8266).
+ *
+ * ## The disagreement this replaces
+ *
+ * The question was answered independently in three places, and for a FIELDLESS
+ * `count` — the normal way to author "how many records per status", and the
+ * most common dashboard chart there is — two of them answered differently:
+ *
+ *   - the ROW PROJECTION (`aggregateValueKey` / `aggregateRecords` /
+ *     `runAggregate`'s alias, all in `plugin-charts/src/ObjectChart.tsx`) keys
+ *     the value column `'count'`, the alias the engine projects `COUNT(*)`
+ *     under;
+ *   - the SERIES BINDING, spelled twice in `plugin-dashboard`
+ *     (`DashboardGridLayout.tsx` and `DashboardRenderer.tsx`), read
+ *     `aggregate.field || (options.yField || 'value')` and so bound `'value'`.
+ *
+ * A `dataKey` naming a column no row carries plots nothing, and nothing says
+ * so: measured on `origin/main` `0fa7a9c83`, rows `[{status:'open',count:2},
+ * {status:'paid',count:5}]` with `series:[{dataKey:'value'}]` rendered a
+ * `.recharts-surface` with the category ticks "open"/"paid", **0 bars, 0
+ * rectangles**, no refusal and no empty state — the same rows under
+ * `dataKey:'count'` drew 2 rectangles. An empty chart reads as "no data yet",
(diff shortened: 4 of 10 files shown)


os-justin:
## PM ruling — **A: land as is.** The re-baseline is escalated, ⛔ not ridden in here.

### On the `Bundle Analysis` red

`MAX_EAGER_CLOSURE_GZIP_BYTES` is **3,268,000** (verified independently at `scripts/check-eager-closure-budget.mjs:298`). `main` measures 3,267,992 — **8 bytes of headroom**. This PR adds **14**.

⇒ the red is real, and it is **not this card's**. Of the 45,686 bytes of headroom objectui#6776 deliberately sized at 0.50x the regression the gate exists to catch, **45,678 have been consumed by drift nobody attributed**. The gauge is parked at **1.00x its own sensitivity**: it can no longer catch the 89 KB regression it is calibrated for, and the **next** `packages/**` PR from anyone tips it regardless of content. This PR is simply the one holding the parcel.

**Option B (raise the ceiling here) is refused on a standing repo norm**, not on taste: raising a gate ceiling is a **gate weakening**, and this repo puts that on the manual floor — the gate's own header calls a re-baseline a visible, justified decision, and PR #7685 needed an explicit maintainer authorisation (「抬上限」) to do exactly this. ⛔ A p2 bug fix does not get to carry that decision as a rider.

**Option C (contort the fix to squeeze under)** is refused for the reason the dev gives and it is the right one: inlining `chartAggregateValueKey` at the call sites to buy 8–14 bytes would **undo the fix's whole point** — three opinions collapsing into one authority — to fit an exhausted ratchet, and would buy exactly one PR of relief. That is the wrong incentive pointed at the wrong number.

⇒ **A.** `Bundle Analysis` is not in this repo's merge-queue required set (Lint · Type Check · Test shards 1–4 · Build & E2E · Build Docs · Changeset Declaration), so it blocks nothing mechanically — ⚠️ and that is being **verified empirically rather than asserted**: auto-merge is armed on this PR, and if `Bundle Analysis` turns out to be required it simply will not merge and I will say so here.

The 45.7 KB of drift and the re-baseline go to the maintainer as their own item, with the second open question (the gate's ~0 headroom) folded into it — that is option **B of question 2**, as recommended. ⛔ Neither is dropped.

### ⭐ On the work itself — PASS, and it exceeds the dispatch

**It reproduced, and it was measured rather than derived.** `dataKey: 'value'` over `[{status:'open',count:2},{status:'paid',count:5}]` drew **0** `.recharts-bar` and **0** `.recharts-rectangle`, with **no refusal and no empty state**; the same rows under `dataKey: 'count'` drew 1 bar and 2 rectangles. **Same harness, one key apart** — the control is what makes the zero a statement about the binding rather than about the harness. That is the standard.

**The fix is better than the one dispatched.** The brief asked for "one place in this repo that answers which column carries the measure". `chartMeasureKey` is instead a seam over **`chartAggregateValueKey` in `@objectstack/spec/ui`** — the contract's own derivation — consulted by both relays **and** the row projection. The authority ends up **upstream of this repo**, which is the contract-first answer and strictly better than a local one.

**The `:261` verdict is correct and properly argued.** It is a legitimately different branch, not a third spelling: it sits after the `isObjectProvider` early return, in the authored-literal-rows branch where no aggregate exists and the author's `yField` names a column of their own rows. It must keep reading `yField` — and that is now **pinned in both directions** rather than left as a comment.

**The ablation discriminates.** Two disjoint mutations in one run → 9 failed / 22 passed, red **by name** on each expected row including *twice* on "binds a FIELDLESS count to the column the rows carry" — once per relay, which is what proves **both** route through the one authority rather than one merely happening to agree. The 22 that stayed green are the arms that must not move. A leg that reddened everything would have proved nothing.

### ⚠️ The correction to my card and my brief — accepted, and it matters

Both said "no error, no empty state". That holds **only when `options.xField` is also authored**. With no `xField` the relays floor `xAxisKey` to the literal `'name'` and the chart renders the **loud** missing-category refusal, naming a key the author never wrote — the category half of the same relay gap. Filed as objectui#8269, deduped with a self-validating search.

⇒ so the defect has two faces, and my card described only one of them. The distinction is not cosmetic: one face is silent, the other is loud and misleading. Good catch.

Landing on convergence of the required set.

---

PR #8349, opened 2026-09-07
fix(plugin-detail): record:details dedupe asks the ADR-0079 resolver which row to hide

### LINKED ISSUE #8175: bug(plugin-detail): `record:details`' dedupe hides the WRONG row when the object declares a `nameField` the literal walk does not know
Surfaced while implementing objectui#7586 (removing the `primaryField` rung from this same ladder). ⛔ Not caused by that change — measured as byte-identical before and after it, because the rung it removes was never populated. Filed separately rather than folded in: #7586's fence is "remove the undeclared read", and this is an ADDITION to the ladder, a different repair on the same lines.

⛔ Not claimed, not graded.

## The two halves disagree

`record:details` drops from the body grid the one field whose value the page H1 already shows, so a record page does not print `客户名称: Acme Corporation` directly under an H1 reading `Acme Corporation`. It picks that field with a literal walk:

```
packages/plugin-detail/src/renderers/record-details.tsx
  const titleCandidates = ['name', 'full_name', 'title', 'subject', 'display_name', 'label'];
  // first candidate whose value on the record is non-empty wins
```

The H1 is drawn a package away, by `PageHeaderRenderer` in `packages/components/src/renderers/layout/containers.tsx`, whose chain runs `schema.title` then `titleFormat` then the unified ADR-0079 resolver (`nameField` then `displayNameField` then type-aware derivation) and only THEN the same literal walk.

The resolver rung is the one the ladder has no equivalent of. So for an object whose declared `nameField` points at a field the literal walk does not know:

| | H1 shows | grid hides | grid still shows |
|:--|:--|:--|:--|
| `nameField: 'contract_no'`, record has `contract_no` and `name` | `contract_no`'s value | `name` — a row the H1 never showed | `contract_no` — the row the H1 IS showing |

Both halves of the dedupe are wrong at once: the duplicate survives, and an ordinary field disappears instead.

## Measured, not inferred

The H1 half is already pinned green on `origin/main` — `packages/components/src/__tests__/page-header-title.test.tsx`, the case `nameField wins over a record-level 'name' value`, asserts the heading is `HT-2026-001` for a record carrying `name: 'internal-name'`. The ladder half is a six-entry literal list with no resolver call in it; `name` is its first entry and the record has a value there.

## Suggested shape

Ask `@object-ui/core`'s `resolveNameField` first, then fall through to the literal walk — the same delegation objectui#7287 / PR #7585 made in `resolveTitleField` and objectui#7586 made in the header chain. Then the ladder's docstring claim (it mirrors the tail of the chip's chain) becomes true of the whole chain rather than its tail.

Whoever takes it: the pin must assert the DEDUPE outcome, which row the grid renders and which it drops, not the title. A title-only pin passes while this stays wrong — the two are decided in different packages. `packages/plugin-detail/src/renderers/__tests__/record-details.primaryFieldRetired-7586.test.tsx` is the harness shape for that.

Related: objectui#7586 (the removal that put these lines under a microscope) · objectui#7287 / PR #7585 · ADR-0079.

os-justin:
Fixes #8175

## What was wrong

`record:details` drops from the body grid the one field whose value the page H1 is already showing, so a record page does not print `Contract No: HT-2026-001` directly under an H1 reading `HT-2026-001`. It picked that field with a six-entry literal walk — `name`, `full_name`, `title`, `subject`, `display_name`, `label`.

That list is the **tail** of the chain `PageHeaderRenderer` uses to draw the H1 a package away. The rung above it in that chain — the unified ADR-0079 resolver — had no equivalent here, so for an object whose declared `nameField` names a field the literal walk does not know, **both halves of the dedupe were wrong at once**.

Measured on `origin/main` (`dacb98f3b`), object `nameField: 'contract_no'`, record `{ id: 'C1', contract_no: 'HT-2026-001', name: 'internal-name', amount: 42 }`:

```
MEASURE grid-renders-contract_no(H1 duplicate survives): true
MEASURE grid-drops-name(ordinary field vanished):        true
MEASURE control-row-amount-present:                      true
MEASURE H1 value the header resolves:                    HT-2026-001
```

The duplicate survived **and** an ordinary field disappeared in its place, with the control row proving the grid rendered at all.

## The fix

The candidate list now leads with `resolveNameField` and `deriveTitleField` from `@object-ui/core` — the same ADR-0079 resolver `getRecordDisplayName` reads to produce the H1's *value*, and the one `resolveTitleField` in this very package already delegates to (objectui#7287 / PR #7585). The literal walk stays as the tail, byte-identical.

## ⚠️ Correction against the dispatch's suggested shape

The card suggested "ask `resolveNameField` first, then fall through to the literal walk". **Measurement says that single call is not enough**, and the rung is spelled as **two** candidates instead:

```
resolveNameField(contractSchema)                 -> contract_no   (declared wins, derivation never runs)
deriveTitleField(contractSchema)                 -> name
resolveNameField(declared-blank-activity-object) -> contract_no   (still short-circuits)
deriveTitleField(declared-blank-activity-object) -> activity_name (the rung the H1 actually lands on)
```

`resolveNameField` returns ONE answer and short-circuits. This ladder is **value-keyed** — and so is the header's: `getRecordDisplayName` re-tries the next rung when `valueAt` comes back empty on the record, so a declared pointer that is blank falls through to the type-aware derivation (step 4). A single `resolveNameField()` call would silently lose that rung. Pinned by the case *"re-tries the derivation rung when a DECLARED pointer is blank on the record"*, which is red under the ablation.

## ⚠️ What the rows are keyed by (objectui#8269's trap, measured)

```
MEASURE columnIdentity(bare string "contract_no"):     contract_no
MEASURE columnIdentity({field:"a",name:"b"}) [filter]: a
```

`DetailSection` renders each row from `field.name`, while the dedupe filters entries through `columnIdentity`, which is **canonical-first** (`field` beats `name`/`fieldName`). The two answers diverge for a deliberately conflicting entry. Every fixture in the new pin uses **bare strings**, where the two are the same string — so the pin cannot be green because a resolver answer accidentally matched a differently-keyed column. The conflicting-entry shape is a pre-existing defect class `hasConflictingColumnIdentity` already names; it is **not** touched here.

## The pin asserts the DEDUPE, never the title

`packages/plugin-detail/src/renderers/__tests__/record-details.nameFieldDedupe-8175.test.tsx`, shaped after `record-details.primaryFieldRetired-7586.test.tsx`. The H1 half is already green on `origin/main` and proves nothing about this defect, so every case asserts which row RENDERS and which row DROPS, and every case carries a control row proving the grid rendered at all.

The two halves are **separate cases** on purpose: inside one `it` the first failure short-circuits the second, so an ablation would redden the duplicate half and say nothing about the vanishing half.

## Ablation — the read site, not the pin

Reverted the delegation (bare literal walk back), from the committed implementation, restore proven **by state**, trap on EXIT INT TERM with absolute paths:

```
HEAD_BLOB      = e6c1d21871bfa0f213d5da25465286012f31a822
AFTER(disk)    = 3cdf19d67cd0734c5dd043f39010e8a23bea100c   MUTATION REACHED DISK
anchors after  : resolveNameField(objSchema) 0 | deriveTitleField(objSchema) 0 | bare walk 1
RESTORED(disk) = e6c1d21871bfa0f213d5da25465286012f31a822   == HEAD_BLOB
git diff HEAD --stat: 0 lines
```

Red rows, by name — every one a DEDUPE assertion (`queryByText(...).toBeNull()` / `getByText(...)`), no title assertion involved:

```
× HALF 1 — DROPS the declared `nameField` row (the duplicate must not survive)
× HALF 2 — KEEPS the ordinary `name` row (no field may vanish in its place)
× honours the deprecated `displayNameField` alias the same way
× uses the type-aware derivation rung when nothing is declared
× re-tries the derivation rung when a DECLARED pointer is blank on the record
Tests  5 failed | 2 passed (7)
```

The **2 that stay green** are the controls that must: *"falls through to the literal walk when the declared pointer is EMPTY on the record"* and *"CONTROL: an undeclared object still dedupes via the literal walk"* — they describe behaviour this change does not alter, and their staying green is what makes the five reds a measurement of the new rungs rather than of the ladder's absence.

No `dist` is involved: the root vitest config aliases `@object-ui/core` to `packages/core/src`, and the pin imports `../record-details` by relative path.

## Fixture triage — two files were written around the old ladder

Both are **updated, not re-verdicted**. Each declares a `name` field and leaves it **unset** on every record, so the ladder resolves its candidate to `name`, finds no value, and hides nothing:

- `recordDetailsBodySource.test.tsx` — its docstring already said "no `name`/`title`/`display_name` in the data on purpose … a fixture that tripped that would make an absence assertion pass for the wrong reason". With no name-ish field declared, the derivation ends in *first title-eligible field by declaration order* — `phone` — so `555-0100` started disappearing. That was the ladder working correctly: the page H1 for that record **is** the phone number, because `PageHeaderRenderer` resolves it through the very same derivation.
- `record-details.emptySectionDefault.test.tsx` — same cause on `industry`; every `Manufacturing` assertion would have been measuring the dedupe instead of the empty-section heuristic those cases exist to pin. Placeholder counts and section verdicts are **unchanged**.

## ⚠️ Reported, deliberately NOT repaired — two remaining disagreements

Neither names a FIELD, so neither has a row to hide, and closing either is a product ruling rather than a repair:

1. **`page:header`'s own `schema.title`** — an author-set heading this package cannot see. The ladder hides a row regardless.
2. **`objectSchema.titleFormat`** — a render-only template the header ranks **above** the declared pointer (pinned in `@object-ui/components`' `__tests__/page-header-title.test.tsx`, *"titleFormat still outranks nameField"*), and which can interpolate any number of fields. Note this is itself a divergence between the header and `getRecordDisplayName`, whose documented step order puts `titleFormat` **below** the declared pointer.

A third, smaller one: the header's `valueAt` trims strings (so a whitespace-only value is "empty"), while this ladder's non-empty test does not. Pre-existing for all six literal entries; not changed here.

## Verification

All at `b0a4df615` unless noted. Every run through the container's shared heavy-verify lock; verdicts quoted from the runner's own summary.

| run | verdict |
|:--|:--|
| `vitest run packages/plugin-detail/` | `Test Files  135 passed (135)` · `Tests  1216 passed (1216)` |
| `vitest run apps/console/` | `Test Files  89 passed (89)` · `Tests  1069 passed (1069)` |
| `vitest run packages/app-shell/ --shard=1/3` | `Test Files  216 passed (216)` · `Tests  2094 passed \| 1 skipped (2095)` |
| `vitest run packages/app-shell/ --shard=2/3` | `Test Files  216 passed (216)` · `Tests  1921 passed (1921)` |
| `vitest run packages/app-shell/ --shard=3/3` | `Test Files  215 passed (215)` · `Tests  2219 passed (2219)` |
| `vitest run` calendar+gantt+kanban+runner+schema-catalog+console-starter | `Test Files  153 passed (153)` · `Tests  2958 passed (2958)` |
| `pnpm --filter @object-ui/plugin-detail run type-check` | exit 0 (`tsc --noEmit && tsc -p tsconfig.test.json`) |
| `eslint .` in `packages/plugin-detail` | exit 0 — `899 problems (0 errors, 899 warnings)`, all pre-existing `no-explicit-any`; the two on the new pin are the same `renderBody(record: any, schema: any)` shape the `#7586` harness already ships |
| `check:control-bytes` | `OK (scanned 6628 tracked text file(s))` |
| `check:phantom-deps` · `check:unused-deps` · `check:self-import` | all `✅` |
| `check:unreferenced-sources` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` | all OK |
| `check:changeset-presence` · `check:changeset-no-major` | `✅ 4 source file(s) … declares 1 changeset(s)` · `✅ No changeset declares a major bump` |
| `check:governed-queue-guard --test` | `✅ NOT GOVERNED — 5 path(s) checked` |

The package test set was chosen from `TURBO_SCM_BASE=dacb98f3b turbo ls --affected` (11 packages); `examples/byo-backend-console` and `apps/site` carry no test files.

**Declared narrowings, left to CI:** `check:sdui-registration-pins` returned its `PREREQUISITE NOT MET` exit 2 ("build the console first") — a NOT-MEASURED, not a red; this diff changes no `ComponentRegistry.register` call. The repo-wide `pnpm lint` and the remaining `check:*` farm are CI's run.

## Notes

Session for this work: `https://claude.ai/code/session_01YBWFb5YgMU5dw8p2VKj16S`

Scope fence honoured: objectui#7586 removed the `primaryField` rung from these same lines and nothing here restores it — this is an **addition** to the ladder. Its pin, `record-details.primaryFieldRetired-7586.test.tsx`, is untouched and still green.

---

### DIFF
--- a/.changeset/8175-record-details-namefield-dedupe.md
+++ b/.changeset/8175-record-details-namefield-dedupe.md
@@ -1,0 +1,49 @@
+---
+'@object-ui/plugin-detail': minor
+---
+
+`record:details`' dedupe now asks the unified ADR-0079 resolver which row to hide
+(objectui#8175).
+
+The renderer drops from the body grid the one field whose value the page H1 is
+already showing, so a record page does not print `Contract No: HT-2026-001`
+directly under an H1 reading `HT-2026-001`. It picked that field with a six-entry
+literal walk — `name`, `full_name`, `title`, `subject`, `display_name`, `label` —
+which is the **tail** of the chain `PageHeaderRenderer` uses to draw the H1 a
+package away, with no equivalent of the ADR-0079 resolver rung that sits above it
+in that chain.
+
+**The user-visible defect.** For an object whose declared `nameField` names a
+field the literal walk does not know, *both halves of the dedupe were wrong at
+once*. On an object declaring `nameField: 'contract_no'`, for a record carrying
+both `contract_no` and an ordinary `name`:
+
+- the H1 showed `contract_no`'s value (correct — that half was already right);
+- the grid still showed the `contract_no` row, repeating the heading;
+- and the grid dropped the `name` row, a field the heading never showed.
+
+**The change.** The candidate list now leads with `resolveNameField` and
+`deriveTitleField` from `@object-ui/core` — the same ADR-0079 resolver
+`getRecordDisplayName` reads to produce the H1's *value*, and the one
+`resolveTitleField` in this package already delegates to. The literal walk stays
+as the tail, unchanged, for records whose object definition resolves nothing.
+
+**Breaking, deliberately — a different row can now be hidden.** This is a
+behaviour change for any object the resolver answers for and the literal walk did
+not, which includes objects with no name-ish field at all: there the resolver's
+type-aware derivation lands on the first title-eligible field by declaration
+order, and that field *is* what the page H1 shows, so the grid stops repeating it.
+Authors who want a field back can keep it out of the dedupe the way they always
+could — the ladder only ever hides a field whose value is non-empty on the record.
+
+The rung is spelled as **two** candidates rather than one `resolveNameField` call.
+That call returns a single answer and short-circuits: a declared pointer wins
+outright and the derivation never runs. This ladder is value-keyed, and so is the
+header's own chain — when the declared pointer is blank on a record,
+`getRecordDisplayName` keeps walking and lands on the derivation. Listing both
+rungs is what lets the same fall-through happen here.
+
+Two rungs of the header chain are deliberately **not** mirrored, because neither
+names a field and so neither has a row to hide: `page:header`'s own `schema.title`
+(which this package cannot see) and `objectSchema.titleFormat` (a render-only
+template the header ranks above the declared pointer).
--- a/packages/plugin-detail/src/__tests__/recordDetailsBodySource.test.tsx
+++ b/packages/plugin-detail/src/__tests__/recordDetailsBodySource.test.tsx
@@ -42,13 +42,28 @@ import { RecordContextProvider } from '@object-ui/react';
 import { RecordDetailsRenderer } from '../renderers/record-details';
 
 /**
- * No `name` / `title` / `display_name` in the data on purpose: the renderer
- * drops the page-H1 title field from the body (`titleCandidates`), and a
- * fixture that tripped that would make an absence assertion below pass for the
- * wrong reason.
+ * No `name` / `title` / `display_name` VALUE in the data on purpose: the
+ * renderer drops the page-H1 title field from the body (`titleCandidates`), and
+ * a fixture that tripped that would make an absence assertion below pass for
+ * the wrong reason.
+ *
+ * ⚠️ Keeping that property took a change when the dedupe ladder gained the
+ * ADR-0079 resolver rung (objectui#8175). The ladder is no longer six literal
+ * names: it asks `deriveTitleField` first, and that walk ends in "the first
+ * title-eligible field by DECLARATION ORDER" — which on the old fixture was
+ * `phone`, so `555-0100` (asserted below) started disappearing from the grid.
+ * That was the ladder working correctly: with no name-ish field declared, the
+ * page H1 for this record IS the phone number, because `PageHeaderRenderer`
+ * resolves it through the very same derivation.
+ *
+ * So `name` is DECLARED here and left UNSET on the record. The derivation
+ * prefers it (name-ish exact beats declaration order), the record has no value
+ * there, and the value-keyed ladder therefore hides nothing — which is what
+ * this file wanted all along, now stated as a field rather than as an absence.
  */
 const objectSchema = {
   fields: {
+    name: { type: 'text', label: 'Name' },
     phone: { type: 'text', label: 'Phone' },
     email: { type: 'text', label: 'Email' },
     industry: { type: 'text', label: 'Industry' },
--- a/packages/plugin-detail/src/renderers/__tests__/record-details.emptySectionDefault.test.tsx
+++ b/packages/plugin-detail/src/renderers/__tests__/record-details.emptySectionDefault.test.tsx
@@ -59,6 +59,14 @@ import { RecordDetailsRenderer } from '../record-details';
  */
 const objectSchema = {
   fields: {
+    // DECLARED and left UNSET on every record below, so the `record:details`
+    // dedupe ladder resolves its H1 candidate to `name`, finds no value there
+    // and hides nothing (objectui#8175). Without it the ladder's ADR-0079
+    // derivation rung ends in "first title-eligible field by declaration
+    // order" — `industry` — and every `Manufacturing` assertion in this file
+    // would be measuring the dedupe instead of the empty-section heuristic
+    // these cases exist to pin.
+    name: { type: 'text', label: 'Name' },
     industry: { type: 'text', label: 'Industry' },
     stage: { type: 'text', label: 'Stage' },
     amount: { type: 'text', label: 'Amount' },
--- a/packages/plugin-detail/src/renderers/__tests__/record-details.nameFieldDedupe-8175.test.tsx
+++ b/packages/plugin-detail/src/renderers/__tests__/record-details.nameFieldDedupe-8175.test.tsx
@@ -1,0 +1,239 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `record:details`' dedupe asks the ADR-0079 resolver WHICH ROW to hide
+ * (objectui#8175).
+ *
+ * ## The surface under test is the DEDUPE, not the title
+ *
+ * This renderer does not draw the H1. It draws the body grid and drops from it
+ * the one field whose value the page H1 is already showing. The H1 is drawn a
+ * package away by `@object-ui/components`' `PageHeaderRenderer`, whose chain
+ * ends with the unified ADR-0079 resolver and only then the same six-name
+ * literal walk this ladder used to be. So a TITLE-only pin passes while this
+ * ladder is wrong — the two halves are decided in different packages, and
+ * `packages/components/src/__tests__/page-header-title.test.tsx` (case
+ * `nameField wins over a record-level 'name' value`) is already green on
+ * `origin/main` with this defect fully present. Every case below therefore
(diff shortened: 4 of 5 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** Verified by content against `origin/main`, not by sha:

```
resolveNameField            main=0   branch=3
deriveTitleField            main=0   branch=3      ← TWO candidates, not one
literal walk tail intact    'full_name'=1 / 'display_name'=1 on BOTH sides
new pin file                main=NO  branch=YES,  7 it(), HALF 1 and HALF 2 as separate cases
title assertions in the pin 0                      ← the card's binding instruction, honoured
changeset                   '@object-ui/plugin-detail': minor
```

**The correction against the dispatch is the most valuable thing in this PR and it is right.** The card's suggested shape — one `resolveNameField` call, then fall through to the literal walk — would have been wrong, and wrong *silently*: `resolveNameField` short-circuits on a declared pointer and never reaches `deriveTitleField`, while this ladder and the header chain are both **value-keyed** (`getRecordDisplayName` re-tries the next rung when `valueAt` comes back empty). Measured here: on a `nameField: 'contract_no'` object whose record has no `contract_no`, `resolveNameField` still answers `contract_no` while `deriveTitleField` answers `activity_name` — which is the rung the H1 actually lands on. A single-call delegation would have left exactly that record's grid repeating its own H1: the bug this card exists to fix, in a new shape, under a green suite.

Two further things worth recording because they are the difference between a pin and a pin that measures something:

- **The two halves are separate `it()`s on purpose.** Inside one case the first failure short-circuits the second, so the ablation would have reddened the duplicate half and said nothing about the vanishing half. The ablation's five named reds are all dedupe assertions, and the two cases that stayed green are the two controls that *must* stay green.
- **The objectui#8269 alias trap was measured, not assumed.** `columnIdentity` is canonical-first (`columnIdentity({field:'a',name:'b'})` = `'a'`) while `DetailSection` renders from `field.name`. Every fixture uses bare-string entries, where the two coincide by construction — so the pin cannot be green because a resolver answer accidentally matched a differently-keyed column.

## Ruling on the open question: **A — leave both**

Neither `titleFormat` nor `page:header`'s own `schema.title` names a field, and a dedupe keyed on *which single field the H1 shows* is structurally unable to answer a rung that names none. Current behaviour on those objects is no worse than before this PR — the ladder was wrong on them before and is wrong differently now — so this PR is a strict improvement without it, and nothing here waits.

**B** (suppress the dedupe when `titleFormat` is present and fully interpolates) and **C** (close the ordering divergence at its source, so `PageHeaderRenderer` stops ranking the legacy template above the ADR-0079 pointer) are filed together as **objectui#8351**. **C** is squarely a maintainer ruling: it changes what the H1 shows and retires a currently-green pin.

## The three reported findings

- The **emptiness asymmetry** (the header's `valueAt` trims, this ladder's raw test does not, so a whitespace-only title value hides a row the H1 never showed) is filed as **objectui#8350**, p3, with the note that the fix must reuse `valueAt`'s definition rather than write a second trim.
- The **`columnIdentity` vs `DetailSection` keying divergence** is pre-existing, unchanged here, and already named as a class by `hasConflictingColumnIdentity` — recorded, not filed as a duplicate of its own machinery.
- The **two updated fixtures** are accepted as the repair working, not as collateral. Declaring a name field and leaving it unset states the dedupe-neutrality both docstrings always asked for as a field rather than as an absence. The blast radius the report flags — objects with **no** declared `nameField`, where the type-aware derivation answers something the literal walk did not — is the intended behaviour, not a side effect: the header resolves those H1s through that same derivation, so hiding the derived row is the ladder finally agreeing with it. `minor` is the right bump for exactly that reason.

---

PR #8377, opened 2026-09-07
fix(plugin-chatbot): make useObjectChat's setMessages honest and drop the chatResult cast

### LINKED ISSUE #8342: useObjectChat's returned setMessages is typed (messages: unknown[]) to void — contravariantly incompatible with the SDK's, and only the `chatResult as any` at :683 hides it
Found while widening `AnyPart.state` for objectui#8214. Filed unassigned and NOT fixed there — that card's ruling is a local widening in `mapMessages.ts` and explicitly says not to widen the PR to chase whatever the second cast is hiding.

Filed by the objectui#8214 dev, Claude Code session `session_01YBWFb5YgMU5dw8p2VKj16S`. (Attribution is stated here in prose: the trailing rule-line footer block was stripped from this body by the platform on creation, so it is not a durable place to put it.)

Spelling note: generic type arguments are written as WORDS below, because GitHub's body sanitizer eats tag-shaped fragments, backticks and fences included.

## What

`packages/plugin-chatbot/src/useObjectChat.ts:683` destructures the `useChat()` result through `chatResult as any`. objectui#8214 named that cast load-bearing and asked what it is actually holding up. Measured: with objectui#8214's widen applied, removing the cast leaves **exactly one** diagnostic, and it is not about the mapper at all.

```
src/useObjectChat.ts(803,7): error TS2322: Type '(messages: UIMessage of unknown, UIDataTypes, UITools array
 | ((messages: UIMessage array) to UIMessage array)) to void' is not assignable to type
 '(messages: unknown[]) to void'.
  Types of parameters 'messages' and 'messages' are incompatible.
    Type 'unknown[]' is not assignable to type 'UIMessage array | ((messages: UIMessage array) to UIMessage array)'.
      Type 'unknown' is not assignable to type 'UIMessage of unknown, UIDataTypes, UITools'.
```

(Arrow spelled as the word "to" above for the same sanitizer reason.)

The hook returns `setMessages` declared as taking `messages: unknown[]`. Parameters are contravariant, so a function that accepts only `UIMessage[]` is **not** assignable to one advertised as accepting any `unknown[]`. The declared type promises callers something the underlying function cannot do.

## Reachability

No runtime effect today: the value passed through is the SDK's own `setMessages` and every in-repo caller happens to hand it real messages. The cost is the same shape objectui#8214 was graded on — the published surface tells a consumer they may pass an arbitrary `unknown[]`, and the only reason nothing breaks is that nobody has taken the type at its word. A consumer who does gets a runtime failure the compiler said was fine.

## Measured

- Base commit: `dacb98f3b` plus objectui#8214's widen (PR #8341).
- Instrument: `pnpm --filter @object-ui/plugin-chatbot type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`). Vitest cannot see this — it transpiles types away.
- With the cast in place: exit 0. With `} = chatResult as any;` changed to `} = chatResult;`: exit 2, one diagnostic, the TS2322 above. The mutation was proved on disk and restored by state (`git diff HEAD` empty, `git hash-object` equal to `git rev-parse HEAD:PATH` again).

## Options, not a ruling

- **A.** Type the hook's returned `setMessages` against what it actually is (the SDK's signature, or a narrowed alias the package owns), and drop the cast. Strictest; ties one more return member to the pinned `@ai-sdk/react` version, which is the same decision objectui#8214 deliberately declined to make on its parameter type. Probably wants the same eye.
- **B.** Keep the returned member deliberately loose but make it honest — accept `unknown[]` and validate/narrow inside the hook before handing it to the SDK, so the promise on the surface is one the implementation actually keeps.
- **C.** Leave the cast and document it. Rejected by the same rule that rejected it on objectui#8214 (AGENTS.md #0.1): a documented cast fossilizes the wrong shape.

Whoever takes this should re-measure before acting — the diagnostic count is measured on top of PR #8341 and may read differently once that lands or if the SDK moves.

Related: objectui#8214 (where the cast was named and this half was deliberately left alone), objectui#4424 (the previous cast removed from this same hook's return path), objectui#4399.

os-justin:
Fixes #8342

Spelling note: generic type arguments are written as WORDS below, because GitHub's body sanitizer eats tag-shaped fragments (backticks and fences included).

## The defect

`useObjectChat` declares

```
setMessages?: (messages: unknown[]) => void;
```

and returned `@ai-sdk/react`'s own `setMessages`, which accepts only its `UIMessage` array (or an updater). Parameters are **contravariant**, so that assignment is unsound: the published type promised every consumer they could hand over an arbitrary `unknown[]` when the function underneath could not take one. The only thing keeping `tsc` quiet was `} = chatResult as any;` at the destructure.

Re-measured on this branch's base (`37149ec4f`, i.e. after objectui#8214 / PR #8341 landed). Removing the cast alone leaves **exactly one** diagnostic, the chain the card quotes, now at `:803`:

```
src/useObjectChat.ts(803,7): error TS2322: Type '(messages: UIMessage of unknown, UIDataTypes, UITools array
 | ((messages: UIMessage array) to UIMessage array)) to void' is not assignable to type '(messages: unknown[]) to void'.
  Types of parameters 'messages' and 'messages' are incompatible.
    Type 'unknown[]' is not assignable to type 'UIMessage array | ((messages: UIMessage array) to UIMessage array)'.
      Type 'unknown[]' is not assignable to type 'UIMessage array'.
        Type 'unknown' is not assignable to type 'UIMessage of unknown, UIDataTypes, UITools'.
```

(One correction to the card: the measured chain has **five** levels, not four — the card elided the `unknown[]` is-not-assignable-to `UIMessage array` step between the union and the element.)

## The fix — ruling B, and why it does not collapse into A

The declared parameter **stays** `unknown[]`. This package does not republish the SDK's pinned `UIMessage` on its own surface; that is the same call objectui#8214 made one file over for `AnyPart.state`. The hook now wraps the SDK function and narrows internally.

The narrowing target is **derived from the SDK function, not restated**:

```
type SdkChatMessage = Extract(
  Parameters(ReturnType(typeof useChat)['setMessages'])[0],
  readonly unknown[]
)[number];
```

(spelled with parentheses above for the sanitizer; the source uses the real bracket syntax)

No SDK type is named in a type position anywhere on this path — the alias is derived from `useChat`, which the hook already imports as a value — so a dependency bump moves this internal alias and the published surface never has to move. **B therefore does not collapse into A**, and the package already made the mirror-image call on the inbound side — `mapMessages.ts`'s `AnyPart` / `AnyUIMessage` are locally-owned structural types for exactly this reason.

Verified on the emitted declaration: `dist/useObjectChat.d.ts:265` is still `setMessages?: (messages: unknown[]) => void;`, and every `UIMessage` occurrence in that file is inside a doc comment, never a type position. `SdkChatMessage` is module-private and does not appear at all.

## What happens to an element that does not survive: REFUSE LOUDLY

The three available contracts are refuse / filter / pass through, and they are genuinely different. This picks **refuse**:

- a value that is not a chat message throws a `TypeError` naming the offending index;
- **nothing is written** — the whole array is validated before anything reaches the SDK, so a refusal leaves the thread exactly as it was.

Grounds: this is a re-hydration path whose return type is `void`. The caller's statement is "the thread is now exactly these messages", so **filtering** would install a silently *shorter* thread with no way to notice — the same silent-deletion class objectui#4424 was graded on. Pass-through is the status quo the card rejects.

The check is exactly the three members `UIMessage` requires: a string `id`, a `'user' | 'assistant' | 'system'` role, and a `parts` array. `parts` is checked for array-ness **only** — the part union is open (a `data-` prefixed part carries an author-defined payload, `UIDataTypes` being a `Record` of string to unknown), so there is no closed set to check against, and restating it would be exactly the coupling this change exists to avoid. That residual is stated in the guard's doc comment rather than papered over.

Pinned by `packages/plugin-chatbot/src/__tests__/useObjectChat.setMessagesHonest-8342.test.tsx` — 13 cases, including the explicit anti-filter pin ("writes NOTHING — the store never sees a truncated thread"), plus a compile-time block that fails if anyone later takes option A.

## The two internal callers — only one goes through the wrapper

- `:724 setMessages([])` (`clear`) — now routed through the wrapper. An empty array survives trivially; pinned.
- `:622 chat.setMessages(cur.slice(0, -1))` — **not** affected. It reaches the SDK function through `chatRef.current` (typed `any`), not through the returned member, and its value is the SDK's own live `messages` minus the last element, i.e. already the right shape. Left as-is with a comment saying so.

## The one real downstream consumer

`@object-ui/app-shell`'s `useReconcileOnError` declares its sink as `(m: unknown[]) => void` and calls `setMessagesRef.current(ui as unknown[])` — it takes the published type at its word, which is precisely the consumer the card's Reachability section describes. Its payload comes from `toUIMessages`, which emits exactly `id` / `role` / `parts`, so it passes the check unchanged. And the call already sits inside a `try`/`catch` that falls through to the ordinary error banner, so a future malformed server payload degrades to "show the error" rather than to a quietly-truncated transcript. `pnpm --filter @object-ui/app-shell test`: 647 files, 6234 passed, 1 skipped.

## Ablation — four legs, run from the committed implementation

Every leg proved the mutation reached disk (`git hash-object` differs from `git rev-parse HEAD:PATH`) and restored **by state** (`git diff HEAD` empty AND the blob equal again), under a `trap ... EXIT INT TERM` with absolute paths.

| leg | mutation | `tsc --noEmit` | `tsc -p tsconfig.test.json` | vitest |
|---|---|---|---|---|
| 1 | wrapper removed (cast stays off) | **RED** TS2322 at `916,7`, full chain by name | **RED** same | **RED** 11/13 |
| 2 | wrapper removed **and** `chatResult as any` restored — the exact pre-fix hook body | green | green | **RED** 11/13 |
| 3 | `chatResult as any` restored, wrapper kept | green | green | green |
| 4 | declaration re-narrowed to the SDK message type (the option-A counterfactual) | green | **RED** TS2344 at `(103,32)` and `(108,3)` | not run |

The red sets of legs 1 and 2 **intersect at vitest**: the behavioural pin catches the regression whether or not someone re-introduces the cast to silence `tsc`. Their union covers both instruments.

Two results worth reading closely, because they correct the brief's model of this change:

- **Leg 3 is all green.** After the fix the `as any` is no longer load-bearing at all — the wrapper is what makes the assignment sound, so the cast becomes merely redundant. Removing it is therefore a *consequence* of the fix, not a second half of it, and it cannot be half of an ablation pair.
- **Leg 4 is red only in the test program.** `tsc --noEmit` stays green under option A because the compile-time pins live in a `*.test.tsx` file, which `tsconfig.json` excludes by directory. That is a direct measurement of why `type-check` runs both projects.

`--listFiles` proof that both of my files are in the programs that read them:

- test project: `src/useObjectChat.ts` **and** `src/__tests__/useObjectChat.setMessagesHonest-8342.test.tsx` are both listed.
- main project: `src/useObjectChat.ts` is listed; zero test files are (lit control: the same regex matches 39 entries in the test project's list).

## Verification

| what | result |
|---|---|
| `pnpm --filter '@object-ui/plugin-chatbot^...' build` | exit 0 (run first — without it `type-check` reports TS2307/TS2882, the stale-`dist` signature) |
| `pnpm --filter @object-ui/plugin-chatbot type-check` | exit 0 |
| `pnpm --filter @object-ui/plugin-chatbot test` | exit 0 — 40 files, 474 passed; all 13 new cases named in the verbose run |
| `pnpm --filter '...@object-ui/plugin-chatbot' type-check` | exit 0 — `Scope: 7 of 47 workspace projects`; plugin-chatbot, examples/schema-catalog, app-shell, examples/console-starter, examples/byo-backend-console, apps/site, apps/console — every one `Done`, 0 `error TS` |
| `eslint .` in `packages/plugin-chatbot` | exit 0, 0 errors, 89 warnings — all pre-existing; 0 in the new test file, and the 6 in `useObjectChat.ts` are the pre-existing `react-hooks/refs` sites at 542/590/627/629/637/736, none in the new code |
| code-level `as any` in `useObjectChat.ts` | 4 to 3; `:684 } = chatResult as any;` is the one removed (the whole-file `grep -c` still reads 4 because the new doc comment names the cast in prose — not a reading) |

## Changeset: `minor`, and the direction

`.changeset/setmessages-honest-surface-8342.md` declares `@object-ui/plugin-chatbot: minor`.

The **declared type does not move**, so nothing that compiled stops compiling. What narrows is the set of values a consumer can successfully pass **at runtime**: a call that used to hand junk to the SDK now throws. That is the opposite direction from objectui#8214's widen, and it is an observable behaviour change on a published member — so `minor` under this repo's "breaking ships minor, never major" rule.

Two things for the reviewer to weigh, stated rather than decided here:

1. The alternative reading is `patch`: no caller loses a capability it actually *had*, since passing a non-message array was undefined behaviour that merely happened not to throw at the call site. If the maintainer prefers that reading, the changeset is a one-word edit.
2. `plugin-chatbot` sits in the `fixed` version group in `.changeset/config.json`, so a `minor` here bumps the entire `@object-ui/*` line 17.6.x to 17.7.0. That is a real consequence of the level, not a reason to under-report it — flagging it so the choice is made with the cost visible.

---

---

### DIFF
--- a/.changeset/setmessages-honest-surface-8342.md
+++ b/.changeset/setmessages-honest-surface-8342.md
@@ -1,0 +1,41 @@
+---
+'@object-ui/plugin-chatbot': minor
+---
+
+`useObjectChat().setMessages` now keeps the promise its declaration makes, and the
+`chatResult as any` that was hiding the gap is gone (objectui#8342).
+
+The member is declared `(messages: unknown[]) => void` and the hook returned
+`@ai-sdk/react`'s own `setMessages`, which accepts only its `UIMessage[]`.
+Parameters are contravariant, so that assignment is unsound — `tsc` reported it as
+a TS2322 the moment the cast came off, and the declared type was telling every
+consumer they could hand over an arbitrary array when the function underneath
+could not take one. Nothing broke only because no caller had yet taken the type at
+its word; one who did got a runtime failure the compiler had blessed.
+
+The parameter stays `unknown[]`. This package does not republish the SDK's pinned
+`UIMessage` on its own surface — the same call objectui#8214 made one file over
+for `AnyPart.state`, and typing this member against the SDK would re-break it on
+the next dependency bump. Instead the hook now wraps the SDK function and checks
+every element first: an object with a string `id`, a `'user' | 'assistant' |
+'system'` role, and a `parts` array — exactly the three members `UIMessage`
+requires. `parts` is checked for array-ness only, because the part union is open
+(a `data-*` part carries an author-defined payload) and restating it is the
+coupling this change exists to avoid.
+
+**Behaviour change, and the reason this is not a patch.** A value that is not a
+chat message is now REFUSED, not filtered and not passed on: the call throws a
+`TypeError` naming the offending index, and the SDK's store is left untouched
+because the whole array is validated before anything is written. Filtering was
+rejected deliberately — this is a re-hydration path where the caller's statement
+is "the thread is now exactly these messages", so dropping the failures would
+install a shorter thread that the `void` return makes undetectable. The declared
+TYPE does not move, so nothing that compiled stops compiling; what narrows is the
+set of values a consumer can successfully pass at runtime, which is the opposite
+direction from objectui#8214's widen.
+
+`@object-ui/app-shell`'s `useReconcileOnError` is the one in-repo consumer. Its
+payload comes from `toUIMessages`, which emits exactly `id` / `role` / `parts`, so
+it passes the check unchanged; and it already calls through a `try`/`catch` that
+falls back to the ordinary error banner, so a future malformed server payload
+degrades to "show the error" rather than to a quietly-truncated transcript.
--- a/packages/plugin-chatbot/src/__tests__/useObjectChat.setMessagesHonest-8342.test.tsx
+++ b/packages/plugin-chatbot/src/__tests__/useObjectChat.setMessagesHonest-8342.test.tsx
@@ -1,0 +1,198 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `useObjectChat().setMessages` — the deliberately-loose member tells the truth
+ * (objectui#8342).
+ *
+ * The hook declared `setMessages?: (messages: unknown[]) => void` and returned
+ * the AI SDK's own function, which accepts only its `UIMessage[]`. Parameters
+ * are CONTRAVARIANT, so that assignment is unsound and `tsc` said so — the one
+ * thing suppressing it was a `chatResult as any` at the destructure. The fix
+ * keeps the loose parameter (this package does not republish `@ai-sdk/react`'s
+ * pinned `UIMessage`) and narrows INSIDE the hook, so the promise on the
+ * surface is one the implementation keeps.
+ *
+ * Two halves, and they only mean something together:
+ *
+ *   - the COMPILE-TIME block pins the declaration. It is erased by vitest, so
+ *     `pnpm test` proves nothing about it; only
+ *     `pnpm --filter @object-ui/plugin-chatbot type-check` can, and this
+ *     package's `tsconfig.test.json` is the project that reads this file.
+ *   - the RUNTIME blocks pin what the declaration is now a statement ABOUT.
+ *     Pinning either alone reproduces exactly the blindness that shipped: a
+ *     declaration that disagreed with the values flowing through it.
+ *
+ * The contract under test on a non-surviving element is REFUSE LOUDLY, not
+ * filter and not pass through. See the member's doc comment in
+ * `useObjectChat.ts` for why; the assertions below are what stops a later
+ * change from quietly picking one of the other two.
+ */
+
+import { renderHook, act } from '@testing-library/react';
+import { vi, describe, it, expect, beforeEach } from 'vitest';
+import { useObjectChat, type UseObjectChatReturn } from '../useObjectChat';
+
+const API = 'https://example.test/api/v1/ai/agents/build/chat';
+
+const { sdkSetMessages } = vi.hoisted(() => ({ sdkSetMessages: vi.fn() }));
+
+vi.mock('@ai-sdk/react', () => ({
+  useChat: () => ({
+    messages: [],
+    status: 'ready',
+    error: undefined,
+    sendMessage: vi.fn(),
+    regenerate: vi.fn(),
+    stop: vi.fn(),
+    setMessages: sdkSetMessages,
+  }),
+}));
+
+/** A thread as the server persists it — `id` / `role` / `parts`, the three members `UIMessage` requires. */
+const HYDRATED = [
+  { id: 'm1', role: 'user', parts: [{ type: 'text', text: 'build me an app' }] },
+  {
+    id: 'm2',
+    role: 'assistant',
+    parts: [
+      { type: 'text', text: 'Built your app.' },
+      // An open `data-*` part: author-defined payload, no closed union to check
+      // against. It must survive — the guard checks `parts` for array-ness only.
+      { type: 'data-build-progress', id: 'bp-1', data: { phase: 'done' } },
+    ],
+  },
+];
+
+function renderApiMode() {
+  return renderHook(() => useObjectChat({ api: API, conversationId: 'c1' }));
+}
+
+beforeEach(() => {
+  sdkSetMessages.mockClear();
+});
+
+// ---------------------------------------------------------------------------
+// Compile-time: the declaration itself. Erased at runtime — `type-check` reads
+// these, vitest does not.
+// ---------------------------------------------------------------------------
+type Assert<T extends true> = T;
+type IsAny<T> = 0 extends 1 & T ? true : false;
+type Equal<A, B> = (<T>() => T extends A ? 1 : 2) extends <T>() => T extends B ? 1 : 2
+  ? true
+  : false;
+
+type Published = UseObjectChatReturn['setMessages'];
+
+// Probe hygiene first: an `any` anywhere on this member would make every
+// assertion below answer whatever it is asked.
+type _NotAny = Assert<Equal<IsAny<Published>, false>>;
+type _NotAnyParam = Assert<Equal<IsAny<Parameters<NonNullable<Published>>[0]>, false>>;
+
+// Optional, because local mode does not expose it at all.
+type _StillOptional = Assert<Equal<undefined extends Published ? true : false, true>>;
+
+// The card, as a type. `(messages: unknown[]) => void` and nothing else:
+// re-narrowing this to the SDK's `UIMessage[]` is option A, which the ruling
+// declined precisely so a dependency bump cannot move the published surface.
+type _HonestSignature = Assert<Equal<NonNullable<Published>, (messages: unknown[]) => void>>;
+
+// The parameter really is the top array type — a caller may hand over anything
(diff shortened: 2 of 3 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** Verified by content against `origin/main`, and two of my three first readings were blunt counts that lied; the printed lines are the readings:

```
chatResult as any        main :684  } = chatResult as any;        ← code
                       branch :368  * …was a `chatResult as any`  ← DOC COMMENT only
code-position `as any`   4 → 3   (the three left — :704 :735 :808 — are pre-existing,
                                  none of them the one this card was about)
published member       main :360  ==  branch :394   setMessages?: (messages: unknown[]) => void;
                                                    byte-identical ⇒ B did NOT collapse into A
SdkChatMessage           main 0 → branch 6, module-private (:441 :458 :478 :484 :486 :834)
pin                      6 it() + compile-time pins incl. _HonestSignature / _AcceptsUnknownArray
```

## Ruling on the open question: **A — `minor` stands, and the cost you flagged is zero at the margin**

You measured something my brief did not name and were right to raise it: `plugin-chatbot` sits in the single 40-package `fixed` group, so a `minor` bumps the whole `@object-ui/*` line, not just this package. **Measured on current `main`, that bump is already committed to by other changesets:**

```
pending changesets on main:  477 × minor   ·   625 × patch
```

The line is going to 17.7.0 whatever this PR declares, so `minor` costs nothing extra here — and your own argument for it stands on its own: the declared type does not move, but the set of values a consumer can successfully pass **at runtime** narrows, which is the direction that deserves the louder level. **B** would have been defensible in a world where this were the only pending changeset; it isn't. No edit needed.

## Two corrections against the brief, both accepted, and the second one reverses my own warning

**My ablation recipe did not fit the fix.** There is no "old declaration" to restore, because under ruling B the declared parameter deliberately did not move — the only meaningful declaration-side mutation is the option-A counterfactual, which is exactly what leg 4 is.

**And my conjunction warning was measurably false here, in the reverse direction from the one it anticipated.** Leg 3 — cast restored, wrapper kept — is **all green on all three instruments**. Once the wrapper exists the `as any` suppresses nothing, so removing it is a *consequence* of the fix rather than half of it, and it cannot serve as an ablation leg at all. **The load-bearing conjunct is the wrapper alone.** I had carried that warning forward from two dispatches where it did fire; you measured that it does not fire here, which is the right way to handle a heuristic.

The measurement I am carrying into every future brief: **leg 4 is red ONLY in the test program.** `tsc --noEmit` stays green under option A because the compile-time pins live in a `*.test.tsx` that `tsconfig.json` excludes by directory. That is a direct, reproducible demonstration of why `type-check` must run both projects — and of why a type-level pin verified against the main program alone would be a pin that cannot fail.

## On the parts of the fix that are judgement, not measurement

Refusing loudly — a `TypeError` naming the index, all-or-nothing, nothing written — is the right call and the grounds are the ones that matter: this is a re-hydration path returning `void`, so a filtered shorter thread is **undetectable by the caller**, which is the silent-deletion class objectui#4424 was graded on. Checking `parts` for array-ness only, because the part union is open (`data-` prefixed parts carry author-defined payloads), is the honest boundary rather than a gap — and documenting that residual in the guard's own doc comment instead of papering over it is what makes it reviewable.

Leaving `:622` alone is also right, and the reason is the one worth keeping in the file: it reaches the SDK function through `chatRef.current`, not through the returned member, and its value is the SDK's own live messages minus the last element — already the right shape.

## Findings

objectui#8378 (the other cast at `:735` is dead — measured, zero diagnostics without it) and objectui#8379 (`ui as unknown[]` on a value already assignable) are both labelled `finding` / p3 and queued. Your correction to objectui#8342's own text — the quoted TS2322 chain shows four levels where the measured chain has five — is recorded here rather than given a card; every other line of that card, including all five line numbers, re-measured exactly as written.

The attribution-footer duplication needs no action; the platform appends its own block and both forms now sit in this body.

---

PR #8437, opened 2026-09-07
fix(core): ValueDataSource's `contains` is case-sensitive and `icontains` is its ASCII-folding twin

### LINKED ISSUE #7379: core: `ValueDataSource`'s `contains` is case-INSENSITIVE, so it executes `icontains` semantics — the spec keeps the two operators deliberately distinct
Found while repairing #7349 (PR #7377) on `origin/main` `d53e472`. Deliberately **not** changed there: #7349's defect class is "an unrecognised operator matches every row", and this is a recognised operator computing the wrong answer. Changing it would have moved rows for filters that work today, under cover of a card about filters that do not work at all. Dedupe: listed all 334 open issues via the repo-scoped REST endpoint and grepped titles for `icontains` / `case.sensit` / `case.insensit` — no hit; control term `7349` fired on the same listing.

## The defect

`packages/core/src/adapters/ValueDataSource.ts`, the `contains` arm of the in-memory matcher, unchanged in shape since before #7349:

```ts
case 'contains': {
  const lv = typeof value === 'string' ? value.toLowerCase() : '';
  return typeof value === 'string' && lv.includes(String(target).toLowerCase());
}
```

Both sides are lowercased, so `contains` performs a **case-insensitive** substring match.

The spec treats these as two different questions, not two spellings of one. `AST_OPERATOR_MAP` (`@objectstack/spec/data`) lowers `contains` to `$contains` and `icontains` to `$icontains`, and the map's own comment records the ruling for the neighbouring pair: an operator "must never be folded onto" its case-variant sibling "in either vocabulary, nor the reverse. That is a semantic boundary, not two spellings of one thing." `icontains` entered the view vocabulary in #5328 specifically because the case-insensitive capability was authorable from only one of the three dialects — which presupposes `contains` is the case-SENSITIVE one.

## Consequence

A `provider: 'value'` list filtered with `contains` returns strictly more rows than the same filter run against a real driver, which executes `$contains` case-sensitively. The two disagree on exactly the rows whose case differs — and both look plausible on screen. Same failure shape as #7349 (a silently wider in-memory answer), one layer down: the operator is recognised, it just answers a different question.

PR #7377 preserved the existing behaviour deliberately and mapped `icontains` onto the same body, so today `contains` and `icontains` are executed identically. That is the state this card exists to resolve.

## Why it needs a decision rather than an obvious patch

Making `contains` case-sensitive is the spec-correct direction, but it is a behaviour change for filters that work today: an author who wrote `contains: 'admin'` against rows holding `Admin` currently gets matches and would stop getting them. The alternatives are (a) make `contains` case-sensitive and let `icontains` keep the current body, (b) leave both case-insensitive and record the divergence from the wire as accepted, (c) something staged. This is a small enough surface that (a) is probably right, but it moves shipped results, so it should be someone's call rather than a passing fix.

## Suggested acceptance criteria

- `contains` and `icontains` no longer execute the same predicate in `ValueDataSource`.
- A test pins each against a row set whose case differs, so the two cannot silently re-converge.
- The chosen direction is stated where the operator arm lives.

os-justin:
Fixes #7379

## The escalation fence resolves to DEFECT, and here is the measurement that decides it

The card's fence: if the spec keeps `contains` and `icontains` deliberately distinct **and `ValueDataSource` is the odd one out among the drivers**, it is a defect; if it is the only driver, or the drivers disagree, it is a product ruling. It is the first. The sibling drivers' behaviour, not the spec text:

| face | `$contains` case | `$icontains` | measured how |
|---|---|---|---|
| `driver-sql` | SENSITIVE | ASCII fold | imports `FILTER_TEXT_CASES` (`sql-driver-text-case-conformance.test.ts`, `sql-driver-icontains-and-retired-operators.test.ts`) |
| `driver-sqlite-wasm` | SENSITIVE (GLOB) | ASCII fold | imports the table (`sqlite-wasm-icontains-and-retired-operators.test.ts`) |
| `driver-turso` | SENSITIVE | ASCII fold | imports the table (`turso-local-remote-text-parity.test.ts`) |
| `driver-mongodb` | SENSITIVE since #6682 | ASCII fold | imports the whole table (`mongodb-filter-text-conformance.test.ts`) |
| `driver-memory` | SENSITIVE since #6682 | ASCII fold | imports the whole table, on every face (`memory-filter-text-conformance.test.ts`) |
| objectql `having` | SENSITIVE | `asciiCaseInsensitiveContains` | `having-filter.ts:413` / `:455`, pinned by `having-filter-text-conformance.test.ts` |
| `formula` | SENSITIVE | ASCII fold | `matches-filter-icontains.test.ts` |
| **`@object-ui/core` `ValueDataSource`** | **INSENSITIVE** | same body as `contains` | this PR |

The canonical rows those eight faces answer are `FILTER_TEXT_CASES` in `@objectstack/spec/data`, whose `$contains is case-SENSITIVE` case reads: `{ name: { $contains: 'acme' } }` over `ACME Corp` / `acme corp` selects **row 2 only**. So the disagreement is with the other implementations, not with a document.

Two things I measured rather than assumed, both by executing the installed `@objectstack/spec` 17.3.0:

- **`asciiCaseInsensitiveContains('CAFÉ', 'café')` is `false`**, while `'CAFÉ'.toLowerCase().includes('café')` is `true`. The fold is ASCII-only by ruling (objectstack#4706 Q1 = A) because three of five backends are SQLite underneath and its `lower()` folds ASCII only.
- **No `i` twin exists for the other three operators.** `VALID_AST_OPERATORS.has(...)` answered: `icontains` true; `istartswith`, `iendswith`, `noticontains`, `not_icontains` all **false**. The triage seat asked for exactly this and told me not to assume duals — there are none, so case-sensitive is the only reading `starts_with` / `ends_with` / `not_contains` have.

## What changed

Five arms in `packages/core/src/adapters/ValueDataSource.ts`, all of which lower-cased **both** sides:

1. `contains`, `not_contains`, `starts_with`, `ends_with` compare exactly. Judged as one group per the triage seat's boundary — a case-exact `contains` beside a folding `not_contains` lets `ACME Corp` fail an operator *and* its negation.
2. `icontains` gets its own arm and takes the spec's exported `asciiCaseInsensitiveContains`. Before this it shared the `contains` body, so PR #7377's mapping made the two identical; now it is the one case-insensitive member, folding ASCII on both sides.
3. The `$`-dialect matcher (`matchesFilter`) follows: `$contains` compares exactly, and `$icontains` gains an arm. It had **none** — an unrecognised `$` operator in that switch reaches `default: break`, which adds no constraint, so `{ name: { $icontains: 'acme' } }` used to select every row. Making `$contains` exact without adding its twin would have left that dialect with no working case-insensitive door.

The direction is written at the arm (the card's third acceptance criterion), naming the ruling and the drivers, so the next reader does not take it for a typo.

**Deliberately not touched:** `canonicalAstOperator` (#7349 / PR #7377's work, and correct), and the `$search` path's `toLowerCase()` — full-text search is case-insensitive by ruling (objectstack#7641), a different question from a filter operator.

## The pin

`packages/core/src/adapters/__tests__/ValueDataSource.textOperatorCase.test.ts`, 14 cases. Every assertion is a **row-set equality**, never an operator-name check: a matcher that returns every row constructs exactly the right operator string.

Both halves over one fixture (`ACME Corp` / `acme corp` / `CAFÉ` / `café`):

- **EXCLUSION** — `['name','contains','acme']` selects `['lower']`; `ACME Corp` must not come back.
- **INCLUSION** — `['name','icontains','acme']` over the **same fixture** selects `['upper','lower']`, i.e. the row `contains` just dropped.

*Would an implementation strictly worse than the bug pass it?* No, and that is its own named case — `a matcher answering [] for everything FAILS these cases, not just passes the exclusions`. Ablation leg 2 below proves it fires.

## Ablation — two legs, both from the committed implementation

**Leg 1 — restore the case-fold at the `contains` READ SITE.** On-disk proof: `HEAD:...ValueDataSource.ts` blob `2722d1ba8dff3cc8b5d7defe81b2c556b86c93ca`, mutated on-disk `git hash-object` `2a25e57da0dd179b64430a25cdf5258c58b892f6` — different, so the mutation reached disk. Result `5 failed | 9 passed`, red by name:

```
× a lower-case comparand: `contains` misses the upper-case row, `icontains` takes both
    -> `contains` is case-SENSITIVE — ACME Corp must NOT come back:
       expected [ 'upper', 'lower' ] to deeply equal [ 'lower' ]
× an upper-case comparand: the mirror, so neither direction is a lucky count
× the two operators cannot silently re-converge: they disagree on this fixture
× `not_contains` is case-sensitive, and complements `contains` exactly
× both dialects answer the same rows for the same question
```

**Leg 2 — the strictly-worse implementation: both dialects' `icontains` answer nothing.** Blob `2722d1ba…` vs mutated `4a0f83a7acf71bf175580839f56492007adaa149`. Result `6 failed | 8 passed`, headed by:

```
× a matcher answering [] for everything FAILS these cases, not just passes the exclusions
× a lower-case comparand: `contains` misses the upper-case row, `icontains` takes both
× É does not fold to é, in either direction
× $contains excludes the case-differing row and $icontains includes it
```

Both legs restored **by state**, not by exit code: `git diff HEAD` empty, `git status --short` empty, and on-disk `git hash-object` back to `2722d1ba…`. The pin re-ran `14 passed (14)` afterwards. No rebuild step is involved: the pin imports `../ValueDataSource` (relative source), so there is no `dist/` hop that could stale the reading.

## Verification, on the final commit `c889b18a7`

- `pnpm exec vitest run packages/core/` plus the 67 other test files that reach a value-provider datasource: **`Test Files 190 passed (190)` · `Tests 3364 passed (3364)`**.
- That 67-file set is a **measured narrowing, not a guess**: the change is confined to two functions in `ValueDataSource`, so a test can only observe it by (a) reaching that adapter and (b) passing one of the six operator spellings. The union above is condition (a) over the whole repo — 76 files across 12 packages, every one run. `turbo ls --affected` names 38 packages because `@object-ui/core` is a base dependency; those are declared to CI's four shards.
- `pnpm --filter @object-ui/core type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`): exit 0. The new file is in the program — `tsc -p tsconfig.test.json --listFiles` lists it (1 hit), lit control `adapters/ValueDataSource.ts` 1 hit, negative control 0 hits.
- `eslint .` in `packages/core`: `518 problems (0 errors, 518 warnings)`, all pre-existing `no-explicit-any`.
- `node scripts/check-changeset-presence.mjs`: `✅  2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/wise-poems-behave.md.`
- `node scripts/check-changeset-no-major.mjs`: `✅  No changeset declares a 'major' bump.`
- Control-character self-scan over the three touched source files: zero hits.

## Changeset level: `minor`

Scored `minor` because this **moves shipped results** — metadata relying on the lenient matching gets fewer rows and no error — and the repo forbids `major` (every package is in one fixed group, so one major carries all of them off the pinned `@objectstack` major; `check-changeset-no-major.mjs` enforces it). `patch` would be wrong: this is not a fix that leaves behaviour where it was. The changeset body names the migration explicitly — a filter that means "match regardless of case" is authored as `icontains` or `$icontains`, and both now execute.

## Dependency range: checked, no bump needed

`packages/core` declares `"@objectstack/spec": "^17.2.0"`. `asciiCaseInsensitiveContains` is present in **17.2.0**'s published `api-surface/data.json` and its `dist/data/index.d.ts` (read from the 17.2.0 tarball; positive control `canonicalAstOperator` present, negative control absent). So the declared range is honest and stays.

## Two out-of-scope findings, both MEASURED, neither fixed here

Reported for filing — `search_issues` returned `API rate limit already exceeded for user ID 323634890` on both dedupe queries, and repo-scoped REST is 403 for this seat, so I could not search and therefore did not file.

**A. `matchesFilter` waves through every `$` operator it does not recognise.** Same defect class as objectui#7349, in the dialect #7349 did not touch. Measured over rows `n` (`score: 5`) and `s` (`score: '5'`):

```
{ score: { $nin: [5,'5'] } }      -> ['n','s']   (must be [])
{ score: { $startsWith: 'zzz' } } -> ['n','s']   (must be [])
{ score: { $null: true } }        -> ['n','s']   (neither row is null)
{ score: { $exists: false } }     -> ['n','s']
```

The implemented set is `$gt $gte $lt $lte $ne $in $contains` plus the `$icontains` this PR adds; everything else is silently no-constraint. Not fixed here: closing it means refusing unknown operators the way `matchesASTFilter` does, which moves results for every one of those spellings — its own card, its own escalation.

**B. `not_contains` answers NO for a stored value that is not a string, so a row fails the operator AND its negation.** Measured: `['score','contains','5']` selects `['s']` (correct — a number cannot contain a substring) but `['score','not_contains','5']` selects `[]`, dropping row `n` from the negation too. The platform ruled this cell on 2026-09-05 (objectstack#14079, option A): a non-string never satisfies a positive text operator and **does** satisfy `$notContains`. `FILTER_TEXT_CASES` pins it on every face; `ValueDataSource` has exactly the shape the reference matcher had before that card. Different axis from this one (stored-value type, not case folding), so it is not ridden in here.

---

### DIFF
--- a/.changeset/wise-poems-behave.md
+++ b/.changeset/wise-poems-behave.md
@@ -1,0 +1,17 @@
+---
+"@object-ui/core": minor
+---
+
+`ValueDataSource`'s text operators answer the case question the wire answers: `contains` is case-SENSITIVE, `icontains` is its ASCII-folding twin (objectui#7379)
+
+The in-memory matcher lower-cased BOTH sides of `contains`, `not_contains`, `starts_with`, `ends_with` and — since the `icontains` arm was stacked onto the `contains` one — `icontains` too. So `contains` executed `icontains`, the two spellings named a single predicate, and a `provider: 'value'` list filtered with `contains` returned strictly more rows than the same filter run against a real driver. Nothing errored; the list was just longer, and both answers looked plausible.
+
+`$contains` is contractually case-sensitive (objectstack#4706 Q2 = A) and `$icontains` is the case-insensitive member, folding **ASCII only** (Q1 = A, because three of the five backends are SQLite underneath and its `lower()` folds ASCII only). All five drivers — `driver-sql`, `driver-sqlite-wasm`, `driver-turso`, `driver-mongodb`, `driver-memory` — plus objectql's `having` matcher import `FILTER_TEXT_CASES` from `@objectstack/spec/data` and answer its case rows. This adapter was the last face that did not.
+
+What changed:
+
+- `contains`, `not_contains`, `starts_with`, `ends_with` compare exactly. There is no `i` twin for the last three — `VALID_AST_OPERATORS` has `icontains` and nothing else with an `i` prefix — so case-sensitive is the only reading available to them, and it is the one `not_contains` needs so that no row can fail an operator *and* its negation.
+- `icontains` now uses the spec's own `asciiCaseInsensitiveContains`, so `CAFÉ` no longer matches `café`. `String.prototype.toLowerCase()`, which this arm used, is the full Unicode fold — a promise the SQL family cannot keep.
+- The `$`-dialect matcher follows: `$contains` compares exactly, and `$icontains` gains an arm. It had none, and an unrecognised `$` operator in that switch adds no constraint at all, so a `$icontains` filter used to select every row.
+
+**Behaviour change.** Metadata that relied on the lenient matching gets fewer rows and no error. A filter that means "match regardless of case" should be authored as `icontains` (AST/view dialect) or `$icontains` (`$` dialect); both now execute, and both fold ASCII case on either side.
--- a/packages/core/src/adapters/ValueDataSource.ts
+++ b/packages/core/src/adapters/ValueDataSource.ts
@@ -18,7 +18,7 @@ import type {
   AggregateParams,
   AggregateResult,
 } from '@object-ui/types';
-import { canonicalAstOperator } from '@objectstack/spec/data';
+import { asciiCaseInsensitiveContains, canonicalAstOperator } from '@objectstack/spec/data';
 import { emulateBatchTransaction } from './batchTransaction.js';
 
 // ---------------------------------------------------------------------------
@@ -122,23 +122,41 @@ function matchesComparisonNode(
       return Array.isArray(target) && target.includes(value);
     case 'nin':
       return Array.isArray(target) && !target.includes(value);
+    // -- Text operators. THE DIRECTION, so the next reader does not take it for
+    // a typo and fold it back (objectui#7379): the `$contains` FAMILY is
+    // case-SENSITIVE and `icontains` is its one case-insensitive member. That
+    // is the platform's ruling (objectstack#4706 Q2 = A), not this file's
+    // preference, and it is what every backend executes — `driver-sql`,
+    // `driver-sqlite-wasm`, `driver-turso`, `driver-mongodb` and
+    // `driver-memory` all import `FILTER_TEXT_CASES` (`@objectstack/spec/data`)
+    // and answer its `$contains is case-SENSITIVE` rows. These four arms used
+    // to lower-case BOTH sides, so `contains` executed `icontains`, the two
+    // were one predicate, and a `provider: 'value'` list answered a filter with
+    // strictly more rows than the same filter run against the wire.
+    //
+    // There is no `i` twin for the other three: `VALID_AST_OPERATORS` has
+    // `icontains` and NOTHING else with an `i` prefix — no `istartswith`, no
+    // `iendswith`, no `not_icontains` (the `$` dialect has no `$notIcontains`,
+    // and the AST table mirrors the executed set rather than widening it). So
+    // case-sensitive is the ONLY reading available to them, and it is the one
+    // `$notContains` needs for complementarity: a folding `not_contains` beside
+    // a case-exact `contains` lets one row fail an operator AND its negation.
     case 'contains':
-    case 'icontains': {
-      const lv = typeof value === 'string' ? value.toLowerCase() : '';
-      return typeof value === 'string' && lv.includes(String(target).toLowerCase());
-    }
-    case 'not_contains': {
-      const lv = typeof value === 'string' ? value.toLowerCase() : '';
-      return typeof value === 'string' && !lv.includes(String(target).toLowerCase());
-    }
-    case 'starts_with': {
-      const lv = typeof value === 'string' ? value.toLowerCase() : '';
-      return typeof value === 'string' && lv.startsWith(String(target).toLowerCase());
-    }
-    case 'ends_with': {
-      const lv = typeof value === 'string' ? value.toLowerCase() : '';
-      return typeof value === 'string' && lv.endsWith(String(target).toLowerCase());
-    }
+      return typeof value === 'string' && value.includes(String(target));
+    // The fold is ASCII-ONLY (objectstack#4706 Q1 = A) and it runs on BOTH
+    // sides, which is why this borrows the spec's own predicate instead of
+    // spelling one here. `String.prototype.toLowerCase()` — what this arm used
+    // to reach for — is the FULL Unicode fold, so it matched `CAFÉ` against
+    // `café`; three of the five backends are SQLite underneath, whose `lower()`
+    // folds ASCII only, so a Unicode promise here is one the wire cannot keep.
+    case 'icontains':
+      return typeof value === 'string' && asciiCaseInsensitiveContains(value, String(target));
+    case 'not_contains':
+      return typeof value === 'string' && !value.includes(String(target));
+    case 'starts_with':
+      return typeof value === 'string' && value.startsWith(String(target));
+    case 'ends_with':
+      return typeof value === 'string' && value.endsWith(String(target));
     case 'between':
       return Array.isArray(target) && target.length === 2 && value >= target[0] && value <= target[1];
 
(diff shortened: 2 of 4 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.**

## The one thing you flagged for me: keep `icontains` in, do not split

You fixed a second axis the card never named — `icontains` folded via `String.prototype.toLowerCase()` (full Unicode) where the contract folds **ASCII only** (objectstack#4706 Q1 = A, because three of five backends are SQLite and its `lower()` folds ASCII only) — and asked whether the PM wants it split out.

**Keep it.** Grounds:

- Splitting ships a **half-fix that makes the documented escape hatch wrong**. This card's whole migration story is "a filter meaning *match regardless of case* is authored as `icontains`". Landing `contains` alone would point authors at an operator that is, on this face, still wider than every driver — a worse state than today, because today at least the two are consistently wrong together.
- The narrowing you introduce is not a new divergence, it **removes one**: non-ASCII rows already fail to match under `icontains` on every real driver. This face was the outlier in both directions.
- It meets the bounded in-place bar on every count you list — same defect class, same file, mechanical, shape already pinned by the spec's own exported `asciiCaseInsensitiveContains`, no new verification surface.

The `minor` bump and the migration note in the changeset are right for exactly this reason: shipped results move.

## Three corrections against my brief, all accepted

**① My pin recipe was insufficient — `ValueDataSource` has TWO matchers.** A pin written only in the AST dialect passes green while `$contains` keeps folding. You pinned both. My recipe assumed one seam.

**② My "an operator-name assertion passes on a driver that returns everything" warning has a mirror I did not name.** An **inclusion** assertion is equally satisfied by a matcher returning **everything** — which is precisely what `$icontains` did before this PR (no arm at all, `default: break`, no constraint). So "the inclusion half rules out the `[]` implementation" is only half the guard. Your explicit not-everything assertion for that dialect is the other half, and I am carrying it into future briefs.

**③ My escalation fence would have misfired on a survey scoped to this repo.** Inside objectui, `ValueDataSource` very nearly *is* the only driver — `data-objectstack` is a translator, `ApiDataSource` and the runner's mock implement no text operator at all. The deciding siblings all live in **objectstack**. A seat that read my fence literally and surveyed only objectui would have found no sibling and **escalated a defect as a ruling**. The fence should have said *survey the drivers wherever they live*.

Also noted: an existing green test **pinned the bug** — `{ name: { $contains: 'ali' } }` expecting Alice was reachable only under folding, and its comment shows the author's model was plain substring. Retriaging it to `'Ali'` rather than deleting it is the right call.

## Your two findings

Both are real and neither belongs in this PR. I am filing them:

- **A** — the `$`-dialect matcher's `default: break` waves through every unrecognised operator with no constraint (`$nin`, `$startsWith`, `$null`, `$exists` all measured returning every row). Same class as objectui#7349, in the dialect that card did not touch.
- **B** — `not_contains` on a non-string stored value fails the operator **and** its negation, against the platform's 2026-09-05 ruling (objectstack#14079 option A).

Both move results for spellings beyond this card's scope, so each wants its own card and its own fence — exactly as you scoped them.

---

PR #8456, opened 2026-09-08
fix(core): lower $and / $or to real AST group nodes in convertFiltersToAST

### LINKED ISSUE #6948: convertFiltersToAST has no branch for $and / $or / $not, so spec-legal FilterCondition combinators never reach the wire
Split out of #4664 (PR #6946). Pre-existing on every consumer of the shared
filter sink; that card newly makes it reachable by an author following the
spec's own type, which is why it is being written down now.

## What happens

`convertFiltersToAST` (`packages/core/src/utils/filter-converter.ts`) — the
repo's ONE lowering from the MongoDB-style filter object to the ObjectQL AST,
reached through `toFilterNode` / `mergeFilterNodes` — has no branch for the
logical combinators. Measured against `packages/core/dist` at
`40c479af2`, composing each condition with a parent scope:

```
$or    ->  ["and",["task_version","=","tv-1"],["$or","=",[{"status":"open"},{"status":"blocked"}]]]
$and   ->  ["and",["task_version","=","tv-1"],["$and","=",[{"status":"open"},{"is_active":true}]]]
$not   ->  THROWS FilterOperatorError: Unknown filter operator 'status' for field '$not'.
plain  ->  ["and",["task_version","=","tv-1"],["status","!=","archived"]]   (correct)
```

Two distinct failures, neither of them the right answer:

- `$and` / `$or` fall through to the **simple-equality** branch (their value is
  an array, so the operator loop is skipped) and produce a leaf naming a field
  literally called `$and` / `$or`. That is a well-formed AST node carrying a
  nonsense field, so the server refuses it (`400 INVALID_FILTER`) — loud, but
  the diagnostic points at the author's field list rather than at the missing
  lowering.
- `$not` enters the operator loop with its OWN nested object's keys read as
  operators, so the thrown message names a nonsense operator
  (`Unknown filter operator 'status' for field '$not'`) and lists the supported
  operators — none of which is what the author needs to hear.

## Why it matters now

`Field.relatedListFilter` (`@objectstack/spec` 17.1.0, objectstack#8704 /
PR #8955) is typed `FilterConditionSchema`, and `FilterCondition` explicitly
declares all three:

```ts
$and?: FilterCondition[];
$or?: FilterCondition[];
$not?: FilterCondition;   // NULL-safe per #5146
```

So the spec tells a metadata author these are legal in the very key #4664 wired
up, while this repo's lowering cannot carry any of them to the wire. The same
gap applies to every other producer that reaches this sink — the component-level
`record:related_list.filter` (objectstack#7118), `ListView.filter` /
`ViewTab.filter` via `buildEffectiveFilter`, and `plugin-view`'s ObjectView —
so this is a sink-level bug, not a related-list one.

Also worth deciding as part of the fix: `$not` is NULL-safe by ruling
(objectstack#5146, maintainer 2026-08-04 — `NOT (…) OR col IS NULL`, with the
guard on each LEAF inside the negation, not hoisted). Whatever lowering is
chosen has to preserve that or hand the negation to the server intact.

## Why it was not fixed in #4664

Changing the shared sink changes what every one of those callers puts on the
wire, so it needs its own card, its own fixture set and its own reverse
verification. Working around it renderer-side — accepting the combinators in
one consumer and lowering them there — is precisely the second de-facto
contract AGENTS.md #0.1 forbids, so #4664 carried the value verbatim to the
existing sink and named this boundary in its PR body instead.

## Where

- `packages/core/src/utils/filter-converter.ts` — `convertFiltersToAST`
  (the missing combinator branch), `toFilterNode`, `mergeFilterNodes`

Filed unassigned by the #4664 execution seat. No dedupe hit: a targeted issue
search for the combinator lowering returned only objectui#4744 (a different
function, `ListView.convertFilterGroupToAST`, closed), with a control query in
the same session returning #4664 itself.

---

os-justin:
Fixes #6948

## The card's premise is half wrong, and that changes what this PR is

Re-derived on `ca3942729` (the branch base) against `@objectstack/spec` **17.3.0** — the card measured `packages/core/dist` at `40c479af2`.

**Still holds, byte for byte.** Composing each condition with a parent scope, `mergeFilterNodes({task_version:'tv-1'}, f)` on the base commit reproduces every emission the card printed:

```
$or    ->  ["and",["task_version","=","tv-1"],["$or","=",[{"status":"open"},{"status":"blocked"}]]]
$and   ->  ["and",["task_version","=","tv-1"],["$and","=",[{"status":"open"},{"is_active":true}]]]
$not   ->  THROWS FilterOperatorError: Unknown filter operator 'status' for field '$not'.
plain  ->  ["and",["task_version","=","tv-1"],["status","!=","archived"]]
```

The file, the three functions and the two mechanisms (array value skips the operator loop; `$not`'s own nested keys read as operators) all hold. `filter-converter.ts` has not moved on `main` since `2a9513d81`.

**Falsified.** The card says the `$and` / `$or` leaf is *"a well-formed AST node carrying a nonsense field, so the server refuses it (`400 INVALID_FILTER`)"*. Measured against the spec's own doors:

```
isFilterAST(['$or','=',[{status:'open'},{status:'blocked'}]])   -> true
parseFilterAST(['$or','=',[{status:'open'},{status:'blocked'}]]) -> {$or:[{status:'open'},{status:'blocked'}]}
```

A `[field, '=', value]` node lowers to `{ [field]: value }`, and `$or` is a legal `FilterCondition` key — so the leaf round-trips into a *real* combinator and the server was never refusing anything. This repo already knew: `data-objectstack/src/filter-entry-translation.test.ts` pinned that exact round trip with the note "verified, and the reason this is NOT rewritten here."

**So the failure mode being fixed is neither "never reaches the wire" nor "reaches it and is ignored."** It is: *reaches the wire and is interpreted correctly there, and is mis-evaluated — silently, to zero rows — by every AST evaluator inside this repo.* `['$or','=',[...]]` is a well-formed **comparison** node, so `ValueDataSource`'s `matchesComparisonNode` reads `$or` as a field name, looks up `record['$or']`, finds nothing, and excludes every row. No throw, no console line, an empty list where the author asked for a union.

## Reachability — which consumers the claim covers

The card called this sink-level rather than related-list-level. That holds and is stronger than stated: `$and` has live in-repo producers today, not only the newly-typed `Field.relatedListFilter`.

| producer | shape emitted |
|---|---|
| `core/src/utils/merge-filters.ts` `mergeFilters` | `{ $and: [a, b] }` |
| `plugin-dashboard/src/DashboardRenderer.tsx:851,858` | via `mergeFilters` (scope-filter broadcast) |
| `plugin-report/src/DatasetReportRenderer.tsx:1429,1447` | via `mergeFilters` (report / block runtime filter) |
| `plugin-dashboard/src/ObjectMetricWidget.tsx:413`, `DrillDownDrawer.tsx:106` | `{ $and: [existing, resolved] }` |
| `fields/src/widgets/FilterConditionField.tsx:209,213` | `{ $or: [...] }` / `{ $and: [...] }` |

The only in-repo AST **evaluator** is `ValueDataSource`'s `matchesASTFilter` (grep for `node[0]`); every other consumer forwards the node to a `DataSource`.

## The change

`$and` / `$or` lower to `['and'|'or', ...children]` — the spelling the spec's own `FILTER_ARRAY_LOGIC_KEYWORDS` declares — with children lowered recursively. The wire condition is **unchanged** (pinned both ways); what changes is that the node is executable as well as parseable, and that operators inside a combinator branch now meet the same guard every other operator meets (a `$bogus` or a `$regex` inside a `$or` branch used to travel to the wire unchecked inside the leaf's value slot).

Boolean identities (objectstack#5322) are handled at the boundary of the new branch, because a new branch has to decide them:

- `{ $and: [] }` is TRUE, so it **drops out**. Not `['and']`: measured, `isFilterAST(['and'])` is `false` and `parseFilterAST(['and'])` is `undefined` — no filter at all, i.e. **every row**, the one direction that must never happen.
- `{ $or: [] }` is FALSE, which the AST cannot spell, so it **keeps the leaf it already had** — measured to answer zero rows at the wire (`{$or: []}`) *and* in the in-memory matcher. Documented in place, with the alternatives and why each is worse.
- A `{}` disjunct is TRUE and absorbs its `$or`; a `{}` conjunct drops out of its `$and`.

## `$not` is refused, and it is an open contract question

The ObjectQL AST has **no negation**: `FILTER_ARRAY_LOGIC_KEYWORDS` is `['and','or']` and `VALID_AST_OPERATORS.has('not')` is `false` (measured, 53 members). Rewriting the negation inward is not available either — `startswith`, `endswith`, `between` and `icontains` have no negated counterpart in that set, so a De Morgan lowering would be silently partial and would quietly drop the NULL-safe rule of objectstack#5146.

So `$not` throws with an accurate message instead of one naming the author's own nested field as a bogus operator. **It threw before this change too, so the verdict is unchanged and only the diagnostic moved** — this PR does not decide the contract. Whether the AST should gain a negation, or whether this sink may hand a `FilterCondition` object to the wire when a `$not` is present, is raised on the card for the maintainer.

## The pin — row sets, both directions, and what a worse implementation would do

`packages/core/src/utils/__tests__/filter-combinators-6948.test.ts` (24 cases). Nothing greps the source; every assertion is either the node that reaches the wire put through `isFilterAST` / `parseFilterAST`, or the row set a real `ValueDataSource` returns.

Both halves are asserted because each alone passes on something worse than the bug:

- **inclusion half** — `selectedIds(...)` is `['open-active','blocked-idle']`, exactly. A converter that emits nothing usable leaves `$filter` unread and returns **every** row (`selectedIds(undefined)` is asserted equal to `ALL_IDS` as the lit control).
- **exclusion half** — `done-active`, `null-status`, `no-status-key` are absent. A converter emitting a node no evaluator reads returns **no** row — which is the bug itself, pinned as `PRE_FIX_OR_NODE` selecting `[]`.
- a case asserting the answer is neither `ALL_IDS` nor `[]`, and that the fixture is not trivially either.
- a scoped case where one row is in the union but fails the scope and another passes the scope but is outside the union — both directions in one assertion.

Refusals assert the `INVALID_FILTER` / `400` envelope plus the message clause, never a bare `toThrow()`.

## Ablation — the read site, from the committed implementation

Removed the combinator dispatch block from `convertFiltersToAST` (the read site, not the helper), proved it reached disk, ran, restored by state.

```
HEAD blob = 5f99be95293258673b88816bdffa2004d8deec6f
disk blob = aec6ab2b7d1cc723422658edfc677bbe12fca336   (differs -> mutation on disk)
anchor 'const logicKeyword = AST_LOGIC_KEYWORD[field];'  BEFORE 1 -> AFTER 0
marker 'ABLATED objectui#6948' AFTER 1, at line 234
```

**19 named red rows across both packages**, including `lowers $or to an AST group node`, `INCLUDES both branches of the union — not empty`, `$and intersects, both directions`, `a parent scope still narrows a union it wraps`, `nested combinators evaluate as written`, `drops $and: [] — the TRUE identity constrains nothing`, `runs the unknown-operator guard on combinator children`, and both routes of `lowers a top-level Mongo logical node to an AST group`.

Two rows deliberately stayed **green** under ablation and that is the point: `EXCLUDES the rows outside the union` (the ablated converter excludes *everything*, so it trivially excludes those) and `the pre-fix node selected NOTHING` (a control on a literal node, independent of the implementation). That is exactly why the inclusion half exists.

The `data-objectstack` pin going red under a `packages/core` source edit is also the proof that **no rebuild leg is owed**: the root `vitest.config.mts` aliases `@object-ui/core` to `packages/core/src`, so there is no `dist` hop to stale.

Restore verified by state, not by exit code: `git diff HEAD` empty and `hash-object` equal to `rev-parse HEAD:path`.

## Changeset

`.changeset/filter-ast-combinators-6948.md` — `@object-ui/core: minor`, `@object-ui/data-objectstack: patch`. Verdict line:

```
✅  2 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s): .changeset/filter-ast-combinators-6948.md.
```

`minor` because shipped results move: a list filtered by a combinator through any in-process data source goes from zero rows to the rows the author asked for, and an unknown operator inside a combinator branch is now refused rather than shipped. Same shape as the landed precedent PR #8437. `major` is forbidden here (fixed group); `check-changeset-no-major.mjs` exits 0. `skip-changeset` deliberately not applied — it is a phantom label in this repo.

## Verification run

At `df25f4b1a`:

| command | result |
|---|---|
| `pnpm exec vitest run packages/core/ packages/data-objectstack/ packages/fields/ packages/plugin-dashboard/ packages/plugin-report/` | `VERDICT command-exit 0` — 429 files, 6775 tests |
| `pnpm exec vitest run <the two pins> packages/plugin-list/ packages/plugin-grid/ packages/plugin-detail/ packages/plugin-form/ packages/plugin-view/` | `VERDICT command-exit 0` — 442 files, 4360 passed / 1 skipped |
| `pnpm exec vitest run packages/components/ packages/react/` | `VERDICT command-exit 0` — 313 files, 3100 tests |
| `pnpm --filter '@object-ui/core' --filter '@object-ui/data-objectstack' type-check` | exit 0 (both test files proven in-program via `--listFiles`, with a lit and a dark control) |
| `pnpm --filter '@object-ui/core' --filter '@object-ui/data-objectstack' lint` | exit 0 — 0 errors |
| `node scripts/check-control-bytes.mjs` | `✅ OK (6668 tracked text files)` |
| `node scripts/check-governed-queue-guard.mjs --test <the 4 paths>` | `✅ NOT GOVERNED` |

**Declared narrowing:** 1184 test files run locally, covering every direct consumer of the sink and every in-repo producer of `$and` / `$or` enumerated above. `app-shell`, `console`, `site`, the examples and the remaining plugins are declared to CI — their `$and` references (`ObjectFieldInspector`, `datasetFilterCondition`) are metadata-authoring conversions that never reach `convertFiltersToAST`.

**Merge note.** `ValueDataSource.ts` moved on `main` after this branch was cut (objectui#7379: text-operator case sensitivity). The pin here uses only `=` and `>=` on non-string and string-equality fields, so it touches none of the drifted arms; `filter-converter.ts` itself is unmoved since `2a9513d81`.

---

### DIFF
--- a/.changeset/filter-ast-combinators-6948.md
+++ b/.changeset/filter-ast-combinators-6948.md
@@ -1,0 +1,30 @@
+---
+'@object-ui/core': minor
+'@object-ui/data-objectstack': patch
+---
+
+`convertFiltersToAST` lowers the `$and` / `$or` combinators to real ObjectQL AST
+group nodes (objectui#6948).
+
+`FilterCondition` declares `$and` / `$or` / `$not`, and this repo's one lowering
+had no branch for any of them. `$and` / `$or` fell through to the
+simple-equality branch and became a leaf naming a field literally called `$and`
+/ `$or`. That leaf reached the server intact — `parseFilterAST` reads
+`['$or', '=', [...]]` back as a real `$or`, so the wire condition was correct
+and is unchanged by this release — but it is a well-formed *comparison* node, so
+every AST evaluator in this repo read `$or` as a field name, found no such key
+on any record, and returned an EMPTY list with no error. Producers that reach
+this today include `mergeFilters` (dashboard scope broadcast, dataset report
+blocks), `FilterConditionField`, and `Field.relatedListFilter`.
+
+`minor` rather than `patch`: shipped results move. A list filtered by a
+combinator through any in-process data source went from zero rows to the rows
+the author asked for, and an unknown or refused operator *inside* a combinator
+branch — which used to travel to the wire unchecked inside the leaf's value slot
+— is now refused at the same door as every other operator.
+
+`$not` is refused with an accurate message instead of translated: the AST has no
+negation keyword (`FILTER_ARRAY_LOGIC_KEYWORDS` is `['and', 'or']`) and several
+operators it carries have no negated counterpart, so a rewrite would be silently
+partial. It threw before this change too, naming the author's own nested field
+as a bogus operator; the verdict is unchanged, only the diagnostic.
--- a/packages/core/src/utils/__tests__/filter-combinators-6948.test.ts
+++ b/packages/core/src/utils/__tests__/filter-combinators-6948.test.ts
@@ -1,0 +1,345 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `$and` / `$or` / `$not` through the repo's ONE lowering — objectui#6948.
+ *
+ * ## What this file asserts, and why it is row sets and not source shape
+ *
+ * The card asked for a branch. A branch existing is not the property that
+ * matters, so nothing here greps the source: every assertion is either the node
+ * that REACHES THE WIRE, put through the spec's own doors (`isFilterAST` /
+ * `parseFilterAST`, `@objectstack/spec/data`), or the ROW SET a real
+ * `ValueDataSource` returns for it.
+ *
+ * Both halves are pinned in both directions, because each half alone passes on
+ * an implementation strictly worse than the bug:
+ *
+ *   - an assertion that a combinator is PRESENT passes on a builder that emits
+ *     everything;
+ *   - an assertion that a row set is EMPTY passes on a converter that emits
+ *     nothing at all — `$filter: undefined` returns EVERY row, and a node no
+ *     evaluator reads returns NONE.
+ *
+ * So every row-set case names the exact ids, and {@link ALL_IDS} is asserted to
+ * be neither of them.
+ *
+ * ## What was actually wrong — NOT what the card said
+ *
+ * The card stated that the pre-fix emission for `$and` / `$or` — a leaf naming a
+ * field literally called `$and` / `$or` — was refused by the server with `400
+ * INVALID_FILTER`. Re-measured against `@objectstack/spec` 17.3.0 that is FALSE,
+ * and this repo already knew: `data-objectstack/src/filter-entry-translation.test.ts`
+ * pins the same round trip with the note "verified". `parseFilterAST` lowers
+ * `['$or', '=', [...]]` to `{ $or: [...] }` — the FilterCondition the author
+ * wrote — because a `[field, '=', value]` node becomes `{ [field]: value }` and
+ * `$or` is a legal FilterCondition key. So the combinator DID reach the wire and
+ * WAS interpreted correctly there. {@link WIRE_UNMOVED} pins that it still is.
+ *
+ * The real defect is one door further in, and it is silent: `['$or', '=', [...]]`
+ * is a well-formed COMPARISON node, so every AST evaluator in this repo reads
+ * `$or` as a field name. `ValueDataSource`'s matcher looks up `record['$or']`,
+ * finds nothing, and excludes every row — no error, no console line, an empty
+ * list where the author asked for a union. That is the failure this file fixes
+ * in place, and {@link PRE_FIX_OR_NODE} is the control that it really did.
+ *
+ * ## `$not` is refused, and that is an open contract question
+ *
+ * The ObjectQL AST has no negation: `FILTER_ARRAY_LOGIC_KEYWORDS` is
+ * `['and', 'or']`. Rewriting the negation inward is not available either —
+ * `startswith`, `endswith`, `between` and `icontains` have no negated
+ * counterpart in `VALID_AST_OPERATORS`, so a De Morgan lowering would be
+ * silently partial, and `$not` is NULL-safe by ruling (objectstack#5146), which
+ * a partial rewrite would quietly drop. So `$not` is refused with an accurate
+ * message instead of translated. It threw before this card too — with a message
+ * naming a nonsense operator — so this changes the diagnostic, not the verdict;
+ * whether the AST should GAIN a negation is a spec decision, not this file's.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { isFilterAST, parseFilterAST } from '@objectstack/spec/data';
+import { convertFiltersToAST, mergeFilterNodes } from '../filter-converter';
+import { ValueDataSource } from '../../adapters/ValueDataSource';
+
+// ---------------------------------------------------------------------------
+// Fixture — one row per branch of the combinators under test
+// ---------------------------------------------------------------------------
+
+const ROWS = [
+  { id: 'open-active', status: 'open', is_active: true, age: 30 },
+  { id: 'blocked-idle', status: 'blocked', is_active: false, age: 20 },
+  { id: 'done-active', status: 'done', is_active: true, age: 40 },
+  { id: 'null-status', status: null, is_active: null, age: null },
+  { id: 'no-status-key' },
+];
+
+/** What "no filter reached the evaluator" looks like. Never an expected answer. */
+const ALL_IDS = ['open-active', 'blocked-idle', 'done-active', 'null-status', 'no-status-key'];
+
+async function selectedIds(filter: unknown): Promise<string[]> {
+  const ds = new ValueDataSource({ items: ROWS });
+  const result = await ds.find('tasks', { $filter: filter as any });
+  return result.data.map((r: any) => r.id as string);
+}
+
+/** The authored filter of the card: a two-branch union. */
+const AUTHORED_OR = { $or: [{ status: 'open' }, { status: 'blocked' }] };
+
+/** Its emission BEFORE this card — a comparison node on a field named `$or`. */
+const PRE_FIX_OR_NODE = ['$or', '=', [{ status: 'open' }, { status: 'blocked' }]];
+
+// ---------------------------------------------------------------------------
+// 1. The lowering — the node that reaches the wire
+// ---------------------------------------------------------------------------
+
+describe('objectui#6948 — the lowering', () => {
+  it('lowers $or to an AST group node, children lowered recursively', () => {
+    expect(convertFiltersToAST(AUTHORED_OR)).toEqual([
+      'or',
+      ['status', '=', 'open'],
+      ['status', '=', 'blocked'],
+    ]);
+  });
+
+  it('lowers $and the same way, and lowers operator children too', () => {
+    expect(convertFiltersToAST({ $and: [{ status: 'open' }, { age: { $gte: 18 } }] })).toEqual([
+      'and',
+      ['status', '=', 'open'],
+      ['age', '>=', 18],
+    ]);
+  });
+
+  it('nests, so a combinator inside a combinator survives as a group', () => {
+    expect(
+      convertFiltersToAST({
+        $and: [{ $or: [{ status: 'open' }, { status: 'done' }] }, { is_active: true }],
+      }),
+    ).toEqual([
+      'and',
+      ['or', ['status', '=', 'open'], ['status', '=', 'done']],
+      ['is_active', '=', true],
+    ]);
+  });
+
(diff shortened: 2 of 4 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** Two rulings below; neither is a product call, both are recordings of what your measurement already settled.

## The card's account of the failure is falsified, and that is the headline

The card said the `$and`/`$or` leaf is refused by the server with 400 INVALID_FILTER. Measured: `isFilterAST(['$or','=',[...]])` is **true**, `parseFilterAST` returns `{$or:[...]}` — the exact `FilterCondition` the author wrote. And **this repo already knew**: `filter-entry-translation.test.ts` pinned that round trip with the word *"verified"*.

⇒ The real failure is **neither of the two modes my brief named**. It is a third: the node **reaches the wire and is interpreted correctly there**, then is **mis-evaluated to zero rows by the in-repo AST evaluator**, because a well-formed comparison node makes `matchesComparisonNode` read `$or` as a *field name*. `selectedIds(['$or','=',[...]])` → `[]` before, `['open-active','blocked-idle']` after.

My brief told you to establish which of two failure modes you were fixing. The correct answer was "neither" — that is the most useful thing this report contains.

## Ruling on `$not`: **D now, A next — and B and C are ruled OUT explicitly**

You asked for B and C to be ruled out by name rather than left as "not chosen". Agreed, and here is why each is a regression rather than a cheap answer:

- ⛔ **B (hand a `FilterCondition` object to the wire)** — measured worse, not neutral: `mergeFilterNodes` composes as `['and', parentScope, child]`, so an object in AST child position makes `isFilterAST` false for the **whole** filter; and `ValueDataSource` routes a non-array `$filter` to `matchesFilter`, whose switch has no `$not` arm and ends in `default: break`. The negation is silently dropped and **the list WIDENS**. A filter that quietly returns more rows than asked is the worst outcome available here.
- ⛔ **C (emit the accidental `['$not','=',{...}]` leaf)** — correct at the wire, wrong at every in-repo evaluator: a field named `$not`, zero rows. That is **the exact defect this PR just fixed for `$and`/`$or`, re-introduced for `$not`**, and it would enshrine an accident as a contract.

**D stands** because it decides nothing: `$not` threw before this card (naming the author's own nested field as a bogus operator) and throws now with an accurate message and the same envelope. Improving a diagnostic without moving a verdict is not a contract decision, and keeping the arm is right — a misleading error is a second defect.

**A is not this repo's to make.** `FILTER_ARRAY_LOGIC_KEYWORDS` is `['and','or']` and `VALID_AST_OPERATORS.has('not')` is false, and you established that a De Morgan rewrite cannot complete either — `startswith` / `endswith` / `between` / `icontains` have no negated counterpart, and a partial rewrite would silently drop objectstack#5146's NULL-safe rule, whose guard belongs on each leaf. That is an `objectstack` spec card. It goes to the maintainer with the rest of the cross-repo queue rather than being filed by this seat into the wrong repository.

## Ruling on the flipped pin: **A — accept the flip**

You were right to flag it rather than do it quietly; a green pin carrying the word *"verified"* is exactly what a dev should not overturn on their own. Accepted, because its stated concern was **the wire**, and that half is not merely preserved but promoted from narration to assertion — you added the companion case proving both spellings lower to the same `FilterCondition`. What the note never covered is the evaluator one door in, which is the whole defect. B (a second lowering for one shape) is the AGENTS.md #0.1 shape and correctly rejected.

## The ablation reproduced my own warning, live

Two rows stayed **green** under ablation: *'EXCLUDES the rows outside the union'* (the ablated converter excludes everything) and *'the pre-fix node selected NOTHING'* (a control on a literal node). That is precisely the hazard my brief named — **an emptiness assertion passes on a builder that emits nothing** — demonstrated on this PR's own pin rather than argued. It is why the inclusion half asserting the exact ids is load-bearing, and it is worth more than a paragraph of reasoning would have been.

Also noted: reachability is **stronger than the card claimed** — `$and` has live in-repo producers today (`mergeFilters` into `DashboardRenderer` / `DatasetReportRenderer`, `ObjectMetricWidget`, `DrillDownDrawer`, `FilterConditionField`), not only `Field.relatedListFilter`.

## Your finding is already filed

The `matchesFilter` `default: break` finding is objectui#8447, filed from objectui#7379's return before your search was refused. Your version adds the full unconstrained set and the *"one adapter, two answers, selected by whether `$filter` arrives as an array or an object"* framing — I am adding both to that card rather than opening a duplicate.

The `FilterConditionField.isMatchAllCriteria` observation (it treats `{ $or: [] }` as match-all where objectstack#5322 rules it FALSE) is recorded as you framed it: the divergence is in the **safe** direction for a sharing-rule warning guard, and it is noted so the two readings are not later mistaken for a contradiction that needs "fixing".

---

PR #8460, opened 2026-09-08
fix(plugin-detail): copy object-valued cells as JSON, not `[object Object]`

### LINKED ISSUE #8395: DetailSection's click-to-copy writes `[object Object]` to the clipboard for every object-valued cell (address, geolocation, JSON, attachment)
Filed by the objectui#8376 dev (branch `claude/issue-8376-detailsection-emptiness-authority`) while converging that file's emptiness tests. ⛔ Not claimed. Out of scope there: #8376 is about which rows count as EMPTY; this is about what the copy affordance WRITES for a row that is filled, and fixing it does not move any emptiness predicate.

## What

`packages/plugin-detail/src/DetailSection.tsx`:

```ts
const handleCopyField = React.useCallback((fieldName: string, value: any) => {
  const textValue = value !== null && value !== undefined ? String(value) : '';
  navigator.clipboard.writeText(textValue).then(...)
```

`String(value)` on an object is `'[object Object]'`. The affordance is offered for every non-empty value — objects included — so on a detail page a reader who clicks the copy button (or the row, or presses Enter on it) on an **address**, **geolocation**, **JSON**, **attachment/file**, **multiselect** or expanded **lookup** cell silently gets `[object Object]` on the clipboard.

Reachable in read mode on any non-editable object-valued row, which is the common case: the pencil affordance replaces copy only when the field is inline-editable AND the parent threaded `onEnterInlineEdit`.

## Why it is worth fixing rather than removing

The cell next to the button is already rendering that value **correctly** — `AddressCellRenderer` prints a formatted postal address (objectui#4037), `LocationCellRenderer` prints `30.2741, 120.1551`, `SelectCellRenderer` prints option badges. So the rendered text the reader is looking at is exactly what they expect to copy, and the clipboard gets a placeholder string instead. The failure is silent: nothing errors, and it is only noticed on paste.

## Fix shape (a suggestion, not a ruling)

Two candidates, and a taker should decide rather than assume:

- **A — copy the rendered text.** Read the row's rendered `textContent`. Matches what the reader sees, and needs no per-type table. Costs a DOM read at click time and depends on the renderer's formatting (a truncated value would have to be handled — `TruncatedText` is in this path).
- **B — a shared value→text formatter.** Serialize per type (`formatAddress` for address, `lat, lng` for geolocation, `JSON.stringify` for json, the option labels for multiselect). Deterministic and testable, but it is a **second** formatting authority next to the cell renderers — the exact drift shape objectui#4037 closed for `formatAddress`, so it should reuse those functions rather than re-spell them.

⛔ Not "don't offer copy for objects": the value IS copyable and the reader has a legitimate reason to want it.

The pin should assert what reaches `navigator.clipboard.writeText`, per type, from a rendered `DetailSection` — with a control asserting an ordinary string row still copies its own value unchanged.

## Measured

Read on branch `claude/issue-8376-detailsection-emptiness-authority` at `b6ed5cdaf`. `canCopy` is `hasCellValue(value)` as of #8376, and that function classifies every object as a value (deliberately — see its docblock), exactly as the raw test it replaced did. So #8376 neither introduces nor widens this; the `String(value)` line is untouched by it.

## Dedup

`search_issues` once, self-validating (it returned a card that IS about this file's object-valued cells): the clipboard / `[object Object]` / detail copy wording returned **exactly one** issue, objectui#4037 — the address cell renderer, closed. #4037 is read-side only and says nothing about the clipboard; its own body scopes itself to the display registry. Nothing names the clipboard write.

## Related

objectui#4037 (`formatAddress`, the shared address formatting authority a fix B would reuse) · objectui#8376 (`hasCellValue`, which decides whether the affordance is offered at all)

os-justin:
Fixes #8395

`DetailSection`'s click-to-copy built its payload with `String(value)`, and `String()` of an object is the literal text `[object Object]`. The affordance is offered for every non-empty value — objects included (`canCopy` is `hasCellValue`, objectui#8376) — so a reader clicking the copy button, the row, or pressing Enter on an address, geolocation, JSON, file, expanded-lookup, repeater or image cell silently got a placeholder on the clipboard **while the cell beside the button rendered that same value correctly**. Nothing errors; it is noticed only on paste.

**The change is one read site**: objects are serialized with `JSON.stringify`, everything else keeps `String()`.

## Per-kind payloads — measured, not reasoned

Rendered via a real `DetailSection` at `innerWidth` 1280; the observed value is the ARGUMENT handed to `navigator.clipboard.writeText`.

| field kind | the cell renders | clipboard BEFORE | clipboard AFTER |
|---|---|---|---|
| address | `1 Main St, Springfield, IL 62704, USA` | `[object Object]` | `{"street":"1 Main St","city":"Springfield","state":"IL","postalCode":"62704","country":"USA"}` |
| geolocation | `30.2742, 120.1551` | `[object Object]` | `{"latitude":30.2741567,"longitude":120.1551234}` |
| json | `{"a":1,"b":["x","y"]}` | `[object Object]` | `{"a":1,"b":["x","y"]}` |
| file (single) | `contract.pdf` | `[object Object]` | `{"name":"contract.pdf","url":"https://cdn.example.com/contract.pdf"}` |
| lookup (expanded) | `Acme Corp` | `[object Object]` | `{"id":"acct-1","name":"Acme Corp"}` |
| file (array) | `2 files` | `[object Object],[object Object]` | `[{"name":"a.pdf"},{"name":"b.pdf"}]` |
| repeater | a count | `[object Object],[object Object]` | `[{"label":"one"},{"label":"two"}]` |
| image | an `img` element, no text | `[object Object]` | `{"name":"logo.png","url":"https://cdn.example.com/logo.png"}` |
| multiselect | `AlphaBeta` | `alpha,beta` | `["alpha","beta"]` |

⚠️ **One payload moves without having been broken** — the multiselect. An array IS an object, so `['alpha','beta']` now copies as a JSON array. The new form is lossless where the old one was ambiguous for any value containing a comma. It is called out here rather than buried: it is the only row whose BEFORE column was not already a placeholder.

## ⭐ The non-regression half

Every correct-today scalar payload is **byte-identical**, and the pin asserts the rendered text beside it, so it is red for any later "just copy what the cell renders" rewrite:

| field kind | the reader SEES | the clipboard GETS (unchanged) |
|---|---|---|
| number (scale 2) | `16.00` | `16` |
| currency | `1,234.50` | `1234.5` |
| percent | `12%` | `0.123` |
| date | `Mar 4` | `2026-03-04` |
| select | `Closed Won` | `won` |
| text | `Plain String Value` | `Plain String Value` |
| boolean | (a checkbox, no text) | `true` |

## What this deliberately does NOT decide

- ⚠️ **The JSON blob is a defensible default, not a settled contract.** The read site now says so in place, naming objectui#8395's **option B** — a shared value-to-text formatter reusing the cell renderers' own formatters (`formatAddress`, objectui#4037) — as where the real per-kind contract gets decided. That needs product answers that do not exist yet (address under which locale, geolocation at stored precision or displayed rounding, json compact or pretty, file filename or list or URL, multiselect labels or values, lookup name or id, image URL or filename) and is a separate card.
- ⛔ **Copying the rendered text (option A) was measured and rejected.** It is the worse contract for 9 of 17 field types and loses data silently: `date` renders `Mar 4` (the year is gone), `percent` renders `12%` against a stored `0.123` (a different quantity), `datetime` concatenates to an unparseable `3/4/20265:06 am`, `image` and `boolean` render no text at all — so those two would copy the EMPTY STRING, an implementation strictly worse than the defect. It would also make the payload display-locale dependent.
- ⛔ **Withdrawing the affordance from text-less cells (option D) stays out**: it would narrow `canCopy` away from `hasCellValue`, whose three readers must agree per objectui#8376's docblock.
- ⛔ **objectui#8440 is not addressed here** and its code path is byte-for-byte untouched: the `password` / `secret` branch of this same handler (copy writes the raw secret while the cell renders the mask) is in the maintainer's decision box. The pin asserts nothing about it on purpose, so whichever way that card is answered, no case in this file turns red. That card remains open.

## The one addition beyond the literal one-liner

`JSON.stringify` throws on a cycle, and this runs inside a click handler. The stringify is therefore wrapped in the SAME `try` / `catch` that `JsonCellRenderer` already applies to this exact operation on this exact value, falling back to today's string form. It is pinned (a self-referencing value still copies `[object Object]` and the click does not throw) and is the only input for which that text is still written. Strike it if you would rather have the bare call.

## Ablation — proof the pin can fail

Run from the committed implementation, restoring the copy handler's bare `String(value)`. ⚠️ There is a SECOND `String(value)` in this file — `displayValue`'s no-renderer fallback — which is NOT this defect; the mutation is anchored on the handler body and that second site is asserted unmoved.

```
HEAD blob            014ec9dc9f356d7c2f0267c5c4f1389cfca70324
on disk, mutated     6f5bde22876b224f753718e4a028eb41a9d932c3   ⇒ the mutation reached disk
anchors  JSON.stringify(value)                1 -> 0
         bare one-liner (copy handler)        0 -> 1
         '      return String(value);' (L420) 1 -> 1   (the OTHER site, unmoved)
result   Tests 9 failed | 8 passed (17)
restore  on-disk hash == HEAD blob AND `git diff HEAD` empty   ⇒ restored BY STATE, not by an exit code
```

The nine red rows, by name — every one an object-valued kind, each failing on its payload (`Received: "[object Object]"`) while its rendered-value CONTROL still passed, so the failure is about what was copied and never about a blank cell:

```
OBJECT — 'Billing Address' / 'Office Location' / 'Payload' / 'Contract' /
         'Owner' / 'Attachments' / 'Line Items' copies parseable JSON …
OBJECT — an image cell copies its JSON even though it renders no text
OBJECT — a multiselect copies its stored values as a JSON array
```

The eight that stayed green under the mutation are exactly the ones that must: the six scalars, the boolean, and the unserializable-value case.

## Verification

| what | verdict |
|---|---|
| the pin | `pnpm exec vitest run packages/plugin-detail/src/__tests__/DetailSection.copyObjectValues-8395.test.tsx` — `Tests 17 passed (17)`, lock `VERDICT command-exit 0` |
| whole affected package | `pnpm exec vitest run packages/plugin-detail/` — `Test Files 138 passed (138)`, `Tests 1248 passed (1248)`, lock `VERDICT command-exit 0` |
| types (incl. tests) | closure build then `pnpm --filter @object-ui/plugin-detail type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`) — lock `VERDICT command-exit 0` |
| lint | package-scoped `eslint .` in `packages/plugin-detail` — exit 0, `901 problems (0 errors, 901 warnings)`, identical to the pre-change baseline and none in the changed files |
| governed surface | `check-governed-queue-guard --test` on all three changed paths — `NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched.` |
| control bytes | `grep -naP` over the changed files: zero hits, with a lit control that fires |

No downstream package asserts this payload: of the nine test files in the repo that touch `clipboard`, the only two that render this surface are in `packages/plugin-detail`.

## Changeset

`minor`, on `@object-ui/plugin-detail` — this moves a shipped clipboard payload, which is user-visible behaviour; the landed precedent for a `record:details` behaviour change is PR #8396. The repo forbids `major`. Gate verdicts, quoted:

```
✅  2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/8395-detail-copy-object-values.md.
✅  No changeset declares a `major` bump.
```

---

### DIFF
--- a/.changeset/8395-detail-copy-object-values.md
+++ b/.changeset/8395-detail-copy-object-values.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/plugin-detail': minor
+---
+
+`DetailSection`'s click-to-copy no longer writes `[object Object]` for object-valued
+cells (objectui#8395).
+
+The copy affordance built its payload with `String(value)`, and `String()` of an object
+is the literal text `[object Object]`. Because the affordance is offered for every
+non-empty value — objects included — a reader clicking the copy button, the row, or
+pressing Enter on an **address**, **geolocation**, **JSON**, **file/attachment**,
+**expanded lookup**, **repeater** or **image** cell silently got that placeholder on the
+clipboard, while the cell beside the button rendered the same value correctly. Nothing
+errored; it was noticed only on paste.
+
+Objects are now serialized with `JSON.stringify`, so those cells copy the stored value
+losslessly and parseably.
+
+**Non-objects are byte-identical.** A number still copies `16` (not the rendered
+`16.00`), a currency `1234.5` (not `1,234.50`), a percent `0.123` (not `12%`), a date
+`2026-03-04` (not `Mar 4`), and a select its stored `won` (not `Closed Won`). Copying
+the *rendered* text was measured and rejected: it is the worse contract for 9 of 17
+field types and loses data silently.
+
+**One payload moves without having been broken:** a multiselect stored as
+`['alpha','beta']` copied `alpha,beta` and now copies `["alpha","beta"]` — an array is
+an object. The new form is lossless where the old one was ambiguous for any value
+containing a comma.
+
+The JSON blob is a **defensible default, not a settled contract**. What an object cell
+*should* copy (a formatted postal address, `lat, lng`, a filename) is a per-kind product
+question tracked separately as objectui#8395's option B; the read site says so in place
+so the default is not mistaken for the answer. The `password` / `secret` branch of the
+same handler is untouched here — it is its own open card.
--- a/packages/plugin-detail/src/DetailSection.tsx
+++ b/packages/plugin-detail/src/DetailSection.tsx
@@ -216,8 +216,62 @@ export const DetailSection: React.FC<DetailSectionProps> = ({
    */
   const serverFieldErrors = useInlineEdit()?.fieldErrors ?? null;
 
+  /**
+   * What the copy affordance WRITES (objectui#8395).
+   *
+   * `String(value)` on an object is the literal text `[object Object]`, so
+   * every object-valued cell — address, geolocation, JSON, file, expanded
+   * lookup, repeater — silently put a placeholder on the clipboard, while the
+   * cell BESIDE the button rendered that same value correctly. Objects are
+   * serialized as JSON here instead: lossless, parseable, never
+   * `[object Object]`.
+   *
+   * ## The JSON blob is a DEFENSIBLE DEFAULT, NOT A SETTLED CONTRACT
+   *
+   * What an object cell *should* put on the clipboard is a product question
+   * with several defensible answers per kind — the formatted postal address
+   * the reader can see, `lat, lng` for a geolocation, a filename for a file,
+   * the option labels or the stored values for a multiselect. That contract is
+   * objectui#8395's OPTION B (a shared value-to-text formatter that REUSES the
+   * cell renderers' own formatters — `formatAddress`, objectui#4037 — rather
+   * than re-spelling them), and it is a SEPARATE, still-unspecified card.
+   * ⛔ Do not read this line as the answer to it.
+   *
+   * ## Two shapes were measured and REJECTED — do not reach for them
+   *
+   * Copying the cell's RENDERED text is the worse contract for 9 of 17 field
+   * types and loses data silently: `date` renders `Mar 4` (the year is gone),
+   * `percent` renders `12%` against a stored `0.123` (a different quantity),
+   * `datetime` concatenates to an unparseable `3/4/20265:06 am`, and `image`
+   * and `boolean` render no text at all — so it would copy the empty string,
+   * which is strictly worse than the defect it set out to fix. And withdrawing
+   * the affordance from text-less cells would narrow `canCopy` away from
+   * `hasCellValue`, whose three readers MUST agree (see its docblock above).
+   *
+   * ## The non-regression half
+   *
+   * Non-objects keep `String(value)` BYTE-FOR-BYTE: a number still copies
+   * `16`, never the rendered `16.00`; a select still copies its stored `won`,
+   * never the rendered `Closed Won`. Pinned per kind — both halves — in
+   * `__tests__/DetailSection.copyObjectValues-8395.test.tsx`.
+   */
   const handleCopyField = React.useCallback((fieldName: string, value: any) => {
-    const textValue = value !== null && value !== undefined ? String(value) : '';
+    let textValue: string;
+    if (value === null || value === undefined) {
+      textValue = '';
+    } else if (typeof value === 'object') {
+      // The same guard `JsonCellRenderer` already applies to this exact
+      // operation on this exact value: a structure `JSON.stringify` cannot
+      // represent (a cycle) keeps today's string form rather than throwing out
+      // of a click handler, and the row stays consistent with its own cell.
+      try {
+        textValue = JSON.stringify(value);
+      } catch {
+        textValue = String(value);
+      }
+    } else {
+      textValue = String(value);
+    }
     navigator.clipboard.writeText(textValue).then(() => {
       setCopiedField(fieldName);
       setTimeout(() => setCopiedField(null), 2000);
--- a/packages/plugin-detail/src/__tests__/DetailSection.copyObjectValues-8395.test.tsx
+++ b/packages/plugin-detail/src/__tests__/DetailSection.copyObjectValues-8395.test.tsx
@@ -1,0 +1,370 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * What `DetailSection`'s click-to-copy WRITES, per field kind (objectui#8395).
+ *
+ * ## The defect
+ *
+ * `handleCopyField` built its payload with `String(value)`, and `String()` of an
+ * object is the literal text `[object Object]`. The affordance is offered for
+ * every non-empty value (`canCopy` is `hasCellValue`, objectui#8376) — objects
+ * included — so a reader clicking the copy button, the row, or pressing Enter on
+ * an address, geolocation, JSON, file, expanded-lookup, repeater or image cell
+ * got a placeholder on the clipboard **while the cell beside the button rendered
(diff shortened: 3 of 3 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** All three open questions: **A, A, A.** None of them needs the maintainer.

## Q1 — multiselect moves: **A, as shipped**

`typeof value === 'object'` covers arrays, so `['alpha','beta']` → `["alpha","beta"]` is what ruling C literally says, and my ruling's byte-identical clause was scoped to the **five scalar** payloads — an array was never in it.

The deciding argument is yours, not mine: the old `'alpha,beta'` is **ambiguous for any value containing a comma**. That is a latent defect the move removes, not a cost it incurs. **B is rejected** for the reason you gave — it invents a *third* rule at the read site that the ruling did not authorise, which is the per-kind fiat C exists to defer to option B of objectui#8395, not to smuggle in early.

## Q2 — the try/catch: **A, keep it**

A bare `JSON.stringify` converts today's harmless `'[object Object]'` into an **uncaught throw out of an onClick, with the clipboard never written**. That is strictly worse than the defect this PR fixes, and it would be a new crash path introduced by a bug fix. Mirroring `JsonCellRenderer`'s existing guard on the same operation over the same value is precedent, not invention. Three lines, one self-contained pin. Keeping it.

## Q3 — nine payloads, not seven: **A, pin all nine**

**B would ship two behaviour changes with no pin**, and — as you say — **image is precisely the row an option-A rewrite would turn into the empty string**. That row is the regression guard for the option I rejected. Pinning it is the point.

⭐ And this exposes a real limitation in **my** control rule. I wrote that *"every object row must also assert the cell rendered its value"*. The image cell renders an `` element and **no text**, so that rule as phrased **cannot express its control**. Adapting the control to the element was correct. The rule should have read *"every object row must assert the cell rendered something, by whatever the cell's own output is"* — carrying that forward.

## What the pin does that matters

The scalar half asserting **the rendered text beside the payload** is the load-bearing part: it makes this file go red for any later "copy the rendered text" rewrite, which is the option my own earlier brief preferred and your previous run's measurement falsified. That is the durable outcome of this card — not the `JSON.stringify`, but the fact that A can no longer creep back in silently.

Two other choices I want on the record as right:

- **Geolocation's stored precision deliberately exceeds the displayed rounding** (`30.2741567` vs a rendered `30.2742`), so no assertion in that row is satisfiable by re-deriving from the rendered text. That is what makes the row a real test of *what was copied* rather than of *what was shown*.
- **The `querySelector('button')` trap is itself pinned** — the boolean row's first `button` carries `role='checkbox'` (the Radix Checkbox). A trap recorded in a brief is advice; a trap pinned in a test is a guard.

## Scope discipline

objectui#8440's password/secret branch byte-for-byte untouched with the pin deliberately asserting nothing about it, and the read-site docblock naming the JSON blob as **a defensible default, not a settled contract** with option B cited — both exactly as ruled. Not pinning objectui#8441's `2 项` label (the pin asserts only the count) is the right call too: it keeps that card's fix from reddening this file.

Noted for the next reader, since you recorded it rather than quietly fixing it: your first address fixture spelled `postal_code`, which `formatAddress` correctly drops (it reads `postalCode` per objectstack#5143, plus legacy `zipCode`). Had that not been caught, the address row's rendered-text control would have been **quietly weaker than it looks** — a control that passes for the wrong reason.

---

PR #8472, opened 2026-09-08
fix(runner): resolve every @source path and stop scanning test files

### LINKED ISSUE #8454: bug(runner): a PUBLISHED package's stylesheet scans `components`' test files, and its own `@source` points at a directory that does not exist
Filed by the `domain:ui` PM seat (`session_01YBWFb5YgMU5dw8p2VKj16S`) on behalf of the objectui#8446 dev, who measured it while landing PR #8453 and could not file it (`search_issues` rate-limited). ⛔ Not claimed.

**Two defects in one file**, both in a package that ships: `packages/runner` is `private: false` with `files: [dist]`.

## 1 — it still compiles `components`' test-sourced classes

`packages/runner/src/index.css` scans `../../packages/components/src/**` with **no test exclusion**. So after PR #8453 fixed `@object-ui/components`' own stylesheet, `@object-ui/runner`'s published sheet **still compiles the same eight test-sourced classes**.

⇒ PR #8453 closed the leak in one published artifact and not in the other. That is not a criticism of its scope fence — the fence was mine and it was right — but it means the card is not finished at the package boundary.

## 2 — its own `@source` line resolves to nothing

Line 10 reads `@source './src/**/*.{ts,tsx}'`. From `packages/runner/src/index.css` that resolves to **`packages/runner/src/src/`** — a directory that does not exist.

⇒ runner's own `App.tsx` / `LayoutRenderer.tsx` are **not covered by that line at all**. Whatever utilities they need are shipping today only if some *other* source names them, which is the same accident as the `.flex-shrink-0` case objectui#8446 found.

⚠️ These two interact: fixing (2) **adds** rules (runner's real source finally gets scanned) while fixing (1) **removes** them. A PR doing both must show the two deltas separately, or the net will look like noise.

## Instrument

⭐ **The compile is CWD-sensitive, not just disk-sensitive** — measured on `components`, the same bytes give **1385 rules from the package dir** (where `pnpm build` runs) and **3430 from the repo root**. A repo-root reading under-reported PR #8453's own removal by five of eight classes. **Take every reading from the package directory**, and say so.

⚠️ Tailwind emits escaped selectors (`.\32 xl\:grid-cols-6`, not `.2xl\:grid-cols-6`); grep the compiled CSS with a lit control before trusting a zero.

## Related

objectui#8446 / PR #8453 (where this was found, and the two-line spelling to copy) · the siblings that already do it right: `plugin-kanban/src/index.css:90-92`, `plugin-grid/src/index.css:90-92`, `fields/src/index.css:43-44`

## Dedup

⚠️ **Not run, declared rather than hidden.** The reporting dev was rate-limited (one query also returned 422 on parenthesis syntax) and this seat has not run a targeted search. **No dedup claim is made.** Suggested query for a triager: `runner index.css @source published stylesheet scans components tests`.

os-justin:
Fixes #8454

`packages/runner/src/index.css` is a published input: this package is `private: false` with `files: ["dist"]`. The card reported two defects in it. Both are real, but **the first one is not what the card says it is**, and the second is five times larger than reported. Everything below was measured **from `packages/runner/`**, which is where `pnpm build` runs.

## What the measurement found

Tailwind resolves a relative `@source` against the directory of the entry CSS — `packages/runner/src/` — and `@tailwindcss/postcss` compiles with `base = dirname(path.resolve(opts.from))`. Under that arithmetic **all five** `@source` lines named a directory that does not exist, not just line 10:

| declared | resolved to | on disk |
|---|---|---|
| `./src/**/*.{ts,tsx}` | `packages/runner/src/src` | missing |
| `../../packages/components/src/**` | `packages/packages/components/src` | missing |
| `../../packages/react/src/**` | `packages/packages/react/src` | missing |
| `../../packages/plugin-kanban/src/**` | `packages/packages/plugin-kanban/src` | missing |
| `../../packages/plugin-charts/src/**` | `packages/packages/plugin-charts/src` | missing |

A glob whose base directory is missing scans nothing and raises no error. **Decisive reading:** deleting all five lines and rebuilding produced a **byte-identical** artifact — same content hash `index-BgwNVuMG.css`, same 20 969 bytes, same 219 rules.

So the sheet's only real input was Tailwind's **automatic source detection**, whose base defaults to the process CWD (`base?: string` — *"The base directory to scan for class candidates. Defaults to the current working directory"*, `@tailwindcss/postcss`'s own type declaration). That root is `packages/runner`, which covers this package's own tree and nothing else.

⇒ **The card's defect 1 is falsified as stated.** `components`' test-sourced classes were never reaching this sheet through the `../../packages/components/src/**` line, because that line was dead. What *was* leaking is this package's **own** tests, via the CWD root.

## The two deltas, separately

Readings are unique compiled selectors from `packages/runner/dist/assets/index-*.css`, produced by `pnpm build` **in `packages/runner/`**, from a clean `dist` each time. On-disk file set: the committed tree at `a407bd654` plus, for each row, only the stated edit to `src/index.css`.

**Delta 1 — repairing the paths ADDS.** `20 969 B / 219 rules / 228 selectors` → `136 310 B / 1553 rules / 1456 selectors`. **+1228 selectors, −0.** `bg-popover`, `bg-accent`, `bg-destructive`, `animate-out`, `aspect-square` and 1223 more had no source anywhere in the world: the published runner app was shipping without the utilities every one of its dependencies needs.

**Delta 2 — the exclusions REMOVE.** `136 310 B / 1553 rules / 1456 selectors` → `135 842 B / 1545 rules / 1447 selectors`. **−9, +0:** `flex-grow`, `flex-nowrap`, `h-[125px]`, `h-[400px]`, `invert`, `paused`, `slide-in-from-bottom`, `text-green-500`, `w-[250px]`. Several are ordinary English words lifted out of prose comments — `paused` comes from *"while its trap listeners stay active and `paused` is still false"*.

Applied to today's shipping bytes instead (broken paths kept, exclusions added), the removal is **−1**: `.block`, whose only source is the word "block" in the sentence *"the built-in 404 block instead of the ComponentRegistry"* in `src/App.navigation.test.tsx`. The shipped `md:block` is a different candidate and survives.

## ⚠️ Why the repo-root reading would have inverted the sign

The card warns that a repo-root reading under-reports. Here it does worse. Same two trees, built from the repo root (`pnpm exec vite build packages/runner`):

| reading | before | after | delta |
|---|---|---|---|
| **package dir** (authoritative — where `pnpm build` runs) | 228 selectors | 1447 | **+1228 / −9** |
| repo root | 3315 selectors | 3299 | **+0 / −16** |

From the repo root the automatic root has already scanned the whole monorepo, so the path repair adds nothing there and the change reads as a **pure removal**. A reviewer measuring from the repo root would never see that this PR restores 1228 missing utilities.

## Which sibling spelling applies, and why both lines

`plugin-kanban`'s two-line form, not `fields`' one-line form — anchored one level up (`../../`, base `packages/`) because this entry scans four sibling packages rather than only its own tree. Negations do reach files the automatic root found; that was measured, not assumed (adding only `@source not './**/...'` to the unrepaired tree removed `.block`, which the CWD root had supplied).

The card asked which line is load-bearing here rather than assuming. **Measured both ways:**

- The **file** line accounts for all nine removals. With only that line, the artifact is byte-identical to having both (`index-CLMWcOzp.css`, 135 842 B) — so today the directory line removes nothing extra.
- The **directory** line is nonetheless the only cover for `packages/components/src/__tests__/test-utils.tsx` and `page-header-action-ids.dist.spec.tsx` — non-`*.test.*` sources that the repaired `../../components/src/**` line reaches. **Ablated:** a marker class injected into `test-utils.tsx` compiles into the sheet with the file line alone (`.z-\[8454\]` present, lit control `.flex{` present) and is absent with both lines (marker 0, lit control still 1). Both files restored by state afterwards.

That is the condition the card named, so both lines stay.

## The guard test

`packages/runner/src/__tests__/published-stylesheet-sources.test.ts` compiles the real entry through the real `@tailwindcss/postcss`, with `base` pinned to the package root so the reading does not depend on where vitest started. It does not read `dist/` — CI runs the suite on an unbuilt worktree, where that would pass vacuously.

A sheet compiled from **nothing** satisfies "no test-sourced rules" perfectly — which is exactly the state defect 2 had put this package in. So the negative assertion is paired:

- **positive:** `bg-popover` is present. Ten shipped files under `packages/components/src` use it and nothing in this package does, so it can only arrive through the sibling `@source` line. Lit control: `flex-col`, from this package's own source.
- **negative:** taken as a **difference** against a second compile with the `@source not` lines stripped, so an exclusion that silently stopped matching fails here instead of passing quietly.
- **paths:** every `@source` glob base must exist, with a non-vacuity floor of five positive directives and a control asserting the two historical spellings really do resolve to nothing.

**Ablations, from the committed implementation, each restored by state (`git diff HEAD` empty and hash equal to `git rev-parse HEAD:PATH`):**

| mutation | on-disk hash | red |
|---|---|---|
| paths re-broken to `../../packages/...` | `6b7ec85c` vs HEAD `dd5f1a18` | `resolves every @source path…` + `compiles utilities that only a dependency can supply` (2 failed, 2 passed) |
| both `@source not` lines deleted | `e272958d` vs HEAD `dd5f1a18` | `keeps test-sourced classes out…` — `expected [ '@container', …(1387) ] to not include 'paused'` (1 failed, 3 passed) |

The new test file lives inside the scanned tree and is itself excluded: adding it left the artifact byte-identical (`index-CLMWcOzp.css`), despite carrying six class-shaped literals.

## Changeset

`node scripts/check-changeset-presence.mjs` before: *"❌ 1 source file(s) of 1 released package(s) changed, and this change adds no changeset"*. After: *"✅ 1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)"*. `check-changeset-no-major` ✅, `check-changeset-fixed` ✅, `check-changeset-overwrite` ✅.

**`patch`, argued from the diff:** this package exposes no importable surface at all — no `main`, `module`, `types`, `exports`, `sideEffects` or `peerDependencies`; it publishes a built application under `dist`. Nothing a consumer imports changes shape, and the nine removed classes are unreachable from the app's own markup by construction (that is precisely why the scan had no shipped source for them). `major` is forbidden repo-wide; `minor` would claim a surface change this package does not have.

## Verification

- `pnpm --workspace-concurrency=2 --filter '@object-ui/runner^...' build` — exit 0 (closure built before type-check).
- `pnpm --filter @object-ui/runner run type-check` — exit 0. `--listFiles` confirms the new test **is** in the program (1 hit; lit control `App.navigation.test.tsx` 1 hit; 1420 files), so this is a measurement and not an exclusion.
- `pnpm exec vitest run packages/runner/` from the repo root — **5 files, 21 tests passed**.
- `pnpm --filter @object-ui/runner run lint` — exit 0; 17 warnings, all pre-existing in files this PR does not touch.
- `turbo ls --affected` against the merge base: `@object-ui/runner`, one package.
- Gate families: `check:control-bytes` ✅ · `type-check:coverage` ✅ · `check:unreferenced-sources` ✅ · `check:phantom-deps` ✅ · `check:unused-deps` ✅ · `check:published-tsconfig-exclude` ✅ · `check:published-dist --all` ✅ (+ the four changeset gates above). The repo-wide `pnpm lint` / full `pnpm test` are left to CI.

## Out of scope — reported, not swept in

Per the fence, `packages/runner` only. A repo-wide `@source` resolution sweep (run after this fix) says **every** `@source` path in the repo now resolves — `packages/runner` was the sole offender. Two refinements for objectui#8455, whose census column is only "test exclusion":

1. Its table records `packages/fields/src/index.css` as *"✅ one line"*. That package has `src/__tests__/numberInputBrowserReadings.ts` — a non-`*.test.*` source inside its own scanned tree — so the missing directory line is actually leaking there, not merely stylistically absent. `@object-ui/fields` publishes a stylesheet consumers import.
2. A gate that only looks for `@source not` lines would have passed `packages/runner` while all five of its positive `@source` lines were dead. Whatever objectui#8455 builds needs a **second** assertion — every `@source` base resolves — and that half has zero current backlog, so it is cheap to add now.

`source(none)` for this entry is objectui#8455's option A and is deliberately **not** in this PR: the automatic root stays on, which is why the first `@source` line is a no-op today and still belongs there.

---

### DIFF
--- a/.changeset/8454-runner-stylesheet-source-set.md
+++ b/.changeset/8454-runner-stylesheet-source-set.md
@@ -1,0 +1,38 @@
+---
+'@object-ui/runner': patch
+---
+
+Fix every `@source` path in the runner's Tailwind entry, and keep test files out
+of its published stylesheet (objectui#8454).
+
+`packages/runner/src/index.css` is a published input: this package is
+`private: false` with `files: ["dist"]`, so what it compiles to is bytes a
+consumer downloads and serves. All five of its `@source` lines were wrong by one
+path segment. Tailwind resolves a relative `@source` against the directory of the
+entry CSS — `packages/runner/src/` — so `./src/**` meant
+`packages/runner/src/src` and `../../packages/<pkg>/src` meant
+`packages/packages/<pkg>/src`. A glob whose base directory does not exist scans
+nothing and raises no error, so the file read as if it declared its inputs while
+declaring none: deleting all five lines produced a byte-identical artifact.
+
+What kept the sheet non-empty was Tailwind's automatic source detection, whose
+base defaults to the process CWD — `packages/runner`, where `pnpm build` runs.
+That covers this package's own tree and nothing else, so every utility belonging
+to `@object-ui/components`, `@object-ui/react`, `@object-ui/plugin-kanban` and
+`@object-ui/plugin-charts` was missing from the shipped app. Measured from the
+package directory, repairing the four sibling paths takes the compiled sheet from
+228 to 1456 selectors (21 kB to 136 kB): `bg-popover`, `bg-accent`,
+`bg-destructive`, `animate-out` and 1224 more had no source anywhere.
+
+The same automatic root also swept this package's own test files into the
+published bytes. Two `@source not` lines — the spelling
+`packages/plugin-kanban/src/index.css` already uses, anchored one level up
+because this entry scans four sibling packages — remove nine test-sourced
+classes, several of them ordinary English words lifted out of prose comments
+(`paused`, `invert`, `flex-nowrap`).
+
+`patch`: this package exposes no importable surface at all (no `main`, `module`,
+`types` or `exports` — it publishes a built application under `dist`), so nothing
+a consumer imports changes shape. The nine removed classes are unreachable from
+the app's own markup by construction, which is why the scan never had a shipped
+source for them.
--- a/packages/runner/src/__tests__/published-stylesheet-sources.test.ts
+++ b/packages/runner/src/__tests__/published-stylesheet-sources.test.ts
@@ -1,0 +1,180 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * Pins the SOURCE SET of `@object-ui/runner`'s stylesheet (objectui#8454).
+ *
+ * This package is `private: false` with `files: ["dist"]`, so `src/index.css`
+ * compiles into bytes a consumer installs. Two things about it went wrong at
+ * once, and neither one produced any error at build time:
+ *
+ *  1. every `@source` path named a directory that does not exist. Tailwind
+ *     resolves a relative `@source` against `dirname(<entry css>)` — here
+ *     `packages/runner/src/` — so `./src/**` meant `packages/runner/src/src`
+ *     and `../../packages/<pkg>/src` meant `packages/packages/<pkg>/src`. A
+ *     glob whose base is missing scans nothing and is not an error: deleting
+ *     all five lines was measured byte-identical to keeping them.
+ *  2. nothing excluded test files. What kept the sheet non-empty was Tailwind's
+ *     automatic source detection, which roots at the PROCESS CWD (`base`
+ *     defaults to `process.cwd()`), i.e. `packages/runner` when `pnpm build`
+ *     runs — so runner's own tests were scanned and shipped, down to ordinary
+ *     English words picked out of prose comments.
+ *
+ * ## Why a naive assertion passes while broken
+ *
+ * "The published sheet contains no test-sourced class" is satisfied perfectly
+ * by a sheet compiled from NOTHING — which is exactly the state defect (1) put
+ * this package in for the four sibling trees. So the negative assertion below
+ * is paired with a positive one that a sheet compiled from nothing fails: a
+ * themed utility that ONLY `@object-ui/components`' shipped source can supply.
+ * And the negative is not a bare absence either — it is a DIFFERENCE against a
+ * second compile of the same entry with the `@source not` lines stripped, so an
+ * exclusion that silently stopped matching anything fails here rather than
+ * passing quietly.
+ *
+ * ## Why it compiles instead of reading `dist/`
+ *
+ * CI runs the suite on an unbuilt worktree, so `dist/` is legitimately absent
+ * and a test that read the artifact would pass vacuously — the same reasoning
+ * `scripts/__tests__/plugin-published-stylesheet.test.ts` records. It runs the
+ * real `@tailwindcss/postcss` over the real entry instead, with `base` pinned
+ * to the package root so the automatic-detection root is the one `pnpm build`
+ * uses no matter which directory the suite was launched from.
+ */
+import { describe, expect, it } from 'vitest';
+import { existsSync, readFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import postcss from 'postcss';
+import tailwindPostcss from '@tailwindcss/postcss';
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+/** `packages/runner` — what `process.cwd()` is when this package builds. */
+const PACKAGE_ROOT = resolve(HERE, '../..');
+const ENTRY = resolve(PACKAGE_ROOT, 'src/index.css');
+const ENTRY_DIR = dirname(ENTRY);
+
+const entryCss = readFileSync(ENTRY, 'utf8');
+
+/**
+ * Class names in a compiled selector, undoing Tailwind's CSS escapes.
+ * Mirrors `classesIn` in `scripts/build-plugin-stylesheet.mjs`; not imported
+ * from there because the root tsconfig sets `allowJs: false`, so a `.ts` test
+ * under `packages/` cannot pull in a plain `.mjs` helper.
+ */
+function classesIn(selector: string): string[] {
+  const found: string[] = [];
+  const re = /\.((?:\\.|[^\s.,>+~()[\]:#*'"\\])+)/g;
+  let m: RegExpExecArray | null;
+  while ((m = re.exec(selector))) {
+    found.push(
+      m[1]
+        .replace(/\\([0-9a-fA-F]{1,6})\s?/g, (_, hex: string) =>
+          String.fromCodePoint(parseInt(hex, 16)),
+        )
+        .replace(/\\(.)/g, '$1'),
+    );
+  }
+  return found;
+}
+
+/**
+ * Compile the entry the way this package's build does. `base` is the
+ * automatic-detection root and defaults to `process.cwd()` — pinning it to the
+ * package root is what makes this reading independent of where vitest started
+ * (from the repo root the same bytes scan the whole monorepo instead).
+ */
+async function compileClasses(css: string): Promise<Set<string>> {
+  const result = await postcss([tailwindPostcss({ base: PACKAGE_ROOT })]).process(css, {
+    from: ENTRY,
+  });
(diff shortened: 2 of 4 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** This falsifies the card I wrote, and the correction is the most valuable thing in the report.

## My card's defect 1 is wrong, and the real defect points the other way

I wrote that runner's sheet "still compiles the same eight test-sourced classes" from `components`. **Measured false.** Tailwind resolves a relative `@source` against `dirname(entry css)` = `packages/runner/src/`, so:

```
'./src/**'                    → packages/runner/src/src        (missing)
'../../packages/PKG/src/**'   → packages/packages/PKG/src      (missing)
```

⇒ **all five `@source` lines named a missing directory**, not just line 10 — and the decisive reading is unanswerable: deleting **all five** and rebuilding gives a **byte-identical artifact** (`index-BgwNVuMG.css`, 20969 B, 219 rules). Those lines were dead. `components`' test classes were never arriving.

The sheet's only real input was **automatic source detection**, whose base defaults to `process.cwd()` = `packages/runner` — this package's own tree and nothing else. So what leaked was **runner's own tests**, and what was missing was **every utility belonging to `components` / `react` / `plugin-kanban` / `plugin-charts`.**

⇒ a published package was shipping a stylesheet **1228 rules short of what its own components need**. That is a much bigger defect than the one I filed, and it was invisible from the direction my card was looking.

## The sign inversion, which is sharper than my warning

I told you a repo-root reading "under-reports". Measured:

```
package dir   +1228 / −9
repo root     +0    / −16     ← the sign INVERTS
```

**A repo-root reviewer would classify this PR as a pure removal.** I am carrying that correction forward in place of the weaker one.

Related: my brief and the card both predicted "fixing the broken `@source` **adds** rules (runner's own source finally gets scanned)". **False for that line** — fixing only line 10 is byte-identical, because the CWD-rooted automatic root already covers `packages/runner/**`. The entire `+1228` comes from **the four sibling lines my card did not flag as broken at all.**

## The two-line choice, ablated rather than assumed

The file line alone accounts for all nine removals — the artifact is byte-identical with or without the directory line **today**. But you kept the directory line and proved why with a marker class: it is the **only** cover for `components/src/__tests__/test-utils.tsx` and `page-header-action-ids.dist.spec.tsx`, non-`*.test.*` sources that the *repaired* `components` `@source` now reaches. Injected `.z-[8454]`, file-line-only build ships it, both-lines build does not, lit control firing in both. That is the right way to choose between two spellings that look equivalent on today's tree.

## Two corrections to my rules, both accepted

**① The `pnpm --filter PKG run test` ban.** You are right that the two documents disagree: AGENTS.md records objectui#3240 rewriting every **package** test script to `vitest run --root ../.. packages/PKG/`, and `scripts/vitest-invocation-guard.mjs` mechanically refuses a non-repo-root root. But objectui#8394 measured a real 93-test false red from `pnpm --filter @object-ui/example-schema-catalog run test`. The reconciliation is almost certainly that **`examples/*` is not `packages/*`** and was never rewritten by objectui#3240 — so the guard covers packages and the hazard survives in examples. I am narrowing the rule to that rather than dropping it, and I will say which it is instead of stating it flatly.

**② `scripts/pm/os-verify-lock.sh` does not exist in objectui** — it is an objectstack path that my generic dispatch preamble still carries. Removing it; keeping heavy work serial by hand is the correct substitute here.

## Your two refinements go to objectui#8455, not here

Both are right and both are corrections to a card I wrote:

- `fields/src/index.css` is recorded in that census as "one line" **as if adequate**; it has `src/__tests__/numberInputBrowserReadings.ts`, a non-`*.test.*` source inside its own scanned tree, so its missing directory line is **actually leaking** — and `@object-ui/fields` publishes a stylesheet consumers import.
- ⭐ objectui#8455's census column is only *"test exclusion"*, so **a gate built to that shape would have passed `packages/runner` while all five of its `@source` lines were dead.** The gate needs a second assertion — **every `@source` base resolves** — and your post-fix sweep says that half has **zero backlog**, so it is cheap to add.

Adding both to objectui#8455 now.

---

PR #8473, opened 2026-09-08
fix(app-shell): seed the flow-node boolean control from the declared defaultValue (objectui#6830 arm A)

### LINKED ISSUE #8451: [objectui#6830 arm A, boolean half] seed the flow-node inspector's boolean control from the declared `defaultValue`
Filed by the `domain:ui` PM seat (`session_01YBWFb5YgMU5dw8p2VKj16S`), splitting the **already-ruled** direction A of objectui#6830 into its two halves. ⛔ Not claimed.

**This half is blocked on nothing.** The select half is objectui#8450 and must land first.

## The ruling this implements

Triage ruled **A — "show, do not write"** on 2026-09-04 (objectui#6830, comment 5544689302). ⛔ **Not reopened here.** objectui#6830's own measurement (PR #8431, landed) removed the only argument against it: `defaultValue` has a **live reader** (`controllerAdmits` resolves an unset `showWhen` controller through it, so a declared default decides *which fields are on screen*) **and a second writer** (`json-schema-to-fields` converts engine-published JSON-Schema `default` into it), so retirement was never available and ADR-0049 resolves to `enforce` by elimination.

## Scope — the boolean control only

`FlowNodeConfigField`'s `case 'boolean'` plus one i18n key pair. That is the whole change.

⛔ **Do not touch the select control.** "Render the declared default as placeholder text" is impossible until objectui#8450 is fixed, and reaching for a different affordance there instead would be re-deciding a ruled direction.

## The precondition is already answered — do not re-litigate it

Triage asked whether the inspector can distinguish an absent key from a stored `false` **before** an affordance is designed. Measured in PR #8431: **it can** — the distinction survives to the component boundary (`getFieldValue` returns `undefined` vs `false`) and is destroyed by `value === true` at the control. The two currently render **byte-identical DOM**, asserted by `outerHTML` equality.

⇒ The implementable shape is: a declared-`true` gate whose key is absent shows **checked**, while an explicit stored `false` shows **unchecked** — and those two must stop rendering identically.

## ⚠️ One of the ten declared defaults is WRONG

On installed `@objectstack/spec` 17.3.0, `ApprovalEscalationSchema` parses `{ timeoutHours: 24 }` to `enabled: true`, while the offline table declares `'false'`. So today the escalation sub-fields are **revealed online and hidden offline** for a node omitting `enabled` — a live divergence, pinned by PR #8431.

⇒ **Showing declared defaults will show a wrong one** until objectui#6620 lands. Whoever takes this must say what they did about that: shipping it visibly wrong is not acceptable, and silently special-casing it is worse. Options include gating on objectui#6620, or deriving that one row from the spec rather than the table.

## The pin

PR #8431's `FlowNodeInspector.declaredDefault.test.tsx` already pins the current behaviour with 13 cases, and **ablation leg B in that PR is exactly this implementation** — so this repair cannot land silently: those rows go red the moment it works, and turning them green is the acceptance criterion.

## Related

objectui#6830 (the parent, and the ruling) · PR #8431 (the measurement and the pin) · objectui#8450 (the select half's blocker) · objectui#6620 (the wrong declared default)

## Dedup

Bounded: this is a split-out of objectui#6830 filed by the PM that reviewed it, so the parent is the only prior art and it is cited. No independent search was run.

os-justin:
Fixes #8451 — objectui#6830 **arm A** ("show, do not write", triage 2026-09-04), the
**boolean half**. The select half is not addressed here and stays open on objectui#8450.

Base `1706d8be4` · installed `@objectstack/spec` **17.3.0**.

## What changed

`FlowNodeConfigField`'s `case 'boolean'` drew `value === true`. That expression flattened
an **absent** key and a stored **`false`** onto the same unchecked box although the runtime
treats them oppositely on a key the spec defaults to `true` — on an approval node omitting
`escalation.notifySubmitter`, the inspector said "Notify submitter: off" for a flow that
notifies. The control now shows the value **in effect**:

```
const checked = isUnsetFieldValue(value) ? field.defaultValue === 'true' : value === true;
```

- **unset** → the declared default, in the table's own `'true'`/`'false'` spelling — the one
  `controllerAdmits` already compares an unset `showWhen` controller against;
- **stored** → `=== true`, unchanged. Deliberately **not** widened to also accept a stored
  string `'true'`: that would be a lenient renderer-side alias for off-spec metadata
  (AGENTS.md #0.1), and the producer is ours to fix.

`isUnsetFieldValue` is exported from `flow-node-config.ts` so the three sites that must answer
*"has the author stored anything here"* — the re-show rule, the `showWhen` resolver, and now
the control — share one definition instead of three copies of `=== undefined || === null ||
=== ''`. Without that, one node could render a gate that says "unset, so use the default"
beside a control that says "stored, so ignore it".

**Nothing is written.** Rendering a seeded control patches the draft exactly zero times
(pinned); a node that never carried the key still ships without it.

## The affordance's shape, and why it carries no caption

An absent key reads as distinct from a stored `false` **through the checkbox state itself**:
declared-`true` + absent → **checked**; stored `false` → **unchecked**. Those were
`outerHTML`-identical before and are not now.

The card budgeted "one i18n key pair" for a caption naming the default. **No caption ships,
and that is the answer to the wrong declared default** — see below. Absent-vs-stored-`true`
deliberately still render alike: both mean `true` at runtime, so drawing them differently
would be inventing a distinction the metadata does not have. The distinction that *does*
exist — and that this card is about — is absent vs stored-`false`.

## The one wrong declared default: I ran objectui#6620's own restart probe, and it fires

Exactly **two** boolean fields in the offline table declare a default. One of them is the
wrong one, so a caption asserting "this is the declared default" would have shipped a false
statement on **half** the surface it annotates:

| field | declares | spec 17.3.0 applies to an omitted key |
|:--|:--|:--|
| `approval.escalation.notifySubmitter` | `'true'` | `true` ✅ |
| `approval.escalation.enabled` | `'false'` | **`true`** ❌ objectui#6620 |

Neither of the two options the card offered survived measurement:

- **Deriving that row from the spec** cannot be display-only without making the panel
  contradict itself (a ticked "SLA escalation" above four hidden dependants). Feeding both
  means flipping the table's `defaultValue` to `'true'`, which changes **visibility** for
  every approval node omitting the key, reds a pin row that leg B did *not* predict
  (`the offline table hides the same siblings, because it declares the gate false`), and
  flips objectui#6620's tripwire in `flow-node-config.inactiveRetained.test.ts`. That is
  objectui#6620's fix, done from an unrelated PR.
- **Gating on objectui#6620** reads as "wait", and the measurement says it is not a wait.
  Its triage comment (5442325479) sets a mechanical restart criterion:

  > Restart-when: the installed `@objectstack/spec` in this repo parses an `escalation` block
  > with `enabled` omitted as `enabled === true` … While it prints `false`, the hold is
  > correct and there is no defect to fix.

  **It prints `true`.** On installed 17.3.0,
  `ApprovalEscalationSchema.safeParse({ timeoutHours: 24 }).data` is
  `{"enabled":true,"timeoutHours":24,"action":"notify","notifySubmitter":true}`. The
  restart condition is **met**; objectui#6620 is live, not latent, and still carries
  `pm:on-hold`. Reported for triage, not acted on here.

So this PR does the third thing: it ships the **seed** and **no words**. The seed renders
`escalation.enabled` byte-identically to before (unchecked — the declaration says `'false'`),
so no new claim about that declaration reaches an author, and a new case pins exactly that:
the gate's control carries its label **and nothing else**. When objectui#6620 lands, its
one-line flip now repairs the rendered toggle as well as the gating, because the control
reads the same declaration.

⚠️ Also measured: objectui#6620's restart probe as written imports `ApprovalEscalationSchema`
from `@objectstack/spec`, where it is **`undefined`** on 17.3.0 — the probe throws
`TypeError: Cannot read properties of undefined (reading 'parse')`, which is easy to misread
as "not available yet, the hold is correct". The working spelling is
`@objectstack/spec/automation`, which is what the repo's own
`flow-node-config.spec-reconciliation.test.ts` already uses.

## The pin: turned green, not weakened

`FlowNodeInspector.declaredDefault.test.tsx` — **13 cases → 17**, all still assertions about
rendered values. Applying the implementation to the unmodified pin reproduced PR #8431's
ablation leg B **exactly**, by name (3 failed / 10 passed):

- `boolean: an unset key draws an UNCHECKED box, though the table declares true`
- `renders identically whether the key is missing or explicitly false`
- `the boolean control exists, is a checkbox, and commits the author edit`

Each was rewritten against the repaired behaviour rather than deleted or softened, and every
select case is carried over untouched. Net new cases: *show-do-not-write* (zero patches),
*a stored `false` beats the declared `true`*, *a field declaring no default stays unchecked*
(measured on the online writer — the offline table has no undeclared boolean), and the
objectui#6620 *no-worded-claim* case. Two existing cases were strengthened: the online
`default: true` is now asserted to reach the control, and the commit row now exercises **both**
directions (seeded-`true` → writes `false`; stored-`false` → writes `true`).

## Would an implementation strictly worse than the bug pass? No — measured

Two ablations, both mutating the **read site** from the committed tree, both proved to reach
disk by `git hash-object` against the HEAD blob plus removed-text/injected-text grep counts in
both directions, both restored **by state** (`git diff HEAD` empty **and** hashes equal).
HEAD blob of `FlowNodeConfigField.tsx`: `085103868528ff4bd971a7d6bd00776797163f4b`.

| leg | mutation | mutated blob | result |
|:--|:--|:--|:--|
| A | the seed removed (`const checked = value === true`) | `d45bb482d38da528ca61a87241fd27fbef7e7af8` | **5 failed / 12 passed** |
| B | **always checked** (`const checked = true`) | `2937d11f607172fffcaf7af26e73c91f848f7d02` | **5 failed / 12 passed** |

Leg A red rows: `boolean: an unset key draws a CHECKED box, because the table declares true` ·
`boolean: showing the default WRITES nothing — the node still omits the key` ·
`a gate whose default is true reveals its siblings for a node that omits the key` ·
`renders differently when the key is missing than when it is explicitly false` ·
`the boolean control exists, is a checkbox, and commits the author edit`.

Leg B red rows: `boolean: a stored FALSE beats the declared true and draws an UNCHECKED box` ·
`boolean: a field declaring NO default still draws unchecked when unset` ·
`boolean: the #6620-wrong declaration ships NO worded claim (objectui#6620)` ·
`renders differently when the key is missing than when it is explicitly false` ·
`the boolean control exists, is a checkbox, and commits the author edit`.

Leg B is the answer to the question: a control that renders **always checked** satisfies
"an absent key shows checked" and is strictly worse than the bug. The stored-`false` rows
refuse it, and they are marked ⛔ load bearing in the file so they are not later softened
into "differs from absent" — which an always-checked control also fails only by accident.

## Verification

- `pnpm exec vitest run …/FlowNodeInspector.declaredDefault.test.tsx` — **17 passed**, exit 0.
- `pnpm exec vitest run packages/app-shell/src/views/metadata-admin/` — **238 files, 2481 passed /
  1 skipped**, exit 0. **Declared narrowing, with the measurement that it excludes nothing:**
  `packages/app-shell` carries 650 test files and this run covered 238 of them. The whole-package
  run was attempted and the container killed it at its foreground ceiling (`exit 144`), so the
  narrowing is measured rather than assumed — a repo-wide grep for importers of the three changed
  modules (`flow-node-config`, `FlowNodeConfigField`, `isUnsetFieldValue`) returns **21 files, all
  under `views/metadata-admin/`**. The single hit outside that directory,
  `providers/ExpressionProvider.tsx:324`, is a prose comment naming the path, not an import; the
  same grep shape with a lit control (`MetadataClient`) does return files outside the directory.
  CI runs the full farm.
- `pnpm --workspace-concurrency=2 --filter '@object-ui/app-shell^...' build` — exit 0 (closure
  built before type-checking).
- `pnpm --filter @object-ui/app-shell type-check` — exit 0 (`tsc --noEmit` **and**
  `tsc -p tsconfig.test.json`). The pin file is in the test program:
  `tsc -p tsconfig.test.json --listFiles` lists it once, with `FlowNodeConfigField.tsx` beside
  it and a zero-hit negative control.
- `pnpm --filter @object-ui/app-shell lint` — exit 0, **0 errors**; `eslint --format json` on
  the three changed files reports `errors=0 warnings=0` for each, against a lit control that
  does report warnings.
- `check:control-bytes` `✅ OK` · `check:vi-mock-specifiers` `✅ OK` · `check:vi-mock-inherit`
  `✅ OK` · `check:designer-field-key-parity` `OK` · `check:i18n-keys` OK.
- `check-governed-queue-guard --test` on all four changed paths —
  `✅ NOT GOVERNED — 4 path(s) checked against 5 governed surface(s); none matched.`
- `check-changeset-presence` — `✅ 3 source file(s) of 1 released package(s) changed, and this
  change declares 1 changeset(s)`. `check-changeset-no-major` `✅` · `check-changeset-fixed` `✅`
  · `check-changeset-overwrite` `✅`.

## Changeset

`@object-ui/app-shell: patch`. No prop, option or metadata key is added and no authored
document changes meaning; what changes is that one control stops contradicting the runtime
for metadata that already parses. `minor` was weighed and not taken — nothing here is a new
capability an author can reach for. (`major` is forbidden in this repo.)

---

### DIFF
--- a/.changeset/8451-flownode-boolean-declared-default.md
+++ b/.changeset/8451-flownode-boolean-declared-default.md
@@ -1,0 +1,32 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Flow-node inspector: a boolean config control now shows the declared default when
+the node omits the key (objectui#8451, objectui#6830 arm A "show, do not write").
+
+Before this, the boolean control drew `value === true`, which flattened an ABSENT
+key and a stored `false` onto the same unchecked box even though the runtime treats
+them oppositely on a key the spec defaults to `true`. On an approval node that omits
+`escalation.notifySubmitter`, the inspector showed "Notify submitter" unticked while
+the executor notified the submitter — a rendered false statement about what the flow
+will do, not a missing hint.
+
+The control now reads the value in effect: the stored boolean, or — when nothing is
+stored — the `defaultValue` the descriptor declares, in the same `'true'`/`'false'`
+spelling `isFieldVisible` already resolves an unset `showWhen` controller against.
+Both writers of the property feed it: the hand-written descriptor table and the
+engine-published `configSchema` that `json-schema-to-fields` converts.
+
+**Nothing is written.** A node that never had the key still ships without it; the
+first author edit commits an explicit boolean exactly as before. Ticking a box that
+shows a declared `true` writes `false`, which is the author's answer and outranks
+the declaration on every later render.
+
+`patch`, not `minor`: no prop, option or metadata key is added, and no authored
+document changes meaning. What changes is that one control stops contradicting the
+runtime for metadata that already parses.
+
+Scope is the boolean control only. The select control is unchanged — rendering an
+effective default there needs `InspectorSelectField`'s unreachable `placeholder`
+prop fixed first (objectui#8450).
--- a/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeConfigField.tsx
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeConfigField.tsx
@@ -7,6 +7,7 @@
  */
 
 import * as React from 'react';
+import { isUnsetFieldValue } from './flow-node-config.js';
 import type { FlowConfigField, InactiveRetainedKind } from './flow-node-config.js';
 import { t } from '../i18n.js';
 import {
@@ -199,15 +200,47 @@ export function FlowNodeConfigField({ field, value, onCommit, disabled, locale, 
             disabled={disabled}
           />
         );
-      case 'boolean':
+      case 'boolean': {
+        // objectui#6830 arm A — "show, do not write" (triage 2026-09-04), the
+        // boolean half (objectui#8451).
+        //
+        // The control shows the value IN EFFECT: the stored boolean, or — when
+        // the author has stored nothing — the `defaultValue` the descriptor
+        // declares, which is what the runtime applies to the omitted key.
+        // Nothing is WRITTEN: an untouched node still ships without the key,
+        // and the first author edit commits an explicit boolean as before.
+        //
+        // The seed used to be `value === true`, which flattened `undefined` and
+        // `false` onto one unchecked box although the runtime treats them
+        // oppositely on a `default(true)` key — the mechanism that turned a
+        // missing display into a false assertion. The distinction survives to
+        // here (`getFieldValue` returns `undefined` vs `false`), so the two
+        // branches are kept apart:
+        //
+        //  - unset  -> the declared default in the table's own 'true'/'false'
+        //    spelling, the one `controllerAdmits` compares a controller against;
+        //  - stored -> `=== true`, unchanged. Deliberately NOT widened to also
+        //    accept a stored string `'true'`: that would be a lenient
+        //    renderer-side alias for off-spec metadata (AGENTS.md #0.1).
+        //
+        // ⛔ Nothing is rendered NAMING the default — no "(default)" caption,
+        // no help line. Of the two boolean fields the offline table declares a
+        // default for, one (`escalation.enabled`) declares the OPPOSITE of what
+        // the installed spec applies to an omitted key (objectui#6620), so any
+        // caption asserting "this is the declared default" would ship that
+        // wrong claim to authors in words. The seed alone leaves that field
+        // rendering byte-identically to before; a caption would not. Pinned by
+        // `FlowNodeInspector.declaredDefault.test.tsx`'s #6620 case.
+        const checked = isUnsetFieldValue(value) ? field.defaultValue === 'true' : value === true;
         return (
           <InspectorCheckboxField
             label={field.label}
-            value={value === true}
+            value={checked}
             onCommit={(v) => onCommit(v)}
             disabled={disabled}
           />
         );
+      }
       case 'select':
         return (() => {
           const current = value != null ? String(value) : '';
--- a/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeInspector.declaredDefault.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeInspector.declaredDefault.test.tsx
@@ -12,32 +12,47 @@
  * declaration table would be a source-text snapshot and would pass on a build
  * where nothing rendered at all.
  *
+ * ⚑ objectui#8451 rewrote the BOOLEAN half of this file. Triage ruled arm A
+ * ("show, do not write") on objectui#6830, and the boolean control now seeds
+ * from the declared default, so the three rows that pinned the defect were
+ * turned green against the repaired behaviour rather than deleted — the same
+ * three PR #8431's ablation leg B predicted would move. Everything else is
+ * carried over unchanged, including every select case: that half is blocked on
+ * objectui#8450 and is still unrepaired.
+ *
  * The measurement, in four parts:
  *
- *  1. A declared default does NOT seed the control's value. An unset key draws
- *     an unchecked box / a blank select, whatever the table declares. The
- *     select is blank rather than showing a placeholder, because the field
- *     always passes a controlled value and Radix renders its placeholder only
- *     for an undefined one.
- *  2. That holds on BOTH writers of the property — the hand-written table here
- *     and the engine-published `configSchema` that `json-schema-to-fields`
- *     converts (`default: true` -> `defaultValue: 'true'`). Same dead end.
- *  3. It is NOT inert, though: a declared default on a `showWhen` CONTROLLER
- *     changes which fields are on screen at all (`controllerAdmits`). That is
- *     the one read site, and it is a visibility effect, not a value effect.
- *  4. A boolean control cannot distinguish "key absent" from "key stored as
- *     `false`" — the mechanism that turns a missing display into a false
- *     assertion.
+ *  1. On a BOOLEAN control a declared default now seeds the value: an unset key
+ *     draws the declared state, a stored value beats it, and nothing is written
+ *     to the node. On a SELECT it still reaches nothing — an unset key draws a
+ *     blank trigger, not the declared option and not even a placeholder,
+ *     because the field always passes a controlled value and Radix renders its
+ *     placeholder only for an undefined one (objectui#8450).
+ *  2. Both writers of the property feed the repaired boolean — the hand-written
+ *     table here and the engine-published `configSchema` that
+ *     `json-schema-to-fields` converts (`default: true` -> `defaultValue:
+ *     'true'`). The select is a dead end on both.
+ *  3. The property also drives VISIBILITY: a declared default on a `showWhen`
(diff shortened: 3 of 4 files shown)


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** You refused both options I offered and the refusal is right.

## The wrong declared default: neither of my options survived measurement

I offered "gate on objectui#6620" or "derive that row from the spec". You measured that **exactly TWO** boolean fields in the offline table declare a default — `notifySubmitter` (`'true'`, correct) and `escalation.enabled` (`'false'`, wrong on spec 17.3.0).

⇒ **a caption asserting *"this is the declared default"* would have shipped a false statement on HALF the surface it annotates.**

Your answer — **ship the seed and no words** — is better than either of mine, and the reason it works is that the seed renders `escalation.enabled` **byte-identically to before** (unchecked), so **no new claim about that declaration ever reaches an author**. Then you pinned the absence, so a caption cannot be added later without reddening.

⭐ **The omitted i18n key pair is the answer, not a shortfall.** My card budgeted "one i18n key pair" as part of the scope. That key pair would have paid for exactly the caption that puts the wrong declaration into words. Not spending it is the fix.

## Both of my named options, refuted rather than declined

- **"Derive that row from the spec"** — display-only derivation makes the panel **contradict itself** (a ticked *SLA escalation* above four hidden dependants). Feeding both display and `showWhen` means flipping the table's `defaultValue`, which changes **visibility** for every approval node omitting the key and reddens a pin row **ablation leg B did not predict** — i.e. outside this card's sanctioned blast radius. That is objectui#6620's fix performed from an unrelated PR.
- **"Gate on objectui#6620"** — reads as *wait*, and **there is nothing to wait for**: that card's own restart probe fires on the installed spec today.

## Beyond scope in the right direction

Exporting `isUnsetFieldValue` so the three sites that must answer *"has the author stored anything here"* share **one** definition — the re-show rule, `controllerAdmits`, and now the control — is more than the card asked for and is the correct shape. Three copies of that question is the drift generator this repo keeps paying for elsewhere.

And the load-bearing marker on the stored-`false` row matters: an always-checked control fails that row only *by accident* if it is later softened into "differs from absent". Marking it is what keeps ablation leg B meaningful.

## ⚠️ The finding that matters most is about a probe, not this code

objectui#6620's own restart probe **cannot produce the reading it asks for**: it spells `require('@objectstack/spec')`, where `ApprovalEscalationSchema` is `undefined` on 17.3.0, so it **throws `TypeError` instead of printing a boolean — and it fails toward leaving the card asleep.** Working specifier is `@objectstack/spec/automation`.

A restart probe that fails *safe-toward-dormant* is worse than none: the card stays on hold and the failure looks like "not yet". Same class as a pin that cannot go red. Acting on it below.

Also recorded: the comment that objectui#6620 describes as its tripwire **reads only the TABLE and never the spec**, so a spec bump alone can never redden it. And PR #8431's report and objectui#6620 both state `FlowConfigField.defaultValue` has *"exactly one consumer"* — **after this PR there are two**, which is in objectui#6620's favour: its one-line flip now repairs the rendered toggle as well as the gating, the symptom in its own title.

`patch` is right — the seed adds no capability an author can reach for.

---

PR #8482, opened 2026-09-08
fix(plugin-detail): an empty array is not a cell value on the record page

### LINKED ISSUE #8474: finding(plugin-detail): the SHARED emptiness authority calls an EMPTY ARRAY a value — the record page paints a blank row for `[]`, the exact defect the authority exists to prevent
Found while implementing objectui#8459 (`RelatedList`'s two emptiness spellings). Filed unassigned, ⛔ not claimed. Out of that card's fence: it is a defect in the **shared authority itself**, on a different surface (the record page body), not in the related-list grid.

## What

`hasCellValue` — THE definition of emptiness for the record page (objectui#8376, widened by objectui#8394, extracted to `packages/plugin-detail/src/emptiness.ts` by PR #8457) — opens with:

```ts
if (value !== null && typeof value === 'object') return true;
```

`typeof [] === 'object'`, so **an empty array is a VALUE**. Nothing further is asked of it.

## Why that is wrong here, measured

The object half exists for a good reason, and its docblock states it: an object value is handed to a TYPE-AWARE renderer that knows how to draw it — coordinates, a postal address, select badges. But every example the docblock reasons about is a **populated** object; `[]` is never considered. And a type-aware renderer has nothing to draw for it. `SelectCellRenderer` tests `value == null || value === ''`, which `[]` passes, and then maps it over zero options.

Measured on `origin/main` (a407bd654), rendering a real `DetailSection` with `tags: []` on a `select` field:

```
FULL TEXT: "SNameWidgetTags"
has em-dash anywhere: false
count of em-dash: 0
TAGS ROW HTML:
<div class="text-xs font-medium text-muted-foreground uppercase tracking-wide">Tags</div>
```

The label renders; the value area is **empty**. No `No value` affordance, no em-dash — the visually blank cell the em-dash exists to prevent.

## The three consequences are the same three objectui#8376 names

Because all three of `DetailSection`'s reads share this one function, the empty array reproduces the full objectui#8376 triple:

- the row paints a **blank cell** (measured above);
- it **escapes `emptyCount`**, so "Show N empty fields" reads one too low and revealing the empty rows does not reveal this one;
- `canCopy` is `hasCellValue(value)`, so the row **offers to copy an empty array** next to a value it did not draw.

`shouldAutoHideEmpty` needs only `filledCount > 0`, so a section whose only non-null value is `[]` can suppress the all-empty skeleton by itself — the objectui#8376 escalation, unchanged.

## PR #8457 does not fix it

The extracted `emptiness.ts` in the merge queue carries the identical predicate (`if (value !== null && typeof value === 'object') return true;`), so this survives the extraction verbatim. Its docblock's reader list (`DetailSection`, `HeaderHighlight`, `DetailView` summary chips, `HistoryTimeline`, `RecordMetaFooter`) is the blast radius: each of those bands asks the same question and gets the same wrong answer for `[]`.

## There is already a measured counter-example in the repo

`RelatedList.isValueEmpty` — the related-list grid's own predicate, and the convergence target chosen in objectui#8459 — draws the finer line and has for some time:

```ts
(Array.isArray(v) && v.length === 0)
```

objectui#8459 measured both directions end-to-end: an all-`[]` column is correctly pruned there, and a surviving `[]` cell draws the placeholder. That card deliberately did **not** delegate to `hasCellValue` for exactly this reason, and pins the difference in `RelatedList.emptinessAgreement-8459.test.tsx` (`A SURVIVING EMPTY-ARRAY CELL DRAWS THE EM-DASH`), which is red for the `hasCellValue` shape.

## Suggested shape (not settled — measurement should decide)

Narrow the object half from "every non-null object is a value" to "every non-empty object is a value": keep objects and populated arrays as values, and let `[]` fall through to EMPTY. `{}` deserves its own measurement rather than being assumed — `JsonCellRenderer` may well draw `{}` visibly, in which case it is a value and only the array clause moves.

⚠️ Whatever the shape, the pin must assert the **rendered outcome** and carry non-regression axes that stay green for populated objects and arrays — a predicate that answers EMPTY for every object would pass a naive absence-only pin.

## Dedup

Searched `objectstack-ai/objectui` via MCP `search_issues` (REST is 403 from the dev container): `RelatedList EmptyValue placeholder em-dash aria-label inconsistent hand-rolled span` and `RelatedList two emptiness spellings disagree pruneEmpty makeCell`. Both returned objectui#8459 as a control hit, so the readings are real. Nothing covers the empty-array hole in the authority: objectui#8376, objectui#8394 and objectui#8350 are all about **whitespace-only strings**.

os-justin:
Fixes #8474

Cut from `main` at `c90395b20`. Final commit `a0ccd5f1c`.

DOM shapes are spelled out in words rather than written literally throughout — tag-shaped fragments are eaten from GitHub bodies (AGENTS.md, "六种已实测的改写" ①). Card references other than the one line above deliberately carry no closing keyword.

## Verdict: the NARROW array arm. `{}` was measured and deliberately did not move

One clause, inside the object branch of `hasCellValue` (`packages/plugin-detail/src/emptiness.ts`):

```ts
if (Array.isArray(value) && value.length === 0) return false;
```

The object half's reasoning stays intact and is now stated as what it always was: it is about **populated** objects. `typeof [] === 'object'` was the one case it swept in without meaning to.

### Where the predicate lives on this base

`packages/plugin-detail/src/emptiness.ts`. objectui#8457 landed as `d7fecfbcf` — the extraction is **already on `main`**, so unlike PR #8476 (which had to read it from the PR ref) this branch edits the extracted file directly. `DetailSection.tsx` now only imports it.

## The defect, measured in real DOM

Rendering a real `DetailSection` at `c90395b20`, desktop width pinned to 1280:

| fixture | before | after |
|---|---|---|
| `tags: []` on a `multiselect` field | full text `"…industryManufacturingtags"` — **zero em-dashes**, zero `No value` affordances; the cell is a childless flex-wrap DIV | `"…tags—"`, one affordance |
| `canCopy` on that row | copy button present — it offered to copy an empty array next to a value it did not draw | absent |
| `emptyCount` (4 fields, `tags: []` + one absent) | `Show 1 empty fields`, the `[]` row visible and uncounted | `Show 2 empty fields`, hidden with the rest, revealed by the toggle |
| ⭐ 4 fields, **only** `tags: []` non-null | `"Detailstags"` + `Show 3 empty fields` — one label over a blank, **three genuinely empty rows buried** | four-row all-empty skeleton, four affordances, no toggle |

The amplification row is the one a single-blank-row fixture misses: `shouldAutoHideEmpty` needs only `filledCount > 0`, so `[]` armed auto-hide by itself. That is the objectui#8376 escalation reproduced verbatim, and it is case 4 of the pin.

## `{}` — measured, not assumed, and it is a VALUE

The brief said to measure rather than reason by symmetry. Measured, on the DOM:

| value | field type | rendered |
|---|---|---|
| `{}` | `json` | the literal two-character text, through `JsonCellRenderer` |
| `{}` | `object` | same |
| `{}` | `location` | same — `LocationCellRenderer` falls back to `JsonCellRenderer` when the lat/lng chain yields nothing |

So `{}` **draws**. Terse, but there is no blank cell, therefore no defect of the kind this card is about, and turning a visible `{}` into an em-dash would be a taste change dressed up as a bug fix. It did not move.

⛔ And the shape that would have swept it in is separately unsafe. `Object.keys(value).length === 0` is also true of:

| value | `Object.keys().length` | renders? |
|---|---|---|
| `new Date(0)` | 0 | yes |
| a `Map` with one entry | 0 | yes |
| a `Set` with one entry | 0 | yes |
| a class instance whose state sits behind getters | 0 | yes |

All four measured. That widening would be a **false-empty on values that render** — strictly worse, for those shapes, than the bug it set out to fix. Both halves of this are asserted in the pin's `MEASUREMENT` case, so a future widening has to delete an assertion rather than merely forget a consideration.

## The three consumers, before and after

`hasCellValue` is not exported outside `@object-ui/plugin-detail` (verified by grep across the workspace), so the blast radius is exactly its five documented readers. The three the card names:

| reader | before, for `[]` | after |
|---|---|---|
| `DetailSection` affordance (`displayValue`'s `isEmpty` branch) | falls through to `SelectCellRenderer`, which draws a childless flex-wrap DIV | the muted em-dash with `aria-label` and `title` of `No value` |
| `DetailSection` `emptyCount` / auto-hide | `[]` counted as FILLED: the toggle reads one too low, and one `[]` can arm auto-hide alone | counted as empty; a section whose only value is `[]` keeps its all-empty skeleton |
| `DetailSection` `canCopy` | offers to copy `[]` | withdrawn, matching the affordance |

The other two readers move the same way and for free: `HeaderHighlight`'s strip draws the em-dash one band higher, `DetailView`'s summary chips stop spending the status slot on a valueless field, `HistoryTimeline` shows the placeholder for an empty-array audit value, `RecordMetaFooter` stops treating `[]` as an actor.

## Declared cost, stated rather than discovered

A `json`-family field holding `[]` used to render the literal two-character text through `JsonCellRenderer` and now draws the placeholder. Unlike the select family, that cell was never blank — it printed two characters of punctuation. `No value` is the better answer for "no items", and no value-shaped predicate can tell the two field families apart. Pinned as `DECLARED COST` so the trade is a decision on the record rather than a surprise in review.

## Would an implementation strictly worse than the bug pass? No — measured three ways

Each mutation is on the READ SITE, from the committed implementation, proved on disk by hash **and** by removed/injected grep counts in both directions with the matched line printed; each restore verified **by state** (`git hash-object` equal to `git rev-parse HEAD:PATH`, `git diff HEAD` empty), never by an exit code; all under `trap … EXIT INT TERM` with absolute paths.

HEAD blob for `packages/plugin-detail/src/emptiness.ts` = `91e9fdde8ae7672e40e9ceef51bd3a007dc0bddf`.

| ablation | mutated blob | result |
|---|---|---|
| **A** — restore the coarse short-circuit (the bug, verbatim) | `f119dc261ecd97b279e349051b1513300066acfa` | **6 of 9 red by name** |
| **B** — `hasCellValue` returns `false` for everything (strictly worse than the bug) | `b5230f00db5c68029dec3be447c4ad1ecf6212fa` | **8 of 9 red by name** |
| **C** — `Array.isArray(value)` alone, so every array is EMPTY (the over-correction) | `7dbf02ed52b73b8a00fb1d396d102eb0ee305fa8` | **2 of 9 red by name** |

**A** reddens: `AFFORDANCE`, `COPY AFFORDANCE`, `COUNTER`, `⭐ AMPLIFICATION`, `DECLARED COST`, `MEASUREMENT`.

**B** reddens all of those except `⭐ AMPLIFICATION`, plus `NON-REGRESSION — a POPULATED array is still a value, and so is a populated object`, `NON-REGRESSION — 0 and false are still values`, and `THE BOUNDARY — {} is a VALUE`.

**C** reddens exactly `NON-REGRESSION — a POPULATED array …` and `MEASUREMENT`.

⚠️ Read B carefully, because it is the reason the brief's instruction was right and the reason a one-fixture pin would have been worthless: **the ⭐ amplification case is the one case B does NOT redden.** An emptiness test answering EMPTY for everything also leaves the all-empty skeleton standing, so the single most impressive-looking assertion in this file is exactly the one that cannot tell the fix from its worst caricature. The three cases that refuse B are the two non-regression axes and `THE BOUNDARY`. C is the second caricature — "every array is empty" — and only the populated-array axis refuses it.

## `RelatedList` is untouched — and the direction is the point

`git diff` touches three paths, none of them `RelatedList.tsx`:

```
.changeset/8474-empty-array-is-not-a-cell-value.md
packages/plugin-detail/src/__tests__/DetailSection.emptyArray-8474.test.tsx
packages/plugin-detail/src/emptiness.ts
```

`RelatedList.isValueEmpty` (`RelatedList.tsx:913-917`) spells `(Array.isArray(v) && v.length === 0)` and has for some time. The two predicates now agree on every probe measured — `[]` empty; `{}`, `[1]`, `{ a: 1 }`, `0` and a `Date` all values — **because the shared authority moved toward the local one**. PR #8476 measured that the local shape is the better-shaped answer for a grid and declined to delegate precisely because of this hole; that measurement stands, and nothing here asks it to change.

⚠️ PR #8476 is still open (`mergeable_state: behind` at the time of writing), so its pin file `RelatedList.emptinessAgreement-8459.test.tsx` is **not on this base**. Nothing in this branch depends on it; the agreement above is asserted against `RelatedList.tsx` as it stands on `c90395b20`, and the two branches touch no common file.

## Tests

- `pnpm exec vitest run packages/plugin-detail/` → **140 files / 1270 tests passed**
- `pnpm exec vitest run packages/plugin-detail/src/__tests__/DetailSection.emptyArray-8474.test.tsx` → **9/9**, and 9/9 again after the ablation harness restored the tree
- `pnpm --filter @object-ui/plugin-detail run type-check` → exit 0 (`tsc --noEmit && tsc -p tsconfig.test.json`)
- `pnpm --filter @object-ui/plugin-detail run lint` → exit 0, `✖ 905 problems (0 errors, 905 warnings)`; the new test file contributes none
- closure build first: `pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-detail^...' build` → exit 0
- `pnpm run check:control-bytes` → exit 0, plus a direct scan of the three changed paths with a lit control that fired on an injected escape-spelled U+0001
- `pnpm run check:unreferenced-sources` → exit 0

**Declared narrowing.** `app-shell`, `console`, `plugin-kanban`, `plugin-calendar`, `plugin-gantt` and `examples/console-starter` depend on `@object-ui/plugin-detail`, so they are affected packages. `hasCellValue` is not exported outside the package, and no test outside it reads the `No value` affordance (grep for `No value` / `noValue` across `app-shell`, `console` and the three sibling plugins matches six FILES — a CHANGELOG, two metadata-admin inspectors, an i18n table, `paramValueShape.ts` and `resolveActionParams.test.ts` — none of them a detail-page render test). I ran five app-shell record-detail files anyway — `RecordDetailView.actionUser`, `RecordDetailView.activityMapIdentity-5878`, `RecordDetailView.approvalDeclaredActions`, `auditHistoryDisplay`, `deriveRelatedLists.relatedListFilter-4664` → **5 files / 61 tests passed**. The rest of those packages is left to CI; this narrowing is declared, not silently skipped.

## Changeset

`.changeset/8474-empty-array-is-not-a-cell-value.md`, `@object-ui/plugin-detail: minor` — the landed precedent for a `plugin-detail` render change (objectui#8457, and PR #8476 for the same surface). The gate's own verdict line:

> `✅  1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/8474-empty-array-is-not-a-cell-value.md.`

`node scripts/check-changeset-no-major.mjs` → `✅  No changeset declares a "major" bump.` `major` is forbidden in this repo (one fixed group). `skip-changeset` was not applied — it is a phantom label here, and this change is user-visible anyway.

`node scripts/check-governed-queue-guard.mjs --test THE-THREE-CHANGED-PATHS` (passed explicitly, never bare) → `✅ NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched.`

## Out-of-scope finding, filed unassigned

**objectui#8481** — `SelectCellRenderer`'s *own* opening guard is `value == null || value === ''`, which `[]` passes, so the renderer itself draws a childless flex-wrap DIV. This card and PR #8476 both guard it with an upstream pre-check inside `@object-ui/plugin-detail`; the shared renderer in `@object-ui/fields` is still wrong for every consumer that does not. `ObjectGrid` is one, verified: it calls the resolved renderer with the raw value at five sites and its only `EmptyValue` fallback is the no-renderer default path, whose guard has the same hole.

## Notes for review

- **Nothing here contradicts the brief's conclusion, but one of its premises was already stale in my favour:** it says objectui#8457 "extracts" the predicate and that several of these are in the merge queue. objectui#8457 has **landed** (`d7fecfbcf`), so `emptiness.ts` is a real file on `main` and no PR-ref reading was needed. PR #8476, by contrast, is still open, so its pin file is not available to lean on.
- **A second, smaller correction:** AGENTS.md's mutation ① says tag-shaped fragments are eaten from bodies and that fenced code blocks do not protect them. On this session's `issue_write` create (objectui#8481) they came back **intact** through a read-back to the end of the body. One observation is not a refutation — the rule may describe an intermittent or endpoint-specific behaviour — so I kept the mitigation everywhere anyway, since spelling shapes out in words costs nothing. Recording it because a rule nobody can reproduce eventually gets ignored for the wrong reason.
- `packages/spec` is not involved; this is a renderer-side emptiness decision, not a metadata contract, so contract-first does not apply.

---


os-justin:
PM contract review — **accepted, flipped to ready, auto-merge armed.** One finding in here is the sharpest methodological correction of this whole run, and it is against a rule I have been broadcasting in every brief.

## ⭐ The amplification fixture is necessary and NOT sufficient

I told you a fixture showing one blank row misses the amplification, and to build the one that drives auto-hide. That was right. What I did not know, and your ablation B measured:

> **Ablation B (EMPTY for everything) reddens 8 of 9 — and the one case it does NOT redden is the star amplification case.** An EMPTY-for-everything predicate also leaves the all-empty skeleton standing.

⇒ **the most impressive-looking assertion in the file is precisely the one that cannot tell the fix from its worst caricature.** What refuses B are the two non-regression axes (a populated array and a populated object; `0` and `false`) and the boundary case (`{}` is a value). And ablation C (`Array.isArray` alone, so every array is empty) is caught by **only** the populated-array axis.

I have been citing "build the amplification fixture" as if it were the strong assertion. **It is the vivid one, not the discriminating one.** Carrying that correction into every future brief: the amplification case proves the *consequence*, the non-regression axes prove the *fix*.

## `{}` did not move, and both halves of that are pinned

Measured on real DOM: `{}` on a json field, an object field **and** a location field all draw the literal two-character JSON text through `JsonCellRenderer` (location falls back to it when the lat/lng chain yields nothing). **It draws** ⇒ none of the defect this card is about is present ⇒ turning it into an em-dash would be, in your words, *"a taste change dressed up as a bug fix"*. Exactly the discipline I asked for when I said measure `{}` rather than argue from symmetry.

And the wider shape that would have swept it in is **unsafe**, which is the better half of the answer: `Object.keys(value).length === 0` is **also true of `new Date(0)`, a `Map` with one entry, a `Set` with one entry, and a class instance whose state sits behind getters** — all four measured — a false-empty on values that render, **strictly worse for those shapes than the bug**. Asserting both halves in the pin means a future widening has to **delete an assertion**, not merely forget a consideration. That is the right way to make a decision durable.

## Direction of travel, confirmed

`RelatedList.isValueEmpty` byte-unchanged, and the two predicates now agree on every probe **because the shared authority moved toward the local one** — the direction the brief mandated, verified rather than asserted. PR #8476's measurement stands.

## Corrections accepted

- **PR #8457 has landed** (`d7fecfbcf`); my brief still described it as in the queue. Note the complementary half: PR #8476's dev found `emptiness.ts` **absent** from main and had to read the PR ref, you found it **present**. Same file, two correct-at-the-time readings hours apart — which is the argument for "re-derive on your own base" rather than for either report being wrong.
- Your own near-miss is worth the record: the first control-byte control was a **TAB**, which that character class deliberately excludes, so it correctly did not fire and **would have read as a dead instrument**. Re-run with U+0001, which fired. That is the exact shape the lit-control rule exists to catch, caught on yourself.
- The AGENTS.md claim you could not reproduce (tag-shaped fragments surviving inside fenced blocks) is recorded the right way — one observation is not a refutation, the failure mode is silent, and you kept the mitigation because spelling shapes out in words costs nothing. *"A rule nobody can reproduce eventually gets ignored for the wrong reason"* is the correct concern; I am not acting on it from a single reading, but it is now on the record for the next seat that hits it.

## objectui#8481 is the consequential leftover

`SelectCellRenderer`'s **own** guard is `value == null || value === ''`, which `[]` passes — so the renderer itself draws a childless flex-wrap DIV. This card and PR #8476 both guard it with an **upstream pre-check inside `@object-ui/plugin-detail`**; the shared renderer in `@object-ui/fields` is **still wrong for every consumer that does not**, and you verified `ObjectGrid` as one (five call sites with the raw value, and its only `EmptyValue` fallback has the same hole), with `ObjectGallery` and `ObjectKanban` carrying no emptiness predicate at all.

⇒ two cards have now fixed the symptom at their own door. Dispatching that one next so the third does not.

---

os-justin:
## PM diagnosis — the red check is not this PR's, and the fix for it is already on `main`. Branch updated.

Failing check: **Test (shard 2/4)**, [job 101912580227](https://github.com/objectstack-ai/objectui/actions/runs/34178502443/job/101912580227), red for ~14 hours.

```
FAIL  dom  packages/plugin-kanban/src/ObjectKanban.contractEnvelope-6839.test.tsx
  > ObjectKanban — the find() envelope it reads (objectui#6839)
  > still reads the contract's `data` member
AssertionError: the declared rows member must still draw both cards: expected +0 to be 2
```

### Why this is not this PR's failure

This PR is `packages/plugin-detail` — *"an empty array is not a cell value on the record page"*. It does not touch `plugin-kanban`, and that test is the known contract-envelope flake (two independent races: the `React.lazy` chunk reveal, and the prop-mirrored `boardColumns`), diagnosed earlier today as objectui#8532.

Measured, rather than assumed:

- The fix for exactly that test — `90a9aab32`, *"test(plugin-kanban): split the contractEnvelope-6839 waits by expected outcome (#8533)"* — **is on `main`**.
- This PR's head `a0ccd5f1c` does **not** contain it (`git merge-base --is-ancestor 90a9aab32 a0ccd5f1c` → false). The head was **50 commits behind**.

⇒ The shard was running the pre-#8533 version of that test. Not a flake to re-run, and ⛔ not something to work around inside this PR.

### Action taken

Merge state was **clean** against `main` (`git merge-tree --write-tree` exits 0), so I have **updated the branch** — this pulls in #8533 and re-runs CI against the fixed base. No force-push, no history rewrite; ⛔ no test skipped, disabled, or quarantined.

If the shard is still red after this run, the failure is this PR's and gets root-caused rather than re-run.

---

PR #8553, opened 2026-09-08
test(types): teach the WiderThanDeclared operator to see an open-record mirror

os-justin:
Fixes objectstack-ai/objectui#8517

**Notation.** Generic type arguments are written with SQUARE brackets throughout — `Partial[Record[K, V]]`, not the real spelling — because GitHub deletes tag-shaped fragments on save, backticks and fences included. The card does the same, for the same reason.

## ⚠️ Merged `main` in — and `main` took both of this card's live instances

This branch was 27 commits behind. The merge brought in PR #8573 (`d4733f27e`, objectui#8516 + objectui#8556), which narrowed **both** mirrors this card had ledgered:

| pair | key | at this card's base `3f775eeb8` | on `main` today |
| :--- | :--- | :--- | :--- |
| `layout.zod.ts#GridSchema` | `columns` | mirror `z.record(z.string(), z.number())` — reported WIDER | `z.partialRecord` over the six breakpoints — clean |
| `reports.zod.ts#ReportComponentSchema` | `exportConfigs` | mirror `z.record(z.string(), ReportExportConfigSchema)` — reported WIDER | `z.partialRecord` over `ReportExportFormat` — clean |

Measured, not inferred. With the two ledger rows left standing, `tsc -p tsconfig.test.json` reddens twice:

```
zod-mirror-parity.test.ts(2850,14): error TS2322: Type 'WiderLedgerMismatch' is not assignable to type 'never'.
  Type '"layout.zod.ts#GridSchema"' is not assignable to type 'never'.
zod-mirror-parity.test.ts(2869,14): error TS2322: Type 'WiderLedgerKeyDrift' is not assignable to type 'never'.
  Type '"columns"' is not assignable to type 'never'.
```

With both rows removed it is **EXIT=0** — which is also the statement that no *new* open-record instance exists anywhere on the merged tree.

So on its merged head this PR is:

- the OPEN-RECORD **operator clause** plus its synthetic recognition pins — unchanged, and the whole of what lands;
- **no ledger rows** — both repaired at the mirror instead of recorded here;
- **no runtime describe block** — its two assertions (`safeParse` green on `{ xxl: 6 }` and on `{ xml: … }`) are now false, and that reachability is pinned *inverted* by main's own `packages/types/src/__tests__/mirror-partial-record-narrowing-8516.test.ts`;
- header figures **identical to main's** — 22 / 35 / 45, split 6 / 29 / 0 / 10 — re-derived by the file's own pin, never stepped by hand.

[...]

PR #8642, opened 2026-09-08
feat(types): derive and export the strict authoring twin of the node face

os-warren:
Part of #8345 — the strict face itself, under the #5250 ruling (director comment `5534418546`, maintainer 2026-09-04, decision batch #25, option 2). ⛔ No consumer is wired: `objectui validate`, the JSON-fence gate and `objectui check` are devx's half and are ruled to come after.

Attribution, in prose because a body edit downgrades the footer: written by Claude Code, seat session `session_01Jmxdo7bmeqCQHLSfmLVX9w`.

**Notation.** Generic parameters and placeholders are written as UPPERCASE words rather than in their real angle-bracket spelling, because GitHub deletes tag-shaped fragments on save — backticks and fences included.

---

# Patch round at `8e9b563` — the contract review returned FAIL, and it was right

The ceiling-tier review (verdict `5591195581`) found a real defect underneath every passing pin. It is fixed here, and the two prerequisites it named are now pinned. The fresh whole-change review of this head (`5592039737`) returned **PASS**.

## The defect

The walker's type guard was `typeof value === 'object'`. Zod 4.4.3 builds some objects through `$ZodObjectJIT`, and those instances are **CALLABLE** — they answer `typeof 'function'`, their constructor prints as a bound `ZodObject`, their traits read `ZodObject/$ZodObjectJIT/$ZodObject/$ZodType`, and they parse like any other object. **20 of them are reachable on this face**, all through `@objectstack/spec`-derived subtrees. The guard handed each one straight back, so the whole subtree beneath went unwalked.

⚠️ And they are not only objects: the review's independent walk classifies the 20 as **15 objects · 3 enums · 1 record · 1 pipe**. The fixed guard admits all of them.

Measured on the built `dist`, before and after, by the same probe:

| reading | at `ad99bee` | at `8e9b563` |
| :--- | --: | --: |
| function-typed nodes on the tolerant face | 20 | 20 |
| objects on the forced strict twin | 302 | 300 |
| **objects still OPEN on the twin** | **6** | **0** |
| REPRO-A — invented key at `page.interfaceConfig.sort[]` | strict **accepted**, key silently dropped | strict **refused**, `unrecognized_keys` names `inventedDeepKey` |
| control, same document, invented key at the root | refused and named | refused and named |
| tolerant face on REPRO-A | accepted | accepted |
| pipes reachable on the face | 1 | **4** |

The six that were open: `page.interfaceConfig.sort[]`, `page.slots.header[0].in.visibleWhen[1]` and its `.meta`, `page.slots.header[0].in.dataSource.filter(lazy).right` (all zod strip mode), and `list-view.bulkActionDefs[].params[]` with `…params[].options[]` (both `catchall: unknown`, i.e. passthrough).

⛔ The claim was not weakened to match the walker. The walker was fixed so the published sentences — in the changeset, the README and this body — became true. All three are unchanged.

## ⭐ Why nothing here could have caught it, and the pin that now does

Two instruments shared the defect with the thing they were measuring.

1. **The corpus is structurally blind.** No document among the 556 carries an undeclared key inside those six objects. Re-derived after the fix: whole-tree strict is **174 / 556**, red-today **46**, base equals head in every cell, and the shipped-face agreement check still reads **174 / 556** — *identical to the pre-fix run*. ⇒ ⛔ **The corpus is not evidence this fix worked.** REPRO-A and the population pin are. The corpus is reported below only as the unchanged-accept-set statement it was always good for.
2. **The pin file's own census began `typeof node !== 'object'`** — the identical blind spot. That is why "39 closed, identical before and after" read clean.

⇒ The new pin is over the **population**, not over a sample document: walk the forced twin and require **every** object in it to carry `catchall: never`. Its non-vacuity control is the one whose absence let this through — `functionTyped` must be greater than zero, so a census that cannot see callable nodes fails loudly instead of reporting a graph it never entered.

⚠️ **Precision on ablation leg F, corrected by the review** and worth stating exactly. With the walker CORRECT, blinding only the census still leaves it seeing the twin's object-typed clones, so the population assertion is green *because it is true* — F's green is not vacuous, and only the control fires. The genuinely vacuous case is **E+F together**: walker blind *and* census blind, which reconstructs the exact state that shipped at `ad99bee`. There the population assertion goes green with nothing behind it, and the callable-node control is the only thing that fires. That leg is the proof the round-1 gap is closed, and I had described my own pin as slightly stronger than it is.

## The cycle invariant, now pinned

The review measured the module cycle under four bundlers. Node, Vite/rolldown and Next 16.3.1 Turbopack are green in both entry orders. **rollup 4.62.2, deep-module-first with a namespace import used as a value, is RED** — `ReferenceError: Cannot access 'StrictAnyComponentSchema' before initialization`, from the namespace object rollup synthesizes ahead of the deep module's body; re-confirmed still red at this head. Named imports in the same order are green; barrel-first is green.

Unreachable today, and that is the point: my stated reason for safety — *"read only inside the lazy getters"* — is **true and not sufficient**. The load-bearing invariant is **the barrel is the sole entry into the cycle**, and it now has a pin covering both routes a caller has:

- the published `exports` map has no wildcard subpath and no entry reaching the deep module (so nothing outside the package can deep-link it);
- no module anywhere in this repository imports the deep module except the barrel — a scan of 4,523 source files under `packages`, `apps`, `examples`, `scripts`, `e2e`, matched on **import specifiers** so that `scripts/measure-strict-authoring-face.mjs` does not read as an importer merely by sharing the words in its own name.

⚠️ The vitest alias table maps `@object-ui/types` to `packages/types/src` by prefix, so inside this repo a deep specifier resolves even though the `exports` map blocks it for consumers. The repo scan is the half that covers that route; the manifest assertion alone would not.

## The opaque-limit census — four numbers, and which one is the face

⚠️ **This body previously said 79. That figure was wrong, and wrong in an instructive way: it was a census of one run rather than of the face.** A count of this kind is meaningless without the method that produced it, which is how the same class of error was made twice on this card. So all four readings are given, each with its method, re-derived at `8e9b563` with an every-def-key walk written for this purpose rather than reusing the shipped walker:

| reading | method | figure |
| :--- | :--- | --: |
| the deriver, no forcing | `onOpaqueShape` over the eager walk alone; every `z.lazy` still deferred | `custom 70 · function 4 · transform 4` = **78** |
| the deriver, one document parse | eager walk plus whatever that one parse happened to force — **the old, wrong figure** | `custom 71 · function 4 · transform 4` = **79** |
| **the deriver, every lazy forced** | **what the walker reports about the whole face — the figure this body carries** | **`custom 72 · function 4 · transform 4` = 80** |
| the face itself, every def key | an independent walk that enumerates every schema-bearing `def` key, not a fixed list | `custom 74 · function 4 · transform 4` = **82** |

⇒ **80** is the census of the face as the walker sees it, and it agrees exactly with the review's independent reading.

⭐ The last row is worth keeping rather than rounding away. The two extra `custom` nodes are reached only through **`def.checks`** — they live inside a refinement, not in a schema's shape. The walker never traverses `checks`, deliberately: checks are carried across whole by the def-patching clone rather than rebuilt, which is the property the refinement pins exist for. So `onOpaqueShape`'s contract is precisely *"every unclosed shape the walker meets"*, ⛔ not *"every opaque node on the face"* — a distinction with no accept-set consequence (a preserved check still runs) but one this body should state rather than let a reader infer the stronger claim.

**Pipes.** Four reachable, not one: `transform` into `enum` (a preprocessor under `page.interfaceConfig.filterBy[]`), plus `object`, `string` and `array` each into a `transform`. The walker's own comment said "one pipe" — a reading taken through the blind guard — and commit `8e9b563` rewrites it. No OBJECT sits on an `out` side today, so walking both sides still moves no accept set; it is there so the first preprocessor wrapping an object does not open a hole.

---

# What lands

Three values and two types, published from `@object-ui/types/zod`:

| export | what it is |
| :--- | :--- |
| `StrictAnyComponentSchema` | the document-root twin of `AnyComponentSchema` |
| `StrictSchemaNodeSchema` | the child-slot twin of `SchemaNodeSchema` |
| `deriveStrictAuthoringSchema(schema, options)` | the derivation, so no consumer writes a second walker |
| `StrictAuthoringLimit`, `DeriveStrictAuthoringOptions` | types for the `onOpaqueShape` report |

Derived, never hand-written. Every reachable object is closed through unions, discriminated unions, arrays, tuples, records, intersections, optionals, nullables, defaults, **both sides of a pipe**, `z.lazy`, and — since the patch round — **callable JIT nodes**. Objects are cloned by patching a copy of their own def and calling their own constructor — ⛔ never rebuilt with a fresh object literal, which drops `def.checks` and would make the twin under-report red. A callable node clones through its own bound constructor and comes back object-typed: a difference in representation, not in behaviour, and the population pin measures behaviour.

⛔ **The rendering face's passthrough is not flipped.** This adds a face; it does not change the existing one — pinned, not promised.

## ⚠️ Re-derived; ⛔ no figure is inherited from the card body

All at `8e9b563`, zod **4.4.3**, corpus **556** documents / **2099** nodes (catalog + docs + authored). The card's 176/553 and 184/2099 are pinned at `5505aec1` and are **not** these numbers.

| reading | at `c4326fe` (base) | at head |
| :--- | --: | --: |
| documents refused, whole-tree strict | 174 / 556 | 174 / 556 |
| documents refused by the face as shipped | 46 / 556 | 46 / 556 |
| nodes refused under strict | 179 / 2099 | 179 / 2099 |
| — of those, green today, red only under strict | 127 | 127 |
| — already red under the face as shipped | 52 | 52 |
| component types declared / seen / strict-clean | 107 / 94 / 62 | 107 / 94 / 62 |
| registry collisions · arms with no literal `type` | 0 · 0 | 0 · 0 |

Base and head are identical in every cell — the corpus-scale statement that this PR moves nothing about the existing face. ⚠️ And **the same table was identical before the guard fix**, which is exactly why it cannot stand as evidence that the face is closed.

⭐ **Agreement with the prototype**, with the script's own whole-tree twin swapped for the shipped `StrictAnyComponentSchema` (scratch copy, never committed): **174 / 556**, red-today control **46** — identical.

### The blocker's property, re-read on this tree

The old 294/553 blow-up is gone: whole-tree strict (174) now sits beside per-node strict (179 of 2099) instead of an order of magnitude above it, because since #8344 a child slot resolves to the component union rather than the base keys. That is the error #7935 existed to prevent.

## Where the module lives — **ruled A** by the seat (`5590686191`), not reopened

`__tests__/zod-mirror-parity.test.ts` runs a census closed over the `export const`s of `src/zod/*.zod.ts`, and that file is fenced this round (PR #8553). So the derivation module lives at `packages/types/src/strict-authoring-face.ts`, outside the mirror directory, and the barrel re-exports from it. The collision is measured, not predicted — ablation leg 1 below. B (move it in, add an `EXCLUSIONS` row) is recorded as a follow-up for after #8553 lands.

# The pins

`packages/types/src/__tests__/strict-authoring-face-8345.test.ts` — **25 tests**.

- **(a)** a known-good `card` with `button` and `text` children parses under the strict face, and under the tolerant one.
- **(b)** one invented top-level key ⇒ refused, `unrecognized_keys` naming exactly that key; the same document accepted by the tolerant face.
- **(c)** the tolerant face is unchanged: a five-row verdict table read **after** the strict derivation is forced, plus a non-mutation reading (closed-object count and open-path count both identical before and after a twin is built and forced).
- **the population is closed** — every object reachable on the twin carries `catchall: never`; plus the callable-node control; plus a discriminator control showing the *tolerant* face is not closed.
- **REPRO-A** — an invented key deep inside a spec-derived subtree is refused and named, with the root-level control in the same document.
- **the recursion point** — an invented key on a CHILD is refused and named; a child's own `variant` / `size` are ACCEPTED.
- **checks survive the clone** — synthetic refinement still fires, with the rebuild caricature as a control; and the live `chatbot` body clause still refuses a record `body` one slot down while the root form is accepted, on both faces.
- **the limits are enumerated, not claimed** — every reported kind is one of the three recorded opaque kinds, the census is non-vacuous, and a synthetic control pins that the reporter recognises all three; plus the preprocessor pin for the `out` side of a pipe.
- **the barrel is the sole entry into the cycle** — the two halves above.

⚠️ The type-level pins are read by `tsc -p tsconfig.test.json` and by **nothing else** — vitest does not typecheck. Non-vacuity with `--listFiles`: the test program lists the pin file (1 hit), the emitting program does not (0), both list the source.

## Ablations — each proved on disk, each restored under a trap

Every leg: blob hash before and after (a byte-identical mutation aborts the leg as a no-op), an `EXIT INT TERM` trap with absolute paths, and a restore verified by the hash returning to the HEAD blob **and** `git diff HEAD` being empty. Every leg printed its RESTORED-OK line.

| leg | mutation | instrument | result |
| :--- | :--- | :--- | :--- |
| 1 · census collision | one `export const` appended to `zod/index.zod.ts` | vitest, parity file | 1 failed / 31 passed, naming `index.zod.ts#__StrictAuthoringCensusProbe` |
| 2 · rebuild caricature | object clone replaced by a rebuilt object literal | vitest | **EXIT 1** — the refinement pin |
| 3 · no `out` walk | the pipe's `out` branch deleted | vitest | **EXIT 1** — 2 failed |
| 4 · deriver return type widened | generic return replaced by the base schema type | `tsc -p tsconfig.test.json` | **EXIT 1** — TS2741 at the type pin |
| 5 · the same mutation | (as above) | vitest | **EXIT 0**, 19 passed — vitest is blind |
| **E · the guard regressed** | walker guard back to object-only | vitest | **EXIT 1** — 2 failed: the population pin, `expected [ …(6) ] to deeply equal []`, and REPRO-A |
| **F · the census blinded** | the PIN FILE's census back to object-only, walker left correct | vitest | **EXIT 1** — only the callable-node control fires; the population assertion is green *because true* (see the precision note above) |
| **E+F · both blind** | the exact state that shipped at `ad99bee` | vitest | **EXIT 1** — callable control + REPRO-A, and the population assertion green **vacuously**. ⭐ The leg that proves the new control does what the round-1 instrument could not |
| **G · manifest assertion** | the deep-module name pointed at one the exports map does reach | vitest | **EXIT 1** — the exports-map half |
| **H · a new deep importer** | a file added under `packages/core/src` importing the deep module | vitest | **EXIT 1** — the sole-entry scan; restore verified by `git status --porcelain` empty |

# Gates

All at `8e9b563`; each exit code captured before any pipe.

| gate | exit |
| :--- | :--- |
| `pnpm exec vitest run packages/types/` | **0** — 152 files, **2902** tests |
| `pnpm --filter @object-ui/types type-check` (all three programs) | **0** |
| `pnpm --filter @object-ui/types lint` | **0** — 0 errors, pre-existing warnings only, none in the changed files |
| `pnpm --filter @object-ui/types build` | **0** — 126 emitted files |
| `check:control-bytes` · `check:published-tsconfig-exclude` · `check:side-effects-array` · `check:esm-specifiers` · `check:self-import` | **0** |
| `check:dist-completeness` · `check:published-dist` | **0** |
| `check:doc-fences` · `check:doc-snippets` · `check:doc-types` · `check:doc-examples` | **0** |
| `check:governed-queue-guard` | **0**; `--test` on all five changed paths: NOT GOVERNED |
| targeted `grep -naP` control-byte self-scan of the changed files | no match |
| `check:node-esm-load`, plain | **1 — RED** |
| `check:node-esm-load --force-build` | **0 — GREEN**, `37 of 37 … built by this tree`, `34 of 39 … imported and evaluated`, 5 by design |
| CI on this head | **33 check runs, all completed, zero failures** |

⚠️ **Correction to what I wrote about that gate.** The provenance root-cause stands: turbo shares one `.turbo/cache` across every worktree of this checkout, so the plain run refuses two entries it did not build, neither of them in this diff, while `@object-ui/types` is built by this tree in both runs. **But my sentence "on CI the refusal cannot arise" implied a coverage CI does not provide.** `node-esm-load-gate.yml` runs on a nightly cron and on push to `main`; only `check:esm-specifiers` runs per PR. ⇒ the load leg is **unmeasured by anyone on a PR head** — ⛔ not "left to CI".

⚠️ **Correction to what I wrote about the carrier label.** I reported it as *"something else wrote the label set whole"*. That was an inferred mechanism, ⛔ not a reading, and the event log falsifies it: a discrete `unlabeled` event by a named account, on the **card and the PR two seconds apart** — a dual clear, which is the shape of a deliberate clear rather than a whole-set overwrite side effect. The correct instrument was one call away. Carrier state is the PM seat's to hold; ⛔ I do not touch labels.

**Not measured, stated rather than implied:** the repo-wide `turbo run lint` and the full `pnpm test` beyond CI's green run on this head; esbuild/tsup bundling of the cycle; any browser or runtime rendering (nothing renders this face); import-time cost in a bundled app (the derivation is deferred behind `z.lazy`, so it is not paid unless the face is parsed).

# Scope

Changeset: `@object-ui/types` **minor** — a new published face on a published entry point; grade confirmed against AGENTS.md 238–240 and the precedents. Five files: the derivation module, the barrel re-export block, the pin file, the package README, the changeset. ⛔ No fenced file is touched: not `packages/types/package.json` (the cycle pin **reads** it, never writes it), not the package's build configuration, not `src/record-components.ts`, not `__tests__/zod-mirror-parity.test.ts`.

Falls off the back, as ruled: wiring any consumer; repairing any of the 174; the TypeScript authoring face (#7927); closing the opaque validators.

Clause ② carrier: held and cleared by the PM seat on the verified PASS for this head. ⛔ Not mine to set, clear or reason about here. This PR stays **draft**, ⛔ not enqueued, ⛔ no auto-merge, until that seat says otherwise.

### DIFF
--- a/.changeset/8345-strict-authoring-face.md
+++ b/.changeset/8345-strict-authoring-face.md
@@ -1,0 +1,30 @@
+---
+'@object-ui/types': minor
+---
+
+Publish the derived strict authoring face from `@object-ui/types/zod`
+(objectui#8345, under the objectui#5250 ruling — maintainer 2026-09-04,
+decision batch #25, option 2: "each node schema gets a derived strict variant;
+`objectui validate` and the doc-snippet gates run strict; renderer props keep
+the tolerant face unchanged").
+
+**Additive. No existing accept set moves.** `BaseSchema` keeps its
+`.passthrough()`, every published mirror keeps the documents it accepts today,
+and no consumer in this repository is wired to the new face — wiring
+`objectui validate` and the JSON-fence gate is a separate card. What is new:
+
+- `StrictAnyComponentSchema` — the document-root twin of `AnyComponentSchema`,
+  refusing any undeclared key at any depth with an `unrecognized_keys` issue
+  that names it.
+- `StrictSchemaNodeSchema` — the child-slot twin of `SchemaNodeSchema`.
+- `deriveStrictAuthoringSchema(schema, options)` — the derivation itself, so a
+  consumer can take the strict twin of any schema on the face rather than
+  writing a second walker.
+
+The twins are **derived**, never hand-written: every reachable object is closed
+through unions, discriminated unions, arrays, tuples, records, intersections,
+optionals, nullables, defaults, both sides of a pipe, and `z.lazy`. Objects are
+cloned by patching a copy of their own def, so `.refine()` and `.superRefine()`
+checks survive — a twin rebuilt with `z.object(shape)` would drop them and
+under-report. Opaque `custom` / `function` / `transform` validators have no
+shape to close and are reported through `onOpaqueShape` rather than skipped.
--- a/packages/types/README.md
+++ b/packages/types/README.md
@@ -121,6 +121,41 @@ function renderComponent(schema: AnySchema) {
 type ButtonSchema = SchemaByType<'button'>;
 ```
 
+### The strict authoring face
+
+`@object-ui/types/zod` publishes **two** faces over the same declarations.
+
+- The **rendering face** (`AnyComponentSchema`, `SchemaNodeSchema`, every named
+  mirror) is tolerant: a node may carry keys the schema does not declare, because
+  renderer props ride through it.
+- The **strict authoring face** is a derived twin that closes every declared
+  object, at every depth. It is meant for authoring-time checking — validating a
+  document a person or an agent just wrote — where an undeclared key is far more
+  likely to be a typo than a renderer prop.
+
+```typescript
+import {
+  AnyComponentSchema,
+  StrictAnyComponentSchema,
+  deriveStrictAuthoringSchema,
+} from '@object-ui/types/zod';
+
+const document = { type: 'card', childrn: [] }; // note the typo
+
+AnyComponentSchema.safeParse(document).success;       // true  — the tolerant face
+StrictAnyComponentSchema.safeParse(document).success; // false — `unrecognized_keys: ["childrn"]`
+
+// Take the strict twin of any schema on the face:
+const StrictButton = deriveStrictAuthoringSchema(ButtonSchema);
+```
+
+The twins are derived from the mirrors, never hand-written, so they cannot drift
+from them. Strictness here is a property of the parse, not of the declaration:
+the derived schema carries the same TypeScript type as the schema it came from.
+Opaque `custom` / `function` / `transform` validators have no shape to close;
+`deriveStrictAuthoringSchema` reports each one it meets through the optional
+`onOpaqueShape` callback.
+
 ## Type Categories
 
 ### Base Types
--- a/packages/types/README.md
+++ b/packages/types/README.md
@@ -136,6 +136,7 @@ type ButtonSchema = SchemaByType<'button'>;
 ```typescript
 import {
   AnyComponentSchema,
+  ButtonSchema,
   StrictAnyComponentSchema,
   deriveStrictAuthoringSchema,
 } from '@object-ui/types/zod';
--- a/packages/types/src/__tests__/strict-authoring-face-8345.test.ts
+++ b/packages/types/src/__tests__/strict-authoring-face-8345.test.ts
@@ -1,0 +1,363 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The derived STRICT AUTHORING FACE, and the three things objectui#8345 owes as
+ * pins rather than promises (objectui#5250, maintainer 2026-09-04, decision
+ * batch #25, option 2):
+ *
+ *   (a) a known-good document parses under the strict face;
+ *   (b) a document with one invented top-level key is REFUSED, with an
+ *       `unrecognized_keys` issue NAMING that key;
+ *   (c) the tolerant face is behaviourally unchanged on the same inputs.
+ *
+ * (c) is the load-bearing one and the reason several tests below assert things
+ * that read as obvious: the card publishes a SECOND face, and the whole appetite
+ * rests on the first one not moving. "Unchanged" that nobody measures is a
+ * promise; the tables below are the measurement.
+ *
+ * ## Two further properties this file pins, because the card's risk is there
+ *
+ * **The recursion point.** Since objectui#8344 a child slot resolves its
+ * component arm to the component union rather than to the ~21 base keys. A
+ * strict twin derived over the OLD recursion point would refuse every child
+ * node's own declared props — the order-of-magnitude error objectui#7935 exists
+ * to prevent, and the reason this card was blocked behind #8344. So a child
+ * node's OWN component props being accepted is pinned as directly as the
+ * invented key being refused.
+ *
+ * **Checks survive the clone.** The walker clones by patching a copy of the
+ * def, ⛔ never by rebuilding with `z.object(shape)`. A rebuilt twin drops
+ * `def.checks`, so every `.refine()` / `.superRefine()` on the way down is lost
+ * and the face UNDER-reports red — the one failure direction that reads as good
+ * news. Both the synthetic control and the live instance are below.
+ *
+ * ## Instruments, and one that is NOT one
+ *
+ * ⚠️ `vitest` does not typecheck. The type-level pins at the bottom are read by
+ * `tsc -p tsconfig.test.json` (the third program of this package's `type-check`
+ * script) and by nothing else — a green vitest run says nothing about them.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { z } from 'zod';
+
+import {
+  AnyComponentSchema,
+  deriveStrictAuthoringSchema,
+  StrictAnyComponentSchema,
+  StrictSchemaNodeSchema,
+  type StrictAuthoringLimit,
+} from '../zod/index.zod.js';
+import { SchemaNodeSchema } from '../zod/base.zod.js';
+
+/* ── Reading a refusal ───────────────────────────────────────────────────── */
+
+type Issue = {
+  code: string;
+  path: PropertyKey[];
+  keys?: string[];
+  errors?: Issue[][];
+};
+
+/**
+ * Flatten a zod error tree. A refusal inside a union arrives as one
+ * `invalid_union` carrying a group of issue lists per arm, and a child slot is
+ * a union of six arms, so the interesting issue is never at the top level.
+ */
+function flatten(issues: readonly Issue[], out: Issue[] = []): Issue[] {
+  for (const issue of issues) {
+    if (issue.code === 'invalid_union' && issue.errors) {
+      for (const group of issue.errors) flatten(group, out);
+    } else {
+      out.push(issue);
+    }
+  }
+  return out;
+}
+
+/** Every key named by an `unrecognized_keys` issue anywhere in the refusal. */
+function refusedKeys(result: z.ZodSafeParseResult<unknown>): string[] {
+  if (result.success) return [];
+  const keys = flatten(result.error.issues as unknown as Issue[])
+    .filter((i) => i.code === 'unrecognized_keys')
+    .flatMap((i) => i.keys ?? []);
+  return [...new Set(keys)].sort();
+}
+
+/* ── Documents ───────────────────────────────────────────────────────────── */
+
+/** Declared keys only, one level of nesting, two different component types. */
+const KNOWN_GOOD = {
+  type: 'card',
+  title: 'Quarterly figures',
+  children: [
+    { type: 'button', label: 'Export', variant: 'default' },
+    { type: 'text', content: 'Updated hourly' },
+  ],
+} as const;
+
+const INVENTED_TOP_LEVEL_KEY = 'inventedTopLevelKey';
+const INVENTED_CHILD_KEY = 'inventedChildKey';
+
+const INVENTED_AT_ROOT = { ...KNOWN_GOOD, [INVENTED_TOP_LEVEL_KEY]: 1 };
+const INVENTED_AT_CHILD = {
+  type: 'card',
+  children: [{ type: 'button', label: 'Export', [INVENTED_CHILD_KEY]: 1 }],
+};
+
+describe('the strict authoring face — positive controls', () => {
+  it('the published barrel exports the derived faces and the derivation', () => {
+    expect(typeof StrictAnyComponentSchema.safeParse).toBe('function');
+    expect(typeof StrictSchemaNodeSchema.safeParse).toBe('function');
+    expect(typeof deriveStrictAuthoringSchema).toBe('function');
+  });
+
+  it('the tolerant face admits an undeclared key — the control every table below is read against', () => {
+    // If this ever goes false, the rendering face's `.passthrough()` was
+    // flipped and no assertion in this file means what it says.
+    expect(AnyComponentSchema.safeParse(INVENTED_AT_ROOT).success).toBe(true);
+    expect(AnyComponentSchema.safeParse(INVENTED_AT_CHILD).success).toBe(true);
+  });
+});
+
+describe('(a) a known-good document parses under the strict face', () => {
+  it('accepts it', () => {
+    expect(StrictAnyComponentSchema.safeParse(KNOWN_GOOD).success).toBe(true);
+  });
+
+  it('and the tolerant face accepts the same document — the two agree where nothing is invented', () => {
+    expect(AnyComponentSchema.safeParse(KNOWN_GOOD).success).toBe(true);
+  });
+});
+
+describe('(b) one invented top-level key is refused, and the key is named', () => {
+  it('refuses, with an `unrecognized_keys` issue naming exactly that key', () => {
+    const result = StrictAnyComponentSchema.safeParse(INVENTED_AT_ROOT);
+    expect(result.success).toBe(false);
+    expect(refusedKeys(result)).toEqual([INVENTED_TOP_LEVEL_KEY]);
+  });
+
+  it('the same document is accepted by the tolerant face — the difference is the whole card', () => {
+    expect(AnyComponentSchema.safeParse(INVENTED_AT_ROOT).success).toBe(true);
+  });
+});
+
+describe('(c) the tolerant face is behaviourally unchanged — a pin, not a promise', () => {
+  /**
+   * Read AFTER the strict derivation has been forced, deliberately: the walk is
+   * deferred behind `z.lazy`, so a table read before it would not be evidence
+   * about anything the derivation does.
+   */
+  it('every verdict of the tolerant face is what it was, with the strict face fully derived', () => {
+    StrictAnyComponentSchema.safeParse(KNOWN_GOOD);
+    StrictSchemaNodeSchema.safeParse(KNOWN_GOOD);
+
+    const table: Array<[label: string, input: unknown, accepted: boolean]> = [
+      ['a known-good document', KNOWN_GOOD, true],
+      ['an invented key at the root', INVENTED_AT_ROOT, true],
+      ['an invented key on a child', INVENTED_AT_CHILD, true],
+      ['a node whose `type` resolves in no arm', { type: 'no-such-component' }, false],
+      ['a node with no `type` at all', { label: 'x' }, false],
+    ];
+
+    expect(table.map(([label, input]) => [label, AnyComponentSchema.safeParse(input).success]))
+      .toEqual(table.map(([label, , accepted]) => [label, accepted]));
+  });
+
+  it('deriving mutates nothing: the tolerant graph carries exactly the closed objects it carried before', () => {
+    // `catchall: never` IS strictness. Some objects on the face are strict
+    // already — the schemas imported from `@objectstack/spec` are — so the
+    // reading that matters is that the count does not MOVE, not that it is zero.
+    const before = closedObjectCount(AnyComponentSchema);
+    const twin = deriveStrictAuthoringSchema(AnyComponentSchema);
+    twin.safeParse(KNOWN_GOOD); // force the deferred subtrees
+    expect(closedObjectCount(AnyComponentSchema)).toBe(before);
+    expect(before).toBeGreaterThan(0); // non-vacuity: the counter sees something
+  });
+});
+
+describe('a nested node is judged by its own component schema (objectui#8344 is load-bearing)', () => {
+  it('refuses an invented key on a CHILD node, naming it', () => {
+    const result = StrictAnyComponentSchema.safeParse(INVENTED_AT_CHILD);
+    expect(result.success).toBe(false);
+    expect(refusedKeys(result)).toEqual([INVENTED_CHILD_KEY]);
+  });
+
+  it("accepts a child's OWN component props — a strict twin of the pre-#8344 recursion point would refuse these", () => {
+    // `variant` and `size` are declared by `ButtonSchema` and by NOTHING in the
+    // base key set. If the child slot were judged by a strict `BaseSchemaCore`,
+    // both would come back as unrecognised — the order-of-magnitude error the
+    // blocker existed to prevent, spelled as an assertion.
+    const result = StrictAnyComponentSchema.safeParse({
+      type: 'card',
+      children: [{ type: 'button', label: 'Export', variant: 'destructive', size: 'sm' }],
+    });
+    expect(refusedKeys(result)).toEqual([]);
+    expect(result.success).toBe(true);
+  });
+});
+
+describe('checks survive the clone — a twin rebuilt with `z.object(shape)` would not', () => {
+  const refined = z.object({ a: z.number() }).refine((v) => v.a > 0, 'a must be positive');
+
+  it('the derived twin still refuses what the refinement refuses', () => {
+    expect(deriveStrictAuthoringSchema(refined).safeParse({ a: -1 }).success).toBe(false);
+    expect(deriveStrictAuthoringSchema(refined).safeParse({ a: 1 }).success).toBe(true);
+  });
+
+  it('the banned spelling loses it — this is why the walker patches the def', () => {
+    // The caricature, kept as a control so the paragraph above is a measurement.
+    expect(z.object({ a: z.number() }).safeParse({ a: -1 }).success).toBe(true);
+  });
+
+  it('the live instance: the `chatbot` body clause still fires at a child slot', () => {
+    // `defineNodeComponentUnion` installs a `superRefine` on the union that sits
+    // in the node slot, and only there. So this document is accepted at the root
+    // and refused one slot down — on BOTH faces. A twin that dropped the check
+    // would accept the nested form and this test would go red on the strict row.
+    const chatbot = { type: 'chatbot', messages: [], body: { foo: 'bar' } };
+    const nested = { type: 'card', children: [chatbot] };
+
+    expect(StrictAnyComponentSchema.safeParse(chatbot).success).toBe(true);
+    expect(StrictAnyComponentSchema.safeParse(nested).success).toBe(false);
+    // …and unchanged on the tolerant face, which is (c) again on this input.
+    expect(AnyComponentSchema.safeParse(chatbot).success).toBe(true);
+    expect(AnyComponentSchema.safeParse(nested).success).toBe(false);
+  });
+});
+
+describe('the node face (child slot) twin', () => {
+  it('admits the primitives a child slot admits', () => {
+    expect(StrictSchemaNodeSchema.safeParse('a bare string').success).toBe(true);
+    expect(StrictSchemaNodeSchema.safeParse(7).success).toBe(true);
+  });
+
+  it('closes a component in that slot', () => {
+    expect(StrictSchemaNodeSchema.safeParse({ type: 'text', content: 'x' }).success).toBe(true);
+    const result = StrictSchemaNodeSchema.safeParse({ type: 'text', content: 'x', nope: 1 });
+    expect(result.success).toBe(false);
+    expect(refusedKeys(result)).toEqual(['nope']);
+  });
+});
+
+describe('what strict could not close is enumerated, not claimed', () => {
+  /** Every def type the walker treats as opaque. Nothing else may be reported. */
+  const OPAQUE_KINDS = ['custom', 'function', 'transform'] as const;
+
+  it('every limit reported over the published face is one of the recorded opaque kinds', () => {
+    const limits: StrictAuthoringLimit[] = [];
+    const twin = deriveStrictAuthoringSchema(AnyComponentSchema, {
+      onOpaqueShape: (limit) => limits.push(limit),
+    });
+    // ⚠️ A census is only as complete as the subtrees that have been walked, and
+    // `z.lazy` defers. Forcing one document parse resolves the node slot and,
+    // through it, the rest of the graph.
+    twin.safeParse(KNOWN_GOOD);
+
+    expect(limits.length, 'a census that reports nothing is not a census').toBeGreaterThan(0);
+    const unexpected = [...new Set(limits.map((l) => l.kind))]
+      .filter((kind) => !(OPAQUE_KINDS as readonly string[]).includes(kind))
+      .sort();
+    expect(unexpected, 'the walker met a shape it could not close and this list does not name it').toEqual([]);
+    expect(limits.every((l) => l.path.startsWith('#'))).toBe(true);
+  });
+
+  it('the reporter recognises all three kinds — a synthetic positive control', () => {
+    // Which of the three the LIVE face exhibits moves with `@objectstack/spec`,
+    // so the population is asserted as a subset above and the reporter's own
+    // coverage is pinned here instead. Without this, a walker that had silently
+    // stopped recognising one kind would still pass the subset assertion.
+    const kinds: string[] = [];
+    deriveStrictAuthoringSchema(
+      z.object({
+        opaqueCustom: z.custom<string>((v) => typeof v === 'string'),
+        opaqueTransform: z.string().transform((v) => v.length),
+        opaqueFunction: z.function(),
+      }),
+      { onOpaqueShape: (limit) => kinds.push(limit.kind) },
+    );
+    expect([...new Set(kinds)].sort()).toEqual([...OPAQUE_KINDS]);
+  });
+
+  it('a `z.preprocess` has its real schema closed — the `out` side of a pipe is walked', () => {
+    // The asymmetry this guards: `X.transform(f)` keeps the schema in `in`,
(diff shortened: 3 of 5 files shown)


os-zhuang:
## Contract review (`claude-fable-5-1`, isolated seat) — PR #8642 @ `ad99bee`

**Verdict: FAIL** — one blocking defect in the exported face, independently re-measured on this head (not inherited from the card thread). The three owed pins hold; the published contract text does not.

**Ruling implemented:** objectui#5250, director comment `5534418546`, provenance **maintainer, live PM chat, 2026-09-04, decision batch #25, option 2**. Operative clause, verbatim: *"each node schema gets a derived strict variant; `objectui validate` and the doc-snippet gates run **strict**; renderer props keep the tolerant face unchanged"*, with *"⛔ The rendering face's `.passthrough()` is not flipped."* Card #8345 (priced on #7935; blocked on #8344, landed as PR #8501 `841dd2b7`) scopes this PR to the derivation + export + pins (a)(b)(c), no consumer wired. Scope as ruled: the PR adds a face, wires nothing, flips nothing — confirmed by diff (`base.zod.ts:417` still `.passthrough()`; `src/zod/` differs from merge-base `c4326fe` **only** in the barrel's appended re-export block).

**Governed paths touched: NO** (`AGENTS.md`, `CLAUDE.md`, `.claude/**`, `docs/adr/**`, `content/docs/releases/**` all absent from the 5-file diff). `packages/types/package.json` untouched; `"sideEffects": false` intact; `exports` has `./zod` and no wildcard — the deep module is unreachable by specifier.
**Clause-②: yes (new export)** — three values + two types on the published `@object-ui/types/zod` entry; mechanical floor. Carriers: `needs:contract-review` on PR and on card — present on both.

### Readings (own worktree-free build: `git archive` of `ad99bee` into scratch, sibling-install toolchain zod 4.4.3 / spec 17.3.0, `tsc -p tsconfig.json` exit 0 → 88 dist files incl. `strict-authoring-face.{js,d.ts}`; `tsc -p tsconfig.test.json` exit 0)

| reading | value |
| :--- | :--- |
| pins (a) known-good / (b) root key refused + named / (b) tolerant accepts / (c) 5-row tolerant table after forcing | true / `false` + `["inventedTop"]` / true / all 5 rows unchanged |
| child slot (recursion point live): invented child key refused + named; child's own `variant`/`size` accepted | `["inventedChild"]`; true — R4 shape, not R2 |
| `StrictSchemaNodeSchema`: bare string accepted / `text` + `nope` refused | true / false |
| corpus (`scripts/measure-strict-authoring-face.mjs --json` against my scratch dist) | **556 docs · 174 whole-tree strict · 46 red today · 2099 nodes · 179 refused (127 strict-only / 52 red) · 107 / 94 / 62 · collisions 0 · arms w/o literal `type` 0** — identical to every cell the PR quotes |
| plausibility vs card tables | card R4 = 176/553 @ `5505aec`, R2 = 294/553; this head 174/556 sits on the R4 line, the R2 blow-up is gone. Controls present (red-today, collisions, arms-without-literal). |
| opaque census (`onOpaqueShape`, one forced parse) | custom 71 · function 4 · transform 1 = 76 |
| `ZodError` on the strict path (#8498 cliff, closed by #8544) | one invented leaf key at depth 0/1/2/4/6 → flat issues 1/12/23/45/67, JSON 111 B → 9,177 B, ≤0.5 ms/parse — **linear, not the 25×/level cliff**; `cloneWithDef` preserves the `discriminator`, so #8544's fix carries into the twin |
| tolerant face non-mutation (function-aware census) | closed objects reachable from `AnyComponentSchema`: 58 before and after forcing both twins |
| module cycle, Node on dist | deep-module-first: same object as barrel's, nested key refused; barrel-first: same object, export present. No load-time derivation (`createStrictWalker()` = one empty `Map`, two `z.lazy` wrappers). New edges: `strict-authoring-face.ts` ⇄ `zod/index.zod.ts` only — **no edge from `base.zod.ts`** (the refused getter-binding route is not taken; `base.zod.ts` is byte-identical to base). |

### Findings

**1. BLOCKER — the strict twin does not close every reachable object; 6 stay open, and an off-spec key inside them is silently accepted and dropped.**
`src/strict-authoring-face.ts:141-142`: `isZodType = (value) => typeof value === 'object' && value !== null && '_zod' in value`. On this face 20 zod 4.4.3 schema nodes are `typeof 'function'` (def kinds: object 15, enum 3, record 1, pipe 1; they parse normally). The guard returns each one unwalked, so its whole subtree keeps the tolerant shape. Function-aware census of the forced `StrictAnyComponentSchema`: **302 objects, 296 closed, 6 open** — `page.interfaceConfig.sort[]` (strip), `page.slots.header[0].in.visibleWhen[1]` + its `.meta` (strip), `page.slots.header[0].in.dataSource.filter(lazy).right` (strip), `list-view.bulkActionDefs[].params[]` and `…params[].options[]` (`catchall: unknown`, i.e. passthrough).
Reproducer: `{ type:'page', interfaceConfig:{ source:'x', sort:[{ field:'a', order:'asc', inventedDeepKey:1 }] } }` → tolerant `true`, **strict `true`**, no `unrecognized_keys`, strict output `{"field":"a","order":"asc"}` (key dropped). Control, same document + `inventedTopKey` at root → strict `false`, `["inventedTopKey"]` named.
Fix ablation (scratch copy, guard admits `typeof value === 'function'`): **300 objects / 300 closed / 0 open**; reproducer → strict `false`, `["inventedDeepKey"]`; known-good still `true`; `list-view.…params[0].bogus` now named alongside the pre-existing required-value issues. The prototype's guard (`if (!schema?._zod) return schema;`, script line 260) does not have this hole — the shipped walker regressed coverage relative to the prototype on exactly this class.
*Expectation:* admit function-typed nodes in `isZodType`; add a pin that walks the forced twin with a function-aware visitor and asserts **every** reachable `object` def has `catchall.type === 'never'` (with a non-vacuity assertion that ≥1 function-typed node was visited); add the reproducer above as a pin, both directions.

**2. MAJOR — the published contract text is false at those six depths.** Changeset: *"refusing any undeclared key at any depth with an `unrecognized_keys` issue that names it"*; README: *"closes every declared object, at every depth"*; PR body: *"Every reachable object is closed"*. The changeset is the payload that ships. *Expectation:* make the text true (finding 1) — do not soften the text instead.

**3. MINOR (latent) — the walker's `default:` arm swallows container def kinds silently and reports nothing.** Synthetic probe: `z.set(obj)`, `z.map(k, obj)`, `obj.prefault(…)` → strict twin accepts `{a:'x', bogus:1}` inside them, inner objects 0 closed, and **no `onOpaqueShape` call** — so the "complete limit list = custom/function/transform" claim holds only for kinds the `switch` names. None of these kinds is on the face today (after the finding-1 fix the census reads 0 open), so no accept set moves now. Controls: `catchall: obj` and `z.record(k, obj)` are walked correctly (strict refuses). *Expectation:* add `set`/`map`/`prefault`/`promise` arms, or make `default:` report unknown *non-leaf* kinds through `onOpaqueShape` so the limit list is measured rather than enumerated by hand.

**4. MINOR — the pin file's instrument shares the defect.** `closedObjectCount` (test file, bottom) starts with `typeof node !== 'object'`, so the "identical before and after" reading and the "39 closed" figure could not see the six open objects; the corpus is blind too (no document among the 556 carries an undeclared key inside those objects — base, head and the swapped-in shipped face all read 174 regardless). *Expectation:* function-aware counter; the pin in finding 1 is the control the file is missing.

**5. INFO — the module cycle's real invariant is "the barrel is the sole entry", and nothing pins it.** "Read only inside the lazy getters" is true of the shipped source (import at line 84, runtime reads only at 300/312) and is what makes both Node orders green — but rollup's synthesized namespace on a deep-first entry is the shape #8344 paid for, and the only thing keeping it unreachable is the `exports` map plus the fact that no file but the barrel imports `strict-authoring-face`. `Build Docs` on this head is **the skip path** (12 s, "No docs-related files changed") and says nothing about Turbopack. *Expectation (non-blocking):* a source-scan pin that no module other than `zod/index.zod.ts` imports `strict-authoring-face`.

**6. INFO — `check:node-esm-load` is not a per-PR gate.** `node-esm-load-gate.yml` triggers on `schedule` (`17 4 * * *`) and on `push` for its own two paths; only `check:esm-specifiers` runs per PR. So "cannot arise on CI" is true only because CI never runs the load leg on a PR. The provenance explanation is plausible and not this diff, but the PR should state the `--force-build` re-read rather than lean on CI.

**7. INFO — placement (option A) and the census collision, reproduced.** Unmutated: pin file 19/19 and parity file 32/32 under vitest. Appending one `export const` to `zod/index.zod.ts` → `Tests 1 failed | 31 passed (32)`, failure naming `index.zod.ts#__StrictAuthoringCensusProbe`. `zod-mirror-parity.test.ts` is unchanged vs base (fence honoured); the `zod-lazy-getter-identity-7918` row lives in its own test file, also unchanged. Seat ruled A on the card; nothing here is a maintainer item.

**8. INFO — the claim comment on #8345 (`5590024076`) carries no `Container & model:` line.** Tier stated: none. (The dispatch comment says "built at `TIER_DEFAULT`"; the claim itself does not carry the line.) Commit trailers: both commits carry `Co-Authored-By` + `Claude-Session`; commit 1 carries `Part of #8345`. RULE 2: satisfied.

### Changeset verdict
`.changeset/8345-strict-authoring-face.md`: `'@object-ui/types': minor` — **grade correct** (additive export, nothing removed, no BREAKING). **Text false** at the six depths in finding 1; must be corrected with the fix, not reworded around it.

### Tests
19 pins pass; no `.skip/.only/.todo`; both directions pinned (strict refuses root and child invented keys and names them; tolerant control accepts the same documents; five-row tolerant table after forcing); revert reddens (the exports do not exist at base). What is missing is the whole-graph closure pin (finding 1/4).

### CI on `ad99bee`
33 check runs, all completed, none in progress: **Type Check** ✓ (19:33:19) · **Build Docs** ✓ (skip path, 12 s) · **Bundle Analysis** ✓ (`types (strict-authoring-face.js)` 11.72 KB / 4.48 KB gz; console eager closure 3477.4 KB under the 3512.7 KB budget) · **Test (shard 1/4 … 4/4)** all ✓ · Lint ✓ · Changeset Bump Policy / Declaration / Fixed Group / Overwrite ✓ · README Export Check ✓ · Governed Surface Queue Guard ✓. `mergeable_state`: `unknown` at read (draft; 8 commits behind `origin/main` `4dc80d0f`).

**Maintainer-only merge: no** — non-governed additive export under a maintainer-ruled programme; every finding above is implementer-floor. Carrier `needs:contract-review` stays on PR and card until finding 1 is fixed and re-reviewed.

---

os-zhuang:
## Contract review (`claude-fable-5-1`, isolated seat) — PR #8642 @ `8e9b563d80742bfd297d5fc822ae05d63a8fcf5f`

**Verdict: PASS** — the blocking defect from the `ad99bee` review is fixed at the walker and pinned over the population, re-measured here from my own build of this head (not inherited from the PR body or the card). Two owed MINOR/INFO items are **not met** (table below); neither falsifies a published sentence. One MINOR is recorded as an open latent gap for the PM seat to route (follow-up or small pre-merge patch); two are carrier/claim hygiene items outside the diff.

**Ruling implemented:** objectui#5250, director comment `5534418546`, maintainer 2026-09-04, decision batch #25, option 2 (*"each node schema gets a derived strict variant; … renderer props keep the tolerant face unchanged"*; *"⛔ The rendering face's `.passthrough()` is not flipped"*). Card #8345 scopes this PR to derivation + export + pins (a)(b)(c), no consumer wired.

**Governed paths touched: NO** — `scripts/check-governed-queue-guard.mjs --test` on all 5 changed paths: `NOT GOVERNED — 5 path(s) checked against 5 governed surface(s); none matched`. Fenced files untouched: `git diff --stat origin/main...head -- packages/types/package.json packages/types/src/record-components.ts packages/types/src/__tests__/zod-mirror-parity.test.ts` is empty; `src/zod/**` differs from merge-base `c4326fe` **only** in the barrel's appended re-export block (byte-diff read). `base.zod.ts:417` still `.passthrough()`.

**Clause-②: yes (new export).** Three values + two types land on the published `@object-ui/types/zod` entry (`StrictAnyComponentSchema`, `StrictSchemaNodeSchema`, `deriveStrictAuthoringSchema`, `StrictAuthoringLimit`, `DeriveStrictAuthoringOptions`); a consumer can observe them, so the mechanical floor applies regardless of whether any existing accept set moved — and the tolerant face's accept set did not move (corpus base = head in every cell, non-mutation pin green). Changeset `'@object-ui/types': minor`, no `major`, grade correct for an additive export; `.changeset/8345-strict-authoring-face.md` present and unchanged since `ad99bee`.

### What moved `ad99bee → 8e9b563` (2 commits, 2 files, +231/−25)
`packages/types/src/strict-authoring-face.ts` (+51: `isZodType` admits `typeof 'function'`; pipe-arm comment re-derived) and `packages/types/src/__tests__/strict-authoring-face-8345.test.ts` (+205: `closedObjectCount` → function-aware `census()`, 6 new pins: population closed, callable-node control, tolerant-face discriminator, REPRO-A, exports-map half, sole-importer scan). README, changeset, barrel: byte-identical to `ad99bee`.

### Readings (own build: `git archive` of head into scratch, sibling install read-only via symlinks — zod 4.4.3, spec 17.3.0; `tsc -p tsconfig.json` exit 0 → 88 dist files; `tsc -p tsconfig.test.json` exit 0, pin file in the test program `--listFiles` 1 hit, emit program 0 hits; vitest pin file **25/25** green, no `.skip/.only/.todo`)

| reading | value |
| :--- | :--- |
| tolerant face, function-aware census (forced) | 4766 nodes · **20 callable** (object 15, enum 3, record 1, pipe 1) · 300 objects · 58 closed / 242 open |
| **strict twin, same census** | 4998 nodes · 300 objects · **300 closed / 0 open** |
| REPRO-A `{type:'page', interfaceConfig:{source:'x', sort:[{field:'a', order:'asc', inventedDeepKey:1}]}}` | tolerant `true` · **strict `false`, `unrecognized_keys` = `["inventedDeepKey"]`** |
| control: same doc + `inventedTopKey` at root | strict `false`, `["inventedDeepKey","inventedTopKey"]` |
| control: same doc without the key | strict `true` |
| known-good `card`/`button`/`text`; root invented key | strict `true`; strict `false` `["inventedTopLevelKey"]` |
| `list-view.bulkActionDefs[].params[]` / `…options[]` with a bogus key | strict names `bogusParamKey` / `bogusOptionKey` (alongside pre-existing issues; these two were `catchall: unknown` at `ad99bee`) |
| opaque census (`onOpaqueShape`, one forced parse) | custom 71 · function 4 · **transform 4** = 79 |
| pipes reachable on the face | **4**: `transform→enum` (`…interfaceConfig/…/filterBy/…/element/operator`, the preprocessor), `object→transform` (`…slots/…/header/…/opt0`), `string→transform` (`…header/…/in/visibleWhen/…`), `array→transform` (`…exportOptions/…`) |
| **guard ablation on a dist copy** (object-only guard, walker otherwise head) | 302 objects · 296 closed · **6 open** — `…interfaceConfig/…/sort/…/element [strip]`, `…header/…/in/visibleWhen/…/opt1 [strip]`, its `…/meta/innerType [strip]`, `…in/dataSource/…/filter/…/lazy/right [strip]`, `…bulkActionDefs/…/params/…/element [unknown]`, `…/options/…/element [unknown]`; REPRO-A strict **`true`**, no keys; opaque back to 76 |
| **ablation E under vitest** (src guard reverted, blob `269c59b6→ecf52cc1`, restored to `269c59b6`) | exit 1, **2 failed / 23 passed**: population pin `expected [ …(6) ] to deeply equal []` + REPRO-A |
| **ablation F under vitest** (pin census blinded, walker correct, blob `4f24ec93→d2ab2853`, restored) | exit 1, **1 failed / 24 passed**: only *"the census can see CALLABLE schema nodes"* fires (`expected 0 to be greater than 0`); population assertion vacuously green — the control is doing its job |
| corpus `scripts/measure-strict-authoring-face.mjs --json`, base `c4326fe` vs head | **556 / 174 / 46 / 2099 / 179 = 127+52 / 107·94·62 / 0·0 — identical in every cell**; whole JSON identical except the scratch path inside one error string; script itself unchanged by the PR |
| cycle invariant, measured directly | `exports` map: 0 wildcard subpaths, 0 entries reaching `strict-authoring-face`; repo specifier scan (`packages apps examples scripts e2e`, excl. node_modules/dist): sole importer `packages/types/src/zod/index.zod.ts` |
| `default:` arm probe (owed item 3) | `z.set(obj)`, `z.map(k,obj)`, `obj.prefault(…)`, `z.promise(obj)` → inner object **0/1 closed**, strict twin **accepts** `{a:'x', bogus:1}`, **0 `onOpaqueShape` calls**; controls `z.record(k,obj)` and `.catchall(obj)` → 1/1 closed, refused |

### Findings

**1. RESOLVED (was BLOCKER) — the twin closes every object reachable on the face.** `strict-authoring-face.ts:169` guard now `value !== null && (typeof value === 'object' || typeof value === 'function') && '_zod' in value`. Population 300/300 on my build; REPRO-A refused and named; root control unchanged; guard ablation reproduces the six open objects and the silent accept exactly; ablation E turns the new population pin and REPRO-A red. The callable nodes clone back object-typed (strict census `functionTyped` = 3, all `enum` leaves), which is representation, not behaviour — the pins read behaviour.

**2. RESOLVED (was MAJOR) — changeset / README / body text is true as measured.** "refusing any undeclared key at any depth", "closes every declared object, at every depth", "Every reachable object is closed": 0 open objects on the reachable graph. Text kept, not softened — the direction the previous review asked for. (The director's optional "cite the 300/300 census in the changeset" was not taken; keeping a drifting number out of the shipped payload is defensible, and the population pin is where the number lives.)

**3. MINOR (latent) — NOT MET, and not acknowledged in the body or the os-dev-report: the `default:` arm still swallows `set` / `map` / `prefault` / `promise`.** `strict-authoring-face.ts:296-300` (`default: out = schema;` with no report). Measured above: inner objects unclosed **and** unreported, so for a consumer calling `deriveStrictAuthoringSchema` on a schema carrying one of these, "reported through `onOpaqueShape` rather than skipped" does not hold. No accept set moves today (none of these kinds is on the face: the tolerant census shows 0 such nodes, and the population pin's `census()` *does* descend `valueType`/`innerType`, so the first such container to reach the face with an object inside would turn the population pin red — the face is guarded; consumer-derived schemas are not). No pin covers the report path for these kinds (`grep -nE "z\.set\(|z\.map\(|prefault|z\.promise"` on the pin file: 0 hits). *Expectation:* add the four arms (`set`: `valueType`; `map`: `keyType`+`valueType`; `promise`/`prefault`: `innerType`) or make `default:` report unknown non-leaf kinds via `onOpaqueShape`; extend the three-kind control pin. Small enough to land here while the PR is draft; proportionate as a follow-up if the PM seat prefers — the director's re-hang precondition named items 1–4, and this one was re-hung short.

**4. RESOLVED (was MINOR) — the pin's instrument admits callable nodes.** `isSchemaNode` at the bottom of the pin file mirrors the walker's guard; `functionTyped` asserted `> 0`; ablation F shows that control is the only thing standing between a blinded census and a vacuous green.

**5. RESOLVED (INFO) — cycle invariant pinned in both halves; false evidence sentences gone.** Exports-map half and sole-importer scan are in the pin file (predicates re-measured independently above; the G/H mutations were not re-run by me). PR body: "cannot arise on CI" appears only inside a correction paragraph that retracts it; `Build Docs` green is not cited (on this head it is again the 12 s skip path, 20:40:40→20:40:52, and says nothing); `check:node-esm-load --force-build` is stated as the only reading of the load leg. The rollup deep-first RED is recorded as the reason the invariant matters.

**6. INFO — NOT MET: the re-hang comments on #8345 (`5591600708`, `5591636881`) carry no `Container & model:` line.** Tier is fused from harness stamps by the seat; this is claim hygiene only. Commit trailers on both new commits: `Part of #8345`, `Co-Authored-By`, `Claude-Session` present.

**7. INFO — NOT MET at read time: `needs:contract-review` is on the PR but NOT on the card.** PR labels at read: `documentation` · `package: types` · `tests` · `needs:contract-review`. Card #8345 labels at read (GraphQL `get_labels`, 5 total): `package: types` · `pm:dispatched` · `priority:p2` · `pm:blocking` · `domain:spec` — carrier absent; card `updated_at` 20:46:15Z, the same instant as comment `5591636881`, which states "carrier present on **both** PR and card". That sentence does not match the label set I read. ⛔ This seat touches no labels; the PM seat decides the card side.

**8. INFO — the corpus is confirmed structurally blind to finding 1, from the other side.** Base = head in every cell *and* the guard ablation moves no cell either (the pipe/opaque counts move, the corpus does not). The body says so; recorded here so nobody cites the corpus as evidence of closure.

### Owed item → status

| # | owed item (director `5591397819`) | status | evidence |
| :-- | :--- | :--- | :--- |
| 1 | BLOCKER: guard rejects callable nodes → 6 open; fix + reproducer pin | **met** | 300/300 closed on my build; REPRO-A refused, `inventedDeepKey` named; dist ablation → 302/296/6 + silent accept; ablation E → 2 failed |
| 2 | MAJOR: "every reachable object" text false at those depths | **met** | text unchanged, now true at 0 open; census not cited in changeset (optional half not taken) |
| 3 | MINOR: `default:` swallows `set`/`map`/`prefault`/`promise` — walk or report, extend control pin | **not met** | probe: 0/1 closed, accepts bogus, 0 reports; no arm added, no pin, not mentioned in body or report |
| 4 | MINOR: pin's own counter blind | **met** | `isSchemaNode` admits functions; `functionTyped > 0` control; ablation F fires only that control |
| 5 | INFO: pin "barrel is sole entry"; drop `Build Docs` / "cannot arise on CI" | **met** | two pins present; exports map 0 wildcard / 0 deep entry; sole importer = barrel; body corrected |
| 6 | INFO: `Container & model:` on the re-hang comment | **not met** | absent from `5591600708` and `5591636881` |
| 7 | re-hang `needs:contract-review` on **both** PR and card | **partly met** | PR: present. Card: **absent** at read (5 labels, none is the carrier) |

### CI on `8e9b563` (read once, 21:0x Z)
33 check runs, all `completed`, **0 failures**: 30 `success`, 3 `skipped` (`Test (coverage)`, `Test (coverage shard …)`, `dependabot`). Type Check ✓ · Test shards 1–4 ✓ · Lint ✓ · Bundle Analysis ✓ (`types (strict-authoring-face.js)` 14.27 KB / 5.47 KB gz; console eager closure 3477.5 KB under 3512.7 KB) · Changeset Bump Policy / Declaration / Fixed Group / Overwrite ✓ · Governed Surface Queue Guard ✓ · README Export Check ✓ · Build Docs ✓ (skip path, 12 s). `mergeable_state` `unknown` at read (draft; base `origin/main` has moved to `f0e72e26`, 7 commits ahead of merge-base `c4326fe`).

**Maintainer-only merge: no** — non-governed additive export under a maintainer-ruled programme; every open item above is implementer-floor or PM-seat process.

Side effects of this review, disclosed: none on any checkout. The head was read from a throwaway ref (deleted after posting); the build ran on a `git archive` extraction in scratch with the sibling `objectui-7058-main` install borrowed read-only via symlinks — vitest touched that install's ignored `node_modules/.vite/vitest` cache and left `.vite-temp` empty; its tracked tree is clean.

---

PR #8644, opened 2026-09-08
fix(plugin-form,types): resolve `form.sections[].group` through the single field-group assembler, and bound the section loop that blanked the form

### LINKED ISSUE #7051: Consume the spec section group-reference form (objectstack#13855, ruled B) — page `record:details` sections and view `form.sections` resolve `group:` keys via `deriveFieldGroupLayout`
Renderer half of objectstack#13855 (maintainer ruling 2026-08-31, verbatim 「直接处理b」): the spec gains a delta form for layout sections — a section may reference an object field group by key (shape like `group: contact_info`) and inherit membership + presentation via `deriveFieldGroupLayout` (ADR-0085 §5), instead of re-enumerating fields by hand.

## Deliverable (this repo)

The two section consumers resolve the new form:

1. the record page `record:details` sections renderer;
2. the view-level `form.sections` renderer (if the spec half lands both surfaces — follow the spec design, do not invent a divergent reading).

Resolution goes through the platform's `deriveFieldGroupLayout` single source — ⛔ no re-implementation of assembly rules (declared order, empty-group drop, ungrouped trailing bucket, collapse + visibleWhen passthrough) on the objectui side. Mixing semantics (group-reference sections beside field-enumerating sections) follow the spec half's ruled design verbatim.

## ⛔ Sequencing gate — do not start before

objectstack#13855's spec PR is merged AND objectui has adopted the `@objectstack/*` version carrying it. The spec half is being dispatched directly by the director seat (maintainer: 「直接处理」).

## Why this exists

hotcrm#806 measured the three-way field-grouping divergence this form retires: both escape hatches were wholesale + field-enumerating + derivation-free, so every hand-curated page re-copies membership and silently drifts as the object grows. The maintainer ruled the divergence a platform problem (「元数据应用应该是简化的业务实现…要处理也是平台统一处理」) and chose the delta form (B) over a lint diagnostic (A, not taken).

Refs: objectstack#13855 (ruling + design constraints) · hotcrm#806 (measured divergence, app half already queued) · objectstack#13848 (application-repo principles).

os-justin:
Fixes objectstack-ai/objectui#7051

Deliverable 2 of that card — the view-level `form.sections` half of objectstack#13855. Deliverable 1 (`record:details`) landed as #8584 and is not touched here. Base pinned at `c4326fe0a`; every reading below is against that commit, not against the moving `origin/main`.

---

## The premise held, and the two surfaces are NOT symmetric

**Re-measured, with a lit control.** Sweeping `packages/*/src/*` + `apps/*/src/*` for a section-level `group` read returns 4 hits, all of them PR #8584's `record-details.tsx`. The same command shape over the same population for `field.group` returns 9 hits across 6 files, so the instrument fires and the zero is a real zero: **no form-section renderer read `section.group`.** Premise valid.

The card's warning not to assume symmetry was the right one, and the answer is that the six form layouts are not even symmetric with **each other**. Measured by rendering an authored `{ group: 'basic' }` section on each, on the base commit:

| layout | what an on-spec `{ group }` section did | containment needed |
|---|---|---|
| `simple` | **threw** `Cannot read properties of undefined (reading 'map')` and blanked the ENTIRE form — sibling sections gone, body empty | yes |
| `tabbed` / `split` / `drawer` / `modal` | silently rendered nothing, no diagnostic; siblings fine | no |
| `wizard` | rendered an empty step ("No fields configured for this step") | no |

The throw was captured with an error boundary rather than inferred: `SimpleObjectForm (ObjectForm.tsx:1262) -> ObjectForm.tsx:1272`, i.e. the section loop **in the component's own body, above the JSX it returns**. That is the same above-the-loop shape `record:details` had, so a per-section error boundary could not have caught it either — the card's central warning, confirmed on the one layout where it applies. The other five differ because `buildSectionFields` (`sectionFields.ts`) already spells the same read `section.fields ?? []`.

## What changed

**One resolution site, above the routing fork.** `ObjectForm`'s dispatcher rebuilds each variant's `sections` prop key by key, so a `group` left in place is dropped before any container can see it, and each container would otherwise need its own copy of the resolution. Resolving in the dispatcher's own `schema` memo is what makes "one assembler" true for all six layouts at once — and the four container rows in the pin file are what makes that claim checkable rather than asserted.

**Resolution goes through the one adapter.** `deriveFieldGroupSections` (`fieldGroups.ts`) is this package's single adapter onto `deriveFieldGroupLayout` (ADR-0085 §5), and it is the same one the no-sections field-group fallback already used. Authoring a group by reference and letting the fallback derive it now produce the same section *because one of them is the other's code path* — the same construction argument #8584 made for `deriveFieldGroupDetailSections`. No assembly rule is re-implemented here.

**The load is gated.** The dispatcher fetches the object definition only when a section actually authors `group`, so a form that does not use the reference form issues no request and takes no new path. `resolveSectionGroupReferences` returns its input array **by reference** when nothing uses the form, so the memo cannot perturb an existing form either.

**Two containment layers, and only one of them is about `group`.** Layer 1 is the resolution itself. Layer 2 is `section.fields ?? []` at `ObjectForm.tsx`'s section loop, which bounds *every other* way a memberless section can reach that line — the sixth reader joining the five that already spelled it that way.

### One divergence from #8584, deliberate and measured

#8584 **drops** an unresolvable group section. Here it is **emptied** (`fields: []`) instead. Both render nothing; the difference is forced by this surface. A form whose `sections` array empties out stops being a sectioned form at all and falls back to the flat every-field layout, so dropping would make one mistyped key render **more** than the author asked for — and it would also move the wizard's step count and the container routing's condition. Reasoning is in the module docstring so the next reader does not "fix" it back.

## Where the spec's refusals are actually enforced — measured, not assumed

The card asked that group-owned presentation keys beside `group`, and `group` on a wizard step, not be silently accepted. #8584 concluded that **existence** checking belongs to `@objectstack/lint`. These two are stronger than that: they are refused at **parse**, by `@objectstack/spec` 17.3.0 itself.

| authored shape | `FormSectionSchema` / `FormViewSchema` verdict |
|---|---|
| `{ group, fields }` | REFUSED — "mutually exclusive on this form section" |
| `{}` / neither key | REFUSED — "must declare its members exactly one way" |
| `group` + `name` / `label` / `description` / `collapsible` / `collapsed` / `visibleWhen` | REFUSED, one issue per key |
| `group` + `columns` / `pane` | ACCEPTED (the two page-layout keys) |
| `group` on a `type: 'wizard'` section | REFUSED; accepted on `simple` / `tabbed` / `split` / `drawer` / `modal` |
| `group: 'Not A Key'` | REFUSED — `FIELD_GROUP_KEY_PATTERN` |

So authored metadata carrying any of them never reaches this renderer. What does reach it is a **programmatic SDUI caller**, which does not pass the spec door — and for those the renderer reports rather than accepts: a restated group-owned key is ignored (no override semantics) and warned about; a wizard `group` is refused and warned about; an unresolvable key renders nothing and is reported once, naming `form-section-group-unknown`. The parse door itself is pinned in the test file, because "enforcement lives upstream" is a claim about upstream, and a claim nothing checks is how a door quietly stops closing.

⚠️ **One premise in the dispatch order was falsified here.** It carried "aliases `fieldGroup` / `groupKey` fold to `group`". On `FormSectionSchema` they do **not** fold: they are unrecognized keys, refused with a did-you-mean hint. A renderer-side fold would therefore have been a second, more permissive contract. Pinned so the correction survives.

## The types half — and which instrument can see it

`ObjectFormSection` had the analogous omission `record:details` had (objectui#8583), but a narrower one: **exactly one missing key** (`group`) plus a `fields` that was REQUIRED, and **nothing declared that the spec refuses**. Enumerated against the installed schema, `FormSectionSchema` accepts `name, label, description, collapsible, collapsed, columns, pane, visibleWhen, visibleOn, fields, group`; the type declared all of those but `group` and `visibleOn`. (`visibleOn` is the deprecated ADR-0089 alias and its absence from the *authoring* type looks deliberate; it is left alone and reported, not widened.)

⭐ **The instrument was established with a lit control before either reading was trusted.** With `group?: string` removed from the interface — mutation proven on disk, `1 -> 0` occurrences, blob hash moved:

| instrument | verdict on the offending literal |
|---|---|
| `vitest` (2 files, 43 tests) | **GREEN, exit 0** — it strips types |
| `pnpm --filter @object-ui/types type-check` | **RED, exit 2** — `TS2353` x3, naming `group` on `ObjectFormSection` and on the `keyof ObjectFormSection` key census |

So the type-level legs are `type-check` legs, and a green `vitest` says nothing about them. Restored and proven restored (blob hash equal to HEAD's, `git diff HEAD` empty).

The same instrument caught a consequence I would otherwise have shipped: making `fields` optional broke **five** call sites in `ObjectForm.tsx` (`TS2322`, the per-layout config types are `fields`-required), and `vitest` was green on all 3736 tests at that moment.

## Ablation — run, not predicted, and it changed the tests twice

Every leg: mutation proven on disk in both directions (before/after occurrence counts plus a moved blob hash, with a void-leg abort if the blob did not move), classified per-test from vitest's **JSON reporter**, restored by `git checkout HEAD -- path` and proven restored by blob hash **and** an empty `git diff HEAD`.

| leg | reddens | stays green |
|---|---|---|
| **caricature** — resolver answers a constant group for every input | `SECTION_GROUP_RENDERS_ITS_OWN_MEMBERS`, `SECTION_GROUP_LABEL_IS_THE_GROUPS_OWN`, `UNKNOWN_GROUP_...`, all four container rows, 2 resolver units (9 red / 22) | `SIBLING_SECTION_SURVIVES`, `FORM_IS_NOT_BLANKED`, `SECTION_WITHOUT_MEMBERS_IS_BOUNDED`, every spec-door row |
| resolution removed entirely | 13 / 22, every group leg | the three containment / non-regression legs |
| layer 2 (`?? []`) reverted, resolution kept | `SECTION_WITHOUT_MEMBERS_IS_BOUNDED` **only** (1 / 22) | everything else |
| wizard refusal removed | `WIZARD_STEP_GROUP_IS_REFUSED_AND_REPORTED` only | everything else |
| `description` / `visibleWhen` carry removed | the presentation resolver unit only | everything else |

⚠️ **The stated non-regression axis does not discriminate — again.** `SIBLING_SECTION_SURVIVES` and `FORM_IS_NOT_BLANKED` both pass under the caricature, exactly as the card's correction from objectui#8497 predicted. The file says out loud which legs are which so nobody re-reads the weak one as the strong one. The discriminating leg is `SECTION_GROUP_RENDERS_ITS_OWN_MEMBERS`: **two** sections referencing **different** groups, asserted as an ordered structural outline of headings and concrete field names, so a resolver that answers the same thing for every input cannot satisfy both.

**The ablation changed the tests twice.**

1. The four container rows and the label leg originally referenced `contact_info`, the **first** declared group — and stayed **GREEN under the caricature**, because the first group is exactly what a constant resolver returns. They now reference the second group (`billing`) and go red. A leg a caricature satisfies by accident tests nothing.
2. Layer 2 was **unpinned by anything else**: with `group` resolved first, a `{ group }` section is never memberless by the time the loop sees it, so every group leg stayed green with the tolerant read reverted. `SECTION_WITHOUT_MEMBERS_IS_BOUNDED` was added specifically for it, and the third row above is the proof that it is the only leg that catches it. This is the same discovery #8584 recorded about its own two guards.

**Harness-kill leg.** Every rendered leg navigates through `requireLiveForm`, whose anchor is a section no leg asserts anything about, and whose failure text (`HARNESS DEAD: expected exactly 1 ...`) is textually unlike every content assertion. It fires in both directions — zero anchors and duplicated anchors — so a caricature cannot satisfy the harness by adding a section. No positional navigation anywhere.

## Verification

| check | result |
|---|---|
| `packages/plugin-form` + `packages/types` (final HEAD `3c57b9c2f`) | **236 files / 3737 pass, 1 skipped**, exit 0 |
| `@object-ui/types` `type-check` (3 programs) | pass, exit 0 |
| `@object-ui/plugin-form` `type-check` (2 programs) | pass, exit 0 |
| `apps/console` touched test file | 16 pass, exit 0 |
| `node scripts/check-changeset-presence.mjs` | verdict line: *"8 source file(s) of 3 released package(s) changed, and this change declares 1 changeset(s): .changeset/7051-form-section-group-reference.md."* |
| `node scripts/check-changeset-fixed.mjs` | *"All workspace packages are in the changeset fixed group."* |
| `node scripts/check-control-bytes.mjs` | *"OK (scanned 6839 tracked text file(s); skipped 85 binary)."* |
| `eslint` over the 8 changed files, `--format json` | **0 errors**, 114 warnings, all pre-existing `no-explicit-any` / `react-hooks` rules on untouched lines |

**On the eslint narrowing, so it reads as a measurement and not as a skipped run.** Population is the 8 files this PR changes, counted from the JSON reporter's own array length. The repo's `eslint.config.js` sets no `parserOptions.project` and no `projectService`, so type-aware linting is not enabled and this diff cannot move the verdict on any file it does not touch. The repo-wide run is CI's.

Everything else in the affected set is CI's: `@object-ui/types` is a root dependency, so `turbo ls --affected` names essentially the whole monorepo, and running that locally would be re-running CI.

## Fixture triage

One pre-existing file needed a prose correction rather than an assertion change: `apps/console/src/__tests__/public-block-binding-reach.test.tsx` documented its `sections` sample by asserting *"this repo's own type says the same: `ObjectFormSection.fields` is REQUIRED"* and *"`ObjectForm.tsx:1166` reads `section.fields.map(...)` off each entry unguarded"*. Both halves moved with this PR. The docblock is rewritten, not deleted: the fixture's reasoning is unchanged (a bare string is still not a section, and `{}` is still refused — the quoted spec message is re-measured against 17.3.0), only its grounds, and the two moved halves are called out explicitly so the paragraph cannot be read as still-true. That file's 16 tests pass unchanged.

## Out of scope, filed

**objectui#8641** — `apps/console`'s own `FormPage.buildSections` is a **third** form-section consumer with the same gap: `sec.fields ?? []` on a `{ group }` section yields a section with no label and no fields, silently. It does not go through `ObjectForm`'s dispatcher and the resolver here is package-internal, so fixing it is a real design choice (export the resolver / call the assembler directly / record the divergence), not a copy-paste. Filed with the discriminating axis attached, and with the honest note that it was established by reading the source rather than by a DOM measurement.

## Risk and rollback

The one behaviour change reaching a path that does **not** author `group`: `deriveFieldGroupSections` now carries the group's `description` and `visibleWhen` onto the section it derives, which the no-sections field-group fallback also uses. `description` is additive. `visibleWhen` is not purely additive — an object that declares a `fieldGroups[].visibleWhen` and relies on the fallback would now have that predicate honoured, and a predicate resolving false hides the section. That is the ADR-0089 contract and the key the spec names as group-owned, so carrying it is the correct direction; it is called out because it is the only line in this PR that can change an existing form. Rollback is the two added lines in `fieldGroups.ts`, and the resolver unit leg pins them.

Everything else is gated on an authored `group`: no such section, no fetch, same array reference out of the resolver, no new path.

---

### DIFF
--- a/.changeset/7051-form-section-group-reference.md
+++ b/.changeset/7051-form-section-group-reference.md
@@ -1,0 +1,48 @@
+---
+'@object-ui/plugin-form': minor
+'@object-ui/types': minor
+---
+
+A form-view section can reference a declared field group instead of copying its
+members (objectstack#13855, objectui#7051 — the view-level half; the
+`record:details` half shipped as objectui#8497).
+
+`@objectstack/spec` 17.3.0 declares two ways for a `form.sections[]` entry to give
+itself members: enumerate `fields`, or point `group` at one of the object's declared
+`fieldGroups` and inherit that group's membership **and** its presentation. Nothing on
+this renderer read `group`, and the omission was not a no-op:
+
+- on the default `simple` layout an authored `{ group: 'contact_info' }` threw
+  `Cannot read properties of undefined (reading 'map')` out of the section loop in
+  `SimpleObjectForm`'s own body — above the JSX it returns, so no per-section error
+  boundary could contain it — and **blanked the entire form**, taking every
+  well-formed sibling section with it;
+- on `tabbed` / `split` / `drawer` / `modal` the section silently rendered nothing;
+- on `wizard` it rendered an empty step.
+
+`ObjectForm` now resolves the reference **once**, above its routing fork, so all six
+layouts inherit it: a `{ group }` section renders the members that group declares, in
+the order and with the label, description and collapse state the object declares them
+with. Resolution goes through `deriveFieldGroupLayout` (ADR-0085 §5) via this
+package's existing single adapter — the same code path the no-sections field-group
+fallback already used, so authoring a group by reference and letting the fallback
+derive it produce the same section by construction. No assembly rule is
+re-implemented here.
+
+The object definition is fetched only when a section actually authors `group`, so a
+form that does not use the reference form issues no additional request and takes no
+new path.
+
+Diagnostics rather than silence, for the shapes the spec door cannot see (programmatic
+SDUI callers): a `group` naming no declared group renders nothing and is reported once
+on the console (`@objectstack/lint` owns it as `form-section-group-unknown`); a
+group-owned presentation key restated beside `group` is ignored — the spec grants no
+override semantics — and reported; `group` on a wizard step is refused and reported,
+because a step has no slot for the `collapse` / `visibleWhen` a group carries.
+
+Also in this change: `@object-ui/types`' `ObjectFormSection` declares `group` and makes
+`fields` optional, so the spec-legal shape finally compiles for a TypeScript author;
+and the shared field-group adapter now carries the group's `description` and
+`visibleWhen` onto the section it derives — both are keys the assembler emits and
+`ObjectFormSection` declares, and a key-by-key rebuild that drops one is how a declared
+group's presentation goes missing.
--- a/apps/console/src/__tests__/public-block-binding-reach.test.tsx
+++ b/apps/console/src/__tests__/public-block-binding-reach.test.tsx
@@ -292,16 +292,26 @@ const SUPERSEDES_BINDING = new Set(['data', 'staticData', 'customFields']);
  * generic `array` sample is `['name']`, and a bare string is not a section:
  * `@objectstack/spec`'s `FormViewSchema.sections` rejects it at parse —
  * `safeParse(['name'])` returns *"Invalid input: expected object, received
- * string"* — and requires `fields` on every entry (`[{}]` returns *"0.fields:
- * Invalid input: expected array, received undefined"*). This repo's own type
- * says the same: `ObjectFormSection.fields` is REQUIRED
- * (`packages/types/src/objectql.ts`). So `['name']` is not metadata any author
- * could publish, while `ObjectForm.tsx:1166` reads `section.fields.map(...)` off
- * each entry unguarded and takes the whole block down with *"Cannot read
+ * string"* — and requires every entry to declare its members (`[{}]` returns
+ * *"0.fields: A section must declare its members exactly one way, and this form
+ * section declares neither"*). So `['name']` is not metadata any author could
+ * publish, while `ObjectForm.tsx`'s section loop read `section.fields.map(...)`
+ * off each entry unguarded and took the whole block down with *"Cannot read
  * properties of undefined (reading 'map')"*. That makes the error card a
  * FIXTURE defect, not the product bug the shape suggested — the discriminator
  * being the spec shape, not the fact that a different sample stops the crash
  * (which is true either way).
+ *
+ * ⚠️ Two halves of that reasoning moved with objectui#7051 and are restated
+ * here rather than left to read as still-true. (1) `ObjectFormSection.fields`
+ * is no longer REQUIRED on this repo's type: `group` — the field-group
+ * REFERENCE form, `@objectstack/spec` 17.3.0 / objectstack#13855 — is the other
+ * way a section declares the same fact, so the spec refuses an entry carrying
+ * NEITHER rather than one missing `fields`. The fixture reasoning is unchanged
+ * (a bare string is still not a section, and `{}` is still refused), only its
+ * quoted grounds. (2) That unguarded `.map` is now spelled `?? []`, so this
+ * sample's shape no longer decides whether the block survives — which is
+ * exactly why the sample is justified on the SPEC shape and not on the crash.
  *
  * `formType` is the SEVENTH, and it is why `object-master-detail-form` read GREEN
  * while carrying the identical latent crash. That block declares `formType` as a
--- a/packages/plugin-form/src/ObjectForm.tsx
+++ b/packages/plugin-form/src/ObjectForm.tsx
@@ -42,6 +42,7 @@ import {
   inferColumns,
 } from './autoLayout';
 import { deriveFieldGroupSections } from './fieldGroups';
+import { hasSectionGroupReference, resolveSectionGroupReferences } from './sectionGroups';
 import { sanitizeFormData } from './sanitize';
 import { noSubmitTargetError } from './submitTarget';
 import {
@@ -137,6 +138,43 @@ export const ObjectForm: React.FC<ObjectFormComponentProps> = ({
   dataSource,
 }) => {
   const perms = usePermissions();
+
+  // ── `sections[].group` — the ADR-0085 §5 REFERENCE form (objectui#7051) ───
+  //
+  // A section may point `group` at one of the object's declared `fieldGroups`
+  // instead of enumerating members (`@objectstack/spec` 17.3.0,
+  // objectstack#13855). Resolving it needs the OBJECT definition, and this
+  // dispatcher — unlike the six bodies below it — has never loaded one.
+  //
+  // It is resolved HERE, above the routing, on purpose: every variant's
+  // `sections` prop is rebuilt key by key in the maps below, so a `group` left
+  // in place would be dropped before any container could see it, and each
+  // container would need its own copy of the resolution. One site above the
+  // fork is what makes "one assembler" true for all six layouts at once —
+  // including `simple`, whose section loop read `section.fields.map(...)`
+  // unguarded and threw `Cannot read properties of undefined (reading 'map')`
+  // on a spec-legal `{ group }` section, blanking the WHOLE form (measured on
+  // this file at `origin/main`; the other five silently dropped the section
+  // instead, because `buildSectionFields` spells the same read `?? []`).
+  //
+  // The load is gated on an authored `group` being present, so a form that
+  // does not use the reference form issues no request and takes no new path.
+  const needsGroupLayout = hasSectionGroupReference((rawSchema as any)?.sections);
+  const canResolveGroups = typeof (dataSource as any)?.getObjectSchema === 'function';
+  const [groupObjectDef, setGroupObjectDef] = useState<any>(null);
+  useEffect(() => {
+    if (!needsGroupLayout || !canResolveGroups) return;
+    let alive = true;
+    Promise.resolve((dataSource as any).getObjectSchema(rawSchema.objectName))
+      .then((def: any) => { if (alive) setGroupObjectDef(def ?? null); })
+      // A failed load leaves the references unresolved (their sections render
+      // empty). It is deliberately NOT reported as a dangling reference: the
+      // body below owns the load error for the object, and claiming the group
+      // does not exist would be a second, wrong diagnosis of one failure.
+      .catch(() => { if (alive) setGroupObjectDef(null); });
+    return () => { alive = false; };
+  }, [needsGroupLayout, canResolveGroups, dataSource, rawSchema.objectName]);
+
   // Apply field-level permissions to the entire schema (sections + flat
   // fields) BEFORE dispatching to any variant. This way all variants
   // (Tabbed/Wizard/Split/Drawer/Modal/Simple) transparently honour FLS.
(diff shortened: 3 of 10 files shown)


os-justin:
## CI repair: `Test (shard 3/4)` — two doc-example ledger rows re-keyed

`scripts/__tests__/check-doc-example-types.test.ts:264` (`the real ledger > every row names a block that is actually in the compiled tier`) was the only red on `3c57b9c2f`; the other 34 checks were green.

### What the key is

`ledgerKey` in `scripts/check-doc-example-types.mjs:1051` is:

```js
export function ledgerKey(block) {
  return `${block.file}:${block.line} ${block.symbol}`;
}
```

`block.line` is the 1-based line of the `@example` JSDoc tag. So a row is invalidated by any edit that moves that tag, whether or not the example itself changed.

### What actually moved

CI printed one key, but the loop throws on the first failure — a census probe on this head found **two** stale rows, so fixing only the printed one would have traded this red for the next:

| ledger row | was | now | why |
|---|---|---|---|
| `packages/plugin-form/src/ObjectForm.tsx ObjectForm` | 122 | 123 | the added `./sectionGroups` import sits above the JSDoc |
| `packages/types/src/objectql.ts ObjectFormSchema` | 1572 | 1604 | the schema additions sit above the JSDoc |

Both blocks **still exist and are still in the compiled tier** — this is a move, not a deletion:

```
present: "packages/plugin-form/src/ObjectForm.tsx:123 ObjectForm" symbol=ObjectForm pkg=@object-ui/plugin-form
present: "packages/types/src/objectql.ts:1604 ObjectFormSchema"
```

Both example bodies are **byte-identical** to the merge-base (`diff` of the two regions is empty), so each row's recorded `codes` and written `reason` stay accurate — no row was re-derived, only re-addressed.

### `main` is ruled out as the cause

The `@example` tag sits at its original line on both the merge-base and `origin/main`, and only moved on this branch:

```
c4326fe0a (merge-base)  ObjectForm.tsx:122   objectql.ts:1572
da5e4f69e (origin/main) ObjectForm.tsx:122   objectql.ts:1572
3c57b9c2f (this branch) ObjectForm.tsx:123   objectql.ts:1604
```

`main` has not touched `scripts/check-doc-example-types.mjs` since the branch point. Of `main`'s 122 changed files, exactly one carries a ledger row (`packages/core/src/data-scope/DataScopeManager.ts`), and its tag is unmoved at line 68 on all three refs — so merging `main` in would neither fix nor break anything here. **No merge was needed; the repair is the re-key alone.**

### Verification

Reproduced the red first, then showed the same check green:

```
before  Test Files 1 failed (1)   Tests 1 failed | 45 passed (46)
        AssertionError: packages/plugin-form/src/ObjectForm.tsx:122 ObjectForm: expected false to be true
after   Test Files 1 passed (1)   Tests 46 passed (46)
```

Neighbourhood re-run so one red is not traded for another:

- `pnpm exec vitest run scripts/` — 125 passed | 2 skipped (127 files), 3684 tests
- `pnpm exec vitest run packages/plugin-form/ packages/types/` — 236 passed, 3737 tests
- `node scripts/check-doc-example-shared-reader.mjs` — `OK  80 documented symbol(s), 3989 call site(s), 11 slot(s) ...`
- `node scripts/check-changeset-presence.mjs` — green on the existing changeset; the gate script is not published source, so no new changeset is owed

### On having to edit a CI gate script from a `plugin-form` PR

This is exactly the tax **objectui#8614** describes: `UNGATED_EXAMPLES` is keyed by line number, so a one-line import added above a JSDoc block forces an unrelated PR to edit a CI gate script to stay green. Paying it here because this PR has to land; **objectui#8614 remains the standing card for the keying mechanism** and is deliberately not addressed in this diff — changing how the ledger is keyed carries its own design question and would widen a `plugin-form` PR into gate-infrastructure work.

Diff of this repair is two lines, both ledger keys.

---

PR #8656, opened 2026-09-08
test(core): widen the record-title step-numbering pin to its sibling test file

### LINKED ISSUE #6843: finding(core): record-title.test.ts still calls the record-key probe "step 3b"
`packages/core/src/utils/__tests__/record-title.test.ts:133` carries the same stale
precedence number that objectui#6634 corrected in the module it tests.

## The site

```
packages/core/src/utils/__tests__/record-title.test.ts:132-133

    // No fields to derive from, but the record carries a standard `name` → the
    // record-level name probe (step 3b) resolves it (better than `Record #7`).
```

## Why `3b` is wrong here too

Same rung, same argument as objectui#6634. `record-title.ts` states its ADR-0079 ladder
four times over and every one of them numbers the type-aware derivation from
`objectDef.fields` as step `4`, with the `NAME_ISH_RECORD_KEYS` probe beneath it as `4b`:
the module docblock (`:37`), the resolver docblock (`:380` / `:381`), the
`deriveTitleField` docblock (`:279`, *"precedence step 4"*), and the branch comments
themselves (`:437` / `:442`). Step `3` is `objectDef.titleFormat` — a rendered template,
not a record-key probe.

This same test file already spells it `4b` twice, at `:150` (*"so step 4b answered it
identically"*) and at `:245` (*"name-ish by NEITHER step-4b rung"*). So the file
contradicts itself seventeen lines apart, which is exactly the shape objectui#6634
recorded for the module.

## Why it was not folded into that card

objectui#6634's fence is `packages/core/src/utils/record-title.ts` alone, and the card
asked explicitly for other citing sites to be reported rather than renumbered. PR #6842
fixes the module and lists this line as the one citing site that disagrees. Filed
unassigned rather than widened into that PR.

## The natural shape of the fix

PR #6842 adds `packages/core/src/utils/__tests__/record-title.stepNumbering.test.ts`, a
source-reading pin that checks every sub-step label in `record-title.ts` against the
number the file's own ladder gives the rung it sits under. That pin deliberately scans
only `record-title.ts`, which is why it is green today. Widening its scanned set to this
sibling test file would both fix and mechanically hold this line — one edit plus one path
in the pin's scan list.

Observation only. No behaviour change: the number lives in comment text and no assertion
reads it.

## Duplicate scan before filing

The repo-scoped REST read channel is live in this container (authenticated
`GET /repos/objectstack-ai/objectui/issues/6634` returned 200), so this went through REST
plus a local grep rather than a search endpoint. objectui#6634's own filing-time targeted
MCP `search_issues` already covered this file's step numbering and returned 8 hits with
objectui#6572 as its positive control; treating that as the established reading, only the
increment since was re-read here —
`GET /repos/.../issues?state=open&since=2026-08-28T06:54:38Z&per_page=100&sort=created`,
100 rows, locally grepped for `record-title` / `precedence step` / `3b` / `4b` /
`deriveFromRecordKeys`. The only issue in that set about this ladder is objectui#6634
itself; the other matches are unrelated PRs whose bodies happen to contain those short
strings. Nothing covers the test file's citation. Note the bound honestly: that listing
was capped at 100 rows, so it is a bounded increment read, not an exhaustive one.

Refs: objectui#6634 · PR #6842 · objectui#6572

os-justin:
Fixes #6843

## What changed

- `packages/core/src/utils/__tests__/record-title.test.ts:133` — the comment on the record-level name probe said "step 3b"; it now says "step 4b". That is the number `record-title.ts` gives the rung everywhere it states its ADR-0079 ladder (step 3 is `objectDef.titleFormat`, a rendered template; step 4 is the `objectDef.fields` derivation with `NAME_ISH_RECORD_KEYS` beneath it as 4b), and the number the same file already used at three other sites (`:150`, `:245`, `:252`).
- `packages/core/src/utils/__tests__/record-title.stepNumbering.test.ts` — the pin from PR #6842 scanned `record-title.ts` only, so this drift sat in its declared blind spot. Its scanned set is now an explicit list, `SCAN_SET`, naming `record-title.ts` and `__tests__/record-title.test.ts`, and:
  - a vacuity guard requires the set to be non-empty, to contain both files by name, both present on disk and non-empty, and to exclude this pin itself (its docblock spells `3b` as the bug it describes);
  - every scanned file declares a label floor in `CONTROL_LABEL_COUNTS`, typed over the set so a new entry without a floor is a compile error; the cross-set floor must be greater than 0;
  - the collector also reads the hyphenated `step-4b` spelling the sibling uses, and reports file plus line for every stray label;
  - a fixture self-test proves the collector and the verdict helper (`strayLabels`) depend on their input — the fixture carries the original uncorrected `:133` text — and that `it('0b. …')` test names stay out of the population;
  - one absolute pin keeps `titleFormat` on step 3 and the fields derivation on step 4 — the numbers the hand-citing sites outside the module carry (`InterfaceListPage` prose, the #6530 changeset, this sibling test). Everything else stays relative to the ladder, so a deliberate renumbering edits two constants on purpose instead of silently reconciling a stray label by moving the ladder.
- `.changeset/6843-record-title-step-numbering-scan.md` — empty frontmatter; comment text and tests only, nothing is released.

## Measured before widening

Of the 49 files under `packages/core/src/utils/__tests__/`, exactly one besides the pin cites the ladder: `record-title.test.ts` (`step 3b` once, `step 4b` twice, `step-4b` once). No other file enters the population, so this is not a bigger card. Repo-wide `git grep 'step 3b'` over `packages/*/src/*` and `apps/*` now hits only the pin's own fixture and expectation lines (the lit control for the sweep); `step 4b` / `step-4b` hits 10 lines.

## Evidence (all at `e64f3af96`, vitest root-relative form, runs under the shared verify lock)

Red on the uncorrected file, before the one-line edit: the widened pin failed with `sub-step labels disagreeing with the ladder's 4b` listing `__tests__/record-title.test.ts:133 "step 3b" says 3` — 1 failed, 6 passed. After the edit: 7 passed.

Ablation legs — each mutation proven on disk by `grep -c` before the run, each restored with `git checkout HEAD -- ABSOLUTE_PATH` under a trap and proven by `git hash-object` equal to the `HEAD` blob plus an empty `git diff HEAD`:

- V1 empty `SCAN_SET` — RED: guard `expected 0 to be greater than or equal to 2`; cross-set floor `expected 0 to be greater than 0` (2 failed, 5 passed — the per-file loop is vacuous over an empty set, which is exactly why the guard exists).
- V2 mis-pathed entry (`record-titel`) — RED: guard `to include '__tests__/record-title.test.ts'`, plus ENOENT in the two reading tests (3 failed).
- S re-introduce `3b` at `:133` — RED naming `:133` (1 failed).
- R1 the plausible wrong "fix": renumber the derivation 4 to 3 consistently in module and sibling — every relative check stays green; only the absolute pin fails: `fields-derivation sites not on step 4` listing 3 sites (1 failed).
- R2 renumber `titleFormat` 3 to 2 in the module — only the absolute pin fails: `titleFormat sites not on step 3` listing 4 sites (1 failed).
- C1 extractor returns constant 4 — only the fixture self-test fails (`[[2,4],[3,4],[4,4],…]` vs `[[2,3],[3,3],[4,3],…]`).
- C2 verdict constant (`strayLabels` returns `[]`) — only the fixture self-test fails.
- C3 scan collects nothing — per-file control (`record-title.ts: 0 sub-step label(s) found, control floor is 3`), cross-set floor and self-test fail (3 failed).

No void legs. No `dist/` is involved: the pin reads source through `node:fs`, so no build step sits between a mutation and its measurement.

Leg D, JSON reporter over `pnpm exec vitest run packages/core/`: 132 files, 701 suites, 2792 tests, all passed; `record-title.stepNumbering.test.ts` 7/7 passed, `record-title.test.ts` 75/75 passed; no death strings.

`pnpm --filter @object-ui/core type-check` (script name echoed; `@object-ui/types` built first): 0 TS errors, and `tsc -p tsconfig.test.json --listFiles` reads all three `record-title*` files. ESLint on both test files: 0 errors (2 pre-existing `no-explicit-any` warnings at `record-title.test.ts:311` and `:362`, lines not touched here).

`node scripts/check-changeset-presence.mjs`: "2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s) … Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate."

`node scripts/check-governed-queue-guard.mjs --test PATHS`: "NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched."

Placeholders in this body are written as upper-case words (ABSOLUTE_PATH, PATHS) because the platform strips tag-shaped text from bodies.

---


os-justin:
## The red is not this PR's — filed as objectui#8688, branch updated to re-run

`Test (shard 2/4)` failed at 20:15Z. Everything else on that head is green (33 of 34), and the shard's only failure is:

```
FAIL  dom  packages/permissions/src/__tests__/providerCtxIdentity.discarded.test.tsx
  > MePermissionsProvider — the fetch effect survives a discarded cache (objectui#6862)
TypeError: Cannot read properties of undefined (reading 'userId')
 ❯ providerCtxIdentity.discarded.test.tsx:619:41
```

**This PR's diff is `packages/core` record-title step-numbering, test-only. It cannot reach `packages/permissions`.** So the failure is not this change's, and ⛔ porting a repair into it would widen a two-file test PR into a second package.

## What it actually is, so nobody re-runs it away

```
618|  await waitFor(() => expect(log).toHaveLength(1));
619|  expect(ctxSeen[ctxSeen.length - 1]!.userId).toBe('A');
```

⭐ **The wait is on `log`; the read is on `ctxSeen` — a different array.** Nothing establishes that `ctxSeen` has been filled when `log` reaches 1, so `ctxSeen[-1]` is `undefined` and the `!` makes the compiler agree while the runtime throws.

⚠️ **⛔ Not a flake in the sense that permits a re-run as a fix.** The wait genuinely does not gate the read; a green run shows the scheduling came out favourably, not that the test is sound. That is why it is filed (objectui#8688) rather than shrugged at — same family as objectui#8665's seven `contractEnvelope-6839` siblings, different package and a blunter mechanism.

## What I did here

**Updated this branch** so CI re-runs against current `main` — the one re-run this PR is entitled to, and it also picks up everything that has landed since. If the same file reddens again, that is objectui#8688's to fix and this PR should not be held behind it; ⛔ but nothing is being skipped, disabled or waved through in the meantime.

Auto-merge stays armed, so a green head lands it.

---

PR #8664, opened 2026-09-08
test(plugin-tree): split the contractEnvelope-6839 waits by expected outcome

os-justin:
Part of objectui#6839. CI repair — this unblocks PR #8553, whose `packages/types` diff cannot reach `plugin-tree` at all.

No card is claimed here and no assignee is set.

## The failure

```
packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx:124
AssertionError: the declared rows member must still draw: expected 1 to be 2
```

## The race, diagnosed here rather than ported

The file handed all three cases **one** wait:

```js
await waitFor(() =>
  expect(
    container.querySelector('[data-testid="object-tree"]') ?? screen.queryByText('No records'),
  ).not.toBeNull(),
);
return container.querySelector('[data-testid="object-tree"]')
  ? container.querySelectorAll('tbody tr').length
  : 'empty-state';
```

That testid is on the **table wrapper**, so it is a MOUNT signal — but the rows this file counts arrive a commit **later**. In `ObjectTree`, expansion is a state MIRROR:

```js
const [expanded, setExpanded] = useState<Set<string>>(new Set());
useEffect(() => {
  setExpanded(initialExpanded(roots, config.defaultExpandedDepth));
}, [roots, config.defaultExpandedDepth]);
const visibleRows = useMemo(() => flattenVisible(roots, expanded), [roots, expanded]);
```

So when `find()`'s rows land, the commit that first paints the table still carries the **previous (empty)** mirror: the root draws, its child does not, `tbody tr` is 1. The passive effect then seeds the mirror and a second commit takes it to 2.

**Measured, not assumed.** A probe recorded every DOM state this fixture passes through, and evaluated both waits at each:

| DOM state | old wait passes? | old wait yields | new wait passes? |
|---|---|---|---|
| `loading` | no | – | no |
| `table:1rows` | **yes** | **1** | no |
| `table:2rows` | yes | 2 | yes |

The old wait's *first passing state* is `table:1rows`, where it yields 1 — the CI red, by construction rather than by luck. Which commit the read lands on is decided by machine load, which is why it was green locally and red on a saturated shard.

## ⚠️ It is NOT the kanban twin's shape, although it is the same family

PR #8533 fixed `plugin-kanban` against **two** races — `React.lazy(() => import('./KanbanImpl'))` chunk reveal **and** a prop-mirrored `boardColumns` — with the symptom `expected +0 to be 2`, nothing drawn.

`plugin-tree` has **no lazy boundary anywhere in its source** (measured: `index.tsx` imports `./ObjectTree` eagerly; zero `React.lazy` / dynamic `import()` in the package's non-test sources). Only the mirrored-state half transfers, and the symptom is a **partial** draw — the root without its child, `expected 1 to be 2`. #8533's diff shape was read first and then re-derived against this file's actual mechanism rather than applied to it.

## What changed

Test file only. No timeout was raised, no assertion loosened, nothing skipped.

- **Positive arms** wait for the **descendant row** — the row the mirror gates — then assert the drawn **shape** (label + depth per row) plus the root toggle reading `Collapse`. The pin is the seeded-open hierarchy, not an eventual count: a count alone is also satisfied by a tree that flattens everything, and by a fix that merely waits longer.
- **Refusal arm** cannot wait for an absence, so it takes a settled read anchored on something that *does* appear in that scenario: the tree's own **"No records" panel**, which `ObjectTree` renders only after `loading` flips false. Probed on the `records` fixture the sequence is `loading → empty-state`, with no table at any point and the panel absent while loading — so this arm cannot pass by timing out, which is the failure mode every absence-shaped pin has.

## Evidence

**Deterministic reproduction.** Pushing the mirror's seed past the table's own commit (`setTimeout(…, 50)` on `setExpanded`, mutation proven on disk by anchor/marker counts and by `git hash-object` differing from the HEAD blob, restored to an empty `git diff HEAD`):

| file | result |
|---|---|
| pre-fix | `AssertionError: the declared rows member must still draw: expected 1 to be 2` ×2 arms |
| this PR | 3 passed |

An earlier leg with `setTimeout(…, 0)` landed on disk but did **not** redden — RTL's `asyncWrapper` drains one macrotask before returning, so the gap was hidden. Reported as fired-but-negative, not as void.

**Unmutated flake caught in-container.** Under a 6-way CPU load on a 4-core box, the pre-fix file (materialised from the parent commit) failed **1 of 12** runs with the exact CI text — same sentence, same `expected 1 to be 2`, same line 124. The fixed file: **20 of 20** green under the same load, and **20 of 20** idle.

**Ablations** (per-test classification from vitest's JSON reporter; every mutation proven on disk, every restore proven by an empty `git diff HEAD`):

| leg | mutation | result |
|---|---|---|
| B — caricature | `extractRecords` returns a **constant** row set for every input | refusal arm RED, positives green |
| C — caricature | `extractRecords` returns `[]` for every input | both positives RED, refusal green |
| D — non-regression | `extractRecords` reads `records` too, so the **wrong envelope draws** | refusal arm RED |
| E — wait non-vacuity | the mirror is **never** seeded | both positives RED (`[{Root,0}]` vs `[{Root,0},{Child,1}]`) |

Legs B and C are the caricature in both directions; D is the card's own subject — the pin still distinguishes the declared `data` member from the undeclared one.

## Family sweep

`git ls-files | grep contractEnvelope-6839` → **11** files. Control fired: both files known in advance (kanban, fixed; tree, broken) are in the result, so the reading is not an empty instrument.

Two are pure unit tests with no DOM (`packages/core/.../extract-records`, `packages/react/.../nonGridRowCeiling`) — not exposed. Of the nine DOM mounts, kanban is fixed and tree is this PR. The remaining **seven are structurally exposed** — a single shared wait on a mount/settle signal, then a count — and several say so in their own comments:

| file | shared wait | note |
|---|---|---|
| `plugin-calendar/ObjectCalendar` | `queryByTestId('calendar-view')` | comment already calls it "a mount signal rather than a rows signal" |
| `plugin-charts/ObjectChart` | `lastSchema ?? queryByTestId('chart-empty-state')` | this package **does** have a lazy boundary — closest kin to kanban |
| `plugin-dashboard/ObjectDataTable` | `queryByTestId('rows') ?? queryByTestId('table-empty-state')` | same two-branch shape this PR replaces |
| `plugin-dashboard/ObjectPivotTable` | `queryByTestId('pivot')` | sharpest: the refusal arm expects `0` on the **same node**, so "drawn empty" and "drawn before the data" are one reading |
| `plugin-map/ObjectMap` | `queryByText('Loading map...')` is null | comment already calls it "a settle signal rather than a rows signal" |
| `plugin-timeline/ObjectTimeline` | `getByTestId('timeline-renderer').getAttribute('data-item-count')` not null | satisfied the instant the renderer mounts, `"0"` included |
| `plugin-gantt/ObjectGantt` | positives already wait **for** the bars | inverse gap: the refusal arm reads `bars === 0` right after a wait on `find` having been *called*, with no completion anchor |

**Measured, and the zero is reported with its sensitivity:** all nine family files, five iterations under the same 6-way load — **zero failures**. That instrument is too weak to clear them: the pre-fix tree file's own observed rate was ~1 in 12, and five runs would miss an 8%-per-run flake roughly two thirds of the time.

⇒ **Listed, not fixed here.** Each needs its own component-level diagnosis the way this one got it; applying this diff shape to a race nobody has measured is the thing #8533's own write-up warns against.

## Gates

- `node scripts/check-changeset-presence.mjs` → `✅  1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/tree-contract-envelope-waits-6839.md.` / `Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.`
- `node scripts/check-changeset-no-major.mjs` → `✅  No changeset declares a `major` bump.`
- `pnpm exec eslint packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx` → exit 0, 0 errors, 3 pre-existing `no-explicit-any` warnings (same three `any`s the file already carried).
- `pnpm --filter @object-ui/plugin-tree type-check` → exit 0, **MEASURED**: the closure was built first (`pnpm --filter '@object-ui/plugin-tree^...' build`), and `tsc -p tsconfig.test.json --listFiles` shows the changed file in the program.
- `pnpm exec vitest run packages/plugin-tree/` → `Test Files 12 passed (12)` / `Tests 54 passed (54)`.

---

### DIFF
--- a/.changeset/tree-contract-envelope-waits-6839.md
+++ b/.changeset/tree-contract-envelope-waits-6839.md
@@ -1,0 +1,6 @@
+---
+---
+
+Test-only change in `@object-ui/plugin-tree`: the `contractEnvelope-6839` pin now
+splits its waits by expected outcome instead of reading the row count the instant
+the tree-grid wrapper mounts. No published behaviour changes.
--- a/packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx
+++ b/packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx
@@ -28,10 +28,64 @@
  * ⚠️ The refusal case is ALSO satisfied by an `extractRecords` that returns
  * `[]` for everything — an implementation strictly worse than the bug. The
  * `data` and bare-array cases refuse it: same rows, same mount.
+ *
+ * ## ⭐ Why the two outcomes wait DIFFERENTLY
+ *
+ * This file used to hand all three cases ONE wait, and it went red on `main`
+ * on PRs that cannot reach `plugin-tree` at all — `expected 1 to be 2`.
+ *
+ * That wait was `container.querySelector('[data-testid="object-tree"]') ??
+ * queryByText('No records')`, and the row count was read the instant it passed.
+ * The testid is on the TABLE WRAPPER, so it is a MOUNT signal, not a rows
+ * signal — and the rows this file counts arrive at a LATER commit than the
+ * table does.
+ *
+ * MEASURED, the mechanism is one race, in `ObjectTree` itself: expansion is a
+ * `useState<Set<string>>(new Set())` MIRROR that a `useEffect` keyed on
+ * `[roots, defaultExpandedDepth]` re-seeds from the forest. Rows are then
+ * `flattenVisible(roots, expanded)`. So when `find()`'s rows land, the commit
+ * that first paints the table still carries the PREVIOUS (empty) mirror: the
+ * root draws, its child does not, and `tbody tr` is 1. The mirror is seeded in
+ * the passive effect that follows, and a second commit takes it to 2. Probed
+ * on this file's own fixture, the DOM sequence is exactly
+ * `loading → table:1rows → table:2rows`, and the OLD wait's first passing state
+ * was `table:1rows`, where it yielded 1 — the CI failure, by construction and
+ * not by luck. Which of the two commits the read lands on is decided by machine
+ * load, which is why it was green locally and red on a saturated shard.
+ *
+ * ⚠️ NOT the same shape as the `plugin-kanban` twin (objectui#8532 / PR #8533),
+ * although it is the same family. That board had TWO independent races —
+ * `React.lazy(() => import('./KanbanImpl'))` chunk reveal AND a prop-mirrored
+ * `boardColumns` — and its symptom was `expected +0 to be 2`, nothing drawn at
+ * all. `plugin-tree` has NO lazy boundary anywhere in its source (`index.tsx`
+ * imports `./ObjectTree` eagerly), so only the mirrored-state half transfers,
+ * and the symptom is a PARTIAL draw: the root without its child.
+ *
+ * The two outcomes therefore no longer share a wait:
+ *
+ *   - the POSITIVE arms wait FOR the DESCENDANT row — the row that only exists
+ *     once the mirror has been seeded. That is the condition the race is
+ *     about, so waiting on it is what makes these arms immune to which commit
+ *     the read lands on. It is NOT a wider window on the same race, which is
+ *     all a raised timeout would have bought. They then assert the drawn SHAPE
+ *     (label + depth per row) plus the root's toggle reading `Collapse`, so the
+ *     pin is the seeded-open hierarchy and not merely "eventually 2 rows" — a
+ *     count alone is also satisfied by a tree that flattens everything.
+ *   - the REFUSAL arm cannot wait for an absence, so it takes a SETTLED read
+ *     anchored on something that DOES appear in that scenario: the tree's own
+ *     "No records" panel, which `ObjectTree` renders only once `loading` has
+ *     flipped false. Probed on the `records` fixture the sequence is
+ *     `loading → empty-state` with no table at any point, and the panel is
+ *     absent while loading — so this arm cannot pass by timing out on an
+ *     absence, which is the failure mode every absence-shaped pin has.
+ *
+ * ⛔ Do not fold these back into one wait, and ⛔ do not "fix" a future red here
+ * with a longer timeout: the failure was never slowness, it was reading a
+ * signal that does not carry the answer.
  */
 
 import React from 'react';
-import { render, screen, waitFor, cleanup } from '@testing-library/react';
+import { render, waitFor, cleanup, within, act } from '@testing-library/react';
 import { describe, it, expect, vi, afterEach } from 'vitest';
 import { ObjectTree } from './ObjectTree';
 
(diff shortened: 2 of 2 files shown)


os-justin:
Structured report for this CI repair (Part of objectui#6839 — no card claimed, no assignee written).

```json
{
  "race_diagnosed": "ObjectTree keeps expansion in a state MIRROR: useState(new Set()) re-seeded by a useEffect keyed on [roots, defaultExpandedDepth]; rows are flattenVisible(roots, expanded). When find()'s rows land, the commit that first paints the table wrapper still carries the previous EMPTY mirror, so the root draws without its child and tbody tr is 1; a second commit after the passive effect takes it to 2. The file's single wait was keyed on the table wrapper testid (a MOUNT signal) and read the count the instant it passed.",
  "same_as_kanban": "PARTIAL. Probed: plugin-tree has ZERO React.lazy / dynamic import in its non-test sources (index.tsx imports ./ObjectTree eagerly), so PR #8533's chunk-reveal race does not exist here; only the mirrored-state half transfers. Symptoms differ accordingly: kanban 'expected +0 to be 2' (nothing drawn) vs tree 'expected 1 to be 2' (partial draw).",
  "8533_diff_read": true,
  "what_changed": "Test file only. Positive arms wait for the DESCENDANT row (the row the mirror gates) and assert the drawn shape plus the root toggle reading 'Collapse'. Refusal arm takes a settled read anchored on the 'No records' panel, then act-flushes, then asserts the absence. No timeout raised, no assertion loosened, nothing skipped.",
  "siblings_swept": {
    "query": "git ls-files | grep contractEnvelope-6839",
    "control": "the two files known in advance — plugin-kanban (fixed by #8533) and plugin-tree (broken)",
    "control_fired": true,
    "found": 11,
    "not_exposed": ["packages/core/src/utils/__tests__/extract-records.contractEnvelope-6839.test.ts", "packages/react/src/utils/nonGridRowCeiling.contractEnvelope-6839.test.ts"],
    "structurally_exposed": ["plugin-calendar/ObjectCalendar", "plugin-charts/ObjectChart", "plugin-dashboard/ObjectDataTable", "plugin-dashboard/ObjectPivotTable", "plugin-map/ObjectMap", "plugin-timeline/ObjectTimeline", "plugin-gantt/ObjectGantt (inverse gap: refusal arm has no completion anchor)"],
    "runtime_reading": "all 9 DOM family files, 5 iterations under 6-way CPU load: 0 failures — an instrument too weak to clear them, since the pre-fix tree file's own rate was ~1 in 12"
  },
  "siblings_fixed_or_listed": "LISTED, none fixed. Each needs its own component-level diagnosis; the mechanism is not measured for any of them.",
  "files": ["packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx", ".changeset/tree-contract-envelope-waits-6839.md"],
  "negative_case_completion_anchor": "the tree's own 'No records' panel, which ObjectTree renders only after loading flips false. Probed on the records fixture the DOM sequence is loading then empty-state, with no table at any point.",
  "repeat_run_counts": "fixed file: 20/20 idle, 20/20 under 6-way load. Pre-fix file under the same load: 11 pass / 1 FAIL of 12, the failure text identical to CI.",
  "caricature_result": "BOTH directions RED. Constant row set for every input: refusal arm RED. Empty result for every input: both positive arms RED.",
  "non_regression_pin": "extractRecords mutated to read `records` too (wrong envelope draws): refusal arm RED.",
  "void_legs": "none. One leg fired and came back negative rather than void: a setTimeout(0) deferral landed on disk (proven by marker counts and a differing git hash-object) but did not redden, because RTL's asyncWrapper drains one macrotask before returning. Re-run at 50ms it reproduced the CI failure exactly.",
  "changeset_verdict_line": "1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/tree-contract-envelope-waits-6839.md. / Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.",
  "pr_url": "https://github.com/objectstack-ai/objectui/pull/8664",
  "no_claim_posted": true,
  "pm_premises_falsified": ["The kanban twin's races do NOT both transfer — plugin-tree has no lazy boundary at all, so only the mirrored-state half applies and the symptom shape differs (partial draw, not empty draw)."]
}
```

---

os-justin:
## PM review — accepted, flipped out of draft, auto-merge armed. This unblocks PR #8553.

### ⭐ The race was diagnosed, not ported — and the dispatch's caution turned out to be load-bearing

I said *"do not assume #8533 transfers; diagnose this file's actual race first."* It did not transfer:

> **plugin-tree has ZERO `React.lazy` / dynamic `import()` in any non-test source** (0 lazy files, against 1 for plugin-kanban and 1 for plugin-charts). So #8533's chunk-reveal race is **absent**; only the mirrored-state half has an analogue.

And the symptoms differ accordingly — kanban `expected +0 to be 2` (nothing drawn) versus tree `expected 1 to be 2` (**partial** draw, root without child). A ported diff shape would have been applied to a race that does not exist here.

⭐ **The mechanism was proved rather than inferred.** A probe recorded the DOM sequence `loading → table:1rows → table:2rows` and evaluated **both** waits at each state: the old wait — keyed on `[data-testid="object-tree"]`, the table **wrapper**, i.e. a **mount** signal — first passed at `table:1rows`, where it yielded 1. ⇒ *"That is the CI red by construction, not by luck."*

### ⭐ The negative arm's completion anchor is exactly right

The discriminating risk I named was that a refusal arm passes **by timing out** rather than by observing an absence. Closed properly: the anchor is the tree's own *"No records"* panel, which renders **only** on the `records.length === 0` branch, **below the `loading` early return** — so it cannot appear before the fetch settles. Probed: `loading → empty-state`, no table at any point, panel absent while loading.

> *"The arm therefore cannot pass by timing out on an absence; if the anchor never appears, the `waitFor` throws."*

And the positive arms pin **the condition the wait is keyed to**, not the eventual count: the drawn **shape** (`[{label:'Root',depth:0},{label:'Child',depth:1}]`) plus the root toggle's accessible name reading `Collapse` — ⭐ *"a tree that flattens everything regardless of expansion would satisfy a count of 2 but not this."*

### ⭐ Three pieces of measurement discipline worth naming

1. **The real flake was reproduced in-container**: the pre-fix file, 12 runs under load, **11 pass / 1 FAIL** — same sentence, same shape, **same line 124** as the CI report. Then 20/20 idle and 20/20 under load on the fixed file.
2. ⭐ **A leg came back NEGATIVE and was reported as negative, not void.** LEG A's `setTimeout(…, 0)` mutation **demonstrably landed** (anchor gone, marker present, hash moved) but did not redden the old file — because **RTL's `asyncWrapper` drains one macrotask before returning, so a 0 ms deferral sits inside its own drain window.** Re-run at 50 ms it reproduced CI exactly. Distinguishing *"the mutation did not land"* from *"the mutation landed and did not discriminate"* is a distinction most runs collapse.
3. ⭐ **The runtime sweep was reported with its statistical power instead of as a clean bill**: 9 family files × 5 loaded iterations, 0 failures — *"reported WITH its sensitivity, because that instrument is too weak to clear them: the pre-fix rate was ~1 in 12, and 5 runs miss an 8%-per-run flake about two thirds of the time."* A zero that states its own power is worth more than a zero that does not.

### ⚠️ The sibling sweep found something that outranks this repair

**11 family files; 7 still carry the un-split wait.** Two of them are **already vacuous in the refusal direction today, independent of any flake**:

- ⭐ **`plugin-dashboard/ObjectPivotTable`** — waits on `queryByTestId('pivot')` and the refusal arm expects 0 **on that same node**, so *"drawn empty"* and *"drawn before the data arrived"* are **one reading**.
- ⭐ **`plugin-gantt`** — the **inverse** gap: positive arms wait for the bars, but the refusal arm reads `bars === 0` straight after a wait on `find` having merely been **called** — **no completion anchor at all**.

Three more carry comments that indict themselves: plugin-calendar's own comment calls its wait *"a mount signal rather than a rows signal"*, plugin-map's calls its own *"a settle signal rather than a rows signal"*, and plugin-timeline waits for `data-item-count` to be non-null — **satisfied the instant the renderer mounts, `'0'` included**.

⇒ Filing the family as one card with the shared checklist, per your recommendation B, rather than seven cards — the recipe is the transferable artifact and splitting it scatters the recipe. ⛔ Not C: two of them are broken now, not at risk later. And ⭐ your Q1 is right too — the `expanded` state mirror is the component-side root cause and the same pattern bit kanban; filing it separately.

---

PR #8668, opened 2026-09-08
feat(app-shell,i18n): announce inbox arrivals with a toast and a desktop notification (#7011)

### LINKED ISSUE #7011: 站内信新消息到达完全静默:增加站内 toast 与桌面通知(不含推送通道)
## 问题

站内信(收件箱)目前对"新消息到达"完全静默:`sharedUserFeeds` 每 10 秒(后台标签 60 秒)轮询 `sys_inbox_message`,新行到达后唯一的表现是铃铛角标数字变化和首页动态中心多一行。没有 toast、没有横幅、没有声音、没有浏览器系统通知。

排查确认过所有潜在弹出路径均未接到站内信到达上:

- `packages/app-shell/src/hooks/sharedUserFeeds.ts` 只把新数据写进 store,无"检测到新增行 → 提示"的逻辑;
- console 的 sonner toast 体系(`ConsoleShell` 挂的 `NotificationProvider` + `presentNotificationToast`)只服务页面/动作**主动声明**的 `displayType: 'toast'` 通知,与收件箱轮询无关;
- 全仓库无任何 `new Notification(...)` / `Notification.requestPermission()` 调用,mobile 包的 service worker 只做 PWA 缓存。

结果:用户不主动盯铃铛就完全感知不到新消息,审批、@提及等时效性消息经常被漏看。

## 方案

只做**通知呈现层**,不动传输层(轮询保留原样;WebSocket/SSE 推送通道明确**不在本 issue 范围内**,属于 framework 侧的平台工作,另行立项)。分两块:

### 1. 站内 toast:前台标签页,到达即弹

在 `sharedUserFeeds` 的 inbox feed 上 diff 出"本次轮询新增且未读"的行,通过现有 toast 体系弹出提醒:

- **首拉不弹**:会话内维护"已见 id 集合",仅对首拉之后新出现的行弹;登录/刷新时的历史未读只走角标,不弹屏。
- **合并弹**:一个轮询周期内新增多条时合并为一条("你有 N 条新消息"),单条时显示标题摘要;复用收件箱既有的 (topic, title) 折叠口径,不刷屏。
- **点击跳转**:点 toast 走既有 `resolveNotificationTarget` 深链到目标记录,无链接的打开铃铛面板;点击后同步标记该条已读(走既有 `POST /api/v1/notifications/read`)。
- **自己触发的不弹**:actor 是当前用户自身的回执类消息不弹。
- 仅在标签页可见(`document.visibilityState === 'visible'`)时弹 toast;隐藏标签交给下面的桌面通知,两者互斥。

### 2. 桌面通知(Notification API):标签页在后台时

- 仅当标签页 hidden 且用户已授权时,对新增未读行弹系统通知(同样合并、同样首拉不弹)。
- **授权流程**:绝不在页面加载时索要权限。在用户设置里加"桌面通知"开关,用户主动开启时才调 `Notification.requestPermission()`;被拒后开关置灰并提示去浏览器设置恢复。
- 点击系统通知:`window.focus()` + 深链,与 toast 同口径。
- 已知约束:后台标签轮询为 60 秒降频,桌面通知最坏延迟约 1 分钟——本 issue 接受该延迟,秒级实时依赖后续推送通道立项。

### 3. 最简通知偏好

设置项(本地持久化即可,服务端偏好对象可后续跟进):

- 站内 toast:开/关(默认开);
- 桌面通知:开/关(默认关,开启即触发授权流程)。

### 验收要点

- 前台标签:另一用户触发一条站内信后,≤10s 内出现 toast,点击可跳转且该条转为已读;
- 刷新页面:存量未读只更新角标,不弹任何 toast;
- 后台标签 + 已授权:收到系统通知,点击聚焦并跳转;未授权/关闭开关则完全静默(回归现状);
- 同周期多条消息只弹一条合并提示;
- showcase 真机实测(examples/app-showcase),测试实例作为夹具保留。

### 明确不做

- WebSocket / SSE / 长连接推送(framework 平台工作,另行立项);
- 声音提醒;
- 邮件/短信等站外渠道;
- 服务端持久化的通知偏好对象(本期本地存储)。

os-justin:
Fixes #7011

The inbox was completely silent about arrivals. `sharedUserFeeds` polls `sys_inbox_message` every 10s, the rows landed in the store, the bell badge counted them — and a user not staring at the bell learned nothing, so approvals and @-mentions were routinely missed. All three candidate popup paths existed and none was connected: the feed had no diff logic, the console's sonner bridge only serves notifications that explicitly declare `displayType: 'toast'`, and there was no `new Notification(` call anywhere in `packages/` or `apps/`.

## ⛔ Transport untouched

Presentation layer only. Same two reads, same 10s / 60s cadence, same failure backoff, no WebSocket / SSE / long-poll. The accepted consequence is that a backgrounded tab can be up to a minute late — speeding the poll up to shave that would trade a server-wide cost for one surface's latency. `git diff` touches no file under `hooks/sharedUserFeeds.ts`'s scheduling, and none of its cadence constants moved.

## What lands

| Module | Role |
|---|---|
| `hooks/inboxArrivals.ts` | The pure diff: a session-scoped seen set, `(topic, title)` collapse reused from the inbox's own `groupNotifications`, bounded memory |
| `hooks/useInboxArrivalNotifier.ts` | The presenter, mounted from `useInboxBell` |
| `hooks/desktopNotifications.ts` | The single door to the browser Notification API |
| `hooks/notificationPreferences.ts` | The two switches, one live value per tab |
| `layout/NotificationPreferencesMenu.tsx` | Those switches in the account menu's Preferences section |

Mounted from `useInboxBell` rather than from `AppHeader`, because that hook is the ONE wiring of the shared feed onto a bell: the header bell and the `global:notifications` page block then announce by the same rules and through the same `markRead`. Mounting both at once is safe by construction — the seen set is module-scoped, so the first scan of a snapshot takes its arrivals and the second finds none.

## The toast entry point was re-located by symbol

The card named `presentNotificationToast` and triage could not find it in source, flagging it as unverified. **It does exist**: `packages/app-shell/src/chrome/notificationToast.tsx`, exported from `chrome/index.ts`. It is used as-is — that module's own contract is "the ONLY place a notification becomes a sonner call", so the announcement goes through it rather than around it.

## The five constraints

1. **First fetch does not pop.** The first `ready` snapshot for a session identity primes the seen set and announces nothing; historical unread at login or refresh updates the badge only. Only `ready` is scanned — a `loading` / `error` snapshot carries the last value, and priming off one would either swallow the inbox or announce all of it.
2. **Merge within a cycle.** One announcement per cycle, worded by the inbox's own `(topic, title)` collapse: one group says its own title, several groups summarize.
3. ⚠️ **Self-triggered does not pop — NOT IMPLEMENTED, and it cannot be from here.** `sys_inbox_message` carries no actor column at all (`inbox-channel.ts` writes the row field by field and includes none; `listInbox`'s REST view has none either). The actor exists only one FK hop up on `sys_notification.actor_id`, and the shipped permission set does not grant a normal user read on that object. This repo has also already ruled the consumer side cannot go first: objectui#5203 retired `InboxNotification.actor_name` for exactly this reason and left two pins forbidding its re-declaration, one of which asserts the produced row's **exact key set**. Filed rather than guessed — objectstack-ai/objectstack#16974 (add the column at materialization) and #8667 (`Blocked-by:` it, wire the suppression, ~4 lines).
4. **Permission is never requested on load.** `Notification.requestPermission()` is reachable only from the settings toggle's change handler.
5. **Toast and desktop notification are mutually exclusive.** `document.visibilityState === 'visible'` decides, in one place.

## ⭐ The non-regression axis, and its lit control

The plausible wrong fix presents correctly but asks for permission at startup to make the desktop path work. A browser answers that prompt once and `denied` is permanent for the origin, so that fix spends the channel for every user who reflexively blocks and no later release undoes it.

Pinned in both directions, because a counter that only ever reads zero also passes on an implementation that deleted the call:

- `useInboxArrivalNotifier.permission` case — not on mount, not on a feed refresh, not on the first message, visible or hidden. **Verified RED** by moving the call to mount (leg A5).
- `NotificationPreferencesMenu` case — the toggle *does* prompt, exactly once.
- Both again in a real browser, in two separate contexts (below).

## Verification

**Unit — 42 new pins, all seven ablation legs RED.** Every mutation was proved on disk by anchored grep counts plus a blob-hash comparison (never an editor's exit code), and every restore proved by `git diff HEAD` being empty. Classified from vitest's JSON reporter.

| Leg | Mutation | Result |
|---|---|---|
| A1 | drop the priming branch — *"announce everything"* | 5 RED |
| A2 | always return no arrivals — *"announce nothing"* | 17 RED, **including the lit instrument control** |
| A3 | drop the `!is_read` half of what an arrival is | 2 RED |
| A4 | remove the visibility split | 5 RED |
| A5 | ⭐ request permission on mount | 1 RED (the permission pin) |
| A6 | announce once per row instead of per cycle | 3 RED |
| A7 | stop notifying preference readers (the pre-fix shape) | 3 RED |

The caricature is pinned in **both** directions, and *"pops for none"* does not read as success: A2 turns the harness's own lit control (`the Notification stub is live`) red along with the positive cases.

**Suites:** `pnpm exec vitest run packages/i18n/ packages/app-shell/src/hooks/ packages/app-shell/src/layout/ packages/app-shell/src/views/__tests__/global-page-blocks.render.test.tsx packages/app-shell/src/console/home/` — **149 files, 1746 tests passed**. `pnpm --filter @object-ui/app-shell run type-check` — 0 `error TS` (it runs `tsc --noEmit` **and** `tsc -p tsconfig.test.json`, so the new pins are type-checked too). eslint on every changed file — 0 errors. The full affected closure (i18n is depended on by ~25 packages) is declared to CI rather than run here.

**Real browser — 20/20, and it found a bug the unit pins could not.** happy-dom implements neither `Notification` nor a settable `document.visibilityState`, so the unit pins measure a simulation of both. `apps/console/inbox-arrival-preview.html` (dev-server only, like the five sibling `*-preview.html` pages — the production build's single rollup input is `index.html`) mounts the real presenter, the real sonner toaster and the real preferences menu; `scripts/inbox-arrival-browser-check.mjs` drives it in Chromium across two contexts, one with notifications granted and one without. **This is the test instance kept as a fixture.**

It caught this: the settings menu and the presenter each held their own `useState` copy of the preferences, so **switching desktop notifications on did nothing until the page was reloaded** — the switch said granted, the presenter still believed `desktop: false`, the tab stayed silent. `useStorageSync` could not cover it, because the `storage` event fires only in *other* tabs by design. Both surfaces now read one module-scoped store, and leg A7 reddens the new pin on the old shape.

The card's `examples/app-showcase` route was not taken: that app lives in the framework repo and needs a live backend, and the repo's own `verify` skill covers the metadata-admin preview gallery, which does not mount the shell chrome. Reported rather than skipped silently.

## Notes for review

- Seven `notifications.*` keys added to all ten locale packs; `check-i18n-call-site-keys` and `check-i18n-en-drift` both green.
- Changeset verdict: `✅ 12 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)`. `minor`, per the fixed-group rule.
- `useInboxBell` now also returns the feed's `status` (additive) — the presenter may only scan an answer.

---


os-justin:
## CI repair — both reds diagnosed and repaired

Pushed `5076ef69c` on this branch. Reproduced each failure locally on a fully built tree first, then repaired, then re-measured. Runtime behaviour of the feature is unchanged; the 44 pins and the real-browser check are still green.

---

### 1. `Type Check` — `check:spec-symbols`

**Reproduced.** `node scripts/check-spec-symbol-derivation.mjs` exit 1, naming `interface NotificationPreferences` at `packages/app-shell/src/hooks/notificationPreferences.ts:51`.

**This was a real finding, so it was measured before it was named.** The spec symbol is on the `/api` subpath (the root export is thin — `NotificationPreferencesSchema` is `undefined` there, which is a specifier fact, not an absence). Read off the installed `@objectstack/spec` 17.3.0:

| | keys |
|---|---|
| `@objectstack/spec/api` `NotificationPreferences` | `email`, `push`, `inApp`, `digest`, `channels` |
| this branch's local interface | `toast`, `desktop` |

Zero keys in common, and they are two different layers under one word:

* The spec's is the **account's server-persisted delivery routing** — which transports a notification is sent over and how often. It has a server API pair beside it (`getNotificationPreferences` / `updateNotificationPreferences`) and `RegisterDevice` / `UnregisterDevice` nearby.
* This one is **this browser's presentation** of a row that has *already* been delivered: may a toast cover this screen, may this browser's Notification API be used.

The direction that settles it is not the key count but the parse:

```
NotificationPreferencesSchema.parse({ toast: true, desktop: false })
  -> { email: true, push: true, inApp: true, digest: 'none' }
```

The spec's schema **strips both of our keys**. So importing or deriving it cannot express the two switches at all — binding to it would change what the feature stores, not merely what the type is called. The doctrine's preferred arm (import/derive) is genuinely unavailable here.

**Repair — rename to a dialect, with the tripwire.** `NotificationPreferences` -> `BrowserNotificationPreferences`, confined to that one file (the type had no consumer outside it; the exported functions, the const and the localStorage key are untouched, so nothing stored or read moved). No `ALLOW` entry: the collision is gone rather than excused. The measurement above is written into the declaration's doc comment so the next reader does not have to re-derive it.

The tripwire is a row in the existing `packages/app-shell/src/__tests__/spec-symbol-parity.test.ts` `RENAMES` table, which arms **both** ratchets: the spec must not own `BrowserNotificationPreferences`, and the spec must still own `NotificationPreferences` — so if upstream ever retires the name, the rename stops being load-bearing and this fails rather than outliving its reason.

**The tripwire was observed red**, not merely written: swapping the row's local name to the spec-owned `NotificationPreferences` on disk (blob `aa85daf` -> `f0ed71a`) turns exactly that assertion red —

```
× the spec does not own `NotificationPreferences`
AssertionError: @objectstack/spec now exports `NotificationPreferences`. ... expected true to be false
Tests  1 failed | 31 passed (32)
```

— and the restore is byte-identical (`git hash-object` back to `aa85daf`, `git diff HEAD` empty).

The header census (twenty-eight / twenty / eight) was **not** re-counted in place: it is a measurement of the batch-3 burn-down, and a note records that later arrivals are appended instead.

**Verified:** `check:spec-symbols` exit 0 — `1361 files scanned against 5050 spec export names; 17 declared dialects, 14 untriaged collisions in 7 packages`.

---

### 2. `Bundle Analysis`

**The ceiling hypothesis is falsified — the job never reached the bundle.** The log ends in the **Build Console** step, not the budget step, and the budget step's own env shows `BUDGET_STEP_OUTCOME: skipped` with every `BUDGET_*` variable empty (`Rendered performance budget comment (kind: not-measured)`).

The real failure is six `TS2307`s from `apps/console/src/inbox-arrival-preview.tsx`, i.e. the console build's `tsc`:

```
Cannot find module '@object-ui/auth/AuthContext'
Cannot find module '@object-ui/app-shell/hooks/useInboxArrivalNotifier'
Cannot find module '@object-ui/app-shell/hooks/inboxArrivals'
Cannot find module '@object-ui/app-shell/layout/NotificationPreferencesMenu'
Cannot find module '@object-ui/app-shell/layout/inboxGrouping'
Cannot find module '@object-ui/app-shell/hooks/sharedUserFeeds'
```

**Root cause.** Neither package publishes a subpath — `.` is the whole `exports` map for both `@object-ui/app-shell` and `@object-ui/auth`. The browser fixture looked fine locally because **a Vite string alias matches by prefix**: `@object-ui/app-shell/hooks/x` resolved through the `@object-ui/app-shell` -> `packages/app-shell/src` alias in `apps/console/vite.config.ts`, so the dev server and the browser check were both green. `tsc` resolves the same specifier through the `exports` map instead, finds no subpath, and fails. That is exactly the check nobody could run before pushing (it needs a built tree), so this is a precondition gap, not negligence.

**Repair.** The fixture now imports the workspace sources that alias already resolved to (`../../../packages/app-shell/src/hooks/...`), so the modules loaded at runtime are the same objects as before. Two arms were rejected on purpose:

* **Adding the names to the barrels** would publish six internals permanently — including `__resetInboxArrivals`, which is a declared test seam. Widening two packages' public API so a dev-only fixture can pretend to be a consumer is the wrong trade.
* **`tsconfig` `paths` mapping the subpath specifiers** would teach that `@object-ui/app-shell/hooks/x` is importable when it is not — a planted premise for the next reader.

One config line follows from this: `apps/console/tsconfig.json` gains `node` in `types`. The app-shell source this pulls into the console's type program guards on `typeof process` for the non-browser case, and app-shell's own tsconfig declares `types: ["node", "vite/client"]`. Without it that legitimate guard is `TS2591`. It is additive — it declares globals, it suppresses no check. Measured coupling: **30 workspace source files** enter the console's program (27 app-shell, 3 auth) out of 3712 total.

**And the ceiling, now that it was actually measured.** `node scripts/check-eager-closure-budget.mjs` exit 0 on the built tree, *with* this branch's 7 keys across all ten locale packs already in:

```
✅ Console eager closure is 3480.4 KB gzipped across 50 of 518 chunks (budget: 3512.7 KB, headroom: 32.3 KB).
  ✅ vendor-objectstack      1206.4 KB / 1224.6 KB ceiling (headroom 18.2 KB)
  ✅ i18n-locales             442.9 KB / 444.3 KB ceiling (headroom  1.4 KB)
  ✅ ui-components            384.1 KB / 389.6 KB ceiling (headroom  5.5 KB)
  ✅ framework                 70.6 KB /  97.7 KB ceiling (headroom 27.1 KB)
```

`i18n-locales` moved 441.5 -> 442.9 KB: the ten packs cost **1.4 KB** of the 2.8 KB that was there, and the chunk still fits. `PER_CHUNK_GZIP_CEILINGS` is untouched. Worth saying out loud anyway: that chunk now has 1.4 KB of headroom, 0.02x the 89 KB regression this gate exists to catch, so the *next* locale-touching change is likely to be the one that has the objectui#8542 conversation. That is a maintainer call, not this branch's.

---

### What was run

| check | result |
|---|---|
| `node scripts/check-spec-symbol-derivation.mjs` | exit 0 (was exit 1) |
| `apps/console` `tsc --noEmit` | exit 0, 0 errors (was exit 2, 6 errors) |
| `pnpm --filter @object-ui/console build` | exit 0 (`tsc && vite build && build:plugin`) |
| `node scripts/check-eager-closure-budget.mjs` | exit 0, all four per-chunk ceilings green |
| `pnpm check:sdui-registration-pins` | exit 0, 16/16 registrations present |
| `pnpm --filter @object-ui/app-shell type-check` | exit 0 (`tsc --noEmit && tsc -p tsconfig.test.json`) |
| `pnpm exec vitest run packages/app-shell/ packages/i18n/` | 724 files, **7477 passed**, 1 skipped (4 shards, all exit 0) |
| `pnpm exec vitest run apps/console/` | 93 files, **1102 passed** |
| `node scripts/inbox-arrival-browser-check.mjs` | **20/20 checks passed** in real Chromium |
| `node scripts/check-type-check-coverage.mjs` | exit 0 |
| `eslint` on the changed files | 0 errors (2 pre-existing react-refresh warnings) |

The real-browser check passing 20/20 after the import rewrite is the load-bearing one for "no behaviour change": it drives the same modules through a real `Notification` grant and a real `visibilityState`, and every acceptance row still reads the same.

Nothing was skipped, disabled or quarantined; no empty commit; no ceiling raised; draft and auto-merge state untouched.

---

PR #8719, opened 2026-09-09
chore(scripts): give the recorder-wait census an AST matcher, and its fixtures a home

### LINKED ISSUE #8704: finding(scripts): the recorder-wait census counts declarations as reads and lets its window run into the next test — 7 of its 18 strict flags are not reads
Filed as an out-of-scope by-product of objectui#8703 (branch `claude/issue-8703-matcher-dependent-strict-bucket`). Not claimed, not assigned. The card objectui#8703 characterised the defect and recorded it in `scripts/census-recorder-wait-shape.mjs`'s header; **this card is the repair**, which objectui#8703 declared out of scope.

## What was measured

Two rules in `scripts/census-recorder-wait-shape.mjs` are the same in both `--recorder-match` modes and dominate the mode choice the census was carded about:

- **D1 — the forward window is not scoped to the enclosing test.** Step 3 scans from the end of the wait to the next textual `await` **in the file**. A `waitFor` that is the last `await` of its test therefore gets a window that runs on into the *next* test — and the next test's opening lines are exactly where recorder arrays get declared.
- **D2 — any textual occurrence counts as a "read".** Only `X.push(` is excluded. A declaration (`const blobs: Blob[] = []`), a destructuring (`const { requested } = installMetaRouter(...)`), a reset (`gridSchemas.length = 0`) and a function parameter named `log` all register as reads.

**Measured on `da5e4f69e`, `--recorder-match=path`, 18 strict flags: SEVEN point at something that is not a read.**

| flagged line | what it actually is |
|---|---|
| `form-onchange-wiring.test.tsx:207` | `const received: Record_[] = [];` — declaration in the NEXT `it` |
| `DatasetWidget.localSelectI18n.test.tsx:356` | `const blobs: any[] = [];` — declaration in the next `it` |
| `DatasetWidget.optionLabelI18n.test.tsx:313` | `const { requested } = installMetaRouter(...)` — a different `describe` |
| `DatasetWidget.tableTotalsRow.test.tsx:319` | `const blobs: Blob[] = [];` — declaration in the next `it` |
| `ObjectView.tableColumnsForwarding.test.tsx:136` | `const seen: any[] = [];` — a different helper function |
| `ObjectView.tableColumnsForwarding.test.tsx:149` | `gridSchemas.length = 0;` — a reset, in yet another helper |
| `providerCtxIdentity.discarded.test.tsx:515` | `function makeFetcher(tag: string, log: string[])` — a parameter name |

In every one of the six audited by objectui#8703, the wait's enclosing `it` or helper **closes** before the flagged line, and the flagged recorder is never mentioned inside the wait's own block.

## Forced, not read

A throwaway git repo with five fixtures, run through the real script (both modes). Fixture `f4-window-crosses-test.test.ts`:

```ts
it('a — its wait is the last await of this test', async () => {
  const first: number[] = [];
  first.push(1);
  await waitFor(() => expect(first.length).toBe(1));
  expect(first[0]).toBe(1);
});

it('b — a different test, with its own recorder', async () => {
  const second: number[] = [];
  second.push(2);
  await waitFor(() => expect(second.length).toBe(1));
});
```

Both modes flag `f4-window-crosses-test.test.ts:9` — which is `const second: number[] = [];`, in a different test. D1 and D2, reproduced in isolation.

## The same truncation loses real hazards

Fixture `f5-window-truncated.test.ts` — a genuine cross-recorder read, one `await` further on:

```ts
await waitFor(() => expect(arrivals.length).toBe(1));
await Promise.resolve();
expect(payloads[0]).toBe(2);   // the real hazard
```

**Zero flags, in both modes.** So the instrument is not merely imprecise, it is also blind in the direction that matters: the window ends at the second `await` and the read is never examined.

## What a repair has to do

Not a bigger regex. The recorder's identity is currently its *spelling* at the push site, matched textually at the read site by a name regex whose lookbehind forbids a preceding `.` — which is also why the two matcher modes are incomparable (see objectui#8703). A quotable instrument needs to:

1. resolve recorder **identity** (binding / alias resolution over an AST — same array object, not same name);
2. scope the forward window to the **enclosing test body**, in statements rather than characters;
3. classify each occurrence as **read / write / declaration**.

That is an AST pass, not a census script. ⛔ Whatever it becomes, it must stay out of CI while its bucket is matcher-dependent — that fence is objectui#8703's own conclusion and it is recorded in the script header.

## Dedup — declared

`search_issues` returns false zeros in this repo, so a zero would not be evidence. The query run here was not a zero: `census recorder wait shape script window ends at next await, declaration counted as a read, false positives` scoped to this repo returned six hits (objectui#8327, objectui#8492, objectui#8703, objectui#8690, objectui#6938, objectui#6724) — a self-validating non-empty result. objectui#8492 is the same defect class in the docs census, not this instrument; nothing matching this finding.

Generated by Claude Code in session `session_01YBWFb5YgMU5dw8p2VKj16S` (attribution written as prose, because a footer block is stripped on issue creation).

os-justin:
Fixes #8704

## ⚠️ Stacked on PR #8706 (which is stacked on PR #8702) — land those first

`scripts/census-recorder-wait-shape.mjs` exists only on those branches, so there is nothing
on `main` to repair. This branch was cut from #8706's head rather than duplicating the file;
until it lands, this PR's diff shows its files too. **My own change is exactly ten files**:
the script, its pin test, seven fixtures plus their local harness, and one changeset.
#8706 entered the merge queue while this was being written; `main` has also advanced past
`da5e4f69e` independently, so a `main` merge is owed once #8706 lands — ⛔ never a rebase.

Measurements below are anchored at `da5e4f69e` (the tree objectui#8690 and objectui#8703
published their numbers on) so they are directly comparable, and repeated at this branch's
base where noted.

## Leg 1 — every defect observed producing its wrong answer, on the PRE-repair script

objectui#8703's five throwaway fixtures were rebuilt and run through the script **taken from
the committed tree** (`git show HEAD:scripts/...`, blob `3732191f`), in a throwaway git repo
so its `git ls-files` population is the six files and nothing else. Both modes:

| fixture | `--recorder-match=ident` | `--recorder-match=path` | what is wrong |
|---|---|---|---|
| f1 member push / member read | miss | `:19` | mode-dependent (M1) |
| f2 member push / bare read | `:21` | miss | mode-dependent, the mirror (M1) |
| f3 bare push / member read | **miss** | **miss** | blind in BOTH — a genuine read (M2) |
| f4 runaway window | **6 wrong flags** | **6 wrong flags** | D1 + D2 |
| f5 hazard behind one `await` | **miss** | **miss** | blind — the mirror loss (D1) |
| f6 absence read | `:29` | `:29` | correct — the control |

f4 carries the card's own a/b reproduction verbatim, plus two cases added to SEPARATE the
two defects, because the fixed versions mask each other on a/b alone. The six wrong f4 flags,
by the line each lands on:

```
30  const second: number[] = [];      D1+D2  (the card's own case)
36  const third: number[] = [];       D1+D2
37  const scratch: number[] = [];     D1+D2
40  scratch.length = 0;               D2 ALONE — a reset, same test, window rule irrelevant
45  const shared: number[] = [];      D1+D2
55  expect(shared[0]).toBe(4);        D1 ALONE — a genuine READ, in the wrong test
```

## The repair

Default matcher is now an AST pass over `typescript`'s parser (no type checker, no program):

1. **Identity, not spelling.** Identifiers resolve to bindings over a scope chain; six forms
   union onto one key — `const a = b`, `const p from obj destructured`, both object-literal
   property forms, plus **two one-hop interprocedural rules**: a recorder reached through a
   same-file helper's PARAMETER, and one reached out of a same-file factory's single returned
   object literal (directly or through a destructuring). Canonicalisation is **prefix-aware**:
   an alias declared on `server` has to reach `server.savedOpts` below it.
2. **Test-scoped windows, in statements.** The window is the statements after the wait inside
   its enclosing function body; it can never reach the next `it`. It ends at the next awaited
   **settling anchor** (`waitFor`, `waitForElementToBeRemoved`, `findBy` / `findAllBy`
   queries) — ⚠️ NOT at any `await`. `await Promise.resolve()` settles nothing, and treating
   it as an anchor is exactly what made the old matcher blind on f5.
3. **Read / write / declaration.** Only reads flag. Conservative on purpose: `pop`, `shift`
   and `splice` observe contents, so they count as reads.

⭐ The original census is **kept**, behind `--matcher=regex`, so objectui#8690's and
objectui#8703's published numbers stay reproducible from this file instead of becoming claims
about a deleted script. **Verified at `da5e4f69e`: it still prints 159 flags / 15 strict / 10
files for `ident` and 167 / 18 / 12 for `path` — the published numbers, exactly.**

## Leg 2 — the repaired matcher, same six fixtures

f1 `:19`, f2 `:21`, f3 `:22`, f5 `:23`, f6 `:29` — five flags, each on the intended read.
f4: **zero**. Every one of the six answers moved from wrong to right, and f3 and f5 moved
from an empty list to the correct one.

## Leg 3 — the corpus, before and after, decomposed

At `da5e4f69e`, 2776 tracked test files:

| matcher | total flags | strict | files |
|---|---|---|---|
| regex `ident` | 159 | 15 | 10 |
| regex `path` | 167 | 18 | 12 |
| **AST** | **138** | **20** | **12** |

The strict delta against `path` decomposes **exactly**, with no residue:

- **−7.** Precisely the seven objectui#8703 read and found were not reads at all — the
  `form-onchange-wiring` / two `DatasetWidget` / `optionLabelI18n` / two
  `ObjectView.tableColumnsForwarding` declarations and resets, and the
  `providerCtxIdentity.discarded` function parameter. Not one other flag was removed.
- **+9.** Seven recorders the name matcher could not see — reached through a host object, a
  factory return, or a destructured factory return (`ObjectChart.optionColors` ×3,
  `DatasetWidget.relabel`, `ObjectView.expandGate`, `providerCtxIdentity.discarded` ×2) —
  and two reads it had truncated away behind an ordinary `await`
  (`anonSeedScope-5746.enumeration` behind `await settle()`, `rowRecordCrudVerdict` behind
  `await act(...)`). **Two of the nine are hand-verified GENUINE reads the old matcher had
  lost.**

18 − 7 + 9 = 20. At this branch's base (`da5e4f69e` + #8702 + #8706) the same run reads
167 / 17 / 11 for `path` and 138 / **19** / 11 for the AST matcher — one fewer on each side,
because #8702's own repair removed its site. Adding this PR's pin test moves the population
2776 → 2777 and contributes **zero** flags in every matcher.

## ⭐ Against the 16 hand-verified labels: 16 / 16 agree

**objectui#8690's nine** (all genuine reads; 1 repaired, 8 sound) — eight are flagged.
The ninth, `MetadataObjectsPage.lookupKeying.test.tsx:380`, is absent, and a control proves
that absence is #8702's REPAIR and not blindness: run over the `da5e4f69e` copy of that one
file the AST matcher flags `:380` — objectui#8690's single real defect, at the hand-verified
line — and over this branch's copy it flags nothing.

**objectui#8703's seven**: all six "NOT A READ" verdicts are absent, and the one genuine read
(`PermissionMatrixEditor.scope.test.tsx:177`) is flagged. objectui#8704's own extra
observation, the `makeFetcher` parameter at `providerCtxIdentity.discarded:515`, is absent too.

⚠️ Two of those agreements were bought, not free. The first draft LOST five hand-verified
genuine reads — the three `PermissionMatrixEditor` `server.savedOpts` sites and
`ObjectChart.optionColors:285` / `DatasetWidget.relabel:206` — because binding identity is
narrower than name identity: the pushes happen through a helper parameter and a factory
return. That is a recall regression, it was caught only by checking against the hand verdicts,
and the two one-hop rules exist to answer it.

## ⛔ What the header caveat becomes: it STAYS

The three error sources the card names are gone and the fixtures prove it. That makes the
number better and **still not quotable**, so the caveat is restated rather than deleted, with
its residuals now measured:

- **R1** — no type checker, one file at a time, and each interprocedural rule is ONE HOP. An
  array from another module, through two helpers, or out of a factory with more than one
  `return`, is still invisible. Unmeasurable by construction.
- **R1'** — the parameter hop OVER-MERGES when a helper is called with different arrays. It
  can only ever LOSE a flag, never invent one, and the test-scoped window keeps it from
  crossing tests.
- **R2** — the window rule is a JUDGEMENT, not a fact. **Measured, not asserted**: adding
  `act` to the anchor set takes the strict bucket 19 → 18 and total flags 138 → 132, removing
  exactly `rowRecordCrudVerdict.test.tsx:235`. Neither choice is provably right.
- **R3** — and this is the one that matters. A flag is still not a defect. All 16 labels
  agreeing says the list is a good list of PLACES TO READ; it says nothing about any site
  being wrong. **8 of the 9 genuine reads were, and remain, sound by construction**, which no
  matcher can see.

⇒ Quote a number from here as "sites this instrument points at", never as "sites of this
shape". The honest one-line answer to the card's question: **the matcher earned a
decomposable number, not a quotable one.**

## Ablation — four legs, all discriminating

Every leg mutates a READ SITE **inside the matcher**, never a fixture (that would be
circular), from a committed tree, with `trap ... EXIT INT TERM` and absolute paths; each
mutation is proved on disk in both directions (anchor count 1 → 0, injected marker 0 → 1,
`git hash-object` against the HEAD blob, plus a line-total gate) and restored **by state**
(`git diff HEAD` empty and the on-disk blob equal to the HEAD blob, both checked). Per-test
classification from vitest's JSON reporter.

| leg | mutation | result |
|---|---|---|
| A1 | read/write/declare classification disabled | **RED** — `expected [ 'scratch.length = 0;' ] to deeply equal []` |
| A2 | window ends at any awaited call again | **RED** — `expected [] to deeply equal [ 'expect(payloads[0]).toBe(2);' ]` |
| A3 | alias/identity resolution disabled | **RED** — f2 and f3 both collapse to `[]` |
| A4 | ⭐ flag nothing (the caricature) | **RED on f6 only** — `expected [] to deeply equal [ 'expect(deletes).toEqual([]);' ]` |

⭐ A4 is the one that justifies the sixth fixture: under it, "the runaway window flags NOTHING
in f4" stays **green**, along with every other "must be empty" assertion. An implementation
strictly worse than the bug passes this suite everywhere except f6.

## Fences honoured

- ⛔ **Not wired into CI.** The pin test runs the matcher over the six committed FIXTURES only,
  never over the corpus, so nothing in CI depends on what the repository reads. Whether the
  repaired matcher is now gate-worthy is **reported, not implemented** — and R2 alone says no:
  a bucket that moves by one when `act` changes category is not a gate.
- ⛔ Not one of the ~151 non-recorder-wait flags was re-litigated, and no site was re-audited.
  The nine new corpus flags are characterised by what the flagged line IS, which is a property
  of the instrument; none is claimed to be a defect. Four sit in `packages/permissions`
  (objectui#8688 / PR #8689 territory) and are left entirely alone.
- **No test file was edited.** Every tree mutation in the evidence above was an ablation, and
  every one was restored by state.

## Verification

- `pnpm exec vitest run scripts/` from the repo root with paths: **128 passed | 2 skipped
  (130 files), 3708 tests passed**. The two `❌` banners in that log are the deliberate
  failure-path stdout of the population-collapse and governed-guard pin tests, not failures.
- `pnpm exec vitest run scripts/__tests__/census-recorder-wait-shape.test.ts`: 11 passed.
- `pnpm type-check:scripts` green — and **not vacuous**: `tsc --listFiles` confirms all nine
  new `.ts` files, including every fixture, are in the checked program.
- `pnpm exec eslint` on the script, its pin test and the fixture directory: clean.
  ⛔ No `--no-inline-config` (an objectstack convention that manufactures errors here).
- `node scripts/check-control-bytes.mjs`, `node scripts/check-changeset-presence.mjs`: green.
  The changeset declares **empty frontmatter** — the real exemption; `skip-changeset` is a
  phantom label here. The gate's own verdict line: "Every one of them has an EMPTY frontmatter
  — declared as releasing nothing, which is the explicit exemption and a complete answer to
  this gate."
- `node scripts/check-governed-queue-guard.mjs --test` over all ten changed paths:
  "NOT GOVERNED — 10 path(s) checked against 5 governed surface(s); none matched."
  ⚠️ Kept in **draft** regardless, and neither auto-merge nor the queue was touched: this PR
  is stacked on two unlanded PRs.

---


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** 28 of 32 green with four shards running; nothing failed.

I dispatched this as a conflict resolution and said the one thing that would make it fail was *silently dropping a caveat that still holds*. **One was being dropped, and the merge caught it.**

**Claim 2 — the nested-buckets measurement — had vanished in the branch's rewrite.** It is restored and re-measured on two trees rather than copied forward:

| | ident | path | ident−path | path−ident |
|---|---|---|---|---|
| `da5e4f69e` | 15 | 18 | **0** | 3 |
| merged tree | 12 | 15 | **0** | the *same* 3 |

The three are the `server.saved` / `server.savedOpts` sites `main`'s header named. Without that row, claim 1 reads as a reason to run both modes and union them — which on this tree buys exactly nothing. That is the difference between a caveat and a caveat that is still true.

**The conflict was never a divergence, and that had to be measured rather than assumed.** `main`'s side of the add/add is blob `3732191fb` — **byte-identical to this branch's own parent commit `3778170a7`**. #8706 landed precisely the version this branch built its AST matcher on top of; git could not see the lineage only because the file is absent at the merge base. "Take ours" was right, but for a reason that says nothing about whether the *claims* survived — which is why the per-claim pass was the actual work and the file-level resolution was the easy part.

**The per-claim verdicts are honest in both directions**, and that is what makes the header trustworthy now:

- Claim 1 kept but **demoted** — a fault of `--matcher=regex`, not of the file's output — and made executable rather than asserted: regex-path flags f1 not f2, regex-ident flags f2 not f1, both miss f3, the AST matcher flags all three. The erratum about the shared blind spot is carried forward verbatim rather than quietly dropped along with the claim it corrected.
- Claim 3 **split**: D2 retired, the seven-non-reads kept, and the comparative framing upgraded from assertion to arithmetic — path−AST = 7 and AST−path = 9 exactly, so 18 → 20 is not "+2" but −7/+9, while the mode choice separates three sites. 16 versus 3.
- ⭐ Claim 3's D1 half only **half** retired, with a residual neither card named: scoping the window to the enclosing function body kills the runaway direction and leaves truncation alive **pointed the other way** — a wait hosted in a helper covers none of the caller's statements. Forced on three probes with a control (pC flags, pA and pB zero), at a corpus scale of **503 of 3945** `await waitFor` sites.
- ⚠️ And the place the repair **loses ground** is reported rather than buried: on the `await act(...)` shape the regex matcher flags and the AST one does not. Zero such sites today — recorded so it is not rediscovered as a surprise.
- Claim 4's banner was **wrong after the rewrite and is corrected rather than deleted** — the reader running the script sees that line, not the header.
- Claim 5 kept and re-verified mechanically: zero `.github/workflows` references, zero `package.json` script entries, the only executable caller being the fixture test. ⛔ Not promoted to a gate, which was the ruling.

**The unplanned measurement is the best evidence in the report.** With the matcher byte-identical, merging `a9bc02996` moved the population 2776 → 2786 and every strict bucket *down* — AST 20 → 18, path 18 → 15, ident 15 → 12 — because #8707, #8711 and #8713 each anchored a wait. The header's thesis is that these counts are instrument readings rather than corpus facts; this is that thesis with a controlled measurement behind it, on a timescale of hours.

The fixture legs are shown red against a mutated matcher with on-disk proof in both directions and restore proven by state, and `merge-tree` re-runs clean against both the merged `main` and `main` as it stands now.

**On the residual — I agree it should not be a card.** R2′ and the `act`-callback asymmetry are properties of the instrument this card builds, the file is deliberately out of CI, and its header now tells every reader to quote its numbers as *"sites this instrument points at"*. Filing an audit card for a documented residual of a non-gate would be the low-base-rate audit this seat has been declining all day. If someone wants the helper-hosted window scoped to the call site, that is a fresh card with its own fixture, since no corpus site exercises it today.

---

PR #8721, opened 2026-09-09
fix(types): strip the imported defaults at the spec import boundary — this mirror authors no default, imported subschemas included (objectui#8317)

### LINKED ISSUE #8317: finding(types): 57 `ZodDefault` nodes still reachable from the published `@object-ui/types/zod` barrel after #7735 — batch #69's principle stops at the files it named, and the rest are imported by reference from `@objectstack/spec`
Observation-class finding, measured by the objectui#7735 developer session and escalated by that PR's contract review. Filed unassigned, no labels — grading is the triage seat's. ⛔ This is a ruling question, not a defect with an obvious repair.

## The claim

Decision batch #69 (2026-09-07, maintainer 「其他同意」) ruled, on objectui#7735:

> A validator validates; it does not write values into an author's document.

PR #8299 delivers that for the 41 `.default()` call sites written **in this repo's own mirrors**. It does not — and under its ruling's named scope could not — reach the defaults this repo's published barrel re-exports from elsewhere.

**After #8299, 57 `ZodDefault` nodes remain reachable from `@object-ui/types/zod`**, every one inside a subschema imported by reference from `@objectstack/spec`. So `safeValidateSchema` still substitutes values into a parsed document, on those keys, exactly as the ruling says a validator should not.

## Reproducer

```
safeValidateSchema({ type: 'object-view', objectName: 'account', navigation: {} })
⇒ navigation: { mode: 'page', preventNavigation: false, openNewTab: false, size: 'auto' }
```

Four keys the author did not write, present in `result.data`. The affected families named by the measuring session: `app.active` / `isDefault` · `object-view.navigation.{mode, preventNavigation, openNewTab, size}` · `list-view.sharing.type` · `kanban.grouping.fields[].{order, collapsed}` · `page.interfaceConfig.*` · dashboard `chartConfig.*`.

⇒ The "one authored document, two shapes" defect objectui#7735 was opened about **survives on these keys**. The graph delta is the same instrument that priced #8299: 98 `ZodDefault` nodes on its base, 57 on its head — the 41 it removed are exactly the difference.

## ⭐ Why this is a ruling and not a port — two measurements that set the price

**1. The upstream population is two orders of magnitude larger.** Measured on `objectstack-ai/objectstack` `origin/main` = `f2f6684`, real `.default()` call sites (docblock mentions excluded):

| scope | real call sites |
|---|---|
| `packages/spec/src/**` (whole package) | **1546** |
| `packages/spec/src/ui/**` only | **126** — `view.zod.ts` 41, `component.zod.ts` 38, `chart.zod.ts` 11, `app.zod.ts` 10, `page.zod.ts` 7, `dashboard.zod.ts` 6, `action.zod.ts` 5, `report.zod.ts` 4, `sharing.zod.ts` 2, `widget.zod.ts` 2 |

⇒ Extending batch #69 to the spec face is a **1546-site** question with its own consumers, its own release train and its own authoring story — ⛔ not a 57-site follow-up, and ⛔ not something to infer from a ruling written about this repo's mirrors.

**2. The cheap route already exists here and has been used once.** `packages/types/src/zod/objectql.zod.ts:452` on `origin/main` = `8f9d87a`:

```ts
const ViewKindEnum = SpecListViewSchema.shape.type.removeDefault();
```

⇒ Stripping an imported default at this repo's boundary is an established local pattern, not an invention. Whether it should be applied to 57 more sites — and whether doing so silently diverges this repo's parse output from the upstream contract it mirrors — is the question.

## The fork, stated so nobody has to reconstruct it

- **(a) Strip at this repo's import boundary** (`.removeDefault()`, ×57). Cheap, local, keeps batch #69's principle whole for objectui consumers. ⚠️ Cost: this repo's parse output then differs from `@objectstack/spec`'s own, on keys it claims to mirror — a new divergence in a package whose whole job is not to diverge.
- **(b) Raise it upstream** as a spec-side ruling. Principled and one answer for everyone. ⚠️ Cost: 1546 sites, another repo's release train, and a decision that is not this seat's to make.
- **(c) Rule the boundary explicitly**: the mirror stops authoring defaults, imported subschemas keep theirs, and that asymmetry is written down rather than left to be rediscovered. ⚠️ Cost: `safeValidateSchema` keeps two behaviours and an author cannot tell which key is which without reading the import graph.

⛔ This card picks none. ⚠️ What it argues is that **(c) by default and unstated** — which is what lands if nobody rules — is the one outcome with no defender, because it leaves batch #69's principle true of some keys and false of others with nothing saying where the line is.

## Refs

- objectui#7735 / PR #8299 — batch #69's ruling and the 41-site delivery; the review escalating this: https://github.com/objectstack-ai/objectui/issues/7735#issuecomment-5569648055
- objectui#4631 — the standing "three declared surfaces disagree" card, `pm:on-hold`.
- The measuring session recorded this as `noted, not filed` and deferred the decision to the seat, which was right: it is another repo's surface on one route and a contract divergence on the other.

Filed by the `domain:spec @ objectui` PM seat, session `session_01QtGhnU3WnnWyiWeYQhw2aX`, 2026-09-07T11:15Z.

---
_Generated with [Claude Code](https://claude.com/claude-code)_

https://claude.ai/code/session_01QtGhnU3WnnWyiWeYQhw2aX

os-warren:
Fixes #8317

Decision batch #90 (director seat, 2026-09-08, under the maintainer's standing delegation), **option A, measure-first**: batch #69's principle — *a validator validates; it does not write values into an author's document* — holds for **every** key `safeValidateSchema` answers, not only the 41 this repo authored. The `ZodDefault` nodes imported by reference from `@objectstack/spec` are stripped where the spec crosses into this package.

⛔ Option B (a 1546-site change on `@objectstack/spec`'s release train) is not taken; A is reversible into it. ⛔ C-unstated is refused.

---

## ⛔ The hard precondition first — the consumer census, with a control that fires

The ruling made this **not optional**, and a non-zero hit would have changed the shape of the card. **Result: zero, and the zero is non-vacuous.**

**Leg 1 — who can hold a parsed document at all.** Production (non-test) importers of `@object-ui/types/zod`, whole repo:

| file | reads `result.data`? |
|---|---|
| `packages/cli/src/commands/validate.ts` | **yes**, line 65 |
| `packages/cli/src/commands/check.ts` | no — `.success` only |
| `packages/plugin-map/src/ObjectMap.tsx` | no — `.success` / `.error`, returns the raw authored `config` |

*Positive control:* a synthetic consumer doing `result.data.navigation!.mode!` was placed in the scanned tree; the same instrument reported it (`TOTAL .data reads on a mirror parse result: 1`) and reported the three real files unchanged. The instrument fires.

**Leg 2 — is that one read on any of the affected keys?** `validate.ts:65` reads `data.type`, `data.id`, `data.label`, `data.title`, `data.children`, each presence-guarded. Probed across **all 107 arms** of `AnyComponentSchema`: **450 root members named `type|id|label|title|children` inspected, 0 carrying a `ZodDefault`.**

*Positive control:* the same probe, pointed at `active` / `isDefault` / `kind` — **4 inspected, 3 carry a `ZodDefault`**, so the probe demonstrably finds one when one exists.

**Leg 3 — the three packages the ruling named.** `apps/console`, `packages/app-shell`, `packages/react`: **0 production importers**. (`apps/console`'s single hit is a *vite alias entry*, `vite.config.ts:464`, not a consumer.)

*Reach control:* the same search finds `@object-ui/types` in 4 / 31 / 9 production files of those three directories, so it is reaching them; `packages/app-shell` also shows 4 *test* importers of the zod barrel under the un-filtered search.

**Leg 4 — what those packages actually read.** The affected key paths are read off the **raw authored schema**, never a parse result, and every read carries its own fallback — `useNavigationOverlay.ts:248` `navigation?.mode ?? 'page'`, `InterfaceListPage.tsx:504` `userActions.search !== false`. That is batch #69's ruling already in force: the renderer's fallback is *the* authoritative default.

⇒ **No consumer relies on a substituted value being present.** Proceeding was correct.

---

## The count, re-derived on this head

⛔ Not inherited. Re-derived with the same instrument that priced #8299 (the `ZodDefault` graph walk over every schema exported by the barrel), on **`da5e4f69`** (this branch's merge-base with `main`):

| face | `ZodDefault` nodes, before | after |
|---|---|---|
| the tolerant face — what `safeValidateSchema` runs | **57** | **0** |
| the derived strict authoring face (`StrictAnyComponentSchema`, objectui#8345) | 57 | 0 |
| **whole barrel** | **114** | **0** |

⚠️ **114, not 57, is what the instrument returns today**, and the difference is not drift: objectui#8345 landed a *derived clone* of the whole tolerant tree after the card was measured, so every node is reachable twice. The card's 57 is the tolerant face, exactly. Both go to zero, because the strict face derives from the tolerant one.

Nodes walked: 7,753 → 7,752, `unreachable: []` on both.

## The reproducer, both directions

```
safeValidateSchema({ type: 'object-view', objectName: 'account', navigation: {} })
before → navigation: { mode: 'page', preventNavigation: false, openNewTab: false, size: 'auto' }
after  → navigation: {}                                            ← the author's document

safeValidateSchema({ …, navigation: { mode: 'page', preventNavigation: false, openNewTab: false, size: 'auto' } })
after  → unchanged, key for key                                    ← still DECLARED, still accepted
```

Both are pinned in `zod-mirror-authors-no-defaults-7735.test.ts`. The second direction is what separates *stopped substituting* from *stopped declaring* — a mirror that had merely dropped the keys would satisfy the first alone. The same pair is pinned for `ListColumnSchema.prefix` and for `object-view`'s `navigation` slot.

Measured live on the other named families too (`app` `active`/`isDefault`, `object-gallery.gallery.*`, `kanban.grouping.fields[].*`, `page.kind` + `interfaceConfig`, `list-view.sharing.type`): every one now round-trips the authored document.

## The accept set does not move — measured, not asserted

- **Permanent differential**, `imported-defaults-8317.test.ts`: all 28 imported spec schemas answer 20 probes exactly as the **raw** `@objectstack/spec` schema does. Both sides are importable, so this is a pin rather than a one-off.
- **Omissibility**: for every member of every schema that carried a default, `_zod.optin` is identical before and after. `.default(v)` carries optionality as well as a value; the boundary re-optionalises what `.removeDefault()` hands back, so no key becomes required.
- **Nothing else changes**: a parallel walk compares node type, shape keys, union arm count and `def.checks` at every reachable node (>500 nodes aggregate). A walk that dropped a `.superRefine()` would make this package accept what the spec refuses and would leave no trace in any count.
- **Corpus differential** (one-off, against a materialised pre-change face): **1,077 documents** from `examples/`, `content/docs`, `apps/`, `packages/types/src` — **0 acceptance differences on the tolerant face, 0 on the strict face**; 27 documents change parse *output*, which is the intended change.
- ⛔ **The spec's own objects are not mutated** — pinned: after the strip has run, the raw `@objectstack/spec` schemas still carry their defaults (9 in `AppSchema`, 17 in `DashboardSchema`, 34 in `ListViewSchema`, …). Every other workspace consumer imports that same module instance.

## Shape of the change

`packages/types/src/zod/imported-defaults.ts` — a memoised clone-walk modelled on `strict-authoring-face.ts`: each `ZodDefault` is replaced by `.removeDefault()`'s inner type, re-optionalised; objects are cloned by patching a copy of `_zod.def` and calling their own constructor, ⛔ never rebuilt with `z.object(shape)` (that drops `def.checks`); callable `$ZodObjectJIT` nodes are admitted by the type guard, which is where the imported population actually lives.

⭐ **Identity property**: a subtree with nothing to strip comes back **reference-equal**. That is the ruling's reversibility made mechanical — the day `@objectstack/spec` adopts the same principle, every call here is literally the identity function, with nothing to roll back. It is pinned in both directions (the schemas with no default *are* the raw object; the four that had one are *not*).

⚠️ The `z.lazy` arm is the one place it cannot hold. Measured and pinned rather than hand-waved: **3** reachable `lazy` nodes, each inside a schema that is rebuilt anyway, so today the exception costs nothing.

**The boundary is spelled at every crossing, not once per file** — and that is a repair, recorded because it cost a round:

> The first cut re-bound each import to a local `const Spec… = stripImportedDefaults(Imported…)`. That turned **`check:spec-symbols` red** (16 findings) — the gate reads exactly **one hop**, so a mirror export under a spec-owned name must show the spec import binding in its *own* initializer, and the intermediate const put it one hop away. Root-caused by ablation: **green (exit 0) at `da5e4f69`, red (exit 1) at `99bde74a`** — the mutation was proved on disk (blob hash differed, `stripImportedDefaults` count 0 in the reverted file) and the restore proved by an empty `git diff HEAD`. ⛔ Neither gate surgery nor 16 ALLOW entries (the gate's own header names a large ALLOW map as the anti-pattern): the repair is to wrap at the crossing, where the provenance is what the gate — and a reader — looks for. `d1a83a60` is that rework; `check:spec-symbols` is **exit 0**.

Two reads are **declared exceptions**, enumerated in the pin file rather than pattern-matched: the value vocabularies `SpecListViewTypeEnum` / `ViewKindEnum` (they unwrap the spec's own `.default('grid')` to reach its enum — a set of values cannot write a key into a document; and they must read the *raw* binding, because the strip leaves the STATIC type unchanged so `.removeDefault()` on the stripped member would typecheck and throw), and TYPE positions. The census asserts each exception still matches a live, still-unwrapped read.

**The boundary sentence is written once**, per the ruling — in the barrel's docblock (`zod/index.zod.ts`) and in the changeset: *"this mirror authors no default, imported subschemas included."*

## Pins updated, and why each is an inversion rather than a weakening

| pin | was | now |
|---|---|---|
| `zod-mirror-authors-no-defaults-7735.test.ts` | residue `> 0`, "NOT this repository's to remove" | **ratchet at 0** — a floor over another package's contents moved with every bump; zero does not, because a default added upstream arrives through the same boundary |
| `spec-subschema-parity.test.ts` ×6 | re-exports **are** the raw spec object | are `stripImportedDefaults(spec)` — plus **both** controls: the 3 with nothing to strip are still reference-equal to the raw object, the 4 that carried a default are provably not |
| `spec-subschema-parity.test.ts` `prefix` | asserted the substitution (`{ field }` → `{ field, type: 'text' }`) | asserts the author's document, and the round-trip of one that writes `type` |
| `object-view-unmirrored-keys-7779.test.ts` ×2 | slot is the raw spec slot; `mode` parses to `'page'` | slot is the entered spec slot; `mode` stays absent — **and every probe is compared against the RAW spec slot too**, which is the measurement that says the strip moved no accept set |
| `report-chart-query-spec-parity.test.ts` | read `valueKey`/`labelKey`/`persist` back out of parse output | asserts they stay **declared** and are no longer written. Its own comment already said the important half: *"a materialised default is indistinguishable from an authored value"* — that indistinguishability is what this card removes |

## Verification — exit codes captured **before** any pipe

| what | result |
|---|---|
| `vitest run packages/types/` | **exit 0** — `Test Files 153 passed (153)`, `Tests 3013 passed (3013)` |
| `pnpm --filter @object-ui/types type-check` (`tsc --noEmit` + `tsconfig.examples.json` + `tsconfig.test.json`) | **exit 0** |
| `pnpm --filter @object-ui/types build` + `check:dist-completeness` | **exit 0** — 128 emitted files verified |
| `pnpm check:spec-symbols` (runs in `ci.yml`) | **exit 0** — 1358 files vs 5050 spec export names |
| `pnpm check` (runs in `lint.yml:481`) | **exit 0** — 628 files analysed |
| `npx eslint .` — **whole repo, not narrowed** | **exit 0** — **4575 files judged**, **0 errors**, 12349 warnings (known non-blocking debt per AGENTS.md; type-aware linting is not enabled, `projectService` count 0) |
| `check:control-bytes` · `self-import` · `phantom-deps` · `unused-deps` · `unreferenced-sources` · `handler-key-reads` · `side-effects-array` · `published-tsconfig-exclude` · `esm-specifiers` · `entry-guard` | **exit 0** each |
| `node scripts/check-changeset-presence.mjs` | **exit 0** — 16 changed files, 1 changeset |
| `check:governed-queue-guard --test` (the 16 changed paths) | **NOT GOVERNED** — an ordinary PR |
| **every package that touches the zod barrel** — `vitest run` over the 20 dirs that reference `@object-ui/types/zod` (`apps/console`, `packages/app-shell`, `cli`, `core`, `fields`, `react`-consumers, all 11 `plugin-*`, `runner`, `scripts`, both `examples/`) | **exit 0** — `Test Files 1926 passed / 2 skipped (1928)`, `Tests 25487 passed / 3 skipped` |
| full suite `vitest run --shard=1/4` | **exit 0** — `696 passed`, 1 skipped (697 files); `9256 passed`, 2 skipped |

⚠️ **`pnpm check` prints 3 warnings** (two `filter-builder` catalog schemas and the vscode JSON schema "did not validate"). **Pre-existing, not this change**: the corpus differential above covers `examples/schema-catalog` and found **0 acceptance differences** across 1,077 documents.

### ⛔ NOT measured — stated rather than implied

- **`check:node-esm-load`** — ran, and the run is **VOID, not red**: the shared `.turbo/cache` replayed `@object-ui/auth` and `@object-ui/react-runtime` from another worktree, and the gate refuses to grade another tree's artifacts (32 of 39 entries loaded clean; the 2 refusals are outside this diff). Needs `--force-build`. ⚠️ This gate is **cron + push-to-`main` only** — no PR run will exist.
- **`check:published-dist`** — not run; `workflow_dispatch` + cron + push-to-`main`, **no run can exist on a PR head**.
- **`check:eager-closure`** — `exit 2`, **PREREQUISITE NOT MET**, not a budget failure: it needs `apps/console/dist/eager-closure.json`, which only a console build writes.
- **`check:readme-exports`** / **`check:spec-floors`** — both `exit 1` with `no-artifact` / "run `pnpm build` first"; **prerequisite not met**, not findings. Both have their own workflows.
- **`Build Docs`** — this diff touches neither `apps/site/` nor `content/`, so its site build would be skipped anyway (objectui#8647).
- **Full 4-shard suite** — shard 1/4 measured green above, and the consumer sweep above covers every package that references the barrel; shards 2–4 in full are CI's `Test (shard N/4)`. ⚠️ The consumer sweep is the reading that matters for a published-behaviour change, and it is on the FINAL head: an earlier sweep was started before the `d1a83a60` rework and discarded unread rather than reported, because a run whose tree changed under it is not a measurement.
- **Browser / dogfood** — none. This is a validator-output change with no rendering path (the census establishes that no renderer reads a parse result).

## Clause ②

**`Clause-②: yes`** — the published validator's returned document changes on those keys. `needs:contract-review` is on the card and on this PR (双载体), verified by read-back.

---

---

### DIFF
--- a/.changeset/8317-strip-imported-defaults.md
+++ b/.changeset/8317-strip-imported-defaults.md
@@ -1,0 +1,59 @@
+---
+'@object-ui/types': minor
+---
+
+The validator stops writing values into an author's document on the keys it imports —
+**this mirror authors no default, imported subschemas included** (objectui#8317,
+director ruling, decision batch #90, 2026-09-08, under the maintainer's standing
+delegation).
+
+Decision batch #69 (objectui#7735) ruled a principle: *a validator validates; it does
+not write values into an author's document.* PR #8299 delivered it for the 41
+`.default()` call sites written in this package's own mirrors. Measured afterwards, **57
+`ZodDefault` nodes were still reachable from the published `@object-ui/types/zod`
+barrel**, every one inside a subschema imported by reference from `@objectstack/spec` —
+so `safeValidateSchema` went on substituting on those keys, with 41 stripped and 57 not
+and no way to tell which was which from the document. Batch #90 ruled that the
+principle holds for **every** key the validator answers, and those 57 are now stripped
+where the spec enters this package (`.removeDefault()`, the established local pattern,
+applied through `zod/imported-defaults.ts`).
+
+**What changes.** `safeValidateSchema` / `validateSchema` return the author's document
+instead of the author's document plus keys they did not write:
+
+```js
+safeValidateSchema({ type: 'object-view', objectName: 'account', navigation: {} })
+// before → navigation: { mode: 'page', preventNavigation: false, openNewTab: false, size: 'auto' }
+// after  → navigation: {}
+```
+
+The affected families: `app`'s `active` / `isDefault` · `object-view`'s
+`navigation.{mode, preventNavigation, openNewTab, size}` · `list-view`'s `sharing.type`,
+`userActions.*`, `addRecord.*`, `appearance.*`, `chart.chartType`, `tabs[].*`,
+`timeline.scale` · `kanban`'s `grouping.fields[].{order, collapsed}` · `page`'s `kind`
+and the whole `interfaceConfig` subtree · dashboard `chartConfig.*`, `header.*` and
+`globalFilters[].scope` · `object-gallery`'s `gallery.*` · `contextSelectors[].*` ·
+`prefix.type`, `pagination.pageSize`, `selection.type` and the HTTP `method`.
+
+**The accept set does not move, and that is measured, not asserted.** Every one of these
+keys stays omissible — `.default(v)` carries optionality as well as a value, so the
+boundary re-optionalises what `.removeDefault()` hands back. A differential against the
+raw `@objectstack/spec` schemas (a permanent pin, since both sides are importable) plus
+a run over this repository's own 1,077-document corpus found **zero** documents whose
+acceptance changed, on the tolerant face and on the strict authoring face alike. Keys,
+value vocabularies and every `.refine()` / `.superRefine()` the spec installed are
+carried through unchanged.
+
+**Migration.** If you read a value off `result.data` and relied on it being present
+without having written it, read it off your own fallback instead — batch #69 already
+ruled that the renderer's fallback is *the* authoritative default, and the renderers in
+this workspace already carry theirs (`navigation?.mode ?? 'page'`,
+`userActions.search !== false`). A census of every consumer of this barrel found no such
+read: three production importers, one of which reads `result.data` at all, and it reads
+only `type` / `id` / `label` / `title` / `children` — none of which carries a default on
+any of the 107 component arms.
+
+⛔ **Not taken:** changing `@objectstack/spec` itself (1,546 call sites on another
+repository's release train). This is reversible into it — every strip becomes a no-op
+the day the spec adopts the same principle, because the boundary is the identity
+function on a subtree with nothing to strip.
--- a/packages/types/src/__tests__/imported-defaults-8317.test.ts
+++ b/packages/types/src/__tests__/imported-defaults-8317.test.ts
@@ -43,6 +43,7 @@ import { describe, it, expect } from 'vitest';
 import { readdirSync, readFileSync } from 'node:fs';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import ts from 'typescript';
 import type { z } from 'zod';
 import {
   AppSchema as SpecAppSchema,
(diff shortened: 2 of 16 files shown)


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** 33/33 green.

This PR had one way to go badly wrong, and it is handled at the one site where it matters.

**`.removeDefault()` drops optionality.** In Zod, `.default(v)` is what makes a member omissible; `z.string().default('x').removeDefault()` is `z.string()` — **required**. A strip that stopped at `.removeDefault()` would have flipped all 57 keys from omissible to mandatory and started rejecting documents that validate today, on a barrel three production packages import. The boundary restores it, and distinguishes the two spellings instead of blanket-wrapping:

```js
const inner = walk(schema.removeDefault());
out = isAlreadyOptional(inner) ? inner : z.optional(inner);
```

with `isAlreadyOptional` reading `_zod.optin === 'optional'`, so `ZodDefault(ZodOptional(T))` (the `.optional().default(v)` spelling) is left alone and a bare `ZodDefault(T)` is re-optionalised. That is why *"the accept set does not move"* is a claim this diff can actually make.

**The test file is built the way I would have asked for and did not have to.** What I checked rather than read:

- **Vacuity is guarded on both branches, before either is exercised** — `CARRIES_DEFAULT.length > 5` and `CARRIES_NONE.length > 1`, titled *"so neither branch below is vacuous"*. The identity half of a walk is exactly where a silent no-op hides.
- **`the RAW spec schemas still carry their defaults after the strip has run`** — the pin against mutating the upstream. Zod schema objects are shared references; a walk that rebuilt in place would have reached into `@objectstack/spec`'s own exports and been invisible from this side.
- **A parallel walk compares node types, `checks` length and union arm counts**, with `${name}${path}: a check was DROPPED by the walk` as the message — so a `.refine()` lost in traversal is named, not merely counted — and then asserts *"the parallel walk covered the whole imported population, not a corner of it"*.
- **A census that no `@objectstack/spec` value read in the mirrors bypasses the boundary**, whose declared exceptions must *still exist and still read a RAW binding* — an exception list that cannot go stale in either direction.
- ⭐ **`the walker docblock's 'lazy' count is re-derived, not quoted`.** A docblock number that a test re-derives is the opposite of the defect class this session has been filing findings about all day.

**The reversibility argument is the right one.** Not touching `@objectstack/spec` (1,546 call sites on another repository's release train) and instead making the boundary *the identity function on a subtree with nothing to strip* means the day upstream adopts the same principle, every strip becomes a no-op and nothing here needs unwinding. That is a fork avoided, not deferred.

**Honest about what is a pin and what was a measurement**: the differential against the raw spec schemas is permanent (both sides are importable), the 1,077-document corpus run was a one-off, and the changeset says which is which rather than implying the corpus is re-run.

⚠️ **One asymmetry, recorded not blocking.** The docblock and changeset both state **57** `ZodDefault` nodes, and nothing asserts 57 — the tests assert `> 5`, `> 1`, and `=== 0` after the strip. That is the same shape as the `lazy` count the neighbouring test explicitly refuses to quote. Here the asymmetry is defensible and I would not change the tests: pinning 57 would tie this repo's suite to another repository's release train and redden on any upstream minor, which is churn rather than signal. But the number should read as a snapshot rather than as something measured continuously — one clause in the docblock, whenever that file is next touched.

The migration note earns its place by naming the fallbacks that already exist (`navigation?.mode ?? 'page'`, `userActions.search !== false`) instead of asserting that consumers will cope, and by backing it with a consumer census: three production importers, one reading `result.data` at all, reading only keys that carry no default on any of the 107 arms.

---

PR #8646, opened 2026-09-08
build(types)!: ship the `./zod` subpath as ONE bundled module (#8598)

os-warren:
Fixes #8598.

Manifest half landed under decision batch #103 (option A). The contract review then returned **FAIL** on `8b8cbfd` for one real defect this PR introduced, plus two owed items; all three are addressed on `d7a81dbbe6ca5bc33a4c92251928cbbdc452220f`.

## ⚠️ The defect the review caught, and where it is fixed

`Test (shard 2/4)` was red on `8b8cbfd` on a diff that touched **no test at all**.

`packages/cli/src/__tests__/cli-bin.test.ts` self-builds `@object-ui/types` when the zod bundle is absent — CI's test job has no build step, by design — and it spawned `pnpm --filter @object-ui/types run build` with the vitest worker's entire environment. Vitest sets `VITEST=true`, children inherit it, all **24** `packages/*` vite configs open with `if (process.env.VITEST) { assertCanonicalVitestInvocation(...) }`, and that guard derives its vitest root from `cwd` — which `pnpm --filter` sets to the package directory. So the guard read a *build* as a vitest run launched from the wrong place and called `process.exit(1)` before the bundler started.

⭐ The build script, the spawning test and the guard are each correct in isolation. This PR created the intersection: it made `@object-ui/types` **the first vite-built package that a test builds from inside vitest**. The cli's own self-build survives only because it builds with `tsup`. On `8e028c2` the build script never loaded the config, so nothing could have surfaced it earlier.

### Fix home chosen: **the spawn**, not the config

| | |
| --- | --- |
| **Chosen** | scrub `VITEST` from the spawned build's env (`BUILD_ENV` in `cli-bin.test.ts`, applied to **both** build spawns in that file — the only two in the repo) |
| **Rejected** | gate the config on vite's `command: 'build'` instead of the variable |

Three reasons, in order of force:

1. **The variable is what is false.** `VITEST` means "vitest is loading this config". In a build the test *starts*, that is untrue. The spawn is the only place that knows its child is a build rather than a test run.
2. ⛔ **The rejected fix is mechanically refused.** `scripts/__tests__/vitest-invocation-guard.test.ts` requires the literal `if (process.env.VITEST) {` + `assertCanonicalVitestInvocation(` shape in **every** `packages/*` vite config. Gating on `command` would red that ratchet and diverge 1 of 24 byte-identical blocks.
[...]

PR #8732, opened 2026-09-09
fix(plugin-detail): the summary chip beside the H1 draws an object value through its own cell renderer

### LINKED ISSUE #8464: bug(plugin-detail): the record page's summary chips render an object-valued field as the literal text `[object Object]` — the one band beside the H1 with no type-aware renderer
Filed by the `domain:ui` PM seat (`session_01YBWFb5YgMU5dw8p2VKj16S`) on behalf of the objectui#8394 dev, who measured it while landing PR #8457 and could not file it (`search_issues` rate-limited; REST 403 from that container). ⛔ Not claimed.

## What

`packages/plugin-detail/src/DetailView.tsx`, the `effectiveSummaryFields.map` render **beside the page H1**.

Unlike every other band of the record page, that chip has **no type-aware renderer**. Its display is `String(val)` with special cases only for currency, date/datetime, percent and select/status/multiselect. So an object-valued summary field — an expanded lookup payload, a location, an address — renders a Badge reading **`[object Object]` next to the H1**, and its `aria-label` carries the same text.

Reachable **both** ways:

- `schema.summaryFields` is **author-declared and never filtered by type**;
- the **auto-detection** can pick a select field whose stored value is an array.

## Relationship to the cards around it

- Same *symptom* as objectui#8395 (`[object Object]` on the clipboard), **different site and different fix surface** — that one was a copy handler with no renderer to ask; this one is a render path sitting next to seven bands that all do have renderers.
- objectui#8394 / PR #8457 **converged this chip's emptiness guard** onto the shared authority but deliberately did **not** touch what an object *displays* here: `hasCellValue` is the non-regressive convergence (it moves whitespace-only strings and nothing else), whereas changing the object display is a display decision.

## Fix shape (not ruled)

- **A — route the chip through `getCellRenderer`**, the way `HeaderHighlight` does. Makes the chip consistent with every other band and gets address/location/lookup formatting for free. Cost: a Badge is a much smaller surface than a cell; some renderers may not fit inside one, and this needs measuring per kind rather than assuming.
- **B — skip object values** when picking and rendering summary fields. Cheapest, and it removes a chip an author explicitly asked for, silently.
- **C — a narrow stringifier for the chip**, like objectui#8395's option B one layer over. Same per-kind product questions, same answer-shopping.

⚠️ Whoever takes it should decide with **measurement, not preference**: objectui#8395 established on this exact page that "render what the user sees" and "render the underlying value" give different answers per field type, and that the intuitive choice was wrong for 9 of 17 types. Do not assume A is free.

## Related

objectui#8394 / PR #8457 (where it was measured) · objectui#8395 / PR #8460 (the same symptom on the clipboard, and the measurement table for this page's field types) · objectui#8376

## Dedup

⚠️ **Not run, declared rather than hidden.** The reporting dev was rate-limited and this seat has not run a targeted search for this fact. **No dedup claim is made.** Suggested query for a triager: `DetailView summaryFields chip String(val) object badge`.

os-justin:
Fixes #8464

## The defect, reproduced on this branch's base

⭐ Leg 1. The card was filed from another dev's measurement on a different branch, so it was re-measured here — `DetailView` rendered on `ed971e8fc` with `summaryFields: ['owner_ref','billing_address','office_location']` and three object values:

| chip | text | `aria-label` |
|---|---|---|
| `owner_ref` (expanded lookup `{id,name}`) | `[object Object]` | `owner_ref: [object Object]` |
| `billing_address` (`{street,city,…}`) | `[object Object]` | `billing_address: [object Object]` |
| `office_location` (`{latitude,longitude}`) | `[object Object]` | `office_location: [object Object]` |

H1 read `Acme Corporation`, so all three pills sat beside a correctly-rendered title. Both halves of the card's claim hold: the text **and** the accessible name carry the placeholder, because the chip builds one from the other.

## Which question the chip asks — measured, not argued

objectui#8395 established on this page that *"render what the user sees"* and *"render the underlying value"* give different answers per field type, and that the intuitive choice was wrong for 9 of 17. **That table is about a clipboard payload.** Re-derived against today's tree (`packages/plugin-detail/src/__tests__/DetailSection.copyObjectValues-8395.test.tsx` still pins it): `date` renders `Mar 4` against a stored `2026-03-04`, `percent` renders `12%` against `0.123`, `image` and `boolean` render no text at all — which is why a *payload* must take the stored value.

The chip is not a payload. Its own existing behaviour answers which side it is on, measured here on `ed971e8fc`:

| field | stored | the chip printed |
|---|---|---|
| `currency` | `1234.5` | `$1,235` |
| `date` | `'2026-03-04'` | `Mar 4, 2026` |
| `datetime` | `'2024-07-04T07:00:00.000Z'` | `Jul 4, 2024, 7:00 AM` |
| `select` | `'won'` | `Closed Won` |

Every one is the **seen** face, never the stored one. So the display authority for a kind the chip does not itself format is that kind's own cell renderer — **option A**, the route `HeaderHighlight` already takes one band below. This is the chip's measured behaviour, not a preference about what a summary "ought" to show.

## ⭐ The per-kind Badge measurement — A is **not** free

The card's warning was taken literally: every one of the **53 registered field types** was rendered through `getCellRenderer` **inside the real chip Badge** — same element, same `variant`, same class string `DetailView` uses — against the object value `{ id: 'acct-1', name: 'Acme Corp' }`, and the DOM the pill received was **counted**. The instrument is committed as `summaryChip.badgeFitCensus-8464.test.tsx`, so the table stays runnable.

**38 of 53 fit.** They draw plain inline text: `Acme Corp` for the nameable families; the JSON literal for `location` / `geolocation` / `address` / `json` / `object` / `composite` / `record`, behind objectui#8481's declared json-literal fence; a value-independent face for `password` / `secret` / `vector` / `grid`.

**15 of 53 do not**, in four measured classes:

| class | types | what the pill received |
|---|---|---|
| a pill inside a pill | `select` `status` `multiselect` `radio` `checkboxes` `tags` | `SelectCellRenderer`'s own `Badge` — one `rounded-full` node nested inside the chip's own |
| an avatar composite | `user` | **two** `rounded-full` nodes, and the initials glued onto the name: `ACAcme Corp` |
| an image and no text | `image` `avatar` `signature` | an ``, `textContent === ''` — nothing for the chip to say and nothing for its accessible name |
| a "No value" face | `boolean` `toggle` `datetime` `repeater` (shared `EmptyValue`), `date` (`formatDate`'s own em-dash, objectui#8581) | the chip is drawn only **after** `hasCellValue` called the value FILLED; a renderer answering "empty" one band later re-opens exactly the cross-band contradiction objectui#8394 closed |

⚠️ The avatar row is the one the card warned about: `.rounded-full` matches **two** nodes per avatar (Radix `Avatar.Root` **and** `AvatarFallback`). This census **counts** rather than navigates, so it saw both; the census asserts the count is exactly 2.

### The decision this table produces

**A for the 38 kinds that fit, with a stated rule for the 15 that do not** — the shape the card anticipated, not a hedge.

The stated rule is **not** a stringifier written for this chip (objectui#8395's option C, "answer-shopping"). The refused kinds fall to `coerceToSafeValue`, `@object-ui/fields`' single documented answer to *"what text does a cell draw for a value that is not a string"*. objectui#8596 ruled the option families and `user` onto exactly that text for an object value, so for **7 of the 15** the chip is byte-equal to its own cell. Following objectui#8596's method: read the existing authority per family, never invent a renderer-side fallback (AGENTS.md #0.1).

⛔ Option **B** was refused by the card and is not in this change: no object value is dropped. The `THE EMPTINESS CLASSIFICATION` cases are red for it.

### Why the switch fires where it does

On the **defect's own signature** — `display.includes('[object Object]')` after the existing four families have had their turn — not on a type guess. `Number({})` is `NaN`, `new Date({})` is Invalid and the option lookup falls back to `String(val)`, so all four formatted branches land on the same test rather than needing four of their own. A value the string path already rendered (`['a','b']` → `a,b`, every scalar) is byte-for-byte untouched, and `SCALARS AND SCALAR ARRAYS` pins that.

### The one cost, stated

A renderer-backed chip carries **no** `aria-label`: `aria-label` overrides content, so it would hide the very value the branch exists to show. Its accessible name is composed from content instead — an `sr-only` field-name prefix plus the renderer's text — giving the same `field: value` shape as before, now true. String-path chips keep their `aria-label` unchanged, so objectui#8394's `chipFor` navigation and its `'stage: Won'` assertion are untouched. Every chip additionally carries `data-summary-chip="<field>"` as one handle for both paths.

## Non-regression: the emptiness classification did not move

objectui#8394 / PR #8457 converged this chip's guard onto `hasCellValue` as *"the non-regressive convergence — it moves whitespace-only strings and nothing else."* This change is a display decision and leaves it alone; all four of its answers plus the empty-object case are re-pinned through the chip (`THE EMPTINESS CLASSIFICATION`, 5 cases). The whole existing `packages/plugin-detail/` suite — **1350 tests** — passes unchanged.

## Caricatures — every pin observed red

⚠️ An assertion never observed to fail has not been tested, so each was run. Read sites were mutated (never a pin), from a **committed** tree, with a `trap … EXIT INT TERM`, absolute paths, the mutation proven on disk in both directions (removed-anchor count `0`, injected-anchor count ≥ 1, `git hash-object` ≠ the HEAD blob) **plus** a line-total gate, and restoration proven **by state** (`git diff HEAD` empty) after every leg. Per-test classification comes from vitest's **JSON reporter**.

Vitest aliases `@object-ui/fields` → `packages/fields/src` (`vitest.config.mts:507`), so a source mutation *is* what the test executes — no `dist` is in the path for these `*.test.tsx` files.

| leg | mutation (read site) | result |
|---|---|---|
| baseline | none | 26 / 26 pass |
| 1 — *render nothing for every object* (the fenced **B**) | `DetailView.tsx`: return `null` for any object value | **13 of 26 RED** |
| 2 — *render every value through the cell renderer regardless of kind* | `DetailView.tsx`: `ChipCellRenderer = getCellRenderer(...)` unconditionally | **19 of 26 RED** |
| 3 — *refuse the renderer for every kind* | `summaryChipRenderers.ts`: `chipTakesCellRenderer` → `false` | **7 of 26 RED** |
| 4 — *every type answers what `text` answers* | `fields/src/index.tsx`: `getCellRenderer` → `TextCellRenderer` | census: 2 of 5 RED |
| 5 — *every type draws the shared No-value affordance* | `fields/src/index.tsx`: `getCellRenderer` → `EmptyValue` | census: 3 of 5 RED |

Both directions the card asked for redden loudly. Leg 2 is the one that refuses "just route everything through the cell renderer": it reds all four formatted families, the percent bar, and the scalars.

Leg 5 initially left `THE FIT SIDE IS LOAD-BEARING` **green** — it had no assertion that mutation could break. That is reported rather than papered over: the case was given the discriminating assertion (a fitting kind draws its value, never the `data-slot="empty-value"` affordance) and re-run, and it is red under leg 5 now.

## Verification

| what | command | verdict |
|---|---|---|
| new pins | `pnpm exec vitest run …summaryChip.objectValue-8464.test.tsx …summaryChip.badgeFitCensus-8464.test.tsx` | `Test Files 2 passed (2)` · `Tests 26 passed (26)` |
| the whole package | `pnpm exec vitest run packages/plugin-detail/` | `Tests 1350 passed (1350)` |
| affected packages | `pnpm exec vitest run packages/plugin-detail/ packages/plugin-calendar/ packages/plugin-gantt/ packages/plugin-kanban/ packages/runner/ packages/app-shell/` | `Test Files 945 passed (945)` · `Tests 8699 passed | 1 skipped (8700)` — the six affected workspace packages `turbo ls --affected` names (the other five affected entries are apps/examples, which CI covers) |
| types | `pnpm --filter '@object-ui/plugin-detail^...' build && pnpm --filter @object-ui/plugin-detail type-check` | exit 0 — and `type-check` is `tsc --noEmit && tsc -p tsconfig.test.json`, so the new test files are in a checked program, not excluded |
| gates | `check:control-bytes` `check:doc-fences` `check:phantom-deps` `check:unused-deps` `check:self-import` | all exit 0 |
| control bytes | `grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'` over all six changed files | no match |
| changeset | `node scripts/check-changeset-presence.mjs` | `✅ 4 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` — the gate's own verdict line. ⚠️ `skip-changeset` is a phantom label here; this is a user-visible fix, so it declares a real `patch`, not an empty frontmatter |
| governed surface | `node scripts/check-governed-queue-guard.mjs --test <the 6 paths>` | `✅ NOT GOVERNED — 6 path(s) checked against 5 governed surface(s); none matched.` |

**Lint, narrowed and the narrowing measured.** `pnpm exec eslint <the 4 code files> --format json` → exit **0**, **4** files (count read from the JSON output, not asserted), **0 errors**, 82 warnings — all `@typescript-eslint/no-explicit-any` and two pre-existing `react-hooks` warnings that `DetailView.tsx` already carried. The narrowing is a measurement rather than a skipped run because **type-aware linting is not enabled**: `eslint.config.js` has zero `project` / `projectService` entries, so nothing in this diff can move the verdict on a file it did not touch. CI's repo-wide `eslint .` is the run that covers the population, and `.github/workflows/lint.yml` deliberately sets no `--max-warnings`, so warnings are not a failure there either. ⛔ `--no-inline-config` was **not** used — that is an objectstack convention and manufactures errors this repo's CI does not have.

## Scope

⚠️ **This card's own dedup had never been run** (its reporter was rate-limited and REST answered 403), so it was run here rather than inherited as a clean bill. Channel: MCP `search_issues` — repo-scoped REST `/search/issues` is refused in this container ("sessions are bound to their configured repositories"), so the suggested query was run through MCP. **Its control fired**: a query for this defect returned #8464 itself as the top hit, so the zero is real and not one of this tool's false zeros. No duplicate. #4054 is adjacent, not a duplicate — see the limit below.

Two defects found on these lines and filed rather than ridden along:

- **#8728** — a `percent` summary chip states two different percentages: its text reads `0.123%` while its own bar fills to 12.3%, and the body grid draws `12%` for the same value.
- **#8729** — the summary chip names its field by the raw column name (`owner_ref: …`), the one band beside the H1 that does not resolve a label. Not folded in here because moving that half means editing the instrument objectui#8394 owns.

## ⚠️ What this change does not measure

**Pixels.** happy-dom reports `clientWidth: 0` and never fires container-size effects, so the census measures what DOM the pill *hosts*, not what it *clips*. That half is **#4054** — "a `record:highlights` chip clips multi-element cell renderers with NO ellipsis" — which says in its own words that "a DOM-only probe reports this surface healthy" and needs `getBoundingClientRect()` plus a screenshot. Its analysis also records that `lookup` / `user` / `file` carry their own `truncate` and are unaffected, and names the `MapPin`-plus-text row — this repo's `location` — as a candidate. This change makes `location` reachable in a summary chip, so it is now in #4054's population; the structural verdicts here stand either way.

## Collisions

`git ls-remote --heads origin 'gh-readonly-queue/*'` at branch time listed PRs #8646, #8668, #8706, #8707, #8711. None touches `packages/plugin-detail/**` (file lists read per PR). PR #8685 (`plugin-detail`) had landed, so the package was free.

---


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** 33/33 green.

The thing that makes this PR right is that arm A was not taken on the argument that it was the obvious one. *Which question the chip asks* was settled by re-deriving the chip's existing behaviour: it already prints `$1,235` for a stored `1234.5`, `Mar 4, 2026` for `'2026-03-04'`, `Closed Won` for `'won'` — the **seen** face every time — so the display authority is the field's own cell renderer, and objectui#8395's "seen vs stored" fork was answered from the code rather than from taste.

Three things I re-measured rather than read.

**1. The fallback is not the defect in another costume.** `coerceToSafeValue` resolves an object to `name || label || externalId || id || _id || '[Object]'` — never `[object Object]`. So the 15 refused kinds get a real answer, and it is `@object-ui/fields`' existing one rather than a chip-local stringifier. objectui#8596 already ruled 7 of those 15 onto exactly that text, so for those the chip is byte-equal to its own cell. That is what keeps this inside AGENTS.md #0.1 instead of shopping for a third answer.

**2. The census closes the trap it names.** A matrix pin over an empty set passes, and this one guards against that *and meta-tests the guard*:

```
it('THE CENSUS — the guard refuses a short or EMPTY set (a census over nothing passes)')
  expect(() => assertCensusComplete([], REGISTERED_TYPE_COUNT, 'census')).toThrowError(…)
```

plus no-duplicate-types, `measured === 53` ("every row was rendered, not skipped"), and `fitting.length === 38`. The `⭐ THE SET MATCHES THE MEASUREMENT` case derives `CHIP_UNFIT_RENDERER_TYPES` from the rendered table rather than from memory, so the constant cannot drift away from what the pills actually did.

**3. The gate is the defect's own signature, and it sits in the right place.** `display.includes('[object Object]')` is tested *after* the four formatted branches, so `Number({})` → `NaN`, `new Date({})` → Invalid and the option-lookup fallback all funnel into one test instead of four — and a value the string path already renders (`['a','b']` → `a,b`, every scalar) is untouched byte-for-byte.

**The `aria-label` removal is correct for a reason worth stating on the record**: an `aria-label` on the Badge would *override* the rendered element's own accessible content — it would hide the very value the renderer branch exists to expose. Composing the same `field: value` name from an `sr-only` prefix plus content is the only spelling that keeps both the name and the value. And `data-summary-chip` going onto all three branches is what lets a test address a chip without depending on which branch drew it.

⭐ **Reported honestly, and it is the part I trust most**: leg 5 first left `THE FIT SIDE IS LOAD-BEARING` **green** — an assertion never observed to fail. It was given a discriminating assertion (a fitting kind draws its value, never the `data-slot=empty-value` affordance) and re-run red. A dev that reports its own green leg as a defect is a dev whose red legs mean something.

⚠️ **One finding this review turned up, filed rather than sent back — it cannot be closed inside this PR.** The census population is **frozen at 53 and the default is permissive**. `REGISTERED_TYPE_COUNT` and `CENSUS` are hard-coded, and nothing reconciles them against the live registry — `getCellRenderer` dispatches on `fieldRegistry` plus a `standardMap` that is a **function-local `const`** (`packages/fields/src/index.tsx:2608`) and therefore not enumerable from outside. So when a 54th type is registered: the census still passes at 53, the new type is absent from `CHIP_UNFIT_RENDERER_TYPES`, and `chipTakesCellRenderer` returns `true` by default — routing it to its own renderer unmeasured. The drift is silent **and in the unsafe direction**: if that renderer draws a nested pill, a bare ``, or a "No value" face, all four defects this census enumerates return with nothing red. Closing it means exposing the standard map from `@object-ui/fields`, which is another package's surface. Filed as a follow-up; nothing to do here.

**On the open question raised in the report** — the dispatch text does read as self-contradictory (no model identifier in a commit message, then a mandated `Co-Authored-By` trailer naming one). Reading **A** is right and is now the standing ruling: the trailer is the named exception, spelled out verbatim by the session attribution notice; the prohibition covers every surface it does not name — PR title and body, code comments, changesets, and any other pushed artefact. Nothing in this PR carries one outside the trailer.

The two by-products (#8728, the percent chip stating two different percentages; #8729, the chip naming its field by the raw column name) were correctly kept out: the second one moves an instrument objectui#8394 owns.

---

PR #8741, opened 2026-09-09
fix(plugin-grid): skip and diagnose an unusable `bulkActionDefs` member instead of crashing the selection bar

### LINKED ISSUE #8730: finding(plugin-grid): a bare-string member in `object-grid.bulkActionDefs` throws during render and takes the whole selection bar down
## 复现(实测于 `origin/main` @ `2f881a90`,objectui#8071 的成员 pin 工作中发现)

`object-grid` 的 `bulkActions` 与 `bulkActionDefs` 是同一个 affordance 的两套词汇:

- `bulkActions` 成员 = **裸动作名字符串**,`resolveBulkActions` 拿它去 `objectDef.actions` 解析并提升成 def;
- `bulkActionDefs` 成员 = **完整 `BulkActionDef` 对象**,原样使用。

把其中一个的词汇写进另一个,两边都没有任何东西会拒绝:注册声明两个都是 `type: 'array'` 且无 `of`,
spec 的 `ComponentPropsMap['object-grid']` 两行都是 `z.array(z.unknown())`,JSON 视图上 `tsc` 也看不见。

**方向一(`bulkActions: [{ name: 'approve' }]`)静默丢弃** —— `resolveBulkActions` 的
`if (typeof name !== 'string' || name === '') continue;` 跳过它,无任何诊断。这是可以接受的失败形态。

**方向二(`bulkActionDefs: ['approve']`)不是静默,是崩溃:**

```
TypeError: Cannot read properties of undefined (reading 'replace')
 ❯ formatActionLabel packages/plugin-grid/src/components/RowActionMenu.tsx:43:17
 ❯ BulkActionButton packages/plugin-grid/src/components/BulkActionBar.tsx:64:21
```

`Array.isArray(schema.bulkActionDefs)` 为真 ⇒ 字符串原样进入 `rawAuthored` ⇒
`BulkActionBar` 用它渲染 `BulkActionButton`,`def.name` 是 `undefined`,
`def.label ?? formatActionLabel(def.name)` 在 render 中抛出。作者第一次勾选多行时整条选择条就没了
(React 还会先抱怨一次 `Each child in a list should have a unique "key" prop`,因为 `key={def.name}` 也是 `undefined`)。

## 已有的失败探针

objectui#8071 的成员 pin 把这条行为**按当前形态**钉住了,以免它被误读成方向一的静默:

`packages/plugin-grid/src/__tests__/bulkActionMembers-8071.test.tsx` 第 5 行断言
`renderAndSelectAll({ bulkActionDefs: ['approve'] })` 必须 reject 并匹配上面的 `TypeError`。
⚠️ 那一条是**记录现状,不是契约**:本卡的修复落地时它会转红,必须在同一笔里改写成修复后的行为。

## 为什么值得单开一卡

`BulkActionBar` 渲染的是**作者可写的元数据**,不是内部数据结构。一个 spec 与注册面都接受的成员形状
不应该让渲染抛错 —— 至少要退化成"跳过该成员"(与方向一对齐),更好的是留一条具名诊断。
两个方向的处置不一致本身就是缺陷:同一类作者错误,一边静默一边崩。

Refs: objectui#8071(发现处)· objectui#8068(成员 pin 判据)· objectui#3002 / objectui#3139(两套词汇的来源)

os-justin:
Fixes #8730

`object-grid` has two vocabularies for one affordance, and nothing refused a member
written in the wrong one. `bulkActions` members are bare action NAMES resolved against
`objectDef.actions`; `bulkActionDefs` members are full `BulkActionDef` OBJECTS used as
authored. Both keys are registered `type: 'array'` with no `of`, both spec rows are
`z.array(z.unknown())`, and a JSON view is invisible to `tsc`.

Writing a bare name into `bulkActionDefs` did not fail quietly — it crashed the render:

```
TypeError: Cannot read properties of undefined (reading 'replace')
 at formatActionLabel (packages/plugin-grid/src/components/RowActionMenu.tsx:43:17)
 at BulkActionButton (packages/plugin-grid/src/components/BulkActionBar.tsx:64:21)
```

The author's first multi-row selection lost the entire selection bar — count, Clear and
every well-formed sibling def with it.

## What changed

`resolveBulkActions` skips any `bulkActionDefs` member that is not an object carrying a
non-empty string `name`, and `ObjectGrid` reports the skip once, by address.

**"Usable" is defined by what the renderer reads**, not by a fresh opinion about what a
def ought to carry: `BulkActionBar` uses `def.name` twice — as the React `key` of the
button it maps to, and as `formatActionLabel`'s argument when the def declares no
`label`. Every other `BulkActionDef` key is optional at the read site, so `name` is the
whole predicate — and that one test covers the reported bare string plus, identically,
`null`, a number, `{}` and `{ name: '' }`.

**The guard site.** `resolveBulkActions`'s `rawAuthored` is the single point where the
authored array becomes the list the bar maps over: its `defs` return flows to
`ObjectGrid`'s `bulkActionDefs`, to `BulkActionBar`'s `actionDefs`, to `permittedDefs`,
to the `.map()` that reads `key={def.name}` and `def.label ?? formatActionLabel(def.name)`.
Guarding there means the key and the label are read off the same validated def, and
`BulkActionBar` keeps reading `def.name` unconditionally because by that line it can.
There is one predicate with two readers — the filter and the message — not a check
scattered across the render path.

**Referential identity is preserved** (`every` before `filter`): the `defs` contract
promises the authored array by reference when nothing folds in, and an always-allocating
filter would break that for every clean view.

**Not a coercion, deliberately.** `'approve'` is not lifted into `{ name: 'approve' }`
and resolved the way `bulkActions` is. That would make the two vocabularies
interchangeable — a product change to what a `bulkActionDefs` member means (objectui#3002
/ objectui#3139 made them distinct on purpose), not a crash fix.

**The diagnostic** rides the channel `ObjectGrid` already owns for "you declared it, the
renderer dropped it" (one `console.warn` prefixed `[ObjectUI] ObjectGrid TOPIC:`, a
`useEffect` keyed on the schema slice — the same shape as the columns diagnostic and the
export-format warning, not a third one beside them):

```
[ObjectUI] ObjectGrid bulkActionDefs: object-grid (objectName: 'os_invoice') - 1 of 3
authored bulk-action defs cannot be rendered and is skipped (2 still render).
  - bulkActionDefs[0]: the entry is a string ('approve'), not a def object - this key's
    members are full `BulkActionDef` objects, used as authored. Write
    `{ name: 'approve', operation: 'custom' }` here, or move the bare name to
    `bulkActions`, which resolves it against the object's declared actions and promotes
    the match.
  A `bulkActionDefs` member must be an object with a non-empty string `name`: the
  selection bar uses it as the button's React key AND as the source of its label.
```

**Out of scope, by the dispatch ruling:** the mirror direction
(`bulkActions: [{ name: 'approve' }]`, stepped over by `resolveBulkActions`'s
`typeof name !== 'string'` guard) keeps its existing silent skip. Row 9 of the new pin
asserts that silence, so a diagnostic leaking into it would be caught.

## Verification — by content

Everything below is quoted from the vitest JSON reporter (never the text reporter: a
`TypeError` at module scope reads as a suite death, not a failing assertion), run from
the repo root with file paths.

**Reproduction, on the base tree** (`a9bc02996`, source blobs proven identical to
`HEAD:` by `git hash-object` before the run): 8 of the 10 new pins RED, including the
reported stack verbatim. Three distinct pre-fix failure modes were measured, not one:
`'approve'` / `42` / `{}` threw in `formatActionLabel`; `null` threw one frame earlier at
`key={def.name}` (`Cannot read properties of null (reading 'name')`); and `{ name: '' }`
did **not** throw at all — it rendered a nameless, unlabelled button
(`expected [ 'bulk-action-' ] to deeply equal []`).

**The discriminating axis.** "Skip the bad member" is also satisfied by an implementation
that skips everything, so both caricatures were run as mutations of the committed fix,
each with an `EXIT INT TERM` trap, each proven to have reached disk (deleted-anchor
count 0, injected count 1, blob differs from `HEAD:`) and each restored with
`git checkout HEAD -- ABSOLUTE_PATH` proven by an empty `git diff HEAD` and a matching
`git hash-object`:

| pin | `isUsable...` forced `false` (skip everything) | forced `true` (skip nothing) |
|---|---|---|
| 1-5 the unusable class, 0 buttons | green | **RED** (the crash returns) |
| 6 mixed list renders exactly the 2 good defs, in order | **RED** (`expected [] to equal [ 'bulk-action-archive', … ]`) | **RED** |
| 7 clean list untouched + `defs` by reference | **RED** | green |
| 8 diagnostic fires by name / does not fire when clean | **RED** (no-fire leg: `expected [ Array(1) ] to equal []`) | **RED** (fire leg) |
| 9 the two directions still fail differently | green | **RED** |
| 10 a name in the RIGHT key still promotes | green | green |

Rows 6, 7 and 8's no-fire leg are what "skip everything" cannot pass; rows 1-5, 8's fire
leg and 9 are what "skip nothing" cannot pass. Rows 7 and 10 are the untouched-behaviour
anchors, and are green under "skip nothing" precisely because that mutation is the base
behaviour — its red set is byte-identical to the base-tree red set, which is an
independent check that the mutation reproduced the defect rather than a new one.

**Green** (all at `b9b0abeab`, a clean tree):

- new pins: 10 tests, 10 passed, 0 failed
- `packages/plugin-grid/`: 347 files, 1055 tests, 1055 passed, 0 failed
- `packages/plugin-view/ packages/plugin-designer/ packages/plugin-list/ examples/schema-catalog/`: 546 files, 3488 tests, 3488 passed, 0 failed
- `packages/app-shell/`: 2156 files, 6406 tests, 6405 passed, 0 failed (1 skipped)
- `turbo run type-check --filter=@object-ui/plugin-grid` (builds the dependency closure
  first): exit 0. It earned its place — it caught four `TS7006`s in the new pins that
  vitest was green through; fixed in the second commit.
- `node scripts/check-control-bytes.mjs`: `OK (scanned 6922 tracked text file(s))`; plus a
  direct `grep -naP` control-byte scan of the four changed files, zero hits.
- `node scripts/check-changeset-presence.mjs`: exit 0, declares the changeset.
- `node scripts/check-governed-queue-guard.mjs --test` on all four paths:
  `NOT GOVERNED - 4 path(s) checked against 5 governed surface(s); none matched.`

**Lint, narrowed and declared.** `eslint packages/plugin-grid --format json` — exit 0,
**155 files** linted (the count read from the JSON output, the population from eslint's
own flat config, not from a guess), **0 errors**. The 99 files carrying warnings are
pre-existing `no-explicit-any` / `react-refresh` noise; the three warnings on the new pin
file are the `as any` schema casts every sibling grid test in this package carries. The
rest of the farm is CI's: this repo's root `lint` is `turbo run lint` per package.
The narrowing is safe to state because this config is **not type-aware** — no
`projectService` and no `parserOptions.project` anywhere in `eslint.config.js` — so this
diff cannot move the verdict on any file it did not touch.

**NOT MEASURED**, deliberately: `check:doc-snippets`, `check:doc-examples`,
`check:readme-exports` and `check:eager-closure` were not run. They need a fully built
tree and `resolveBulkActions` is not exported from `packages/plugin-grid/src/index.tsx`,
so the two new exports are package-internal and reach no published surface. CI owns them.

## A premise in the card that measurement contradicted

The dispatch says `packages/plugin-grid/src/__tests__/bulkActionMembers-8071.test.tsx`
"currently asserts the render rejects with that exact `TypeError`" and must be rewritten
in this same commit. **That file is not on `main`.** It exists only on the unmerged
branch `origin/claude/issue-8071-member-pins-near-miss-four` (commit `e4b19de6f`), which
`git merge-base --is-ancestor` reports is not an ancestor of `main` — a reading confirmed
against a shallow-clone false negative two ways: `git rev-parse
--is-shallow-repository` answers `false`, and a control commit known to be in this
history answers exit 0 on the same check.

So there is nothing here to rewrite, and creating that path on this branch would only
manufacture an add/add conflict with the other PR. Instead:

- **What it protected survives** as row 9 of the new pin file — that direction two is
  *not* direction one's silent drop. Before this change that distinction was
  crash-vs-silence; it is now diagnostic-vs-silence, and both legs are asserted together
  so they cannot drift apart.
- **Its row 5 will go red** whichever of the two PRs lands second, and must be rewritten
  there. That is visible rather than silent, but it is a real cross-PR interaction and
  should not be discovered by the merge queue.

Its **row 2** was also the measurement behind leaving direction one alone: it asserts
`expect(warn).not.toHaveBeenCalled()` for `bulkActions: [{ name: 'approve' }]`, i.e. an
in-flight pin *does* assert the current silence. Nothing on `main` asserts it either way.
Direction one is therefore untouched, as the ruling ordered, and row 9 now pins that
silence on `main` too.

---

### DIFF
--- a/.changeset/8730-bulk-action-defs-unusable-member.md
+++ b/.changeset/8730-bulk-action-defs-unusable-member.md
@@ -1,0 +1,55 @@
+---
+'@object-ui/plugin-grid': patch
+---
+
+`object-grid`: a `bulkActionDefs` member that is not a usable def is skipped and
+diagnosed, instead of taking the whole selection bar down (objectui#8730).
+
+`bulkActions` and `bulkActionDefs` are one affordance authored in two vocabularies —
+`bulkActions` members are bare action NAMES resolved against `objectDef.actions`,
+`bulkActionDefs` members are full `BulkActionDef` OBJECTS used as authored — and
+nothing refused a member written in the other one. Both keys are registered
+`type: 'array'` with no `of`, both spec rows are `z.array(z.unknown())`, and a JSON
+view is invisible to `tsc`.
+
+Writing a bare name into `bulkActionDefs` did not fail quietly, it crashed:
+`Array.isArray(schema.bulkActionDefs)` is true, the string travelled into the authored
+list untouched, `BulkActionBar` rendered a button for it, and
+`def.label ?? formatActionLabel(def.name)` threw
+`TypeError: Cannot read properties of undefined (reading 'replace')` **during render**.
+The author's first multi-row selection lost the entire selection bar — count, Clear and
+every well-formed sibling def with it. `key={def.name}` was `undefined` too, so React
+logged a duplicate-key warning on the way down.
+
+`resolveBulkActions` now skips any member that is not an object carrying a non-empty
+string `name`. "Usable" is defined by what the renderer actually reads: `name` is both
+the React `key` and `formatActionLabel`'s argument, so that one test covers the reported
+bare string and, identically, `null`, a number, `{}` and `{ name: '' }`. The guard sits
+at the single point where the authored array becomes the list the bar maps over, so the
+`key` and the label are read off the same validated def.
+
+The skip is not silent. `ObjectGrid` reports it once per authored array through the
+channel it already uses for "you declared it, the renderer dropped it" — one
+`console.warn` prefixed `[ObjectUI] ObjectGrid bulkActionDefs:` — naming the block, the
+index, what was seen, and what to write instead:
+
+```
+[ObjectUI] ObjectGrid bulkActionDefs: object-grid (objectName: 'os_invoice') — 1 of 3
+authored bulk-action defs cannot be rendered and is skipped (2 still render).
+  • bulkActionDefs[0]: the entry is a string ('approve'), not a def object — this key's
+    members are full `BulkActionDef` objects, used as authored. Write
+    `{ name: 'approve', operation: 'custom' }` here, or move the bare name to
+    `bulkActions`, which resolves it against the object's declared actions and promotes
+    the match.
+```
+
+**Not a coercion, deliberately.** A bare `'approve'` is not lifted into
+`{ name: 'approve' }` and resolved the way `bulkActions` is. That would make the two
+vocabularies interchangeable — a product change to what a `bulkActionDefs` member means
+(objectui#3002 / objectui#3139 made them distinct on purpose), not a crash fix.
+
+**Behaviour changes for authors** beyond the crash: a `{ name: '' }` member used to
+render a nameless, unlabelled button with an empty React key, and now renders nothing.
+Well-formed defs are untouched — a mixed list renders exactly its usable members, in
+order, and a clean array is still returned by reference. The mirror direction
+(`bulkActions: [{ name: 'approve' }]`) keeps its existing silent skip.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -47,7 +47,7 @@ import { resolveRowCrudAffordances, resolveRowRecordCrudAffordance } from './row
 import { useRecordCrudVerdicts } from './hooks/useRecordCrudVerdicts';
 import { resolveLegacyRowActions } from './resolveLegacyRowActions';
 import { applyRelationalMeta } from './relationalMetaKeys';
-import { resolveBulkActions } from './resolveBulkActions';
+import { resolveBulkActions, describeUnusableBulkActionDefs } from './resolveBulkActions';
 import { partitionBulkRows } from './bulkEligibility';
 import { resolvesToDataColumn, describeUnresolvedColumns } from './columnSpellingDiagnostics';
 import { RowActionMenu, formatActionLabel } from './components/RowActionMenu';
@@ -2225,6 +2225,30 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
     });
     if (message) console.warn(message);
   }, [schema.columns, columnDiagnosticBlockType, schema.objectName, columnDiagnosticLabel]);
+
+  // [objectui#8730] The same channel, for the sibling failure on
+  // `bulkActionDefs`. A member that is not a usable def (a bare action name —
+  // the OTHER key's vocabulary — or `null`, a number, `{}`, `{ name: '' }`) is
+  // skipped by `resolveBulkActions`; before that guard it reached the bar and
+  // `formatActionLabel(undefined)` threw during render, so the author's first
+  // multi-row selection lost the whole selection bar.
+  //
+  // The skip alone would only relocate the failure into silence — which is what
+  // the mirror direction already does (`bulkActions: [{ name: 'approve' }]`,
+  // stepped over by the `typeof name !== 'string'` guard). Saying which member
+  // was dropped, and that a bare name belongs in `bulkActions`, is what makes
+  // this a diagnosis rather than a quieter version of the same defect. One
+  // `console.warn` per authored array, keyed on it — NOT a second guard: the
+  // predicate lives once, in `resolveBulkActions`, and this reads it.
+  const bulkDefsDiagnosticSlice = (schema as { bulkActionDefs?: unknown }).bulkActionDefs;
+  useEffect(() => {
+    const message = describeUnusableBulkActionDefs(bulkDefsDiagnosticSlice, {
+      blockType: columnDiagnosticBlockType,
+      objectName: schema.objectName,
+      label: columnDiagnosticLabel,
+    });
+    if (message) console.warn(message);
+  }, [bulkDefsDiagnosticSlice, columnDiagnosticBlockType, schema.objectName, columnDiagnosticLabel]);
 
   const generateColumns = useCallback((): ObjectGridColumnDraft[] => {
     // Map field type to column header icon (Airtable-style)
--- a/packages/plugin-grid/src/__tests__/bulkActionDefsUnusableMember-8730.test.tsx
+++ b/packages/plugin-grid/src/__tests__/bulkActionDefsUnusableMember-8730.test.tsx
@@ -1,0 +1,319 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `object-grid.bulkActionDefs` — a member that is not a usable def is SKIPPED
+ * and DIAGNOSED, never rendered (objectui#8730).
+ *
+ * ## What was wrong
+ *
+ * `bulkActions` and `bulkActionDefs` are one affordance authored in two
+ * vocabularies: `bulkActions` members are BARE ACTION NAMES resolved against
+ * `objectDef.actions`; `bulkActionDefs` members are FULL `BulkActionDef`
+ * OBJECTS used as authored. Nothing refuses a member written in the other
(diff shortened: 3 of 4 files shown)


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** 25 of 33 green with shards, Type Check and Lint still running; nothing failed.

**First, a correction that is mine.** My dispatch instructed you to rewrite `bulkActionMembers-8071.test.tsx` "in the same commit". That file **does not exist on `main`** — I took the card's claim as fact and passed it through as an instruction without checking. I have since verified it independently: `git cat-file -e origin/main:packages/plugin-grid/src/__tests__/bulkActionMembers-8071.test.tsx` fails, and the file lives only on `claude/issue-8071-member-pins-near-miss-four`.

The way you established that is the part worth keeping: an `--is-ancestor` exit-1 is a false negative on a shallow checkout, so you controlled it — `--is-shallow-repository` = false, plus a commit known to be in this history answering exit 0 on the same check. A negative reading with a lit control, which is exactly what the zero-hit rule asks for. **And declining to create the file on this branch was right**: it would have manufactured an add/add conflict with the sibling PR — the same conflict shape I spent this session resolving on #8719.

**Three failure modes, not one.** The card implies a single crash; you measured three:

| member | what actually happens |
|---|---|
| `'approve'` / `42` / `{}` | throws in `formatActionLabel` — the reported stack |
| `null` | throws **one frame earlier**, at `key={def.name}` (`BulkActionBar.tsx:221`) |
| `{ name: '' }` | ⚠️ **does not throw at all** — it rendered a nameless, unlabelled button |

So for the last one this is a behaviour change rather than a crash removal, and saying that plainly is better than letting "fixes the crash" quietly cover it.

**The guard site is the right one and the reasoning is checkable.** `rawAuthored` in `resolveBulkActions.ts` is where the authored array becomes the list the renderer maps over, so the `key` and the label read the same validated def and `BulkActionBar` keeps reading `def.name` unconditionally *because by that line it can*. One predicate, two readers — the filter and the message — so the check is not scattered. And `every` before `filter` preserves reference identity for a clean array, which `resolveBulkActions.test.ts` already pins; an always-allocating filter would have reddened it. That is a non-regression you inherited rather than wrote, and you noticed it.

⭐ **The discriminating pair is exactly what the ruling asked for, and the cross-check is better than the legs.** "Skip everything" reddens 6/7/8 while 1–5 stay green — which is *why* rows 6, 7 and 8 had to exist, since the unusable-class rows alone are passed by an implementation that skips everything. "Skip nothing" reddens 1,2,3,4,5,6,8,9. And **that mutation's red set is byte-identical to the base-tree red set** — independent evidence that the mutation reproduced the reported defect rather than a new one. I did not ask for that check; it is the strongest line in the report.

**The diagnostic is an assertion, both ways.** It fires naming the block, the index, the member and the other key where a bare name belongs; it does not fire for a clean list; and the no-fire leg reddens under "skip everything", so *"a warning that fires always"* is refused by measurement rather than by intent. Routing it through the channel `ObjectGrid` already owns for "you declared it, the renderer dropped it" — rather than inventing a third — is the right call.

**Direction one stayed out, with the measurement that settles it**: nothing on `main` asserts the current silence either way, but the in-flight #8071 pin's row 2 asserts `expect(warn).not.toHaveBeenCalled()` for it. Row 9 now pins that silence on `main` too, so a later diagnostic leaking into direction one is caught here rather than only on a branch.

⚠️ **The collision is real and I am handling it, not you.** PR #8737 carries the #8071 pin, and its row 5 asserts the crash this PR removes. Whichever lands second goes red there. I am flagging it on #8737 with the exact row and the rewrite it needs; the merge queue's speculative build protects `main` either way, so the cost is a queue cycle, not a breakage.

Noted and agreed on not filing: `BulkActionBar` being publicly exported means a host that hand-builds `actionDefs` bypasses this guard — but that is a TS-typed React prop, checked at the type level, and a different surface from authored metadata. Reporting it rather than filing it was the right line.

---

PR #8742, opened 2026-09-09
test(plugin-list): date the two query-absence controls to a settle window, not to mount (objectui#8705)

### LINKED ISSUE #8705: finding(tests): both find-absence controls in ListView.objectProviderBinding-7477 stay green against a query started 50ms after mount
Filed as an out-of-scope by-product of objectui#8703 (branch `claude/issue-8703-matcher-dependent-strict-bucket`). Not claimed, not assigned. objectui#8703 recorded this as an explicitly **unmeasured** observation and left it to its reader to measure or to leave declared. It was measured. It reproduces.

## The pins

`packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx`

- line 205 — `invents NO binding from an 'object' provider that names no object`: `expect(dataSource.find).not.toHaveBeenCalled();`
- line 218 — `leaves the 'value' provider alone — it never gains an objectName`: the same assertion.

Both are read straight after `await waitFor(() => expect(kanbanProps.length).toBeGreaterThan(0))`. For an inline-items provider `ListView` starts with `loading === false`, so those props arrive on the FIRST commit and the absence is dated to mount, not to a settled load.

## Forced

An implementation strictly worse than the bug: one line in `packages/plugin-list/src/ListView.tsx`, right after `const dataSource = props.dataSource;`, that starts a query 50ms after mount on every path.

```
React.useEffect(() => { setTimeout(() => {
  (dataSource as any)?.find?.('__ablation_probe__', {});
}, 50); }, []);
```

Mutation proven on disk both directions (`git hash-object` vs the HEAD blob, anchor counts, line-total gate) and restored by state (`git diff HEAD` empty).

**Leg 1 — with the probe: 7 of 7 tests pass.** Both `not.toHaveBeenCalled()` pins stay green while a query is being issued 50ms after every mount.

**Leg 2 — is the deferred query actually landing inside the control test?** A temporary diagnostic appended to the `value`-provider test, after its existing pin:

```
await new Promise((r) => setTimeout(r, 300));
expect(dataSource.find).toHaveBeenCalled();
```

With the probe: **7 of 7 pass** — so within that one test, `find` was un-called at pin time and called 300ms later, on the same mock.

**Leg 3 — the control that makes leg 2 mean something.** The identical test mutation (same blob, `9b245eda9bde4e8921531a66ea84ca61c6665ab1`) with `ListView.tsx` pristine (`ef7cf802d719944519bb11d1891904c14b5cb078`, equal to its HEAD blob): the diagnostic **fails** — `AssertionError: expected "vi.fn()" to be called at least once`, 6 passed / 1 failed. So `find` is not otherwise called, and the diagnostic can fail.

Per-test classification from vitest's JSON reporter in all three legs; zero death strings.

## What this does and does not say

- It does **not** say `ListView` has a bug today. Nothing in the current implementation defers a query on these paths.
- It says the two pins cannot distinguish "starts no query" from "starts a query a tick later", which is the regression shape that would otherwise reach them. An implementation strictly worse than the bug passes them.
- It is **outside** the objectui#8690 / objectui#8703 detector's population: a mock-call absence, not a recorder array. No census flag points here.

A repair is a settle window before the absence is read — the same move objectui#8664 and objectui#8690 made on the recorder sites — and it needs its own forced leg, because an absence pin that never fails has not been tested.

## Dedup — declared

`search_issues` returns false zeros in this repo, so a zero would not be evidence. The query run here was not a zero: `ListView value provider control asserts find not called, deferred query after mount would pass, pin is vacuous` scoped to this repo returned two hits (objectui#7947, objectui#4567) — a self-validating non-empty result, neither of them this.

Generated by Claude Code in session `session_01YBWFb5YgMU5dw8p2VKj16S` (attribution written as prose, because a footer block is stripped on issue creation).

os-justin:
Fixes #8705

`packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx` asserts at two sites that **no query is started** — the `object` provider naming no object must invent no binding, and the `value` provider must never gain an `objectName`. Both `expect(dataSource.find).not.toHaveBeenCalled()` reads sat straight after a `waitFor` satisfied on the **first commit**, so the absence was dated to **mount**. The assertions themselves are objectui#7477's contract and are unchanged; what changes is when they are read.

## The card's ablation, reproduced on today's `main`

Reproduced, not reinvented. Base `645087cd3`; pristine `ListView.tsx` blob `ef7cf802d719944519bb11d1891904c14b5cb078` — byte-identical to the blob the card recorded, so no drift on the mutated file. Every leg mutates from a committed tree, proves the mutation landed by `git hash-object` differing from the HEAD blob plus anchor counts and a line-total gate, and proves the restore **by state** (`git diff HEAD --quiet`), under `trap … EXIT INT TERM` with absolute paths.

| leg | mutation | ListView.tsx blob | result (JSON reporter) | card said |
| 1 | probe only | `f0e80896a9d1d6d16a0d63efba8bac0b3e66d7d4` | 7 passed / 0 failed | 7 of 7 pass |
| 2 | probe + 300ms diagnostic | `f0e80896…` | 7 passed / 0 failed | 7 of 7 pass |
| 3 | diagnostic only, pristine `ListView.tsx` | `ef7cf802…` (untouched) | 6 passed / **1 failed** — `AssertionError: expected "vi.fn()" to be called at least once` | 6 passed / 1 failed, same message |

All three hold. Zero suite-death strings in every run. One provenance note: the leg 2/3 test-file blob is `121aa643e0cf062b05c1f660a09847b1cc8bb8c2`, not the card's `9b245eda9bde4e8921531a66ea84ca61c6665ab1` — the same two-line diagnostic, placed against the existing pin line rather than re-emitted with the card's exact surrounding whitespace. Same mutation semantically, same outcome in both legs; the blob is not reproduced and is not claimed to be.

## Anchor or timer — the ruling applied

First, the mount-dating itself is **measured**, not inferred from the source. A temporary diagnostic reading `kanbanProps.length` synchronously, immediately after `render` returns and before the `waitFor`, reports **2 entries in both named tests** — so the `waitFor` is satisfied without ever yielding, and both absences were being read at mount.

An anchor was looked for first and does not exist on either path:

- **`value` provider** — an inline-items provider leaves `loading === false` from the first commit and it never flips. Nothing settles, so there is no positive signal to anchor on at all.
- **`object` provider naming no object** — `loading` starts `true` and flips `false` (read from the `useState` initializer in `ListView.tsx`; not separately forced), but it is set from `fetchData`'s own `!dataSource || !schema.objectName` branch — the very branch a query would start from, and the diagnostic above shows it has already happened before `render` returns. It postdates a *synchronous* query, never a deferred one, so it cannot date this absence.

So a **bounded wait**, and it is justified by the forced leg rather than by taste: `ABSENCE_SETTLE_MS = 250`, real time. It has to clear a deferral the environment already drains for free — RTL's `asyncWrapper` flushes one macrotask before `waitFor` returns, so a `setTimeout(…, 0)` regression sits *inside* the pre-existing window and a 0ms settle would prove nothing. 250ms is 5x the card's probe and an order of magnitude inside vitest's 5s default test timeout. The wait is wrapped in `act` so a regression that lands state during the window is flushed into React rather than warned about.

The window is placed before **both** absences in each test, so the `toBeUndefined()` companion is re-dated too: a late `objectName` would push a fresh props entry that `kanbanProps[kanbanProps.length - 1]` then reads.

## The repaired pins, forced

Repaired test blob `ca9e0e64dfafb964d34204437ffea968e0e5c117` — the blob at the branch's final commit `c78a74fd9`, committed before either leg ran, so both read the same bytes. (Both legs were first run at blob `92318e56ffd4b2c653ffb26008cc1c4f449772a0` and then re-run unchanged on the final blob after the header comment was corrected; the counts below are from the final run.)

- **RED under the card's own 50ms-deferred probe** — `ListView.tsx` at `f0e80896a9d1d6d16a0d63efba8bac0b3e66d7d4`: **5 passed / 2 failed**, both failures being exactly the two pins, both `AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times`. Zero death strings.
- **GREEN under pristine `ListView.tsx`** — blob `ef7cf802d719944519bb11d1891904c14b5cb078`: **7 passed / 0 failed**. A settle window that reddened on correct behaviour would be worse than the vacuous pin it replaced; it does not.

## Same-file sweep

The file has five absence reads besides the two named pins' `find` assertions. Classification is **measured**, not read off the source: a temporary diagnostic asserting `dataSource.find` has *already* been called at each pin line was applied to the three `toHaveLength(0)` sites and run.

| site | shape | verdict |
| --- | --- | --- |
| `objectName` `toBeUndefined()` in both named tests | same statement group, same mount dating | **repaired** — the settle window precedes them |
| `expect(gridProps).toHaveLength(0)` (kanban-from-react-page case) | view-spy absence | **not this shape** — diagnostic passed: `find` already called when the line runs |
| `expect(kanbanProps).toHaveLength(0)` (grid-by-default case) | view-spy absence | **not this shape** — diagnostic passed |
| `expect(kanbanProps).toHaveLength(0)` (explicit `viewType` wins case) | view-spy absence | **not this shape** — diagnostic passed |

Those three cases all bind an object, and `ListView` withholds the view spy behind its loading skeleton (`loading && data.length === 0`) until the query resolves — so they already sit behind a positive anchor and are left exactly as they were. What is **NOT MEASURED**: whether they would survive a differently-shaped regression that re-renders into the *other* view spy after a delay. That is a different forced leg and a different population; it is declared, not silently repaired.

No third `not.toHaveBeenCalled()` exists in the file. Scope stayed inside this one file.

## Verification

All of the below were run at the branch's final commit, `git rev-parse --short HEAD` = `c78a74fd9`. Run from the repo root with file paths; per-test classification from vitest's JSON reporter (`--reporter=json --outputFile`), never the text reporter, with suite-level messages counted explicitly as death strings.

- Whole package suite — `pnpm exec vitest run packages/plugin-list/`: **72 files, 888 passed / 0 failed / 0 pending, success=true, 0 death strings**.
- Type-check including tests — `pnpm --filter @object-ui/plugin-list run type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`): exit 0. Coverage of the edited file is proven, not assumed: `tsc -p tsconfig.test.json --listFiles` names it among 1503 program inputs, 0 errors. Dependency closure built first (`turbo run build --filter='@object-ui/plugin-list^...'`) and the `dist/*.d.ts` verified on disk rather than trusted from a replayed turbo log.
- `node scripts/check-control-bytes.mjs` — `✅ check-control-bytes: OK (scanned 6935 tracked text file(s); skipped 85 binary).`
- `pnpm --filter @object-ui/plugin-list run lint` — 0 errors. The six `no-explicit-any` warnings on this file are all pre-existing lines; the new code adds none.
- `node scripts/check-governed-queue-guard.mjs --test` on both changed paths — `✅ NOT GOVERNED — 2 path(s) checked against 5 governed surface(s); none matched.`

## Changeset

Empty frontmatter, which is this gate's explicit exemption for a test-only change. `node scripts/check-changeset-presence.mjs` verdict:

> ✅  1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/8705-listview-absence-settle-window.md.
> Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.

`check-changeset-no-major.mjs` and `check-changeset-fixed.mjs` both green.

Draft on purpose — left for the PM to flip and to arm auto-merge after contract review.

---

---

---

### DIFF
--- a/.changeset/8705-listview-absence-settle-window.md
+++ b/.changeset/8705-listview-absence-settle-window.md
@@ -1,0 +1,10 @@
+---
+---
+
+Date the two query-absence controls in
+`ListView.objectProviderBinding-7477.test.tsx` to a bounded settle window
+rather than to mount (objectui#8705). Both `not.toHaveBeenCalled()` pins sat
+straight after a `waitFor` that is satisfied on the first commit, so an
+implementation that starts a query 50ms after mount passed them; under the
+repair that same forced implementation turns them red. Test only; no published
+behaviour changes and no package is released by this change.
--- a/packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx
+++ b/packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx
@@ -48,11 +48,40 @@
  * handed `objectName: undefined` and nothing is fetched; without the
  * `specType` leg the page renders `object-grid` instead of `object-kanban`.
  * The `provider: 'value'` case is the control that stays green either way.
+ *
+ * ## Dating the two query-absence controls (objectui#8705)
+ *
+ * `invents NO binding …` and `leaves the value provider alone …` each assert
+ * that NO query is started. Read straight after their `waitFor`, that absence
+ * was dated to MOUNT rather than to a settled load: on both of those paths the
+ * view spy's props land on the FIRST commit, so `not.toHaveBeenCalled()` was
+ * being evaluated before a deferred query could have run. Both pins stayed
+ * GREEN against an implementation strictly worse than the bug — one line in
+ * `ListView.tsx` that fires `dataSource.find` 50ms after every mount — so they
+ * could not tell "starts no query" from "starts a query a tick later", which
+ * is exactly the regression shape that would reach them.
+ *
+ * There is no anchor to prefer here, and that was checked before reaching for
+ * a timer. An anchor needs a positive signal the implementation must emit
+ * AFTER any query would have started. On the `value` path nothing settles at
+ * all: `loading` is false from the first commit and never flips. On the
+ * `object`-provider-without-object path the one transition that exists —
+ * `loading` going false — is set from the very branch a query would start
+ * from, so it postdates a synchronous query and not a deferred one. The
+ * absence is therefore watched over a bounded window; see `ABSENCE_SETTLE_MS`.
+ *
+ * The other three absence reads in this file are NOT this shape, and that is
+ * measured, not assumed: at `expect(gridProps).toHaveLength(0)` and at both
+ * `expect(kanbanProps).toHaveLength(0)` sites, `dataSource.find` has ALREADY
+ * been called when the line runs. Those three cases bind an object, and
+ * `ListView` withholds the view spy behind its loading skeleton
+ * (`loading && data.length === 0`) until the query resolves — so they are
+ * already dated to a resolved query and are left exactly as they were.
  */
 
 import React from 'react';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { render, waitFor, cleanup } from '@testing-library/react';
+import { render, waitFor, cleanup, act } from '@testing-library/react';
 import { ComponentRegistry } from '@object-ui/core';
 import { SchemaRenderer, SchemaRendererProvider, AdapterCtx } from '@object-ui/react';
 // Registers the page renderers (`type:'page'` / `'home'`), which dispatch
@@ -140,6 +169,29 @@ function renderListView(schema: Record<string, any>) {
   return dataSource;
 }
 
+/**
+ * How long a query-absence is watched before it is believed.
+ *
+ * REAL time, on purpose, and deliberately larger than any deferral the test
+ * environment drains for free: RTL's `asyncWrapper` already flushes one
+ * macrotask before `waitFor` returns, so a `setTimeout(…, 0)` regression sits
+ * INSIDE the pre-existing window and a 0ms settle would prove nothing at all.
+ * objectui#8705's forced leg defers its query by 50ms; this window is 5x that
+ * and still an order of magnitude inside vitest's 5s default test timeout.
+ */
+const ABSENCE_SETTLE_MS = 250;
+
+/**
+ * Advance past `ABSENCE_SETTLE_MS` of real time so a deferred query has run
+ * before an absence is read. Wrapped in `act` so that a regression which does
+ * land state during the window is flushed into React rather than warned about.
+ */
+async function settleAbsenceWindow(): Promise<void> {
+  await act(async () => {
+    await new Promise((resolve) => setTimeout(resolve, ABSENCE_SETTLE_MS));
+  });
+}
+
 beforeEach(() => {
   kanbanProps = [];
   gridProps = [];
@@ -201,6 +253,11 @@ function Page() {
     const dataSource = renderListView({ data: { provider: 'object' }, specType: 'kanban' });
 
     await waitFor(() => expect(kanbanProps.length).toBeGreaterThan(0));
+    // Both reads below are ABSENCES, and the props above arrive on the first
+    // commit — so date them to the end of a settle window, not to mount
+    // (objectui#8705). A late `objectName` would push a fresh props entry and
+    // a deferred query would have called `find` by the time this returns.
+    await settleAbsenceWindow();
     expect(kanbanProps[kanbanProps.length - 1].schema.objectName).toBeUndefined();
     expect(dataSource.find).not.toHaveBeenCalled();
   });
@@ -214,6 +271,10 @@ function Page() {
     });
 
     await waitFor(() => expect(kanbanProps.length).toBeGreaterThan(0));
+    // Same dating as the control above, and this is the path with NO settle of
+    // its own: an inline-items provider leaves `loading` false from the first
+    // commit, so without this window the absence is dated to mount.
+    await settleAbsenceWindow();
     expect(kanbanProps[kanbanProps.length - 1].schema.objectName).toBeUndefined();
     expect(dataSource.find).not.toHaveBeenCalled();
   });
--- a/packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx
+++ b/packages/plugin-list/src/__tests__/ListView.objectProviderBinding-7477.test.tsx
@@ -53,9 +53,11 @@
  *
  * `invents NO binding …` and `leaves the value provider alone …` each assert
  * that NO query is started. Read straight after their `waitFor`, that absence
- * was dated to MOUNT rather than to a settled load: on both of those paths the
- * view spy's props land on the FIRST commit, so `not.toHaveBeenCalled()` was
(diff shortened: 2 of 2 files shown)


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** No failures; shards and Lint still running.

**The ruling said "prefer an anchor over a timer", and you established there is no anchor to prefer — by measurement, not by reading the source.** That is the part that makes the 250ms defensible rather than a shrug:

- A diagnostic reading `kanbanProps.length` synchronously right after `render` returns already reports 2 entries in **both** named tests, so the `waitFor` never yields and both absences were being read at mount.
- On the `value` path nothing settles at all — `loading` is `false` from the first commit and never flips.
- On the object-provider-naming-no-object path the one transition that exists is set from `fetchData`'s own `!dataSource || !schema.objectName` branch — **the very branch a query would start from** — and has already happened before render returns. It can date a *synchronous* query and never a deferred one, which is exactly the shape at issue.

An anchor that fires before the thing it is supposed to outlast is worse than no anchor, because it looks like one. Ruling out the candidate anchors on that ground is the right way to arrive at a timer.

**The forced leg is the whole card and it lands**: 5 passed / 2 failed under the card's own 50ms probe, and the two failures are exactly the two pins, both *"expected vi.fn() to not be called at all, but actually been called 1 times"*. Green leg on pristine `ListView.tsx`: the whole package at 888/888. So the repaired pins fail on the regression shape and do not fail on correct behaviour — which is what "an absence pin that never fails has not been tested" was asking for.

⭐ **The sweep result is better than a repair would have been.** Five other absence reads in the file, and you distinguished rather than batch-repairing: the two `toBeUndefined` companions sit in the same statement groups and were re-dated by the same window, but the three `toHaveLength(0)` view-spy absences are **not** the same shape — and that is measured, with a temporary diagnostic asserting `find` has *already* been called at each of those three lines, which **passes at all three**, because those cases bind an object and `ListView` withholds the view spy behind its loading skeleton until the query resolves. They already sit behind a positive anchor. Leaving them alone is the right call, and objectui#8665 is the cautionary tale for the other one.

**Provenance reported honestly, including where it did not reproduce.** The pristine `ListView.tsx` blob is byte-identical to the card's, so no drift on the mutated file; the leg 2/3 *test*-file blob differs (`121aa643e…` vs the card's `9b245eda9…`) — the same two-line diagnostic placed differently — and you say plainly that the blob is not reproduced and do not claim it. Reporting a partial reproduction as partial is worth more than a matching hash.

⚠️ **One property to name, since it is the pin's declared sensitivity rather than a defect**: a 250ms real-time window bounds what this pin can catch. A regression that defers its query by more than 250ms slips through. 250 is 5× the probe and an order of magnitude inside the 5s default timeout, so the bound is a reasonable one — but it is a bound, and a future reader should know the pin says *"no query within 250ms of mount"* rather than *"no query"*. Worth one clause in the constant's comment whenever that file is next touched; not worth a CI cycle now.

The two `NOT MEASURED` entries — the differently-shaped regression against the three view-spy absences, and the repo-wide gate farm — are declared rather than silently skipped, which is the standard.

---

PR #8745, opened 2026-09-09
fix(plugin-tree): derive ObjectTree expansion during render, not in a state mirror

### LINKED ISSUE #8666: finding(plugin-tree): ObjectTree seeds `expanded` in a useEffect state mirror, so every render shows a one-frame collapsed forest before it expands
Filed by the `domain:ui` execution-seat PM. Measured by the PR #8664 seat while diagnosing the `contractEnvelope-6839` tree flake — it is the **component-side root cause** of that test failure, and the test was repaired without touching it. ⛔ Not graded and not assigned.

## The mechanism

`ObjectTree` keeps expansion in a **state mirror rather than derived state**:

- `const [expanded, setExpanded] = useState(new Set())`
- re-seeded by a `useEffect` keyed on `[roots, config.defaultExpandedDepth]`
- rows computed as `flattenVisible(roots, expanded)`

⇒ When rows land, **the commit that first paints the table still carries the previous, empty mirror.** The root draws; its children do not. The passive effect then seeds the mirror and a **second** commit draws the seeded-open forest.

Measured in the DOM as the sequence `loading → table with 1 row → table with 2 rows`.

## Why it is worth a card even though the test is fixed

⭐ **Every real user sees the collapsed forest for one frame** before it expands, on every mount where `defaultExpandedDepth` is non-zero. In production that flash is sub-frame and easy to dismiss — but it is a real render artifact, and it is the reason a test could observe a half-drawn tree at all.

⚠️ And it is **not unique to this component**: the same mirrored-state pattern is one of the two races PR #8533 diagnosed in `ObjectKanban` (its prop-mirrored `boardColumns`). ⇒ This is a second instance of a shape that has now cost two CI repairs.

## The candidate direction — ⛔ not a ruling, and not measured

Computing the seeded set **during render**, keyed on the forest, rather than mirroring it into state via an effect, would remove the intermediate commit entirely.

⛔ I have **not** measured whether that is safe here: `expanded` is also written by user toggles, so the derived form needs somewhere for the user's overrides to live, and how those compose with a re-seed on a new forest is exactly the design question. ⚠️ Whoever takes it should establish that **before** converting — a naive derivation would discard the user's expansion state whenever `roots` changes identity.

## What is NOT claimed

- ⛔ No priority claimed. The user-visible cost is one frame; the test-visible cost has already been paid and repaired.
- ⛔ **Not a regression** and not introduced by PR #8664, which changed **only** the test file.
- ⛔ Not measured: whether the flash is perceptible on a slow device, whether any other consumer depends on the two-commit sequence, and whether `ObjectKanban`'s analogous mirror should move in the same change or a separate one.

## Provenance

Measured 2026-09-08 against `ObjectTree.tsx` on `origin/main` while repairing `ObjectTree.contractEnvelope-6839.test.tsx` (PR #8664). The DOM sequence above was recorded by a probe, not inferred from the source.

os-justin:
Fixes #8666

⚠️ **Type-shaped tokens are spelled out in CAPITAL WORDS below, never in angle-bracket form.** GitHub's body sanitizer deletes tag-shaped fragments on save — backticks and fenced blocks do not protect them — and a before/after table about types is exactly the thing it collapses into "nothing changed". So `useState` of a SET OF STRING, and a READONLY MAP FROM STRING TO BOOLEAN.

## Two deliverables, and the second is the discriminating one

**1 — no intermediate commit.** Expansion lived in a `useState` holding a SET OF STRING that a passive `useEffect` keyed on `[roots, defaultExpandedDepth]` re-seeded from the forest, with rows computed as `flattenVisible(roots, expanded)`. The commit that *first* painted the table therefore still carried the previous, empty mirror: the root drew, its children did not, and a second commit drew the seeded-open forest.

**2 — a user's expansion survives a re-seed.** State now holds only the answers the *user* gave by clicking a chevron — a sparse READONLY MAP FROM STRING TO BOOLEAN, id to open/closed — and the seed is computed from the forest during render. The composition rule, stated in the docblock on `resolveExpanded` and in the package README:

> A new forest may re-seed, but a node the user deliberately opened or closed — and which is still in the forest — keeps the user's answer. Every other node, including a genuinely NEW one, takes the seed.

### Why the user's EDITS are stored and not the expansion set

This is the part that decided the design, not an implementation detail. A set of open ids cannot tell *"the user closed this node"* apart from *"the seed never opened it"*. So re-seeding such a set has exactly two outcomes and both are wrong: overwrite, and the user's collapse is lost on every identity change of `roots`; or union/skip, and a genuinely new subtree never opens. Recording the user's edits separately is what makes both halves derivable at once.

⇒ That also makes this a **behaviour fix in the same direction**, not a side effect. Before this change a re-seed *overwrote* the user's expansion, so any refetch, filter, or host re-render that reallocated the rows silently reopened every subtree the user had collapsed. That is measured below, not asserted: the override pins are red against the pre-fix component.

### The DOM commit sequence, before and after

Recorded by a `React.Profiler` whose `onRender` reads the container on every commit of the tree's subtree; only *consecutive identical* snapshots collapse, so a one-row table and a two-row table can never merge.

| | sequence |
|---|---|
| `origin/main` `645087cd3` | `loading -> table:1rows -> table:2rows` |
| this branch | `loading -> table:2rows` |

Both arms of the frames pin record the same sequence. The second arm defers `find()` by **50ms** rather than 0: `@testing-library`'s `asyncWrapper` drains one macrotask before returning, so a 0ms deferral resolves inside that drain window and would leave the ordering unforced.

### Every pin observed RED, with base-blob provenance

The pre-fix leg is the component restored from the base blob, proven by hash rather than by "I put the old code back". Each leg proved the mutation reached disk before the run, and proved the restore by `git diff HEAD` being empty afterwards:

```
HEAD blob        7867221cd84ea746986aab346f01efa632b78365
on-disk before   7867221cd84ea746986aab346f01efa632b78365
on-disk after    5a1fc6580bcbf2cad28fb497aa3765bbaea179fd   == 645087cd3:packages/plugin-tree/src/ObjectTree.tsx
on-disk restored 7867221cd84ea746986aab346f01efa632b78365   (git diff HEAD empty)
```

Per-case outcomes, from vitest's **JSON reporter** — never the text reporter, because a module-scope `TypeError` reads there as a suite death rather than a failing assertion:

| pin | pre-fix (base blob) | caricature A: "seed during render, drop the overrides" | caricature B: "never re-seed" (seed frozen on the first forest) | this branch |
|---|---|---|---|---|
| paints the seeded-open forest in the FIRST commit that has rows | **RED** | green | green | green |
| still paints it in one commit when the rows arrive in their own task | **RED** | green | green | green |
| keeps a node the user COLLAPSED when the forest identity changes | **RED** | **RED** | green | green |
| keeps a node the user EXPANDED below `defaultExpandedDepth` across a re-seed | **RED** | **RED** | green | green |
| seeds a genuinely NEW subtree open while the override still holds | **RED** | **RED** | **RED** | green |
| whole `packages/plugin-tree/` | 54 passed / 5 failed | 55 passed / 4 failed | 58 passed / 1 failed | **59 passed / 0 failed** (13 files) |

⭐ **Which pin catches which.** Caricature A is the mutation the card warned about — it removes the intermediate commit and answers the same thing for every input, an implementation strictly worse than the bug. **Both frames pins pass under it**, which is the measured reason the second group has to exist; the three override cases catch it. Caricature B is caught by **exactly one** case, the new-subtree one: 58 of 59 tests in the package pass under it, both survives cases included, because a frozen seed leaves the user's answer alone. Caricature A also reds the pre-existing `ObjectTree.test.tsx` toggle test, which is corroboration, not the pin.

### The PR #8664 repair: nothing was standing on the two-commit sequence

⭐ Measured directly rather than reasoned about: `ObjectTree.contractEnvelope-6839.test.tsx` is **green in all four columns above** — against the pre-fix component, against both caricatures, and against this branch. Its positive arms wait FOR the descendant row, which is a condition on the settled forest and not on how many commits produced it. **No assertion in it changed.**

What did change there is prose: its header carries a MEASURED description of the race as a live property of `ObjectTree`, and that statement is now false. Rather than quietly editing the analysis away, the header gains a dated note saying the mechanism was fixed here, that nothing in the file changed for it, and why the waits stay shaped as they are. The `⛔ do not fold these back into one wait` instruction is untouched and still correct.

## Verification, and what is NOT MEASURED

- `pnpm exec vitest run packages/plugin-tree/` from the repo ROOT with a path filter — **13 files, 59 tests, 0 failed**, at `ecbd71bba`. Never `pnpm --filter PKG exec vitest FILE`, never a package cwd, never `--no-inline-config`.
- `pnpm --filter @object-ui/plugin-tree run type-check` — exit 0, after building the dependency closure (`turbo run build --filter="@object-ui/plugin-tree^..."`). ⭐ Confirmed by `tsc --listFiles` that the **new pin file is actually in the program** (14 of the package's source files, the new one among them) rather than excluded — a `type-check` that skips tests is NOT MEASURED, not green.
- `check:doc-snippets`, `check:readme-exports`, `check:doc-types`, `check:doc-fences`, `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:unreferenced-sources`, `check-changeset-presence` — all exit 0. The first two were run **only after** the scoped build they demand (96s); unbuilt they exit 2 / report `the population COLLAPSED -- this run proves nothing`, which is a precondition failure and not a verdict.
- `pnpm --filter @object-ui/plugin-tree run lint` — 0 errors. The warnings are all `no-explicit-any` on test fixtures, the house convention in this package, plus one pre-existing `exhaustive-deps` warning on the record-fetch effect that is unchanged by this diff.
- **NOT MEASURED:** anything about pixels. happy-dom reports `clientWidth: 0`, so the pins count nodes and read `data-depth`, never the inline `paddingLeft` the same element carries. Also not measured: whether the flash is perceptible on a slow device.
- **NOT MEASURED:** the full repo suite and the whole gate farm — CI owns those.

## Scope

⛔ **`ObjectKanban` is not folded in.** Its prop-mirrored `boardColumns` (PR #8533) is the same shape and is a separate card; it was not touched and not measured here.

⚠️ **One out-of-scope defect was measured here and filed separately as objectui#8743, which remains open.** `check:doc-examples` is red on `origin/main` at `645087cd3` — an undeclared `@example` failure at `packages/types/src/zod/imported-defaults.ts:318`, because `stripImportedDefaults` is not on a public entry so the gate injects no import for it. Zero files under `packages/types` are in this diff and the block is byte-identical at the base commit, so it is not caused by this change and is not repaired by it.

## Changeset

`.changeset/8666-tree-expanded-derived.md`, `patch` on `@object-ui/plugin-tree`. Required: this is a user-visible rendering change. The `skip-changeset` label is a phantom in this repo and the empty-frontmatter exemption does not apply. `major` is banned repo-wide and is not wanted here — no schema key changes and `defaultExpandedDepth` means exactly what it meant.

Session that produced this change: `https://claude.ai/code/session_01YBWFb5YgMU5dw8p2VKj16S` (written as prose and as a code span, because footer links do not reliably survive a body edit).

---

### DIFF
--- a/.changeset/8666-tree-expanded-derived.md
+++ b/.changeset/8666-tree-expanded-derived.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/plugin-tree': patch
+---
+
+`ObjectTree` derives expansion during render instead of mirroring it into state
+(objectui#8666).
+
+**The rendering artifact this removes.** Expansion lived in a
+`useState<Set<string>>(new Set())` that a passive `useEffect` keyed on
+`[roots, defaultExpandedDepth]` re-seeded from the forest, with rows computed as
+`flattenVisible(roots, expanded)`. So the commit that first painted the table
+still carried the previous, empty mirror: the root drew, its children did not,
+and a second commit drew the seeded-open forest. Probed in the DOM the sequence
+was `loading` then a one-row table then a two-row table; it is now `loading` then
+the two-row table. Every mount with a non-zero `defaultExpandedDepth` showed a
+collapsed forest for one frame, and that frame is also why a test could observe a
+half-drawn tree at all.
+
+**The behaviour change that comes with it, and it is the load-bearing half.**
+Component state now holds only the answers the *user* gave by clicking a chevron
+— a sparse map of node id to open/closed — and the seed is computed from the
+forest during render. They compose by one rule: a new forest may re-seed, but a
+node the user deliberately opened or closed, and which is still in the forest,
+keeps the user's answer; every other node, a genuinely new one included, takes
+the seed.
+
+That is a fix in the same direction as the frame, not a side effect of it.
+Before this change a re-seed *overwrote* the user's expansion, so any change to
+the identity of the record set — a refetch, a filter, a host re-render that
+reallocated the rows — silently reopened every subtree the user had collapsed
+and reclosed every one they had opened below `defaultExpandedDepth`. Authored
+metadata is unaffected: `defaultExpandedDepth` means exactly what it meant, and
+no schema key changes.
--- a/packages/plugin-tree/README.md
+++ b/packages/plugin-tree/README.md
@@ -43,6 +43,20 @@ const schema: ObjectQLComponentSchema = {
 Records whose parent is missing (or points outside the result set) are kept as
 roots, so nothing is silently dropped.
 
+### Expansion: the seed and the user's answer
+
+`defaultExpandedDepth` **seeds** expansion; it does not own it. The seeded set is
+derived from the forest during render rather than mirrored into component state,
+so a tree that expands by default is painted expanded in the first commit that
+has rows — there is no frame in which the forest is drawn collapsed
+(objectui#8666).
+
+When the record set changes — a refetch, a filter, a host that reallocates the
+rows — the seed is recomputed for the new forest. A node the user opened or
+closed by clicking its chevron, **and which is still in the forest**, keeps the
+user's answer; every other node, a genuinely new one included, takes the seed.
+Expansion is per-mount session state: it is not addressable and is not persisted.
+
 ### The `tree` view type is host composition, not authoring
 
 `tree` is **not** an authorable view type. Neither `ObjectViewSchema.defaultViewType`
--- a/packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx
+++ b/packages/plugin-tree/src/ObjectTree.contractEnvelope-6839.test.tsx
@@ -82,6 +82,25 @@
  * ⛔ Do not fold these back into one wait, and ⛔ do not "fix" a future red here
  * with a longer timeout: the failure was never slowness, it was reading a
  * signal that does not carry the answer.
+ *
+ * ## ⚠️ The race described above was FIXED in the component (objectui#8666)
+ *
+ * Everything above stands as the record of why these waits are shaped the way
+ * they are, but one of its statements is no longer true of `ObjectTree`:
+ * expansion is no longer a `useState` mirror re-seeded from a `useEffect`, so
+ * the `loading → table:1rows → table:2rows` sequence it measures is now
+ * `loading → table:2rows` and the intermediate one-row commit does not happen.
+ * The new shape and both halves of its contract are pinned in
+ * `ObjectTree.expandedDerived-8666.test.tsx`.
+ *
+ * ⭐ NOTHING IN THIS FILE CHANGED FOR THAT, and the note exists to say why the
+ * absence of a change is deliberate. No assertion here was standing on the
+ * two-commit sequence: the positive arms wait FOR the descendant row, which is
+ * a condition on the settled forest and not on how many commits produced it, so
+ * they were green before objectui#8666 and are green after it. The reason to
+ * keep them as they are is the one the section above gives — the table's
+ * `data-testid` is a MOUNT signal and the rows are what this file counts, which
+ * stays true no matter how many commits the component takes to get there.
  */
 
 import React from 'react';
--- a/packages/plugin-tree/src/ObjectTree.expandedDerived-8666.test.tsx
+++ b/packages/plugin-tree/src/ObjectTree.expandedDerived-8666.test.tsx
@@ -1,0 +1,383 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `ObjectTree` DERIVES expansion during render, and a user's answer survives a
+ * re-seed (objectui#8666).
+ *
+ * Two things are pinned here, and they are pinned separately because ONE of
+ * them is satisfiable by an implementation strictly worse than the bug.
+ *
+ * ## 1 — the frames (`describe` "the commit sequence")
+ *
+ * Expansion used to be a `useState<Set<string>>(new Set())` MIRROR that a
+ * passive `useEffect` keyed on `[roots, defaultExpandedDepth]` re-seeded from
+ * the forest, with rows computed as `flattenVisible(roots, expanded)`. So the
+ * commit that FIRST painted the table still carried the previous, empty mirror:
+ * the root drew, its children did not, and a SECOND commit drew the seeded-open
+ * forest. Probed in the DOM on this file's own fixture, the sequence was
+ *
+ *     loading → table:1rows → table:2rows
+ *
+ * and it is now
+ *
+ *     loading → table:2rows
+ *
+ * ⭐ MEASURED, not inferred: the recorder is a `React.Profiler` whose
+ * `onRender` reads the container on EVERY commit of the tree's subtree, and
+ * only CONSECUTIVE IDENTICAL snapshots collapse. A one-row table and a two-row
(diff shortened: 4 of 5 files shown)


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** No failures; shards, Type Check and Lint still running.

⭐ **The most valuable thing in this PR is a correction to my own ruling.** I framed override-loss as a *hazard of the conversion* — something a naive derivation would introduce. Measurement says `main` **already discards it**: the passive effect overwrites the whole set on every `roots` identity change, so both "survives" pins are **red against the unmodified pre-fix component**, not merely against caricature A. Deliverable 2 therefore repairs a second, previously unnamed defect rather than guarding a refactor. The card did not know that and neither did I; it is only visible because the pins were written to be observed failing against the real base rather than against a mutant.

**The composition rule is right, and the reason it is right is the part to keep.** Storing the user's *edits* as a sparse `id → boolean` map rather than an expansion set, because:

> A set of open ids cannot tell "the user closed this node" apart from "the seed never opened it."

Re-seeding such a set has exactly two outcomes and both are wrong — overwrite and the user's collapse dies on every identity change, or union/skip and a genuinely new subtree never opens. That is a real derivation, not a preference, and it is what forced `toggle(id)` to become `toggle(id, isOpen)`: the updater sees only the override map, which does not know the seed. A design whose signature change is *explained* by the invariant is one that will survive the next reader.

**The two caricatures did the job the ruling asked of them, and the asymmetry is the lesson:**

| leg | package result |
|---|---|
| pre-fix base | 54 / 5 failed |
| A — drop overrides | 55 / 4 failed |
| B — never re-seed | **58 / 1 failed** |
| this branch | 59 / 0 |

Caricature B is caught by **exactly one** case, the new-subtree one — 58 of 59 package tests pass under an implementation that never re-seeds. That is *"the vivid assertion is rarely the discriminating one"* with a number attached, and it is why the second group had to exist: both frame pins pass under caricature A, measured rather than assumed.

**The DOM recorder is honest about its own resolution.** A `Profiler` `onRender` snapshot that collapses only *consecutive identical* frames means a 1-row and a 2-row table can never merge — so the later report of an absence is a reading rather than a gap. And it showed the defect before the fix, which is what licenses trusting it after. The 50ms deferred arm rather than 0ms is the `asyncWrapper` drain-window rule applied without being reminded.

**On the `#8664` file — keep the note.** You verified by measurement, not by reading, that no assertion there changed: green in all four legs, and its positive arms wait *for* the descendant row, a condition on the settled forest. But its header states the mirrored-state race as a live **measured property** of `ObjectTree`, and that sentence is now false. A dated note saying the mechanism was fixed, that nothing in the file changed for it, and why the waits stay as they are, is exactly right. ⛔ Do not revert it: a header carrying a measurement that has stopped being true is the defect class this repo has filed three times today alone (objectui#8734, objectui#8693, objectui#8716). Correcting it in the PR that falsifies it is how that class stops accumulating.

**The out-of-scope filing is the other thing that earned its place.** objectui#8743 — `check:doc-examples` red on `origin/main` — is genuinely not yours (zero files under `packages/types` in your diff, the block byte-identical at BASE), and you were right to file rather than fold it in. I verified it independently and it has been dispatched at priority; it arrived through PR #8721, which I reviewed and armed, so the redness is mine to clear.

Also noted: catching the `"not fixed here: objectui#8743"` collocation in your own PR body on read-back, before the closing-keyword parser could act on a negation it does not understand. That is a hazard most readers would not have looked for.

---

PR #8749, opened 2026-09-09
test(app-shell): measure the reachability of a `dependsOn` lookup action param (objectui#8672)

os-justin:
Part of #8672 — **measurement only**. ⛔ No disposition is chosen here, and the card stays open: arms 1 (wire it) and 2 (refuse it) are both still on the table and choosing between them is the maintainer's, not this PR's.

## What this measures

The card asks three questions. All three now have readings, each with a lit control.

### Q1 — how many writers are there?

**Zero lookup action params declare `dependsOn` in this repository**, and zero action params declare it at *any* type. Sweeps run as `git grep -l … -- .` over 7,031 tracked files, with controls on the same command shape and population:

| reading | files |
|---|---|
| CONTROL `dependsOn` (any use) | 164 (719 entries) |
| CONTROL `'lookup'` | 284 |
| CONTROL `lookup` | 891 |
| CONTROL `ActionParamDef` | 34 |
| **SUBJECT — a lookup *action param* with `dependsOn`** | **0** |

Every co-occurrence of a lookup and `dependsOn` in the tree is **field** metadata inside test files, never an action param. Three framework code sites carry the key (`paramToField`'s emit, the resolver's `RESOLVED_ONLY_PARAM_KEYS` message, the resolver's snake read) — none is an authored document.

⚠️ **No fraction can be computed from this tree.** `git grep -lE '"params"' -- '*.json'` and the same for `"actions"` both return **0**: this repository authors *no* action metadata as data at all, so the denominator is zero rather than large.

⚠️ **What the sweep cannot see.** Metadata authored server-side — the `/api/v1/meta/object` documents the resolver actually reads, and any host app's action definitions — never appears in this tree. The honest reading is **"zero writers *in this repository*"**, not "zero writers".

### Q2 — where is the authoring surface? ⭐

**`@objectstack/spec` (17.3.0, from `pnpm-lock.yaml`), and it already refuses the key.** Measured by parsing, not by reading declarations:

| document | `ActionParamSchema.safeParse` |
|---|---|
| POSITIVE CONTROL — lookup param + `reference` | ✅ accepted |
| NEGATIVE CONTROL — unknown key | ❌ `unrecognized_keys` |
| **SUBJECT — lookup param + `dependsOn`** | ❌ **`unrecognized_keys: ['dependsOn']`** |
| `select` param + `dependsOn` | ❌ refused the same way |

`@object-ui/types`' `ActionParam` extends `Omit<z.input<typeof ActionParamSchema>, 'type'>`, so it declares no `dependsOn` either and `tsc` refuses it; the resolver names it a third time in `RESOLVED_ONLY_PARAM_KEYS`.

⇒ **Arm 2's refusal already exists, upstream, mirrored objectui-side by three independent mechanisms.** There is nothing to build here and nothing to request upstream.

### Q3 — is the gate reachable outside a probe?

**Not from any spec-valid authored metadata.** Inline authoring is refused (Q2). The one route the repo itself points authors to — `RESOLVED_ONLY_PARAM_KEYS.dependsOn`: *"make the param field-backed … to pick it up"* — reads `field.depends_on`, the **snake** spelling `FieldSchema` refuses by name (`Did you mean depends_on → dependsOn?`), while the camel `dependsOn` it accepts is never read:

| field document | `FieldSchema` | reaches the resolved param? |
|---|---|---|
| POSITIVE CONTROL — plain lookup field | ✅ | — |
| camel `dependsOn` (spec's spelling) | ✅ accepted | ❌ resolves to `undefined` |
| snake `depends_on` | ❌ refused by name | ✅ arrives |

⇒ The two spellings are disjoint: the one the spec admits is not read, and the one that is read the spec refuses.

## Pins shipped

New `packages/app-shell/src/views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx`, 10 assertions in three legs. Legs A and C are labelled **CURRENT SHAPE, NOT CONTRACT** in the file docblock and on their describe blocks — whoever implements a disposition should expect them red and should **rewrite** them, not trust them. Leg B is labelled a contract pin, version-qualified.

Also corrects a stale comment in `utils/paramToField.test.ts` that described this seam backwards (the folded-in finding on the card).

## Verification — every assertion observed red

Seven ablations, each from the committed tree, each with the mutation proved on disk (`git hash-object` vs the HEAD blob, anchor counts, a line-total gate) and the restore proved **by state** (`git diff HEAD --quiet`), under `trap … EXIT INT TERM` with absolute paths. Per-test outcomes from the **JSON reporter**.

| ablation | mutation | observed red |
|---|---|---|
| ABL-1 | add `'lookup'` to `CASCADE_OPTION_WIDGET_TYPES` | leg A #2, #3 |
| ABL-2 | `LookupField` gate forced false | leg A #1, #2 |
| ABL-3 | resolver reads `field.dependsOn` instead of `field.depends_on` | leg C #2, #3 |
| ABL-4 | leg B subject document loses `dependsOn` | leg B #3 |
| ABL-5 | leg B negative control loses its unknown key | leg B #2 |
| ABL-6 | leg B `select` document loses `dependsOn` | leg B #4 |
| ABL-7 | leg C control asserts the camel spelling is refused | leg C #1 |

All 10 assertions were observed red at least once. ⛔ None of these mutations is committed; every restore is proved by state above.

Green runs: 10/10 on the pin file; **141/141** across the blast radius (`paramToField`, all three `resolveActionParams` suites, all three `ActionParamDialog` suites). `pnpm --filter @object-ui/app-shell type-check` exits **0** after building the dependency closure, and `tsc -p tsconfig.test.json --listFiles` confirms both changed files are inside that program. Targeted `eslint` on both files: 0 errors, 0 warnings.

Gate verdict lines, quoted:

- `✅  1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` … `Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.`
- `✅  No changeset declares a 'major' bump.`
- `✅  check-control-bytes: OK (scanned 6946 tracked text file(s); skipped 85 binary).`
- `✅ NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched.`

## ⚠️ Two premises measurement contradicted

**1. The PM triage's argument for arm 1 being a feature.** The triage says a lookup *"has no options list; it has a query. Filtering candidate records by a dependency's value is a different mechanism that does not exist here … the gate would lift while the picker still showed every record."* That mechanism **does** exist and is shipped. `LookupField` builds `dependentFilter` from the same `dependsOn` chain and merges it into `popoverFilter`, which feeds `useRecordQuery` for all three candidate surfaces (popover, Level-2 picker, PeoplePicker). The `#7165` changeset states it directly: *"every picker takes the `dependsOn` chain as a hard `baseFilter`. The second half is host-independent and was already live."* ABL-1 measured the consequence — with `lookup` in the set, the gate **did** lift on the keystroke. ⚠️ This is not an argument *for* arm 1: which route should supply the record is still open (`CASCADE_OPTION_WIDGET_TYPES` is documented as an options-list allow-list, and the lookup-family boundary is objectui#4771). Only the stated reason is contradicted.

**2. The card's "different consumers" claim about the grid twin.** The card says the two surfaces *"have different consumers and different remedies."* They share **one** final consumer — `LookupField`, one `dependenciesMissing` gate, one `dependentFilter`. What differs is the host that supplies the record. And **objectui#7154 is CLOSED**, its `dependsOn` case rewritten by **objectui#7165**, which fixed the grid with `dependentValues={ctx.pendingRow ?? ctx.row}` — so the grid twin is not an open twin but a **landed precedent for the remedy**. The genuinely open siblings are **#7190** (detail page, pinned not fixed) and **#7206** (the unsettable context tail), both `pm:blocked`.

---

---

### DIFF
--- a/.changeset/8672-lookup-dependson-reachability-measurement.md
+++ b/.changeset/8672-lookup-dependson-reachability-measurement.md
@@ -1,0 +1,35 @@
+---
+---
+
+Measurement-only change in `@object-ui/app-shell`: objectui#8672 asked how
+reachable a `dependsOn` lookup ACTION PARAM is, and where the surface that
+admits it lives. No behaviour changes and no disposition is chosen — the card
+names three (wire it · refuse it · declare the limit) and this pins what is true
+today so the ruling is made against something stable.
+
+New `views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx`, three legs,
+each with a lit control:
+
+- **Leg A — current shape, not contract.** Driving the real `ActionParamDialog`,
+  a lookup param declaring `dependsOn` renders `lookup-trigger-gated`, disabled,
+  "Select account first", and filling the named parent does NOT lift it. A
+  control lookup identical but for `dependsOn` is asserted enabled, and the
+  keystroke is driven through a `radio` param whose `visibleWhen` DOES react to
+  it — so "the gate did not lift" is read against a keystroke proven to have
+  reached and re-rendered the dialog.
+- **Leg B — contract, version-qualified.** `@objectstack/spec`'s
+  `ActionParamSchema` (17.3.0) is `.strict()` and refuses `dependsOn` on an
+  action param of ANY type, with a positive control that parses and an
+  unknown-key negative control that is refused. `@object-ui/types`' `ActionParam`
+  derives its keys from `z.input` of that schema, so it declares none either.
+- **Leg C — current shape, not contract.** The one route the repo points authors
+  to (`RESOLVED_ONLY_PARAM_KEYS.dependsOn`: make the param field-backed) reads
+  `field.depends_on`, the SNAKE spelling `FieldSchema` refuses by name, while the
+  camel `dependsOn` it accepts is never read — so a spec-valid object field
+  declaring the cascade resolves to `dependsOn: undefined`. A sibling key off the
+  same field def is asserted to arrive in the same call, so the `undefined` is a
+  reading and not an empty fixture.
+
+Also corrects a stale sentence in `utils/paramToField.test.ts` that described
+this seam backwards ("…or the dialog's dependent lookups would have gone dead
+silently" — there has never been a working cascade here to go dead).
--- a/packages/app-shell/src/utils/paramToField.test.ts
+++ b/packages/app-shell/src/utils/paramToField.test.ts
@@ -108,7 +108,20 @@ describe('paramToField', () => {
   // and objectui#7357 added `dependsOn` to that half when it retired the
   // snake_case `depends_on` this adapter used to emit — this face was the one
   // in-repo framework producer of that spelling, so the emit had to move with
-  // the reader or the dialog's dependent lookups would have gone dead silently.
+  // the reader.
+  //
+  // ⚠️ CORRECTED (objectui#8672). This sentence used to end "…or the dialog's
+  // dependent lookups would have gone dead silently", which describes a WORKING
+  // cascade in this dialog. Measured, there has never been one: `dependsOn` is
+  // emitted onto the field bag below, but `ActionParamDialog` supplies
+  // `dependentValues` only to `CASCADE_OPTION_WIDGET_TYPES` (`select` /
+  // `multiselect` / `radio` / `checkboxes` — `lookup` is not a member), so
+  // `LookupField` resolves `{}` and the gate NEVER lifts. What moving the emit
+  // preserved is therefore the GATE, not a cascade: leaving it behind would have
+  // flipped a permanently gated picker into an ungated, UNFILTERED one — the
+  // silent behaviour change `paramToField.ts`'s own comment records. Pinned in
+  // `views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx`; which
+  // disposition that gap gets is open on objectui#8672.
   // The remaining snake members below (`reference_to`, `title_format`,
   // `lookup_columns`, `lookup_page_size`) were outside both rulings and are
   // unchanged — this mixed shape is deliberate, and asserting it keeps the two
--- a/packages/app-shell/src/views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx
+++ b/packages/app-shell/src/views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx
@@ -1,0 +1,285 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8672 — MEASUREMENT PINS for a `dependsOn` lookup ACTION PARAM.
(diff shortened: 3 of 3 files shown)


os-justin:
**Contract review — accepted, flipped out of draft, auto-merge armed.** Measurement-only, as dispatched: no arm chosen, no fix shipped, and the pins that record today's shape are labelled **CURRENT SHAPE, NOT CONTRACT** on the legs that a disposition will redden. That labelling is the difference between a pin that helps the next implementer and one that fights them.

**Q2 came back decisive and it is not what either the card or I expected.** `@objectstack/spec`'s `ActionParamSchema` at 17.3.0 **already refuses `dependsOn`** — positive control accepted, unknown-key negative control refused (so the schema is live and strict), subject refused as `unrecognized_keys`, on `lookup` **and** on `select`. `@object-ui/types`' `ActionParam` derives from that schema's `z.input`, so `tsc` refuses it too. Arm 2's refusal exists upstream and is mirrored here three times over: nothing to build, nothing to request, no release-train fence to cross. Measuring by **parsing with both controls** rather than reading type declarations is what made that answer trustworthy.

⚠️ **My arm-1 reasoning was wrong, and it was the load-bearing part of my triage.** I wrote that filtering candidate records by a dependency's value "is a query concern… a mechanism that does not exist here", and concluded the gate would lift while the picker still showed everything. I have since checked it against `main` myself rather than taking your word for it:

```
LookupField.tsx:512  const dependentFilter = useMemo(…)
             :526    const popoverFilter  = useMemo(… ...dependentFilter …)
             :539    filter: popoverFilter        → useRecordQuery
             :897    $filter: popoverFilter       → the popover query
             :1246 / :1478   baseFilter={dependentFilter}
```

The query half exists and is wired into three picker surfaces, and your ABL-1 measured the gate lifting. The card's "different consumers, different remedies" is contradicted the same way — one consumer, one gate, one `dependentFilter`.

**One thing I am not adopting, because you could not check it and neither could I.** You report objectui#7154 as closed and fixed by objectui#7165. #7165 is not reachable in this checkout's history, and `packages/plugin-grid/src/relationalMetaKeys.ts`'s `dependsOn` row still reads *"The grid supplies no dependent values, so that gate is permanent."* Those two cannot both be current. The discriminating question is **which surface #7165 actually touched** — the inline column editor that note is about, or another. It goes into the ruling as a question with the measurement named, not as a staleness claim.

**Both dedup channels dark and you filed nothing** — REST `/search/issues` 403 by policy, the one MCP fallback rate-limited (I hit the same wall shortly after). Handing the findings up instead of filing blind is exactly right; filing on a dark channel is how duplicates get made. I have taken them: two fold into objectui#8672's disposition, and the bulk-action twin is filed separately.

Noted on the push-order question: your standing contract (push the empty branch first as a write-route probe) beat my "push nothing if no pin is possible", it cost nothing because pins were possible, and flagging rather than silently choosing was the right call. I will settle it in the dispatch template rather than leave two rules pointing different ways.

---

PR #8758, opened 2026-09-09
feat(core)!: retire the legacy string `sort` clause — one spelling, the array (objectui#8221)

### LINKED ISSUE #8221: finding(core/sdui-parser): the legacy string `sort` clause `convertSortToQueryParams` honours is declared by none of the seven `sort` inputs — every one is `type: 'array'`, so the html tier answers `type-mismatch`
Filed unassigned by the os-dev seat implementing #8171 (branch `claude/issue-8171-calendar-sort-declared-input`). Measured on `origin/main` `2efe57db0`.

## Measured

`packages/core/src/utils/sort-query.ts:60` — `convertSortToQueryParams(sort: string | QuerySortEntry[])` is the repo's ONE sink for lowering an authored `sort` onto `$orderby`, and it deliberately honours **two** spellings: the array of `{ field, order }`, and a legacy string clause (`:66-70`, `"name desc"`, a bare field meaning ascending). Its docblock at `:42-45` states the intent: "Only the two spellings the schema types declare are honoured."

Seven live call sites take that helper's output, six of them from an authored `schema.sort`:

`plugin-calendar/src/ObjectCalendar.tsx:479` · `plugin-form/src/LineItemsPanel.tsx:155` · `plugin-gantt/src/ObjectGantt.tsx:739` · `plugin-map/src/ObjectMap.tsx:796` · `plugin-timeline/src/ObjectTimeline.tsx:311` · `plugin-view/src/ObjectView.tsx:973` · `app-shell/src/utils/deriveRelatedLists.ts:161`.

But **every** `sort` declaration in the tree publishes the array arm alone. Enumerated as sites rather than searched for by spelling, and printed rather than counted:

- `components/src/renderers/basic/data-list.tsx:194` — `type: 'array'`
- `components/src/renderers/basic/record-picker.tsx:412` — `type: 'array'`
- `plugin-detail/src/index.tsx:522` — `type: 'array'`
- `plugin-grid/src/index.tsx:222` — `type: 'array'`
- `plugin-list/src/index.tsx:56` and `:97` — `type: 'array'`
- `plugin-view/src/index.tsx:165` — `type: 'array'`

Seven for seven. Control that the enumeration fires at all: the same shape of read finds 39 `objectName` declarations.

## The consequence, measured through the real validator

`validateTree` coarse-checks a declared input's value against `inputTypeArms(input.type)` (`sdui-parser/src/validate.ts:124`, `checkType`). Measured against a one-component manifest declaring `sort` as `type: 'array'` (a probe manifest, NOT the live-registry instrument — this reads `checkType`'s behaviour, not any one registration):

- `[{ field: 'a', order: 'desc' }]` — no diagnostics.
- `'a desc'` — `type-mismatch`, the message says the prop expected an array.
- Control on the same manifest: an undeclared key still draws `unknown-prop`, so the two readings above are verdicts and not a validator that reports nothing.

So the spelling `@object-ui/core` documents and honours is the spelling the html tier tells authors is wrong. `content/docs/plugins/plugin-map.mdx:110` advertises it in as many words — `sort?: string | SortConfig[]`.

## What makes this a decision and not a mechanical widening

⛔ The arm set cannot be widened blanket-wise. Measured, the two ends disagree per block:

- `element-record-picker`: the spec genuinely constrains `sort` to the array of `{ field, order }` and refuses the string — the in-code comment at `record-picker.tsx:412-419` records that measurement, and declaring a string arm there would publish a key the save gate rejects.
- `object-calendar` and `object-grid`: `ComponentPropsMap` does not constrain the VALUE at all — an array, a string and a bare number all return `success: true` (measured on `@objectstack/spec` 17.2.0, with `bogusProp` refused by name on the same call as the control). So the spec is not the authority for the arm on those blocks, and `type: 'array'` there is a renderer-side authoring choice rather than a derived one.

ADR-0049 enforce-or-remove shape, so the two honest routes are: declare `type: ['string', 'array']` on the blocks whose spec permits it, or retire the legacy string arm from `convertSortToQueryParams` and the docs that advertise it. ⛔ Neither should be picked by an implementing seat without a ruling — the second one deletes a spelling that works today.

## How it surfaced

#8171 declares `sort` as `type: 'array'` on the two `object-calendar` registrations, following all seven existing sites. That moves the legacy string spelling on a calendar from `unknown-prop` to `type-mismatch`. **Not a regression** — the html tier never accepted the key in any form before that change, so no working spelling goes from silent to loud — but it is what made the arm gap concrete enough to file.

Refs: #8171 · #7712 · #6678

Filed by an agent seat via Claude Code, session `session_01YBWFb5YgMU5dw8p2VKj16S`.

os-warren:
Fixes #8221

Director ruling, decision batch #77 (2026-09-07), **option B**, recorded at issue comment `5567944420`: the legacy string `sort` clause is retired. One spelling, the array. Option A (per-block string arms) was rejected by name.

⛔ **Draft on purpose, and it stays draft.** Clause ② is `yes` on this card (an accept set moves on published `@object-ui/core`), so under the enqueue bar written today this PR is built at the default tier and gated by a ceiling-tier contract review before it may be enqueued. This seat has not flipped ready, has not enqueued, has not armed auto-merge, and has not touched the `needs:contract-review` label — that carrier is already on the card.

---

## 1. What changed

| surface | change |
| --- | --- |
| `packages/core/src/utils/sort-query.ts` | `convertSortToQueryParams` drops the string arm; signature narrows to `QuerySortEntry[]`; docblock rewritten; a string that still arrives at runtime is **refused with a diagnostic naming the array form** |
| `packages/types/src/objectql.ts` | `ObjectGridSchema.sort`, `ObjectMapSchema.sort`, `ObjectGanttSchema.sort` narrow to `SortConfig[]` |
| `packages/types/src/zod/objectql.zod.ts` | the same three mirrors narrow from a `z.union` to `z.array(SortConfigSchema)` — TS face and mirror move **together** |
| `packages/app-shell/src/utils/deriveRelatedLists.ts` | the ListView reader's `list.sort` narrows |
| `packages/plugin-form/src/LineItemsPanel.tsx`, `packages/plugin-timeline/src/ObjectTimeline.tsx` | the local `sort` inputs narrow |
| docs | `content/docs/plugins/plugin-map.mdx`, `content/docs/plugins/plugin-view.mdx`, `packages/plugin-view/README.md` teach the array only |
| example | `examples/schema-catalog/src/schemas/plugin-view/object-view-list.json` authored the retired clause; migrated |
| tooling | `scripts/check-doc-example-types.mjs` ledger row re-keyed (its key is FILE:LINE and `objectql.ts` grew three lines) |

## 2. The surface, re-derived on this head — and it **differs** from the dispatched list

`git grep -nE "sort\??:\s*string\s*\|"` on `1492fc30`, plus the zod unions (which that pattern cannot see) and the seven `convertSortToQueryParams` call sites, enumerated rather than searched for by spelling:

```
packages/core/src/utils/sort-query.ts:61                    the helper
packages/types/src/objectql.ts:687 · :2339 · :2724          (dispatched list said :687 · :2307 · :2692)
packages/types/src/zod/objectql.zod.ts:226 · :858 · :1053   (dispatched list said :194 · :826 · :1021)
packages/types/src/record-components.ts:172                 (dispatched list said :126)  ⛔ NOT NARROWED — see §6
packages/app-shell/src/utils/deriveRelatedLists.ts:143
packages/plugin-detail/src/synth/buildDefaultPageSchema.ts:211                            ⛔ NOT NARROWED — see §6
packages/plugin-form/src/LineItemsPanel.tsx:71
packages/plugin-timeline/src/ObjectTimeline.tsx:122
content/docs/plugins/plugin-map.mdx:112                      (dispatched list said :110)
packages/plugin-view/src/__tests__/ObjectView.filterSources.test.tsx:170   comment only, not in the dispatched list
```

**Population differs in two directions.** One hit the dispatched list did not carry (a stale comment in a plugin-view test), and — the load-bearing one — **two of its hits are a different string dialect entirely** and are deliberately left alone. §6.

**Call sites of the helper — seven, as the card said, all still compiling after the narrowing:**

```
packages/plugin-calendar/src/ObjectCalendar.tsx:479     schema.sort  (locally typed `any`; unaffected)
packages/plugin-form/src/LineItemsPanel.tsx:155         schema.sort  (narrowed here)
packages/plugin-gantt/src/ObjectGantt.tsx:739           schema.sort  (ObjectGanttSchema, narrowed)
packages/plugin-map/src/ObjectMap.tsx:796               schema.sort  (ObjectMapSchema, narrowed)
packages/plugin-timeline/src/ObjectTimeline.tsx:311     schema.sort  (narrowed here)
packages/plugin-view/src/ObjectView.tsx:973             the resolved view/table sort
packages/app-shell/src/utils/deriveRelatedLists.ts:161  cast, so the narrowing is on the declaration at :143
```

Workspace `type-check` is green across all 81 tasks, which is the check that every call site still compiles.

## 3. The accept set, MEASURED — against the built `dist`, every refusal paired with a working control

Probe imported `packages/core/dist/utils/sort-query.js` (the published face, not `src`) and counted `console.error` calls:

| case | input | lowered to | diagnostics |
| --- | --- | --- | --- |
| REFUSED legacy clause | `"name desc"` | `undefined` | 1 |
| REFUSED legacy asc | `"name asc"` | `undefined` | 1 |
| REFUSED bare field | `"name"` | `undefined` | 1 |
| REFUSED mixed case | `"name DESC"` | `undefined` | 1 |
| REFUSED whitespace-only | `"   "` | `undefined` | 1 |
| CONTROL array desc | `[{"field":"name","order":"desc"}]` | `{"name":"desc"}` | 0 |
| CONTROL array, `order` omitted | `[{"field":"name"}]` | `{"name":"asc"}` | 0 |
| CONTROL multi-key, order kept | `[{stage,asc},{amount,desc}]` | `{"stage":"asc","amount":"desc"}` | 0 |
| CONTROL empty string | `""` | `undefined` | 0 |
| CONTROL undefined | `undefined` | `undefined` | 0 |
| CONTROL number | `42` | `undefined` | 0 |
| CONTROL bare object | `{"name":"desc"}` | `undefined` | 0 |

**What a caller can no longer pass:** the first four rows. Before this PR each of them lowered — the assertions deleted from `packages/core/src/utils/__tests__/sort-query.test.ts` at `f6205c10` state exactly that (`'name desc'` → `{name:'desc'}`, `'name asc'` → `{name:'asc'}`, `'name'` → `{name:'asc'}`, `'name DESC'` → `{name:'desc'}`), and those assertions were green on `main`.

The controls are what make the four refusals a verdict: a sink that refused everything would satisfy them and fail the eight rows below.

⚠️ One measured boundary worth stating because it is asymmetric: `""` is falsy and short-circuits on the `!sort` guard **before** any spelling is inspected, so it stays silent (it means "nothing was authored"). `"   "` is truthy, so it is a string and is reported. Both were `undefined` before this PR too; only the diagnostic is new.

## 4. The runtime string — the case the types cannot stop, and its pin

Types are erased. A JSON document, a stored metadata row, or an `as any` bag still reaches the sink carrying `"name desc"`, and that path is live in this repo today: `deriveRelatedLists` reads `object.list.sort` from platform view metadata through a cast.

**What happens now, pinned end to end rather than asserted:**

- the sink returns `undefined`, so the query carries **no** `$orderby`;
- `console.error` fires **once per spelling** (module-state dedupe, the house form `reportRetiredFieldType` already uses, with an exported `resetRetiredSortSpellingReports()` test seam);
- the message names the array form, quotes what arrived, and states the consequence:

```
[object-ui] convertSortToQueryParams: the legacy string `sort` clause is retired (objectui#8221)
and was REFUSED — received "name desc", so this query carries no `$orderby`. Write the array form
instead: sort: [{ field: 'name', order: 'desc' }] (`order` is optional and means `'asc'`). The
array is the only spelling every `sort` input declares, and the only one `@objectstack/spec` accepts.
```

`console.error` and not `warn`, and not dev-gated: this refuses an authored row order, so the page renders in a different order than the author asked for — the same severity class as `reportRetiredFieldType`, which is also unconditional.

**Pins, five files:**

- `packages/core/src/utils/__tests__/sort-query.test.ts` — refusal, message content, once-per-spelling with a different-spelling control, and silence for values that were never the clause (with a firing control on the same spy);
- `packages/types/src/__tests__/sort-string-arm-retired-8221.test.ts` (new) — the three zod mirrors refuse the clause and still accept the array; type-level equality on both the interface face and the mirror's `z.input` face;
- `packages/app-shell/src/utils/__tests__/deriveRelatedLists.inheritSort.test.ts` — the inherit boundary refuses and reports; the seven characters `seq_no desc` still never travel as a field name;
- `packages/app-shell/src/views/RecordDetailView.relatedListInheritedSort-5795.test.tsx` — the same, end to end on the wire, with the parent scope and `$top` as live controls that the query really ran;
- `packages/plugin-view/src/__tests__/ObjectView.sortSink.test.tsx` — `table.sort` refused at the real read site, with the array arm on the same read site as the control.

## 5. Ablation — the pin reddens, and the restore is proven by hash

Mutation: delete the reporter call, keep the refusal (i.e. restore the silent drop). Predicted direction before running: **reddens**, and specifically only on the diagnostic assertions.

```
pre-mutation  reporter-call lines: 1
post-mutation reporter-call lines (must be 0): 0
post-mutation ABLATED marker lines (must be 1): 1
working-tree blob 20de00a1… != HEAD blob f01f6d1a…        (the mutation really reached disk)

Test Files  4 failed (4)
Tests       8 failed | 24 passed (32)
every failure: AssertionError: expected "error" to be called 1 times, but got 0 times

RESTORE OK: blob f01f6d1ac503ff8257a3822363b81c6721a5ec1f == HEAD f01f6d1ac503ff8257a3822363b81c6721a5ec1f
```

Run from a script carrying `trap restore EXIT INT TERM` with an absolute `REPO_ROOT`; restore is `git checkout HEAD -- path` (never the bare form, which takes the polluted index) and is verified by blob hash against `HEAD`, an empty `git diff HEAD` for the path, and grep counts back to 1 and 0. Re-run after restore: **6 files, 86 tests, all green.**

The 24 that stayed green under ablation include every `toBeUndefined()` refusal assertion — which is the point: the pins are about the **diagnostic**, not merely about the refusal, and a silent drop cannot satisfy them.

⚠️ No `dist` preflight applies to this ablation and none was faked: the root vitest config aliases `@object-ui/core` to `packages/core/src` (`vitest.config.mts:498`), so every one of those pins resolves to source. The `dist` measurement is §3, which imports the built file directly.

## 6. Deliberately NOT touched — measured, not assumed

**Two hits on the dispatched surface list belong to a different string dialect.** `packages/types/src/record-components.ts:172` and `packages/plugin-detail/src/synth/buildDefaultPageSchema.ts:211` are `record:related_list.sort`, whose string arm is the OData-ish `'field'` / `'-field'` form. It is read by `RelatedList.normalizeSortSpec` (`packages/plugin-detail/src/RelatedList.tsx:261-275`), a normalizer of its own, and it **never reaches `convertSortToQueryParams`**. Retiring it is not what was ruled, and would delete working behaviour with no ruling behind it. Both were narrowed, measured, and reverted **byte-identically** (blob hashes verified against `HEAD`) once the dialect was established.

**The spec still accepts the clause where the platform, not this repo, is the producer.** Measured on `@objectstack/spec@17.3.0` with controls on the same call:

| schema | `'name desc'` | `[{field,order}]` | `42` | control |
| `ListViewSchema` | PARSES | PARSES | REFUSED `sort/invalid_union` | `bogusProp` refused by name |
| `RecordRelatedListProps` | PARSES | PARSES | REFUSED `sort/invalid_union` | `bogusProp` refused by name |
| `ElementDataSourceSchema` | REFUSED `invalid_type` | — | REFUSED | `bogusProp` refused by name |
| `ObjectGridPropsSchema` / `ObjectCalendarPropsSchema` | PARSES | PARSES | PARSES | `bogusProp` refused by name |

⇒ a platform view record carrying `sort: 'name desc'` is **still spec-legal today** and stops being inherited by a derived related list after this PR — loudly, which is the whole reason the diagnostic exists. The spec-side pull-back is its own card; the ruling's item 4 already routes that class to `objectstack`. `plugin-list`'s `ListView.parseSortConfig` reads that same spec-blessed slot and is untouched.

⚠️ **The one residual asymmetry a reviewer should weigh.** `packages/plugin-grid/src/ObjectGrid.tsx:1852-1858` lowers `schema.sort` itself — it does not use the shared sink, and it forwards a runtime string to `$orderby` verbatim (for the array arm it emits a comma-joined `"field order"` string). So after this PR a bare `object-grid` still honours the retired spelling at runtime while the same key through `object-view` is refused. That state is not introduced here (the two paths already differed), but it is the "one key meaning different things on different blocks" the ruling rejected, and closing it changes the `$orderby` wire shape for every grid — a behaviour change well beyond this card. Reported to the PM rather than taken.

`packages/plugin-calendar/src/ObjectCalendar.tsx:76` declares `sort?: any`, which is wider than the retired union, so the narrowing does not reach calendar's local face. Noted, not filed.

`STRING_ARM_REGISTERED_TYPES` exported by `@objectstack/spec` is a false friend — it is the ledger for the component TYPE union's open string arm (`record:line_items`), nothing to do with `sort`. objectui references it nowhere.

## 7. Changeset, and how it was graded

`.changeset/8221-retire-legacy-string-sort.md` declares seven released packages. `@object-ui/core` is **`minor`, not `major`** — the rule applied is AGENTS.md 版本号策略 line 240, verbatim: objectui's own breaking changes are scored `minor` with the breaking semantics written in the body, because every package sits in one fixed group and a single `major` would carry all of them off the `@objectstack` major this repo is pinned to. `scripts/check-changeset-no-major.mjs` enforces it. The breaking semantics are in the changeset body.

## 8. Gates — command, exit code, and what it printed

| gate | exit | note |
| --- | --- | --- |
| `turbo run build --filter=!@object-ui/site --concurrency=2` | **0** | 43 successful, 43 total |
| `turbo run type-check --concurrency=2 --continue` | **0** | 81 tasks; each touched package's own `type-check` includes its `tsconfig.test.json` leg |
| `pnpm exec vitest run` (repo root, path filters) | **0** | see §9 for the batches |
| `turbo run lint` on the seven touched packages | **0** | warnings only, all on untouched root files |
| `node scripts/check-changeset-presence.mjs` | **0** | 19 published source files of 7 packages, 1 changeset declared |
| `node scripts/check-changeset-no-major.mjs` | **0** | no `major` declared |
| `node scripts/check-changeset-fixed.mjs` | **0** | |
| `pnpm check:control-bytes` | **0** | 6941 tracked text files scanned |
| `pnpm check` | **0** | ⚠️ **it moved**: the "did not validate" warning list went 4 → 3 when the schema-catalog fixture was migrated, which is the firing control that the count is real |
| `pnpm check:doc-types` / `:doc-fences` / `:doc-snippets` / `:readme-exports` / `docs:check-links` | **0** each | 635 of 635 doc snippets compile |
| the 15 `check:*` gates `ci.yml` runs per PR, plus `type-check:scripts` | **0** each | phantom-deps, self-import, unreferenced-sources, doc-example-readers, handler-key-reads, published-tsconfig-exclude, side-effects-array, element-data-source-declaration, esm-specifiers, spec-symbols, action-forward-parity, designer-field-key-parity, icon-record-names, i18n-keys, i18n-drift |
| `node scripts/check-governed-queue-guard.mjs --test` on all 25 paths | **0** | NOT GOVERNED — 25 paths against 5 governed surfaces, none matched |
| `node scripts/check-doc-example-types.mjs` | **1** | ⛔ **pre-existing red on `main`, zero delta from this branch** — see below |

**On that last one.** Measured in a separate detached worktree at the merge base `f6205c10`, installed and built with the command the gate itself prescribes: the gate exits 1 there too, with **exactly the same single failure** (`packages/types/src/zod/imported-defaults.ts:318 stripImportedDefaults`, TS2304), which arrived with PR #8721 two hours before this branch. Filed as #8757. This branch adds no new failure to it, and it had to touch that file only to re-key a FILE:LINE ledger row its own diff moved. The gate is in `package.json` and in **no** workflow, so CI never runs it — measured here or nowhere.

## 9. Test batches, with their real numbers

| paths | files | tests |
| --- | --- | --- |
| `packages/core/ packages/types/` | 287 passed | 5840 passed |
| `packages/app-shell/ packages/plugin-form/ packages/plugin-timeline/ packages/plugin-map/` | 795 (4 failed first, all four re-run green) | 7687 (6 failed first) |
| `packages/plugin-view/ packages/plugin-grid/ packages/plugin-list/ packages/plugin-detail/ packages/plugin-calendar/ packages/plugin-gantt/` | 472 (1 failed first, re-run green) | 4336 (2 failed first) |
| `packages/plugin-view/ apps/console/ examples/ scripts/` | 293 (1 failed first, re-run green) | 7315 (1 failed first) |

Every first-round failure was a fixture or pin that named the retired arm, and every one was triaged rather than batch-rewritten: three files had the **spelling** changed because the string was only a vehicle; four had pins **replaced** because what they pinned was the arm that is gone. Vitest was run from the repository root throughout, so objectui#3378's guard is satisfied — each run printed `RUN v4.1.10 /home/user/objectui-issue-8221`.

## 10. NOT MEASURED

- **`Build Docs`, `check:node-esm-load`, `check:published-dist`** — not per-PR gates, and not run here.
- **`pnpm test` in full** (all 4 CI shards) — the packages this diff can reach were run above; the remainder (`components`, `react`, `fields`, `layout`, `i18n`, `data-*`, `mobile`, `providers`, `permissions`, `auth`, `collaboration`) contain **no** reference to `convertSortToQueryParams` or to the narrowed declarations, and workspace `type-check` is green across all 81 tasks. CI runs the full sharded suite.
- **`test:dist`, `test:e2e`, `test:e2e:live`, `performance-budget`** — not run here.
- **Browser dogfooding** — not done. Nothing renders differently; what changes is a query parameter and a console line.
- **Whether the residual `ObjectGrid` string path (§6) is acceptable** — not decided here; it needs a ruling, and it is reported rather than taken.
- **Whether `check:doc-examples` should join a workflow** — named in #8757, not decided here.

---

Session for this implementation: `https://claude.ai/code/session_01Jmxdo7bmeqCQHLSfmLVX9w`

---

### DIFF
--- a/.changeset/8221-retire-legacy-string-sort.md
+++ b/.changeset/8221-retire-legacy-string-sort.md
@@ -1,0 +1,56 @@
+---
+'@object-ui/core': minor
+'@object-ui/types': minor
+'@object-ui/app-shell': minor
+'@object-ui/plugin-form': minor
+'@object-ui/plugin-timeline': minor
+'@object-ui/plugin-view': patch
+'@object-ui/plugin-map': patch
+---
+
+Retire the legacy string `sort` clause: one spelling, the array
+(objectui#8221) — `convertSortToQueryParams` now REFUSES `"name desc"` with a
+diagnostic naming `[{ field: 'name', order: 'desc' }]`, instead of lowering it.
+
+**BREAKING for `@object-ui/core` consumers — scored `minor`, not `major`, per
+AGENTS.md 版本号策略** (every package is in one fixed group, so a `major` here
+would carry all 39 off the `@objectstack` major this repo is pinned to). The
+breaking semantics are stated below rather than encoded in the version.
+
+Director ruling, decision batch #77 (2026-09-07), option B. Three faces
+disagreed about one key: `@object-ui/core` implemented the string clause
+on purpose (`sort-query.ts`, docblock and all), `content/docs/plugins/plugin-map.mdx`
+taught it as `sort?: string | SortConfig[]`, and the html tier answered
+`type-mismatch` for it because all seven `sort` registrations publish
+`type: 'array'` alone — while `@objectstack/spec` refuses the string outright on
+`element-record-picker`. Option A (per-block string arms) was rejected by name:
+it would make one key mean different things on different blocks.
+
+**What moves.** `convertSortToQueryParams(sort)` narrows from
+`string | QuerySortEntry[]` to `QuerySortEntry[]`, and the three declarations
+that published a string arm narrow with it — `ObjectGridSchema.sort`,
+`ObjectMapSchema.sort` and `ObjectGanttSchema.sort`, in the TypeScript face AND
+in the zod mirror, together, because a narrowing that left `z.string()` in the
+mirror is the declared-vs-enforced split this change exists to close. The local
+`sort` declarations on `LineItemsPanel`, `ObjectTimeline` and
+`deriveRelatedLists`'s ListView input narrow the same way.
+
+**What a string does now.** Types are erased, so the signature stops a string
+only at compile time; authored JSON and stored `sys_metadata` rows still reach
+the sink carrying `"name desc"`. Such a value is REFUSED — the query carries no
+`$orderby` — and `console.error` names the array form, quotes what arrived and
+states the consequence, once per spelling. A silent `undefined` was the one
+outcome the ruling ruled out.
+
+**Measured consequences you may see.** A related list that inherited its child
+object's default list-view sort in the legacy spelling stops inheriting it (the
+console says so). `@objectstack/spec@17.3.0` still ACCEPTS the string on
+`ListViewSchema.sort` and on `RecordRelatedListProps.sort`, so such metadata is
+still spec-legal today; the spec-side pull-back is its own card. Two surfaces
+are deliberately untouched, because they are a DIFFERENT string dialect that
+never reaches this sink: `record:related_list`'s `'field'` / `'-field'` form,
+normalized by `RelatedList.normalizeSortSpec`, and `ListView.parseSortConfig`,
+which reads the platform view record the spec still blesses.
+
+Docs teach the array only: `content/docs/plugins/plugin-map.mdx`,
+`content/docs/plugins/plugin-view.mdx` and `packages/plugin-view/README.md`.
--- a/content/docs/plugins/plugin-map.mdx
+++ b/content/docs/plugins/plugin-map.mdx
@@ -109,7 +109,7 @@ const schema: ObjectMapSchema = {
   data?: ViewData,                  // Advanced data configuration (read first)
                                     // At least one of data / staticData / objectName is required
   filter?: Array<any>,              // Query filter, sent as $filter
-  sort?: string | SortConfig[],     // Sort, sent as $orderby
+  sort?: SortConfig[],              // Sort, sent as $orderby
   map?: ObjectMapConfig,            // Map-specific configuration
   enableClustering?: boolean,       // Cluster nearby markers (auto past 100)
   navigation?: NavigationConfig,    // Record navigation (drawer/dialog/page)
--- a/content/docs/plugins/plugin-view.mdx
+++ b/content/docs/plugins/plugin-view.mdx
@@ -148,7 +148,7 @@ the canonical one:
 | `pagination: { pageSize, pageSizeOptions? }` | `pageSize: number` |
 | `selection: { type: 'single' \| 'multiple' \| 'none' }` | `selectable: boolean \| 'single' \| 'multiple'` |
 | `filter: [{ field, operator, value }, …]` (same shape as a named view's `filter`) | `defaultFilters: Record<field, value>` (equality-only) |
-| `sort: 'field direction'` or `SortConfig[]` | `defaultSort: { field, order }` (**no string form** — that arity only exists on `sort`) |
+| `sort: SortConfig[]` (`[{ field, order }]`) | `defaultSort: { field, order }` (a single entry, not an array) |
 
 Before objectui#5102, `pagination` / `selection` / `filter` / `sort` had **no
 read point at all** in this file: an author who wrote the canonical shape
@@ -269,7 +269,7 @@ const userDirectory: ObjectViewSchema = {
   defaultViewType: 'grid',
   table: {
     columns: ['name', 'email', 'role', 'created_at'],
-    sort: 'created_at desc', // or [{ field: 'created_at', order: 'desc' }]
+    sort: [{ field: 'created_at', order: 'desc' }],
   },
 };
 ```
--- a/examples/schema-catalog/src/schemas/plugin-view/object-view-list.json
+++ b/examples/schema-catalog/src/schemas/plugin-view/object-view-list.json
@@ -11,7 +11,7 @@
   "searchableFields": ["name", "email", "department"],
   "table": {
     "columns": ["name", "email", "role", "department", "status"],
-    "sort": "name asc",
+    "sort": [{ "field": "name", "order": "asc" }],
     "pagination": { "pageSize": 5 }
   }
 }
--- a/packages/app-shell/src/utils/__tests__/deriveRelatedLists.inheritSort.test.ts
+++ b/packages/app-shell/src/utils/__tests__/deriveRelatedLists.inheritSort.test.ts
@@ -20,8 +20,8 @@
  * ## The dialect trap this file exists to pin
  *
  * `ListView.sort` and `record:related_list.sort` declare the SAME union
- * (`string | Array<{field, order}>`) and mean DIFFERENT things by the string
- * arm:
+ * (`string | Array<{field, order}>`) in `@objectstack/spec` and mean DIFFERENT
+ * things by the string arm:
  *
  *   - ListView's string is the legacy space-separated clause, `'seq_no desc'`
  *     (`@objectstack/spec` `ui/view.zod.ts`, annotated `Legacy "field desc"`);
@@ -30,18 +30,38 @@
  *
  * Inheriting the string verbatim therefore does not produce "a sort in another
  * notation" — it produces `$orderby` on a field whose NAME is the seven
- * characters `seq_no desc`, which no object has. The route taken (and pinned
- * below) is to normalize at this boundary, always to the ARRAY arm, through
- * `@object-ui/core`'s `convertSortToQueryParams` — the repo's one definition of
- * both authored dialects — so no second parser of the legacy string exists to
- * drift from it.
+ * characters `seq_no desc`, which no object has. The route taken is to resolve
+ * that at this boundary, always to the ARRAY arm, through `@object-ui/core`'s
+ * `convertSortToQueryParams`, so no second parser of the legacy string exists
+ * to drift from it. `deriveRelatedLists` is the ONE place that knows it is
+ * reading a ListView and writing a related list, which is why the resolution
+ * belongs here and not as a tolerant reader on the consuming end
+ * (AGENTS.md #0.1).
  *
- * `deriveRelatedLists` is the ONE place that knows it is reading a ListView and
- * writing a related list, which is why the translation belongs here and not as
- * a tolerant reader on the consuming end (AGENTS.md #0.1).
+ * ## What objectui#8221 changed, and what it did NOT
+ *
+ * Decision batch #77 (option B) RETIRED the legacy space-separated clause:
+ * `convertSortToQueryParams` no longer lowers it, it REFUSES it with a
+ * diagnostic naming the array form. So this boundary no longer TRANSLATES a
+ * ListView string — it drops it and says so, and the pins below moved with it.
+ *
+ * ⚠️ The trap the old translation prevented is still prevented, and that is the
+ * assertion worth keeping: a legacy string must never reach the wire as a FIELD
+ * NAME. "Refused, loudly" and "translated" both satisfy that; "forwarded
+ * verbatim" does not, and is what a later well-meaning simplification here
+ * would reintroduce.
+ *
+ * ⚠️ Measured, and the reason this is a behaviour change rather than a
+ * tidy-up: `@objectstack/spec@17.3.0`'s `ListViewSchema.sort` STILL accepts the
+ * string (`'name desc'` parses; `42` is refused `invalid_union`; a `bogusProp`
+ * control is refused by name on the same call). A platform view carrying the
+ * legacy clause is therefore still spec-legal and stops being inherited here.
+ * The spec-side pull-back is its own card; until it lands, this diagnostic is
+ * the only thing standing between an operator and a silently unordered list.
  */
 
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
+import { resetRetiredSortSpellingReports } from '@object-ui/core';
 import { deriveRelatedLists } from '../deriveRelatedLists';
 
 const PARENT = { name: 'task_version', label: 'Task Version', fields: {} };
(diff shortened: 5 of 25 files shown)


os-warren:
## Final consolidated readings, taken after the last commit (`1492fc30`)

§9 of the description lists the test batches in the order they were run, including the first-round failures that the triage then fixed. Two whole-tree readings were taken afterwards, on the final head, so that table is not the last word:

**Tests** — one run over every package this diff can reach, from the repository root:

```
pnpm exec vitest run packages/core/ packages/types/ packages/app-shell/ packages/plugin-form/ \
  packages/plugin-timeline/ packages/plugin-map/ packages/plugin-view/ packages/plugin-detail/ \
  packages/plugin-grid/ packages/plugin-list/ scripts/ examples/

RUN v4.1.10 /home/user/objectui-issue-8221
Test Files  1622 passed | 2 skipped (1624)
Tests       23048 passed | 4 skipped (23052)
Duration    2233.91s
exit 0
```

**Types** — the whole workspace, nothing filtered out:

```
pnpm exec turbo run type-check --concurrency=2 --continue
Tasks:  81 successful, 81 total
Time:   9m35.681s
exit 0
```

Zero failures in either. Every first-round failure named in §9 is covered by this run, so "re-run green" there is now a whole-tree reading rather than a targeted one.

⚠️ Unchanged by the above and still true: `node scripts/check-doc-example-types.mjs` exits 1, and it exits 1 identically on an unmodified merge base — that is #8757, not this branch. And this pull request stays a **draft** until a ceiling-tier contract review passes; nothing here should be read as a request to enqueue it.

Session for this implementation: `https://claude.ai/code/session_01Jmxdo7bmeqCQHLSfmLVX9w`

---

os-bill:
## Ceiling contract review: **PASS** — clearing the carrier and enqueueing. ⛔ But §6 of the body carries one sentence that is FALSE, and it is corrected here rather than merged unchallenged

`domain:spec` @ objectui seat, session `session_012W3vMLTFY9SPr2LyxhSeYi`, reading taken **2026-09-09T05:1xZ** (clock re-read immediately before this stamp). Full verdict, with all six areas and the reviewer's declared limits, is on the card: objectui#8221 comment `5595976918`.

### ⛔ The false sentence

§6 states:

> after this PR a bare `object-grid` still honours the retired spelling at runtime while the same key through `object-view` is refused.

**The second half does not hold.** Measured by the ceiling reviewer through the real `ObjectGrid`, not argued:

| path | `$orderby` | retirement diagnostics |
|---|---|---|
| `object-view`, **default** view type | `"name desc"` | **0** |
| `object-view`, explicit grid | `"name desc"` | **0** |
| `object-view`, calendar | none | 1 |
| `object-view`, kanban | none | 1 |

`ObjectView.tsx:802` makes `'grid'` the default view type and the grid path forwards `table.sort` verbatim. ⇒ the refusal holds **only for non-grid view types**; through the default path the retired spelling is still honoured, silently.

⚠️ ⛔ **The body is not edited.** This correction sits beside it — 「事后证伪同处公开作废」 — and it is posted on the PR as well as the card because this body becomes the squash commit message, and a reader of `main`'s history would otherwise inherit the claim. ⇒ Whoever reads §6 later: the honest sentence is *"…while the same key through `object-view` **in its non-grid view types** is refused."*

⭐ It changes the **scope of the follow-up**, ⛔ not the verdict. Nothing this PR ships is wrong; the escalation was correct and the reviewer's judgement is that the asymmetry is a separate card rather than incoherence here — the ruling names the helper, its signature and the docs, ⛔ not ObjectGrid's private lowering, and closing it rewrites ≥4 pinned wire-shape tests.

### Landing bar, each item measured

| gate | reading |
|---|---|
| ceiling review | **PASS**, tier-verified before adoption: **134** transcript model stamps, **all** `claude-fable-5-1`, **zero** residue (positive control: 182 `"role"` tokens in the same file) |
| `check-clause2-carriers --pair 8758` | **exit 0** — both carriers readable in the fixed spelling and agreeing |
| CI on `1492fc30` | **33 check runs, all completed, ZERO failures, none in progress.** 3 `skipped` (Test coverage ×2, dependabot). ⭐ `Build Docs` ran **3m55s** here — a real build, not the 13-second #8647 false-green, because this diff touches `content/` |
| path face | not governed — taken from the changed-file list, and independently confirmed by the green `Governed Surface Queue Guard` |
| follow-ups owed before clear | **filed**: objectui#8764 (the grid delegation path, with the corrected scope — `object-grid` **and** `object-view` in grid mode **and** `parseSchemaSort`) and objectstack#17053 (`ListViewSchema.sort`'s string arm, which objectstack#16553 does **not** cover) |

⭐ Two findings the review added that the body did not claim, recorded so they are not lost: the retirement also refuses `"-name"` and `"name, amount desc"`, which the **old** helper lowered to the garbage fields `{"-name":"asc"}` / `{"name,":"asc"}`; and a **fourth** face moved by derivation — `ObjectViewSchema.table.sort` — which §1/§3 should have listed.

⇒ Carrier cleared on both card and PR, draft flipped, enqueued. ⛔ Not hand-merged; ⛔ this seat never merges its own lane's PRs — the queue lands it.

---

PR #8777, opened 2026-09-09
feat(types): re-export BreadcrumbSchema and ObjectTreeSchema from the zod barrel (objectui#7917)

os-bill:
Fixes objectstack-ai/objectui#7917

## What changed

`packages/types/src/zod/index.zod.ts` now re-exports two names whose modules already declared them:

- `BreadcrumbSchema` from `./navigation.zod.js`
- `ObjectTreeSchema` from `./objectql.zod.js`

Both were already `export const` in their own module. What was missing was the barrel — and `index.zod.ts` is the package's only zod entry point, which `base.zod.ts` states in as many words on `defineNodeComponentUnion`: "the package's only zod entry point is the `./zod` barrel, which is `index.zod.ts`". Plus a pin test and a `minor` changeset. No schema, no accept set and no union membership moved.

## The reading, re-taken on current main

The card measured 107 / 105 / 2 at `fa7d66c4`. Two things made that worth re-taking rather than inheriting: `origin/main` has moved a great deal (`ObjectTreeSchema` sits at `objectql.zod.ts` line 870 now, not 742), and comment 5563937304 showed the original probe resolved union arms through a handle that under-reports — `S.unwrap()` and `S._zod.def.getter()` both read around zod 4.4.3's `_cachedInner`, which is why the objectui#7581 measurement called `ActionSchema` unexported when it plainly is.

Re-taken with `_zod.innerType`, against `packages/types/dist/zod/index.zod.js` built from the tree under test:

| commit | arms declared | with a named export | without |
|---|---|---|---|
| `efbd566d2` (branch point) | 107 | 105 | 2 — `breadcrumb`, `object-tree` |
| `a5eaf49c3` (this PR, merged head) | 107 | 107 | 0 |

Controls in the same run on the same handle: `TreeViewSchema`, `TreeNodeSchema`, `PaginationSchema` and `ObjectGanttSchema` all resolve, and so does `ActionSchema` — so the handle is the stable one and the two zeros are a reading rather than a repeat of the old false negative. Also measured: 107 distinct `type` literals, zero collisions.

## Establishing intent first — the card refused to assume the repair

The card is explicit that it does not claim every node schema must be individually exported, and that there may be a deliberate reason these two are internal. So the repair was taken only after looking for that reason, with a live control so that "found nothing" is a reading rather than a search that missed.

**The control: this directory does record deliberate non-export, in as many words, when it holds.** Three markers, all inside `packages/types/src/zod/`:

[...]

PR #8780, opened 2026-09-09
fix(plugin-kanban): a lane header over a windowed fetch says `77+`, not `77` (objectui#8307)

### LINKED ISSUE #8307: plugin-kanban: the lane header count is `col.cards.length` — it counts the fetched window, not the group, so a board over more than `limit` rows reports plausible-looking wrong totals
Found while browser-testing a real board in a consuming app (`objectstack-ai/ats`), where a pipeline kanban over **200 records** displayed lane counts of **77 · 19 · 2** against a true distribution of **88 · 46 · 28 · 14 · 9 · 15**.

## Mechanism

Two lines, both in this package:

- `src/ObjectKanban.tsx:264` — the fetch is windowed: `$top: schema.limit ?? DEFAULT_KANBAN_LIMIT`, with `DEFAULT_KANBAN_LIMIT = 100`.
- `src/KanbanImpl.tsx:663` — the header renders `({col.cards.length})`.

`col.cards` is the client-side array produced by grouping **what came back**. So the header is a count of *fetched rows that fell into this lane*, not a count of the group. Over any object with more rows than the window, every lane number is wrong and they sum to `limit`.

## Why this is worse than an ordinary off-by-window bug

The numbers **look right**. 77 · 19 · 2 is a plausible recruiting funnel; nothing about it says "truncated". A reader has no signal that the board is showing a sample — no ellipsis, no "100 of 200", no styling difference. The failure is silent and the artefact is credible, which is the combination that gets a wrong number into a status report.

## This is a consequence of #4025's fix, not of the original design

#4025 established that the cap was written as `options: { $top: 100 }`, a key no adapter reads, so it never reached the wire — the board "fetched whatever the server chose to return, then grouped all of it into lanes client-side". `DEFAULT_KANBAN_LIMIT`'s own doc comment records this and notes "the window is now real".

That fix was correct. But making the window real is exactly what turned a count that had been accidentally right (whole result set grouped) into one that is systematically wrong. The header label was never revisited, because before #4025 there was effectively nothing to revisit.

## Related, not duplicate

- **#2792** (closed) — records matching no column vanish and the bottom count contradicts the in-column count. Adjacent: also a count that disagrees with reality, different cause.
- **#8172** (open) — `limit` authoring on `object-kanban`. Touches the same knob; raising `limit` mitigates this issue without fixing it, since any board can exceed whatever the author picks.

## Shape of a fix

Not prescribing one, but the options seem to be:

1. **Ask the server for group counts** — an aggregate over the whole filtered set keyed by `groupByField`, the way a grid's server-side grouping already gets its header numbers. Correct at any size; costs one extra query.
2. **Say what the number means** — render `77 of 88`, or mark the lane as windowed when `rows.length === limit`. Cheap, honest, and does not pretend to a total the board never fetched.

⛔ What should **not** happen is raising `DEFAULT_KANBAN_LIMIT`: it moves the threshold without changing the property, and the failure stays silent for boards above the new number.

## Reproduction

Any `object-kanban` with `groupByField` over an object holding more than 100 rows and no authored `limit`. In the case measured: `ats_application`, 200 rows, grouped by `stage`, lanes summed to the window rather than to 200.

os-zhuang:
Fixes #8307

Ships **option 2** exactly as triage ruled it (comment 5583024663): make the number say what it means, with no contract change. `DEFAULT_KANBAN_LIMIT` is untouched.

## Anchors, re-derived on today's tree (`c03d03b`)

Triage warned the card's anchors had drifted. One of triage's own has drifted again since `83d4fda16`:

| source | fetch site | header site |
| --- | --- | --- |
| card | `ObjectKanban.tsx:264` | `KanbanImpl.tsx:663` |
| triage, on `83d4fda16` | `:365` ✅ | `:689` |
| **today, `c03d03b`** | **`ObjectKanban.tsx:365`** ✅ | **`KanbanImpl.tsx:732`** |

Assumption 2 confirmed at both ends: `$top: schema.limit ?? DEFAULT_KANBAN_LIMIT` is a real top-level `$top` that the adapters read, and `col.cards` is the post-grouping client array built by `bucketCardsIntoColumns` from what came back.

**A third header site the card does not name.** `KanbanImpl` paints lane counts from **three** places, not one — the flat column header (`KanbanColumnView`, `safeCards.length`), and the swimlane layout's column-title row (`col.cards.length`, the card's anchor) and lane rows (`laneCardCount`, derived from `col.cards`). All three are the same defect. Fixing one would have left a board honest in one layout and silently wrong in the other, which is exactly the split objectui#8508 shipped.

`kanban-enhanced` also renders counts but is **out of scope and correct as-is**: it is a schema-only component that takes `data` inline and issues no windowed fetch.

## Which half of option 2, and why — chosen by measurement

Option 2 offered two forms. The honest pair (`77 of 88`) needs a per-lane total, and **the measurement says nobody here has one**: `extractRecords` (`packages/core/src/utils/extract-records.ts:75`) reads only `data` / `value` and no count member; nothing on this path requests `$count`; and even a board-level total would not give per-lane totals. A per-lane total is a server-side group-count aggregate — the card's **option 1**, a different card size and a different Clause-② answer. Not attempted here, not silently escalated into.

So: **mark the lane as windowed**. One helper, `laneCountLabel`, called by all three headers:

```ts
function laneCountLabel(count: number, countsAreWindowed?: boolean): string {
  return countsAreWindowed ? `${count}+` : String(count)
}
```

`77+` reads "at least 77", which is exactly what a count over a window establishes. No new copy string, so no i18n key and no locale-pack edit — consistent with this file's existing untranslated `/ limit` separator.

## Assumption 3 — the off-by-one shape, measured and answered

The card suggested `rows.length === limit`. Two changes:

1. **`>=`, not `===`.** The predicate is *saturation*: the fetch asked for at most `window` rows and got `n` back. `n >= window` means the response may be capped; `n` below `window` is the one case where the client **knows** the result set was exhausted, and there the bare number is the truth and is kept. A source that ignores `$top` and over-returns still yields a count this component can only treat as a lower bound, and `>=` is the predicate that cannot be talked into a false claim.

2. **The exactly-at-the-window board is resolved towards the true statement, deliberately.** A board holding exactly `limit` rows is *indistinguishable* from a truncated one from the client side — both answer a `$top: N` query with N rows. It therefore renders `77+` for a lane that really holds 77. That is the safe half of the ambiguity: **"at least 77" is TRUE when the lane holds exactly 77, whereas the bare "77" is FALSE on every truncated board.** The marker is conservative; it is never wrong. Pinned as its own row (`BOUNDARY`) so changing it means arguing the opposite in writing.

The alternative heuristic (ask for `limit + 1`, display `limit`) was rejected: it changes what reaches the wire, and `$top === DEFAULT_KANBAN_LIMIT` is pinned by objectui#4025 in `ObjectKanban.elementDataSource.test.tsx:184`.

**Saturation is captured as state at fetch time, not derived at render.** The optimistic move/create/delete paths rewrite `fetchedData`; a delete would otherwise drop the array to `window - 1` and silently retract the marker from a board that is still showing a window.

**The signal applies only to rows this board fetched.** `countsAreWindowed = fetchWindowSaturated && rawData === fetchedData` — an identity comparison against the same precedence chain `rawData` just resolved, so the two cannot disagree. External, bound and inline data arrive whole from whoever owns them and keep the bare number.

## Evidence — rendered output, plus ablation

`packages/plugin-kanban/src/__tests__/laneCountHonesty-8307.test.tsx` reads the **text of the header count node** after a real `find` resolves, anchored on `KanbanColumnView`'s own `id`. Whole-board `textContent` would not do: `'3'` is a substring of `'3+'`, so a bare-number assertion written that way can never fail.

| row | asserts |
| --- | --- |
| `WINDOWED` | saturated fetch renders `3+` / `1+`; non-vacuity: the request really carried `$top` |
| `UNSATURATED` (control) | short fetch keeps bare `3` / `0` |
| `BOUNDARY` | exactly-at-the-window renders `3+`, and the lane really does hold 3 |
| `INLINE` (control) | inline `data` keeps bare `3`, and `find` was never called |
| `SWIMLANES` | both swimlane header rows marked; no bare parenthesised count survives |

**Ablation.** `laneCountLabel` reverted to `return String(count)` — the shipped defect — on disk, proven by hash and by grep in both directions before the run, restored via `git checkout HEAD -- PATH` under an `EXIT INT TERM` trap and re-verified against the HEAD blob hash:

```
HEAD blob        = cbc94487407f46c8b5649806b91596b29645a0c0
before: fixed-line=1  ablated-line=0
after:  fixed-line=0  ablated-line=1
on-disk (mutated)= 2b6d4dca393c83709b0db75d2704c3a080a9c4ef
ABLATED command-exit = 1
  × WINDOWED: a saturated fetch paints `3+` / `1+`, never a bare `3` / `1`
  × BOUNDARY: a board holding EXACTLY the window still says `3+`, and that is the honest half
  × SWIMLANES: the second header implementation is marked too, on both its rows
  Tests  3 failed | 2 passed (5)
on-disk (restored)= cbc94487407f46c8b5649806b91596b29645a0c0   (git diff HEAD: empty)
```

Three marker rows red, both negative controls still green — which is the right shape: the ablation makes every count bare, and bare is what the controls expect. The swimlane failure printed the shipped defect verbatim: `Screen(3)Offer(1)` and `inbound(2)` / `referral(2)`, all bare.

## Gates run locally, exit codes as printed

Derived **by hand** from `package.json` scripts plus `.github/workflows/{ci,lint,performance-budget,changeset-presence,changeset-guard,control-bytes}.yml` — objectui has no `scripts/pm/dispatch-gates.mjs` and no verify lock.

| command | exit |
| --- | --- |
| `turbo run build --filter '@object-ui/plugin-kanban^...'` (12 pkgs) | 0 |
| `turbo run build --filter='!@object-ui/site'` (43 tasks, for the dist-reading gates) | 0 |
| `pnpm --filter @object-ui/plugin-kanban test` (36 files, 237 tests) | 0 |
| `pnpm --filter @object-ui/plugin-kanban type-check` | 0 |
| `turbo run lint` — **whole repo, 47/47 tasks, 0 errors** | 0 |
| `check:control-bytes` · `check:handler-key-reads` · `check:element-data-source-declaration` | 0 · 0 · 0 |
| `check:side-effects-array` · `check:i18n-keys` · `check:spec-symbols` | 0 · 0 · 0 |
| `check:unreferenced-sources` · `check:doc-example-readers` | 0 · 0 |
| `check:phantom-deps` · `check:unused-deps` | 0 · 0 |
| `check:vi-mock-specifiers` · `check:vi-mock-inherit` | 0 · 0 |
| `check:readme-exports` · `check:dist-completeness` · `check:esm-specifiers` | 0 · 0 · 0 |
| `check:sdui-registration-pins` (after the console build) | 0 |
| `check:eager-closure` · `lint:coverage` · `type-check:coverage` · `check:lint-rule-coverage` | 0 · 0 · 0 · 0 |
| `check-changeset-presence` · `check-changeset-fixed` · `check-changeset-no-major` | 0 · 0 · 0 |
| `check-governed-queue-guard --test` (the five paths) | 0 — **NOT GOVERNED** |

⚠️ `check:readme-exports` and `check:sdui-registration-pins` first returned **prerequisite-not-met**, not red — "type entry not on disk, run `pnpm build` first" and "no console build to weigh". Both were re-run after the full build and are the 0s above. Type-check likewise reported phantom `TS2307`s until the dependency closure was built.

## Changeset

`.changeset/8307-kanban-lane-count-honesty.md`, `@object-ui/plugin-kanban: patch`. Owed and checked, not assumed: `check-changeset-presence.mjs` reports *"3 source file(s) of 1 released package(s) changed"* — the package publishes `dist` and the change moves rendered output.

## Patch round — objectui#7322's spelling pin, kept rather than moved

CI went red on one test, `packages/types/src/__tests__/object-kanban-group-by-limit-7322.test.ts:221` — ours, not a flake. Reproduced locally on `e33d6ff` before touching anything: **1 failed / 28 passed**, identical assertion.

```
× `limit` is still read, as the exact text the docblocks cite
AssertionError: packages/plugin-kanban/src/ObjectKanban.tsx no longer reads `schema.limit`
  as `$top: schema.limit ?? DEFAULT_KANBAN_LIMIT`
```

objectui#7322 pins that literal source text in `READ_TEXT.limit`. The first cut of the saturation reading hoisted the cap into `const rowWindow = schema.limit ?? DEFAULT_KANBAN_LIMIT` and passed `$top: rowWindow`: behaviour identical, pinned spelling gone.

**Siblings, measured not assumed.** `git grep -F` on the literal finds **nine** occurrences. Exactly **one** is executable — the `READ_TEXT.limit` row above. The other eight are prose: docblocks in `packages/types/src/objectql.ts:2861`, `packages/types/src/zod/objectql.zod.ts:1185`, `packages/plugin-kanban/src/index.tsx:379`, `packages/app-shell/.../previews/block-config.ts:237`, a ledger `why` string in `block-config-schema-parity-8216.test.ts:188`, and two changesets. Separately, `object-kanban-record-source-7780.test.ts` pins **three more** literals out of this same file (`const boundData = useDataScope(schema.bind);`, the fetch guard, and the `const rawData = …` line). All three are untouched by this PR and are green — they are also why this file is treated as text-pinned by house style.

**Resolution: constraint 2, and it costs nothing.** The query is now a named object with the cap spelled inline exactly as pinned, and the saturation comparison reads `query.$top` back off it:

```ts
const query = {
    $filter: schema.filter,
    $top: schema.limit ?? DEFAULT_KANBAN_LIMIT,
    ...(expand.length > 0 ? { $expand: expand } : {}),
};
const results = await dataSource.find(schema.objectName, query);
...
setFetchWindowSaturated(data.length >= query.$top);
```

This is **one** spelling, not two — no expression is duplicated to appease a grep. It is also strictly better than the `rowWindow` local it replaces: a second copy of the cap kept only for the comparison could drift from the value actually on the wire, and a windowed-ness marker computed against a window the server was never asked for is precisely the silent wrongness this card exists to remove. The structure now makes that drift unrepresentable.

**No pin weakened, skipped, quarantined or edited.** `packages/types` is untouched by this PR; no file surface was extended. The assertion still fails if the renderer ever stops reading `schema.limit` or stops defaulting to `DEFAULT_KANBAN_LIMIT` — verified by the same ablation discipline that produced it: the row is red on `e33d6ff` and green on the head below.

**Green, on the pushed head:**

| run | result |
| --- | --- |
| `object-kanban-group-by-limit-7322.test.ts` (the failing file) | **29 passed**, incl. `✓ limit is still read, as the exact text the docblocks cite` |
| `+ object-kanban-record-source-7780.test.ts` + `packages/plugin-kanban/` | **38 files / 283 tests passed**, exit 0 |
| `pnpm --filter @object-ui/plugin-kanban type-check` | 0 — `query.$top` types clean |
| `pnpm --filter @object-ui/plugin-kanban lint` | 0 — same 176 pre-existing warnings, none new |

`origin/main` merged before pushing (`4d78b42`); it brought `#8737` and `#8771` and touched **no** file under `packages/types`, so neither #8776 nor #8777 has landed yet and there was no conflict. The ablation above is still anchored to the exact bytes it was run on: `KanbanImpl.tsx` is untouched by this patch round and its blob is still `cbc94487407f46c8b5649806b91596b29645a0c0`.

⛔ Unchanged by this round, as instructed: the `77+` design, the `>=` saturation predicate, the three header sites and the `BOUNDARY` row.

## 验收备注

- `noted, not filed:` the flat column header renders a bare English `Full` badge and the swimlane lane toggle a bare `▶`, neither behind `useKanbanT`. Pre-existing, untouched here, and cosmetic-copy rather than a wrong number. Carrier: whoever next opens `KanbanImpl.tsx`'s header block.
- `noted, not filed:` no tooltip explains the `+` to a first-time reader. Adding one means a new key across ten locale packs, which is outside this card's declared file surface; the marker itself is self-describing and never false, so the gap is legibility, not honesty.
- Option 1 (the server-side group-count aggregate) remains the complete fix and is untouched. This change is a strict improvement even if it later lands, exactly as triage said.

https://claude.ai/code/session_01611D6ZaRaMmwTNQmSbk8MH

---

### DIFF
--- a/.changeset/8307-kanban-lane-count-honesty.md
+++ b/.changeset/8307-kanban-lane-count-honesty.md
@@ -1,0 +1,44 @@
+---
+'@object-ui/plugin-kanban': patch
+---
+
+A Kanban lane header over a windowed fetch now says `77+` instead of `77`
+(objectui#8307).
+
+**The defect.** `object-kanban` fetches with a real `$top` (objectui#4025) and then
+groups **what came back** into lanes client-side, so `col.cards.length` counts fetched
+rows that fell into this lane — not the size of the group. Over any object holding more
+rows than the window every lane number is wrong and they sum to the window. Measured in
+a consuming app on 200 records: the board displayed **77 · 19 · 2** against a true
+**88 · 46 · 28 · 14 · 9 · 15**.
+
+The expensive part is that those numbers look right. 77 · 19 · 2 is a plausible funnel;
+there was no ellipsis, no "100 of 200", no styling difference — nothing prompting the
+reader to distrust a number that then travels into a status report.
+
+**The fix is not a better number, it is an honest one.** This component does not have a
+group total and cannot compute one: that needs a server-side group-count aggregate over
+the whole filtered set, which is a separate change. What it can stop doing is claiming a
+total it never fetched. When the fetch comes back **saturated** — at least as many rows
+as the window allowed — every lane header on the board renders `77+`, which is precisely
+what a count over a window establishes: *at least* 77. A fetch that came back short of
+its window is the one case where the client knows the result set was exhausted, and
+there the bare number is the truth and is kept. Rows handed to the board whole (inline,
+bound or external `data`) never passed through this window and keep the bare number too.
+
+`DEFAULT_KANBAN_LIMIT` is unchanged, deliberately: raising it moves the threshold
+without touching the property, and every board above the new number would fail exactly
+as silently.
+
+**The boundary case is resolved towards the true statement.** A board holding exactly
+the window is indistinguishable from a truncated one from the client side — both answer
+a `$top: N` query with N rows — so such a board renders `77+` for a lane that really
+holds 77. "At least 77" is TRUE there, whereas the bare "77" is FALSE on every truncated
+board. The marker is conservative; it is never wrong.
+
+All three lane headers move together — the flat column header and the swimlane layout's
+column-title row and lane rows — through one `laneCountLabel` helper, so a board cannot
+be honest in one layout and silently wrong in the other. Pinned by
+`laneCountHonesty-8307.test.tsx`, which reads the rendered header text (a threaded flag
+dropped one render short of the DOM would ship the identical silent board) and carries
+the exactly-at-the-window row and both negative controls.
--- a/packages/plugin-kanban/src/KanbanImpl.tsx
+++ b/packages/plugin-kanban/src/KanbanImpl.tsx
@@ -75,6 +75,13 @@ export interface KanbanBoardProps {
   objectFields?: unknown
   /** Field name for swimlane rows (2D grouping) */
   swimlaneField?: string
+  /**
+   * The cards in `columns` are a fetched WINDOW, not the whole group
+   * (objectui#8307) — see `laneCountLabel`. Injected by `ObjectKanban`, the
+   * only entry point that issues the windowed query; a board handed its rows
+   * whole leaves it unset and keeps the bare number.
+   */
+  countsAreWindowed?: boolean
 }
 
 /**
@@ -270,6 +277,43 @@ function columnWidthClasses(columnStyle?: React.CSSProperties): string {
     : "w-[85vw] sm:w-80 shrink-0";
 }
 
+/**
+ * How a lane count is WRITTEN, in ONE place, for every header on this board.
+ *
+ * The number handed in is `col.cards.length` — the count of fetched rows that
+ * fell into this lane. The fetch is windowed at a real `$top` (objectui#4025),
+ * and the board groups what came back, so over any object with more rows than
+ * the window that number is not the size of the group: every lane is short,
+ * the lanes sum to the window, and the result is CREDIBLE — 77 / 19 / 2 reads
+ * as a plausible funnel against a true 88 / 46 / 28 / 14 / 9 / 15, so nothing
+ * prompts the reader to distrust it (objectui#8307).
+ *
+ * The fix is not a better number — this component does not have one, and
+ * getting one means a server-side group-count aggregate over the whole
+ * filtered set. The fix is to stop the number claiming to be something it is
+ * not: `77+` says "at least 77", which is what a count over a window actually
+ * establishes.
+ *
+ * ⚠️ The boundary case is deliberate. `countsAreWindowed` comes from
+ * "the fetch came back with at least as many rows as it asked for", and a
+ * board holding EXACTLY the window is indistinguishable from a truncated one
+ * from this side of the wire — no client-side signal separates them without
+ * asking the server a second question. So such a board renders `77+` for a
+ * lane that really does hold 77. That is the safe half of the ambiguity:
+ * "at least 77" is TRUE when the lane holds exactly 77, whereas the bare
+ * "77" is FALSE whenever the board is in fact truncated. The marker is
+ * conservative; it is never wrong.
+ *
+ * A shared function rather than a template literal repeated per header,
+ * because this board paints lane counts from three places (the flat column
+ * header, and the swimlane layout's column-title row and lane rows) and a
+ * count that is honest on one layout and bare on another is the same defect
+ * with a smaller blast radius. Adding a header means calling this.
+ */
+function laneCountLabel(count: number, countsAreWindowed?: boolean): string {
+  return countsAreWindowed ? `${count}+` : String(count)
+}
+
 function KanbanColumnView({
   column,
   cards,
@@ -280,6 +324,7 @@ function KanbanColumnView({
   objectFields,
   columnStyle,
   suppressEmptyPlaceholder,
+  countsAreWindowed,
 }: {
   column: KanbanColumn
   cards: KanbanCard[]
@@ -298,6 +343,8 @@ function KanbanColumnView({
    * doesn't read as N redundant copies of the same message.
    */
   suppressEmptyPlaceholder?: boolean
+  /** The cards handed in are a fetched window — see `laneCountLabel` (objectui#8307). */
+  countsAreWindowed?: boolean
 }) {
   const { t } = useKanbanT()
   const safeCards = cards || [];
@@ -349,7 +396,7 @@ function KanbanColumnView({
                   : "bg-muted/70 text-muted-foreground",
               )}
             >
-              {safeCards.length}
+              {laneCountLabel(safeCards.length, countsAreWindowed)}
               {column.limit && <span className="text-muted-foreground/70 font-normal">{` / ${column.limit}`}</span>}
             </span>
             {isLimitExceeded && (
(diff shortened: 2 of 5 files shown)


os-zhuang:
## CI red — diagnosed, this PR's, patch round dispatched

`domain:ui` @ objectui seat (`session_01611D6ZaRaMmwTNQmSbk8MH`), 2026-09-09T06:3xZ.

**`Test (shard 4/4)` — 1 failed / 8746 passed.** Read from the job log, ⛔ not inferred:

```
packages/types/src/__tests__/object-kanban-group-by-limit-7322.test.ts:221
AssertionError: packages/plugin-kanban/src/ObjectKanban.tsx no longer reads `schema.limit`
  as `$top: schema.limit ?? DEFAULT_KANBAN_LIMIT`
```

### It is ours, and it is not a flake

objectui#7322 pinned that **literal source text**. This PR extracts `const rowWindow = schema.limit ?? DEFAULT_KANBAN_LIMIT` so the saturation predicate can compare against the window, and passes `$top: rowWindow`. ⇒ the **behaviour is unchanged** — the limit is still read, the default still applies — but the spelling the pin greps for is gone.

⛔ Not re-run, and ⛔ not called a flake: the diff touches exactly the pinned line, so the single failing assertion is caused by this change. One test, named, reproducible.

### What the fix may and may not be

The pin's intent must survive: *`ObjectKanban` still reads `schema.limit` and still defaults to `DEFAULT_KANBAN_LIMIT`*, and it must still fail if a future change really stops doing so. ⛔ Weakening, skipping, quarantining or deleting it to get green is refused outright.

Preference order given to the dev: keep the pinned spelling inline if that costs nothing real; ⛔ **not** two spellings of one expression merely to appease a grep; otherwise update the pin to the new spelling with the reason stated. That file lives in `packages/types`, outside this claim's `packages/plugin-kanban/**` surface — **the extension is authorised by this seat for that one file**, on the ground that the pin is about the file this PR changes.

⚠️ `packages/types` is contended (PRs #8776, #8777 from the `domain:spec` seat) — `origin/main` gets merged before the push.

### Everything else stands

All other 32 checks are green, and the seat review on objectui#8307 is already **ACCEPT**: the `77+` marker, the `>=` saturation predicate, the three header sites and the BOUNDARY row are unchanged by this. ⛔ This is a one-test patch round, not a re-scope.

---

PR #8930, opened 2026-09-10
fix(types): refuse `breadcrumbs` by name on the `page` node (objectui#8871)

os-warren:
Fixes #8871

## Step 1 was a measurement, and its verdict is COVERED

Triage did not rule this retirement — it ruled that a measurement decides whether a ruling is needed at all: *does ADR-0049's enforce-or-remove cover this repository's `packages/types` zod **mirror** face?* It does, and the evidence is on this exact node.

| Evidence | Reading |
|---|---|
| `packages/types/src/zod/tombstone.zod.ts` | Exports `retirementTombstone()`, whose own docstring opens *"Declare an **ADR-0049 RETIREMENT TOMBSTONE**"*, and which is *"Internal to this package's zod modules"* — machinery built **for** this face and no other |
| `git grep -l "ADR-0049" -- .changeset/` | **63** changesets, exit 0. The majority bump `'@object-ui/types'` |
| `PageNodeSchema` itself | Already carries `actions: retirementTombstone(PAGE_ACTIONS_REFUSAL)` — landed by PR #8870 as *"an ADR-0049 refusal arm"*. **Same file, same node, same `.passthrough()` survival mechanism, one key over** |
| `docs/audits/2026-09-plugin-detail-downstream-consumer-census.md:22` | This repo's own audit doctrine: *"Retiring a published component is **ADR-0049 / ADR-0087 (objectstack)** and narrows public surface"* |

⇒ Not a fork requiring a ruling. A retirement executed under standing discipline.

<details><summary><b>Counter-evidence, weighed and reported rather than suppressed</b> — it does not flip the verdict, but a reviewer should see it</summary>

ADR-0049's own **Decision** text is scoped narrowly: *"A spec property that names a **security/access-control boundary** must be in exactly one of three states."* Its **Non-goals** place *"the P2 (**spec hygiene**) clusters of #1878 — non-security, governed separately"* outside. And its 2026-09-04 amendment's Scope note says *"**No new scope is claimed for this ADR by this amendment**."*

In the same breath, that amendment records as fact that *"**the repo cites this ADR as the enforce-or-remove policy for spec-property retirement generally**"* — and does not disown it. objectui adopted it exactly that way: none of `MobileOverrides`, grid `wrap`, `ComponentInput.inputType` or `page.actions` is a security property, and all four cite this ADR.

Triage's question was whether the **mirror face** is in range. It is, demonstrably, by landed precedent on this very node. The property-class nuance is recorded here so it is a known reading rather than a silent assumption.
[...]

PR #8872, opened 2026-09-09
test(console): pin object-calendar member shapes (objectui#8071 slice 5)

os-warren:
Refs objectstack-ai/objectui#8071

## What

Converts both remaining `object-calendar` `MEMBER_PIN_EXEMPTIONS` entries (`calendar`, `dataSource`) into real per-block member pins and lowers `MEMBER_PIN_EXEMPTION_CEILING` 48 -> 46 in the same commit. This closes `object-calendar` outright — every other key it declares (`data`, `filter`, `sort`, `staticData`) was already pinned by earlier slices — the same shape `element:record_picker` (slice 3) and `record:quick_actions` (slice 4) closed in.

| | before | after |
|---|---|---|
| `MEMBER_PINS` | 42 | **44** |
| `MEMBER_PIN_EXEMPTIONS` | 48 | **46** |
| `MEMBER_PIN_EXEMPTION_CEILING` | 48 | **46** |
| population | 90 | 90 |

Census by **TypeScript AST walk** (`ts.createSourceFile`, visiting `VariableDeclaration`/`PropertyAssignment` — not brace-depth counting), taken at BASE (`42ddd7f71`) and on the current tree; both match the dispatched ledger exactly.

## Why this batch

`object-calendar`'s only two remaining exemptions were both left over from objectui#8176 (`AWAITING_A_PIN_NEWLY_JUDGED`), on a block where the other four keys were already pinned — a two-key remainder on an already-mostly-pinned block, closeable outright rather than skimmed for the easy key and left with a straggler. `NEWLY_JUDGED_UNPINNED_MEMBERS` shrinks from 4 to 2 (only `object-kanban`'s pair remain) — that array is checked against the live `MEMBER_PIN_EXEMPTIONS` state by the gate's own `the ceiling correction admits exactly the four keys objectui#8176 made visible` test, so it has to shrink in the same commit, not just the docblock prose.

- `dataSource` — the shared `elementDataSourceBlock` per-element binding (`ELEMENT_DATA_SOURCE_INPUT`, `@object-ui/core`), the same mechanism `element:record_picker.dataSource` and `record:related_list.dataSource` already pin. Promotes a **pre-existing file**, `ObjectCalendar.elementDataSource.test.tsx`, read end to end before being credited.
[...]

PR #8865, opened 2026-09-09
feat(types,plugin-kanban,plugin-gantt)!: retire the bare `kanban` / `gantt` node keys and the `kanban-ui` / `kanban-enhanced` registrations

os-warren:
Fixes #8802
Fixes #8257
Fixes #8008

⚠️ Those three lines are spelled as bare `Fixes #N` on purpose. The cross-repo
form (`Fixes owner/repo#N`) is honoured by GitHub but NOT by this seat's landing
pre-check, whose `closingKeywordRe` accepts only the bare form — on the previous
head `check-clause2-carriers.mjs --pair 8865` therefore exited **2 =
UNJUDGED**, no pair formed, and the gate could not answer at all. ⛔ Do not
"normalise" them back.

Executes three maintainer rulings taken in one batch on 2026-09-09. ⛔ Nothing about the direction is re-opened here.

`Clause-②: yes` — this narrows published accept sets. An in-seat contract-review-tier PASS is the PM's to run, and landing is the PM's and the maintainer's call.

⛔ **CORRECTION, 2026-09-10 — this paragraph previously said the PR "stays **draft**" and that "`needs:contract-review` remains hung". Both are now FALSE and are corrected rather than reversed.** At `05:05:42Z` a maintainer flipped this PR **out of draft** and **removed** the `needs:contract-review` label. Neither act was this seat's; ⛔ neither is reversed here — the label is not re-added and the draft state is not flipped back. What the removal did **not** change is the declaration itself: `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 8865` exits **0** on the current head, agreeing on all three cards (#8802, #8257, #8008), with `--self-test` → **0** (299 cases) as its control. That gate reads the `Clause-②` declaration in both carriers, ⛔ not the label, which is why removing the label did not break the pair.

## 🔁 Merge-repair round, 2026-09-10 — current `main` merged in, and the semantic collision that exposed

A maintainer's merge attempt was rejected. ⛔ **Not a textual conflict** — `git merge-tree --write-tree` exits 0 and a real `git merge --no-ff origin/main` auto-merges with **0 unmerged paths**. The failure is **semantic and appears only on the MERGED tree**, which is what the queue builds and what a `mergeable_state` computed against a stale base cannot see.

[...]

PR #8933, opened 2026-09-10
test(types): pin NamedListView's per-member liveness census (objectui#7924)

os-warren:
Refs #7924 — this PR lands the **census only** and decides nothing. The two dispositions the card names for the unread members (`?: never` tombstones per the objectui#7129 route, or making the delegation read them from the named view) are maintainer rulings and stay with objectui#7928. ⛔ **No declaration face moved**: `packages/types/src/objectql.ts`, every `*.zod.ts` and every renderer read set are byte-identical to `main` in this diff, which touches one test file and one changeset.

## Why none of the card's figures are quoted

The card body says "about 52 members" / "exactly seven read" / "~45", measured on `6a9ee323`. Every figure below is **re-derived on this branch's head** (`3b5053d45`). The reading is **47 declared, 6 declared-and-read, 41 unread**, plus one read-but-undeclared key. "About 52" reproduces only as a hand figure sitting between two instruments and is neither of them — the instrument table below pins that gap as an assertion rather than a sentence.

## The census

`NamedListView` (`packages/types/src/objectql.ts:2025`–`:2202` today — re-located by AST, not by the card's stale line numbers) declares 47 top-level members. The `object-view` node renderer reads 7 names off a named view; 6 of them are declared members.

⚠️ The third column is the distinction the entire finding turns on: a member reached through `activeView?.KEY` — the host's `views` prop — is **not** read off the named view. It is included because it is what the ruling needs: it separates "unread here" from "unread anywhere".

| # | member | optional | read off the named view | reached from `activeView` |
| ---: | --- | --- | --- | --- |
| 1 | `label` | **no** | **read** | `activeView?.label` |
| 2 | `type` | yes | **read** | `activeView?.type` |
| 3 | `columns` | yes | **read** | `activeView?.columns` |
| 4 | `filter` | yes | **read** | `activeView?.filter` |
| 5 | `sort` | yes | **read** | `activeView?.sort` |
| 6 | `options` | yes | **read** | none |
| 7 | `showSearch` | yes | unread | `activeView?.showSearch` |
| 8 | `showSort` | yes | unread | `activeView?.showSort` |
| 9 | `showFilters` | yes | unread | `activeView?.showFilters` |
| 10 | `showHideFields` | yes | unread | fold only |
| 11 | `showGroup` | yes | unread | fold only |
| 12 | `showColor` | yes | unread | fold only |
| 13 | `showDensity` | yes | unread | fold only |
| 14 | `compactToolbar` | yes | unread | `activeView?.compactToolbar` |
| 15 | `allowExport` | yes | unread | `activeView?.allowExport` |
| 16 | `color` | yes | unread | `activeView?.color` |
| 17 | `inlineEdit` | yes | unread | `activeView?.inlineEdit` |
| 18 | `wrapHeaders` | yes | unread | `activeView?.wrapHeaders` |
| 19 | `clickIntoRecordDetails` | yes | unread | `activeView?.clickIntoRecordDetails` |
| 20 | `addRecordViaForm` | yes | unread | `activeView?.addRecordViaForm` |
| 21 | `addDeleteRecordsInline` | yes | unread | `activeView?.addDeleteRecordsInline` |
| 22 | `collapseAllByDefault` | yes | unread | `activeView?.collapseAllByDefault` |
| 23 | `fieldTextColor` | yes | unread | `activeView?.fieldTextColor` |
| 24 | `prefixField` | yes | unread | `activeView?.prefixField` |
| 25 | `description` | yes | unread | none |
| 26 | `showDescription` | yes | unread | `activeView?.showDescription` |
| 27 | `navigation` | yes | unread | `activeView?.navigation` |
| 28 | `selection` | yes | unread | `activeView?.selection` |
| 29 | `pagination` | yes | unread | `activeView?.pagination` |
| 30 | `searchableFields` | yes | unread | `activeView?.searchableFields` |
| 31 | `filterableFields` | yes | unread | `activeView?.filterableFields` |
| 32 | `resizable` | yes | unread | `activeView?.resizable` |
| 33 | `densityMode` | yes | unread | `activeView?.densityMode` |
| 34 | `rowHeight` | yes | unread | `activeView?.rowHeight` |
| 35 | `hiddenFields` | yes | unread | `activeView?.hiddenFields` |
| 36 | `exportOptions` | yes | unread | none |
| 37 | `rowActions` | yes | unread | `activeView?.rowActions` |
| 38 | `bulkActions` | yes | unread | `activeView?.bulkActions` |
| 39 | `bulkActionDefs` | yes | unread | none |
| 40 | `sharing` | yes | unread | `activeView?.sharing` |
| 41 | `addRecord` | yes | unread | `activeView?.addRecord` |
| 42 | `conditionalFormatting` | yes | unread | `activeView?.conditionalFormatting` |
| 43 | `userFilters` | yes | unread | `activeView?.userFilters` |
| 44 | `showRecordCount` | yes | unread | `activeView?.showRecordCount` |
| 45 | `allowPrinting` | yes | unread | `activeView?.allowPrinting` |
| 46 | `emptyState` | yes | unread | `activeView?.emptyState` |
| 47 | `aria` | yes | unread | `activeView?.aria` |
| — | `data` | **not declared** | **read**, through an `as any` cast | `activeView?.data` |

**Totals: 47 declared = 6 read + 41 unread.** Read set off the named view = 7 names (the 6 above plus the undeclared `data`).

"fold only" means the member has no `activeView?.KEY` access, but IS consumed when it appears on `activeView`, through `normalizeListViewSchema(activeView)` (`packages/core/src/utils/normalize-list-view.ts`, its `SHOW_FLAG_TO_USER_ACTION` map). ⚠️ Worth stating because a named-property probe alone reports those four as having no reader at all, which is false — and that mistake would have gone straight into the ruling.

### Three readings the ruling request should carry

1. **`data` is still read-but-undeclared.** It reaches the renderer as `data: (currentNamedViewConfig as any)?.data ?? …`. objectui#7928 requires it to be *declared or its cast removed* and names this card as its home; it is still neither. It is why the arithmetic subtracts **six**, not seven, from 47.
2. **The 41 unread are not one population.** 34 have a host-side twin the renderer honours off `activeView`; 4 more are folded from `activeView`; **3 have no reader on any path** — `description`, `exportOptions`, `bulkActionDefs`. Of those, `exportOptions` and `bulkActionDefs` do not appear in the renderer at all (pinned), and `description` appears only off `schema` / `schema.table` / `schema.form`, never off a view.
3. **`label` is the only required member**, and it is read — so a tombstone route would leave the sole required member intact.

## Instruments, and the control that makes each zero a reading

| instrument | what it walks | reading |
| --- | --- | --- |
| `namedListViewMembers()` **(new)** | `ts.createSourceFile`, the interface's own `PropertySignature` members | **47** |
| `namedListViewMemberCount()` (kept) | strict regex, two-space indent anchor | **47** |
| `namedListViewLooseMemberCount()` **(new)** | same regex without the indent anchor — also matches nested object-literal lines | **59** |

⛔ Not a brace-depth count: a brace parser on this repo has already failed to terminate at the end of an object literal (objectui#8071). The three are now pinned **against each other**, so the instrument swap is itself a measurement and "52 is between two instruments and is neither" stops being prose.

The read set is derived by walking **every route** from `schema.listViews` to a named view — a binding annotated `NamedListView`; the value binding of `Object.entries(record).map(...)` (through an intermediate `const entries = …` too); and element access straight off the record — stripping `( )`, `!` and `as any` at each access, which is the only reason `data` is visible at all.

⭐ That walk found a named-view read the old `currentNamedViewConfig?.KEY` regex **cannot see**: `{view.label || key}` on the named-view tab strip. `label` is read at two sites, and the count is asserted.

**Completeness**, not just coverage: every occurrence of the record binding is classified into a role (`DECLARATION`, `ELEMENT_ACCESS`, `OBJECT_KEYS`, `OBJECT_ENTRIES`, `COMPARISON`, `CONDITION`, `HOOK_DEPENDENCY`). A new syntactic route lands as `UNCLASSIFIED` and fails **before** the census can silently under-count — the test says in its own message not to add the role to make it green.

**Firing controls** (a silent probe is a parser that found nothing, not a reading of zero):

- the query returning nothing for all 41 unread members is the query that returns `columns` — asserted present in both the declared set and the read set;
- its negative twin `rowHeight` is asserted **declared and not read** off the named view, with `rowHeight: activeView?.rowHeight,` pinned as the text that makes it so;
- `stickyHeader`, in neither set, is non-vacuity for both probes at once;
- the two "absent from the renderer entirely" claims carry their own control: the same `grep` that finds nothing for them is asserted to find `rowHeight`.

## Ablation — the pin is non-vacuous in **both** directions

Each leg: mutation proven on disk **before any result was read** (anchor counts on both the removed and the injected text, plus a `git hash-object` differing from the HEAD blob), restore from a `trap … EXIT INT TERM` using `git checkout HEAD -- ABSOLUTE_PATH`, restoration proven **by blob hash**, never by an exit code. HEAD blob of `ObjectView.tsx` = `c1a807af683a0147796780c801ac34ffa97e8c70`.

**Leg A — unread becomes read.** `navigation: activeView?.navigation ?? …` rewritten to read the named view first. Removed-text count 1 to 0, injected 0 to 1, blob `c1a807af` to `fdfe41ac`. Result: **4 failed / 143 passed**, and the only per-member row that moved is the expected one:

- `UNREAD — navigation is declared and NOT read off a named view` ⇒ red
- the two "exactly seven names" assertions and the `6 read + 41 unread = 47` partition ⇒ red
- the other **40** unread rows and all **6** read rows ⇒ green

**Leg B — read becomes unread.** The `currentNamedViewConfig?.type` read deleted. Blob `c1a807af` to `c9e79701`. Result: **4 failed / 143 passed**:

- `READ — type is declared AND read off a named view` ⇒ red
- the same three aggregate assertions ⇒ red
- every other per-member row ⇒ green

⚠️ **Leg B's first attempt did not run, and is reported rather than quietly retried.** Its injected text (`if (activeView?.type) return activeView.type;`) already existed on the next line, so the on-disk proof read injected-count **2, expected 1**, and the script aborted with `exit 94` **before** running anything and restored the file. The reading was void, not green. Re-anchored on a unique marker and re-run; both blob hashes above are from the run that actually happened.

Restoration verified after every leg: `git hash-object` back to `c1a807af683a0147796780c801ac34ffa97e8c70` and `git diff HEAD` empty for that path.

## Readings — commands and exit codes

Exit codes captured into a file before any pipe (`cmd > out 2>&1; RC=$?`), never off the end of a pipeline.

| command | exit | reading |
| --- | ---: | --- |
| `pnpm exec vitest run packages/types/src/__tests__/object-view-unmirrored-keys-7779.test.ts` | 0 | `Test Files 1 passed (1)` · `Tests 147 passed (147)` |
| `pnpm --filter @object-ui/types type-check` | 0 | all three of `tsc --noEmit`, `tsconfig.examples.json`, `tsconfig.test.json` echoed and green — the test file **is** typechecked |
| `pnpm exec eslint packages/types/src/__tests__/object-view-unmirrored-keys-7779.test.ts` | 0 | clean |
| `node scripts/check-changeset-presence.mjs` | 0 | 1 source file of 1 released package changed, 1 changeset declared, empty frontmatter = the explicit exemption |
| `node scripts/check-changeset-no-major.mjs` | 0 | no changeset declares `major` |
| `pnpm check:control-bytes` | 0 | 7111 tracked text files scanned |
| `pnpm check:phantom-deps` | 0 | `typescript` is already a declared devDependency of `@object-ui/types` |
| `node scripts/check-governed-queue-guard.mjs --test` (both changed paths) | 0 | `NOT GOVERNED` — 2 paths, no governed surface matched |

All test runs went through the container's shared heavy-verify lock; each printed `VERDICT command-exit 0`.

**Declared narrowing.** Repo-wide `pnpm lint` and `pnpm test` are CI's runs, not this seat's. The lint reading above is narrowed to the one changed file; `eslint.config.js` configures **no** type-aware linting (no `project` / `projectService`), so this diff cannot move the verdict on any file it does not touch. Test scope is the single pin file because the diff changes exactly that file plus a changeset, and no published byte moves.

## Scope

⛔ The three live branches' files were not touched: `packages/types/src/complex.ts` / `plugin-kanban` / `plugin-gantt` (PR #8865), `packages/types/src/layout.ts` / `page-breadcrumbs-refusal-8871.test.ts` (PR #8930), `registry-inputs-spec-parity.test.ts` / `plugin-calendar` (PR #8872).

The census extends the existing pin rather than adding a file. That file already owned **both** halves of this measurement (the 47 count and the seven-name read set) and is where the regex instrument being improved on lives; a second file would create a second instrument for the same number, which is the staleness shape objectui#7947 was filed about. The header's prose figures are now backed by the assertions underneath them.

⛔ This PR stays **draft** — the PM lands it. The card is **not** closed by it: the disposition is still unruled.

---

---

### DIFF
--- a/.changeset/7924-named-list-view-liveness-census.md
+++ b/.changeset/7924-named-list-view-liveness-census.md
@@ -1,0 +1,15 @@
+---
+---
+
+Measure `NamedListView`'s per-member liveness and pin it (objectui#7924).
+
+The declared set is now walked with the TypeScript parser rather than a regex,
+the named-view read set is derived through every route from `schema.listViews`
+to a named view (with every occurrence of that record classified, so a new
+route fails instead of shrinking the reading), and each of the 47 declared
+members is pinned by name on its side of the partition — 6 read, 41 unread —
+so a member moving between the two sets fails in either direction.
+
+⛔ Measurement only: neither declaration face moved, and the disposition for the
+41 unread members (`?: never` tombstones vs. making the renderer read them)
+stays with objectui#7928. Test only; no package is released by this change.
--- a/packages/types/src/__tests__/object-view-unmirrored-keys-7779.test.ts
+++ b/packages/types/src/__tests__/object-view-unmirrored-keys-7779.test.ts
@@ -57,11 +57,41 @@
  * are asserted against the SPEC schema here, so the day the spec relaxes (or
  * the renderer's read set moves) the measurement — and the stop — is re-taken
  * rather than remembered.
+ *
+ * ## objectui#7924 — the same measurement, now PER MEMBER
+ *
+ * The two figures above were a COUNT and a name list. objectui#7924 asks which
+ * members, by name, are on each side, because that is the input the `listViews`
+ * value-type ruling (objectui#7928) needs: tombstoning the unread ones narrows a
+ * published accept set, and making the renderer read them is capability growth.
+ * ⛔ Neither is dispatched or decided here — this file only measures.
+ *
+ * The census lives in the last describe block and changes the INSTRUMENT as well
+ * as the resolution:
+ *
+ * - the declared set is walked with the TypeScript parser (`ts.createSourceFile`,
+ *   the interface's own `PropertySignature` members), ⛔ not a regex and ⛔ not a
+ *   brace-depth count — a brace parser on this repo has already failed to
+ *   terminate at an object's close (objectui#8071). The two regexes are kept and
+ *   asserted AGAINST the parser, so the instrument swap is itself a reading:
+ *   strict regex 47 = parser 47, loose regex 59.
+ * - the read set is derived by walking every route from `schema.listViews` to a
+ *   named view, and every occurrence of that record is CLASSIFIED — so a new
+ *   route fails as `UNCLASSIFIED` instead of silently shrinking the read set.
+ *   That found a named-view read the old `currentNamedViewConfig?.KEY` regex
+ *   cannot see: `{view.label || key}` on the tab strip.
+ * - each of the 47 members is then pinned BY NAME on its side of the partition,
+ *   so a member moving between read and unread fails in EITHER direction.
+ *
+ * ⚠️ The distinction the whole finding turns on: a member reached through
+ * `activeView?.KEY` — the host's `views` prop — is NOT read off the named view.
+ * `rowHeight: activeView?.rowHeight` is the pinned counter-control below.
  */
 import { describe, it, expect } from 'vitest';
 import { readFileSync } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
+import ts from 'typescript';
 import {
   ListViewSchema as SpecListViewSchema,
   ObjectListViewSchema as SpecObjectListViewSchema,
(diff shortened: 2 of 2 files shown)


os-warren:
## Green on `3b5053d45` — flipped ready and enqueued. ⚠️ The enqueue is **NOT VERIFIED**; the merge is what will settle it.

`domain:spec` @ objectui PM seat, session `session_01Jmxdo7bmeqCQHLSfmLVX9w`.

**Eligibility, measured.** `get_check_runs` on head `3b5053d45f85dd5d2f5767afe38800b676048627`: **34 runs — 31 `success`, 3 `skipped`, 0 `failure`, 0 `in_progress`.** The three skips are by design (`Test (coverage)`, `Test (coverage shard …)`, `dependabot`). ⛔ 入队资格 is every check green, ⛔ not a required subset.

**⚠️ What I can and cannot prove.** The flip out of draft was read back. The enqueue was **not** verified: `enable_pr_auto_merge` returns *"(method: , enabled at )"* with both fields empty, and this session's MCP server exposes no issue-timeline / PR-events tool, so no reading available to me distinguishes "enqueued" from "no-op". Reported as **NOT VERIFIED** rather than dressed up. A check-in will verify by **outcome**, and by content rather than an API field.

### Why this PR is `Clause-②: no` — stated, since the card it serves is a narrowing question

⛔ **No declaration face moved.** The diff is exactly two files: this pin file and its changeset. `packages/types/src/objectql.ts` is **read**, never edited; no `*.zod.ts` is touched; no renderer's read set moves. The disposition for the 41 unread members — `?: never` tombstones (a narrowing) or making the delegation read them (capability growth) — is ⛔ **not** taken here and stays with objectui#7928, where the four-group ruling request is now posted (`5612300243`).

### What is worth keeping from this round

⭐ **The instrument was replaced, and the replacement is itself an assertion.** The strict indent-anchored regex (**47**) and the TypeScript parser (**47**) are pinned as agreeing, and the loose regex (**59**) is pinned as disagreeing *for a stated reason* — it also counts nested object-literal lines inside members' inline types. ⇒ the retired "about 52" is pinned as a hand figure sitting between two instruments and equal to neither, rather than as a sentence someone has to remember.

⭐ **Completeness is enforced, not assumed.** Every occurrence of the named-views record is classified into a syntactic role; a **new** route to a named view lands as `UNCLASSIFIED` and **fails**, instead of silently shrinking the read set. That is what caught the read the old `currentNamedViewConfig?.KEY` regex could not see — `{view.label || key}` on the tab strip — so `label` is now asserted at **two** sites.

⭐ **The ablation runs in both directions**, which is the property a partition pin needs: make an unread member read (`navigation`) → exactly the `UNREAD — navigation` row red, the other 40 unread rows and all 6 read rows green; make a read member unread (`type`) → exactly `READ — type` red. Mutations proven on disk by blob hash **before** any result was read, restored from a `trap … EXIT INT TERM` and proven back by blob hash with an empty `git diff HEAD`.

⚠️ And **leg B's first attempt did not run, and was reported as void rather than retried into silence**: its injected text already existed on the following line, the on-disk proof read `injected = 2` against an expected 1, and the harness aborted **exit 94 before running anything** and restored. Re-anchored on a unique marker and re-run. ⛔ A reading taken after a mutation that did not land the way it was proven is not a green — it is nothing.

---

PR #8942, opened 2026-09-10
test(console,components): pin the last four one-key member blocks (objectui#8071 slice 7)

os-warren:
Refs #8071 — slice 7. The card stays open, and 37 exemptions remain.

## What this lands, in one commit

Every remaining `MEMBER_PIN_EXEMPTIONS` block whose remainder was exactly **one key** becomes a real per-block member pin. The four exemptions are deleted and `MEMBER_PIN_EXEMPTION_CEILING` moves 41 to 37 in the same commit, so the ceiling never holds an unused slot.

| key | pin file | new or promoted |
| --- | --- | --- |
| `element:button.action` | `packages/components/src/__tests__/elementButtonActionMembers-8071.test.tsx` | new |
| `element:number.filter` | `packages/components/src/renderers/basic/__tests__/elementNumberFilterMembers-8071.test.tsx` | new |
| `page:accordion.items` | `packages/components/src/__tests__/pageAccordionItemMembers-8071.test.tsx` | new |
| `page:tabs.items` | `packages/components/src/__tests__/pageTabsItemMembers-8071.test.tsx` | new |

All four blocks now carry **zero** exemptions. The batch is coherent on slice 6's terms: one package (`packages/components`) registers all four — the two `element:` keys share `renderers/basic/elements.tsx`, the two `page:` keys share `renderers/layout/containers.tsx` — and the selection rule is stateable in a sentence.

## AST ledger — measured, not inherited

TypeScript AST walk (`ts.createSourceFile`, `VariableDeclaration`, initializer unwrapped through as / satisfies / parens, `PropertyAssignment`), object boundaries from **AST node spans**. Re-derived on this branch's own head rather than taken from the dispatch.

```
                                 before (b97129e96)      after (4e4e6ca56)
MEMBER_PIN_EXEMPTIONS                    41                      37
MEMBER_PIN_EXEMPTION_CEILING             41  (:2747)             37  (:2818)
MEMBER_PINS                              49                      53
distinct block prefixes                  10                       6
NEWLY_JUDGED_UNPINNED_MEMBERS             2                       2   (untouched)
```

The delta balances on one instrument: **minus 4 exemptions, minus 4 ceiling, plus 4 pins**, with the parent holding exactly one key under each of the four prefixes taken. Parser guards clean: `parseDiagnostics` 0; exactly one `VariableDeclaration` per name, so no shadowed node was picked up; 0 non-`PropertyAssignment` members and 0 computed keys in either object; 0 non-string array elements.

### Instrument cross-check, reported rather than smoothed

A key-anchored regex **bounded to top-level indentation inside the AST span** AGREES on both objects: 37 = 37 and 53 = 53, with an empty only-in-AST and an empty only-in-regex set.

The **depth-blind** key anchor slice 6 caught still reads `MEMBER_PINS` as **159** against the AST's 53. Fully accounted for: values are objects of shape file-plus-pins, so 53 entries carry 106 nested sub-keys, and 53 + 106 = 159 matches to the unit. That is the same failure mode for the fourth time on this file, on new content — recorded, not smoothed.

## The two one-key blocks deliberately NOT taken

- **`record:related_list.actions`** is the `NO_READ_SITE_TO_PIN` sentinel, and the reading was **re-measured here rather than inherited**: on this tree `packages/plugin-detail/src/renderers/record-related-list.tsx` contains **zero** case-sensitive occurrences of `actions` (control on the same file, same run: `import` reads **9**), and the only near-matches are `useRelatedRecordActions` / `relatedActions` — the host bridge the constant already names. Nothing to pin, still. It is **not** "one easy key left".
- **`object-kanban.columns` / `object-kanban.dataSource`** are the `NEWLY_JUDGED_UNPINNED_MEMBERS` pair, held by the unruled contract question in objectui#8913 and by the live branch over `packages/plugin-kanban/**`. Pinning a shape a pending ruling may delete would be worse than leaving it exempt, so `NEWLY_JUDGED_UNPINNED_MEMBERS` is **unchanged** — no block it names was touched, and no file this PR edits is owned by that branch.

## Why four new files rather than promotions

Every candidate was **read end to end** before being rejected, not dismissed on its greps. None covered its key's member **set**: `element-button-action.test.tsx` drives only `type` and `to` on one navigation action; `action-bodyExtra-forward.test.tsx` pins one member across four renderers; `page-accordion-icon.test.tsx` pins `icon` alone; the five `page:tabs` suites pin one member each; and `element-number.contractEnvelope-6726.test.tsx` drives the same `find()` fallback branch but never authors a `filter`, so the locator itself would refuse it.

One had to be **refused** rather than merely passed over. `action-bodyShape-forward.test.tsx` NAMES `element:button` and the key `action`, so it would satisfy the locator — while what it actually says is that `element:button` is deliberately out of its scope. Crediting it would have been a pin asserting the opposite of its claim, which is the false-negative shape objectui#8068 exists to end. Its boundary is instead pinned as **behaviour** in the new file, which authors a non-whitelisted key and asserts it is dropped.

## Ablations — every mutation proven on disk BEFORE any result was read

Each run anchors grep counts on **both** the injected and the removed text, requires `git hash-object` to differ from the HEAD blob, restores with `git checkout HEAD -- ABSOLUTE_PATH` from a `trap ... EXIT INT TERM`, and proves restoration **by blob hash** — never by an exit code.

**Family 1 — the gate mechanism.** One deleted exemption re-added, ceiling untouched. Blob `c5497f50` to `c7e10387` and back to `c5497f50`.

Exactly three rows red, each for its own reason, 195 green:

```
× judges a non-vacuous member-shape census   expected [ 'element:button.action' ] to deeply equal []
× carries no stale member-pin exemption      expected [ 'element:button.action' ] to deeply equal []
× the member-pin exemption list only ratchets DOWN   expected 38 to be less than or equal to 37
```

**Family 2 — one per new pin, each failing for its own reason.** All four pin files run together each time, so "its own reason" is observable rather than asserted.

| pin ablated | mutation | rows red | other three pins |
| --- | --- | --- | --- |
| `element:button.action` | drop `bodyExtra` from the forward whitelist | 1 — the whitelist-SET row | green (23 passed) |
| `element:number.filter` | rewrite the `find()` wrapper key to the flat spelling | 1 — the wrapped-spelling row | green (23 passed) |
| `page:accordion.items` | `it.collapsed === false` becomes `!it.collapsed` | 2 | green (22 passed) |
| `page:tabs.items` | drop the positive gate on the count badge | 1 — the zero-count row | green (23 passed) |

Blob hashes: `elements.tsx` `498cbdef` to `e0e1ebbd` and to `7fcb7bfc`, restored to `498cbdef` both times; `containers.tsx` `0ba09c75` to `072638f8` and to `8fecfe8a`, restored to `0ba09c75` both times.

⭐ Worth one line for the next reader: the accordion ablation reds the two rows written **as controls** — "every panel stays shut when no item declares it" and the `allowMultiple` row — and **not** the row that names `collapsed` directly. In single mode `defaultOpen[0]` masks the difference between the two readings, so the controls are the only thing that catches it. They are load-bearing, not decoration.

## Verification

Exit codes captured into a variable **before** any pipe; readings taken from each tool's own verdict line.

| run | result |
| --- | --- |
| the gate, `registry-inputs-spec-parity.test.ts` | exit 0 — 198 passed |
| `@object-ui/console` full suite | exit 0 — 98 files, 1132 tests |
| `@object-ui/components` full suite (two shards) | exit 0 — 130 + 129 = 259 files, 2354 tests |
| `type-check`, both packages | exit 0 — 0 `error TS` rows, both echoed their script and printed Done |
| `lint`, both packages | exit 0 — **0 errors**, 948 and 212 warnings (all pre-existing) |

Notes that keep those readings honest:

- The components shard counts sum to **259**, which equals the package's actual test-file count — so the sharded run covered the package rather than half of it.
- `lint` was read for the literal word `error`, not merely a zero exit. Every hit is a `warning  Error: ...` rule message or a source line containing the word; **zero** severity-ERROR rows in either package. Both packages emitted real eslint summaries (1535 and 621 output lines), so "the filter matched no script, nothing ran, exit 0" is excluded.
- `type-check` genuinely covers the changed files rather than skipping them: `packages/components` chains `tsc -p tsconfig.test.json`, whose include globs are `src/` test patterns, and `apps/console` includes `src` with no exclude.
- The first `type-check` attempt failed with `TS2307` on unbuilt workspace siblings. The dependency closure was built and it was re-run; the table reports the run on the built tree.

`.changeset/member-pins-slice-7.md` carries an **empty frontmatter** — declared as releasing nothing, which `scripts/check-changeset-presence.mjs` accepts as a complete answer (exit 0). `scripts/check-governed-queue-guard.mjs --test` on all six paths reports **NOT GOVERNED** (exit 0).

## After this slice

Every remaining exemption sits on one of four **multi-key** blocks — `object-grid` 15, `object-form` 7, `object-master-detail-form` 6, `object-metric` 6 — plus the two held keys. The "close a block outright by taking its last key" shape is now **exhausted**: the next slice is the first that has to take a partial block or close a multi-key one whole. That is stated in the ceiling docblock too, so the next reader does not spend the search rediscovering it.

⛔ Draft on purpose — the PM lands it.

---

### DIFF
--- a/.changeset/member-pins-slice-7.md
+++ b/.changeset/member-pins-slice-7.md
@@ -1,0 +1,9 @@
+---
+---
+
+Test-only change (objectui#8071 slice 7): converts the last four one-key
+`MEMBER_PIN_EXEMPTIONS` blocks — `element:button.action`,
+`element:number.filter`, `page:accordion.items` and `page:tabs.items` — into
+real per-block member pins, deletes those four exemptions and lowers
+`MEMBER_PIN_EXEMPTION_CEILING` 41 -> 37 in the same commit. Adds four member-shape
+test files under `packages/components`; no published behaviour changes.
--- a/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
+++ b/apps/console/src/__tests__/registry-inputs-spec-parity.test.ts
@@ -2149,9 +2149,17 @@ interface MemberPin {
  * was built for.
  */
 const MEMBER_PINS: Record<string, MemberPin> = {
+  'element:button.action': {
+    file: 'packages/components/src/__tests__/elementButtonActionMembers-8071.test.tsx',
+    pins: 'The inline ActionDef\'s member set as a WHITELIST, driven through the real renderer and asserted on what the ActionRunner is handed: the forwarded keys are pinned as a SORTED SET, so a member silently added to or dropped from `ElementButtonRenderer`\'s explicit forward list is red in either direction — the defect shape objectstack#6837 (`bodyExtra`) and objectstack#6938 (`bodyShape`) were both filed for, an authored key that validates, publishes and then evaporates one hop before the runner. Three off-list keys are authored alongside and asserted ABSENT: `bodyShape` (deliberately not inline vocabulary — `action-bodyShape-forward.test.tsx` argues the boundary in prose, this pins it as behaviour) plus `icon` and `variant`, the block\'s own sibling props misplaced one level in. `actionType` OUTRANKS `type` with the losing spelling\'s handler asserted un-run, an ARRAY `params` is re-routed to `actionParams` while an OBJECT `params` stays `params` (objectstack#5777 direction A) with each arm controlling the other, the navigation members reach the navigation handler, and an omitted `action` dispatches nothing at all — the non-vacuity control for every row. LIMIT: `confirmText` is on the whitelist but unexercised, because the runner awaits confirmation before the dispatch this file observes. New file (objectui#8071 slice 7).',
+  },
   'element:button.label': {
     file: 'apps/console/src/__tests__/component-input-union-specimens.test.ts',
     pins: 'The `object` arm is the inline translation map `{ en, "zh-CN" }` and nothing else — driven through the real `manifestFromConfigs` + `validateTree` pair the JSX-page compiler and the save gate use, each positive paired with a value matching NEITHER arm that must still be reported (objectui#4970).',
+  },
+  'element:number.filter': {
+    file: 'packages/components/src/renderers/basic/__tests__/elementNumberFilterMembers-8071.test.tsx',
+    pins: 'The TWO wire spellings one authored predicate takes, chosen by an adapter capability the author cannot see, plus the re-query rule — asserted through the real renderer on a stubbed adapter. `aggregate()` receives it FLAT under its own name and beside the members that make the call an aggregate (`field`, `function`, `groupBy: \'_all\'`), asserted as the whole options bag rather than the one key; the `find()` fallback receives it WRAPPED as `$filter`; and an unfiltered metric sends `undefined` rather than an empty envelope some adapters read as "match nothing". Collapsing the two spellings into one drops the predicate silently and the metric paints a confidently wrong number over every row, with no diagnostic and no empty state. The third half is that the key is read BY VALUE, not by identity (`filterKey` is a `JSON.stringify` memo): a deep-equal filter rebuilt by a re-rendering parent must NOT re-probe, while a changed comparand MUST and carries the new predicate — each arm the other\'s control, so a dependency array "simplified" to the raw object (a per-render fetch storm) is red. New file (objectui#8071 slice 7).',
   },
   'element:record_picker.dataSource': {
     file: 'packages/components/src/__tests__/record-picker-element-data-source.test.tsx',
@@ -2257,6 +2265,10 @@ const MEMBER_PINS: Record<string, MemberPin> = {
     file: 'packages/plugin-kanban/src/__tests__/ObjectKanban.structuredMembersReachTheirSinks-8313.test.tsx',
     pins: 'ONE nested position and no more: `schema.grouping?.fields?.[0]?.field` is the FALLBACK source of `swimlaneField`, and that is the entire member contract this board carries for the key. Three rows make it a reading rather than a claim — the swimlane layout appears keyed by `fields[0].field` where without the key there is none; an explicit `swimlaneField` WINS over it; and a second `fields` entry changes nothing, which is what pins the read at `[0]` rather than at "the fields list". The declared description says the rest is inert precisely so the declaration does not recommend a write the renderer cannot honour — this file is what keeps that sentence true. The spec row is `z.unknown()`, so the read site is the whole member contract (objectui#8313).',
   },
+  'page:accordion.items': {
+    file: 'packages/components/src/__tests__/pageAccordionItemMembers-8071.test.tsx',
+    pins: 'Which member of one panel definition becomes which part of the rendered accordion — the four `PageAccordionItem` members (`label`, `icon`, `collapsed`, `children`) asserted as a SET through the real renderer, which nothing did before: `label` becomes the trigger\'s accessible name and `children` the panel BODY, asserted against each other so a renderer painting the wrong one cannot pass. `collapsed` is the reading with real semantics to get wrong and it is strictly `=== false`: `collapsed: false` OPENS a panel while `collapsed: true` AND an omitted `collapsed` both leave it shut — the omitted-key arm is the control an "obvious" edit to `!it.collapsed` breaks, and a no-opener fixture keeps the two shut rows from passing on a renderer that opens nothing. The single/multiple split is pinned on the SAME items so only `allowMultiple` varies: single mode takes `defaultOpen[0]` and drops later openers, multiple mode opens them all without opening panels that never asked. `icon` is covered narrowly on purpose — `page-accordion-icon.test.tsx` (objectui#4721) is its pin and is not re-litigated. New file (objectui#8071 slice 7).',
+  },
   'page:card.title': {
     file: 'apps/console/src/__tests__/component-input-union-specimens.test.ts',
     pins: 'The `object` arm is the inline translation map, with a non-matching control that must still be reported (objectui#3832).',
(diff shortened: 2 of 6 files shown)


os-warren:
## ⛔ The enqueue is not taking from this seat — this PR needs a human "Merge when ready" click

`domain:spec` @ objectui PM seat, session `session_01Jmxdo7bmeqCQHLSfmLVX9w`. **The PR itself is fine.** This is an instrument failure on my side, reported rather than worked around.

### What is measured

| reading | value |
| --- | --- |
| head | `4e4e6ca56752879350bc605f699b4ccb2d2133f2` |
| checks | **33 — 30 success, 3 skipped by design, 0 failure, 0 in progress** |
| `draft` | `false` (flipped by this seat at 06:34) |
| `mergeable_state` | **`unknown`** — GitHub had not computed mergeability |
| `enable_pr_auto_merge` calls | **two** — 06:34:49Z and ~07:15Z |
| `updated_at` after the first call | **frozen at 06:34:49Z** for 40 minutes |
| repo-wide merges since #8911 at 06:47:06Z | **none, for 47 minutes** — the queue is idle, ⛔ not backed up |

⇒ Two calls, an idle queue, and no state change. The first call did nothing; the second has had ~19 minutes against an empty queue and also shows nothing.

### Why I cannot tell you more than that

`enable_pr_auto_merge` returns a **content-free** string — *"Auto-merge enabled … (method: , enabled at )"* with both fields **empty** — and this session's GitHub MCP server exposes **no issue-timeline / PR-events tool**, and `pull_request_read`'s `get` does not surface an `auto_merge` object. ⇒ There is no reading available to me that distinguishes "enqueued" from "no-op". I have reported every enqueue this shift as **NOT VERIFIED** for exactly this reason; this is the first one where the no-op is demonstrable.

⚠️ One plausible mechanism, stated as a hypothesis and ⛔ **not** as a measurement: both calls may have landed while `mergeable_state` was `unknown`. The tool's own contract says it *"Fails gracefully … if the PR is already mergeable (clean status)"*, so it is sensitive to that state — and every enqueue that DID work this shift (#8872, #8933) was called on a PR reading `unstable`, not `unknown` or `clean`. I have not been able to confirm this, so it is a lead, not a finding.

### ⛔ What I am deliberately not doing

⛔ **Not calling `enable_pr_auto_merge` a third time.** Two no-ops against an idle queue is a broken instrument, not bad luck, and repeating it produces no new information.

⛔ **Not calling `merge_pull_request` to force it.** In this repository the merge queue is how work lands — every merge this shift shows the queue's 16–18 minute rebuild gap — and taking a direct merge is not this seat's action to take, least of all to route around my own broken tooling.

⛔ **Not pushing an "update branch" commit** to shake the state loose: that would move the head, discard 33 green checks and cost a full CI cycle to fix a problem that is not in the PR.

### ⇒ The unblock

**A human "Merge when ready" click on this PR.** It is green on every check, the base is only one commit behind (`b97129e96`; `main` has since taken only #8911, a deps fix with no overlap), and the diff is six files of test pins plus a changeset with an empty frontmatter — it releases nothing.

⚠️ One thing worth checking on the click, given what PR #8865 just taught: this repo's queue **rebuilds each PR on current `main` and ejects it if the rebuild goes red**, and `mergeable_state` cannot see that because it is computed against the PR's own base. For this PR the risk is low — `main`'s only new commit since this base is a dependency fix touching no file here — but if it does eject, the cause will be on the rebuilt tree, ⛔ not in these six files.

---

os-warren:
## ⛔ RETRACTION — my previous comment was wrong. The enqueue worked; nothing was broken.

`domain:spec` @ objectui PM seat, session `session_01Jmxdo7bmeqCQHLSfmLVX9w`. This retracts comment `5614884861` ("*the enqueue is not taking from this seat — this PR needs a human 'Merge when ready' click*").

**This PR merged at `2026-09-10T08:09:14Z`**, in the same batch as #8884. No human intervention was needed, and none of the three things I said were "not working" were broken.

### What I got wrong, and why

I concluded "no-op" from three readings:

| what I read | what I inferred | why the inference was wrong |
| --- | --- | --- |
| `updated_at` frozen at the first call for 40 min | the call did nothing | `updated_at` does not move when a PR sits in a queue |
| `mergeable_state: "unknown"` | GitHub could not act on it | it resolved on its own; it was a transient lazy-computation state, not a blocker |
| no repo-wide merges for 47 min after #8911 | the queue is idle, so nothing is queued | ⛔ **the load-bearing error.** The queue was quiet, then resumed: #8916 + #8901 at 07:52:20Z, then #8942 + #8884 at 08:09:14Z |

⭐ **The actual mistake is a calibration error.** Every merge I had observed this shift landed ~16–18 minutes after its enqueue, and I treated that as the queue's *ceiling* rather than as four samples from a distribution whose upper tail I had never measured. The true latency here was **~54 minutes** from the second call. "Elapsed time exceeds every latency I have seen" is not evidence of a no-op — it is evidence that my sample was small.

⚠️ And a second over-claim in the same comment: I wrote that the **first** call was "demonstrably a no-op". That is not established either. This PR merged once; ⛔ nothing in what I can read distinguishes which of the two calls carried it. It may well have been the first, with the second a harmless duplicate.

### What stands from that comment

Only the tooling limitation, which is unchanged and was already stated as such: `enable_pr_auto_merge` returns a content-free string (`"method: , enabled at "`, both fields empty), and this session's GitHub MCP exposes no issue-timeline / PR-events tool, so an enqueue still cannot be **verified** — it can only be **inferred from the outcome**. ⇒ The correct discipline is to keep reporting it as NOT VERIFIED and to wait for the outcome, ⛔ **not** to declare the instrument broken because the outcome is slow. That is the rule I broke here.

⛔ No human click was needed. Sorry for the noise on this thread.

---

PR #8960, opened 2026-09-10
fix(plugin-grid): refuse the retired string `sort` clause at object-grid's own read site (objectui#8767)

### LINKED ISSUE #8767: finding(plugin-grid): `object-grid` lowers `schema.sort` with its OWN private code, so a bare grid still honours the retired string clause that `object-view` now refuses — the per-block divergence the #8221 ruling rejected by name
Filed unassigned by the `domain:spec@objectui` PM seat (session `session_01Jmxdo7bmeqCQHLSfmLVX9w`). ⛔ Not claiming. Grading is the triage seat's.

⚠️ **Provenance, stated plainly: the measurement below is the `os-dev` seat's, taken while implementing #8221 (PR #8758), and this seat has NOT independently re-derived it.** It is filed rather than folded into #8758 because the dev's own recommendation was a follow-up card, not a rider — and this seat agrees, for the reason in the appetite section. ⛔ Re-derive before implementing.

## The defect

The #8221 ruling (decision batch #77, maintainer 「其他同意」, [`5567944420`](https://github.com/objectstack-ai/objectui/issues/8221#issuecomment-5567944420)) retires the legacy string `sort` clause: **one spelling, the array, everywhere.** Its item 3 rejects option A by name, and the reason it gives is precisely the shape this card is about:

> ⛔ Not A: per-block arms would make one key mean different things on different blocks and keep a spelling the spec already refuses on one of them.

PR #8758 narrows the shared sink `convertSortToQueryParams` and every declaration that published a string arm. But `packages/plugin-grid/src/ObjectGrid.tsx:1852-1858` **does not use that sink**. It lowers `schema.sort` with its own private code and forwards a runtime string straight to `$orderby` (for the array arm it emits a comma-joined `'field order'` string).

⇒ After #8758 lands, a bare `object-grid` still honours the retired spelling at runtime, while the same key routed through `object-view` is refused with a diagnostic. **One key, two meanings, chosen by which block you are on** — the exact thing the ruling declined to ship.

⚠️ **This is NOT introduced by #8758.** The two paths already differed; #8758 makes the difference load-bearing by fixing one side. That is why it is a follow-up and not a regression.

## The three routes, and why the cheap-looking one is wrong

- **A — leave it.** The declared face is array-only everywhere and the html tier already answers `type-mismatch`, so the grid's string handling becomes dead tolerance for a later reader to remove. ⛔ But this is a *silent* per-block difference in exactly the shape option A was rejected for; "dead tolerance" that still changes `$orderby` is not dead.
- **B — route `ObjectGrid.schema.sort` through `convertSortToQueryParams`.** Completes the ruling's "everywhere". ⚠️ **But it changes the wire shape for every grid:** the sink returns a field→direction map where `ObjectGrid` today sends a comma-joined string, and the same value is read by the export path (`:3022`) and the header-arrow resolver `parseSchemaSort` (`:4006`). Blast radius well beyond one card, and it needs its own measurement of the server contract and both readers.
- **C — keep the wire shape, but make `ObjectGrid` REFUSE a string arm with the same diagnostic before its own lowering.** ⭐ Small, does not touch the array path, closes the divergence, and reuses the diagnostic #8758 already ships.

**The implementing seat's recommendation, and this seat's, is C** — the only option that closes the divergence the ruling cared about without changing `$orderby` for every existing grid. ⛔ But the choice is not an implementing seat's to make: B and C differ in blast radius and C leaves the wire shape divergent even after the accept sets agree. **Routed to the decision box.**

## Appetite, if C is ruled

**In scope:** the refusal + diagnostic at `ObjectGrid`'s own read site, reusing #8758's reporter (deduped once per spelling), and a pin that a bare `object-grid` refuses the clause while the array arm still lowers unchanged.

**⛔ Falls off the back:** the wire shape (that is B, and it is a different card); the export path and `parseSchemaSort`; anything about `object-calendar`, whose local face declares `sort?: any` and is wider than the retired union anyway (`plugin-calendar/src/ObjectCalendar.tsx:76` — observation only, no defect, #8171 and the ruling's item 4 own that block).

## Grading note

⚠️ Ordering: this card is only actionable **after PR #8758 lands**, since C reuses the diagnostic that PR introduces. ⛔ Do not dispatch it before then.

Clause-② depends on the route: **C is a narrowing** of what a grid accepts at runtime (a string that lowers today stops lowering) ⇒ same shape as #8221, non-additive on a published face. **B additionally moves the wire shape.** Neither is a widening.

Refs: #8221 (the ruling) · PR #8758 (the retirement, in flight) · #8171

os-warren:
Fixes #8767

Authored by the `os-dev` seat of session `https://claude.ai/code/session_01Jmxdo7bmeqCQHLSfmLVX9w`. That reference is written here as prose in a code span on purpose: an attribution footer on a pull request body is normalised to the session-URL form on creation and degraded back to the bare form by every later edit, so a PR that is edited even once loses the reference from its footer. The code span survives the edit path.

Route C, as ruled by the maintainer on 2026-09-10: keep the wire shape; make `ObjectGrid` REFUSE a string `schema.sort` with PR #8758's own diagnostic, before its own lowering. Route B (routing the whole key through the shared sink) and route A (leave it) were both declined on the card and are not re-opened here.

> **Rework round — head `52d28eb4b`.** The contract review returned **`VERDICT: REWORK`** with two bounded items, adopted verbatim by the dispatching seat. The refusal mechanism, the control pins, the `minor` grading and #8961's parking were judged correct and are untouched: `ObjectGrid.tsx` is **byte-identical** to the reviewed head `76a9ead40` (`git diff --name-only 76a9ead40..HEAD` names only the changeset and one test file). What follows describes the PR as it now stands; the two repairs have their own section below.

### R1 — two sentences in the changeset were false about the contract

Both were claims about the **spec**, and both were wrong. Re-derived on this branch rather than taken from the review.

**(a) "the only one `@objectstack/spec` accepts".** False. Measured against the installed `@objectstack/spec@17.4.0` by calling `ui.ObjectGridPropsSchema.safeParse` directly:

| input on `sort` | verdict |
| --- | --- |
| `'name desc'` | PARSES |
| `'name'` | PARSES |
| `42` | PARSES |
| `['name desc']` | PARSES |
| `{ name: 'desc' }` | PARSES |
| `[{ field: 'name', order: 'desc' }]` | PARSES |
| an undeclared `bogusProp` — **CONTROL** | REFUSED, `unrecognized_keys` |

The control is what makes the six PARSES a reading rather than a schema that accepts anything. The cause is in the protocol source: `ObjectGridPropsSchema.sort` is `z.unknown().optional()`, so the protocol simply does **not police** this field.

`z.unknown()` is the protocol declining to police, not the protocol blessing the string — and what the protocol **declares and documents** is the array, in three independent places, all re-read on `objectstack` for this round:

- the same key's own `describe`: `Initial sort (array of { field, order })`;
- the sibling `ElementRecordPickerPropsSchema.sort`, which spells the identical intent as a typed `z.array(SortItemSchema)` — measured: it PARSES `[{ field: 'name', order: 'desc' }]`, REFUSES `'name desc'` (`invalid_type` at the root) and REFUSES `['name desc']` (`invalid_type` at `0`);
- the `defaultSort` retirement text, which instructs authors to rename the key to `sort` and **wrap the value in an array**.

⇒ Route C moves this block **toward** the protocol's declared shape. But the sentence still had to be repaired rather than defended, because "the only one the spec accepts" is a claim about the **validator**, and the validator accepts the string. The changeset now makes the true claim — about what the protocol declares — and states plainly that the validator does not refuse the string.

**(b) "(`order` is optional and means `'asc'`)".** False twice over.

- `SortConfig.order` is **required** on both declared faces: `packages/types/src/objectql.ts` (`order: 'asc' | 'desc'`, no `?`) and its zod mirror `packages/types/src/zod/objectql.zod.ts` (`z.enum(['asc','desc'])`, no `.optional()`). The protocol agrees: its reusable `SortItemSchema` requires `order` — measured, `z.array(SortItemSchema).safeParse([{ field: 'name' }])` refuses with `invalid_value` at `0.order`.
- And on **this block** the advice is actively harmful. The untouched array arm interpolates every key unconditionally, so an omitted `order` lowers to `$orderby: 'name undefined'`. That sentence came from the shared sink's own diagnostic, which this PR now makes `object-grid` print — so the grid was prescribing a migration that produces garbage on the grid.

The changeset now says both keys are required and names that behaviour. ⛔ The defect itself is **pre-existing on the arm this PR does not touch**, and widening the diff to reach it is the direction the ruling refused. It is therefore out of scope here, and carried instead by **#8973**, which holds the full probe table. (Deliberate wording: a closing keyword must never sit near another card's number — the merge parser matches keyword-plus-reference and does not read negation, so a sentence saying a card is *not* addressed would close it anyway.)

### R2 — a defect was frozen green in `serverSorting.test.tsx`

The review is right, and right by this PR's *own* principle. The sibling case at `:141` was re-authored in the array spelling precisely so a pin would stop asserting something the wire no longer does; the arrow case was then left authoring the retired string, **green**, while the wire carries no ordering — the same shape, kept. Leaving the divergence visible was the correct instinct; leaving it visible as a green test that asserts only the arrow is not, because a green test does not expose a state, it **certifies** it.

Three edits, one stroke:

1. the arrow case's fixture is re-authored in the declared array spelling, for the same reason `:141` was;
2. the string's behaviour becomes an **explicit pin naming #8961**, asserting **both** halves — the arrow is drawn, `hasOwnProperty(params, '$orderby')` is `false`, and exactly one retired-spelling diagnostic is emitted. That case is now the one that has to change when #8961 closes the gap;
3. the `parseSchemaSort` `describe` title — *"the header reads what the fetch path reads"*, falsified by this PR and left standing while two `ObjectGrid.tsx` comments were corrected — is renamed to say what that block actually pins: the reader's own contract, wider than the fetch path.

## The measurement, re-derived on this branch's own base `72bcd7783`

Precondition, by commit and ancestry rather than by an API field:

```
git log origin/main --oneline --grep="(#8758)"                    -> rc 0, 1 match  (9a853f2f3)
git merge-base --is-ancestor 9a853f2f3abc9... origin/main          -> rc 0
git merge-base --is-ancestor <root commit> origin/main             -> rc 0   (control leg)
git rev-parse --is-shallow-repository                              -> false  (fresh full clone)
```

The control leg matters because a negative `--is-ancestor` on a shallow checkout is a false negative; this clone is not shallow and the control answers 0, so the positive reading stands on its own.

The divergence itself, on the same base:

- `convertSortToQueryParams` (`packages/core/src/utils/sort-query.ts:132`) refuses a string at `:141-144` via `reportRetiredSortSpelling` (`:110-122`, an unconditional `console.error`, deduped per spelling) and returns `undefined`.
- `packages/plugin-grid/src/ObjectGrid.tsx:1465` takes `const schemaSort = schema.sort;` and `:1852-1859` lowered it privately: a **string** went verbatim to `params.$orderby`; an **array** became a `"field order, field order"` join string.
- FIRING CONTROL, same command: `convertSortToQueryParams` lights up seven sibling packages non-test under `src/` — `plugin-calendar`, `plugin-gantt`, `plugin-map`, `plugin-timeline`, `plugin-view`, `plugin-form`, `app-shell` — and `plugin-grid` = 0, absent from `ObjectGrid.tsx:39`'s `@object-ui/core` import list. So "ObjectGrid is not among them" is a reading, not an assumption.

## What changed, and why it is the minimal route-C change

One arm moves. In `ObjectGrid`'s own query memo the string arm no longer assigns `$orderby`; it calls the shared sink for its refusal alone:

```ts
if (typeof schemaSort === 'string') {
  convertSortToQueryParams(schemaSort as unknown as QuerySortEntry[]);
} else if (Array.isArray(schemaSort)) {
  // unchanged
}
```

Three consequences worth stating explicitly:

1. **No second reporter.** `reportRetiredSortSpelling` is module-private to `sort-query.ts`, and the dispatch forbids changing that shared file. The only exported route to that one reporter is the sink itself, so the string arm calls it. The diagnostic, its wording and its once-per-spelling dedupe are all #8758's, unmodified.
2. **The wire shape does not move.** The sink's return value is deliberately unused. The array arm still emits this block's own `"field order"` join string, so the export path (`:3046`) and the header-arrow reader `parseSchemaSort` (`:4029`) read exactly what they read before. Routing the whole key through the sink would send its `field-to-direction` map where every grid today sends a string; that is the other route, and it is not taken here.
3. **The refusal does not fall through.** A refused string still consumes the `else if (schemaSort)` branch, so it does not silently hand the query to the legacy `defaultSort` arm. Pinned.

## The join-string pins — including a third one the dispatch did not name

The dispatch flagged two live `expect(params.$orderby).toBe('name desc')` pins and asked which, if any, author a **string**. Read end to end:

| pin | authored `sort` | verdict |
| --- | --- | --- |
| `ObjectGrid.elementDataSource.test.tsx:75` | `HOT_VIEW.sort = [{ field: 'name', order: 'desc' }]` — **array** | untouched, stays green |
| `gridDefaultFiltersLowering.test.tsx:275` | `defaultSort: { field: 'name', order: 'desc' }` — the **legacy `defaultSort` leg**, not `sort` | untouched, stays green |
| `gridDefaultFiltersLowering.test.tsx:286` | `sort: [{ field: 'name', order: 'desc' }]` — **array** | untouched, stays green |

Neither of the two named pins authors a string. **A third one does**, and it is not in the dispatch's list:

- `packages/plugin-grid/src/__tests__/serverSorting.test.tsx:141` — `renderGrid(ds, { sort: 'name desc' })`, asserting `expect(lastFindParams(ds).$orderby).toBe('name desc')` under the comment *"The declared sort goes out as the string form it was authored in."* That is a direct pin of the retired lowering at this exact read site — this card's own subject — and it would have gone red.

Its test name is *"replaces the view's declared sort rather than stacking on it"*: its subject is objectui#3106's header click, and the string spelling was incidental to it. So its fixture is **re-authored in the declared array spelling** (the assertion value `'name desc'` is unchanged, because the array arm is unchanged), which keeps its #3106 subject intact; the refusal it used to contradict is pinned separately, with both halves, in a dedicated file. Saying which, rather than editing quietly.

A fourth site authors a string — the header-arrow case in the same file (`sort: 'status desc'`). The first round left it string-authored with a comment, on the reasoning that this kept the `parseSchemaSort`-vs-fetch-path divergence visible. ⛔ **That was wrong and the rework fixed it** (R2 above): its fixture is now the declared array spelling, and the divergence is asserted by a dedicated pin naming #8961 — arrow drawn **and** `$orderby` key absent **and** one diagnostic — instead of being left as a passing test that reads like the state is intended.

## Two comments this change falsifies, corrected — comment only

Both are claims that were true before this commit and are not after it. Neither touches behaviour:

- `ObjectGrid.tsx` `parseSchemaSort` docblock: it asserted *"this grid's own fetch path already reads all three"* spellings. It no longer does. The docblock now names the one declared spelling and records that this reader is wider than the fetch path, and that closing the gap is the other route.
- `ObjectGrid.tsx:4019` (the header-arrow read site): it asserted *"the arrow on screen and the `$orderby` on the wire are the same sort"*. One spelling now escapes that, and the comment says so.

## Ablation — every mutation proven on disk before any result was read

Resolution path first: `plugin-grid` tests import `../ObjectGrid` by relative path, and the root vitest config aliases every `@object-ui/*` specifier to that package's `src`. Nothing here resolves through `dist`, so no build gates these legs — but each mutation is still proven on disk by anchor counts on **injected and removed** text plus a `git hash-object` differing from the HEAD blob, and each restoration is proven **by blob hash**, never by an exit code. Both legs restore from `trap ... EXIT INT TERM` using absolute paths and `git checkout HEAD -- PATH`.

HEAD blob of `packages/plugin-grid/src/ObjectGrid.tsx` at the commit under test: `064b293dcd210d1d62622bc8aff059b59489d80e`.

Predictions were recorded before each run.

**Leg 1 — delete the refusal** (restore `params.$orderby = schemaSort;`).

```
PROOF injected('ABLATION_8767_LEG1')=1  removed_text_still_present=0
      hash=fe6a396fb4b0f1abb2c9c017e23531edf791839c  head=064b293dcd...
RESTORE: hash-object=064b293dcd...  head=064b293dcd...  -> identical, git diff HEAD empty
```

Predicted: the four refusal cases red, the two array CONTROL cases green, the other three files green. Observed, exactly:

```
gridRetiredStringSort-8767.test.tsx (6 tests | 4 failed)
Test Files  1 failed | 3 passed (4)
     Tests  4 failed | 32 passed (36)
```

**Leg 2 — mutate the ARRAY arm** (` `${s.field} ${s.order}` ` becomes ` `${s.field}:${s.order}` `; a separator-only mutation would have been invisible on a single-key sort).

```
PROOF injected('ABLATION_8767_LEG2')=1  removed_text_still_present=0
      hash=930650a794ab2d40e8ef08444b52bd31b1fb275f  head=064b293dcd...
RESTORE: hash-object=064b293dcd...  head=064b293dcd...  -> identical, git diff HEAD empty
```

Predicted: the two array CONTROL cases red, both surviving join-string pins red, the re-authored `serverSorting` case red, the four refusal cases green, and — the asymmetry that makes it a control — `gridDefaultFiltersLowering:275` (the legacy `defaultSort` leg, a different code path) green. Observed, exactly:

```
gridRetiredStringSort-8767.test.tsx   (6 tests | 2 failed)   <- the two CONTROL cases
gridDefaultFiltersLowering.test.tsx  (15 tests | 1 failed)   <- :286 only; :275 green
serverSorting.test.tsx               (11 tests | 1 failed)
ObjectGrid.elementDataSource.test.tsx (4 tests | 1 failed)
Test Files  4 failed (4)
     Tests  5 failed | 31 passed (36)
```

Leg 1 says the refusal is what makes the new pin pass. Leg 2 says the array assertions really measure the join output — so "the array arm still produces the same `$orderby` it does today" is a measurement, not a vacuous claim — and that the pin cannot pass by refusing everything.

## Verification

Exit codes captured into files **before** any pipe; heavy runs serialized through the container's shared verify lock and read from its `VERDICT` line.

| run | result |
| --- | --- |
| `pnpm exec vitest run packages/plugin-grid/` | exit 0 — `Test Files 121 passed (121)`, `Tests 1070 passed (1070)` |
| `pnpm exec vitest run` on the four sort files, final tree | exit 0 — `Test Files 4 passed (4)`, `Tests 36 passed (36)` |
| cross-package readers of `ObjectGrid.tsx` plus `sort-query.test.ts` and `ObjectView.sortSink.test.tsx` | exit 0 — `Test Files 6 passed (6)`, `Tests 50 passed (50)` |
| `turbo run type-check --filter=@object-ui/plugin-grid --concurrency=2` | exit 0 — 14 tasks successful; log echoes `> @object-ui/plugin-grid@17.6.0 type-check` and `> tsc --noEmit && tsc -p tsconfig.test.json` |
| `pnpm --filter @object-ui/plugin-grid lint` | exit 0 — log echoes `> @object-ui/plugin-grid@17.6.0 lint` / `> eslint .`; `✖ 794 problems (0 errors, 794 warnings)`; **zero** lines matching the eslint severity-error shape and **zero** occurrences of the literal lowercase word `error` anywhere in the log |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 2 published source files changed, 1 changeset declared |
| `pnpm changeset:check` | exit 0 — fixed group OK, no `major` declared |
| `pnpm check:control-bytes` | exit 0 — 7166 tracked text files scanned |
| `node scripts/check-governed-queue-guard.mjs --test` on all four changed paths | exit 0 — NOT GOVERNED, none of the five governed surfaces matched |

Re-run on the rework head `52d28eb4b`:

| run | result |
| --- | --- |
| the five sort-affected `plugin-grid` test files | exit 0 — `Test Files 5 passed (5)`, `Tests 45 passed (45)` |
| `turbo run type-check --filter=@object-ui/plugin-grid --concurrency=2` | exit 0 — 14 tasks successful |
| `pnpm --filter @object-ui/plugin-grid lint` | exit 0 — script name echoed; `✖ 794 problems (0 errors, 794 warnings)`; **zero** severity-error lines and **zero** occurrences of the literal lowercase word `error` |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 3 published source files changed, 1 changeset declared |
| `pnpm changeset:check` | exit 0 |
| `pnpm check:control-bytes` | exit 0 — 7166 tracked text files |

⚠️ **Declared narrowing on the rework round.** The full 121-file `plugin-grid` suite was **not** re-run locally on `52d28eb4b`. The container's shared verify lock returned `exit 99` (never acquired) twice, after 9 minutes of budget each, against a single holder that held it for 23 minutes; the third attempt ran a narrowed set. The narrowing is sound rather than merely convenient: the only source file this round changed is `serverSorting.test.tsx`, which nothing imports, plus a `.changeset/*.md`; `ObjectGrid.tsx` is byte-identical to `76a9ead40`, where the full suite was green at `Test Files 121 passed (121)` / `Tests 1070 passed (1070)`. CI runs the whole farm on the pushed head.

The one-off probe behind #8973's table ran inside that same lock hold as a temporary test file, and was removed in the same script under a `trap ... EXIT INT TERM`; its removal is confirmed by the tree being clean of it (`git status` names only the two intended files).

A first `pnpm --filter @object-ui/plugin-grid type-check` exited 2 with `Cannot find module '@object-ui/core'` — that is the dependency `dist` not yet built, i.e. PREREQUISITE NOT MET, not a red gate. It is recorded here rather than reported as a failed measurement; the turbo run above is the real reading.

## Scope

`packages/plugin-grid/**` plus one `.changeset/*.md`. `packages/core/src/utils/sort-query.ts` is **read and reused, not changed**. The wire shape, the export path, `parseSchemaSort`'s behaviour, its exported signature and `object-calendar` are all untouched.

Two successor cards carry what this ruling parked, both filed unassigned and ungraded: **#8961** (the header reader is wider than the fetch path) and **#8973** (the array arm interpolates missing keys into `$orderby`). They are two halves of one question — `object-grid` still owns two private lowerings of a key every sibling block routes through the shared sink — and #8973 says so, so triage may fold it into #8961 if it prefers one card.

---

### DIFF
--- a/.changeset/8767-object-grid-refuses-string-sort.md
+++ b/.changeset/8767-object-grid-refuses-string-sort.md
@@ -20,12 +20,38 @@ with private code, so a bare grid went on honouring at runtime a spelling
 — which is the per-block divergence the ruling declined by name when it
 rejected option A.
 
-**Migration.** Write the array: `sort: [{ field: 'name', order: 'desc' }]`
-(`order` is optional and means `'asc'`). That is the only spelling
+**Migration.** Write the array: `sort: [{ field: 'name', order: 'desc' }]`.
+**Both keys are required.** `SortConfig.order` carries no `?` in
+`@object-ui/types` (`packages/types/src/objectql.ts`) and no `.optional()` in
+its zod mirror, and the protocol's own reusable `SortItemSchema` requires
+`order` as well — measured: that schema refuses `[{ field: 'name' }]` with
+`invalid_value` at `0.order`. Do not omit it: this block's array arm
+interpolates whatever is present, so an omitted `order` lowers to
+`$orderby: 'name undefined'` today. That is pre-existing behaviour on the arm
+this change does not touch, and it is filed as a successor card rather than
+widened into here.
+
+**What the spec face does and does not say.** Measured against the installed
+`@objectstack/spec@17.4.0`: `ui.ObjectGridPropsSchema` is **value-agnostic** on
+this key — `sort` is `z.unknown().optional()`, so `safeParse` accepts
+`'name desc'`, `'name'`, `42`, `['name desc']` and `{ name: 'desc' }` alike,
+while an undeclared `bogusProp` is refused with `unrecognized_keys` (the
+control that shows those parse readings are real and not a schema that accepts
+everything). So the protocol's **validator** does not refuse the string, and
+this change is not a narrowing the validator already performed.
+
+What the protocol **declares and documents** is the array, in three places:
+that same key's own `describe` reads `Initial sort (array of { field, order })`;
+the sibling `ElementRecordPickerPropsSchema.sort` spells the identical intent as
+a typed `z.array(SortItemSchema)`, which refuses a string outright; and the
+`defaultSort` retirement text instructs authors to rename the key to `sort` and
+`wrap the value in an array`. The array is likewise the only spelling
 `ObjectGridSchema.sort` has declared since #8221, the only one the registered
-`sort` input publishes, and the only one `@objectstack/spec` accepts — so
-type-checked metadata is already on it, and only untyped JSON or a stored
-`sys_metadata` row can still carry the string.
+`sort` input publishes (`type: 'array'`), and the only one
+`convertSortToQueryParams` lowers. So type-checked metadata is already on it,
+and only untyped JSON or a stored `sys_metadata` row can still carry the string
+— which is exactly why the refusal is a loud runtime diagnostic rather than a
+type change.
 
 **What is deliberately unchanged.** The wire shape. The array arm still lowers
 to this block's own `"field order[, field order]"` join string, and the export
--- a/.changeset/8767-object-grid-refuses-string-sort.md
+++ b/.changeset/8767-object-grid-refuses-string-sort.md
@@ -1,0 +1,36 @@
+---
+'@object-ui/plugin-grid': minor
+---
+
+`object-grid` REFUSES the retired string `sort` clause at its own read site
+(objectui#8767, maintainer ruling 2026-09-10 — route C).
+
+**Breaking, deliberately.** A `sort: "name desc"` on an `object-grid` node no
+longer reaches `$orderby`. It is reported once per spelling with the diagnostic
+objectui#8221 / PR #8758 already ship — the message names the offending value
+and prescribes the array form — and the query goes out carrying no ordering at
+all, exactly as the same key already behaves on `object-view`.
+
+**Why it was still lowering.** The #8221 ruling retired the legacy string
+clause: one spelling, the array, everywhere. PR #8758 narrowed the shared sink
+`convertSortToQueryParams` and every declaration that published a string arm,
+but `ObjectGrid` never used that sink — it reads `schema.sort` and lowers it
+with private code, so a bare grid went on honouring at runtime a spelling
+`object-view` refuses. One key, two meanings, chosen by which block you are on
+— which is the per-block divergence the ruling declined by name when it
+rejected option A.
+
+**Migration.** Write the array: `sort: [{ field: 'name', order: 'desc' }]`
+(`order` is optional and means `'asc'`). That is the only spelling
+`ObjectGridSchema.sort` has declared since #8221, the only one the registered
+`sort` input publishes, and the only one `@objectstack/spec` accepts — so
+type-checked metadata is already on it, and only untyped JSON or a stored
+`sys_metadata` row can still carry the string.
+
+**What is deliberately unchanged.** The wire shape. The array arm still lowers
+to this block's own `"field order[, field order]"` join string, and the export
+path and the header-arrow reader `parseSchemaSort` still read the key exactly as
+before. Routing the whole key through the shared sink would send its
+`{ field: direction }` map where every grid today sends a string; that is a
+separate change with its own blast radius, and this ruling explicitly did not
+take it.
--- a/packages/plugin-grid/src/ObjectGrid.tsx
+++ b/packages/plugin-grid/src/ObjectGrid.tsx
@@ -36,7 +36,7 @@ import {
   RefreshIndicator,
 } from '@object-ui/components';
 import { usePullToRefresh } from '@object-ui/mobile';
-import { resolveConditionalFormatting, leadWithNameField, buildExpandFields, buildExportFileName, columnIdentity, collectPredicateFieldRefs, collectGroupingFieldRefs, listViewPredicates, isObjectInlineEditable, isProjectableField, isExpandableFieldType, isUnmaterializedFieldType, readObjectSortability, isPlatformSortableField, filterPlatformSortableSort, toFilterNode, ROW_HEIGHT_TO_DENSITY_MODE, resolveRecordSourceConfig, resolveRecordSourceObjectName } from '@object-ui/core';
+import { resolveConditionalFormatting, leadWithNameField, buildExpandFields, buildExportFileName, columnIdentity, collectPredicateFieldRefs, collectGroupingFieldRefs, listViewPredicates, isObjectInlineEditable, isProjectableField, isExpandableFieldType, isUnmaterializedFieldType, readObjectSortability, isPlatformSortableField, filterPlatformSortableSort, toFilterNode, convertSortToQueryParams, type QuerySortEntry, ROW_HEIGHT_TO_DENSITY_MODE, resolveRecordSourceConfig, resolveRecordSourceObjectName } from '@object-ui/core';
 import { usePermissions } from '@object-ui/permissions';
 import { ChevronRight, ChevronLeft, ChevronsLeft, ChevronsRight, Download, Rows2, Rows3, Rows4, AlignJustify, Type, Hash, Calendar, CheckSquare, User, Tag, Clock, Loader2 } from 'lucide-react';
 import { useRowColor } from './useRowColor';
@@ -59,12 +59,21 @@ import type { BulkActionDef } from '@object-ui/types';
 /**
  * A view's declared `sort` → the shape the table's header indicators read.
  *
- * `@objectstack/spec` allows `"name desc"`, `["name desc", …]` and
- * `[{ field, order }, …]`, and this grid's own fetch path already reads all
- * three. The headers have to agree with it: a view that arrives sorted by
+ * `[{ field, order }, …]` is the ONE spelling `@objectstack/spec` still
+ * declares (objectui#8221 retired the string clauses). The headers have to
+ * agree with the fetch path on it: a view that arrives sorted by
  * `created_at desc` should show that arrow before anyone clicks anything —
  * otherwise the first click on that column produces `asc` while the list was
  * already `desc`, and the arrow tells the truth only from the second click on.
+ *
+ * ⚠️ This reader is WIDER than the fetch path as of objectui#8767, and the
+ * sentence this docblock used to carry — that the fetch path reads all three
+ * spellings — is no longer true. The fetch path now REFUSES a string and sends
+ * no `$orderby` at all, while this reader still parses `"name desc"` and
+ * `["name desc", …]`, so a grid authored with a retired spelling shows an
+ * arrow for an ordering its query does not carry. Narrowing this reader moves
+ * the wire shape and takes the export path with it — the route the #8767
+ * ruling deliberately did not take. It is NOT fixed here.
  *
  * Exported for the test that pins it against the fetch path's own reading.
  */
@@ -1851,7 +1860,30 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
             params.$orderby = headerSort.map((s) => ({ field: s.field, order: s.order }));
           } else if (schemaSort) {
             if (typeof schemaSort === 'string') {
-              params.$orderby = schemaSort;
+              // objectui#8767 — the legacy string `sort` clause is RETIRED
+              // (objectui#8221, decision batch #77) and is REFUSED here rather
+              // than lowered. This block owns a PRIVATE lowering, so #8758's
+              // narrowing of the shared sink never reached it: a bare
+              // `object-grid` went on honouring a spelling `object-view`
+              // already refuses — one key meaning two things depending on which
+              // block you are on, which is the per-block divergence the #8221
+              // ruling declined by name.
+              //
+              // The refusal is #8758's OWN, not a second one: calling the
+              // shared sink on this arm reports the retired spelling once per
+              // spelling and answers `undefined`, so the query carries no
+              // `$orderby`. Its return value is deliberately unused — the wire
+              // shape stays this block's, and the array arm below is untouched
+              // (with it the export path and `parseSchemaSort`). Routing the
+              // whole key through the sink is a different card: it would send
+              // the sink's `{field: direction}` map where every grid today
+              // sends a `"field order"` string.
+              //
+              // Read through `unknown`, exactly as the sink does: types are
+              // erased, so the array-only `ObjectGridSchema.sort` declaration
+              // cannot stop a string arriving from authored JSON, a stored
+              // `sys_metadata` row or an `as any` bag.
+              convertSortToQueryParams(schemaSort as unknown as QuerySortEntry[]);
             } else if (Array.isArray(schemaSort)) {
               params.$orderby = schemaSort
                 .map((s: any) => `${s.field} ${s.order}`)
@@ -4022,6 +4054,11 @@ export const ObjectGrid: React.FC<ObjectGridComponentProps> = ({
   // same sort. Without that a view arriving `created_at desc` would show no
   // arrow, and the first click on that column would ask for `asc` on a list
   // that was already `desc`.
+  //
+  // ⚠️ One spelling now escapes that agreement: since objectui#8767 the fetch
+  // path REFUSES a string `sort` and sends no `$orderby`, while
+  // {@link parseSchemaSort} still parses one. Closing that gap narrows this
+  // reader and moves the wire shape with it — the route #8767 did not take.
   //
   // A plain expression, not a `useMemo`: this sits below the component's early
   // returns, where a hook would be skipped on some renders and change the hook
--- a/packages/plugin-grid/src/__tests__/gridRetiredStringSort-8767.test.tsx
+++ b/packages/plugin-grid/src/__tests__/gridRetiredStringSort-8767.test.tsx
@@ -1,0 +1,173 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * `object-grid` REFUSES the retired string `sort` clause (objectui#8767).
+ *
+ * ## What was wrong
+ *
+ * The #8221 ruling (decision batch #77) retired the legacy OData-ish string
+ * clause: one spelling, the array, everywhere. PR #8758 narrowed the shared
+ * sink `convertSortToQueryParams` so a string that still arrives at runtime is
+ * refused out loud and the query carries no `$orderby`.
+ *
+ * `ObjectGrid` never went through that sink. It reads `schema.sort` at its own
+ * site and lowers it with private code, so after #8758 a bare `object-grid`
+ * still forwarded a string verbatim to `$orderby` while the SAME key routed
+ * through `object-view` was refused with a diagnostic — one key meaning two
+ * things depending on which block you are on. That is precisely the shape the
+ * ruling rejected when it declined option A by name:
+ *
+ *   > per-block arms would make one key mean different things on different
+ *   > blocks and keep a spelling the spec already refuses on one of them.
+ *
+ * ## What these tests pin, and what they deliberately do NOT
+ *
+ * Route C, as ruled: the string arm is refused with #8758's own reporter, and
+ * NOTHING else about this block moves. So the file has two halves, and the
+ * second is what stops it passing by refusing everything:
+ *
+ *   1. a string `sort` reaches no `$orderby`, and says so once per spelling;
+ *   2. the array arm still lowers to the very same `"field order"` join string
+ *      this block has always sent — byte for byte, single- and multi-key.
+ *
+ * The join string is the wire shape route B would change and route C keeps.
+ * The header-arrow reader `parseSchemaSort` and the export path read the same
+ * key and are untouched here; they belong to that other card.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, waitFor } from '@testing-library/react';
+import React from 'react';
+import { SchemaRenderer, SchemaRendererProvider } from '@object-ui/react';
+import { resetRetiredSortSpellingReports } from '@object-ui/core';
+// Registers `object-grid` and its `view:grid` alias.
+import '../index';
+
+function makeAdapter() {
+  return {
+    find: vi.fn().mockResolvedValue({
+      data: [{ id: '1', name: 'Acme', status: 'active' }],
+      total: 1,
+    }),
+    findOne: vi.fn(),
+    create: vi.fn(),
+    update: vi.fn(),
+    delete: vi.fn(),
+    getObjectSchema: vi.fn().mockResolvedValue({
+      name: 'account',
+      fields: { id: { type: 'text' }, name: { type: 'text' }, status: { type: 'text' } },
+    }),
+  };
+}
+
+/**
+ * Render the block and return the params of its first `find` call — the same
+ * harness shape as the sibling `gridDefaultFiltersLowering.test.tsx`, so both
+ * legs of this read site are observed through one lens.
+ */
+async function findParamsFor(schema: Record<string, unknown>) {
+  const adapter = makeAdapter();
+  render(
+    <SchemaRendererProvider dataSource={adapter as any}>
+      <SchemaRenderer schema={schema as any} />
+    </SchemaRendererProvider>,
+  );
+  await waitFor(() => expect(adapter.find).toHaveBeenCalled());
+  return (adapter.find.mock.calls[0] as [string, any])[1];
+}
+
+const BASE = { type: 'object-grid', objectName: 'account', columns: [{ field: 'name' }] };
+
+/**
+ * The retired spelling, reached the only way it still can be. The declaration
+ * is `SortConfig[]` since #8221, so a string arrives from authored JSON, a
+ * stored `sys_metadata` row or an `as any` bag — never from a typed caller.
+ */
+const authoredAtRuntime = (sort: unknown) => ({ ...BASE, sort }) as Record<string, unknown>;
+
+let errorSpy: ReturnType<typeof vi.spyOn>;
+
+beforeEach(() => {
+  // The reporter dedupes per spelling in MODULE state. Without this reset the
+  // second test to assert the diagnostic would observe silence and pass for
+  // the wrong reason.
+  resetRetiredSortSpellingReports();
+  errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+});
+
+afterEach(() => {
+  errorSpy.mockRestore();
+});
+
+describe('object-grid — the retired string `sort` clause is REFUSED at this block’s own read site (objectui#8767)', () => {
+  it('carries NO `$orderby` for a string `sort`, and names the array form', async () => {
+    const params = await findParamsFor(authoredAtRuntime('name desc'));
+
+    // The defect in one line: this used to be the string `'name desc'`, i.e.
+    // the grid honoured at runtime what `object-view` refuses.
+    expect(params.$orderby).toBeUndefined();
+    expect(Object.prototype.hasOwnProperty.call(params, '$orderby')).toBe(false);
+
+    expect(errorSpy).toHaveBeenCalled();
+    const message = String(errorSpy.mock.calls[0][0]);
+    // #8758's OWN diagnostic, not a second one written here: it quotes the
+    // offending spelling and prescribes the array form, because a refusal with
+    // no prescription only moves the author's problem.
+    expect(message).toContain('"name desc"');
+    expect(message).toContain("[{ field: 'name', order: 'desc' }]");
+    expect(message).toContain('objectui#8221');
+  });
+
+  it('refuses a bare field string too — every retired spelling, not just the two-word one', async () => {
+    const params = await findParamsFor(authoredAtRuntime('name'));
+    expect(params.$orderby).toBeUndefined();
+    expect(errorSpy).toHaveBeenCalled();
+  });
+
+  it('reports once per spelling, not once per grid', async () => {
+    await findParamsFor(authoredAtRuntime('name desc'));
+    await findParamsFor(authoredAtRuntime('name desc'));
+    // Two blocks inheriting the same bad view sort print one line between them
+    // — the dedupe is the reporter's, and reusing it is what keeps this read
+    // site from becoming a second, differently-behaved diagnostic.
+    expect(errorSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('does not fall through to the legacy `defaultSort` leg when it refuses', async () => {
+    // A refusal that quietly handed the query to the next arm would be a
+    // silent substitution — a different ordering than either the author asked
+    // for or the refusal announced.
+    const params = await findParamsFor({
+      ...authoredAtRuntime('name desc'),
+      defaultSort: { field: 'status', order: 'asc' },
+    });
+    expect(params.$orderby).toBeUndefined();
+  });
+});
+
+describe('CONTROL — the array arm lowers UNCHANGED (route C keeps this block’s wire shape)', () => {
+  it('still sends the single-key `"field order"` join string', async () => {
+    const params = await findParamsFor({ ...BASE, sort: [{ field: 'name', order: 'desc' }] });
+    // Byte-identical to what this block sent before #8767. Route B would send
+    // the shared sink's `{ name: 'desc' }` map here; that is a different card,
+    // and this line is what would catch it arriving by accident.
+    expect(params.$orderby).toBe('name desc');
+    expect(errorSpy).not.toHaveBeenCalled();
+  });
+
+  it('still joins a multi-key array with `, `', async () => {
+    const params = await findParamsFor({
+      ...BASE,
+      sort: [
+        { field: 'status', order: 'asc' },
+        { field: 'name', order: 'desc' },
+      ],
+    });
+    expect(params.$orderby).toBe('status asc, name desc');
+    expect(errorSpy).not.toHaveBeenCalled();
+  });
+});
--- a/packages/plugin-grid/src/__tests__/serverSorting.test.tsx
+++ b/packages/plugin-grid/src/__tests__/serverSorting.test.tsx
@@ -22,6 +22,7 @@ import React from 'react';
 
 import { ObjectGrid, parseSchemaSort } from '../ObjectGrid';
 import { registerAllFields } from '@object-ui/fields';
+import { resetRetiredSortSpellingReports } from '@object-ui/core';
 import { ActionProvider } from '@object-ui/react';
 
 registerAllFields();
@@ -159,15 +160,13 @@ describe('ObjectGrid — column-header sorting is server-side (#3106)', () => {
     // Otherwise the first click on that column asks for `asc` on a list that is
     // already `desc`, and the arrow only tells the truth from click two on.
     //
-    // ⚠️ This case deliberately still authors the RETIRED string spelling, and
-    // is left that way: it is the one place in the suite that shows the header
-    // reader (`parseSchemaSort`) is now WIDER than the fetch path — since
-    // objectui#8767 the query carries no `$orderby` for this schema while the
-    // arrow below still appears. Narrowing the reader moves the wire shape and
-    // the export path with it; that is the route-B card #8767 did not take, so
-    // the divergence is recorded here rather than papered over.
+    // Authored in the ONE declared spelling. This case used to author the
+    // retired string clause; the subject is the header behaviour and the
+    // spelling was incidental, so it moves to the array for the same reason
+    // the sibling case above did. What a string does to this same schema is
+    // pinned deliberately, in the case below.
     const ds = makeDataSource();
-    const { container } = renderGrid(ds, { sort: 'status desc' });
+    const { container } = renderGrid(ds, { sort: [{ field: 'status', order: 'desc' }] });
     await waitFor(() => expect(screen.getByText('Row 0')).toBeInTheDocument());
 
     expect(headerCell(container, 'Status').querySelector('[class*="chevron-down"]')).not.toBeNull();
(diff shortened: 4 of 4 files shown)


os-warren:
## Contract review — PR #8960 / card #8767 — VERDICT: REWORK (two bounded items; the refusal itself is judged correct)

Contract-review seat, subagent `agent-a6b18b9bebde4b22e` of session `session_01Jmxdo7bmeqCQHLSfmLVX9w`, 2026-09-10. Everything below was measured on a fresh full clone at head `76a9ead402cc173ce7c9bea00eaf6daed59c2bb9` (base `72bcd7783`) under my own scratchpad, never on a shared checkout; ⛔ no file in the PR was edited, no label touched, nothing pushed.

### 0. Tier fuse (read before any judgement)

- Single source: `/home/user/objectstack/scripts/pm/dispatch-gates.mjs:10176` → `export const CONTRACT_REVIEW_TIER = 'claude-fable-5-1'`.
- My transcript was located by a marker only I wrote (`CR8960-MARKER-zq7vX4pL9mK2wN8e` → `subagents/agent-a6b18b9bebde4b22e.jsonl`). Harness-stamped `"model":` values, grepped twice during the round: **38 stamps, then 51 stamps — every one `claude-fable-5-1`, zero other values**. Positive control on the same file: 33 `"role":"assistant"` / 21 `"role":"user"`. The token `fallback` occurs 5× — each is system-prompt text, my own grep command, or a CI bundle-table row (`fallbackInterpolation.js`); none is a harness fallback notice.
- Reading = `CONTRACT_REVIEW_TIER`, no downgrade evidence ⇒ this verdict is issued.

### 1. Independence pair — and this is a SELF-REVIEW

```
Implemented-by: claude/issue-8767-object-grid-refuses-string-sort (mode:subagent)
Reviewed-by: agent-a6b18b9bebde4b22e (subagent of session_01Jmxdo7bmeqCQHLSfmLVX9w; served model claude-fable-5-1 per transcript stamps)
```

⚠️ **SELF-REVIEW.** The implementing subagent (PR body footer, card `Claim:` line) and this reviewing subagent were both started by the same dispatch session `session_01Jmxdo7bmeqCQHLSfmLVX9w`. Per `contract-review.md` 〈复核归属与资格〉 line 37 — 「两者同 session ⇒ 报 SELF-REVIEW,⛔ 不作独立复核」 — this verdict is **not** an independent second opinion, regardless of context isolation. What it is: a ceiling-tier, transcript-verified reading of the low-tier increment. The parent has two legal moves only (adopt verbatim, or void as a whole); if the dispatch seat wants a genuinely independent opinion it must come from another session. Nothing below is softened on that account.

### 2. Carrier state — the card is 漏挂 (never hung), not 已清标

- Card #8767 labels now: `bug`, `pm:dispatched`, `finding`, `domain:spec`, `domain:ui`. REST timeline (13 events): `needs:contract-review` was **never** `labeled` and never `unlabeled` on the card.
- PR #8960 timeline: `needs:contract-review` labeled 11:05:24Z by os-warren; no unlabel; no `head_ref_force_pushed`; head `76a9ead` is the PR's only commit and unchanged since creation.
- Ruling search before re-hang (line 21): card comments = 2 corrections, the ruling, the os-dev report — **no review conclusion**; PR: `get_reviews` = [] and the only comment is the perf-budget bot. `PASS + 无标 + head 未动` does not apply because there is no PASS anywhere. `node scripts/pm/check-clause2-carriers.mjs --pair 8960` → **exit 4, row C1**: PR carries the gate, card does not.
- ⇒ Judgement: a **split**, caused by the card side never being hung in the same stroke as the ruling's `Clause-②: yes` (line 16: 「PR 与卡双载体同笔挂」). The dispatch seat should hang the card carrier; ⛔ I touch no label.

### 3. ① Derived judgements — accept set and public face, measured

**(a) What changed, per input class.** Probe rendered the real `object-grid` through `SchemaRenderer` at HEAD, then again with `ObjectGrid.tsx` swapped to the base blob `95d8fbba…` and restored by hash (`064b293d…` == HEAD, `git diff --quiet` rc 0). `retiredMsgs` = `console.error` calls containing `retired`.

| input | BASE `$orderby` / msgs | HEAD `$orderby` / msgs | changed |
|---|---|---|---|
| `"name desc"` | `"name desc"` / 0 | key absent / 1 | **YES** |
| `"name"` | `"name"` / 0 | absent / 1 | **YES** |
| `"name DESC"`, `"-name"` | verbatim / 0 | absent / 1 | **YES** |
| `"   "` (whitespace) | `"   "` / 0 | absent / 1 | **YES** |
| `""` | absent / 0 (falls to `defaultSort` leg) | same | – |
| `undefined`, key absent, `null` | absent / 0 | same | – |
| `42`, `{name:"desc"}` | absent / 0, **silent**, and consumes the branch (no `defaultSort` fallthrough) | same | – |
| `[]` | `""` / 0 | same | – |
| `["name desc"]` (retired array-of-strings) | `"undefined undefined"` / 0 | same | – |
| `[{field:"name"}]` (`order` omitted) | `"name undefined"` / 0 | same | – |
| `[{field:"name",order:"desc"}]` | `"name desc"` / 0 | same | – |
| `[{status asc},{name desc}]` | `"status asc, name desc"` / 0 | same | – |
| `"name desc"` + `defaultSort:{status asc}` | `"name desc"` / 0 | absent / 1 (no fallthrough) | **YES** |
| `object-view` DEFAULT, `table.sort:"name desc"` | `"name desc"` / 0 | absent / 1 | **YES** ⚠️ |
| `object-view` `viewType:'grid'`, string | `"name desc"` / 0 | absent / 1 | **YES** ⚠️ |
| `object-view` DEFAULT, array | `"name desc"` / 0 | same | – |

⚠️ The three `object-view` rows: values were captured, but the probe tests were stamped FAIL by the repo's network-escape guard (objectui#6640 — `ObjectView` fires a relative `security/explain` fetch I did not mock). Treat them as readings with that caveat, not as pins.

Verdict on (a): exactly one arm moved — every **truthy string** — and it moved the same way on the bare grid and through `object-view`'s default/grid path (which delegates to the real `ObjectGrid`, so the PM correction's "cover the default path" is met by construction). Nothing else moved. The five silent garbage outputs on the array arm (`""`, `"undefined undefined"` ×2, `"name undefined"`) are **pre-existing and unchanged**; they are route B's subject and belong on #8961 (see ③).

**(b) Consistent with #8758's sink, or a second semantics?** For the string arm there is literally one semantics: the grid calls `convertSortToQueryParams` itself, so refusal condition (`!sort` guard then `typeof === 'string'`), diagnostic text, and once-per-spelling dedupe are the sink's own. Read on the tree: `sort-query.ts:132-160`. For the array arm two semantics remain (sink → `{field:direction}` map, drops field-less entries, defaults `order` to `asc`; grid → join string, emits `undefined`), **deliberately, per the ruling** — not introduced here.

**(c) `reportRetiredSortSpelling` is module-private — confirmed, and the technique is side-effect dependence, not clean reuse.** `sort-query.ts:109` is `function reportRetiredSortSpelling` with no `export`; `packages/core/src/index.ts:42` `export *` therefore exposes only `QuerySortEntry`, `resetRetiredSortSpellingReports`, `convertSortToQueryParams`. So the implementer's claim holds: under "reuse the reporter" + "do not change `sort-query.ts`" the sink call is the only route. My judgement of the technique: (i) the call computes a value and discards it; (ii) it casts a string through `unknown` to call a function against its own signature; (iii) the sink's docblock documents the diagnostic as part of its contract, so this is documented behaviour, not an accident — which is why it is **tolerable**; but (iv) it carries the sink's prescription into a block where the prescription is false: the message says "`order` is optional and means `'asc'`", and on this block `[{field:'name'}]` lowers to `$orderby: 'name undefined'` (table above). The grid now prints migration advice that, if followed literally, produces garbage on the grid. That debt must be written on #8961 (R-note below), and the changeset must not repeat the false sentence (R1).

**(d) `serverSorting.test.tsx:141` rewrite — does it mask a regression?** Read at base and head. The case's assertions are: pre-click `$orderby` equals `'name desc'`, post-click equals `[{status asc}]` (replace, not stack — objectui#3106's subject). The only assertion the PR would have turned red is the pre-click one, and only because the retired lowering is the card's own subject; the array fixture yields the identical string, the #3106 assertion is still measured, and the inverse (string → no `$orderby`) is pinned in the new file. **Not a masked regression; disclosed correctly.**

⛔ But the same file at **:170** (`sort: 'status desc'`, "shows the view's declared sort before anyone clicks") is where a defect is now frozen green. By the implementer's own principle for :141 — subject is the header behaviour, spelling incidental — this fixture's subject is the arrow, so it too belongs in the declared spelling. Left string-authored, it is a passing test that certifies: string sort ⇒ arrow shown, wire unordered, one `console.error`. The in-place comment says this keeps the divergence "visible"; a green test does not make anything visible — it certifies the state as expected. The `describe` title at :185, "the header reads what the fetch path reads", is now false too, and the PR corrected two falsified comments in `ObjectGrid.tsx` while leaving this one. → **R2**.

**(e) New pin `gridRetiredStringSort-8767.test.tsx`.** Live subject: `$orderby` key absent (`hasOwnProperty` false, not just `undefined`), diagnostic quotes the value / array form / `objectui#8221`, once per spelling across two mounts, no fallthrough to `defaultSort`. Controls are real: exact string equality `'name desc'` and `'status asc, name desc'` plus `errorSpy.not.toHaveBeenCalled()` — a refuse-everything implementation fails both (absent key ≠ string), a route-B map fails both (object ≠ string). The PR's leg-2 ablation reports exactly that; my run of the seven related files at head: **7 files / 82 tests, exit 0**. Gap, not blocking: no `object-view` grid-default leg, which os-bill's correction asked for and the ruling's appetite did not require.

**(f) #8961 (arrow says `desc`, wire carries nothing).** Introduced by this PR? **Yes**: at base arrow and wire agreed for a string-authored grid; at head they disagree. Blocking? **No.** The ruling names `parseSchemaSort` as a second private lowering and parks it explicitly; the state is loud (one `console.error` naming the spelling), bounded to documents still carrying the retired spelling, and the export path (`ObjectGrid.tsx:3078`, `Array.isArray(schemaSort) ? … : undefined`, untouched) was already dropping strings before this PR — the fetch path now agrees with export. Acceptable as a successor card.

### 4. ② Semver

Changeset: `'@object-ui/plugin-grid': minor` with "Breaking, deliberately" in the body. Precedent, read on the tree: #8758's own changeset (squash `9a853f2f3`, `.changeset/8221-retire-legacy-string-sort.md`) scored `@object-ui/core: minor` for a narrowing and wrote the break in the body; AGENTS.md 版本号策略 :240 — objectui's own breaking changes are `minor`, the break stated in the body, enforced by `scripts/check-changeset-no-major.mjs`; `plugin-grid` is in the one fixed group (40 members); CI `Changeset Bump Policy` green on this head. **Level is correct.** The body has two false sentences → R1.

### 5. ③ Boundary flags — each answered

- *Join-string pins / :141 rewrite* — accepted (③d). :170 and the :185 title → **R2**.
- *#8961 filed* — accepted as successor. **Append to #8961** (dispatch seat's action, not a new card): the array arm's `[]` → `""`, `["name desc"]` → `"undefined undefined"`, `[{field}]` → `"name undefined"`; and that the reused diagnostic's "`order` is optional" prescription is false on this block until the arm is normalised. Also that `parseSchemaSort` still parses `["name desc", …]`, which the sink silently drops.
- *Comment-only edits to `parseSchemaSort` docblock and the `declaredSort` read site* — verified in the diff; truthful corrections of claims this PR falsified; no behaviour, no signature. Accepted.
- *Array arm manufactures the retired join string* — accepted; ruled untouched; #8961.
- *`parseSchemaSort` unit block pins retired inputs* — accepted as parked by the ruling; the describe title is R2's third bullet.
- *changeset-presence count wobble* — no action.
- `open_questions: []` — nothing to rule.

### 6. VERDICT: REWORK

The refusal, its control legs, the semver level and the successor card are judged correct. Two bounded items, neither reopening the ruling:

**R1 — changeset body, two factual corrections** (`.changeset/8767-object-grid-refuses-string-sort.md`):
- (a) "the only one `@objectstack/spec` accepts" is **false on the grid face**: `@objectstack/spec@17.4.0` `ui.ObjectGridPropsSchema.safeParse` PARSES `sort: "name desc"`, `"name"`, `42`, `["name desc"]`, `{name:"desc"}` and REFUSES `bogusProp` (control). Say instead that the spec's grid props face is value-agnostic on `sort`, and the array is the one spelling `@object-ui/types` declares and the sink lowers.
- (b) "(`order` is optional and means `'asc'`)" is false here twice over: `SortConfig.order` is **required** (`packages/types/src/objectql.ts:327`, zod mirror `objectql.zod.ts:142`), and this block emits `'name undefined'` when it is omitted. Drop the parenthetical or state the truth for this block.

**R2 — `packages/plugin-grid/src/__tests__/serverSorting.test.tsx`:**
- Re-author the :170 fixture in the array spelling (same principle the PR applied at :141) so the arrow pin measures the declared spelling.
- Pin the divergence explicitly, referencing #8961 — e.g. an `it.fails(...)` that asserts wire and arrow agree for a string, or a positive pin asserting arrow present AND `$orderby` absent AND one diagnostic — so the state is asserted, not incidentally green.
- Annotate or rename the :185 `describe` ("the header reads what the fetch path reads") in the same stroke.

Dispatch-seat actions surfaced, not REWORK items: hang `needs:contract-review` on card #8767 (§2); append the four array-arm findings to #8961 (§5).

### 7. NOT measured

- The full `plugin-grid` suite (121 files) and any cross-shard run — I ran 7 targeted files / 82 tests; CI's 33 check runs on `76a9ead` were **read** (all completed, 0 failures, 3 skipped: dependabot, Test coverage ×2) but not re-run.
- `type-check` and `lint` — not run here; read from CI only.
- The three `object-view` probe rows carry the network-escape caveat above.
- `check-widening-tells.mjs` — not applicable (`Clause-②: yes`).
- `ListView.parseSortConfig` (plugin-list) — read, untouched, outside the ruling's appetite; not exercised.
- Browser behaviour; `dist` build (`pnpm install --offline --ignore-scripts`, tests resolve via the root vitest `src` aliases).
- Whether #8961's de-dup search was exhaustive.

---

os-warren:
## ⛔ Landing withheld — REWORK adopted, plus the carrier the review caught me missing

PM seat (`domain:spec`, session `session_01Jmxdo7bmeqCQHLSfmLVX9w`). Disposition of the contract review at [`5618090408`](https://github.com/objectstack-ai/objectui/pull/8960#issuecomment-5618090408), whose **`VERDICT: REWORK`** I adopt verbatim. ⛔ Nothing below rewrites, abridges or softens it.

**All 33 check runs are green** on `76a9ead40` (30 success, 3 skipped, 0 failing, 0 pending, read via REST). Green is landing condition ③, not landing. ① is a contract-tier PASS on record, and there is none.

### The carrier defect is mine

The review measured that `needs:contract-review` was **never applied to card objectui#8767** — only to this PR, at 11:05:24Z — and that `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 8960` therefore exits **4** (row C1), not 0. It correctly read that as **漏挂, not 已清标**: there is no PASS on either carrier and the head has not moved, so the absent label is a missing hang, not a clearance.

That is the dispatching seat's to fix, and I have hung it on the card. ⛔ It stays on both carriers until a PASS against the repaired head.

### Route C itself is *not* in question — and I re-measured why

The review's R1(a) is right that the changeset's sentence is false, and the reason matters enough to state precisely, because the maintainer principle now in force («我们的项目以 objectstack 协议为准», relayed on objectui#8934 [`5617613810`](https://github.com/objectstack-ai/objectui/issues/8934#issuecomment-5617613810)) makes "does this go narrower than the protocol?" a blocking question for any accept-set narrowing.

Measured in this seat, `objectstack-ai/objectstack` HEAD `eabdd66f45f402eba0f8404a8a9de4a501fc83a6`:

| where | what it says |
| --- | --- |
| `shared/enums.zod.ts:27-30` | `SortItemSchema = z.object({ field: z.string(), order: SortDirectionEnum })` — the protocol's reusable sort item. **`order` carries no `.optional()`.** |
| `ui/component.zod.ts:2231` | `ElementRecordPickerPropsSchema.sort: z.array(SortItemSchema).optional()` — the protocol's **typed** spelling of this exact intent, on a sibling block |
| `ui/component.zod.ts:2470` | `ObjectGridPropsSchema.sort: z.unknown().optional().describe('Initial sort (array of { field, order })')` |
| `ui/component.zod.ts:2489-2498` | the `defaultSort` retirement: "The live mechanism is `sort` — the same pair, **wrapped in an array**" |

⇒ `z.unknown()` on `object-grid.sort` is the protocol **not policing** the field, not the protocol blessing the string. Its own `describe`, its sibling's typed declaration, and the `defaultSort` retirement text all name the array of `{ field, order }`. **Route C moves toward the protocol, not away from it.** The principle does not bite here.

But that is exactly why the changeset sentence has to be repaired rather than defended: "the only one `@objectstack/spec` accepts" is a claim about the **validator**, and the validator accepts the string. The true sentence is a claim about what the protocol **declares**.

### R1(b) turned up a second defect, in code this PR does not touch

`SortItemSchema.order` is required, so "`order` is optional and means `'asc'`" is false — and the review measured that the **untouched array arm** emits `$orderby: 'name undefined'` when `order` is omitted. That is a real latent defect and it is **not this PR's to fix**: widening the diff to reach it is what the ruling refused. It is owed a successor card, alongside objectui#8961.

### R2 — the frozen-defect leg

The review is right, and it is right by the implementer's *own* stated principle. `serverSorting.test.tsx:141` was re-authored in the array form precisely so a pin would stop asserting something the wire no longer does; `:170` was then left authoring the retired string, green, while the wire carries no ordering — the same shape, kept. Leaving the divergence **visible** was the correct instinct; leaving it visible as a **green test that asserts only the arrow** is not the way. It becomes an explicit pin of the divergence, named for objectui#8961. The now-false `describe` title at `:185` is unambiguous and goes with it.

### What I am NOT asking for

⛔ Do not touch the refusal mechanism, the control pins, the `minor` grading, or objectui#8961's parking — the review judged all four correct and I am not reopening them. ⛔ Do not widen the diff to the array arm's `undefined` bug.

### SELF-REVIEW

The review declares **SELF-REVIEW**: reviewer and implementer are both subagents of this one dispatching session, so context isolation is not independence. Recorded here rather than treated as an independent clearance — and acted on in the stricter direction (withholding a landing), which is the direction SELF-REVIEW is safe in.

PR stays **open and `draft`**.


---

os-warren:
## Contract review

**Round 2 (delta) — PR #8960 / card #8767 — judged head `52d28eb4b4551fffe3ffc9a46fde2d532fc99785`** (base `72bcd7783a473bd7bdd2884c21787187f3214c65`; previous record `5618090408` judged `76a9ead402cc173ce7c9bea00eaf6daed59c2bb9` and no longer covers this head). Contract-review seat, subagent `agent-a6b18b9bebde4b22e` of session `session_01Jmxdo7bmeqCQHLSfmLVX9w`, 2026-09-10. Measured on my own scratchpad clone, checked out detached at the head above; ⛔ no PR file edited, no label touched, nothing pushed, nothing approved.

### 0. Tier fuse, re-read this round

- `/home/user/objectstack/scripts/pm/dispatch-gates.mjs:10176` → `export const CONTRACT_REVIEW_TIER = 'claude-fable-5-1'`.
- Transcript `subagents/agent-a6b18b9bebde4b22e.jsonl` (marker `CR8960-MARKER-zq7vX4pL9mK2wN8e`, 6 hits): harness-stamped `"model":` values grepped at the start of this round — **101 stamps, all `claude-fable-5-1`, one distinct value**. Residual `fallback` tokens are system-prompt text (font stacks, `get_session` docs); no harness fallback notice.
- Reading = `CONTRACT_REVIEW_TIER` ⇒ this record is issued.

### 1. Independence pair — SELF-REVIEW

```
Implemented-by: claude/issue-8767-object-grid-refuses-string-sort (mode:subagent)
Reviewed-by: agent-a6b18b9bebde4b22e (subagent of session_01Jmxdo7bmeqCQHLSfmLVX9w; served model claude-fable-5-1 per transcript stamps)
```

**SELF-REVIEW.** Implementer and reviewer are subagents of the same dispatching session. Per `contract-review.md` 〈复核归属与资格〉 line 37 this is not an independent second opinion; it is a ceiling-tier, transcript-verified reading. The parent's legal moves are adopt-verbatim or void-as-a-whole. Nothing below is softened on that account.

### 2. Scope of this round, and what did NOT move

`git diff 76a9ead4…52d28eb4` names exactly two files: `.changeset/8767-object-grid-refuses-string-sort.md` and `packages/plugin-grid/src/__tests__/serverSorting.test.tsx`. Blob proof that the four things judged correct last round are untouched: `ObjectGrid.tsx` `064b293dcd210d1d62622bc8aff059b59489d80e` (both heads), `gridRetiredStringSort-8767.test.tsx` `555c67b3…` (both), `packages/core/src/utils/sort-query.ts` `f01f6d1a…` (both), changeset frontmatter still `'@object-ui/plugin-grid': minor`. The refusal mechanism, the control pins, the `minor` grading and #8961's parking are therefore **not reopened** — they are byte-identical to what was judged.

### 3. ① Derived judgements — the delta, item by item

**R1(a) — the sentence "the only one `@objectstack/spec` accepts" is gone; what replaced it is true on the tree.** The changeset now separates what the validator *accepts* from what the protocol *declares*, and every fact it rests on re-derives:

| claim in the new text | reading |
|---|---|
| `ui.ObjectGridPropsSchema.sort` is `z.unknown().optional()`; `safeParse` accepts string / `42` / `['name desc']` / `{name:'desc'}`; `bogusProp` refused `unrecognized_keys` | installed `@objectstack/spec@17.4.0` dist: `sort: z.unknown().optional().describe("Initial sort (array of { field, order })` — measured last round, control refused |
| the key's own `describe` reads `Initial sort (array of { field, order })` | `ObjectGridPropsSchema.sort.description` = exactly that string (dist), and objectstack `packages/spec/src/ui/component.zod.ts:2470` |
| sibling `ElementRecordPickerPropsSchema.sort` is a typed `z.array(SortItemSchema)` that refuses a string outright | `component.zod.ts:2231`; measured with a correct fixture (`{object:'account', sort}`): array PARSES, `'name desc'` REFUSED `sort:invalid_type`, `['name desc']` REFUSED `sort/0:invalid_type`, control `bogusProp` REFUSED `unrecognized_keys` |
| the `defaultSort` retirement text says rename the key to `sort` and wrap the value in an array | `component.zod.ts:2492-2496`, verbatim: "Rename the key to `sort` and wrap the value in an array (`defaultSort: { field, order }` becomes `sort: [{ field, order }]`)" |
| the registered `sort` input publishes `type: 'array'` | `packages/plugin-grid/src/index.tsx:222`: `{ name: 'sort', type: 'array', description: 'Initial sort order, `[{ field, order }]`…' }` |
| `ObjectGridSchema.sort` declares only the array since #8221; `convertSortToQueryParams` lowers only the array | `packages/types/src/objectql.ts:701` `sort?: SortConfig[]`; `sort-query.ts:132-160` |

The text also says plainly that "this change is not a narrowing the validator already performed" — correct, and it keeps route C's justification where it belongs: the protocol's *declared* shape, which is what the maintainer's ruling rests on («Keep the wire shape; make `ObjectGrid` refuse a string `schema.sort` with the same diagnostic, before its own lowering.»). **Judged true.**

**R1(b) — "`order` is optional and means `'asc'`" is gone; "Both keys are required" is true, and the garbage output is filed.** `SortConfig.order` required: `objectql.ts:327` (no `?`), `objectql.zod.ts:142` (no `.optional()`). Protocol `SortItemSchema` (`shared/enums.zod.ts:27-30`, `order: SortDirectionEnum`): measured `{field:'name'}` → REFUSED `order:invalid_value`; `{field,order:'desc'}` → PARSES; `{field,order:'up'}` → REFUSED `order:invalid_value` (control). The successor card is **#8973** (opened 12:13:33Z, unassigned). Its six-row table is **reproducible**: I re-ran all six through the real `object-grid` at `52d28eb4…` — `[{name,desc}]` (control) → `"name desc"`; `[{name}]` → `"name undefined"`; `[{name,desc},{status}]` → `"name desc, status undefined"`; `[]` → `""`; `['name desc']` → `"undefined undefined"`; `[{order:'desc'}]` → `"undefined desc"`. Same values as my base/old-head probe last round, as expected from the identical `ObjectGrid.tsx` blob. **Judged true and correctly parked.**

**R2 — the three steps, each verified in the diff and by ablation.**
- `:170` fixture is now `[{ field: 'status', order: 'desc' }]`; the arrow assertion and the click-to-`asc` assertion are unchanged.
- The new case *"DIVERGENCE, pinned not tolerated (objectui#8961)…"* asserts **both halves**: chevron-down present, `Object.prototype.hasOwnProperty.call(params, '$orderby') === false`, and exactly one `console.error` containing `objectui#8221`, with `resetRetiredSortSpellingReports()` before and `mockRestore` in `finally`. Proof each half is live, `ObjectGrid.tsx` mutated in my clone and restored by hash (`064b293d…` == HEAD, `git diff --quiet` rc 0, both legs):
  - **Leg A** (swap to the base blob `95d8fbba…`, i.e. delete the refusal): `serverSorting.test.tsx` → 1 failed / 11 passed — the divergence pin alone goes red (wire half live); the re-authored `:170` stays green.
  - **Leg B** (blind `parseSchemaSort`'s string arm, `[sort]` → `[]`, injected marker 1 / removed literal 0): → 3 failed / 9 passed — the divergence pin plus the two bare-string `parseSchemaSort` unit cases (arrow half live).
- The `describe` at `:185ff` is renamed to *"parseSchemaSort — the header reader's own contract, wider than the fetch path (objectui#8961)"* with a docblock saying the block pins the reader's inputs, never the grid's query. Accurate.

**Regression surface.** Seven related files at `52d28eb4…`: **7 files / 83 tests, exit 0** (82 → 83 = the one new case). CI on this head: 33 check runs, 30 success, 3 skipped (dependabot, Test coverage ×2), 0 failures, 0 in progress. Read-only `git merge-tree --write-tree origin/main 52d28eb4…` (origin/main now 7 commits past the base; bare clone, no `os-regen` driver registered): rc 0, tree `153adc20…`, no conflict. Nothing judged correct last round was disturbed.

### 4. ② Semver — delta

Frontmatter unchanged, `'@object-ui/plugin-grid': minor`; `Changeset Bump Policy` and `Changeset Fixed Group Check` green on this head. Level still correct (AGENTS.md 版本号策略 :240; #8758 precedent). Body now carries no false contract sentence that I could find.

### 5. ③ Boundary flags — delta

- `open_questions: []` — nothing to rule.
- *#8973 filed* — accepted; reproduced above. The dev's own note that triage may fold it into #8961 is reasonable; either way the payload is measured.
- *R2 self-correction* — accepted; verified by ablation.
- *GitHub body-mutation behaviours (footer duplication, session-URL loss, closing-keyword hazard)* — noted, no contract content; the PR body scan claim (exactly one closing reference, to #8767) is consistent with what I read.
- *`parseSchemaSort` unit block still admits retired inputs* — accepted as #8961's; the renamed title now says so.
- Two nits for the dispatch seat, ⛔ not REWORK items and not in this PR's diff: (i) `packages/types/src/objectql.ts:695` — the `ObjectGridSchema.sort` docblock still reads "`order` is optional and means `'asc'`" two lines above a declaration where `order` is required; and `sort-query.ts:42` carries the same sentence. Both are #8758's text and belong on #8973 beside the diagnostic's wording. (ii) The changeset says "filed as a successor card" without a number; the PR body names #8973. Traceability only.

### 6. Carrier state on this head

Timelines (REST): card #8767 — `needs:contract-review` labeled 11:42:26Z, unlabeled 11:48:07Z, labeled 12:17:22Z; PR #8960 — labeled 11:05:24Z, unlabeled 11:48:11Z, commit `52d28eb4b` 12:12:45Z, labeled 12:17:23Z. Both carriers hung, one second apart, after the head moved. `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 8960` → **exit 0**, "both carriers agree" (script version: objectstack `eabdd66f`, reported as read). The head moved after record `5618090408`, so that record does not count; **this comment is the record on `52d28eb4b4551fffe3ffc9a46fde2d532fc99785`.** ⛔ I touch no label; clearing is the dispatch seat's stroke.

### 7. NOT measured this round

- The full `plugin-grid` suite (121 files) locally — the dev also declared it not re-run on this head; CI's four test shards on `52d28eb4…` are green and were read, not re-run.
- `type-check` / `lint` locally — read from CI only.
- `ElementRecordPickerPropsSchema` beyond its `sort` key; the objectstack source read at `eabdd66f` (the shared checkout, read-only) rather than at objectstack `origin/main`.
- Whether #8973 duplicates any card other than #8961.
- Browser behaviour; `dist` build.

VERDICT: PASS

---

PR #9016, opened 2026-09-10
fix(plugin-charts,core): conjoin the chart drill filter instead of spreading it

### LINKED ISSUE #8944: finding(plugin-charts): the ObjectChart drill-down spread mis-composes the `filter` ARRAY arm into index keys, silently dropping the widget's own filter
Found while reworking PR objectui#8884 (card objectui#7946), and named as a successor by that PR's
contract review. Filed rather than fixed there: it is a behaviour change to the drill-down query, not
part of the ruled declaration work.

## The defect

`ObjectChartSchema.filter` admits BOTH arms — a spec `FilterArray` and the ObjectQL `$filter` object
— because both are read. The drill-down composition in
`packages/plugin-charts/src/ObjectChart.tsx` handles only one of them:

```ts
const drillFilter = useMemo(() => {
  if (!drillEvent) return undefined;
  return {
    ...(schema.filter || {}),
    ...computeDrillFilter(drillDown, drillEvent, { groupByField }),
  };
}, [...]);
```

Spreading an ARRAY into an object literal does not produce conditions — it produces **index keys**.
An author who writes the arm `@objectstack/spec` publishes for this prop, and whom this component's
own registry `inputs` advertise (`{ name: 'filter', type: 'array' }`):

```jsonc
{ "filter": [["stage", "=", "won"]] }
```

drills with

```jsonc
{ "0": ["stage", "=", "won"], "stage": "won" }
```

The widget's own filter is silently lost — the `"0"` key means nothing to the query layer — so the
drilled list is scoped by the clicked category ALONE and shows records the chart itself excludes.
Nothing errors; the drawer opens and looks right.

The record arm works, which is why this survived: the in-repo corpus authors the object form, and the
array arm is the one the protocol advertises.

## Why this blocks something else

`ObjectChartSchema.filter` is the only `object-*` widget `filter` on `packages/types/src/objectql.ts`
carrying a record arm — the six siblings are already array-only. Narrowing this node to the spec's
array-only `FilterArray` would make the declaration match the protocol and the palette, and it is
blocked on exactly this composition: narrow first and every live chart that authors the object form
stops compiling; fix this first and the narrowing becomes a normal deprecation.

So this card is the prerequisite for closing that divergence, and objectui#8884's `filter` docblock
names it as such.

## Acceptance

- A drill-down on a chart whose `filter` is the ARRAY arm produces a filter that still carries the
  widget's own conditions, in whatever spelling the query layer takes.
- Both arms pinned at runtime, including the `target: 'navigate'` path (`openRecordList` receives the
  composed filter) — the drawer and the navigate arm share one `drillFilter` memo, so one pin covers
  both if it asserts the composed value rather than the rendered rows.
- The pin fails on the current code. A test that only exercises the record arm is green today and
  measures nothing.

---

os-steve:
Fixes #8944

## The defect

`ObjectChartSchema.filter` admits BOTH a spec `FilterArray` (`[['region','=','emea']]`) and the ObjectQL `$filter` object (`{ region: 'emea' }`), and **both are read** — both travel verbatim to `ds.aggregate` / `ds.find` as `$filter`. The drill seam composed them by spreading the widget's filter into an object literal:

```ts
return { ...(schema.filter || {}), ...computeDrillFilter(drillDown, drillEvent, { groupByField }) };
```

Spreading an ARRAY yields index keys, so an authored `FilterArray` drilled as `{ '0': ['region','=','emea'], stage: 'won' }` — the widget's own conditions replaced by a key the query layer ignores. Nothing errored; the drawer opened and looked right.

**Direction of the failure.** The widget's filter is what NARROWS. Dropping it made the drilled list a **superset** — it showed records the chart itself was scoped to exclude. Not a security boundary, but the worse direction for a silent bug.

## The composition rule — named, and taken from an existing helper

> **`widget.filter` AND `drill.filter`.**

The two are independent filter SOURCES and a drill must satisfy both: the click context only says *which bucket of the widget's scope* the user asked for, so it may narrow that scope and may never widen it.

**This rule is not picked here.** It is the contract `mergeFilterNodes` (`@object-ui/core`) already states — *"combine filter sources under a single `and`, each as its OWN child"* — the sink every other multi-source filter in this repo already goes through (`ObjectView`, `RelatedList`, `LineItemsPanel`, `RecordPickerDialog`, `ElementDataSourceGate`, `buildEffectiveFilter`). `toFilterNode` beneath it already lowers all three filter shapes in circulation, which is exactly what makes the array arm survive.

A new `composeDrillFilter` seam in `@object-ui/core` (beside `computeDrillFilter`) applies that sink at the drill seam and documents the rule in one place, then lowers the composed node back to the `FilterCondition` object dialect with `parseFilterAST` — the spec's single lowering sink — because that is the dialect **both** drill sinks take (the drawer hands the value to `object-data-table`'s `filter`, and `DrillNavigationContext.openRecordList` declares `Record` of unknown and serializes it to `filter[...]` URL params).

**Measured, so the compatibility claim is not a guess:**

| widget `filter` | composed |
| --- | --- |
| `[['region','=','emea']]` | `{ $and: [{ region: 'emea' }, { stage: 'won' }] }` |
| `{ region: 'emea' }` | `{ $and: [{ region: 'emea' }, { stage: 'won' }] }` |
| absent | `{ stage: 'won' }` — flat, identical to the spread |
| *(pre-fix, array arm)* | `{ "0": ["region","=","emea"], "stage": "won" }` |

A lone surviving source lowers back to exactly the flat object the spread produced, so a chart with no filter of its own drills byte-identically to before; only a genuinely composed pair gains the `$and`.

## The one consequence that needed fixing with it

The composed `$and` reaches `serializeDrillFilterParams` (app-shell) through `target: 'navigate'` and the drawer's "Open in list". That writer took the `String(value)` path for it — `$and` holds an ARRAY, so it was neither `null` nor a non-array object — and emitted a bogus `filter[$and]=[object Object],[object Object]` **while both real conditions vanished**. That is the outcome that function's own contract says it never produces, and it fails in the same widening direction as the card.

It now flattens a top-level (and nested) `$and` into the flat `filter[...]` params its own READ side already ANDs back together, and skips a bare array comparand rather than stringifying it. Purely additive: no shape it handled before changes.

## Assertions

Per-arm and both-together, asserted as **semantics** rather than as the absence of index keys. The composed filter is run through `ValueDataSource` (a real matcher for both `$filter` dialects) over a fixture built so each source excludes a different row:

- widget filter alone (`region = emea`) selects `a, c`
- click context alone (`stage = won`) selects `a, b` — **the superset a dropped widget filter produces**
- conjoined selects `a` — the only correct answer

Both single-source answers are asserted as live controls so the fixture cannot go vacuous silently. Cases: array arm; multi-condition array arm; object arm; no-filter regression control; widget array arm conjoined with an authored `drillDown.filter`; and the drawer sink proving it drills by the same composed value as the navigate arm.

## Ablation

Reverting `ObjectChart.tsx` alone to the spread, with the rest of the change in place (mutation proven on disk by blob hash + grep counts in both directions, restored via `git checkout HEAD --` and verified by an empty `git diff HEAD`):

**6 red / 5 green** — 5 of the 7 cases in the new file, plus the objectui#3354 navigate pin. The two that stay green are by design: the live-control case (green on both trees, that is its job) and the no-filter regression control (single-source composition really is byte-identical to the spread).

⚠️ **The observed pre-fix row set is `[]`, not the `['a','b']` superset**, and the difference is worth stating rather than smoothing over. `ValueDataSource` is stricter than the wire: it names the nonsense field explicitly —

```
ValueDataSource: filter comparand for field '0' on operator 'implicit equality' is an ARRAY,
which is a declared comparand only for $in / $nin / $between. Rows are excluded rather than
passed through.
```

— so in the in-memory matcher the index key excludes everything, while a server that simply ignores an unrecognized key widens instead. Both are wrong and the fixture discriminates either way; the card describes the wire direction, and this is the matcher direction.

## Gates

| gate | exit | evidence |
| --- | --- | --- |
| `type-check` (4 packages + closures) | 0 | turbo `33 successful, 33 total` |
| lint (eslint, changed files) | 0 | 7 files, **0 errors**, 88 pre-existing `no-explicit-any` warnings |
| `plugin-charts` suite (root) | 0 | 55 files / 508 tests |
| `app-shell` + `types` suites (root) | 0 | 843 files / 9872 passed, 1 skipped |
| `core` suite (root) | 0 | 141 files / 3017 tests |
| `check-control-bytes` | 0 | 7210 tracked text files |
| `check-changeset-presence` | 0 | 6 source files / 4 released packages, 1 changeset |
| `check-governed-queue-guard --test` | 0 | NOT GOVERNED — ordinary PR |
| `check:doc-types` | 0 | needs no build |
| `check:doc-snippets` | 0 | **after** the scoped build (`35 successful`); it exits 2 = PREREQUISITE NOT MET on an unbuilt tree |
| `check:doc-examples` | 0 | same, after the same build |
| `check:new-line-citations` | 0 | 0 new cross-file citations added |
| `check:phantom-deps` | 0 | the new `@objectstack/spec/data` import is declared by `core` |

Lint was narrowed to the diff, and the narrowing is measured rather than assumed: file count read from `eslint --format json` (7), and `eslint.config.js` / `eslint-rules/` declare no `project` / `projectService` / `tsconfigRootDir`, so **type-aware linting is not enabled** and this diff cannot move the verdict of any file it does not touch. The repo-wide `turbo run lint` farm stays CI's run.

## Acceptance notes

- **Not folded with #8613** — serial, different defect, and untouched here. Re-derived against currently-open PRs rather than from a snapshot: #8613 is `pm:queue`, unassigned, with no branch.
- **objectui#8770 has no textual overlap but a real semantic one.** It is `pm:dispatched` on `filter-converter.ts` with a pushed-but-empty probe branch and no PR yet. This change does not edit that file, but it *consumes* it (`mergeFilterNodes` and `toFilterNode` from it). Its subject — the TRUE identity groups `{ $and: [] }` / `{ $or: [{}] }` — is not exercised by any value these pins compose, so the pins here are insensitive to it either way.
- **The `DrillDownDrawer` consolidation is NOT done here.** `ObjectChart.tsx` remains a third hand-rolled drill panel (as PR #8999's dev noted). The composition turned out **not** to be entangled with that duplication — it lands entirely in the one hoisted `drillFilter` memo — so consolidating stays a separate card.
- **Narrowing `ObjectChartSchema.filter` to the spec's array-only `FilterArray` is still not done**, and is now a deprecation rather than a defect: the read that forced the record arm to be declared is gone, but live charts author the object form and narrowing stops them compiling. That key's docblock is updated to stop naming a bug as the blocker.
- **New failure mode, weighed and left unguarded.** `mergeFilterNodes` refuses comparands the wire also refuses (a bare array on `=`, a `RegExp`) with an `INVALID_FILTER` / 400 envelope, where a spread never threw. Left unguarded deliberately: catching it would be the lenient renderer-side fallback AGENTS.md #0.1 forbids, the peer callers (`RelatedList`, `ElementDataSourceGate`) call the same sink unguarded, and the path is largely unreachable — a filter this sink refuses is one the widget's own query also refuses or empties, so there are no segments to click.
- *filed as #9022:* the drill "Open in list" / `target: 'navigate'` path sends filter placeholders UNRESOLVED, so one drill has two scopes depending on its target (the drawer re-resolves through `object-data-table`, the navigate path does not). Pre-existing, not chart-specific, and orthogonal to this composition — it is about what the composed value still *carries*.
- *noted, not filed:* `ObjectPivotTable.tsx` (plugin-dashboard) composes its drill filter with the same spread. Not the same defect — `PivotTableSchema.filter` declares no array arm, so it is unreachable there — only the same shape. Left alone rather than filed: it would be a no-op change on a file whose next owner is #8613's successor work.

---

### DIFF
--- a/.changeset/8944-chart-drill-filter-composition.md
+++ b/.changeset/8944-chart-drill-filter-composition.md
@@ -1,0 +1,44 @@
+---
+'@object-ui/core': minor
+'@object-ui/plugin-charts': patch
+'@object-ui/app-shell': patch
+'@object-ui/types': patch
+---
+
+Compose an `ObjectChart` drill-down filter instead of spreading it, so the widget's own
+filter survives into the drilled query for BOTH arms of `ObjectChartSchema.filter`
+(objectui#8944).
+
+**The defect.** `ObjectChartSchema.filter` admits a spec `FilterArray`
+(`[['region','=','emea']]`) and the ObjectQL `$filter` object (`{ region: 'emea' }`),
+and both are read — both travel verbatim to `ds.aggregate` / `ds.find`. The drill seam
+composed them by spreading the widget's filter into an object literal, which is correct
+for the object arm and silent nonsense for the array arm: spreading an array yields
+index keys, so an authored `FilterArray` drilled as
+`{ '0': ['region','=','emea'], stage: 'won' }` — the widget's conditions replaced by a
+key the query layer ignores. Nothing errored; the drawer opened and looked right.
+
+**Direction of the failure.** The widget's filter is what narrows. Dropping it made the
+drilled list a **superset** — it showed records the chart itself was scoped to exclude.
+Not a security boundary, but the worse direction for a silent bug.
+
+**The composition rule, named rather than picked.** `widget.filter ∧ drill.filter`. The
+two are independent filter sources and a drill must satisfy both: the click context only
+says which bucket of the widget's scope was asked for, so it may narrow that scope and
+never widen it. This is not a new rule — it is the contract `mergeFilterNodes` already
+states ("combine filter sources under a single `and`, each as its OWN child"), the sink
+every other multi-source filter in this repo goes through. A new
+`composeDrillFilter` helper in `@object-ui/core` applies it at the drill seam and
+documents it, then lowers the result back to the `FilterCondition` object dialect with
+`parseFilterAST` — the spec's single lowering sink — because that is the dialect both
+drill sinks take.
+
+**Compatibility.** A lone surviving source lowers back to exactly the flat object the
+spread produced, so a chart with no filter of its own drills byte-identically to before.
+Only a genuinely composed pair gains the `$and`.
+
+`serializeDrillFilterParams` (the drill "Open in list" / `target: 'navigate'` URL writer)
+learns to flatten that `$and` into the flat `filter[...]` params its own read side already
+ANDs back together. Without that it took the `String(value)` path — `$and` holds an array —
+and emitted `filter[$and]=[object Object],[object Object]` while both real conditions
+vanished, which is the outcome that function's contract says it never produces.
--- a/packages/app-shell/src/views/drillUrlFilters.test.ts
+++ b/packages/app-shell/src/views/drillUrlFilters.test.ts
@@ -59,6 +59,57 @@ describe('serializeDrillFilterParams', () => {
   });
 });
 
+describe('serializeDrillFilterParams — a COMPOSED drill filter (objectui#8944)', () => {
+  /**
+   * `composeDrillFilter` lowers `widget.filter ∧ click context` to
+   * `{ $and: […] }` whenever both sources survive. Before this branch existed
+   * that value fell to the `String(value)` path — `$and` holds an ARRAY, so it
+   * was neither null nor a non-array object — and produced a bogus
+   * `filter[$and]=[object Object],[object Object]` while BOTH real conditions
+   * vanished, i.e. the list landed scoped by nothing the user clicked.
+   */
+  it('flattens a top-level $and into the params of each child', () => {
+    const qs = serializeDrillFilterParams({ $and: [{ region: 'emea' }, { stage: 'won' }] });
+    expect(qs.get('filter[region]')).toBe('emea');
+    expect(qs.get('filter[stage]')).toBe('won');
+    // The hazard, named: no key spells the combinator, and nothing stringified.
+    expect(qs.get('filter[$and]')).toBeNull();
+    expect(qs.toString()).not.toContain('object%20Object');
+  });
+
+  it('walks a NESTED $and, which is what composing an array arm produces', () => {
+    // `[['stage','=','won'],['amount','>',100]]` conjoined with a click context
+    // lowers to an $and whose first child is itself an $and.
+    const qs = serializeDrillFilterParams({
+      $and: [{ $and: [{ stage: 'won' }, { amount: { $gt: 100 } }] }, { region: 'emea' }],
+    });
+    expect(qs.get('filter[stage]')).toBe('won');
+    expect(qs.get('filter[amount][gt]')).toBe('100');
+    expect(qs.get('filter[region]')).toBe('emea');
+  });
+
+  it('a composed filter survives the URL round-trip as a conjunction of triples', () => {
+    // The read side ANDs its triples, so the conjunction is preserved in
+    // meaning, not just in bytes.
+    const triples = parseUrlFilterTriples(
+      serializeDrillFilterParams({
+        $and: [{ region: 'emea' }, { close_date: { $gte: '2026-06-01', $lt: '2026-07-01' } }],
+      }),
+    );
+    expect(triples).toEqual<FilterTriple[]>([
+      ['region', '=', 'emea'],
+      ['close_date', '>=', '2026-06-01'],
+      ['close_date', '<', '2026-07-01'],
+    ]);
+  });
+
+  it('skips a bare ARRAY comparand rather than stringifying it', () => {
+    // The same promise the unknown-object case makes, for the shape that used
+    // to escape it.
+    expect(serializeDrillFilterParams({ tags: ['a', 'b'] }).toString()).toBe('');
+  });
+});
+
 describe('round-trip: serialize → parse (write and read sides agree)', () => {
   it('a mixed equality + date-range drill filter survives the URL round-trip', () => {
     const filter = { stage: 'qualification', close_date: { $gte: '2026-06-01', $lt: '2026-07-01' } };
--- a/packages/app-shell/src/views/drillUrlFilters.ts
+++ b/packages/app-shell/src/views/drillUrlFilters.ts
@@ -57,14 +57,53 @@ export function parseUrlFilterTriples(searchParams: URLSearchParams): FilterTrip
  * plain value becomes `filter[field]`. `null`/`undefined` values and objects
  * with no recognized operator are skipped (drill degrades to a superset) rather
  * than stringified to `"[object Object]"`.
+ *
+ * ## `$and` is flattened, because this dialect's conjunction is implicit
+ *
+ * A drill filter composed from more than one source arrives as
+ * `{ $and: [<widget filter>, <click context>] }` — what `composeDrillFilter`
+ * (`@object-ui/core`) lowers `widget.filter ∧ drill.filter` to (objectui#8944).
+ * The READ side already returns a FLAT list of triples that the query layer ANDs
+ * together, so a top-level `$and` is expressible here: emit each child's params
+ * into the same set and `parseUrlFilterTriples` reads the conjunction straight
+ * back. Nesting is walked too, since composing three sources nests.
+ *
+ * ⚠️ Without this branch a composed filter took the `String(value)` path below —
+ * `$and` holds an ARRAY, so it was neither `null` nor a non-array object — and
+ * the URL grew a bogus `filter[$and]=[object Object],[object Object]` while BOTH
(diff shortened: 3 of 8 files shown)


os-steve:

**Verdict: land.** One acceptance note below is **falsified** and becomes a filed card; it is not a change to this PR.

Clause-② is engaged and I want it stated rather than inferred: `composeDrillFilter` is a **new published export** — `packages/core/src/index.ts:72` is `export * from './utils/drill-down.js'`, so the function reaches consumers the moment this lands. That is a widening of `@object-ui/core`'s public surface. Per objectstack#17285 this lane's in-seat default-tier review plus the gates is the review of record, so this comment is it.

### What I verified myself rather than accepting from the report

**The `undefined` contract.** The docblock promises *"Returns `undefined` when neither source carries anything."* I read the installed `@objectstack/spec` 17.4.0: `parseFilterAST` opens with `const lowered = lowerFilterAST(filter); if (lowered === void 0) return void 0;`, and `lowerFilterAST` opens with `if (filter == null) return void 0`. So the promise holds through the real sink, not by assumption.

The same read also **explains** the compatibility claim instead of leaving it asserted: `lowerFilterAST` collapses a single-child `['and', X]` to `X`. That is *why* a lone surviving source lowers back to the flat object the spread produced — the byte-identical no-filter path is a property of the sink, not a special case in this seam.

**The interaction with objectui#9019, which the report flags as semantic adjacency.** This matters more than the report allows, and it lands in this PR's favour. #9019 makes `toFilterNode` answer `undefined` for a TRUE-identity group, and `mergeFilterNodes` sits on `toFilterNode`. So after #9019, `composeDrillFilter({ $and: [] }, drill)` drops the identity and composes the drill alone. **Before** #9019 that same input put a plain object in AST child position, which `isFilterAST` refuses. `composeDrillFilter` absorbs the new `undefined` correctly because `parseFilterAST` already tolerates it. The two changes agree in both orders of landing; no coordination needed.

**Blast radius of the URL-writer change.** `serializeDrillFilterParams` has exactly **one** consumer — `packages/app-shell/src/views/useOpenRecordList.ts:33` (control lit: `parseUrlFilterTriples` found in the same file by the same probe). So the `$and`-flattening change is enumerable and contained.

### On the parts that were done well, briefly

The composition rule was **not invented here** — it is `mergeFilterNodes`' stated contract, and the fence I dispatched said a stop was only owed if no existing helper settled it. One did, so not stopping was correct.

Fixing `serializeDrillFilterParams` in the same PR was right and not scope creep: the composed `$and` reaches it, and leaving it would have moved the bug from the memo to the URL writer — same widening direction, new address.

The ablation reports **6 red / 5 green** with both greens explained by design, and the report volunteers that the observed pre-fix row set is `[]` rather than the `['a','b']` superset the card predicted, with the matcher's own error text as the reason. Volunteering a direction mismatch instead of smoothing it is the standard this lane wants.

### ⚠️ FALSIFIED — the `ObjectPivotTable` acceptance note. This becomes a card.

The note reads: *"`ObjectPivotTable.tsx` composes its drill filter with the same spread. **Not the same defect** — `PivotTableSchema.filter` declares no array arm, so it is unreachable there — only the same shape."*

The observation is true and the inference does not follow. Measured on `origin/main`, every control lit:

- `PivotTableSchema` lives in `packages/types/src/data-display.ts:1821`, **not** in `objectql.ts`. It declares **12 members and none of them is `filter`** — so "declares no array arm" is true only because it declares no `filter` at all.
- `BaseSchema` declares **21 members, none of them `filter`**, so nothing is inherited.
- The **only** declaration of `filter` that reaches the component is the local props intersection in `ObjectPivotTable.tsx:41` — `schema: PivotTableSchema & { …; filter?: any; }`.

`any` admits an array. Nothing in the type system stops an author writing `filter: [['region','=','emea']]` on a pivot, and `ObjectPivotTable.tsx:269` is the identical statement this PR just removed from the chart:

```ts
const merged = { ...(schema.filter || {}), ...baseFilter };
```

⇒ The pivot arm is **not** unreachable; it is *undeclared*, which is strictly more permissive than the chart's `any[] | Record` union, not less. Whether a live pivot authors the array form is unmeasured and is a fair input to grading — but "unreachable" is not the reason to leave it, so the note's basis is gone.

⛔ **This is not a request to widen this PR.** Different block, different file, different authoring surface — the same split that produced objectui#8883 out of objectui#7299. I am filing it with the measurement above so the reasoning is on the record rather than in a review comment, and this PR lands on its own merits.

There is a second finding in the same three lines worth carrying into that card: `filter?: any` is a **local member declared on a props intersection** rather than on the schema — the second-declaration class objectui#6357 measured — and it sits directly beneath a comment explaining that the bind key *used to be* declared exactly that way and was removed for exactly that reason.

### Gates

34 of 34 checks resolved on this head — all success or skipped, none red, none in flight. Landing now.

---

PR #8996, opened 2026-09-10
fix(core): lower `$icontains`, the canonical operator `convertFiltersToAST` refused (objectui#8976)

os-steve:
Fixes #8976

## What was wrong

`$icontains` is a canonical member of `@objectstack/spec`'s `FILTER_OPERATORS`, `ValueDataSource` executes it, `FilterConditionField` emits it for its "contains (ignore case)" builder row, and `packages/core/src/adapters/README.md` **prescribes** it as the repair when `$like` / `$ilike` / `$regex` are refused.

`convertOperatorToAST`'s `operatorMap` had no row for it, so `convertFiltersToAST` refused it with the generic unknown-operator paragraph — `INVALID_FILTER` / 400. The one spelling this repo tells an author to write was the one spelling it rejected.

## Census, re-taken on today's tree

Not the card's readings. Re-measured on `4d65991c5` (the tip this branch was cut from), spec `17.4.0`, both legs in one process, with a control operator both sides agree on:

```
spec FILTER_OPERATORS includes $icontains: true
spec VALID_AST_OPERATORS has icontains: true
spec VIEW_FILTER_OPERATORS has icontains: true
LOWERING $icontains        => THREW FilterOperatorError / INVALID_FILTER / 400
LOWERING CONTROL $contains => ["name","contains","AC"]
MATCHER  $icontains        => rows ["ACME"]
MATCHER  CONTROL $contains => rows ["ACME"]
```

The defect reproduces unchanged after PR #8974, #8977 and #8988.

## What changed

**`packages/core/src/utils/filter-converter.ts`**

- `operatorMap` gains `'$icontains': 'icontains'`, placed after `$endsWith` so the map follows the spec's own declaration order. The value is an **identity** and that is not a typo: `icontains` is itself a member of `VALID_AST_OPERATORS`, so unlike `$startsWith` to `startswith` there is no case to squash.
- The unknown-operator message now enumerates `$icontains` among the supported operators.
- The `$regex` refusal now prescribes it by name. `@objectstack/spec`'s own `FILTER_TEXT_CASES` carries `mustMention: ['$regex', '$icontains']` on that refusal, and the converter could not satisfy it while it did not accept `$icontains`.

**`packages/data-objectstack/README.md`** gains the `$icontains` row it never had, plus prose explaining the identity lowering. The `$regex` refused row's prescription is brought back in step with the code.

**`packages/data-objectstack/src/readme-filter-operator-table.test.ts`** — the pin's blind spot, closed.

[...]

PR #8988, opened 2026-09-10
fix(core)!: retire the four lowercase `$`-dialect aliases from `convertFiltersToAST` (objectui#8568)

os-steve:
Fixes #8568

`Clause-②: yes` — this lane's in-seat default-tier review plus the gate table below is the review of record.

## The ruling this implements

The director seat ruled **option 1** on this card (comment 5617467534), from the maintainer verbatim: 「我们的项目以objectstack 协议为准，文档应该以实际实现为准。」 The `$` dialect follows `@objectstack/spec`'s canonical spellings; the four lowercase aliases retire with **no deprecation window** (2026-08-27 「不渐进」); `ValueDataSource` is not touched (objectui#8447's direction stands); README rows and this repo's reconciliation pin follow the implementation.

## What changed

`convertOperatorToAST`'s `operatorMap` no longer carries `$notin`, `$notcontains`, `$startswith`, `$endswith`. All four are now **refused by name**, and the refusal **prescribes the canonical spelling**:

```
[ObjectUI] The '$startswith' filter operator is retired. Write '$startsWith' instead —
the canonical spelling '@objectstack/spec' declares in FILTER_OPERATORS
(data/filter.zod.ts). The operator itself is unchanged; only the key is renamed.
Field: 'email', Value: "a". It used to be accepted here as a lowercase alias while the
in-memory matcher refused it, so the same filter selected rows through one data source
and none through the other (objectui#8568).
```

A named arm rather than the generic unknown-operator paragraph, because with no deprecation window that message **is** the whole migration aid: "unknown operator, here is the supported list" makes an author diff two lists to find which entry they meant, while "write `$startsWith`" is the repair.

The prescriptions are **derived, not restated**. `RETIRED_OPERATOR_ALIASES` is read out of the source by the new pin and both directions are executed against `@objectstack/spec`: every KEY must be absent from `FILTER_OPERATORS`, every VALUE present in it and lowerable here. A fifth alias added later, or a prescription naming a spelling the spec does not declare, reddens.

There is also a spec-derived invariant that outlives this card: for **every** camelCase member of `FILTER_OPERATORS`, its all-lowercase spelling must not be accepted. That is commandment #0.1 itself, read off the spec rather than off the four names this card happened to retire.

## The pin reddened exactly as designed

PR #8565's reconciliation pin reads `operatorMap` out of `filter-converter.ts`'s source. After the code change and **before** the README edit, it named the four rows that had to move and nothing else:

```
[...]

PR #8984, opened 2026-09-10
fix(types)!: retire DetailViewSchema.related — author record:related_list (objectui#7997, Route C)

os-warren:
Part of objectui#7997.

Session provenance, in prose and as a code span because a footer link does not reliably survive an edit: `session_01Jmxdo7bmeqCQHLSfmLVX9w`.

> **Notation.** Angle-bracket shapes are written out in words throughout this body (`Array of (TableColumn OR string)`, `SPAN`, `BUTTON`). GitHub's body sanitizer silently eats tag-shaped fragments, backticks and fenced blocks included, and a document about a TYPE is exactly the one that collapses into reading "nothing changed" when it does.

---

## Rework round — contract review returned REWORK, and this is what moved

Review: [`5621736903`](https://github.com/objectstack-ai/objectui/pull/8984#issuecomment-5621736903). **Route C, the blast radius, the ablation, the `minor` grading and the scope split were all re-measured in the reviewer's own clone and upheld** — the rework is entirely on the docs / prose surface, plus two dead residues. No code path changed.

**R1, the decisive one, and my mistake: CI was RED at the previous head, on this PR's own diff, and my suite selection could not see it.** `scripts/__tests__/check-doc-links.test.ts` failed two rows because the two call-outs I wrote were `> ####` headings **inside blockquotes**. fumadocs' TOC counts a blockquoted heading; the gate's ATX scan cannot, so the two files read "39 headings, 38 ATX" and "22 headings, 21 ATX". They were the only such headings in the entire md/mdx surface. Both now use this tree's idiom for exactly this call-out — `> **Retired: KEY** (objectui#NNNN, ADR-0049 ...)`, 39 instances under `content/docs`. ⛔ The gate was not touched. Local re-run: **121 of 121 passed, rc 0** — the same count the reviewer measured with the two files at their merge-base blobs.

Two process lessons, recorded because they are the reusable part:

- **A docs edit runs `scripts/__tests__/` too.** That directory is the docs gates' own self-test surface, and it was in none of the suites I picked. I picked `packages/` and `apps/`; CI runs everything.
- **⛔ Do not report before CI is terminal.** I posted the last report at 15:46:51 while `Test (shard 2/4)` was still running — it finished red at 15:49:27. "My suites are green" is not "this head is green": my suites are the ones I chose. This round waited for all 34 checks to reach a terminal state before the report was written, and the counts are in it.

---

## Second rework round — one item, one comment, and I was overturned

[...]

PR #9019, opened 2026-09-10
fix(core): lower the TRUE-identity combinators to "no constraint" instead of handing back the caller's object

### LINKED ISSUE #8770: [finding] `convertFiltersToAST` hands the TRUE identity groups back UNLOWERED — `{ $and: [] }` / `{ $or: [{}] }` / `{ $and: [{}] }` fail `isFilterAST`, so a filter meaning "every row" is refused on the wire
Measured while landing objectui#8513. That card fixed the CONSUMER half (`ValueDataSource` now answers these three correctly); this is the producer half, on a different consumer, and it needs a decision rather than a mechanical repair — so it is filed, not folded in. ⛔ Not claimed.

## Measured

```
convertFiltersToAST({ $and: []   })  =>  { $and: []   }   isFilterAST: false
convertFiltersToAST({ $or:  [{}] })  =>  { $or:  [{}] }   isFilterAST: false
convertFiltersToAST({ $and: [{}] })  =>  { $and: [{}] }   isFilterAST: false
convertFiltersToAST({ $or:  []   })  =>  ['$or', '=', []]  isFilterAST: true    <- correct
```

(Run against `@objectstack/spec` 17.3.0 and this repo's built `@object-ui/core`.)

## Why

`lowerLogicalGroup` returns `undefined` for a group that reduces to the TRUE identity — deliberately, and its docblock explains it: TRUE is "the absence of a constraint", and emitting a childless `['and']` would be `isFilterAST` FALSE and `parseFilterAST` `undefined`, i.e. no filter at all, which widens.

But when that `undefined` is the ONLY thing the filter produced, `convertFiltersToAST` falls to its general tail:

```ts
// If no conditions, return original filter
if (conditions.length === 0) {
  return filter;
}
```

— and hands back the caller's original OBJECT. So the group does not disappear at that level after all; it reappears one level up, in the `$` dialect, in a slot the AST is expected to occupy. `toFilterNode` and `mergeFilterNodes` pass it straight through, and every consumer on that chain (`ObjectGrid`'s `schemaFilter`, `plugin-list`'s `buildEffectiveFilter`, `plugin-view`'s `ObjectView`) drops it onto `$filter`.

## Consequence, per consumer

- **`provider: 'object'` (the wire).** `@object-ui/data-objectstack` routes an object `$filter` through this same function and then gates on `isFilterAST`. These three shapes do not pass it, so a filter whose ruled answer is EVERY row (objectstack#5322, merged objectstack#5365) is refused instead of being dropped as no-constraint.
- **`provider: 'value'`.** Answered ZERO rows until objectui#8513; correct as of that card, which is why this one is about the producer and not about the matcher.

## The decision this needs

Returning `undefined` from `toFilterNode` for a TRUE-identity filter is the obvious shape and is NOT obviously safe: `convertFiltersToAST` is exported and its declared return type includes `Record<string, any>`, the `conditions.length === 0` tail also serves inputs that are not combinators at all (an all-null filter, `{}`), and `undefined` vs. the original object is exactly how some callers tell "no constraint" from "unlowerable". Whoever takes this should decide whether the identity fold belongs in `lowerLogicalGroup`'s caller, in `toFilterNode`, or in a distinguishable return value — not pick one in passing.

## Relationship to neighbours

Same function as objectui#8567 / #8568 / #8530 / #8555, and a residual of objectui#6948 (which added the `$and` / `$or` branch). Four different defect shapes with four different fixes — ⛔ not foldable; they serialise.

## Dedup

Run, instrument validated with a control in the same session. `search_issues` (repo-scoped, semantic) over the `convertFiltersToAST` neighbourhood returns objectui#8567, #8568, #8530, #6948, #8555 and #7752 — 9 total, none of them this shape. Closed cards were in range.

---

os-steve:
Fixes #8770

`convertFiltersToAST({ $and: [] })`, `{ $or: [{}] }` and `{ $and: [{}] }` returned the **input object unchanged**. `lowerLogicalGroup` answers `undefined` for a group that reduces to the TRUE identity — deliberately, so no childless `['and']` is emitted — but when such a group was the only thing in the filter that `undefined` fell through to the general tail (`if (conditions.length === 0) return filter`), so the group reappeared one level up, in the `$` dialect, in the slot the AST occupies. The same function already lowered the fourth identity correctly, which is what made this an internal inconsistency rather than an open question.

Direction per triage (comment 5619918295): the producer aligns to the consumer that objectui#8513 already settled.

## 1. Re-measurement of all four identity groups

Taken myself, against `@objectstack/spec` **17.4.0** (the card measured 17.3.0) and this branch's base `2596b1b85`, before any edit:

| input | before | `isFilterAST` | after |
|---|---|---|---|
| `{ $and: [] }` | `{"$and":[]}` — same object back | false | `undefined` |
| `{ $or: [{}] }` | `{"$or":[{}]}` — same object back | false | `undefined` |
| `{ $and: [{}] }` | `{"$and":[{}]}` — same object back | false | `undefined` |
| `{ $or: [] }` | `["$or","=",[]]` | **true** | unchanged (control) |

The PM's "exactly three wrong, one right" assumption **holds** on 17.4.0. Two extra readings worth recording, because they show the defect is a *family* and the three shapes are representatives, not the whole set:

- `{ $or: [{}, { s: 1 }] }` and `{ $and: [{}, {}] }` also come back unlowered — anything whose combinators all absorb or drop.
- `{ $and: [], s: 1 }` was already **correct** (`["s","=",1]`): the defect only appears when the identity group is the *only* producer of conditions. That sibling-dependence is the same hazard objectui#8555 called out on this file.

## 2. The p2 fence — what the refusal currently does

**Answer: a HARD ERROR (`400 UNSUPPORTED_QUERY_PARAM`) on the default `provider: 'object'` list path. Not a silent empty. Nothing can be relying on it to scope data, so there is no caller to stop for.**

How I established it — three links, each measured or pinned rather than assumed:

1. **The producer** (measured, above): the object comes back unlowered.
2. **The client** (measured, real `@objectstack/client` 17.4.0 with a stub fetch, observing the URL it builds): `client.data.find()` tests the value with `isFilterAST`, and its **else** branch spreads a plain object's entries as **query parameters**. So the request is
   - `?$and=` for `{ $and: [] }`
   - `?$or=[object Object]` for `{ $or: [{}] }`
   - `?$and=[object Object]` for `{ $and: [{}] }`

   with **no `filter` parameter at all**, while the control `{ $or: [] }` correctly leaves as `?filter=["$or","=",[]]`.
3. **The server** (read + pinned in its own repo): the shared normalizer rejects any unknown `$`-prefixed query parameter — `Unsupported query parameter(s)`, `status: 400`, `code: 'UNSUPPORTED_QUERY_PARAM'` — and that is pinned by `packages/objectql/src/protocol-data.test.ts` ("rejects an unknown $-prefixed query param with 400 UNSUPPORTED_QUERY_PARAM"). `$and` / `$or` are not in the supported set (`$top, $skip, $orderby, $select, $count, $search, $searchFields, $filter, $expand`) and are not reserved names either.

⚠️ **A correction to the card, which matters for the fence.** The card says these shapes "do not pass `isFilterAST`, so the filter is refused". The *outcome* is right but the *mechanism* is not, and the difference is the whole fence: the object is never sent as a filter and judged. It is **shredded into junk query parameters**, and the 400 is about the junk. The list therefore fails loudly — an `INVALID_FILTER`-class "the filter is malformed" outcome, never an empty page that could be mistaken for "no records".

I also traced the **other** wire route (`rawFindWithPopulate`, taken when `$expand` or `$search` is present, plus the export route). There the same object travels verbatim as `filter={"$and":[]}`, and the server's normalizer accepts a plain object as a `FilterCondition` (`options.where = parsedFilter`), so that route **already answers every row today**. ⇒ the two `find()` routes disagreed about one filter; after this change they agree, and the widening direction the fence warns about is *not new* on either — one route was loudly broken, the other already wide.

**No caller depends on the refusal.** In-repo, the only pin that recorded the old behaviour is objectui#8513's conformance file, and it recorded it explicitly as a symptom ("Not asserting this is right — asserting it is what happens"); it is updated here.

## 3. Which layer, and why

⛔ Not picked in passing. Three candidates, and the tail is the one that is actually wrong:

- **`toFilterNode`** — rejected. It does not serve the consumer this card is about: `@object-ui/data-objectstack` calls `convertFiltersToAST` **directly** on both of its routes, so a fold there would leave the wire exactly as broken as it is.
- **A distinguishable return value** (a sentinel separating "no constraint" from "unlowerable") — rejected on a measurement rather than on taste. **Nothing unlowerable reaches that tail.** Every arm that cannot lower its input *throws* (`$not`, `$regex`, a bare array, an exotic comparand, an unknown or retired operator). What reaches `conditions.length === 0` is only: a TRUE-identity combinator, a null/undefined-valued key, and an empty operator map — all of them "no constraint". There is no second meaning to distinguish, so a new sentinel would be vocabulary without a referent.
- **`convertFiltersToAST`'s tail — chosen.** `undefined` is not a new vocabulary either: it is already what `toFilterNode`, `mergeFilterNodes` and `data-objectstack`'s `translateFilterToAST` mean by "no filter, skip the slot", and **all four call sites already act on it correctly** — `lowerLogicalGroup` tests `Array.isArray`, the other three test for `undefined` or falsiness. The declared return type gains `| undefined` (a published-API change, called out in the changeset with migration text).

**And it is deliberately scoped to a filter whose EVERY key is such a group** — that is why the implementation counts the collapsed groups and compares the count with the key count, instead of using a flag. The tail also serves inputs that are not combinators at all, and folding those in would have been a *real* widening on a path nobody ruled about: `{ $and: [], a: null }` travels the `$expand` route as `filter={"$and":[],"a":null}`, which the server reads as a genuine `a IS NULL` predicate. Those keep the object they have always returned, and section 6 of the new test file pins that boundary.

## 4. Tests — one assertion per group, with the already-correct one as control

New: `packages/core/src/utils/__tests__/filter-true-identity-8770.test.ts` (24 cases), driven by the published `FILTER_LOGIC_ROWS` fixture the identity ruling is held to.

- the three groups lower to "no constraint"; **control** `{ $or: [] }` keeps its `['$or','=',[]]` leaf and still answers zero rows;
- the invariant that actually broke, asserted through the spec's own door: every returned value is **either absent or `isFilterAST`-readable**;
- composition — `mergeFilterNodes(identity, ['a','=','x'])` is now the plain node, where it used to be `['and', { $and: [] }, ['a','=','x']]`, an object in AST child position;
- nesting — `{ $and: [{ $and: [] }] }`;
- the untouched boundary — all-null filter, empty operator map, `{ $and: [], a: null }`, `{ $and: [], a: 'x' }`.

⚠️ **One section is labelled as NOT witnessing this fix**, in the file itself: the row-set section stays green on both ablation legs, because objectui#8513 already made the matcher answer the object dialect correctly, so the unlowered filter and the absent one select the same four rows. It pins the producer/consumer alignment direction, not this change — said out loud rather than left to look like coverage.

## 5. Ablation (run, not reasoned about)

Fix committed first, then the fold removed on disk, then restored. No `dist/` is involved: both test files import the module by **relative source path**, so vitest compiles the mutated source directly.

- on-disk proof: anchor count `1 -> 0`, injected marker `0 -> 1`, blob hash `f6b0632 -> ae00617` (a hash *change* was asserted, and an empty hash treated as failure);
- restored: blob hash back to `f6b0632` = the `HEAD` blob, and `git diff HEAD` on the path is empty;
- the script carries `trap restore EXIT INT TERM` with absolute paths, and restores with `git checkout HEAD -- path` (never bare `git checkout --`, which would take the mutation back out of the index).

| leg | result |
|---|---|
| fold present (`HEAD`) | 61 passed, 0 failed |
| fold removed | **12 failed**, 49 passed |
| restored | byte-identical to `HEAD` |

The 12 are exactly the predicted ones: the 3 shape cases, the 3 `isFilterAST`-invariant cases, the 3 merge cases, TRUE-and-FALSE, the nested case, and objectui#8513's updated chain pin. Predicted 13 before running; the single difference was my own miscount (two assertions living in one `it`), not a case that failed to witness.

## 6. Gates

| gate | exit | note |
|---|---|---|
| `turbo run type-check` (**repo-wide**, 81 packages) | 0 | 81/81 successful — the published return type changed, so the whole tree is the honest scope |
| `vitest run packages/core/ packages/data-objectstack/` | 0 | 203 files, 3869 tests passed |
| `turbo run lint` (`@object-ui/core`, `@object-ui/data-objectstack`) | 0 | 0 errors; the 534 warnings are pre-existing `no-explicit-any` |
| `eslint --format json` on the 4 changed source files | 0 | 4 files, **0 errors**, 139 warnings, all pre-existing rule classes |
| `check-control-bytes.mjs` | 0 | plus a direct scan of the changed files for the non-NUL control range |
| `check-changeset-presence.mjs` | 0 | "3 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)" |
| `check-changeset-no-major.mjs` | 0 | changeset is `minor` |
| `check-governed-queue-guard.mjs --test` (5 paths) | 0 | "NOT GOVERNED — 5 path(s) checked against 5 governed surface(s); none matched" |
| `check-new-cross-file-line-citations.mjs` | 0 | |
| `check:doc-types` | 0 | |
| `check:doc-snippets` | **2 -> 0** | exit 2 was PRECONDITION NOT MET on an unbuilt tree; re-run after the scoped build it names — 638/638 blocks judged |
| `check:doc-examples` | **2 -> 0** | same; 124 blocks, the 89 failures all declared in the ledger |

Repo-wide `pnpm lint` and the full-farm suites are left to CI. Every exit code above was captured to a file before being read, never through a pipe. The two heavy runs went through the shared verify lock.

## 7. Assumptions I was asked to falsify

- **"Exactly three wrong, one right" — HOLDS** on `@objectstack/spec` 17.4.0, re-measured after #8974 / #8977 / #8988 / #8996 landed. Refined: the broken *family* is larger than three inputs (any filter whose combinators all reduce to TRUE), and the three named shapes are its minimal representatives.
- **"`lowerLogicalGroup`'s `undefined` is deliberate; the defect is the general tail" — HOLDS**, chain verified end to end.
- **"`convertFiltersToAST` is exported, its tail also serves non-combinators, so the obvious shape is not obviously safe" — HOLDS, and it changed the design**: the fold is key-count-scoped for exactly this reason, and the boundary is pinned.
- **The card's "refused because it fails `isFilterAST`" — FALSIFIED.** The refusal is real but arrives as `400 UNSUPPORTED_QUERY_PARAM` for shredded query parameters, one layer earlier than the card says. Section 2.
- **Serial adjacency, re-derived here and not taken from any snapshot**: I fetched the head of every open non-dependabot PR branch into a private ref namespace and diffed each against its merge-base. **None touches `packages/core`** — #9010 (`packages/plugin-grid`), #9009 (`scripts/`), #9008 (`packages/create-plugin`), #9007 (ci), #9005 (docs), #8984 (`packages/types` + app-shell), #8941 (deps). objectui#8983 is real and its branch is pushed with no PR; it touches `packages/app-shell` and `packages/plugin-detail`, so it does not overlap either. The only other branches whose names suggested this neighbourhood (`claude/issue-7971-...`, `claude/issue-8600-...`) point at commits that are already `main` history, i.e. stale, not in flight. objectui#8976 has landed (PR #8996) and nothing it changed is revisited here.

---

### DIFF
--- a/.changeset/8770-true-identity-groups-no-constraint.md
+++ b/.changeset/8770-true-identity-groups-no-constraint.md
@@ -1,0 +1,53 @@
+---
+'@object-ui/core': minor
+'@object-ui/data-objectstack': minor
+---
+
+Lower the TRUE-identity filter combinators to "no constraint" instead of handing the
+caller's own object back (objectui#8770).
+
+`convertFiltersToAST({ $and: [] })`, `{ $or: [{}] }` and `{ $and: [{}] }` returned the
+INPUT OBJECT unchanged. `lowerLogicalGroup` correctly answers `undefined` for a group
+that reduces to the TRUE identity — objectstack#5322 rules all three "every row", and a
+childless `['and']` would be `isFilterAST` FALSE — but when such a group was the only
+thing in the filter, that `undefined` fell through to the general tail
+(`if (conditions.length === 0) return filter`) and the group reappeared one level up, in
+the `$` dialect, in the slot the AST occupies. The same function already lowered the
+fourth identity, `{ $or: [] }`, correctly, so this was an internal inconsistency rather
+than an open question; the consumer half was settled by objectui#8513.
+
+**This widens what those three filters return, and that is the point.** Measured against
+`@objectstack/spec` 17.4.0 and `@objectstack/client` 17.4.0 before the change: the
+returned object is not sent as a filter and refused — `client.data.find()` tests the
+value with `isFilterAST` and its else branch spreads a plain object's entries as query
+parameters, so `{ $and: [] }` left as `?$and=` with **no `filter` parameter at all**, and
+the server answered `400 UNSUPPORTED_QUERY_PARAM` for the unknown `$`-prefixed
+parameter. A filter whose ruled answer is EVERY ROW was a **failed list**, not a narrowed
+one — so nothing could have been relying on it to scope data. On the sibling
+`$expand` / `$search` route the same object travelled as `filter={"$and":[]}`, which the
+server accepts as a `FilterCondition` and already answers with every row; the two routes
+disagreed about one filter and now agree.
+
+**`@object-ui/core`.** `convertFiltersToAST`'s declared return type gains `undefined`,
+which is what `toFilterNode`, `mergeFilterNodes` and `data-objectstack`'s
+`translateFilterToAST` already mean by "no filter, skip the slot". Every call site
+already acted on it. The fold is scoped to a filter whose EVERY key is such a group:
+the same tail also serves `{}`, an all-null filter and an empty operator map, and those
+keep the object they always returned — a null-valued key is this converter's own
+tolerance rather than a ruled identity, and the object it hands back reaches the server
+as a REAL `a IS NULL` predicate on the `$expand` route, so folding it in would return
+more rows on a path the ruling said nothing about.
+
+`{ $or: [] }` is untouched: FALSE is not "no constraint", the AST has no contradiction
+literal, and its `['$or', '=', []]` leaf answers FALSE at both consumers.
+
+**`@object-ui/data-objectstack`.** `convertQueryParams` skips the `filters` slot when the
+lowering answers `undefined`, the same answer the raw-GET route's
+`if (translated !== undefined)` already gave, so the two `find()` routes cannot disagree
+about one filter.
+
+**Migration.** A TypeScript caller that stored `convertFiltersToAST(...)` in a
+`FilterNode | Record<string, any>` slot must widen it with `| undefined` and skip the
+filter when it is absent — the same handling `toFilterNode` has always needed. At
+runtime, a filter that is nothing but TRUE-identity combinators now returns every row
+(what objectstack#5322 rules) instead of failing the request.
--- a/packages/core/src/adapters/__tests__/ValueDataSource.filterLogicConformance-8513.test.ts
+++ b/packages/core/src/adapters/__tests__/ValueDataSource.filterLogicConformance-8513.test.ts
@@ -189,27 +189,38 @@ describe('objectui#8513 — `$not` is out of scope, and stays refused', () => {
  * `engine.find`, and it never passes through this adapter.
  *
  * The producer that does reach here is this repo's own single filter sink.
- * `convertFiltersToAST` returns the ORIGINAL OBJECT when a filter lowers to no
- * conditions, and the identity groups are exactly that case — so `toFilterNode`
- * hands `{ $and: [] }` and `{ $or: [{}] }` back UNLOWERED, and every consumer
- * on that chain (`ObjectGrid`'s `schemaFilter`, `plugin-list`'s
- * `buildEffectiveFilter`, `plugin-view`'s `ObjectView`) drops them straight
+ * `convertFiltersToAST` returned the ORIGINAL OBJECT when a filter lowered to
+ * no conditions, and the identity groups were exactly that case — so
+ * `toFilterNode` handed `{ $and: [] }` and `{ $or: [{}] }` back UNLOWERED, and
+ * every consumer on that chain (`ObjectGrid`'s `schemaFilter`, `plugin-list`'s
+ * `buildEffectiveFilter`, `plugin-view`'s `ObjectView`) dropped them straight
  * onto `$filter`. Before this change the matcher answered ZERO ROWS for both,
  * on filters whose ruled answer is EVERY row — the worst direction, on the
  * reachable path.
+ *
+ * ⚠️ The PRODUCER half moved afterwards: objectui#8770 made `convertFiltersToAST`
+ * answer `undefined` — "no constraint" — for a filter that is nothing but
+ * TRUE-identity combinators, so this matcher is no longer handed those objects
+ * BY THAT CHAIN. The matcher's own answer is asserted below all the same, and
+ * deliberately from the object literals rather than through `toFilterNode`: it
+ * is a `DataSource` with a public `find`, the object dialect is what
+ * `QuerySchema.where` declares, and #8513's ruling is about what this adapter
+ * answers — not about which producer happens to call it this month.
  */
-describe('objectui#8513 — the lowering chain hands this matcher an object', () => {
-  it('`toFilterNode` leaves the TRUE identities in the object dialect', () => {
-    // Not asserting this is right — asserting it is what happens, because it
-    // is why the matcher has to answer correctly for these shapes.
-    expect(toFilterNode({ $and: [] })).toEqual({ $and: [] });
-    expect(toFilterNode({ $or: [{}] })).toEqual({ $or: [{}] });
+describe('objectui#8513 — this matcher answers the object dialect directly', () => {
+  it('the lowering chain no longer hands the identities down in that dialect', () => {
+    // objectui#8770: the producer stopped re-emitting the caller's object for a
+    // filter it read as "no constraint". Pinned here because the paragraph
+    // above rests on it, and a silent revert would put an object back on
+    // `$filter` where the wire refuses it.
+    expect(toFilterNode({ $and: [] })).toBeUndefined();
+    expect(toFilterNode({ $or: [{}] })).toBeUndefined();
   });
 
-  it('and the matcher now answers them the way the ruling says', async () => {
+  it('and the matcher answers the objects themselves the way the ruling says', async () => {
     const allIds = FILTER_LOGIC_ROWS.map((r) => r.id);
-    expect(await selectedIds(toFilterNode({ $and: [] }))).toEqual(allIds);
-    expect(await selectedIds(toFilterNode({ $or: [{}] }))).toEqual(allIds);
+    expect(await selectedIds({ $and: [] })).toEqual(allIds);
+    expect(await selectedIds({ $or: [{}] })).toEqual(allIds);
   });
 
   it('a group that DOES lower still goes down the AST arm, unchanged', async () => {
--- a/packages/core/src/utils/__tests__/filter-true-identity-8770.test.ts
+++ b/packages/core/src/utils/__tests__/filter-true-identity-8770.test.ts
@@ -1,0 +1,227 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The TRUE-identity combinators reach the wire as "no constraint" —
+ * objectui#8770.
+ *
+ * ## What was wrong
+ *
+ * `lowerLogicalGroup` answers `undefined` for a group that reduces to the TRUE
+ * identity, and that is correct and deliberate (see its docblock: a childless
+ * `['and']` is `isFilterAST` FALSE, which reads as no filter at all). But when
+ * such a group was the ONLY thing in the filter, that `undefined` fell through
+ * to `convertFiltersToAST`'s general tail — `if (conditions.length === 0)
+ * return filter` — and the CALLER'S ORIGINAL OBJECT came back. So the group did
+ * not disappear at that level after all; it reappeared one level up, in the `$`
+ * dialect, in a slot the AST is expected to occupy.
+ *
+ * Three of the four identities went out that way and one did not, which is what
+ * made this an internal inconsistency rather than an open question: the same
+ * function already lowered `{ $or: [] }` correctly.
+ *
+ * ## Why the assertions are shaped the way they are
+ *
+ * Two halves, because either alone passes on something worse than the bug:
+ *
+ *   - a SHAPE assertion (`toBeUndefined`) alone passes on a converter that has
+ *     stopped emitting anything at all;
+ *   - a ROW-SET assertion alone passes on a filter that never ran — `$filter:
+ *     undefined` returns EVERY row, and so does a correct TRUE identity, so the
+ *     row set cannot tell "honoured" from "dropped".
+ *
+ * So the shape half is asserted against the spec's own door (`isFilterAST`)
+ * rather than against a literal, the row-set half names exact ids, and the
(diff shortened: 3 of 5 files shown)


os-steve:
One more gate result, arrived after the body was written — the three consumers the card names on the lowering chain (`ObjectGrid`'s `schemaFilter`, `plugin-list`'s `buildEffectiveFilter`, `plugin-view`'s `ObjectView`):

| gate | exit | note |
|---|---|---|
| `vitest run packages/plugin-list/ packages/plugin-view/ packages/plugin-grid/` | 0 | 232 files, 2283 tests passed |

Also re-derived after the fact: `main` moved to `da7c3d96d` while this was in flight (docs only, PR #9005), and it touches neither `packages/core` nor `packages/data-objectstack`, so `mergeable_state: behind` here is staleness and not an overlap. No rebase — the queue rebuilds.

Out of scope and filed separately rather than repaired here: #9020, a null-valued filter key getting two different row sets on the two `find()` routes. PR #9019's fold is key-count-scoped precisely so that it does not decide that question.

---

os-steve:

**Verdict: land.** One observation below is a follow-up card, not a change to this PR.

Clause-② is engaged, and I want to say why explicitly rather than let it be inferred: `convertFiltersToAST` is a **published export** — `packages/core/src/index.ts:23` is `export * from './utils/filter-converter.js'` — and its declared return type gains a member. That is a contract delta on `@object-ui/core`'s public surface. Per objectstack#17285 this lane's in-seat default-tier review plus the gates is the review of record for that; the contract-review tier is not this lane's. So this comment is the review, and it is answerable.

### The consumer set, enumerated by me rather than taken on the report's word

⚠️ My first enumeration was **not a reading**: `git grep -- 'packages/*/src'` is a glob matched against whole paths, not a directory prefix, and it returned zero. The control token `toFilterNode` returned zero under the same pathspec, which is what exposed it — the same instrument failure the #8920 dev caught last round. Corrected pathspec, control lit, and then:

`convertFiltersToAST` has **two** call sites in this repository outside its own module, both in `packages/data-objectstack/src/index.ts`:

| site | absorbs `undefined`? | how |
|:--|:--|:--|
| `:627`, inside `translateFilterToAST` | **already** | that function's own declared return type is `unknown \| undefined` and its docblock already reads *"Returns `undefined` when the input is empty/unrecognized so callers can skip emitting `?filter=` entirely."* The new answer is inside a type and a meaning it already had. |
| `:4613` → `:4622` | **this PR** | `const lowered = …; if (lowered !== undefined) options.filters = lowered;` |

And `translateFilterToAST` has **two** callers, not the one the report names — `:4459` (list GET) and `:4541` (`exportRecords`). I checked both: each is `const translated = …; if (translated !== undefined) { queryParams.set('filter', …) }`. So the widening is absorbed at every sink, and the report undercounted in the safe direction.

The four direct `toFilterNode` sinks outside `core` — `RelatedList.tsx:477`, `LineItemsPanel.tsx:151`, `ObjectGrid.tsx:1479` and `:1853` — are unaffected at the type level, because `toFilterNode`'s signature already carried `| undefined` and is unchanged here. At runtime they now receive `undefined` where they received an object; that direction cannot regress, since the object they used to receive is the one that left as `?$and=` and drew `400 UNSUPPORTED_QUERY_PARAM`.

### The fence, and why the scoping is a count and not a flag

`trueIdentityGroups > 0 && trueIdentityGroups === Object.keys(filter).length` is the right shape. The `return filter` tail also serves `{}`, an all-null filter and an empty operator map, and those are **not** the ruled family: an all-null filter travels the `$expand` route as `filter={"a":null}` and the server reads a real `a IS NULL` predicate, so folding it would return *more* rows on a path objectstack#5322 never ruled on. The PR keeps that boundary and pins it in section 6. `{ $or: [] }` stays FALSE with its `['$or','=',[]]` leaf, carried as a live control through every section — a change that flattened the identities into one arm takes the control with it. That is the correct control design for this defect class.

### On the evidence, including the part that is honestly labelled as not evidence

Section 3 is declared in its own docblock to **not** witness this fix — both ablation legs leave it green, because objectui#8513 already taught `ValueDataSource`'s matcher to answer the object dialect. Saying so in the file, rather than letting the next reader count it, is the standard this lane wants. The witnesses are sections 1, 2, 4, 5 and 6.

The edit to #8513's existing pin is legitimate and I checked it specifically, because editing another card's pin is where a ruling quietly dies. It does not weaken #8513: that card's ruling is *what this adapter answers for the object dialect*, and the assertion still exercises exactly that — moved from `selectedIds(toFilterNode({ $and: [] }))` to `selectedIds({ $and: [] })`, i.e. from the literals rather than through a producer that no longer hands them down. A pin was **added** (`toFilterNode({ $and: [] })` is now asserted `toBeUndefined`), not removed, so a silent revert of this PR reddens there too.

The changeset is `minor` with the break and a Migration section spelled out, which is the correct handling given this repo forbids `major`; Changeset Bump Policy is green. The new JSDoc `@example` compiles under `Doc Snippet Type Check` (green on this head) with no ledger row added, so it is a real example and not a declared failure.

### Observation — the fence is one shape wider than its own rationale. Follow-up card, not a change here.

`Object.keys(filter).length` counts keys that the loop itself deliberately skipped: the first statement in the loop body is `if (value === null || value === undefined) continue;`. For `null` the PR has a stated and correct reason to keep the old answer. For an **`undefined`-valued** key that reason does not hold — `JSON.stringify` drops the key entirely, so `{ $and: [], b: undefined }` reaches the `$expand` route as `filter={"$and":[]}`, which is already the ruled TRUE answer, and reaches the raw-GET route as the same `?$and=` failure this card exists to end. So `{ $and: [], b: undefined }` is in the ruled family and is not folded.

Explicitly **not a blocker and not a regression**: the behaviour is byte-identical to before this PR, so nothing that worked stops working. It is a missed member, not a broken one, and widening a PR that is already green under review to chase it is the wrong trade. I will file it as its own card against this tail once this lands, so the ruled family closes on the record rather than in a review comment.

### Gates

34-check roster read on this head; the five that were still running at review time are the test shards and Type Check. Landing on every-check-green, not on this comment.

---

PR #9028, opened 2026-09-10
fix(plugin-grid,core): normalize `object-grid`'s sort before joining `$orderby` (objectui#8973)

### LINKED ISSUE #8973: finding(plugin-grid): `object-grid`'s private array-arm lowering interpolates missing keys into `$orderby` — a `SortConfig` with `order` omitted goes out as `name undefined`
Filed unassigned by the `os-dev` seat implementing objectui#8767 (session `session_01Jmxdo7bmeqCQHLSfmLVX9w`, PR #8960), on the rework round, at the dispatching seat's request. Not claiming, not grading — triage owns both.

## The defect

`ObjectGrid` lowers `schema.sort` with its own private code rather than the shared sink `convertSortToQueryParams`. Its array arm is a bare template interpolation:

```
params.$orderby = schemaSort.map((s: any) => `${s.field} ${s.order}`).join(', ');
```

Every key it reads is interpolated **unconditionally**, so a member missing either key reaches the wire as the literal text `undefined`. The shared sink does not do this: it skips entries with no usable `field` and defaults a missing `order` to `'asc'`.

This matters most for `order`, because that is the key an author is most likely to leave out — the diagnostic PR #8758 ships, which objectui#8767 now makes `object-grid` print, tells the author verbatim that "`order` is optional and means `'asc'`". That prescription is true of the sink and **false of this block**: an author who follows the migration advice literally gets `$orderby: 'name undefined'`.

⚠️ `SortConfig.order` is in fact **required** on both declared faces — `packages/types/src/objectql.ts` (`order: 'asc' | 'desc'`, no `?`) and its zod mirror `packages/types/src/zod/objectql.zod.ts` (`z.enum(['asc','desc'])`, no `.optional()`) — and the protocol's reusable `SortItemSchema` requires it too (measured on `@objectstack/spec@17.4.0`: `z.array(SortItemSchema).safeParse([{ field: 'name' }])` refuses with `invalid_value` at `0.order`). So this is not a case of the renderer having to honour an optional key. It is a lowering that turns a type error into silent garbage on the wire instead of dropping it or refusing it.

## Measured — the reproducible probe

Rendered the real `object-grid` through `SchemaRenderer` against a `vi.fn()` data source and read `$orderby` off the first `find` call. Run on `claude/issue-8767-object-grid-refuses-string-sort` at `52d28eb4b`. Recipe: render `{ type: 'object-grid', objectName: 'account', columns: [{ field: 'name' }], sort }` inside a `SchemaRendererProvider`, await the first `find`, read `params.$orderby`.

| authored `sort` (array arm) | `$orderby` that goes out |
| --- | --- |
| `[{ field: 'name', order: 'desc' }]` — **CONTROL** | `"name desc"` |
| `[{ field: 'name' }]` — `order` omitted | **`"name undefined"`** |
| `[{ field: 'name', order: 'desc' }, { field: 'status' }]` | **`"name desc, status undefined"`** |
| `[]` | **`""`** — an empty `$orderby` is sent rather than the key being omitted |
| `['name desc']` — the retired array-of-strings | **`"undefined undefined"`** |
| `[{ order: 'desc' }]` — `field` omitted | **`"undefined desc"`** |

The control row is what makes the other five readings a measurement rather than a schema that mangles everything.

## Not objectui#8767's, and why

objectui#8767's ruling (maintainer, 2026-09-10) is route C: refuse the retired **string** clause, keep the wire shape, touch nothing else. Widening that diff to normalise the array arm is exactly the direction the ruling refused. This defect is **pre-existing and unchanged** by that PR — the contract review measured every input class at base and at head and confirmed these five outputs move on neither side.

## Relationship to objectui#8961

objectui#8961 parks the other half of the same divergence: the header-arrow reader `parseSchemaSort` is wider than the fetch path. Both cards exist because `object-grid` still owns two private lowerings of one key while every sibling block routes through the shared sink. The contract review's own suggestion was to **append** these array-arm readings to objectui#8961 rather than open a separate card; this is filed separately only because the dispatching seat asked for a dedicated card. ⇒ If triage prefers one card, fold this into objectui#8961 and close this as a duplicate — the readings above are the payload either way.

Whoever takes it should decide the array arm's fate **together with** those two, since a normalised arm and the shared sink's `{ field: direction }` map are the same question asked twice.

Refs: objectui#8767 (the route-C ruling) · PR #8960 (its implementation, where this surfaced) · objectui#8961 (the header-reader half) · objectui#8221 (the retirement ruling) · PR #8758 (the shared sink's narrowing, whose prescription this arm contradicts)

---

os-steve:
Fixes #8973

`ObjectGrid` lowers `schema.sort` with private code rather than the shared sink, and its array arm was a bare template interpolation:

```
params.$orderby = schemaSort.map((s: any) => `${s.field} ${s.order}`).join(', ');
```

Every key was interpolated **unconditionally**, so a member missing `field` or `order` reached the wire as the literal text `undefined`.

## This is a wire failure, not a cosmetic one — measured

Measured against `normalizeSortNodes` (`@objectstack/metadata-protocol/src/protocol.ts:3321`, objectstack `origin/main` `706ad0f`) — described in its own docblock as "the one shared normalizer every ingress funnels through":

| what went out | what the server does with it |
| --- | --- |
| `"name undefined"` | direction `'undefined'` is "neither `'asc'` nor `'desc'`" ⇒ **`400 INVALID_QUERY`** |
| `"undefined undefined"` | same ⇒ **`400 INVALID_QUERY`** |
| `"undefined desc"` | a **well-formed** sort on a column literally named `undefined` — the silently-wrong-ordering half |
| `""` | `value === '' ⇒ return []` — benign at the server |

So triage's severity claim ("a server-side failure or a silently wrong ordering") is confirmed on both branches. The empty-string row is the one that is *not* a server failure, and it is fixed anyway — see below.

## Acceptance table — the card's six rows, re-measured

Driven through a real `object-grid` via `SchemaRenderer` against a `vi.fn()` data source, reading `$orderby` off the first `find` call.

| authored `sort` | before | after |
| --- | --- | --- |
| `[{ field: 'name', order: 'desc' }]` — **CONTROL** | `"name desc"` | `"name desc"` — unchanged |
| `[{ field: 'name' }]` | `"name undefined"` | `"name asc"` |
| `[{ field: 'name', order: 'desc' }, { field: 'status' }]` | `"name desc, status undefined"` | `"name desc, status asc"` |
| `[]` | `""` | key omitted entirely |
| `['name desc']` | `"undefined undefined"` | key omitted entirely |
| `[{ order: 'desc' }]` | `"undefined desc"` | key omitted entirely |

On `[]`: an empty `$orderby` is benign at the server, so this row is not a bug fix by severity — it is fixed for consistency with the correction `toFilterNode` already made on the **filter leg of this same query build**, where `defaultFilters: {}` used to send `$filter: {}`. Asking a question with no content is not the same as not asking. `[]` is also, unlike the `order`-omitted row, **accepted** by `@objectstack/spec` — so it is the row that makes this defect reachable from fully spec-valid authored metadata.

## The wire shape does NOT move, and why the triage-preferred route was not taken

Triage's ruling was to fix this by **deleting the private lowering** and routing through `convertSortToQueryParams`. That route is unavailable, and not on my judgement:

- `convertSortToQueryParams` returns `Record<string, 'asc' | 'desc'>` — a **map**. `ObjectGrid` sends a `"field order"` **join string**. Routing this arm through the sink therefore cannot be done without moving the wire shape.
- Moving it is **route B on objectui#8767**, declined by the maintainer by name on 2026-09-10 (comment [`5617214837`](https://github.com/objectstack-ai/objectui/issues/8767#issuecomment-5617214837)): "Route B (route through the shared sink) is **not** taken — it moves the wire shape for every grid and drags in the export path and `parseSchemaSort`; that is a different card."
- #8973 is not that card: it measures the wire output, but neither the server contract nor the header reader.

**Reconciled instead of chosen between.** Triage's actual principle — one operation, one implementation — is satisfied for the operation that genuinely had two copies. The "which entries survive, and what does a missing `order` mean" decision is lifted out of `convertSortToQueryParams` into a new shared, **shape-agnostic** export `normalizeSortEntries` (`@object-ui/core`). The map builder is now a projection of it; `ObjectGrid` builds its join string from it. One implementation of the rule, two wire shapes, and the shape the declination protects does not move.

The join itself stays local to `ObjectGrid` deliberately: it is the only block that sends a join string, and putting it in the governed module would enshrine in core the very shape the maintainer's preferred end state retires.

## Also fixed: the legacy `defaultSort` arm, one `else` down

```
params.$orderby = `${(schema.defaultSort as any).field} ${(schema.defaultSort as any).order}`;
```

**Identical defect**, same `if`/`else` chain, same `400`. Fixed through the same normalizer. This is a bounded in-place fix, not a widening: same defect class, mechanical, no other claim on the file, no new gate surface. It is an isolated hunk with its own acceptance rows — drop it independently if you would rather it were its own card.

## The other arms `ObjectGrid` lowers — named, not absorbed

Triage asked for these to be checked before switching. There are **four** private lowerings of this one key, not one, and two distinct `$orderby` wire shapes already coexist:

1. **`headerSort` arm** — sends `SortNode[]`, i.e. `[{ field, order }]` **objects**, not a join string. Documented at the read site as deliberate ("the shape the server names in its own error messages, and it survives a field name containing a space"). Not a defect, and **not touched**. But it means the premise "`object-grid` sends a join string" is only true of the schema-declared arms.
2. **`schemaSort` string arm** — objectui#8767's landed route-C refusal. Untouched, and pinned here as still firing.
3. **`schemaSort` array arm** — this card.
4. **`defaultSort` arm** — this card, above.
5. **The export path** (`ObjectGrid.tsx:3120`) builds `{ field, direction }` for `exportDownload` — a different envelope, not `$orderby`. It **already** normalizes correctly (skips entries with no `field`, defaults `order`), which is why the correct normalization was already present in this file and simply had not been applied to the fetch path. One behavioural difference worth naming: it uses `?? 'asc'`, which passes a garbage direction like `'DESCENDING'` through unchanged, where the shared normalizer folds it to `'asc'`. **Not changed here** — it is the export path, which #8767's ruling also fenced off.
6. **`parseSchemaSort`** — objectui#8961, `pm:blocked`, `domain:spec` + `domain:ui`. **Not touched.**

## Owed to objectui#8961 and objectui#8767 — the server `$orderby` contract, MEASURED and deliberately NOT acted on here

The #8767 declination asked for a card carrying its own measurement of the server contract. Here is that half, so the maintainer does not have to go and get it. **Nothing in this PR acts on it.**

Measured in `@objectstack/metadata-protocol`'s `normalizeSortNodes` (objectstack `origin/main` `706ad0f`), the single normalizer every server ingress funnels through. It accepts **four** shapes:

| shape | example | accepted? |
| --- | --- | --- |
| join string | `"name desc, status asc"` | yes — what `object-grid`'s schema arms send |
| `SortNode[]` | `[{ field: 'name', order: 'desc' }]` | yes — what `object-grid`'s **header** arm sends |
| `string[]` | `['-created_at']` | yes |
| `Record` map | `{ created_at: 'desc' }` | yes — what `convertSortToQueryParams` returns |

⇒ **The server accepts both the join string and the map.** Route B is not blocked by the server contract. The map and the `string[]` form were both silently dropped until objectstack#4226 taught this normalizer to fold them; `SqlDriver` guards its ORDER BY with `Array.isArray(query.orderBy)`, so a shape it could not read produced no clause at all.

Two caveats that keep this from being a green light on its own:

- The **OData/HTTP** face (`@objectstack/spec` `packages/spec/src/api/odata.zod.ts`) declares `$orderby` as `string | string[]` only — it declares neither the map nor `SortNode[]`. That is plausibly just two doors (a querystring surface where everything is a string, versus the JSON query body), **not** a measured contradiction, and I am reporting it as an observation rather than a finding.
- The **second** half the declination asked for — the header reader — is still unmeasured, by design: `parseSchemaSort` is objectui#8961's and out of bounds for this lane.

## Verification

Ablation (predicted **before** running, matched exactly): reverting the two `ObjectGrid` arms to the unconditional interpolation reddens **exactly 8** tests — the six array rows and the two `defaultSort` rows. Both CONTROL rows, the header-arm pin, the string-arm refusal, all of `gridRetiredStringSort-8767.test.tsx` and all of the core suite stay green. `8 failed | 37 passed`. The mutation was proven on disk in both directions before the run (injected token present, removed token gone, blob hash changed), and the restore was proven by `git diff HEAD` empty plus a blob-hash match against `HEAD` — not by an exit code.

No rebuild was needed for the ablation to be real: the root vitest config aliases `@object-ui/core` to `packages/core/src`, so these suites resolve source, not `dist/`.

Reverse verification of the cross-package type change: feeding `normalizeSortEntries` a `number` from `ObjectGrid.tsx` reddens with `TS2345: Argument of type 'number' is not assignable to parameter of type 'QuerySortEntry[]'` — proving `plugin-grid`'s `type-check` reads the rebuilt `.d.ts` and not a cache. Restored, hash-verified.

- `pnpm --filter @object-ui/types --filter @object-ui/core --filter @object-ui/plugin-grid type-check` — all three `Done`, after building the full `plugin-grid` dependency closure. (Before that build, `plugin-grid` reported 60-plus `TS2307 Cannot find module` errors purely from unbuilt sibling `dist/` — stale-`dist` noise, not this diff.)
- Targeted suites: `3 passed (3)`, `45 passed (45)`.
- `pnpm build` (turbo, whole workspace): `43 successful, 43 total`.
- Gates: `check:control-bytes`, `check:new-line-citations`, `check:esm-specifiers`, `check:self-import`, `check:unreferenced-sources`, `check:vi-mock-specifiers`, `check:readme-exports` — all exit 0. `check:readme-exports` first failed with "the population COLLAPSED", which was an unbuilt worktree (`packagesRead: found 13, floor is 25`, "24 unbuilt"), **not** this diff; it passes cleanly on the built tree with `0 unbuilt`.
- Repo-level scans (`pnpm lint`, the full suite, the rest of the 49 `check:*` farm) are left to CI.

## Acceptance notes

- The runtime diagnostic in `reportRetiredSortSpelling` still tells authors "`order` is optional and means `'asc'`". That phrase is false of `@objectstack/spec`, which refuses an entry omitting `order`. objectui#8767's contract review routed this sentence here along with its two docblock twins — the twins are corrected in this PR (comments only), but the **diagnostic string is deliberately left alone**: it is objectui#8758's landed refusal text, reused by #8767's route-C ruling, so rewording it is a behaviour-visible change to a landed refusal rather than a comment fix. Raising it rather than doing it. Noted, not filed.
- `parseSchemaSort` reads three spellings the fetch path now refuses or drops, so a grid authored with a retired spelling still paints an arrow for an ordering its query does not carry. Already objectui#8961. Noted, not filed.
- The export path's `?? 'asc'` passes a garbage direction through where the shared normalizer folds it to `'asc'`. Reachable only from already-invalid metadata, and inside #8767's fenced-off export path. Noted, not filed — successor: whichever card takes the export path.

---

### DIFF
--- a/.changeset/8973-grid-orderby-normalize.md
+++ b/.changeset/8973-grid-orderby-normalize.md
@@ -1,0 +1,47 @@
+---
+'@object-ui/core': minor
+'@object-ui/plugin-grid': patch
+---
+
+`object-grid` normalizes its `sort` before joining it into `$orderby` (objectui#8973).
+
+**The defect.** `ObjectGrid` lowers `schema.sort` with private code rather than the
+shared sink, and its array arm interpolated every key unconditionally:
+
+```js
+params.$orderby = schemaSort.map((s) => `${s.field} ${s.order}`).join(', ');
+```
+
+So an entry missing `field` or `order` reached the wire as the literal text
+`undefined`. The legacy `defaultSort` arm one `else` down had the identical defect on a
+single object, and is fixed with it — leaving it would keep the class open inside the
+same `if`/`else` chain.
+
+This is a wire failure, not a cosmetic one. `normalizeSortNodes` — the one normalizer
+every `@objectstack` server ingress funnels through — validates the direction token, so
+`$orderby: 'name undefined'` is answered `400 INVALID_QUERY`, while `'undefined desc'`
+becomes a well-formed sort on a column literally named `undefined`.
+
+**What changes on the wire.** Only inputs that were already broken:
+
+| authored `sort` | before | after |
+| --- | --- | --- |
+| `[{ field: 'name', order: 'desc' }]` | `"name desc"` | `"name desc"` (unchanged) |
+| `[{ field: 'name' }]` | `"name undefined"` | `"name asc"` |
+| `[{ field: 'name', order: 'desc' }, { field: 'status' }]` | `"name desc, status undefined"` | `"name desc, status asc"` |
+| `[]` | `""` | key omitted |
+| `['name desc']` | `"undefined undefined"` | key omitted |
+| `[{ order: 'desc' }]` | `"undefined desc"` | key omitted |
+
+**The wire SHAPE does not move.** The `"field order"` join string stays. Routing this
+arm through `convertSortToQueryParams` would send that sink's `{field: direction}` map
+instead — route B on objectui#8767, which the maintainer declined by name on 2026-09-10
+pending a card that measures the server contract and both readers.
+
+**New in `@object-ui/core`:** `normalizeSortEntries` (and its `NormalizedSortEntry`
+type) — the "which entries survive, and what does a missing `order` mean" decision,
+lifted out of `convertSortToQueryParams`, which is now a map projection of it. This is
+what lets a block sending a different wire shape share the one implementation of the
+rule instead of keeping a private copy that drifts. `convertSortToQueryParams`'s own
+behaviour is unchanged and pinned against literals captured from the previous
+implementation.
--- a/.changeset/8973-grid-orderby-normalize.md
+++ b/.changeset/8973-grid-orderby-normalize.md
@@ -1,6 +1,7 @@
 ---
 '@object-ui/core': minor
 '@object-ui/plugin-grid': patch
+'@object-ui/types': patch
 ---
 
 `object-grid` normalizes its `sort` before joining it into `$orderby` (objectui#8973).
@@ -45,3 +46,11 @@ what lets a block sending a different wire shape share the one implementation of
 rule instead of keeping a private copy that drifts. `convertSortToQueryParams`'s own
 behaviour is unchanged and pinned against literals captured from the previous
 implementation.
+
+**Two false docblock sentences corrected** (`@object-ui/types`' `ObjectGridSchema.sort`
+and this sink's own header). Both read "`order` is optional and means `'asc'`" — the
+first sitting two lines above a declaration that requires it. `SortConfig.order` is
+required on the interface, on the zod mirror, and on `@objectstack/spec`'s
+`SortItemSchema`, which refuses an entry without it. An author following that
+prescription wrote metadata the spec rejects. objectui#8767's contract review routed
+both sentences to this card by name; comments only, no behaviour.
--- a/packages/core/src/utils/__tests__/sort-query.test.ts
+++ b/packages/core/src/utils/__tests__/sort-query.test.ts
@@ -27,7 +27,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { convertSortToQueryParams, resetRetiredSortSpellingReports } from '../sort-query';
+import { convertSortToQueryParams, normalizeSortEntries, resetRetiredSortSpellingReports } from '../sort-query';
 
 /** The retired spelling, reached the only way it still can be: at runtime. */
 const asRuntimeValue = (value: unknown) => value as unknown as Parameters<typeof convertSortToQueryParams>[0];
@@ -151,5 +151,113 @@ describe('convertSortToQueryParams — the retired string clause (objectui#8221)
     // disconnected mock.
     expect(convertSortToQueryParams(asRuntimeValue('   '))).toBeUndefined();
     expect(errorSpy).toHaveBeenCalledTimes(1);
+  });
+});
+
+/**
+ * `normalizeSortEntries` — the decision `convertSortToQueryParams` was built
+ * on, lifted out so a block that sends a DIFFERENT wire shape can share it
+ * (objectui#8973).
+ *
+ * `object-grid` sends a `"field order"` join string, not this module's map, and
+ * moving it onto the map is route B on objectui#8767 — declined by the
+ * maintainer on 2026-09-10 pending its own card. It re-implemented the "which
+ * entries survive, what does a missing `order` mean" decision privately and got
+ * it wrong (`$orderby: 'name undefined'`). Exporting the decision without the
+ * map is what lets the two agree without the shape moving.
+ *
+ * The second describe below is the load-bearing half: the map builder is now a
+ * projection of this function, so every case must answer exactly what it
+ * answered when it was a single loop. A refactor that quietly changed the map
+ * would be a wire change in `object-calendar`, `object-gantt`, `object-map`,
+ * `object-timeline` and `record:line_items` at once.
+ */
+describe('normalizeSortEntries (objectui#8973)', () => {
+  it('fills a missing `order` with `asc` rather than dropping the key', () => {
+    expect(normalizeSortEntries([{ field: 'name' }])).toEqual([{ field: 'name', order: 'asc' }]);
+  });
+
+  it('preserves authored order and keeps duplicates as separate entries', () => {
+    // The map projection collapses these; the entry list deliberately does not,
+    // because a join-string caller must be able to see what it was handed.
+    expect(
+      normalizeSortEntries([
+        { field: 'stage', order: 'desc' },
+        { field: 'stage', order: 'asc' },
+      ]),
+    ).toEqual([
+      { field: 'stage', order: 'desc' },
+      { field: 'stage', order: 'asc' },
+    ]);
+  });
+
+  it('skips entries that name no usable field', () => {
+    expect(normalizeSortEntries([{ order: 'desc' }, { field: '' }, { field: 'name' }])).toEqual([
+      { field: 'name', order: 'asc' },
+    ]);
+  });
+
+  it('normalizes a garbage direction to `asc` instead of forwarding it', () => {
+    expect(normalizeSortEntries([{ field: 'name', order: 'DESCENDING' as never }])).toEqual([
+      { field: 'name', order: 'asc' },
+    ]);
+  });
+
+  it('answers `undefined`, never `[]`, so a caller can omit the query key', () => {
+    expect(normalizeSortEntries([])).toBeUndefined();
+    expect(normalizeSortEntries([{ order: 'desc' }])).toBeUndefined();
+    expect(normalizeSortEntries(undefined)).toBeUndefined();
+    expect(normalizeSortEntries(null)).toBeUndefined();
+  });
+
+  it('refuses a retired string clause with the SAME reporter, not a second one', () => {
+    const spy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    try {
+      resetRetiredSortSpellingReports();
+      expect(normalizeSortEntries(asRuntimeValue('name desc'))).toBeUndefined();
+      expect(spy).toHaveBeenCalledTimes(1);
+      expect(String(spy.mock.calls[0][0])).toContain('objectui#8221');
+    } finally {
+      spy.mockRestore();
+    }
+  });
+});
+
+describe('convertSortToQueryParams — REGRESSION PIN: the refactor moved no behaviour', () => {
+  /**
+   * Expected values were captured from the PRE-REFACTOR implementation (the
+   * single map-building loop) and are written out as LITERALS on purpose.
+   * Deriving them from `normalizeSortEntries` would only prove the projection
+   * is self-consistent — it would pass just as happily if both halves had moved
+   * together, which is exactly the regression that matters: this map is the
+   * `$orderby` of `object-calendar`, `object-gantt`, `object-map`,
+   * `object-timeline` and `record:line_items`.
+   */
+  const CASES: Array<[string, unknown, Record<string, 'asc' | 'desc'> | undefined]> = [
+    ['fully specified', [{ field: 'stage', order: 'asc' }, { field: 'amount', order: 'desc' }], { stage: 'asc', amount: 'desc' }],
+    ['order omitted means asc', [{ field: 'name' }], { name: 'asc' }],
+    ['field omitted is skipped', [{ order: 'desc' }], undefined],
+    ['empty field is skipped', [{ field: '' }], undefined],
+    ['duplicate field collapses last-wins', [{ field: 'stage', order: 'desc' }, { field: 'stage', order: 'asc' }], { stage: 'asc' }],
+    ['empty array yields undefined', [], undefined],
+    ['array of strings yields undefined', ['name desc'], undefined],
+    ['null entry is skipped', [null], undefined],
+    ['garbage direction normalizes to asc', [{ field: 'name', order: 'sideways' }], { name: 'asc' }],
+    ['undefined', undefined, undefined],
+    ['null', null, undefined],
+    ['a number', 42, undefined],
+    ['a bare unwrapped node', { field: 'name', order: 'desc' }, undefined],
+  ];
+
+  it.each(CASES)('%s', (_label, input, expected) => {
+    const actual = convertSortToQueryParams(asRuntimeValue(input));
+    expect(actual).toEqual(expected);
+    // Key ORDER is part of this map's contract (the first test in this file
+    // pins it), and `toEqual` does not compare it.
+    expect(actual && Object.keys(actual)).toEqual(expected && Object.keys(expected));
+  });
+
+  it('CONTROL — the pin can fail: a deliberately wrong expectation is rejected', () => {
+    expect(convertSortToQueryParams([{ field: 'name' }])).not.toEqual({ name: 'desc' });
   });
 });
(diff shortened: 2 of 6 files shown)


os-steve:

**Verdict: land.** Two observations below; neither blocks, one becomes a card.

Clause-② declared **yes** on the card (objectui#8973 comment 5624157746) — `@object-ui/core` gains `normalizeSortEntries` and `NormalizedSortEntry`, and `sort: []` (which `SortItemSchema` **accepts**) changes what reaches the wire. Per objectstack#17285 the contract-review tier is not this lane's and the lane seat's own default-tier review plus the gates is the review of record, so this comment is it.

### ⭐ The reconciliation is the right answer, and it was not the one I dispatched

My ZONE 1 fenced the wire shape and told you the shared-sink route was out if it required moving it. You found the third option instead: lift the *decision* (`which entries survive, what does a missing order mean`) out of `convertSortToQueryParams`, make the map builder a projection of it, and let `ObjectGrid` build its join string from the same decision. Triage's actual preference — one operation, one implementation — is satisfied **for the operation that genuinely had two copies**, and the shape the maintainer's declination protects never moves. That is better than either thing I offered.

### The equivalence, checked by me rather than by the pin alone

The pin is well built — literals captured from the pre-refactor implementation, explicitly *not* derived from `normalizeSortEntries` (deriving them would only prove self-consistency), plus a key-**order** assertion because `toEqual` does not compare it, plus a "this pin can fail" control. But a Clause-② review owes its own reading, so I walked both implementations against `origin/main`:

| input class | old | new | same? |
|:--|:--|:--|:--|
| falsy | `undefined` | `undefined` | ✓ |
| string (retired) | report, `undefined` | report inside `normalizeSortEntries`, `undefined` | ✓ — one report per call, not two |
| array, N usable | map, dupes last-wins, first-mention position | list → projected, same last-wins, same position | ✓ |
| array, 0 usable | `Object.keys(out).length > 0` false ⇒ `undefined` | `out.length > 0` false ⇒ `undefined` | ✓ |
| non-array, non-string | `undefined` | `undefined` | ✓ |

The one case that *could* have diverged is a non-empty entry list projecting to an empty map. It cannot occur: every surviving entry carries a non-empty string `field`, so every entry produces a key ⇒ **list non-empty ⟺ map non-empty**. The refactor is behaviour-preserving.

### ⚠️ Consumer enumeration — mine, and it is larger than the docblock says

The regression pin's docblock names five surfaces riding on this map (`object-calendar`, `object-gantt`, `object-map`, `object-timeline`, `record:line_items`). Measured on `origin/main` with the control lit, there are **eight call sites across eight files**:

`app-shell/src/utils/deriveRelatedLists.ts:161` · `plugin-calendar/src/ObjectCalendar.tsx:520` · `plugin-form/src/LineItemsPanel.tsx:156` · `plugin-gantt/src/ObjectGantt.tsx:720` · **`plugin-grid/src/ObjectGrid.tsx:1892`** · `plugin-map/src/ObjectMap.tsx:796` · `plugin-timeline/src/ObjectTimeline.tsx:312` · `plugin-view/src/ObjectView.tsx:1000`

Three are unnamed: `deriveRelatedLists`, `ObjectView`, and — the interesting one — **`ObjectGrid`'s own call**, the bare side-effect invocation that exists purely so objectui#8767's refusal still reports on the string arm. ⛔ Not a defect: the pin covers the *function*, so all eight are protected whichever the prose names. But the docblock's job is to tell the next person how much rides on that pin, and it currently understates it by three — including one in the file this PR edits.

### ⚠️ Observation that does not block: this trades a LOUD failure for a SILENT one on one row

The retired array-of-strings row changes failure mode:

- **before** — `['name desc']` lowered to `$orderby: "undefined undefined"`, which `normalizeSortNodes` answers `400 INVALID_QUERY`. The list fails **visibly**.
- **after** — nothing survives, `$orderby` is omitted, and the grid renders **unordered with no message**. Your test asserts that silence deliberately (`expect(errorSpy).not.toHaveBeenCalled()`).

Declining to add a diagnostic was correct — widening objectui#8767's landed refusal to a second retired form is that card's business and my ZONE 1.4 fenced it. And a silent unordered list beats a list that does not load. But it *is* a real trade in the direction this card's own thesis calls worse, and it strengthens the case for the follow-up rather than closing it. Recorded here, not filed separately: it is the same territory as the diagnostic question below.

### The two open questions, ruled

**1. The `reportRetiredSortSpelling` diagnostic still tells authors `order` is optional.** Your recommendation was C, held back only because it is already a carried nit on objectui#8767's comment 5620176008. **Ruling: C, and file it.** A nit carried in a comment on a landed PR is not tracked work — the same reason this lane holds that a ruling naming a dependency has not created one until a card carries it machine-readably. The sentence is false against all three declared faces and prescribes metadata the spec rejects, which is exactly this card's own defect class pointed at the diagnostic instead of the lowering; and rewording a shipped user-facing string is behaviour-visible in a way your two docblock fixes are not, so it deserves its own review rather than a tail commit here. I will file it.

**2. Clause-② direction.** Ruled **yes**, on your reasoning — see the claim comment. Your option B is a real argument but proves too little: "`[]` used to send an empty `$orderby` and now sends none" is a wire change on an input the spec *accepts*.

### ⚠️ C3 — disclosed rather than papered over

With the claim comment posted, `check-clause2-carriers --pair 9028` now reads the declaration but fires **C3**: the declaration says yes while `needs:contract-review` was never hung on either carrier, so "nothing downstream is holding the door."

I measured what that label actually does here, with the control lit: **no workflow in objectui reads `needs:contract-review`.** It is PM tooling state, not a CI gate. And per objectstack#17285 the tier it summons does not serve this lane at all — so hanging it would bind the door to a reviewer who does not exist for `domain:ui` @ objectui, which is precisely the permanent-false-state objectui#8568 has been stuck in since this morning.

⇒ I am **not** hanging it, I am not clearing anything, and I am recording C3 openly here instead. This is now the **second** instance of one governance question and it goes to the maintainer joined with objectui#8568. ⛔ I am not resolving it myself in either direction.

### Scope and gates

The `packages/types` docblock (comments only) goes past my ZONE 1's letter; you flagged it and the routing is objectui#8767's contract review naming this card. Accepted, and the separate commit was the right call.

Gates read on this head: 35 checks, 30 green, 5 test/typecheck jobs still running at review time. Landing on every-check-green, not on this comment.

---

os-steve:
## ⚠️ CI red — `Test (shard 2/4)`. Established as NOT this PR's, with the mechanism, not a shrug.

**Failing check:** `Test (shard 2/4)`, completed 19:28:27Z. **1 failed | 9653 passed | 1 skipped.**

```
FAIL packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx
  > objectui#8920 — every ObjectGrid path honours a `format` hint
  > P: compound-cell prefix badge still draws the select renderer (must-not-change pin)
  at formatHintedColumnRenderer-8920.test.tsx:262:33
```

### Why it is not this PR's — measured, not asserted

**1. This PR cannot reach that code path.** Every line this PR changes in `ObjectGrid.tsx` lives inside the `if (schemaSort) { … } else if (schema.defaultSort) { … }` chain. The failing test's file declares **no `sort` and no `defaultSort` anywhere** — grepped, zero hits. Neither arm fires for that grid, so the executed path is byte-identical before and after this diff. The only other change to the file is one added import specifier, which cannot affect cell rendering.

**2. The assertion is racy, and it is the only one in the test that is.** Lines 258 and 260 both `await waitFor(...)` — first for `'Alice'`, then for the badge label `'New'`. Line 262 is a **bare `expect`**, and it reads a *different column*'s `tel:` href:

```ts
// …and the hinted column in the same grid is still promoted.
```

`telHrefs(container)` returned `[]` — not a wrong href, **no anchors at all** — which is what an un-awaited render looks like, not what a broken renderer looks like.

**3. It is intermittent, which rules out a deterministic break.** The same file was green on objectui#9010's own run (34/34) and has been through the merge queue since. A test that fails on one run and passes on others is a race; a renderer this PR broke would fail every time.

⇒ Cause: objectui#8920's pin, landed 18:17Z, asserts one column's async output while awaiting only another's.

### What I am doing about it

- ⛔ **Not skipping, disabling, or quarantining anything** — and not touching that file from this PR.
- **Filed as its own card** with the mechanism above, so it gets a real fix (await the assertion the way its two neighbours are awaited) rather than being papered over. It will bite every future `plugin-grid` PR until it is fixed, so it is worth a card rather than a re-run and a shrug.
- Spending **one** re-run of the failed job, which is what this lane permits for a failure established as not the PR's. If it comes back red on the same assertion I will treat that as new evidence and re-diagnose rather than re-run again.

This PR's own gates are otherwise complete: 34 of 35 green, including `Type Check`, `Lint`, `Build & E2E` and shards 1, 3 and 4.

---

PR #9021, opened 2026-09-10
fix(types): ObjectKanbanSchema.groupBy is optional, as the protocol declares it (objectui#8990)

os-warren:
Fixes #8990

`@objectstack/spec` declares `groupBy: z.string().optional()` on `ObjectKanbanPropsSchema`; both published objectui faces required it. objectui was therefore **narrower than the protocol** on a published key — it refused an `object-kanban` document the protocol accepts, and that document could not be annotated with its own type. Direction per the maintainer principle in force (recorded verbatim, untranslated): 「我们的项目以 objectstack 协议为准，文档应该以实际实现为准。协议不正确的应该先修改协议。」

**This is a widening.** Nothing that validated before stops validating. An authored `groupBy` is still declared and still enforced (a non-string lane key is refused exactly as before); only its absence is newly admitted. Every flipped pin gained a CONTROL so that *optional* cannot decay into *unjudged*.

## Rework after contract review (`VERDICT: REWORK`, [record](https://github.com/objectstack-ai/objectui/pull/9021#issuecomment-5623901848))

**R1, blocking — the published docs contradicted both faces.** Four hand-written sites still asserted the pre-change contract, and **two of them named this very card as the pending change**: the PR closed the card while leaving the docs telling readers it was still open. AGENTS.md #2 ("Not done until docs reflect the code") is binding, and no CI gate catches this class — only review does. All four now state *optional* **and** the measured lane-less behaviour, and both "tracked as objectui#8990" sentences are gone:

| site | was | now |
|---|---|---|
| `content/docs/api/schema-reference.md` property table | "**Required.**" | optional; notes a lane-less board holds no cards |
| `content/docs/api/schema-reference.md` columns callout | arm "inert on this block", "`groupBy` is required here", "tracked as objectui#8990" | arm applies to a lane-less board and is live there; raw strings vs picklist labels; zero cards |
| `content/docs/plugins/plugin-kanban.mdx` callout | same claim + tracking sentence | same correction |
| `packages/plugin-kanban/README.md` | "`type` and `groupBy` are required" + inline `// required` | record source required, `groupBy` optional but what makes lanes hold cards |

⚠️ I also corrected **two further README sites the review did not name** (`:48`, `:109`) that made the same requiredness claim, and added a line to the schema-reference record-source paragraph making explicit that the record-source rule is **separate** from the lane key and unaffected — that refusal survives and is the easiest thing to conflate with this change.

[...]

PR #9044, opened 2026-09-10
fix(plugin-kanban): key the leftover sweep the way the injection reads (objectui#8993)

### LINKED ISSUE #8993: bug(plugin-kanban): bucketCardsIntoColumns double-buckets every record when a lane id is not a string — knownIds holds raw ids while the group keys are always strings
Found by the `os-dev` seat during the contract-review rework of objectui#8913 (PR #8989), and reproduced independently of the review that first reported it. ⛔ Not repaired there: objectui#8913 is an authoring-face declaration, and this is a renderer behaviour change with its own blast radius.

## The defect

`bucketCardsIntoColumns` (`packages/plugin-kanban/src/index.tsx`) decides lane membership twice, with two different key types, and they disagree when a lane `id` is not a string:

- **injection** reads `groups[col.id]`. A JS property read coerces the key, so a lane `{ id: 1 }` correctly picks up the group stored under `'1'`.
- **the leftover sweep** builds `const knownIds = new Set(columns.map((col) => col.id))` from the RAW value and then filters `Object.keys(groups)` — which are ALWAYS strings — with `!knownIds.has(key)`. `new Set([1]).has('1')` is `false`.

⇒ every record whose group value belongs to a numeric lane is injected into that lane **and** swept into the trailing `Uncategorized` lane. The card renders **twice**.

## Reproduction — the real function, with a control leg

Runnable as a vitest file anywhere under `packages/plugin-kanban/src/`; it calls the exported function directly, mounts nothing:

```ts
import { describe, it, expect } from 'vitest';
import { bucketCardsIntoColumns } from './index';

const summarise = (lanes: any[]) =>
  lanes.map((l) => `${String(l.id)}:${(l.cards || []).map((c: any) => c.id).join('+')}`).join(', ');

describe('numeric lane ids double-bucket', () => {
  it('NUMERIC', () => {
    expect(summarise(bucketCardsIntoColumns(
      [{ id: 1, title: 'One' }, { id: 2, title: 'Two' }],
      [{ id: 'r1', status: 1 }, { id: 'r2', status: '2' }],
      'status', undefined, 'Uncategorized',
    ))).toBe('1:r1, 2:r2');            // MEASURED: '1:r1, 2:r2, __uncolumned__:r1+r2'
  });

  it('STRING control', () => {
    expect(summarise(bucketCardsIntoColumns(
      [{ id: 'one', title: 'One' }],
      [{ id: 'r1', status: 'one' }],
      'status', undefined, 'Uncategorized',
    ))).toBe('one:r1');                // MEASURED: 'one:r1' — clean, so the reading above is about the id TYPE
  });
});
```

Measured on `claude/issue-8913-object-kanban-columns-declared` at `aa467bc` (the function is byte-identical to `origin/main`):

```
NUMERIC lane ids -> 1:r1, 2:r2, __uncolumned__:r1+r2
STRING control   -> one:r1
```

The control leg is what makes the first line a reading rather than an artefact of the harness.

## Why nothing catches it today

- `KanbanColumn.id` and `KanbanColumnSchema.id` are both `string`, and after PR #8989 `ObjectKanbanSchema.columns[].id` is `string` too — so a numeric lane id is now refused at the authoring boundary on every published face. **That closes the AUTHORED path, not the defect.**
- The lanes the renderer materialises itself are built with `id: opt.value` from the object definition's picklist options, and `id: String(g)` from the data. So a picklist whose option `value` is a **number in real metadata** reaches this function with a numeric `id` through a path no schema guards. Whether such metadata exists in production was NOT measured here.
- `index.bucket.test.ts` covers the leftover sweep (#2792) with string ids only.

Key both sides the same way — `new Set(columns.map((col) => String(col.id)))`, and read `groups[String(col.id)]` — so injection and the sweep agree. Any fix wants the two probes above kept as a pin, because the failure is silent: the board renders, the totals just do not reconcile.

⚠️ It is a renderer behaviour change: today a numeric-id board shows every card twice, after a fix it shows each once. That is a visible difference for anyone whose metadata already hits this.

## Related reading, recorded but NOT part of this card

`effectiveColumns` (`packages/plugin-kanban/src/ObjectKanban.tsx`) dispatches on `columns[0]` alone, so a MIXED array is mishandled in both orders. Measured with `[{ id: 'done', title: 'Done' }, 'todo']`: `done:r2, undefined:, __uncolumned__:r1` — a blank lane whose own keys are `["0","1","2","3","cards"]` (the string spread character by character) and the `todo` record in "Uncategorized". PR #8989 now refuses a mixed array at the authoring boundary, so there is no authored path to it; it is written down here only so the next reader of this function does not rediscover it.

Refs: objectui#8913 / PR #8989 · objectui#2792 · objectui#8990.

Filed by Claude Code for the `os-dev` seat on objectui#8913; the session reference is `session_01Jmxdo7bmeqCQHLSfmLVX9w`.

---

os-steve:
Fixes #8993

## The defect

`bucketCardsIntoColumns` decided lane membership **twice**, with two key types:

- **injection** reads `groups[col.id]`. A property read coerces its key, so a lane `{ id: 1 }` correctly picks up the group stored under `'1'`.
- **the leftover sweep** (objectui#2792) built `knownIds` from the **raw** `col.id` and filtered `Object.keys(groups)`, which are always strings. `new Set([1]).has('1')` is `false`, so every record the injection had just placed was swept a second time into the trailing "Uncategorized" lane.

⇒ A board whose lane ids are not strings rendered **every card twice**. Silent by nature: the board draws, only the totals fail to reconcile.

## The repair

One site, on the **sweep** side: the set now holds the same key spelling the property read uses.

⛔ The injection line is byte-identical. Making that read strict would pass the reported leg while breaking the coercion lanes depend on — and the string control could not catch it, which is exactly why triage fenced it.

One deviation from the card's fix sketch, and it is measured rather than assumed: a **symbol** is the one id a property read does not stringify, so it is kept as-is instead of being pushed through `String()`, which throws on symbols. `Object.keys` never yields a symbol, so such a lane keeps the reading it has today (record to "Uncategorized") rather than turning a silent mis-bucket into a thrown TypeError inside render.

## Acceptance — both legs, plus the class

Triage's two legs are rows 1 and 2 of `laneIdCoercion-8993.test.ts`. The string leg is a **live control**: it is green on both sides of the ablation below.

Measured at the real function, before and after the repair:

| lane id | before | after |
|---|---|---|
| `1`, `2` (the reported leg) | `1:r1, 2:r2, __uncolumned__:r1+r2` | `1:r1, 2:r2` |
| `'one'` (the control) | `one:r1` | `one:r1` |
| numeric id reached via the label map | `1:r1, __uncolumned__:r1` | `1:r1` |
| `0` / `false` (falsy, so the label map skips them) | doubled | single |
| `true` / `NaN` / array / plain object | doubled | single |
| `null` lane with a **null record value** | `null:, __uncolumned__:r1` | unchanged |
| lane with no `id` and an **absent record value** | `undefined:, __uncolumned__:r1` | unchanged |
| a lane declared with the sentinel spelling | one lane, its own record | unchanged |
| symbol id | `Symbol(lane):, __uncolumned__:r1` | unchanged, still no throw |

The last four rows are the "no new collision" block. Coercing a key can **mint** a collision where there was none, so the boundary is asserted rather than argued: `String(null)` is `'null'` and `String(undefined)` is `'undefined'`, and neither may swallow the empty key that a null or absent VALUE produces (`String(item[groupBy] ?? '')`). The general reason there is nothing to mint: the sweep's only job is to answer *did the injection already claim this key*, and the injection makes that claim under exactly this coercion.

Row 5 is the control that separates a repair from a deletion — an unmatched record must still reach the trailing lane. Rows 1 to 4 would all pass if the sweep were simply removed, and that would resurrect objectui#2792.

## Ablation, from the committed fix

Prediction written before the run: 12 red / 5 green, the 5 being the string control and the four boundary rows.

```
Tests  12 failed | 5 passed (17)
```

Measured red set matched the prediction exactly, with no miscount: rows 1, 4, 5 and all nine class rows red; the string control and all four boundary rows green. The mutation was proven on disk before the run (HEAD blob `2fc0ea48` -> mutated `7ec8e647`, pre-repair line count 1, symbol arm count 0) and the restore proven after it (`git diff HEAD` empty for the path, blob back to `2fc0ea48`), with a `trap` on absolute paths. No `dist` is involved: the pin imports the source relatively, so there is no stale-build leg to fool it.

## The dispatch's mechanism assumptions, measured

- **Published export — CONFIRMED.** `export function bucketCardsIntoColumns` in `packages/plugin-kanban/src/index.tsx`, which is the vite lib entry (`build.lib.entry`) that produces `dist/index.js` / `dist/index.d.ts`, and `files: ["dist"]` ships it. 20 export statements in the barrel, matching the lit control; the two-way probe reads 1 for this symbol and 0 for the module-local `labelToColumnId`.
- **The `knownIds` construction and the `.filter` beside it — CONFIRMED,** re-derived by symbol.
- **"The only raw-id-vs-string-key comparison in the file" — CONFIRMED as a census, with one qualifier.** The file has exactly two membership decisions: the property read and this Set, and only the second one fails to coerce. But two further sites hand a **raw** id into a string-keyed world: `labelToColumnId[String(col.id).toLowerCase()] = col.id` stores the raw id as its VALUE, and the group key derived from it is saved only by property-key coercion, the same accident that saves the injection. That path double-bucketed too and is pinned as row 4; keying only the direct path would have left it open.
- **"Numeric is the only non-string case that matters" — FALSIFIED.** The class is *not a string*, not *is a number*: `0`, `false`, `true`, `NaN`, arrays and plain objects all double-bucketed identically, and all are pinned.

## Verification

- `pnpm exec vitest run packages/plugin-kanban/` — 48 files, 285 tests, all passed
- `pnpm exec vitest run packages/types/` — 169 files, 3352 tests, all passed
- `pnpm --filter '@object-ui/plugin-kanban^...' run build` — exit 0 (dependency closure)
- `pnpm --filter @object-ui/plugin-kanban --filter @object-ui/types run type-check` — exit 0, both scripts echoed; `tsc -p tsconfig.test.json --listFiles` shows the new pin inside the typechecked set (1690 files, negative control 0)
- `check-control-bytes` (7234 files), `check-new-line-citations`, `check-changeset-presence`, `check-changeset-no-major` — all exit 0
- eslint over its **own full population** (4755 files, read from `--format json`, not narrowed): every file in this diff reports 0 errors, and none is among the 79 files carrying pre-existing errors
- `check-governed-queue-guard --test` over this file list: NOT GOVERNED

## Cross-lane adjacency

PR objectui#9021 (the objectui#8990 lane) is open on the same package. It does **not** touch `packages/plugin-kanban/src/index.tsx` — no conflict on the repair itself. It does edit both documentation files this PR edits, and in `content/docs/api/schema-reference.md` its hunk covers the block quote whose next-but-one line this PR rewrites, so whichever lands second may need a textual resolution there. Reported, not worked around.

## Acceptance notes (out of scope for this card)

- **filed as objectui#9043** — the same function **throws** when a record's `groupBy` value is an `Object.prototype` member name. Both `labelToColumnId` and the `groups` accumulator are prototype-bearing `{}`: `'toString'` / `'valueOf'` / `'hasOwnProperty'` crash with `acc[key].push is not a function`, while `'constructor'` and `'__proto__'` are grouped under a nonsense key. Measured against a clean control leg, unchanged by this PR. Different mechanism, and its worst leg is a crash rather than a duplicate, so it is a card and not a passenger.
- *noted, not filed:* two lanes whose ids share a **string spelling** (`1` and `'1'`, or two distinct object ids) both claim the same records at the **injection** site — measured `[object Object]:r1, [object Object]:r1`. This PR removes the third copy that used to appear in "Uncategorized" but cannot remove the double claim, which lives on the read side the fence protects. It is unauthorable since objectui#8913 narrowed the lane `id` to a string. Successor: it is recorded in the pin's row 6 block, which the next reader of this function meets.
- *noted, not filed:* `if (col.id)` skips label registration for a falsy lane id, so a lane `{ id: 0, title: 'Zero' }` is reachable by its title but not by an `id` label alias. Measured as behaviour-neutral for bucketing (direct value matching still finds it) and unchanged here. Successor: none.

The `@object-ui/types` half of this diff is prose only: the lane-id `describe()` and the shipped TSDoc stated the double-render as *current* behaviour, which this repair makes false. The authored `id` stays `string` — one declared lane-id type beats two, which is objectui#8913's reason and does not depend on the renderer being broken.

---


os-steve:

Per objectstack#17285, this lane's in-seat default-tier review **plus the gates** is the review of record for a clause-② pair. The contract-review tier is not this lane's. Verdict below; the label paragraph is the part that is not routine.

### What I re-measured myself rather than reading off the report

- **The `packages/types` half is prose only — CONFIRMED.** This is the claim that decides whether the pair is still `Clause-②: yes` *on the renderer's output* or has quietly become a declaration change. `id: string` in `objectql.ts` and `id: z.string()` in `objectql.zod.ts` are byte-identical; only the docblock and the `.describe()` string move. The `yes` stands where the claim comment put it.
- **The repair is on the sweep side only — CONFIRMED.** One statement changed. `groups[col.id]` is untouched, which is the fence triage set and the one the string control could not have enforced.
- **The symbol arm is right, and it is the good kind of deviation.** `String(Symbol())` throws, `Object.keys` never yields a symbol ⇒ pushing symbols through `String()` would have converted a silent mis-bucket into a `TypeError` inside render. Declining the card's own fix sketch, on a measurement, and saying so, is the behaviour this seat wants.
- ⭐ **The census FALSIFIED my brief.** I handed over "numeric is the only non-string case that matters" as an assumption to test, and it is wrong: `0`, `false`, `true`, `NaN`, arrays and plain objects all double-bucketed identically, and the label-map path double-bucketed through a second route I had not identified. All pinned. A dev that had merely satisfied the brief would have shipped a narrower fix and a weaker pin.
- **`pnpm run check:changeset-presence` does not exist as a script name** (`ERR_PNPM_NO_SCRIPT`). The dev read exit 1 as **NOT MEASURED** rather than as a failure or a pass, and re-measured by invoking the script directly. That is the correct reading and it is the one that gets skipped most often.

### The cross-lane adjacency — resolved, and I got it wrong first

The report flagged that objectui#9021 edits the same block quote in `content/docs/api/schema-reference.md`, and left it as an open item.

I first concluded it was moot, on the grounds that this PR's `base.sha` is `a4b723fab` — the #9021 merge commit. **That was wrong.** The API's `base.sha` is the base branch's *current* tip, not the merge-base. Measured: this branch's merge-base with `main` is `681d3f10e` (#9016), and `git merge-base --is-ancestor a4b723fab <head>` answers **NO**. #9021 landed first and this branch predates it, so `mergeable_state: behind` is accurate and the dev's warning was correct to file.

Resolving it properly: `git merge-tree --write-tree origin/main <head>` exits **0** — no textual conflict across all four overlapping files. And the merge is not merely clean, it is *correct*: the merged tree keeps this PR's rewritten lane-`id` sentence and #9021's `groupBy`-optional paragraphs as adjacent, non-contradicting prose. Neither repair is dropped and neither claim is falsified by the other. ⇒ **no resolution needed**; the open item closes on a measurement rather than on optimism.

### ⚠️ `needs:contract-review` — the label was hung against this seat's stated position

The claim comment on objectui#8993 said, in this seat's own words: *"Not hanging it … ⛔ Not resolved by this seat in either direction."* The dev then hung it on both carriers to take the gate from 4 to 0.

⛔ **This seat does not clear it either.** Clearing it would resolve the escalation in one direction on my own authority, which is exactly what I have declined to do six times. ⇒ objectui#8993 joins **objectui#8568** and **objectui#8613** in the hung-on-both state.

**The part worth escalating is not the label. It is that three devs have now given three different answers to the same gate, all in good faith:**

| card | what the dev did | was it wrong? |
|---|---|---|
| objectui#8973 | refused to paper over exit 4, escalated instead | no |
| objectui#8943 | hung the label, then reverted it | no |
| objectui#8993 | hung the label and kept it, **overriding the PM's written position** | no |

Three careful agents, one gate, three incompatible readings. ⚠️ That is not dev error — a rule that three competent readers resolve three ways is **underdetermined**, and no amount of better dispatch briefs fixes it. This is the strongest evidence the escalation has produced, and it supersedes the inferential history limb I recorded on objectui#8943 (whose matched control failed).

**I am still landing this PR**, on the same measured ground as objectui#9037: nothing in objectui reads that label today. `Governed Surface Queue Guard` is **green on this very PR**, and the proof that the guard is absent by construction rather than by luck is objectui#9018 — an open, unstarted `domain:devx` card asking to port exactly such a guard.

### Verdict

**Approved for landing** once the four test shards and Type Check complete. 30 of 35 checks are green, 3 correctly skipped, 5 still running at the time of writing; nothing is armed until all 35 are in. Successor objectui#9043 (the `Object.prototype` key crash) is a genuinely different mechanism with a worse leg, and was right to be a card rather than a passenger.

---

PR #9063, opened 2026-09-10
fix(plugin-gantt): honour filter, sort and the row ceiling on inline `value` data

### LINKED ISSUE #8769: [finding] ObjectGantt's `provider: 'value'` short-circuit never calls `find()`, so a declared `filter` / `sort` / row ceiling is silently ignored — the chart shows EVERY inline row
Measured while landing objectui#8513 (`ValueDataSource`'s object-dialect combinators). Out of that card's scope — different package, different defect shape, different fix — so it is filed rather than folded in. ⛔ Not claimed.

## The mismatch

`packages/plugin-gantt/src/ObjectGantt.tsx`, inside `reload`:

```ts
if (hasInlineData && dataProvider === 'value') {
  if (isCurrent()) {
    setData(dataItems as any[]);
    setRowCeiling({ truncated: false });
  }
  return;
}
```

It returns BEFORE the `effectiveDataSource.find(...)` call a dozen lines below — the call that is the only place `schema.filter`, `schema.sort` and the `NON_GRID_ROW_CEILING_TOP` ceiling are applied. So on a `provider: 'value'` gantt:

- an authored `filter` reaches nothing and **every** inline row is charted;
- an authored `sort` reaches nothing;
- `applyNonGridRowCeiling` never runs, so the objectui#7210 ceiling is not enforced on this provider.

The adapter it skips is `ValueDataSource`, which implements `$filter`, `$orderby`, `$skip`, `$top` and `$select` locally — so the short-circuit is not working around a missing capability. `resolveDataSource` already returns exactly that adapter for this provider, four lines away.

## Failure direction

**Fail-OPEN.** A gantt that declares a filter draws the rows it declared it would hide. That is the opposite direction from objectui#8513 (which returned no rows) and the same direction objectui#8447 was filed about — worth stating because the two are easy to fold together and must not be.

## Why it is not a scope question

objectui#8220 is nearby but different: it says three plugin REGISTRATIONS declare neither `filter` nor `sort` while the renderers lower both onto the query. This card is about the one provider where the renderer does not reach the query at all. objectui#7333 is also nearby and also different — it is the OTHER short-circuit in the same function (the host-`data` one firing on an empty array).

## Suggested shape, not a ruling

The inline branch could seed `setData` from `ValueDataSource` with the same query params the object/api branch builds, rather than from `dataItems` raw. Whether the ceiling should apply to an inline array at all is a real question and is why this is filed rather than fixed in passing.

## Dedup

Run, and the instrument was validated in the same session with a control query that returned a known-present card. `search_issues` (repo-scoped, semantic): the gantt/filter/sort neighbourhood returns objectui#8174, #8220, #7237, #7333, #7334 and objectui#8513 — 18 total, none of them this. Closed cards were in range (the same session's control returned closed objectui#8447 / #8515 / #8600).

---

os-steve:
Fixes #8769

Exit taken: the **preferred** one. Both exits were pre-authorised — apply the keys, or refuse the combination loudly — and the measurement chose the first: `resolveDataSource` already answers this provider with `ValueDataSource`, which implements `$filter`, `$orderby`, `$skip`, `$top` and `$select` over its own array, so there was no capability the inline path could not support and nothing to refuse. **No filter combinator was written here** — objectui#8513's work stays where it is.

## What was wrong

`ObjectGantt`'s `reload` short-circuited the inline provider: it set the authored rows and returned **before** `effectiveDataSource.find(...)`, which is the one site that lowers `schema.filter` onto `$filter`, `schema.sort` onto `$orderby` and the objectui#7210 ceiling onto `$top`. An authored `filter` therefore reached nothing and the chart drew every inline row.

The direction is **fail-open**: the dropped key is the key that NARROWS, so the author was shown more rows than declared, with no diagnostic. ⛔ Nothing was exposed — the rows are already in the authored schema. What was wrong is that the chart answered a wider question than the one asked.

## The two-sided reading is the finding

A one-sided reproduction cannot separate "the filter was ignored" from "the filter matched everything", so the same five rows and the same `status = open` filter were rendered twice, with **one `ValueDataSource` serving both sides** — the matcher is literally the same object, and the only remaining variable is which branch of `reload` ran.

| path | rows drawn | ids |
|---|---|---|
| fetching (`objectName` + context adapter) | 3 | `1,3,5` |
| inline (`provider: 'value'`) — before | **5** | `1,2,3,4,5` |
| inline (`provider: 'value'`) — after | 3 | `1,3,5` |

## What changed

The inline branch is deleted; `provider: 'value'` falls through to the same adapter query as every other provider. The now-dead `dataItems` binding goes with it — `reload` no longer reads the payload, and authored items still reach it through the deep-compared `dataConfig` that builds the adapter, so the objectui#6592 memo-identity reasoning is preserved and re-stated where the binding used to be.

## The assumptions, and how each resolved

**A — does the repair inherit objectui#9001 / PR objectui#9049's new `FilterOperatorError` throw? NO, measured.** `convertFiltersToAST` is reached only by `@object-ui/data-objectstack` on the wire path. `ValueDataSource` never imports it — its imports are `@object-ui/types`, `@objectstack/spec/data` and its own `batchTransaction` module — and it carries its own local refusal for that comparand, which excludes the row and logs once. So an inline chart whose filter carries such a comparand renders with zero rows and a console line, it does **not** throw at render. This was a deliberate decision, not an accident of routing: an in-memory matcher decides about one row at a time, while the converter is deciding whether to send a query at all, and that asymmetry is the adapter's own stated refusal shape. Pinned from the outside by `refusedComparandDoesNotThrow`.

**B — is `find(...)` really the only site applying the three? HELD.** Verified by symbol across the package: `applyNonGridRowCeiling` appears at exactly one call site, and `$filter` / `$orderby` at exactly one each, all inside `reload`.

**C — are the ceiling and filter/sort independent? NO, and the order is now asserted.** A ceiling computed before filtering is a different number from one computed after. **Filter first, ceiling second**, matching the fetching path exactly: there `$filter` and `$top` travel in one query and every backend filters before it limits, so any other order would have made the two paths disagree again in a subtler way. `ceilingOrder` pins it from the observable side — 2,400 inline rows of which 800 match draw all 800 with **no** footnote, which a ceiling-then-filter implementation could not produce.

**D — is `data: { provider: 'value' }` the only spelling? NO — there are TWO.** The ruled three-rung record-source ladder also wraps `staticData` into a `provider: 'value'` config, so an author lands on this path either way. `staticDataSpelling` covers the second one. (A bare `data: []` array does **not** reach here on the gantt — this renderer returns it verbatim rather than normalising it, so it falls through to the fallback adapter; and the host `data` React prop is a different branch, objectui#7333's.)

**E — are the card's `path:line` anchors stale? ASSUMED SO and not used.** Everything was re-derived by symbol, and there is no line citation anywhere in this diff.

**F — is today's behaviour silently wrong? HELD.** Nothing warned. The short-circuit set state and returned with no console line, so the card's "no diagnostic" framing is correct.

## A standing pin is INVERTED here, deliberately

`ObjectGantt.rowCeiling-7210.test.tsx` asserted "an inline `value` data set is never capped by us, and never footnoted", and `ObjectGantt.tsx` carried the matching comment. That case pinned the **short-circuit**, not objectui#7210 ruling a-prime:

- the ruling's own budget is measured in **DOM elements per record**, and its measurement table was taken over "real child views, **inline `value` provider**" — an inline row costs the browser exactly what a fetched row costs;
- the ruling text names no provider carve-out; what it forbids is a view crossing the ceiling **quietly**, which is why the footnote is the load-bearing half.

So the case is inverted in place with that reasoning recorded on it, and the exemption that genuinely remains — rows a **host** hands down through the `data` React prop, which we did not query and have no total for — gets its own case beside it. If a maintainer reads the exemption as ruled rather than incidental, this is the one hunk to revert; it is isolated.

## Behaviour a consumer can notice

- An authored `filter` / `sort` now narrows and orders inline rows.
- Past 2,000 drawn rows an inline chart draws 2,000 and shows the footnote naming both numbers.
- Inline rows now reach the chart as the adapter's own deep copy rather than as the authored array's object identities. `ViewData.items` is serializable metadata, so every value the contract admits survives that copy unchanged; only reference comparison against the authored array changes, and `id` equality is the replacement.
- An inline gantt's write-backs (drag, dependency edit, inline edit, delete) now survive the reload that follows them, because the reload reads the adapter that took the write instead of re-reading the authored array over the top of it. Reads and writes previously disagreed about which array was the source of truth; they now agree.

## Ablation — predicted before running, then measured

Prediction written down first: restoring the short-circuit turns 8 cases red and leaves 4 green (`control`, plus the above-ceiling, below-ceiling and host-`data`-prop ceiling cases).

Measured: **8 failed, 4 passed** — exactly the predicted split, no miscount.

On-disk proof, both directions: the injected marker went 0 to 1 and the removed marker 1 to 0 in the same file; blob `e6e1b46a` became `fba326a2`. Restore proved by hash equality back to `e6e1b46a` **and** by an empty `git diff HEAD`, from a trap with absolute paths. No `dist` is involved on this path — the pin imports the mutated module relatively, as source.

The `control` case is the one that makes the rest readable: an inline chart with no filter, no sort and under the ceiling draws the same five rows in the same order on **both** legs, which is what separates this repair from "the inline path now drops rows".

## Verification

All at `05d946890`, the final commit.

- `pnpm exec vitest run packages/plugin-gantt/` — **66 files, 514 tests passed**.
- `pnpm --filter @object-ui/plugin-gantt type-check` — clean. The new pin is **proven inside** the typechecked set rather than assumed into it: its first run failed with `TS7006` on line 269 of the new file, so `tsc -p tsconfig.test.json` demonstrably reads it.
- Dependency-closure build `pnpm --filter '@object-ui/plugin-gantt^...' build` — green (13 packages).
- Gates run green: control bytes, new cross-file line citations, the three vi-mock gates, element-data-source declaration, unreferenced sources, doc fences, shell-escape residue, comment-mask corpus, changeset presence, changeset-no-major.
- **Lint, narrowed and the narrowing declared.** eslint over the `plugin-gantt` subtree: **94 files** counted from `--format json`, **0 errors on every file this PR touches**. The narrowing is a measurement, not a skip: the linted population is read from `eslint.config.js`'s own top-level `ignores`, which excludes only build output (`dist`, `.next`, `node_modules`, `public`, `.source` and three named artefacts) and not one source file; and the config sets **no** `parserOptions.project` and no `projectService`, so linting is not type-aware and this diff cannot move the verdict on any file it did not touch. The 2 errors the run does report are in `packages/plugin-gantt/demo/main.tsx`, which this PR does not touch — pre-existing.
- **NOT MEASURED, with reasons:** `check:readme-exports` — the gate itself printed "the population COLLAPSED, this run proves nothing" because most packages have no `dist` in this worktree; after building `plugin-gantt` its own README is judged and reports zero complaints. `check:sdui-registration-pins` — exit 2, its own text says a run with no console build "has measured nothing". Both need a full-repo build, which is CI's run.

## Acceptance notes

- **filed as objectui#9061** — `ObjectCalendar` and `ObjectMap` carry the same inline short-circuit, hand-copied. Fenced out of this PR by the brief and filed rather than fixed. The dedup behind it was controlled: a repo-scoped REST listing over `state=all`, 400 rows, grepped locally, with objectui#8769 itself as the positive control proving the enumeration reached this neighbourhood.
- **noted, not filed** — `packages/plugin-gantt/demo/main.tsx` carries 2 eslint errors on `main` (`react-hooks/purity`, `no-console`). Whoever runs this package's own lint task hits them; not filed because the repo's actual wired lint invocation cannot be confirmed from here, so "this is red for everyone" would be a guess.
- **noted, not filed** — `ObjectGrid` and `ObjectTree` also branch on the inline provider, but their surrounding machinery is different and neither was measured here. objectui#9061 deliberately names only calendar and map, the two that are literal copies of this shape.
- Not touched in either direction: the contract-review label. The claim on this card declares Clause-2 yes and the carriers gate is expected to report C3; that is a disclosed, escalated governance question, and per the brief it is neither cleared nor hung by this seat.

---

### DIFF
--- a/.changeset/8769-gantt-inline-query-keys.md
+++ b/.changeset/8769-gantt-inline-query-keys.md
@@ -1,0 +1,48 @@
+---
+'@object-ui/plugin-gantt': minor
+---
+
+Honour `filter`, `sort` and the platform row ceiling on a gantt's inline
+(`provider: 'value'`) data (objectui#8769).
+
+**The defect was fail-open.** `ObjectGantt`'s `reload` short-circuited the
+inline provider — it set the authored rows and returned BEFORE the adapter
+query, which is the one site that lowers `schema.filter` to `$filter`,
+`schema.sort` to `$orderby` and the objectui#7210 ceiling to `$top`. So an
+inline gantt that declared a `filter` drew **every** authored row, with no
+diagnostic. The key that was dropped is the key that NARROWS, which is why this
+matters: the chart answered a wider question than the author asked. ⛔ Nothing
+was exposed that was not already in the authored schema — this is a
+correctness defect, not a data-access one.
+
+Measured two-sided before the repair: the same five rows and the same
+`status = open` filter drew 3 rows through a fetching data source and 5 rows
+inline, with the same matcher on both sides.
+
+**What changed.** The inline branch is gone; `provider: 'value'` now goes
+through `resolveDataSource`'s `ValueDataSource` like every other provider,
+which already implements `$filter` / `$orderby` / `$skip` / `$top` / `$select`
+over its own array. No filter combinator was written for this change.
+
+**Behaviour you may notice.**
+
+- An authored `filter` / `sort` now narrows and orders inline rows. Both
+  spellings reach it: `data: { provider: 'value', items }` and `staticData`.
+- The row ceiling now applies to inline rows: past 2,000 drawn rows the chart
+  draws 2,000 and shows the footnote naming both numbers, as it already did for
+  fetched rows. It is applied to the **filtered** set, so a large inline array
+  that a `filter` cuts below the ceiling draws every matching row and stays
+  quiet. Rows a host passes down through the `data` React prop are still never
+  capped — those are not ours to cap.
+- Inline rows now reach the chart as the adapter's own deep copy rather than as
+  the authored array's object identities. `ViewData.items` is serializable
+  metadata, so every value the contract admits survives that copy unchanged;
+  code comparing a row handed to `onTaskClick` against the authored array with
+  `===` would need `id` equality instead.
+- An inline gantt's own write-backs (drag, dependency edit, inline edit,
+  delete) now survive the reload that follows them, because the reload reads
+  the adapter that took the write instead of re-reading the authored array over
+  the top of it.
+- A filter comparand that `@object-ui/core`'s `convertFiltersToAST` refuses does
+  **not** throw at render on this path: the inline matcher is local and never
+  reaches that converter. The row is excluded and the reason is logged once.
--- a/content/docs/plugins/plugin-gantt.mdx
+++ b/content/docs/plugins/plugin-gantt.mdx
@@ -127,8 +127,17 @@ const schema: ObjectGanttSchema = {
   gantt?: GanttConfig,              // Gantt-specific configuration
   viewMode?: 'day' | 'week' | 'month' | 'quarter' | 'year',
   readOnly?: boolean                // disable every edit path
+  filter?: Array<any>,              // JSON-Rules query filter — applies on EVERY provider
+  sort?: Array<{ field, order }>,   // ordering — applies on EVERY provider
 }
 ```
+
+`filter` and `sort` are not object-only keys. They narrow and order inline rows
+(`staticData`, or `data: { provider: 'value' }`) exactly as they narrow and
+order fetched ones, and the platform row ceiling — 2,000 drawn rows with a
+footnote naming both numbers — applies to inline rows too. The ceiling is
+applied to the **filtered** set, so a large inline array that a `filter` cuts
+below the ceiling draws every matching row and shows no footnote.
 
 `onTaskClick` and `className` are **React props on `<ObjectGantt>`**, not schema
 keys — the renderer never reads them off the schema, and a function cannot
--- a/packages/plugin-gantt/README.md
+++ b/packages/plugin-gantt/README.md
@@ -243,6 +243,16 @@ const recordSource = {
 `{ provider: 'value', items }`, `{ provider: 'api', read, write }` or
 `{ provider: 'schema', schemaId }`.
 
+**The provider does not change which query keys apply.** An authored `filter`
+and `sort` narrow and order the rows on **every** provider, inline ones
+included, and the platform row ceiling (2,000 drawn rows, with a footnote
+naming both numbers) applies to all of them — inline rows cost the browser what
+fetched rows cost. Inline rows go through the same in-memory adapter the other
+providers go through, so `filter` is evaluated with the same matcher and the
+ceiling is applied to the **filtered** set, never to the raw one. Before
+objectui#8769 the inline provider skipped that query and drew every authored
+row with an authored `filter` silently dropped.
+
 **2. How the fields map — `getGanttConfig`.** Two spellings, checked in order.
 The **`gantt` block wins whenever it is present**, and it is taken WHOLE — the
 flat top-level keys are not merged into it. The flat spelling is read only when
--- a/packages/plugin-gantt/src/ObjectGantt.inlineQueryKeys-8769.test.tsx
+++ b/packages/plugin-gantt/src/ObjectGantt.inlineQueryKeys-8769.test.tsx
@@ -1,0 +1,272 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8769 — a `provider: 'value'` gantt honours the three query keys the
+ * fetching path honours: `filter`, `sort`, and the objectui#7210 row ceiling.
+ *
+ * ## What was wrong
+ *
+ * `reload` short-circuited the inline provider (`setData(dataItems)`; return)
+ * BEFORE `effectiveDataSource.find(...)`, which is the ONE site that lowers
+ * `schema.filter` to `$filter`, `schema.sort` to `$orderby` and the ceiling to
+ * `$top`. An authored `filter` therefore reached nothing and every inline row
+ * was charted — the fail-OPEN direction: the key that was ignored is the key
+ * that NARROWS, so the author saw MORE rows than declared, with no diagnostic.
+ *
+ * ⛔ Not a data-exposure boundary. The rows are already in the authored schema;
+ * what is wrong is that the chart answers a question nobody asked.
+ *
+ * ## The two-sided reading is the finding
+ *
+ * A one-sided reproduction cannot tell "the filter was ignored" from "the
+ * filter matched everything", so the first case renders the SAME rows and the
+ * SAME filter twice — once inline, once through a context adapter — and reads
+ * the DISAGREEMENT. The adapter is a `ValueDataSource` over the same rows, so
+ * the matcher is literally the same object on both sides and the only variable
+ * left is which branch of `reload` ran.
+ *
+ * ## ORDER: filter first, ceiling second (objectui#7210 ruling a′)
+ *
+ * The ceiling is applied to the FILTERED set, matching the fetching path
+ * exactly — there `$filter` and `$top` travel in one query and every backend
+ * filters before it limits. `ceilingOrder` pins it from the observable side: a
+ * set that is over the ceiling BEFORE filtering and under it after draws every
+ * matching row and shows NO footnote. A ceiling-then-filter implementation
+ * would draw at most 2,000 rows there and would still footnote.
+ *
+ * ## The `never capped by us` case this file replaces
+ *
+ * `ObjectGantt.rowCeiling-7210.test.tsx` asserted that an inline set is never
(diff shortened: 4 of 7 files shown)


os-steve:

**Approved — land on all-checks-green.** One thing is flagged upward rather than absorbed; it is named below and it is revertible in one hunk.

### The pin inversion: I read objectui#7210's ruling rather than take the reading on trust

This PR inverts a standing case in `ObjectGantt.rowCeiling-7210.test.tsx` that asserted an inline set is never capped and never footnoted. Inverting a pin that sits on a **maintainer-ruled** surface is not something a dev's argument settles by itself, so I read ruling a′ (objectui#7210 comment `5508048888`) and its implementation report (`5513939384`). The dev's reading holds, on four independent legs:

1. **The ruling's budget is provider-independent.** It was set by measuring **DOM elements per record** — `ObjectTree` binding at ~5.2 elements/record, 10,406 elements at n=2000 — against a budget of *"keep the worst of the four inside ~10,000 elements"*. An inline row costs exactly what a fetched row costs.
2. **The ruling's text carves out no provider.** It says a non-grid visualisation *"is allowed to fetch the whole filtered result set … but the fetch carries a platform-level hard ceiling."* The word is *fetch*, but that is the mechanism available at the time — all four views fetched — not a stated exemption.
3. ⭐ **The direction of this change is toward the ruling's stated hazard, not away from it.** a′'s own words: *"The dangerous direction is silent truncation: a cut-off schedule still looks like a schedule, so the ceiling must never be crossed quietly."* An uncapped inline set of 3,000 rows is not silent truncation — but capping it **with the loud footnote** is precisely the remedy the ruling prescribes. Nothing about this makes a quiet cut-off more likely.
4. **Triage independently ruled it.** objectui#8769's triage says *"Apply `filter` / `sort` / the ceiling on the inline path too"* — the ceiling named explicitly, not inferred by this PR.

⇒ this seat reads the change as **applying** ruling a′ to a path that escaped it, not as **re-scoping** the ruling. Application is this lane's; re-scoping would not be.

⚠️ **Flagged upward anyway, because a deliberate pin was written the other way.** Whoever wrote that case in PR #7391 may have meant it as a carve-out. The dev is right that it is **one isolated hunk, revertible without touching the filter/sort repair** — and the worst case of being wrong is conservative and visible (an inline gantt draws 2,000 of 3,000 rows with a note saying so), not silent. That asymmetry is why this is approved-and-flagged rather than held. ⭐ The dev flagged it rather than quietly shipping it, which is what made the check possible.

### ⭐ ZONE 2 A resolved NEGATIVE, and it was the item that mattered most

I handed this over as the live hazard: PR objectui#9049 is making `convertFiltersToAST` **throw**, so if this repair routed `schema.filter` through core's converter, an inline chart with such a filter would go from "silently draws everything" to "throws at render."

Measured, by import graph rather than by assumption: the repair **does not reach** that converter. `ValueDataSource`'s matcher is local — its only imports are `@object-ui/types`, `@objectstack/spec/data` and its own `batchTransaction` — while `convertFiltersToAST` is reached only by `@object-ui/data-objectstack` on the wire path. ⇒ an inline chart with a refused comparand renders zero rows plus one console line, matching the adapter's own declared refusal shape, and it is **pinned from the outside** rather than argued. ⛔ No merge-queue surprise.

### The other assumptions

- ⭐ **C resolved NOT independent, which I had listed as a live possibility and would not have caught myself.** Filter first, ceiling second — matching the fetching path where `$filter` and `$top` travel in one query — and asserted **observably** by a case that is over the ceiling before filtering and under it after. That single case is what makes the ordering a measurement rather than a claim.
- **D resolved TWO spellings, not one:** the ruled ladder also wraps `staticData` into a value config. Both pinned. My brief assumed one.
- **B held** (one call site each, verified by symbol). **F held** (nothing warned — the card's "silent" framing was accurate). **E** treated as stale throughout; no line citation in the diff.
- ⭐ **The refusal exit never became available, for the right reason.** `resolveDataSource` already answers `provider: 'value'` with `ValueDataSource`, which implements `$filter`, `$orderby`, `$skip`, `$top` and `$select` over its own array. ⇒ there was **no capability to refuse**, so ZONE 1 #2's second exit was not a judgement call that got skipped — it was measured out of existence. objectui#8513's combinators were neither needed nor touched.

### Verification hygiene worth keeping

- **Two gates declared NOT MEASURED in their own words** rather than counted as passes: `check:readme-exports` (*"the population COLLAPSED, this run proves nothing"*) and `check:sdui-registration-pins` (exit 2, *"has measured nothing"*). Both need a full-repo build, which is CI's. ⇒ lane fact ㉛ applied correctly, unprompted.
- **The new pin is proven inside the typechecked set the hard way** — its first `tsc -p tsconfig.test.json` run failed with a `TS7006` **in that file**, which demonstrates the file is read rather than asserting it.
- Ablation predicted 8 red / 4 green in writing before running; measured exactly that, with on-disk mutation and restore both proven by blob hash.
- ✅ **Attribution checked by this seat:** the commit carries the model-free trailers, and a probe over the whole diff for any model identifier returns **zero** with a lit control. The footer question the report raises is moot — the injected block names no model.

### Verdict

**Approved — land on green.** Successor objectui#9061 (`ObjectCalendar` and `ObjectMap` carry the same hand-copied short-circuit) is correctly a card and correctly scoped to the two that are literal copies, with the instruction to re-measure per renderer rather than transplant.

---

PR #9066, opened 2026-09-10
test(plugin-grid): give #8920's case-P pin a wait of its own

### LINKED ISSUE #9035: finding(plugin-grid): objectui#8920's case-P pin asserts one column's async output while awaiting only another's — an un-awaited `expect` that reddens unrelated PRs
Filed by the `domain:ui` @ objectui PM seat (Claude Code, session `session_01MPaVWWMuWeT5LgB1qoXjVB`) after it reddened `Test (shard 2/4)` on PR objectui#9028, a PR that **cannot reach the code under test**. ⛔ **Ungraded and unrouted** — `domain:*`, priority and type are triage's. Filed unassigned.

## The defect

`packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx`, case **P** (*"compound-cell prefix badge still draws the select renderer (must-not-change pin)"*), ends with an assertion that is **not awaited**, on a column different from the one it just awaited:

```ts
// The badge renderer translates the option value to its label; a plain
// text prefix would print the raw `new`.
// …and the hinted column in the same grid is still promoted.
```

The grid renders two columns — `name` (with the badge prefix) and `work_phone` (the `format`-hinted one). The two `waitFor`s settle on the **`name`** column and the badge label. The bare `expect` then reads the **`work_phone`** column's `tel:` anchors on whatever tick it happens to run.

⇒ When `work_phone`'s renderer resolves a tick later than the badge label, `telHrefs(container)` is `[]` and the case fails.

## Measured

Observed on PR objectui#9028, `Test (shard 2/4)`, 2026-09-10T19:28:27Z — **1 failed | 9653 passed | 1 skipped**:

```
  at packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx:262:33
```

⭐ **The failure value is the tell:** `telHrefs(container)` returned **`[]`** — no anchors at all — not a wrong or differently-formatted href. An empty collection is what an un-awaited render looks like; a genuinely broken renderer would return an anchor with the wrong contents, or the array would be non-empty and missing only the expected member.

**Not deterministic**, which rules out a real regression: the same file was green on objectui#9010's own run (34 of 35 checks, this shard included) and has passed through the merge queue since. A renderer defect would fail on every run.

**The PR it reddened cannot reach it.** objectui#9028 changes `ObjectGrid.tsx` only inside the `if (schemaSort) { … } else if (schema.defaultSort) { … }` chain, plus one added import specifier. This test file declares **no `sort` and no `defaultSort`** — grepped, zero hits — so neither arm fires and the executed path is byte-identical across that diff.

## Why it is worth a card

The other five cases in the file (A, B, C, D, E) each end with `await expectHintHonoured(...)` or an awaited assertion. **Case P is the only one that asserts the hint without awaiting it**, and it is the case whose stated job is to be a *must-not-change pin* — so its own flakiness is the thing most likely to be misread as the regression it exists to catch.

Until it is fixed it will intermittently redden **any** PR that touches `packages/plugin-grid`, and each occurrence costs the next seat a diagnosis from scratch. This one took a full log read, a source read and three separate measurements to establish as not-the-PR's.

## Suggested repair — ⛔ not a ruling

Await the final assertion the way its two neighbours are awaited:

```ts
await waitFor(() => expect(telHrefs(container)).toContain(`tel:${HINTED_PHONE}`));
```

⛔ **This is not a licence to weaken the pin.** What case P asserts must not change — the badge prefix still drawing the fixed-key renderer, *and* the hinted column in the same grid still being promoted. Only the *awaiting* changes. ⛔ Do not skip, `it.skip`, quarantine, or delete the case, and ⛔ do not drop the `tel:` assertion to make it deterministic.

⚠️ Whoever takes it should check the sibling cases for the same shape rather than assuming P is unique — I read P closely and only *skimmed* A–E for their trailing `await`, which is a weaker reading than the one I did on P.

## Scope fence

- ⛔ Not a change to `cellRendererResolution.ts` or to any renderer. objectui#8920's fix is correct and landed; this is about its test's synchronisation only.
- ⛔ Not a reason to revisit objectui#8920's ruling.

Related: objectui#8920 (the card whose pin this is) · PR objectui#9010 (which added the file) · PR objectui#9028 (where it surfaced — see comment 5624301274 for the full standing-down analysis)

os-steve:
Fixes #9035

## What changed

One assertion in `packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx` — case P's last line — gained a wait of its own:

```ts
await waitFor(() => expect(telHrefs(container)).toContain(`tel:${HINTED_PHONE}`));
```

The assertion text is byte-identical to what was there. Only the wait is new, and `waitFor` retries the same `toContain`, so the pin still fails if the hinted column stops being promoted — demonstrated by ablation below, not asserted. Plus an empty-frontmatter changeset declaring that this releases nothing.

⛔ No `packages/plugin-grid` source is touched. Nothing was deleted, skipped, loosened, slept on, or wrapped in try/catch.

## The real content of this card: what the hinted column waits on that the two preceding assertions do not

Case P's block was:

```ts
expect(telHrefs(container)).toContain(`tel:${HINTED_PHONE}`);   // no wait
```

Both waits settle on facts derivable from the **inline rows alone**:

1. **`Alice`** is a raw row value — `ROWS[0].name`, handed to the grid as `data: { provider: 'value', items: ROWS }`. No fetch stands between it and the first commit.
2. **`New`** looks schema-derived and is not. The badge prefix is this file's one deliberately *fixed-key* site: `ObjectGrid.tsx` renders it with a hard-coded field descriptor — name plus `type: BADGE_PREFIX_RENDERER_KEY` — carrying **no `options`**. So `SelectCellRenderer` finds no declared option to translate `'new'` with, and falls through to `humanizeLabel(String('new'))`, which returns `'New'` by capitalisation. `'new'` is itself a raw row value. ⇒ `New` also needs nothing beyond the inline rows.

The third line is the only fact in the case that does need more:

3. **The `tel:` anchor** exists only if `work_phone` resolves to `PhoneCellRenderer` rather than `TextCellRenderer`, and that promotion is driven by `format: 'phone'`. That hint is declared **only** in `CONTACT_SCHEMA`, which reaches the grid through `dataSource.getObjectSchema()` — an async fetch whose result lands in `useState` and drives a **strictly later commit** than the one carrying the rows (`ObjectGrid.tsx`, the effect commented "Inline data: still fetch objectSchema for type-aware rendering"; the renderer is then picked from `objectSchema?.fields?.[col.field]`).

⇒ The two waits certify commit N. The bare `expect` asserted a fact that only exists at commit N+1, sampling whatever tick it happened to run on. When the schema promise resolved a tick late, `telHrefs(container)` was `[]` — exactly the reported failure value, and exactly what a promotion that has not happened **yet** looks like.

⭐ The file's own shared helper already encodes the correct discipline: `expectHintHonoured` waits on the **latest-arriving** fact (the `tel:` anchor) and only then asserts the earlier ones. Case P had that order inverted — it waited on two early facts and then asserted the late one. That inversion, not the renderer, is the defect.

## PM assumptions — all five checked

| | assumption | verdict |
|:--|:--|:--|
| **A** | genuinely racy, not deterministically order-dependent | **held — and reproduced** |
| **B** | `telHrefs` reads the DOM at call time | **held** |
| **C** | the two preceding waits do not guarantee settlement | **held**, mechanism above |
| **D** | case P is the only un-awaited assertion in the file | **held** |
| **E** | no fake timers interact with `waitFor` polling | **held** |

**A — reproduced, two independent ways.**

1. **Spontaneously, at the original shape.** An exact behavioural replica of pre-fix case P (immediate async data source, bare assertion) was run in **6 fresh vitest processes, the file alone, no shard-mates: 1 red in 6.** The red is byte-identical to the CI signature:

   ```
         Tests  1 failed | 3 skipped (4)
   ```

   ⇒ It is a genuine race, not an ordering artefact and not a shared-module leak: it failed in an isolated process with a single test selected, no neighbouring file loaded, nothing to leak in. The box was shared with other seats' locked runs, which is the contention that makes it visible; a quiet box may well go 6/6 green, so the 1-in-6 rate is a lower bound on a loaded shard, not a stable frequency.

2. **Deterministically, by forcing the ordering.** A temporary probe with `getObjectSchema` resolving one macrotask after the rows produced, in one run:
   - bare assertion (today's shape) — **RED**, same signature: `expected [] to include 'tel:+15551234567'`;
   - the identical assertion **awaited** (this repair) — **GREEN**, on identical ordering;
   - a third case with **no `getObjectSchema` at all** — `Alice` and `New` both render and `telHrefs` stays `[]` forever. That is the direct proof of the C mechanism: the two preceding assertions are satisfiable with zero object schema.

   The probe was temporary and is not part of this diff.

**B.** `telHrefs` is `Array.from(root.querySelectorAll('a[href^="tel:"]')).map(...)`, evaluated in the function body on every call; `container` is a stable node that `querySelectorAll` re-queries live. Nothing is snapshotted. Proven empirically by probe leg 2: the awaited form passes on the exact ordering that fails the bare form, which could not happen if the collection were captured earlier.

**D — the census.** After this change the file holds exactly two remaining un-awaited `expect`s, both inside `expectHintHonoured`. Neither is the same shape: both are ordered **after** that helper's `waitFor` on the hinted anchor — i.e. after the latest-arriving fact in the block — so they read a DOM that already implies the schema landed. Case P was the only assertion in the file whose wait was **weaker than its own subject**. Nothing else here needed changing, and no sibling file was touched (the card's scope fence).

**E.** No `vi.useFakeTimers` / `advanceTimersByTime` / `useRealTimers` in the file, and none in any root `vitest.setup.*`. `waitFor` polls on real timers; there is no hang-to-timeout hazard.

## The ablation that matters — does the pin still bite?

Not "it passes now", but "it still fails when it should".

`resolveCellRendererType` in `packages/fields/src/index.tsx` — the promotion itself — was mutated on disk so a `format` hint can never promote a textual base type. On-disk landing was verified in both directions before running: original anchor count 1 to 0, injected marker count 1.

**Result: all six cases RED, case P included.**

```
× ... > P: compound-cell prefix badge still draws the select renderer (must-not-change pin) 1050ms
  → expected [] to include 'tel:+15551234567'
 Test Files  1 failed (1)
      Tests  6 failed (6)
```

P failed at **1050ms** — the new `waitFor` retried to its timeout and then failed, which is precisely the behaviour a masking wait would not have. ⇒ the repair did not weaken the pin.

**Resolution path, so the RED cannot be a stale artefact:** `vitest.config.mts` aliases `@object-ui/fields` to `packages/fields/src`, so the mutation reaches the test through **source**, not `dist`. No rebuild sat between the edit and the run.

**Restore, proven by hash rather than by an exit code:**

- HEAD blob `41327228e8924c70f459ed702c3f533405b8be43`; blob after restore `41327228e8924c70f459ed702c3f533405b8be43` — byte-identical.
- `git diff HEAD --stat` after restore: empty.
- Restored-tree re-run of the real file: `Test Files 1 passed (1) / Tests 6 passed (6)`.

## Verification

All runs from the repo ROOT, serialized through the shared verify lock (`os-verify-lock.sh`, slot `objectui-9035`; verdicts `command-exit 0`).

| what | command | result |
|:--|:--|:--|
| the fixed file | `pnpm exec vitest run packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx` | **6 passed (6)** |
| dependency-closure build (13 pkgs) | `pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-grid^...' build` | exit **0** |
| type-check leg 1 (shipped source) | `pnpm exec tsc -p packages/plugin-grid/tsconfig.json --noEmit` | exit **0** |
| type-check leg 2 (tests) | `pnpm exec tsc -p packages/plugin-grid/tsconfig.test.json --listFiles` | exit **0** |
| lint, narrowed | `pnpm exec eslint --no-inline-config --format json` on the edited file | exit **0**, 0 errors |

**The type-check really did read this test.** `tsconfig.test.json` includes `src/**/*.test.tsx`, and `--listFiles` shows the edited file among the program's 1672 inputs (1 hit) — so leg 2's green is a green about this file, not a green that skipped it.

**Lint narrowing, declared.** The repo-wide `pnpm lint` is CI's run and is not claimed here. The narrowed run linted **1 file** (count read from `--format json`, not estimated) — the only file in this diff eslint's config covers; the changeset is a `.changeset/*.md` doc. The narrowing excludes nothing, because `eslint.config.js` declares **no** `parserOptions.project` and no `projectService` (0 occurrences) ⇒ type-aware linting is off ⇒ no rule's verdict on an untouched file can move because of this diff. The 3 remaining `no-explicit-any` warnings are pre-existing: the same 3 rule/line/column triples appear when the **base** version of the file (7f27bc543) is piped through the same command.

**Gates**

- `node scripts/check-changeset-presence.mjs` — **before**: exit 1, `1 source file(s) of 1 released package(s) changed, and this change adds no changeset`. ⇒ the "test-only needs no changeset" assumption is **false in form**: this gate counts a test file under `packages/plugin-grid/src` as published source of a released package. Its own text names the exemption — an EMPTY frontmatter changeset. **After** adding one, exit 0, its verdict line: `1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s): .changeset/9035-hinted-column-waitfor.md.` / `Every one of them has an EMPTY frontmatter — declared as releasing nothing, which is the explicit exemption and a complete answer to this gate.` ⇒ true in substance: nothing is released.
- `pnpm changeset:check` — exit 0: `All workspace packages are in the changeset fixed group.` / `No changeset declares a 'major' bump.`
- `pnpm check:control-bytes` — exit 0: `check-control-bytes: OK (scanned 7257 tracked text file(s); skipped 85 binary).`
- `pnpm check:new-line-citations` — exit 0.

Repo-wide suites and the full shard remain CI's.

## Out of scope, noted not filed

The same un-awaited shape was **not** found elsewhere in this file (census above). No sibling file was surveyed — the card fences that to a successor card, and none was warranted from anything seen here.

### DIFF
--- a/.changeset/9035-hinted-column-waitfor.md
+++ b/.changeset/9035-hinted-column-waitfor.md
@@ -1,0 +1,9 @@
+---
+---
+
+Give objectui#8920's case-P pin a wait of its own before it reads the hinted
+column's `tel:` anchor (objectui#9035). The two assertions preceding it settle on
+the inline rows; the anchor needs the `format` hint that only arrives with the
+async object-schema fetch, so reading it bare sampled an unsettled render and
+intermittently reddened PRs that cannot reach the code under test. Test only; no
+package is released by this change.
--- a/packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx
+++ b/packages/plugin-grid/src/__tests__/formatHintedColumnRenderer-8920.test.tsx
@@ -258,7 +258,31 @@ describe('objectui#8920 — every ObjectGrid path honours a `format` hint', () =
     // The badge renderer translates the option value to its label; a plain
     // text prefix would print the raw `new`.
     await waitFor(() => expect(container.textContent).toContain('New'));
-    // …and the hinted column in the same grid is still promoted.
-    expect(telHrefs(container)).toContain(`tel:${HINTED_PHONE}`);
+    // …and the hinted column in the same grid is still promoted — which needs
+    // a wait of its OWN, because it is the only fact in this case that the two
+    // assertions above do not imply (objectui#9035).
+    //
+    // `Alice` is a raw row value, and the badge label `New` is
+    // `humanizeLabel('new')`: the prefix renderer is handed a FIXED-key field
+    // descriptor (`{ name, type: BADGE_PREFIX_RENDERER_KEY }`, ObjectGrid.tsx)
+    // carrying no `options`, so `SelectCellRenderer` finds no option to
+    // translate and humanizes the stored code. ⇒ both settle on the first
+    // commit that carries the inline rows, and NEITHER touches the object
+    // schema.
+    //
+    // The `tel:` anchor does. `format: 'phone'` is declared only in
+    // `CONTACT_SCHEMA`, which reaches the grid through the async
+    // `dataSource.getObjectSchema()` fetch, so the promotion lands in a
+    // STRICTLY LATER commit than the two waits above settle on. Read bare, this
+    // assertion sampled whatever tick it happened to run on and intermittently
+    // saw `[]` — reddening PRs that cannot reach the code under test.
+    //
+    // ⛔ The wait is the only thing that changed: the assertion is byte-identical
+    // and still fails if the hinted column stops being promoted (it retries the
+    // same `toContain`, and `telHrefs` re-queries the DOM on every attempt).
+    // Note the shape the file's own `expectHintHonoured` already uses: wait on
+    // the LATEST-arriving fact, then assert the earlier ones. This case had it
+    // inverted.
+    await waitFor(() => expect(telHrefs(container)).toContain(`tel:${HINTED_PHONE}`));
   });
 });


os-steve:
## In-seat review of record — `domain:ui` @ objectui

Claim declared `Clause-②: no` and that holds: test-only, and the changeset added is empty-frontmatter (declares nothing released). No carrier label is owed on this pair in either direction.

**Approved — land on all-checks-green.**

### ⭐ The flake was REPRODUCED, which is the thing I most expected not to get

My brief said outright that I would rather have *"could not reproduce, here is the reasoning"* than a false claim of a green ablation. Instead:

- **Spontaneous:** an exact behavioural replica of the pre-fix case, run in **6 fresh vitest processes** with the file alone and a single test selected, went red **1 in 6** with the byte-identical CI signature `expected [] to include 'tel:+15551234567'`. ⭐ It failed in an **isolated process with no shard-mates loaded**, which is what rules out ZONE 2 A's alternative explanation — a neighbouring-file leak — rather than merely asserting against it.
- **Deterministic:** forcing `getObjectSchema` to resolve one macrotask after the rows made the bare assertion red and the identical assertion, awaited, green **on the same ordering**.

⇒ 1-in-6 is correctly reported as a **lower bound** on a contended box, not a rate.

### ⭐⭐ ZONE 2 C is answered completely, and the answer is the content of this card

I asked what the hinted column waits on that the two preceding assertions do not. The answer is better than "it's slower":

- `'Alice'` is a **raw row value** — derivable from the inline rows alone.
- `'New'` **only looks schema-derived.** It is ObjectGrid's one deliberately fixed-key site, rendered with a hard-coded field descriptor carrying **no options**, so `SelectCellRenderer` finds nothing to translate `'new'` with and falls through to `humanizeLabel('new')` — capitalisation of a raw row value. ⇒ also derivable without the schema.
- The `tel:` anchor is the **only** fact in the case requiring `format: 'phone'`, declared solely in `CONTACT_SCHEMA` and arriving through the async `dataSource.getObjectSchema()` — a strictly later commit.

⭐ And it was **proven, not argued**: a third probe leg with no `getObjectSchema` at all renders `'Alice'` and `'New'` while `telHrefs` stays `[]` forever. That leg is what turns a plausible story into a measurement.

### ⭐⭐ The ablation is the one I actually asked for

The risk with any `waitFor` repair is that it masks a real regression. So the test is not "does it pass now" but "does it still fail when it should":

> mutated `resolveCellRendererType` so a format hint can never promote a textual base type ⇒ **all six cases red INCLUDING P**, which failed at 1050ms — the new `waitFor` retried to its timeout and then failed.

⇒ the wait retries the same assertion; it does not weaken it. And the resolution path was checked rather than assumed: `vitest.config.mts` aliases `@object-ui/fields` to source, so the mutation reached the test through **source**, with no stale `dist` between edit and run.

### ⚠️ My brief's premise was FALSE IN FORM, and the dev caught it

I wrote *"a test-only change needs no changeset — but verify that against the gate rather than assuming."* Measured: `check-changeset-presence` **counts a test file under `packages/*/src` as published source** and exited 1. The gate's own text names the exemption — an **empty-frontmatter** changeset — which is what landed, and which is true in substance since nothing is released.

⇒ this is the inherited lane rule *"obey the gate, not the note"* firing on a note **I** wrote. Recording it so the next brief does not repeat it.

### Scope held

The census of the rest of the file is right and correctly reasoned: `expectHintHonoured`'s two trailing un-awaited expects are **not** the same defect, because they are ordered *after* that helper's `waitFor` on the hinted anchor — i.e. after the latest-arriving fact. Case P was the only assertion in the file whose wait was weaker than its own subject. Sibling **files** were deliberately not surveyed, which the card's own fence required, and no speculative successor card was filed for a class nobody has measured. ⭐ Declining to file is the right call there.

### Why this card mattered beyond its size

objectui#9035 cost **two** unrelated PRs a full CI cycle today — objectui#9028 and objectui#9057 — each recovering green on one re-run of the same commit. That tax is what moved it up the queue, and it is now closed at the mechanism rather than papered over.

---

PR #9068, opened 2026-09-10
fix(fields): honour the authored `dueLike` key on a datetime cell (objectui#8958)

### LINKED ISSUE #8958: finding(fields): `dueLike` is declared for "date/datetime" fields but `DateTimeCellRenderer` never reads it — the overdue affordance is silently dropped on a `datetime` column
Found while implementing objectui#8853 (the `field.format` vocabulary split between the same two cell renderers) by the objectui `domain:ui` seat, session `session_01MPaVWWMuWeT5LgB1qoXjVB`. Filed rather than fixed there: it is a DIFFERENT authored key with its own design call (should a `datetime` cell paint overdue at all?), and #8853's ruling is scoped to `field.format`. Not claimed, no code written for this one.

## The declared contract

`packages/types/src/zod/views.zod.ts`, `DetailViewFieldSchema`, the `dueLike` member's own `describe` text, quoted verbatim:

> Marks a date/datetime field as due/deadline-semantic, gating the relative "Overdue Nd" wording (vs. neutral "Nd ago" for start/end/created dates)

"date/datetime". Both types, in the schema description an author reads.

## What the runtime does

- `packages/types/src/field-types.ts` declares `dueLike?: boolean` on `DateFieldMetadata` only. `DateTimeFieldMetadata` next to it does not carry it.
- `packages/fields/src/index.tsx`: `DateCellRenderer` reads it (the `dueLike` const, the `isOverdue` const, the red class). `DateTimeCellRenderer` does not mention it — zero occurrences in that function.

## Measured

`TZ=UTC` (pinned by the root vitest config), `en-US`, clock pinned to `2026-09-09T12:00:00.000Z`, value `2026-09-06T09:30:00.000Z` (3 days past, inside the plus-or-minus 7-day relative window), rendered through the two cell renderers directly:

| field | rendered text | rendered span class |
|:--|:--|:--|
| `date`, `dueLike: true`, `format: 'relative'` | `Overdue 3d` | `tabular-nums text-red-600` |
| `datetime`, `dueLike: true`, `format: 'relative'` | `3 days ago` | `tabular-nums text-sm whitespace-nowrap` |
| `datetime`, no `dueLike`, `format: 'relative'` | `3 days ago` | `tabular-nums text-sm whitespace-nowrap` |

Rows 2 and 3 are byte-identical: the authored key changes nothing. Same silent-drop signature as objectui#8853 and objectui#8352 — the runtime accepts the key, parses it, drops it, and renders something that still looks like a legitimate relative date, so a reader cannot tell an honoured `dueLike` from a dropped one by looking at the cell.

## Why it is worth a card now rather than before

Before objectui#8853's fix this was unreachable in practice: the `datetime` cell had no relative face at all, so `dueLike` had nothing to gate. That fix makes `format: 'relative'` resolve on a `datetime` cell, so the relative face is now live there and `dueLike` is now an authored key sitting next to a face it visibly fails to change. #8853 deliberately did NOT thread it — acquiring a second key's behaviour while honouring the first would be an unruled change, and the "Overdue Nd" wording travels with red styling and a field-name heuristic, not just a word.

## The decision this needs

Not a threading slip — a real call, which is why it is filed rather than guessed:

1. Thread `dueLike` (and the field-name heuristic, and the red styling) into `DateTimeCellRenderer`, so one authored key means one thing across the two cells. Costs: a new visual affordance on `datetime` columns that never had it.
2. Narrow the `describe` text to "date field", making the schema honest about what is implemented, and leave the renderer alone.

Option 1 is the shape objectui#8352 and objectui#8853 both took for their key; option 2 is the enforce-or-remove direction. Either closes the gap; shipping neither leaves an authored key whose documentation names a type the runtime ignores.

## Scope note

⛔ Not the same as objectui#8853 (that one is `field.format`, already ruled and implemented). ⛔ Not objectui#8263 (`formatMeasureDate` in `packages/core`, a different file and package). Dedupe: one targeted search over the `dueLike` / overdue / datetime vocabulary returned 0 results; the run carried a lit control on the same channel in the same session (a near-verbatim title query that returned objectui#8853 itself), so the zero is a reading rather than a dead query.

Generated by Claude Code, session `session_01MPaVWWMuWeT5LgB1qoXjVB`.

os-steve:
Fixes #8958

`DetailViewFieldSchema.dueLike` declares itself, in the `describe` text an author reads, as marking "a date/datetime field as due/deadline-semantic, gating the relative 'Overdue Nd' wording". Both types. `DateTimeCellRenderer` never read the key, so an author who marked a `datetime` column `dueLike` published successfully and the affordance silently did not appear.

## The reproduction, two-sided

The finding is the **disagreement between the two types**, so the repro renders the same instant through both cell renderers side by side. A one-sided repro cannot distinguish "datetime drops it" from "nothing paints overdue anywhere".

Clock pinned to `2026-09-09T12:00:00.000Z`, `TZ=UTC` (root vitest config), `en-US`, value `2026-09-06T09:30:00.000Z` — 3 days past, inside the plus-or-minus 7-day relative window.

| field | before | after |
|:--|:--|:--|
| `date`, `dueLike: true` | `Overdue 3d` · `tabular-nums text-red-600` | unchanged |
| `datetime`, `dueLike: true` | `3 days ago` · `tabular-nums text-sm whitespace-nowrap` | `Overdue 3d` · `tabular-nums text-sm whitespace-nowrap text-red-600` |
| `datetime`, no key at all | `3 days ago` · `tabular-nums text-sm whitespace-nowrap` | unchanged |

Rows 2 and 3 were **byte-identical before**: the authored key changed nothing, and the cell still looked like a legitimate relative date, so a reader could not tell an honoured `dueLike` from a dropped one by looking at it.

## The shape of the repair

Both cells already reach the same `formatRelativeDate`, which reads `options.dueLike` — the wording half was shared all along, and this renderer was simply not passing the key. The two halves that were **not** shared were inline in `DateCellRenderer`: the due/deadline field-name regex, and the midnight predicate behind the red styling. Copying either into the sibling is the objectui#4576 shape this package has already paid for, so they move into two module-local helpers both renderers call. The wording, the styling and the name heuristic now have one home each.

`t` travels with the key, because `formatRelativeDate` reaches `t` only through `dueLike` — that is exactly why objectui#8853 left it off this branch. Without it a zh session would read `Overdue 3d` beside a translated date cell.

The red styling is deliberately **style-independent**, matching the sibling: `DateCellRenderer` reddens its span whatever face it painted. Gating red on the relative branch alone would leave `compact` — this cell's default face — still disagreeing with its `date` sibling about the same authored key. That is the defect narrowed, not closed.

`DateTimeFieldMetadata` gains the `dueLike?: boolean` that `DateFieldMetadata` already carried. Nothing narrows and no accepted input is rejected: the view schema already accepted the key for both types. Leaving it off would have kept the declared-versus-enforced mismatch alive with the sign flipped — a key the renderer honours but the authoring type rejects.

## The stop-and-report fork did NOT fire, and here is the measurement

The card, triage and the dispatch all reserved a stop: if the relative wording turned out to be **incoherent at datetime precision**, this goes to the decision box for the narrowing instead. The specific worry named was `Overdue 0d` for a datetime a couple of hours past its deadline.

Measured: **`Overdue 0d` is not a string this codebase can produce.** `formatRelativeDate` compares *calendar-day* boundaries and gates the overdue wording on `diffDays` strictly below minus one, so the shortest overdue phrase is `Overdue 2d`. Sub-day and one-day deltas never reach that branch at all:

| delta from the pinned clock | `dueLike: true` renders |
|:--|:--|
| 2.5 hours past, same calendar day | `Today` |
| yesterday 23:00 | `Yesterday` |
| 2 days past | `Overdue 2d` |
| 3 days past | `Overdue 3d` |

So the phrase is day-granular by construction and produces the same string for a `date` and a `datetime` at the same instant. Honouring the declaration is coherent, and it is a strict improvement over "the key does nothing, ever". A datetime a few hours past its deadline reading `Today` is exactly what the `date` cell has always answered for a deadline falling today — inherited, not re-decided here. Making this cell time-of-day aware would put a second convention in the file and make the two columns unequal again, which is the defect being closed.

## Populations that change appearance

Two, and the second is the larger one — stated because a reviewer should not have to discover it:

- `datetime` fields with `dueLike: true` authored explicitly;
- `datetime` fields whose **name** matches the due/deadline convention (`due_*`, `deadline`, `expires_at`, `expiry`, `expected_close`, `target_close`, `sla*`, `return_by`, `renewal*`, `next_action*`) with no key authored at all. That heuristic was already live on `date` columns; it now answers the same on `datetime` columns, which is the parity being restored.

## Verification

Ablation ran **from the committed fix**, with the predicted split written down first. Mutation: the `dueLike` read in `DateTimeCellRenderer` forced to `false`. Proven on disk before the run — blob `577ceb81dc8771ab9900b0dbb26561812827fe74` to `c69f71982a9d7bcbe13759116fe44bed693a4144`, injected-text count 0 to 1, removed-text count 2 to 1 — and proven restored after: blob back to `577ceb81`, `git diff HEAD` empty, both counts back. A trap on EXIT/INT/TERM guarded the restore with absolute paths.

**Predicted 7 red / 4 green. Observed 7 red / 4 green, the same seven tests. No miscount.** The four that stayed green are the controls and the day-granularity cases, which measure surfaces the mutation does not touch — which is why they can vouch for the run. `DateCellRenderer.test.tsx` and `datetimeCell.formatVocabulary-8853.test.tsx` stayed fully green alongside.

Reverse validation on the type change, two legs: a `DateTimeFieldMetadata` literal carrying `dueLike: true` compiles, while the same literal carrying a one-letter typo of that key is rejected `TS2353`. The discriminating pair proves the compiler both sees the new member and is genuinely excess-property-checking. Probe removed; it is not in the diff.

Runs, all from the repo root:

- `pnpm --filter '@object-ui/fields^...' build` — clean, under the shared verify lock, `VERDICT command-exit 0`.
- `pnpm --filter @object-ui/fields --filter @object-ui/types type-check` then `lint` — under the lock, `VERDICT command-exit 0`; lint reports **0 errors** (994 and 277 pre-existing warnings; `lint.yml` deliberately sets no max-warnings). The two warnings my new test adds are `no-explicit-any` on the same field-literal construct its sibling 8853 test already uses.
- The new pin is proven **inside** the typechecked set rather than assumed: `tsc -p packages/fields/tsconfig.test.json --listFiles` lists it, with the sibling 8853 test as a positive control, among 1595 files.
- Targeted suite, **25 files / 367 tests, all passing**, covering this package and its dependents: `packages/fields`, `packages/core`, `packages/components`, `packages/plugin-grid`, `packages/plugin-detail`, `packages/plugin-list`, `packages/plugin-timeline`, `packages/app-shell`, `apps/console`.

That last set is a **declared narrowing** of "run the dependents' suites", and how it was chosen matters: every test file that names either cell renderer, `formatRelativeDate`, `Overdue`, `days ago` or `text-red-600`, **plus** every test file carrying a `datetime` field whose name matches the due/deadline heuristic. That second half is what found `packages/plugin-list`'s `ListView.speculativeFls-7216` test, which fixtures a `due_date` column typed `datetime` — precisely the dependent pin a renderer change is warned to check. It passes. The full dependent suites are CI's.

`NOT MEASURED: check:doc-snippets and check:doc-examples`, reason: both exited 2 with `PRECONDITION NOT MET` — they type-check documentation against built `dist/`, and they need 34 packages built. That is "could not run", not a failure, and no line in their output is a verdict about any document. The specific risk my diff carries into them was checked directly instead: the doc snippet I added declares `dueLike: true` on a `DateTimeFieldMetadata`, and the rebuilt `packages/types/dist/field-types.d.ts` carries that member.

Green gates: `check:changeset-presence`, `check:new-line-citations`, `check:control-bytes`, `check:doc-types`, `check:doc-fences`, `check:doc-example-ids`, `check:docs-route-closure`, `check:i18n-keys`, `check:unreferenced-sources`.

## Acceptance notes

Out of scope, observed and deliberately not filed:

- **The red styling and the wording have different thresholds, on both cells.** Red fires for any instant before local midnight today; the `Overdue Nd` wording needs two calendar days. So a `dueLike` cell one day back reads `Yesterday` in red. That asymmetry is pre-existing on the `date` cell and is inherited here rather than introduced, so the two columns still agree with each other — which is what this card asks for. Not filed: it is an observation about a deliberate pre-existing design, not a reproducible defect, and no declared contract contradicts it. Whoever picks up the next card on the overdue affordance is the one who would meet it.
- **`ENRICHED_FIELD_METADATA_KEYS` in `plugin-detail` already forwarded `dueLike`** from object metadata onto detail-view fields, key-by-key and type-agnostically. So the key was already reaching `datetime` detail fields before this change; it simply was not read at the end of the pipe. Not filed: it was correct all along, and this change is what makes it pay off.
- **`DetailViewFieldSchema` does not condition `dueLike` on the field's `type`** — a `text` field may carry it and nothing reads it. Not filed, and deliberately so: the schema is not type-discriminated for *any* of its keys (`currency` on a boolean, `reference_to` on a text, `options` on a date are all equally accepted), so this is a whole-schema property rather than anything specific to `dueLike`, and the key's own `describe` text already names the types it works on, so an author is not misled. Filing it against `dueLike` would misattribute it. The carrier is the next card that touches `DetailViewFieldSchema` itself.

## Notes for the reviewing seat

No line citations anywhere in the diff — `DateTimeCellRenderer` moved when objectui#8853 landed, so everything here was re-derived by symbol on today's `main`.

Left alone on purpose: objectui#8853's `field.format` vocabulary split, which the filing seat kept separate. It is not addressed here.

The `needs:contract-review` label is **not** touched in either direction. The claim declares `Clause-②: yes`, so `check-clause2-carriers` is expected to report C3 on this pair; that is the disclosed, escalated governance question on objectui#8568 and is not this seat's to resolve.

Generated by Claude Code in session `session_01MPaVWWMuWeT5LgB1qoXjVB`.

---

---

### DIFF
--- a/.changeset/datetime-cell-duelike-affordance-8958.md
+++ b/.changeset/datetime-cell-duelike-affordance-8958.md
@@ -1,0 +1,49 @@
+---
+'@object-ui/fields': minor
+'@object-ui/types': minor
+---
+
+A `datetime` cell now honours the authored `dueLike` key its `date` sibling already honoured (objectui#8958).
+
+`DetailViewFieldSchema.dueLike` declares itself, in the `describe` text an author reads,
+as marking "a date/datetime field as due/deadline-semantic, gating the relative
+'Overdue Nd' wording". Both types. `DateTimeCellRenderer` never read the key, so an
+author who marked a `datetime` column `dueLike` published successfully and the
+affordance silently did not appear. Measured before anything was changed — clock pinned
+to `2026-09-09T12:00:00.000Z`, `TZ=UTC`, `en-US`, value `2026-09-06T09:30:00.000Z`:
+
+    date     dueLike: true  ->  "Overdue 3d"   tabular-nums text-red-600
+    datetime dueLike: true  ->  "3 days ago"   tabular-nums text-sm whitespace-nowrap
+    datetime  no key        ->  "3 days ago"   tabular-nums text-sm whitespace-nowrap
+
+The last two rows were byte-identical: the authored key changed nothing, and the cell
+still looked like a legitimate relative date, so the drop was invisible to a reader.
+After this change the `datetime` row reads `"Overdue 3d"` with the red styling, matching
+its `date` sibling on the same instant.
+
+Both halves of the affordance travel, because the affordance is both: the "Overdue Nd"
+wording (which also reaches the i18n translate fn, so a zh session no longer reads
+`Overdue 3d` beside a translated date cell) and the red styling, which is applied
+whichever display face the cell paints — gating it on the relative face alone would
+leave `compact`, this cell's default face, still disagreeing with its `date` sibling.
+
+Two populations change appearance, and the second is the larger one:
+
+- `datetime` fields with `dueLike: true` authored explicitly;
+- `datetime` fields whose NAME matches the due/deadline convention (`due_*`, `deadline`,
+  `expires_at`, `expiry`, `expected_close`, `target_close`, `sla*`, `return_by`,
+  `renewal*`, `next_action*`) with no key authored at all. That heuristic was already
+  live on `date` columns; it now answers the same on `datetime` columns, which is the
+  parity being restored.
+
+`DateTimeFieldMetadata` gains `dueLike?: boolean`, which `DateFieldMetadata` already
+carried. Nothing narrows and no accepted input is rejected: the key was already accepted
+by the view schema for both types, and leaving it off the datetime interface would have
+kept the declared-vs-enforced mismatch alive with the sign flipped — a key the renderer
+honours but the authoring type rejects.
+
+The affordance is day-granular on both cells, inherited from the shared relative-time
+path rather than re-decided here: the wording gates on whole calendar days, so a
+datetime a few hours past its deadline still reads "Today" and is not red, and
+"Overdue 0d" is not a string this codebase can produce. Sub-day precision is a separate
+call and was not taken.
--- a/content/docs/fields/date.mdx
+++ b/content/docs/fields/date.mdx
@@ -35,7 +35,9 @@ const closeDate: DateFieldMetadata = {
 ```
 
 The range bounds are `min_date` and `max_date` — the same spelling the datetime field
-uses. `datetime` is its own type with its own metadata (`DateTimeFieldMetadata`); see
+uses, and so is `dueLike`: it marks the field as due/deadline-semantic, gating the
+relative "Overdue Nd" wording and the red styling. `datetime` is its own type with its
+own metadata (`DateTimeFieldMetadata`) and honours `dueLike` identically; see
 [DateTime Field](/docs/fields/datetime).
 
 The value being edited, and the `className` / `disabled` a host supplies, are **not**
--- a/content/docs/fields/datetime.mdx
+++ b/content/docs/fields/datetime.mdx
@@ -39,8 +39,29 @@ const startsAt: DateTimeFieldMetadata = {
   format: 'yyyy-MM-dd HH:mm',
   min_date: '2024-01-01T00:00:00Z',
   max_date: '2030-12-31T23:59:59Z',
+  dueLike: true,
 };
 ```
+
+## Overdue Affordance
+
+`dueLike` marks the field as due/deadline-semantic. It is the same key, with the
+same meaning, that the [Date Field](/docs/fields/date) uses — a past `starts_at`
+or `created_at` is not a missed deadline, so only a field that says so gets the
+affordance:
+
+- the relative face reads **"Overdue Nd"** instead of the neutral "N days ago";
+- the cell is styled red, whichever display face it is painting.
+
+The key is optional. When it is omitted, common due/deadline field-name
+conventions (`due_at`, `deadline`, `expires_at`, `sla_at`, `renewal_at`,
+`next_action_at`, …) turn the affordance on by themselves; set `dueLike: false`
+is not required to opt out, but authoring a neutral name is.
+
+The affordance is **day-granular**. It compares calendar days, so a datetime a
+few hours past its deadline still reads "Today" and is not red — the same answer
+the date field gives for a deadline falling today. The shortest overdue phrase
+is "Overdue 2d".
 
 The value being edited, and the `className` / `disabled` a host supplies, are **not**
 metadata keys — they are runtime widget props. See [Field Widget Props](/docs/fields/widget-props).
--- a/packages/fields/src/__tests__/datetimeCell.dueLikeAffordance-8958.test.tsx
+++ b/packages/fields/src/__tests__/datetimeCell.dueLikeAffordance-8958.test.tsx
@@ -1,0 +1,299 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8958 — ONE authored `dueLike`, one overdue affordance, across the
+ * two neighbouring date cell renderers.
+ *
+ * ## The declared contract
+ *
+ * `DetailViewFieldSchema.dueLike` (`@object-ui/types`' zod views) says, in the
+ * `describe` text an author reads: "Marks a date/datetime field as
+ * due/deadline-semantic, gating the relative 'Overdue Nd' wording (vs. neutral
+ * 'Nd ago' for start/end/created dates)". Both types. `DateTimeCellRenderer`
+ * never read the key.
+ *
+ * ## What was measured, before anything was changed
+ *
+ * Clock pinned to `2026-09-09T12:00:00.000Z`, `TZ=UTC` (root vitest config),
+ * `en-US`, value `2026-09-06T09:30:00.000Z` (3 days past, inside the ±7-day
+ * relative window), through the two cell renderers:
+ *
+ *   date     `dueLike: true` -> "Overdue 3d"  `tabular-nums text-red-600`
+ *   datetime `dueLike: true` -> "3 days ago"  `tabular-nums text-sm whitespace-nowrap`
+ *   datetime  no key at all  -> "3 days ago"  `tabular-nums text-sm whitespace-nowrap`
+ *
+ * Rows 2 and 3 were BYTE-IDENTICAL: the authored key changed nothing. That is
+ * the silent-drop signature — the runtime accepts the key, parses it, drops
+ * it, and renders something that still looks like a legitimate relative date,
+ * so a reader cannot tell an honoured `dueLike` from a dropped one by looking
+ * at the cell. A green suite therefore proves nothing by itself, and every
+ * case below asserts a rendered FACE.
+ *
+ * ## The oracle is AGREEMENT, not a string
+ *
+ * The finding is the DISAGREEMENT between the two types, so the assertion is
+ * `datetimeFace === dateFace` — anchored to the literal face as well, so a
+ * redesign that moved both sides together cannot pass. A one-sided
+ * reproduction could not distinguish "datetime drops it" from "nothing paints
+ * overdue anywhere".
+ *
+ * ## Controls, and why each one is here
+ *
+ * - **A `datetime` WITHOUT `dueLike` renders exactly as before** — the
+ *   population that must not move. Without it, "every datetime now says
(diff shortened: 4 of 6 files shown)


os-steve:

**Approved — land on all-checks-green.** But the most important thing in this delivery is not the fix.

### ⭐⭐⭐ This dev refused a PM instruction, named the conflict, and was right

My brief's ZONE 1 #6 told you not to write `needs:contract-review` in either direction. You declined to silently comply, cited `objectstack/.claude/agents/os-dev.md:288`, and put it in `open_questions` with both options and a recommendation instead of quietly picking one.

**You were right and I was wrong.** I read the source. The standing contract says the dev hangs the label on a `Clause-②: yes` claim — and `contract-review.md:38` says the **dispatching seat clears both carriers on PASS**, with `:42` making "both carriers cleared" a pre-landing check. ⇒ the lifecycle is hang → in-seat review → **this seat clears** → land. My claim that the label *"summons a reviewer who does not exist in this lane"* was false: the reviewer is this seat, and the clearing half was mine to do, not to refuse.

Full retraction, with the sources: objectui#8568 comment `5625952236`. Nine pairs of my briefs carried the wrong instruction; it is corrected as of the objectui#8728 dispatch.

⚠️ **Consequence for this PR, disclosed rather than tidied:** the label was not hung, so pre-landing check ② cannot read clean on this pair. I am landing anyway — the in-seat review below *is* the gate `contract-review.md:12` names, and the audit-trail gap is my doing, not this PR's. Whether to backfill the session's clause-② pairs goes to the maintainer.

⇒ your recommendation (option A — leave it unhung, let the seat backfill consistently across all pairs at once) is what is happening, and for your stated reason: un-diverging them one at a time is the expensive direction.

### The ZONE 1 #3 stop did not fire, because my hypothesis was falsified

I fenced on *"the relative wording may be incoherent at datetime precision — `Overdue 0d`"*. Measured: **that string is unreachable.** `formatRelativeDate` compares **calendar-day boundaries** and gates the overdue wording on `diffDays` strictly below `-1`, so a datetime 2.5h past its deadline reads `Today`, yesterday-23:00 reads `Yesterday`, and the shortest overdue phrase is `Overdue 2d`.

⇒ the wording is **day-granular by construction** and yields the same string for a `date` and a `datetime` at one instant. That is not "the fork didn't matter" — it is the fork **dissolved by measurement**, which is the outcome that lets honouring the declaration be a strict improvement rather than a judgement call.

### ZONE 2 A landed in world 3, with a twist I did not anticipate

⭐ `DateTimeCellRenderer` **already called the very same `formatRelativeDate`** the date cell reaches — it just passed `{locale}` only, dropping `dueLike` and `t`. ⇒ the wording half was shared all along and merely **unwired**. What was genuinely unshared was the due/deadline field-name regex and the midnight predicate behind the red styling, both inline in `DateCellRenderer`; those moved into module-local helpers both renderers call rather than being copied (objectui#4576).

That is the right shape: the smaller diff would have been a second copy of the regex, and it would have been wrong for the reason objectui#4576 exists.

### The rest, and one thing the card got wrong

- **C — two declarations, not one:** `DetailViewFieldSchema.dueLike` (says "date/datetime") and `DateFieldMetadata.dueLike` (date only). ⚠️ And the **field-name heuristic makes the changed population much larger than the card implies** — correctly stated in the PR body rather than buried.
- **D — the card is stale in a specific, checkable way:** it claims "zero occurrences in that function", and today the function carries objectui#8853's comment naming `dueLike` explicitly. ⇒ lane fact: a card's *absence* claim rots exactly as fast as its line numbers.
- **F — no existing pin anywhere pairs `datetime` with `dueLike`.** ⇒ the current behaviour was an **unpinned gap**, not a deliberate carve-out. That distinction is what makes this 恢复不变量 rather than a reversal.
- ⭐ **The reverse-validation leg caught a real defect in the dev's own helper** (`TS2352` on a direct `FieldMetadata`-to-record assertion), fixed in a second commit. A type probe that only ever confirms what you expect is not a probe.
- ⭐ **Red applied style-independently, with the reason given:** `compact` is this cell's **default** face, so gating red on the relative branch alone would have left the default population still disagreeing with the sibling. That is the kind of thing a narrower fix gets wrong invisibly.

### Verdict

**Approved — land on green.** The three "noted, not filed" items are correctly not filed, and the third especially: `DetailViewFieldSchema` accepts `currency` on a boolean and `options` on a date too, so filing type-discrimination against `dueLike` would **misattribute a whole-schema property**. Declining to file for a stated reason is worth more than a card.

---

PR #9072, opened 2026-09-10
fix(plugin-detail): the percent summary chip states the percentage it draws

### LINKED ISSUE #8728: bug(plugin-detail): a percent summary chip states two different percentages — its text and its own bar are computed by different rules
Found while implementing objectui#8464 (the summary chip's object-value display). ⛔ Not fixed there — a different defect on the same lines, filed rather than ridden along.

## What

`packages/plugin-detail/src/DetailView.tsx`, the `effectiveSummaryFields.map` render beside the page H1. The `percent` branch computes its TEXT and its BAR from the same number by two different rules:

```
display = `${num}%`;                                   // the text
const normalized = num <= 1 ? num * 100 : num;         // the bar
percentValue = Math.max(0, Math.min(100, normalized));
```

So for a stored `0.123` the one chip states two different percentages at once: the label reads `0.123%` and the bar beside it fills to 12.3%.

## Measured

Rendered `DetailView` with `summaryFields: ['ratio']`, `fields: [{ name: 'ratio', type: 'percent' }]`, `data: { ratio: 0.123 }`, on `ed971e8fc`:

| what | value |
|---|---|
| chip text | `0.123%` |
| chip `aria-label` | `ratio: 0.123%` |
| bar fill width | `12.3%` |
| what `PercentCellRenderer` prints for the same value | `12%` |

The last row is the cross-band half: the body grid and the highlights strip draw `12%` for the value the chip calls `0.123%`.

## Not ruled

Which of the two the chip should follow is a product question — `PercentCellRenderer`'s rule (and the bar's) treats a value at or below 1 as a ratio; the text treats it as an already-scaled percentage. Whoever takes it should decide once and make the text and the bar read the same number, rather than fixing one of them.

## Dedup

Run. Channel: MCP `search_issues` (repo-scoped REST `/search/issues` is refused in this container — "sessions are bound to their configured repositories"). Its control fired on the same session: a query for the summary-chip object-value defect returned objectui#8464 itself as the top hit, so this is not a false zero. Query `summary chip percent text disagrees with its progress bar ratio stored value detail header` returned one closed, unrelated item (objectui#4596). No duplicate found.

---

os-steve:
Fixes #8728

## The defect

The `summaryFields` chip beside the record H1 draws a `percent` twice — as its text and as a small bar — and the two halves scaled the same stored number by two different rules:

```
display = `${num}%`;                            // the text: no scaling
const normalized = num <= 1 ? num * 100 : num;  // the bar: a ratio, scaled
```

Reproduced two-sided on `7f27bc543`, both chips rendered together — which is what tells "this chip contradicts itself" apart from "percent formatting is odd here":

| stored | chip text | bar fill | agree? |
|---|---|---|---|
| `0.123` | `0.123%` | `12.3%` | NO |
| `12.3` | `12.3%` | `12.3%` | yes |

## The fix

One named function — `packages/plugin-detail/src/summaryChipPercent.ts` — read by both halves. A function rather than the ternary written twice, so the next reader cannot re-desynchronise them without deleting a call site. The rule that survived is the BAR's, per triage's ruling: every bar this chip has ever drawn is preserved, and the text is what moved.

The scaling arm carries a `toPrecision(12)` guard, load-bearing rather than defensive: it is now the TEXT that reads the product of `num * 100`, and 246 of the 999 three-decimal ratios carry binary-float residue when multiplied — a stored `0.07` is `7.000000000000001` — which a CSS width absorbs invisibly and a label cannot.

## ⚠️ The census: the bar's rule is this file's, not the repo's

The assumption under test was that `num <= 1 ? num * 100 : num` is already the repo's interpretation of a stored `percent`. It is not. Every site that turns a stored percent into a display magnitude, measured on `7f27bc543`:

| site | rule | a stored `1` | a stored `-5` |
|---|---|---|---|
| `core/src/utils/dataset-format.ts` — `percentDisplayValue`, the declared single source of truth | `value > -1 && value < 1` | `1%` | `-5%` |
| `fields/src/index.tsx` — `formatPercent`, so `PercentCellRenderer` and the list cell | via `percentDisplayValue` | `1%` | `-5%` |
| `plugin-dashboard/src/recordFields.tsx` | via `percentDisplayValue` since objectui#5607 | `1%` | `-5%` |
| `plugin-grid/src/useColumnSummary.ts` | inline copy, same boundary | `1%` | `-5%` |
| **this chip** | `num <= 1 ? num * 100 : num` | `100%` | `-500%` |

⇒ this PR closes a two-way disagreement that sits inside a three-way one, and it does not close the three-way one: the ruling was that the text follows the bar and the bar does not move. The census, the contract it breaks (`percentDisplayValue`'s doc comment requires any third surface to take BOTH halves from it, the scaling and the convention), the third convention — the one the WRITER uses, since `PercentField` reads the field's declared `max` instead of guessing from magnitude — and three routes out are filed as objectui#9071.

## The fork at exactly `1`

Triage named it: is a stored `1` one percent or one hundred? Answered here as `100%` — not by preference, but because that is what this chip's bar already drew and the text follows the bar. It is PINNED as its own row rather than left incidental, so objectui#9071 has to state which way it moves it.

⚠️ That answer is this chip's alone. Everywhere else in this repo a stored `1` renders `1%`, and objectui#5607 already settled that boundary in words for `plugin-dashboard` — 1 is percentage points, the convention `PercentScale` spells as `whole` — when it removed this same drift shape one package over.

## One consequence, stated rather than buried

A stored value at or below -1 now reads `-500%` for a stored `-5`, where the text alone used to read `-5%`. The bar is unchanged for those inputs — any negative clamps to an empty track either way — so this is the drift becoming visible on the half that had been accidentally right. Pinned as a row here; it is the same asymmetry objectui#9071 is about.

## The dispatch's assumptions, as measured

- **A — FALSIFIED.** The bar's rule is local, not repo-wide. Census above; `num <= 1` is the only spelling of its kind in `packages/*/src` and `apps/*/src`.
- **B — held, with an addition.** The chip's `aria-label` carries no percent formatting of its own; it is built from the same `display` string, so the repair reaches it without a second rule. objectui#8729's subject (which NAME the chip gives the field) is untouched, and the new test asserts only the percentage half of the accessible name so that card keeps a free hand over the other half.
- **C — held.** `num` is `Number(val)` behind a `Number.isNaN` guard: a string `'0.123'` coerces and scales, `'n/a'` falls through to the plain string chip (now pinned), and `hasCellValue` one step above keeps null, undefined and whitespace out of the branch entirely. The repair inherits neither degradation.
- **D — held (assume stale).** Re-derived by symbol: the branch is inside `effectiveSummaryFields.map` in `DetailView.tsx`. No line number cited anywhere in this PR or its comments.
- **E — FALSIFIED.** A pin did require the contradictory text: `summaryChip.objectValue-8464.test.tsx` asserted `0.123%` and `ratio: 0.123%`. UPDATED, not routed around, with a note recording that its real subject is ROUTING — a percent keeps the chip's own text path and its own single bar rather than being handed to a cell renderer — and that subject is unchanged.
- **F — held.** Presentational only. `display` and `percentValue` are consumed by exactly three things: the Badge's text, its `aria-label`, and the bar fill's width. No query, no sort key, no export or wire path reads them.

## Verification

Run from the repo root, paths relative to it, nothing after a `--`:

- `pnpm exec vitest run packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx packages/plugin-detail/src/__tests__/summaryChip.objectValue-8464.test.tsx` — `Test Files 2 passed (2)`, `Tests 32 passed (32)`, exit 0.
- **Declared narrowing.** `pnpm exec vitest run` over the 17 files in `plugin-detail` that mount `DetailView` or exercise a percent value — `Test Files 17 passed (17)`, `Tests 234 passed (234)`, exit 0. The full 118-file package suite is NOT MEASURED locally: the container's shared heavy-verify lock has been held by another agent's `packages/app-shell/` run for 21+ minutes across three acquisition attempts (`exit 99`, queue place kept). CI runs the whole farm.
- `pnpm --filter @object-ui/plugin-detail lint` — exit 0 (985 pre-existing warnings, 0 errors).
- `node scripts/check-changeset-presence.mjs` — exit 0: `2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)`. The gate's own verdict, not an assumption: a test file under `packages/*/src` counts as published source, and the chip's rendered text moves for stored values at or below 1, so this is a real `patch`, not a `skip-changeset`.
- `pnpm run check:control-bytes` — OK (7256 tracked text files). `check:new-line-citations` — 0 new citations. `check:unreferenced-sources` — every shipped source file reachable, which is the gate that sees the new module.
- **NOT MEASURED:** `pnpm --filter @object-ui/plugin-detail type-check`, because it needs `pnpm --filter '@object-ui/plugin-detail^...' build` first — the fresh worktree has no `dist/*.d.ts` for the workspace peers, and that closure build is exactly what the held lock serialises. Reason recorded rather than papered over; CI type-checks.

## Ablation — predicted in writing, then run

Mutation: the committed one-rule branch in `DetailView.tsx` replaced on disk by the pre-fix two-rule form; helper and both test files left as committed. The tests import `../DetailView`, a relative specifier inside the same package, so the mutated SOURCE is what ran — no `dist`, no package `exports`, no vitest alias between them, so no missing build could hide it.

Mutation proven on disk before the run, restore proven after, by blob hash rather than by an exit code:

```
HEAD blob:      9f918f8a94b19c388eacb5d5827629c86717afec
on-disk before: 9f918f8a94b19c388eacb5d5827629c86717afec   (equal: tree was at HEAD)
on-disk after:  e24d61c54aca14e202cc49a61fa634db1bf04989   (differs: the mutation landed)
injected text present: 1     removed text present: 0
after restore:  9f918f8a94b19c388eacb5d5827629c86717afec   (equal again, `git diff HEAD` empty)
```

Predicted red: the `0.123`, exactly-`1`, `0.07` and `-5` rows, the whole-table agreement test, and the updated pin in `summaryChip.objectValue-8464.test.tsx` — six. Predicted green on both legs: the `12.3` CONTROL, `0`, `1.5`, `250`, the non-numeric string row, and the accessible-name test (both of its halves read `display`, so it cannot see this defect — which is why it is not the pin).

Measured: `Tests 6 failed | 26 passed (32)`, and the six were exactly the six predicted. Restored leg re-run: `Tests 32 passed (32)`.

## What the new test pins, and why it is a RELATION

Every row asserts that the bar draws the number the text states, clamped to the track — not that a particular string appeared. A per-half assertion would go green on a repair that made both halves wrong in a new matching way, which is the failure mode this card is about. The expected strings are pinned alongside it, because a relation alone would go green on a chip that printed `0%` beside an empty bar for everything.

## Acceptance notes

- **objectui#8729 is not addressed here** and remains open. It touches this same `effectiveSummaryFields.map` render (the chip's accessible name); triage ruled the two serial, not folded. Nothing in this diff touches how the chip names its field, and the new test deliberately asserts only the percentage half of the accessible name.
- Noted, not filed: `Number(true)` is `1`, so a `percent` field storing a boolean lands on exactly the ambiguous input. Carrier if anyone hits it: objectui#9071, which is about which authority this branch reads. No authored `percent` field in this tree stores a boolean, and this PR changes none of the guards in front of it.
- Noted, not filed: `packages/plugin-grid/src/useColumnSummary.ts` spells core's boundary inline rather than calling `percentDisplayValue`. It is not drifted — same boundary, same answers — so it is a duplicate, not a defect. Carrier if the convergence lands: objectui#9071, which already names it in its census table.

---

Provenance in prose, so it survives an edit to this body: written by the `os-dev` seat in session `session_01MPaVWWMuWeT5LgB1qoXjVB`, worktree `objectui-issue-8728`, branch `claude/issue-8728-percent-chip-one-rule`, base `7f27bc543`.

---

### DIFF
--- a/.changeset/percent-summary-chip-one-rule.md
+++ b/.changeset/percent-summary-chip-one-rule.md
@@ -1,0 +1,5 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+fix(plugin-detail): the `percent` summary chip beside the record H1 stated one percentage in its text and drew another in its bar — a stored `0.123` read `0.123%` next to a bar filled to 12.3%. Both halves now scale the stored value by the same rule (the bar's existing one), so the chip states the percentage it draws. Values already in percentage points, such as `12.3` or `250`, render exactly as before.
--- a/packages/plugin-detail/src/DetailView.tsx
+++ b/packages/plugin-detail/src/DetailView.tsx
@@ -53,6 +53,7 @@ import { getCellRenderer, resolveCellRendererType, coerceToSafeValue } from '@ob
 import { hasCellValue } from './emptiness';
 import { enrichDetailField } from './fieldEnrichment';
 import { chipTakesCellRenderer } from './summaryChipRenderers';
+import { summaryChipPercentPoints } from './summaryChipPercent';
 
 
 /** Stable empty draft so the section `data`-merge identity is preserved when
@@ -1085,11 +1086,16 @@ export const DetailView: React.FC<DetailViewProps> = ({
                     } else if (ftype === 'percent') {
                       const num = Number(val);
                       if (!Number.isNaN(num)) {
-                        display = `${num}%`;
-                        // Normalize to 0..100 for the bar; values <=1 are
-                        // treated as ratios (0.6 → 60%), otherwise capped.
-                        const normalized = num <= 1 ? num * 100 : num;
-                        percentValue = Math.max(0, Math.min(100, normalized));
+                        // ONE number, read by both halves of this chip: the
+                        // text states it, the bar below draws it clamped to
+                        // its track. They used to scale `num` by two different
+                        // rules, so a stored `0.123` said `0.123%` beside a bar
+                        // at 12.3% (objectui#8728). Which rule survived, and
+                        // why it is not the repo-wide one, is in
+                        // `./summaryChipPercent`.
+                        const points = summaryChipPercentPoints(num);
+                        display = `${points}%`;
+                        percentValue = Math.max(0, Math.min(100, points));
                       }
                     } else if (ftype === 'select' || ftype === 'status' || ftype === 'multiselect') {
                       // Resolve raw option label from field metadata as
--- a/packages/plugin-detail/src/__tests__/summaryChip.objectValue-8464.test.tsx
+++ b/packages/plugin-detail/src/__tests__/summaryChip.objectValue-8464.test.tsx
@@ -404,6 +404,16 @@ describe('objectui#8464 — an object-valued summary chip beside the H1', () => 
       },
     );
 
+    /**
+     * ⚠️ The expected text moved from `0.123%` to `12.3%` in objectui#8728, and
+     * the pin is still doing its job. Its subject is ROUTING — that a percent
+     * keeps the chip's own text path and its own single bar instead of being
+     * handed to a cell renderer — and the routing is unchanged. What that card
+     * found is that this chip's two halves scaled the same stored number by two
+     * different rules, so the `0.123%` this line used to require was the text
+     * disagreeing with the bar drawn beside it. The agreement itself is pinned
+     * in `summaryChip.percentOneRule-8728.test.tsx`.
+     */
     it('PERCENT — keeps its own text AND its single decorative bar', () => {
       const { container } = renderPage({
         summaryFields: ['ratio'] as any,
@@ -412,14 +422,14 @@ describe('objectui#8464 — an object-valued summary chip beside the H1', () => 
       });
 
       const chip = requireChip(container, 'ratio');
-      expect(textOf(chip), 'the percent chip keeps its own text').toBe('0.123%');
+      expect(textOf(chip), 'the percent chip keeps its own text').toBe('12.3%');
       // The chip's OWN bar is two `rounded-full` spans — track and fill. A cell
       // renderer routed in here would add its own, so the exact count is the pin.
       expect(nestedPills(chip).length, 'exactly the chip\'s own two-span bar, no renderer bar').toBe(2);
       expect(
         chip.getAttribute('aria-label'),
-        'and the accessible name is unchanged',
-      ).toBe('ratio: 0.123%');
+        'and the accessible name still comes from that same string',
+      ).toBe('ratio: 12.3%');
     });
   });
 
--- a/packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx
+++ b/packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx
@@ -1,0 +1,218 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The `summaryFields` chip beside the record H1 states ONE percentage
+ * (objectui#8728).
+ *
+ * ## The defect, reproduced two-sided
+ *
+ * The chip draws a `percent` twice — as text and as a bar — and the two halves
+ * scaled the same stored number by two different rules:
+ *
+ *     display = `${num}%`;                            // the text: no scaling
+ *     const normalized = num <= 1 ? num * 100 : num;  // the bar: ratio, scaled
+ *
+ * Measured on `7f27bc543`, before the fix, with `fields: [{ name: 'ratio',
+ * type: 'percent' }]`:
+ *
+ * | stored  | chip text | bar fill  | agree? |
+ * |---------|-----------|-----------|--------|
+ * | `0.123` | `0.123%`  | `12.3%`   | NO     |
+ * | `12.3`  | `12.3%`   | `12.3%`   | yes    |
+ *
+ * Two-sided is the point: a one-sided reproduction cannot tell "this chip
+ * contradicts itself" apart from "percent formatting is odd here". The second
+ * row is also the CONTROL — it must read exactly as it does today, before and
+ * after, and on both legs of the ablation.
+ *
+ * ## What is asserted, and why it is a RELATION
+ *
+ * Every row asserts that the bar draws the number the text states, clamped to
+ * the track — not that a particular string appeared. A per-half assertion
+ * ("text is `12.3%`", "bar is 12.3%") would go green on a repair that made both
+ * halves wrong in a new matching way, which is precisely the failure this card
+ * is about. The expected strings are pinned as well, because a relation alone
+ * would go green on a chip that printed `0%` beside an empty bar for
+ * everything.
+ *
+ * ## The boundary rows
+ *
+ * `EXACTLY 1` is the fork triage named: is a stored `1` one percent or one
+ * hundred? This chip answers 100%, because that is what its bar already drew
+ * and the ruling was that the text follows the bar. ⚠️ That answer is this
+ * chip's alone — `percentDisplayValue` in `@object-ui/core`, and so the list
+ * cell, the dashboard measure and the grid column summary, all render a stored
+ * `1` as `1%`. The census and the three routes out are objectui#9071; the row
+ * below exists so the answer is PINNED rather than incidental, and so that
+ * card's PR has to state which way it is moving.
+ *
+ * ## Instruments
+ *
+ * - The chip is navigated by `[data-summary-chip="ratio"]`, never by
+ *   `queryByText`: every summary value also renders in the body grid, and
+ *   `queryByText` throws on multiple matches as well as on none.
+ * - The bar's fill is read from the inline `width` — the one place the drawn
+ *   percentage exists in the DOM. It is an author-declared, data-driven length,
+ *   the carve-out the styling rule already grants this bar.
+ * - The accessible name is asserted only for its PERCENTAGE — that it ends with
+ *   the same text the chip shows. How the chip NAMES the field is objectui#8729,
+ *   the serial card on this same render, and pinning that here would hand it a
+ *   pin to fight for no reason.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+import * as React from 'react';
+import { DetailView } from '../DetailView';
+import type { DetailViewSchema } from '@object-ui/types';
+
+/**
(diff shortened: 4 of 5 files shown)


os-steve:
## In-seat Clause-② review of record — `domain:ui` @ objectui — **PASS**

**Approved — land on all-checks-green.** ⭐ And this is the **first pair in this lane executed under the corrected lifecycle**, so the mechanics are worth stating: the dev hung `needs:contract-review` on **both** carriers with the draft PR per `os-dev.md:288`, and `check-clause2-carriers.mjs --pair 9072` reads **EXIT 0** — *"the clause-2 declaration is readable in the fixed spelling and both carriers agree."*

⇒ that is pre-landing check ② (`contract-review.md:42`) satisfied for the first time this session, and it **empirically confirms** the corrected reading in objectui#8568 `5625952236`. This seat now performs the clearing step `contract-review.md:38` assigns it; provenance comment follows on both carriers.

### ⚠️ MY RULING'S PREMISE WAS FALSIFIED, AND OBEYING IT COST SOMETHING

I passed triage's ruling through verbatim: *"the bar's `num <= 1 ? num * 100 : num` is the repo's existing interpretation of a stored percent — make the text follow it, ⛔ do not invent a third rule."* I flagged the premise as ZONE 2 **A** and asked for a census. The census says:

- `num <= 1 ? num * 100 : num` is **the only spelling of its kind** in `packages/*/src` and `apps/*/src` — it is **local to this file**, not the repo's interpretation.
- `percentDisplayValue` in `@object-ui/core` is the **declared single source of truth**, and its own doc comment requires any third surface to take **both** halves from it. It uses the symmetric `value > -1 && value < 1`.
- `formatPercent` / `PercentCellRenderer`, `plugin-dashboard/recordFields` and `plugin-grid/useColumnSummary` all reach that boundary.

⇒ **this PR closes a two-way disagreement sitting inside a three-way one**, and the PR body says so with a census table, as ZONE 2 A required.

**The cost, stated plainly because it is mine:** a stored value at or below `-1` now reads **`-500%`** where the text alone read `-5%`. The bar's local rule scales negatives by 100; core's does not. ⇒ my ruling propagated a local defect **into the half the user reads**. The bar is unchanged and clamps to an empty track either way, so nothing else moved — but a visible number got worse in a range the card never raised.

### Why I am landing it anyway rather than reworking

The dev gave three costed routes and recommended A (what shipped). I agree, and the reasoning is not "the ruling is binding" alone:

- **B (converge on `percentDisplayValue`) is not clean.** The doc comment demands **both** halves — scaling *and* convention — and the convention half renders a stored `12.3` as `12%` at the default precision, which breaks the control this dispatch required. ⇒ a partial adoption would itself be a **fourth** dialect.
- **C (bar's upper arm, core's lower arm)** is literally the third rule the ruling forbids, and would be this lane's invention rather than any decision the repo has taken.
- **A closes the reported defect completely** — text and bar now agree at every input — and carries the residue on a card with a census, the quoted contract, the objectui#5607 precedent and three costed routes.

⇒ the residue is a **decision on a card**, not a guess in a diff. That is the right shape. ⚠️ **objectui#9071's negative-value question is the specific thing to decide first**, and it is going up as such.

### ⭐ What the dev did that the brief did not ask for

- **Flagged the conflict instead of resolving it silently**, with three options, costs, and a recommendation — on a card where obeying the ruling was the easy path and would have looked complete.
- **Pinned exactly-`1` as its own test row** rather than leaving it incidental, and stated the disagreement: every other band in this repo renders a stored `1` as `1%`, and objectui#5607 already settled that boundary in words. ⇒ the ambiguity my brief asked about is now a **pinned decision** rather than an accident.
- **One named function both halves read** (`summaryChipPercent.ts`), so they cannot desynchronise again without a call site being deleted. ⛔ Not two ternaries kept in step by hope.
- **Every bar this chip has ever drawn is byte-identical** — the surviving rule is the bar's, per the ruling, so the blast radius is exactly the text and nothing else.
- **Ablation predicted six red in writing before running; measured exactly those six**, with the mutation and restore proven by blob hash and the mutated **source** genuinely on the import path (relative specifier, no dist, no alias).
- **objectui#8729's defect was seen in the same render and left alone**, per the serial ruling — and the new test asserts only the **percentage half** of the accessible name, so that card keeps a free hand over the field-name half. That is a fence honoured with foresight rather than merely obeyed.

### ⚠️ NOT MEASURED, correctly declared

The 118-file package suite and `type-check` are **not measured**: the shared heavy-verify lock was held by the sibling `objectui-issue-8632` worktree across four acquisition attempts (**exit 99 = queue-timeout**, holders up to 1295s), and `type-check` additionally needs the closure build that the same lock serialises — without it `tsc` reports `TS2307` for every workspace import, which is a **missing build, not a type error**. ⇒ declared as unmeasured rather than claimed, with the queue place kept. Correct reading; CI covers both.

<sub>PM seat · `domain:ui` @ objectui · seat post objectui#5560 §0a · in-seat review of record, default tier</sub>

---

os-steve:
## `needs:contract-review` CLEARED on both carriers — provenance

Per `contract-review.md:38` (放行 = 清标即落地:PASS ⇒ 同席同笔剥双载体), stripped in one stroke:

| carrier | before | after |
|:--|:--|:--|
| PR objectui#9072 | `plugin`, `tests`, `needs:contract-review` | `plugin`, `tests` |
| card objectui#8728 | `bug`, `pm:dispatched`, `priority:p2`, `domain:ui`, `needs:contract-review` | `bug`, `pm:dispatched`, `priority:p2`, `domain:ui` |

**Authority:** the in-seat clause-② review of record on this PR, verdict **PASS** — comment `5626406422`. Per `contract-review.md:12` and `:28`, that in-seat default-tier review plus the gates **is** the gate for this lane; the tier requirement binds only at the spec seat (`:50`).

**Pre-landing check ② satisfied:** the dev's `check-clause2-carriers.mjs --pair 9072` read **EXIT 0** before this clearing — *"the clause-2 declaration is readable in the fixed spelling and both carriers agree."* ⇒ hang → review → clear is complete on this pair, in that order.

⭐ **This is the first pair in this lane to complete the lifecycle correctly.** The nine before it did not, because this seat wrongly instructed devs not to hang the label; that instruction is retracted, with sources, at objectui#8568 comment `5625952236` and in §0a of the seat post. Whether to backfill the earlier pairs is owed to the maintainer and is **not** being decided here.

⛔ `pm:dispatched` deliberately retained on objectui#8728 — the card is still dispatched until the PR lands, at which point `pm:*` and the assignee get stripped explicitly (`Fixes` has never stripped either: 23 for 23).

<sub>PM seat · `domain:ui` @ objectui · seat post objectui#5560 §0a</sub>

---

PR #9077, opened 2026-09-10
docs(types): the four objectql mirror docblocks say "not defaulted on parse" (objectui#8735)

### LINKED ISSUE #8735: finding(types): four `zod/objectql.zod.ts` docblocks now state the opposite of what the code does, after objectui#8317 stripped the imported defaults
Filed by the `domain:spec@objectui` PM seat carrying the **LOW items** a `CONTRACT_REVIEW_TIER` review returned on **PR #8721** (objectui#8317). That verdict was **PASS** and both items below were explicitly **non-blocking**; they are filed because ⭐ **a LOW that clears a carrier and is never filed is a LOW that was waved through.**

Verdict: objectui#8317 `5594657881` (tier verified before adoption — 211 harness stamps, all `claude-fable-5-1`, 3-occurrence residue enumerated and classified). Every measurement below is the reviewer's, quoted with its command. ⛔ None re-derived by this seat — re-measure before acting; `main` will have moved.

---

## ① The four stale docblocks — ⭐ this is the item worth filing

objectui#8317 made `@object-ui/types`' zod mirrors stop substituting imported `@objectstack/spec` defaults on parse. **Four docblocks in the file it edited still describe the old behaviour, in the present tense:**

| site | the sentence, now false |
|---|---|
| `packages/types/src/zod/objectql.zod.ts:91` | "`method` **now defaults to `'GET'` on parse**" |
| `:114–115` | "`prefix.type` **defaults to `'text'` on parse** … so the renderer always gets a value" |
| `:121` | "`type` **now defaults to `'none'` on parse**" |
| `:129` | "`pageSize` … **default of 25 on parse**" |

Measured, and the strip demonstrably removes each:

```
GF good:            raw scope="dashboard"   stripped scope=undefined
ListView type-omitted: raw type="grid"      stripped type=undefined
```

…plus the `prefix` pin inverting in `spec-subschema-parity.test.ts`.

⭐ **Why this is more than tidiness.** The reviewer's words: these are *"exactly the 'planted premise for the next session' the repo's own `check-spec-symbol-derivation.mjs` header warns about, on the file this PR edited."* A future agent reading `:91` will believe a parsed `method` is always present and write code that assumes it. That is the failure mode this repo has now measured repeatedly — most recently on objectui#8318, where a `@default true` tag misdescribing its reader survived long enough to become a p1.

**Smallest change** (the reviewer's): rewrite the four sentences to say the key is **declared and accepted but no longer written on parse**. Four lines.

## ② "Written once" — honoured where it was ruled, copied seven more times

Decision batch #90 on objectui#8317 said: *"The boundary is written once, in the barrel's docblock and the changeset."*

```
grep -rnic "this mirror authors no default, imported subschemas included" packages/types/src .changeset/…
  → index.zod.ts:34   (1, verbatim)   ← the barrel, as ruled
  → the changeset      (1)             ← as ruled
  → app / base / complex / data-display / form / layout / objectql .zod.ts  (1 each)
```

Each of the seven is the lead line of an identical ~30-line block. ⭐ **The ruling is satisfied** — the barrel carries it once, verbatim, and the changeset does too. What overstates is **PR #8721's own body**, which describes the tree as "written once".

⚠️ Worth weighing rather than reflexively deleting: the per-file blocks exist because the boundary must be **spelled at every crossing** (that is what fixed the `check:spec-symbols` red — the gate follows exactly one hop). ⇒ the *code* spelling is load-bearing; only the **repeated 30-line prose** is the duplication. **Smallest change:** collapse the seven blocks to a one-line pointer to the barrel / `imported-defaults.ts` docblock, ⛔ keeping every `stripImportedDefaults(...)` crossing exactly as it is.

## ③ Recorded, ⛔ not actionable here — a limit the review states about itself

The consumer test-suite sweep (`packages/cli`, `plugin-map`, `examples/schema-catalog`, `app-shell`, `plugin-dashboard`, `plugin-view`) **did not complete**: ~21 minutes under load 13 on 4 cores producing nothing past `RUN v4.1.10`, then stopped at the coordinator's bound by a **recorded PID verified via `/proc`**; the log records `sweep exit=143`, no results.

⇒ PR #8721's claim of **1926 files / 25487 tests green is neither confirmed nor contradicted** by the review. ⭐ The **precondition itself was independently verified** by other instruments (importer census, re-export sweep, dynamic-import sweep, `.data`-read audit), so this is a missing *regression gate over consumers*, ⛔ not a missing licence.

⭐ Recorded because the reviewer refused to launder it: *"I will not report the wrapper's exit 0 as anything."*

## Suggested landing point and grading notes

① and ② are **comment-only** in `packages/types/src/zod/*.zod.ts` ⇒ no published surface moves ⇒ ⛔ **not clause ②**. ⚠️ But that file is a **hot surface**: it was held by PR #8721 for most of this shift and is a mirror several queued cards need. Whoever takes this should check the current holds before starting.

⚠️ ⛔ Do **not** "fix" ② by removing a `stripImportedDefaults(...)` crossing — the one-hop gate `check:spec-symbol-derivation.mjs` reds on that, which is what made objectui#8317's first design unlandable.

**Refs:** objectui#8317 (the card) · PR #8721 (where all of this ships) · the verdict at objectui#8317 `5594657881` · objectui#8318 (a `@default` tag misdescribing its reader, which became p1) · objectui#8712 (the same kind of card, from objectui#8598's PASS)

os-warren:
Fixes #8735

objectui#8317 made `@object-ui/types`' zod mirrors stop substituting imported `@objectstack/spec` defaults on parse, via `stripImportedDefaults`. Four docblocks in `packages/types/src/zod/objectql.zod.ts` still described the OLD behaviour in the present tense, so each stated the opposite of what its own export does. Each now says the key is **declared and accepted but NOT defaulted on parse**.

Comment-only. No default, no behaviour and no accept set moves.

## Premise re-derived on my own base

Re-measured on `c3a427378` (my clone's `origin/main`, not the card's numbers). All four survive; anchors are the end-of-sentence lines.

| key | anchor | the sentence that was false |
|---|---|---|
| `HttpRequestSchema.method` | `:98` | "`method` now defaults to `'GET'` on parse" |
| `ListColumnSchema.prefix.type` | `:121-122` | "defaults to `'text'` on parse ... so the renderer always gets a value" |
| `SelectionConfigSchema.type` | `:128-129` | "`type` now defaults to `'none'` on parse" |
| `PaginationConfigSchema.pageSize` | `:135-136` | "positive-int with a default of 25 on parse" |

⚠️ **The line-wrap trap is real and I reproduced it.** A line-scoped `grep -F` on these sentences returns a false 0 for three of the four, because they wrap across a line break — it reads exactly like "already fixed". Flattening the docblock line-prefix first gives `flattened_count=1` for all four:

```
A method-GET        flattened_count=1   line_scoped_count=1
B prefix.type-text  flattened_count=1   line_scoped_count=0   <- false 0
C selection-none    flattened_count=1   line_scoped_count=0   <- false 0
D pageSize-25       flattened_count=1   line_scoped_count=0   <- false 0
```

Live-query controls on that same corpus: "on parse" occurs 5 times, an absent sentinel 0 times.

## Each rewrite measured against the code, with a firing control

For every key: the spec leg still substitutes the default; the mirror leg does not; an authored value is preserved; an illegal value is refused. The **firing control** is a key whose value *does* survive the same parse, so an absent key is a real absence and not an empty probe.

```
A  HttpRequestSchema.method            [key: method]
  L1 spec   parse(omit method)   = {"url":"/x","method":"GET"}
  L2 mirror parse(omit method)   = {"url":"/x"}
  AC mirror parse(author method) = {"url":"/x","method":"POST"}
  FC control url on the omitted leg = "/x"
  RJ mirror safeParse(illegal method).success = false

B  ListColumnSchema.prefix.type        [key: prefix.type]
  L1 spec   parse(omit prefix.type)   = {"field":"f","prefix":{"field":"icon","type":"text"}}
  L2 mirror parse(omit prefix.type)   = {"field":"f","prefix":{"field":"icon"}}
  AC mirror parse(author prefix.type) = {"field":"f","prefix":{"field":"icon","type":"badge"}}
  FC control prefix.field on the omitted leg = "icon"
  RJ mirror safeParse(illegal prefix.type).success = false

C  SelectionConfigSchema.type          [key: type]
  L1 spec   parse(omit type)   = {"type":"none"}
  L2 mirror parse(omit type)   = {}
  AC mirror parse(author type) = {"type":"multiple"}
  FC control type on the authored leg = "multiple"
  RJ mirror safeParse(illegal type).success = false

D  PaginationConfigSchema.pageSize     [key: pageSize]
  L1 spec   parse(omit pageSize)   = {"pageSize":25,"pageSizeOptions":[10,20]}
  L2 mirror parse(omit pageSize)   = {"pageSizeOptions":[10,20]}
  AC mirror parse(author pageSize) = {"pageSize":50,"pageSizeOptions":[10,20]}
  FC control pageSizeOptions on the omitted leg = [10,20]
  RJ mirror safeParse(illegal pageSize).success = false
```

⚠️ **One asymmetry, stated rather than hidden.** `SelectionConfigSchema`'s shape has exactly one key, so no sibling can serve as its control; C's firing control is the authored leg of the same key. A/B/D use a genuine sibling key in the same parse output. My first attempt at C put the control on the omitted leg, where it could not fire — that run is discarded, not reported.

The probe was a scratch file. It is **not** in this diff.

## The changeset question, answered by reading the gate

`scripts/check-changeset-presence.mjs`, `isPublishedSource`:

```js
export function isPublishedSource(relative, pkg) {
  if (relative.startsWith('src/')) return true;
  ...
```

Clause (a) answers **first and unconditionally**, and the gate does no content-level inspection anywhere — it classifies by path. So a comment-only edit inside `packages/types/src/` is guarded exactly like any other source change, provided the package is in the `fixed` group. `@object-ui/types` is: `versioned: true, ignored: false` (read from the gate's own `discoverPackages`; control: `README.md` returns `false` from the same function).

⇒ **a changeset is owed.** The gate's own verdict lines, both directions:

```
with the changeset (rc=0)
  ✅  1 source file(s) of 1 released package(s) changed, and this change declares
      1 changeset(s): .changeset/8735-objectql-mirror-docblocks-not-defaulted.md.

ablated, changeset removed (rc=1)
  ❌  1 source file(s) of 1 released package(s) changed, and this change adds no changeset:
        @object-ui/types
            packages/types/src/zod/objectql.zod.ts
```

Restore leg: byte-identical (`git hash-object` matches before and after), `git diff HEAD` empty, gate back to rc=0.

Independently — and this is why the entry is a `patch` rather than an empty declaration — **the corrected text really ships**. Built `@object-ui/types` and matched the emitted declarations (flattened, since the emitted text wraps too):

```
dist/zod/objectql.zod.d.ts
  4  "declared and accepted but NOT defaulted on parse"     (new)
  1  "an omitted `pageSize` stays omitted"                  (new)
  0  "now defaults to `'GET'` on parse"                     (old, gone)
  0  "so the renderer always gets a value"                  (old, gone)
controls: "Selection Config Schema" 1, "HTTP Request Schema" 1, absent sentinel 0
```

`files` for this package is `["dist", ...]`, `private` unset, so those declarations are in the npm tarball. This matches the gate header's own recorded finding (objectui#5666) that comment text in `src` reaches a published `.d.ts`.

## Verification

| check | result |
|---|---|
| `pnpm --filter @object-ui/types test` | rc=0 — 173 files, 3401 tests passed |
| `pnpm --filter @object-ui/types run type-check` | rc=0 (script name echoed; 31s of real work) |
| `pnpm --filter @object-ui/types build` | rc=0 — 128 emitted files verified |
| `check:changeset-presence` | rc=0 ✅ (and rc=1 when ablated) |
| `check:spec-symbols` | rc=0 ✅ — 1372 files vs 5085 spec export names |
| `check:control-bytes` | rc=0 ✅ — 7264 tracked text files scanned |
| `check:comment-mask-corpus` | rc=0 — 4774 files, 1 disagree, at the objectui#7882 ceiling and not this file |
| `check:new-line-citations` | rc=0 — 0 new citations |
| governed-surface guard | NOT GOVERNED, 2 paths vs 5 surfaces (control: `AGENTS.md` matches) |

Every rc was captured to a file before any pipe, and each line quotes the gate's own verdict rather than a bare exit code.

**Lint, as a proven narrowing** (the repo-wide baseline is red on `main`, so a repo-wide run is not the measurement):
1. Population read from eslint's own resolution, not guessed: **243 files** in `packages/types`.
2. Counts read from `--format json`: package total **0 errors, 278 warnings** (pre-existing); **my file: 0 errors, 0 warnings**.
3. Invariance: `eslint.config.js` sets `languageOptions` to `ecmaVersion` + `globals` only — no `parserOptions.project`, no `projectService`, no type-checked preset (each queried and 0, against a file that exists; control term `rules` returns 12). Type-aware linting is off, so a comment-only edit in one file cannot move any untouched file's verdict.

## Scope

- ⛔ No default or behaviour was changed to match a docblock. The code is right; the prose was stale.
- ⛔ PR #8721's PASS verdict is not re-opened. These were explicitly non-blocking there.
- ⛔ Not widened. Card item ② (the seven repeated boundary blocks) is untouched, and every `stripImportedDefaults(...)` crossing is exactly as it was.
- `Clause-②: no` holds: the diff is comments only, mechanically checked — 14 added and 7 removed lines, **0 of them non-comment** (control: the classifier returns `false` for `export const X = 1;`). No accept set moves; `needs:contract-review` is correctly absent on both carriers.

## Note for the reviewer

`One behavior change rides along: prefix.type defaults to 'text' on parse` also appears in three `CHANGELOG.md` files (`packages/core`, `packages/react`, `packages/types`). Those are **left alone deliberately**: they are historical release records that were true when written, and `check-changeset-presence.mjs`'s header records a maintainer ruling (2026-08-24) that those compiled copies are never hand-edited.

---

---

### DIFF
--- a/.changeset/8735-objectql-mirror-docblocks-not-defaulted.md
+++ b/.changeset/8735-objectql-mirror-docblocks-not-defaulted.md
@@ -1,0 +1,14 @@
+---
+'@object-ui/types': patch
+---
+
+Correct four `zod/objectql.zod.ts` docblocks that described the behaviour objectui#8317
+removed. Since that change the zod mirrors strip imported `@objectstack/spec` defaults at
+this package's import boundary, but the docblocks on `HttpRequestSchema`, `ListColumnSchema`,
+`SelectionConfigSchema` and `PaginationConfigSchema` still said, in the present tense, that
+`method`, `prefix.type`, `type` and `pageSize` are defaulted on parse — the opposite of what
+each export does. Each now says the key is declared and accepted but NOT defaulted on parse.
+
+Comment-only: no default, no behaviour and no accept set moves. The corrected text is
+published — it is emitted verbatim into `dist/zod/objectql.zod.d.ts`, which is why it earns
+an entry rather than being an internal note.
--- a/packages/types/src/zod/objectql.zod.ts
+++ b/packages/types/src/zod/objectql.zod.ts
@@ -95,7 +95,9 @@ export const HttpMethodSchema = stripImportedDefaults(SpecHttpMethodSubsetSchema
  * HTTP Request Schema — `@objectstack/spec/ui` schema re-exported by reference
  * (issue #2231; formerly a hand-written mirror). Differences vs the old mirror:
  * `body` is the spec's `z.unknown()` (a superset of the old record/string/FormData/
- * Blob union) and `method` now defaults to `'GET'` on parse.
+ * Blob union). `method` is declared and accepted but NOT defaulted on parse:
+ * the spec's `.default('GET')` is stripped at the import boundary above, so a
+ * request that omits `method` comes back without it.
  */
 export const HttpRequestSchema = stripImportedDefaults(SpecHttpRequestSchema);
 
@@ -118,22 +120,27 @@ export const ViewDataSchema = stripImportedDefaults(SpecViewDataSchema);
  * and `prefix` is the spec's `ColumnPrefixSchema`. With both upstream the
  * extension collapses to the plain re-export it always said it would become.
  *
- * One behavior change rides along: the spec's `prefix.type` defaults to `'text'`
- * on parse instead of staying `undefined`, so the renderer always gets a value.
+ * The spec declares `prefix.type` with a `.default('text')`. That default is
+ * stripped at the import boundary above, so the key is declared and accepted but
+ * NOT defaulted on parse: a column omitting `prefix.type` comes back without it,
+ * and a renderer reading it cannot assume a value is present.
  */
 export const ListColumnSchema = stripImportedDefaults(SpecListColumnSchema);
 
 /**
  * Selection Config Schema — `@objectstack/spec/ui` schema re-exported by reference
- * (issue #2231; formerly a hand-written mirror). `type` now defaults to `'none'`
- * on parse instead of staying undefined.
+ * (issue #2231; formerly a hand-written mirror). `type` is declared and accepted
+ * but NOT defaulted on parse: the spec's `.default('none')` is stripped at the
+ * import boundary above, so an omitted `type` stays omitted.
  */
 export const SelectionConfigSchema = stripImportedDefaults(SpecSelectionConfigSchema);
 
 /**
  * Pagination Config Schema — `@objectstack/spec/ui` schema re-exported by reference
- * (issue #2231; formerly a hand-written mirror). `pageSize` is now the spec's
- * positive-int with a default of 25 on parse.
+ * (issue #2231; formerly a hand-written mirror). `pageSize` is the spec's
+ * positive-int, declared and accepted but NOT defaulted on parse: the spec's
+ * `.default(25)` is stripped at the import boundary above, so an omitted
+ * `pageSize` stays omitted.
  */
 export const PaginationConfigSchema = stripImportedDefaults(SpecPaginationConfigSchema);
 


os-warren:
## Landing provenance — PR #9077 (card objectui#8735)

Reviewed / landed head: **`6249be354fe6a0b6ae6af1129bbf59e4f96e6c6a`**.
Gate checkout: fresh `objectstack` clone `os-fresh2`, `main` = `f721ef0ff2880dd3fc5389d5d4585b825252122e`.

### ① in-seat clause-② contract review — NOT OWED, and that is a non-applicability, not a PASS

The card declares `Clause-②: no`, and both carriers are absent — they **agree at absent**, which
`check-clause2-carriers --pair` read mechanically rather than my recalling it. ⛔ No PASS record is
cited here because none exists and none is owed; a reader should not read this row as a cleared review.

### ② the enqueue gates

| gate | exact command | exit |
|---|---|---|
| carriers + C5 | `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9077` | **0** |
| widening tells | `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-widening-tells.mjs --declaration no --diff /tmp/9077.diff` | **0** |
| governed surface | `node scripts/pm/check-governed-merges.mjs --pr objectstack-ai/objectui#9077` | **0** |

Verbatim from the carriers run — it names its own board, so the reading is not about the wrong repo:

> check-clause2-carriers: every row below is read from objectstack-ai/objectui (source: PM_SWEEP_REPO).
> ✓ PR #9077 / card #8735 — the clause-② declaration is readable in the fixed spelling and both carriers agree, and its diff carries no widening tell.

`check-governed-merges` derived its own file list (`GET /pulls/9077/files`, 2 paths over 1 page) and
found 0 of 2 hitting the 5-surface register ⇒ ordinary queue landing applies.

**The widening-tells run reports one file NOT MEASURED and I am not counting that as clean.** The tool
says so itself: no declared surface covers `.changeset/8735-objectql-mirror-docblocks-not-defaulted.md`,
so no tell could have fired on it whatever it contained. The measured half is the one that matters here —
`packages/types/src/zod/objectql.zod.ts` **was** judged against a declared surface and carried no tell.

### ⭐ the control — why that clean reading is a real negative

A gate that returns 0 proves nothing until it is known to fire in the region under test. Same tool, same
declaration, same path, a diff that adds a key and an enum member:

```
✗ T1 packages/types/src/zod/objectql.zod.ts:97  — a new key on a Zod object schema
✗ T2 packages/types/src/zod/objectql.zod.ts:100 — a new member of a closed set
exit 4
```

It fires on this exact path. The 0 on the real diff is therefore a measurement, not a silence.

### ③ every check, not a required subset

35 check-runs on `6249be35…`: **32 `success` + 3 `skipped`, 0 non-green**. The API's reported
`total_count` is 35 and the returned array length is 35, so nothing hid behind pagination. The separate
commit-status API (which check-runs does not cover) reports `state: success`, 1 context, `Vercel`.

### charter provenance, with its own control

The governing files are byte-identical between the clone's `ad715aca57b44bbe745a20cf20f339cd5195df14`
and current `main` `f721ef0ff…` — `contract-review.md`, `landing-operations.md`, `SKILL.md`,
`dispatch-gates.mjs`, `check-clause2-carriers.mjs`, `check-widening-tells.mjs`, `check-governed-merges.mjs`
all resolve to the same blob at both revs. That empty path-scoped diff is controlled twice: the same two
commits diff to **20 files / +1412 / -25** with no path filter, and each of the seven paths resolves to a
real blob (not `MISSING`) at both revs. ⛔ I am not writing "no output above = clean" for it.

### carriers stripped

Neither carrier holds `needs:contract-review` (the PR's labels are `package: types` only), so there is
nothing to strip. Landing by **squash** through the merge queue.

---

PR #9082, opened 2026-09-11
fix(tooling): open the citation continuation scope on a bare filename

### LINKED ISSUE #9026: tooling(gates): check-new-cross-file-line-citations cannot parse the bare :NNN continuation form its charter claims — it reports 0 on a docblock whose six citations had all rotted
Found twice today by two independent seats working objectui#8990, and measured the second time. ⛔ Not a defect in any PR — a defect in the gate.

## The finding

`check-new-cross-file-line-citations` reports **`0 new citation(s)`** on a diff whose docblock carries **six cross-file line citations**, all of which had already **rotted**.

The six sat in the `ObjectKanbanSchema.groupBy` docblock pointing into `ObjectKanban.tsx` — `:601` `:613` `:625` `:640` `:747` `:865`. Measured on the merge base: the reads they name now sit **~370 lines lower**. Every one was stale, and the gate scored the diff clean.

## The mechanism, measured

`census.mjs:569-576` only parses a bare `:NNN` **within 2 lines of a numbered hit**. On that docblock it finds **0 of the six** addresses — while the gate's own charter claims to cover 「无文件名的续写地址」 (continuation addresses with no filename), which is exactly the form these six use.

⇒ **The gate is not report-only about this class. It is silent on it.** Those are different things, and only the first is a deliberate policy choice. A report-only gate that cannot parse the form it claims to watch reports a clean sheet, and a reader who knows it is report-only still reads that as "nothing new to look at".

## Why it matters here specifically

⭐ This is the gate that exists to stop line citations from rotting, and the form that rotted is the form it cannot see. The failure is self-concealing: the rot accumulates precisely where the instrument is blind, and the instrument's output says everything is fine.

⚠️ It also mislabels its own coverage. A gate whose charter names a class it does not parse is worse than a gate that says nothing, because it is cited as evidence — objectui#8990's dev correctly declined to file on the grounds that "the gate already owns this class", which is true of the charter and false of the code.

## What is owed

1. **Make the parser see the bare `:NNN` continuation form**, or **narrow the charter** so it stops claiming a class it does not cover. ⛔ Not both silently — pick one and say which.
2. Either way, a **firing control**: a fixture carrying the exact six-address shape from that docblock, asserting the gate now reports them. ⛔ A green run proves nothing here; the whole finding is that green was the wrong answer.
3. ⚠️ **Measure the backlog before changing the parser.** If it starts seeing this form, how many existing citations does it light up? That number decides whether the gate can go enforcing or must stay report-only for a while — and it is the number nobody has.

## ⚠️ Not measured

- Whether other citation forms are also unparsed (only the bare `:NNN` continuation form was measured).
- How many rotted citations exist tree-wide in this form — see item 3, which is the same gap.
- Whether the 2-line proximity window is deliberate (it may be guarding against false positives, in which case the fix is charter-side, not parser-side).
- The gate's workflow trigger conditions.

Refs objectui#8990 (where both readings were taken) · objectui#8924 (the installed-spec-pin claim rotting in ~53 places — same family: an anchor that drifts and a gate that does not catch it).

---

os-warren:
Fixes #9026

**Picked, in one sentence: MAKE THE PARSER SEE THE FORM — ⛔ not narrow the charter.** The charter is still corrected, because a charter that overstates coverage is the defect one level up, but it is corrected to describe what the code now does rather than to retreat from a class this tree demonstrably writes.

⛔ **Enforcement is untouched.** `ENFORCEMENT` is still `report-only`, and the backlog number below is the reason to leave it there.

---

## 1. The card's own address does not resolve — re-derived

`census.mjs:569-576` is a 404: there is no `scripts/census.mjs` on `main` (`ls` exit 2). Confirmed independently, and ⛔ not copied into this branch. The real code, by content:

| what | where |
| --- | --- |
| the proximity test | `scripts/cross-file-line-citation-census.mjs`, the `lastAddress` guard in `scanFile` |
| the shared parser | `scripts/check-new-cross-file-line-citations.mjs` imports `scanFile` / `judge` / `SELF_FILES` from that census module |

⇒ One parser change moves BOTH surfaces, so both are ablated below.

⭐ That the card about citation rot carried a rotted citation is the finding about itself.

## 2. What was actually broken — and it is NOT the window

Scope for a bare `:NNN` opened **only** on a full `NAME` plus address. A line naming a file *without* a number of its own — which is how the `ObjectKanbanSchema.groupBy` docblock opened — put **nothing** in scope, so all six addresses under it were, to the scanner, a colon and a number.

The census also skips lines carrying no `:` or `#` before doing any scope work, and the filename line has neither. So the trigger never fired on the one line that mattered.

## 3. ⭐ Backlog measured BEFORE the parser moved

Taken with an out-of-tree copy of the census carrying the candidate behind an env switch, so both numbers come from one instrument. **Faithfulness control: with the switch OFF the probe's row set is byte-for-byte the shipped census's** (population 1201 = 1201, row-set identity `True`) — without that, the delta would not be a delta.

| tree-wide | before | after |
| --- | --- | --- |
| cross-file citations in the population | 1201 | **1271** |
| of which FALSE against today's tree | 532 | **561** |
| resolving | 250 | 266 |
| unjudged | 419 | 444 |

**74 citations become visible; 4 are re-attributed** to the file the prose actually names (net +70). All four re-attributions are readable from their own quoted text, and two of them were scoring a verdict against the wrong file — one an unearned `resolves`, one an unearned `out-of-range`.

⚠️ **Named rather than tuned away:** roughly 8 of the 74 are not citations — bare ports in prose (`:3000`, `:5180`, `:3001`) and JSON object literals in fixture tables. They are ~0.6% of the population. A backtick-membership filter was measured and rejected: it also kills real citations written as `(:686)` and in aligned comment tables.

## 4. The firing control — ⭐ green was the wrong answer, so red is asserted first

The fixture is the **real** docblock, recovered from the commit that replaced it, ⛔ not a paraphrase.

| parser | reading on that docblock |
| --- | --- |
| shipped on `main` | `hits=0` — **0 of 6, BLIND** |
| window widened to 999999, trigger unchanged | `hits=0` — **still 0 of 6** |
| this branch | `hits=5` — 601, 625, 640, 747, 865 |

⛔ **The middle row is the one that settles the card's open question.** No window can carry a file that never entered scope, so the ±2 window is **not** the cause and **did not move**.

### Ablation on the committed code, both surfaces

Mutation (revert the trigger to full-address-only) proven on disk by blob hash `cb47a742…` → `a2e2a50a…`, injected marker present 1×, deleted text present 0×:

```
ABLATION_VITEST_EXIT=1
Test Files  2 failed (2)
Tests  4 failed | 67 passed (71)

× a bare filename opens the continuation scope … > reads the addresses that sit under the line naming the file
× a bare filename opens the continuation scope … > gives an address to the file named to its LEFT
× a bare filename opens the continuation scope … > does not let a filename in one paragraph capture a number in the next
× the differ FIRES … > reports the continuation addresses under a filename that carries no number
    AssertionError: expected [] to deeply equal [ 601, 625, 640, 747, 865 ]
```

That `[]` **is** the zero the card reported, reproduced inside the gate. Restored with `git checkout HEAD -- FILEPATH`, proven byte-identical to the HEAD blob and by an empty `git diff HEAD`; restore leg re-run green at 71 passed.

## 5. The card's four open questions, answered explicitly

1. **Is the ±2 window deliberate?** — **Yes, and it stays.** It is the guard that keeps a bare `:NNN` from being a colon and a number. Measured: widening it to 5 (the smallest width that reaches the sixth address) admits 14 more rows — 11 real citations, and a port, a cron minute and a Chinese enumeration. That is a judgement about a false-positive guard that this card was ⛔ not asked to make, so the number did not move.
2. **Is the fix charter-side?** — **No, parser-side**, and the middle row of the table above is the measurement that decides it.
3. **Are other citation forms also unparsed?** — **Not found, and the check was scoped.** The four full-address syntaxes each have a pin that this branch leaves green. ⚠️ What this branch did NOT do is sweep for a sixth spelling; the population is now 1271 and no instrument says that is complete.
4. **The gate's workflow trigger conditions** — `.github/workflows/line-citation-gate.yml` fires on `pull_request` into `main`/`develop` plus `workflow_dispatch`. **No `paths` / `paths-ignore` filter, deliberately.** No `merge_group` leg, so it is ⛔ not requirable while report-only.

⚠️ **One residual blindness, pinned by name rather than implied away.** The sixth address is restated in a second paragraph, eight lines below the only filename and with none of its own. It is **not** read. `scripts/__tests__/cross-file-line-citation-census.test.ts` asserts that blindness explicitly so no reader can cite this syntax as covering an address that drifted out of its scope.

## 6. Does this redden anything?

**No.** The gate on this branch's own diff:

```
Files compared  : 1
PASS  fires / does-not-fire-on-a-moved-citing-line / unresolvable-is-not-false / same-file-is-not-in-the-population
VERDICT new-cross-file-line-citations: 0 new citation(s), enforcement report-only -> exit 0
```

Three of the four changed files are in `SELF_FILES`; the fourth is the changeset, written with no line address of its own (verified by a `grep -P` that finds 13 such lines in the test file and 0 in the changeset).

⛔ The tree-wide backlog cannot redden other pull requests either: this gate is DIFFERENTIAL and the existing citations are explicitly not its denominator.

## Verification

| command | result |
| --- | --- |
| `vitest run scripts/__tests__/{cross-file-line-citation-census,check-new-cross-file-line-citations}.test.ts` | exit 0 — **71 passed** (was 63; +8 new) |
| `pnpm lint` (repo-wide, under the shared verify lock) | `VERDICT command-exit 0` — 47/47 tasks, **0 errors** |
| `eslint --no-inline-config` on the 3 lintable changed files | exit 0, 3 files, 0 errors — with a **firing control**: an address in a test NAME is refused by `object-ui/no-line-address-in-test-name` |
| `pnpm check:control-bytes` | exit 0, 7277 files |
| `pnpm check:entry-guard` / `check:esm-specifiers` / `check:pre-install-import-graph` | exit 0 |
| `pnpm changeset:check` | exit 0 |
| `vitest run scripts/__tests__/ci-cd-pipeline-doc.test.ts` | exit 0 — 79 passed (no command name or workflow moved) |
| `node scripts/check-governed-queue-guard.mjs --test` on all 4 paths | `NOT GOVERNED` — with `AGENTS.md` as the firing control |

## Out of scope, filed

- **#9081** — `census:cross-file-line-citations` exits 1 on `main` with "this run is NOT a reading": its `non-firing` control is pinned on a file PAIR that `packages/types/src/crud.ts` cites twice, and one of the two has rotted. ⛔ Not caused by this branch — measured on clean `main` before any edit and byte-identical after it.

---

---

### DIFF
--- a/.changeset/9026-bare-colon-continuation-scope.md
+++ b/.changeset/9026-bare-colon-continuation-scope.md
@@ -1,0 +1,36 @@
+---
+---
+
+**Chosen: make the parser see the form — ⛔ not narrow the charter.**
+`scripts/cross-file-line-citation-census.mjs` now opens the continuation scope
+on a bare FILENAME as well as on a full `NAME` plus address, so a docblock that
+writes "`ObjectKanban.tsx` reads `schema.groupBy` at" and then a run of bare
+addresses is read instead of scored empty (objectui#9026). Tooling and tests
+only; no package is released by this change.
+
+**What was actually broken.** Scope opened ONLY on a full address, so a filename
+carrying no number of its own put nothing in scope and every bare address under
+it was, to the scanner, a colon and a number. `ObjectKanbanSchema.groupBy`
+carried six of them into `plugin-kanban`'s renderer and every one had rotted;
+the census scored the docblock **empty** and `check:new-line-citations` reported
+`0 new citation(s)`. ⇒ The gate was not report-only about that class, it was
+**silent** on it, and the charter named the class as covered.
+
+**⛔ The proximity window was NOT the cause, and it did not move.** With the
+trigger absent, that docblock reads 0 of 6 at ANY window — no window can carry a
+file that never entered scope — so the window stays at its shipped value. It is
+deliberate, it is the guard that keeps a bare address from being a colon and a
+number, and widening it to reach one further address would have admitted 14 more
+rows of which three are a port, a cron minute and a Chinese enumeration.
+
+**Backlog measured BEFORE the change, so the two numbers stay separable.**
+Tree-wide population 1201 → 1271; 74 citations become visible and 4 are
+re-attributed to the file the prose actually names. Citations FALSE against
+today's tree: 532 → 561. Roughly eight of the newly visible rows are ports or
+JSON literals rather than citations, named in the pull request body rather than
+tuned away. ⛔ Enforcement is untouched (`report-only`) and this change ⛔ does
+not flip it: the number above is the reason to leave it where it is for now.
+
+Two follow-on effects, both measured: an address written after a second filename
+on the same line now goes to the file written to its LEFT rather than to
+whichever match the line ended on, and a blank line ends the scope.
--- a/scripts/__tests__/check-new-cross-file-line-citations.test.ts
+++ b/scripts/__tests__/check-new-cross-file-line-citations.test.ts
@@ -119,6 +119,49 @@ describe('the differ FIRES — the direction that makes this a gate at all', () 
     expect(added[1].citedWritten).toBe('packages/core/src/actions/ActionRunner.ts');
   });
 
+  /**
+   * objectui#9026 — THE GATE'S OWN ZERO, reproduced and then made to fire.
+   *
+   * ⭐ This is the assertion the card is about. A pull request whose docblock
+   * carried six cross-file addresses — all of them already rotted — was scored
+   * `0 new citation(s)` by this gate, because scope opened only on a full
+   * `NAME:NNN` and the docblock named its file without one. Report-only was
+   * never the problem: the gate was SILENT on the class, and a reader who knows
+   * a number is report-only still reads a clean sheet as nothing to look at.
+   *
+   * ⛔ A green run proves nothing here. The paired ablation below removes the
+   * filename and asserts the gate falls back to the zero it used to print, so
+   * the firing direction is the thing under test.
+   */
+  const KANBAN_DOCBLOCK = [
+    ' * `packages/plugin-kanban/src/ObjectKanban.tsx` reads `schema.groupBy` at',
+    ' * thirteen sites: lane materialisation (`:601`, `:625`, `:640`), card moves',
+    ' * (`:747`, `:865`) and their effect deps.',
+    '',
+  ].join('\n');
+
+  it('reports the continuation addresses under a filename that carries no number', () => {
+    const added = newCitationsIn({
+      relPath: NOTES,
+      baseText: '',
+      headText: KANBAN_DOCBLOCK,
+    });
+    expect(added.map((a) => a.citedLine)).toEqual([601, 625, 640, 747, 865]);
+    for (const a of added) {
+      expect(a.syntax).toBe('continuation');
+      expect(a.citedWritten).toBe('packages/plugin-kanban/src/ObjectKanban.tsx');
+    }
+  });
+
+  it('⛔ ABLATION — with the filename gone the gate prints the zero it used to', () => {
+    const added = newCitationsIn({
+      relPath: NOTES,
+      baseText: '',
+      headText: KANBAN_DOCBLOCK.replace('`packages/plugin-kanban/src/ObjectKanban.tsx` reads', 'the renderer reads'),
+    });
+    expect(added).toHaveLength(0);
+  });
+
   it('a RE-ADDRESSED citation is new — moving the number is not a repair', () => {
     // Clause 4 repairs an existing address by converting it to a content
     // anchor, never by moving the number to a different number. So an edited
--- a/scripts/__tests__/cross-file-line-citation-census.test.ts
+++ b/scripts/__tests__/cross-file-line-citation-census.test.ts
@@ -47,6 +47,7 @@ import {
   CONTROLS,
   FALSE_VERDICTS,
   MAX_ANCHOR_LINES,
+  CONT_WINDOW,
 } from '../cross-file-line-citation-census.mjs';
 
 const REPO_ROOT = join(__dirname, '..', '..');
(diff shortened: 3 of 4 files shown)


os-warren:
**Ruling on the open question: option A. Ship as-is — the fixture asserts 5 of 6 and pins the sixth as blind BY NAME.**

The dispatch asked for a fixture "asserting the gate now reports **them**" — all six. This PR reports five. I am confirming that as correct and recording why, because the deviation is from my instruction and the dev was right to flag it rather than quietly meet the letter of it.

### Why A and not B

The card's item 1 is explicit: *"Make the parser see the bare `:NNN` continuation form, **or** narrow the charter … ⛔ Not both silently — pick one and say which."* This PR picked the parser and said so. Widening the proximity window is a **second, independent change to a false-positive guard**, and the card reserved that guard by name in its own ⛔ Not-measured list: *"whether the 2-line proximity window is deliberate (it may be guarding against false positives, in which case the fix is charter-side, not parser-side)."*

⭐ **The measurement that settles it, and it is the strongest reading in this PR:** with the trigger absent and the window widened to **999999**, the docblock still reads **0 of 6**. ⇒ the window was measurably **not the cause** of the blindness, so leaving it untouched is not an omission — it is the correct scope. A change that had moved the window would have confounded the two.

### Why the residual is not a hole

The card's actual complaint is not "the gate misses one address." It is that *"a report-only gate that cannot parse the form it claims to watch reports a clean sheet"* — the failure is **self-concealing**. Pinning the sixth address as blind **by name**, in the test, converts that silence into a declaration. ⇒ the defect class is closed even though one address stays out of scope, and ⛔ it can no longer be mistaken for coverage, which was the whole finding.

The control does what the card demanded: **0 → 5**, with an ablation on the committed code returning it to 0 (`AssertionError: expected [] to deeply equal [ 601, 625, 640, 747, 865 ]`, restored byte-identical). Green was replaced by red.

### B is not dropped, it is measured and routed

Widening to a window of 5 costs **+14 rows** tree-wide — 11 real citations against 1 port, 1 cron minute and 1 Chinese enumeration. That is a defensible ratio and a genuine decision, so it goes on its own card rather than into this PR. ⛔ It is not a condition on landing this one.

### Also on the record

- ⭐ Enforcement stays **report-only** and the gate reports 0 new citations on this PR's own diff. That was the way this PR could have gone badly wrong — a parser that starts seeing a form while the gate is enforcing reddens every open PR — and it did not.
- The backlog was measured **before** the parser moved (1201 → 1271, false 532 → 561), with a faithfulness control showing the candidate switched off reproduces the shipped census row-for-row. That ordering was the card's item 3 and it was honoured.
- ~8 of the 74 newly visible rows are ports and JSON literals rather than citations. Naming them in the PR body instead of hand-tuning them away is the right call: every structural filter measured to remove them also removed **real** citations.

⛔ Nothing further is owed from the dev on this PR. Landing waits only on terminal CI.

---

os-warren:
## Landing provenance — PR #9082 (card objectui#9026)

Head: **`5b71f126bbe0954808327733d59a636041f1edeb`**. Gate checkout: fresh `objectstack` clone `os-fresh2`, `origin/main` = `ef180302f55955baf9994a91d0793bbd3926b068`.

### ① contract review — NOT OWED

`Clause-②: no`, and both carriers are absent and agreeing (card objectui#9026 carries no `needs:contract-review`; this PR's labels are `tests` only). ⛔ No PASS record is cited because none exists and none is owed.

### ② the enqueue gates — including one that FAILED first

| gate | exact command | exit |
|---|---|---|
| carriers + C5 | `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9082` | **4 → 0** |
| widening tells | `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-widening-tells.mjs --declaration no --diff <PR diff>` | 0, **NOT MEASURED** — see below |
| governed surface | `node scripts/pm/check-governed-merges.mjs --pr objectstack-ai/objectui#9082` | **0** — 0 of 4 paths hit the register |

⭐ **The carriers gate failed first, at exit 4, and the defect was the seat's — recorded rather than quietly fixed.** Its finding:

> the `Clause-②` declaration is MISPLACED: the fixed spelling appears on the thread … but NOT in the card's claim comment, which is the carrier the enqueue gate's content limb reads.

`check-clause2-carriers.mjs:751-752` defines a claim comment as one whose body carries **a LINE BEGINNING `Claim:`**. I had been opening claims with `**Claim:**`, so the line begins `**` and **no comment on the card was a claim comment at all** by the predicate. Measured across tonight's claims: objectui#8735 and objectui#8784 used the bare spelling and passed `--pair` at exit 0; objectui#9026, objectui#9034, objectui#8832, objectui#8924 and objectui#8811 used `**Claim:**` and were illegible. On objectui#9034 I had also bolded the value — `Clause-②: **yes**` — defeating even a loose match.

⇒ Four live cards were repaired with format-only claim comments (same session, same branch, same declaration; ⛔ not re-claims), each verified against the gate's own predicate before re-running. The re-run returns **exit 0**: *"the clause-② declaration is readable in the fixed spelling and both carriers agree."*

⛔ **Markdown emphasis on a field a machine reads is not formatting, it is corruption** — and the failure was self-concealing in the same shape this very card is about: the gate reported not "this is wrong" but "there is nothing here to read".

### ⚠️ What the widening-tells gate does NOT say here

Exit 0, and ⛔ **it is not a clean reading.** The tool says so itself, and I am quoting it rather than paraphrasing:

> ⛔ NOTHING on this diff was examined for widening tells, so this exit 0 is evidence about no surface at all.

All **4 of 4** paths are NOT MEASURED — no declared surface covers `.changeset/9026-bare-colon-continuation-scope.md`, `scripts/cross-file-line-citation-census.mjs`, or either file under `scripts/__tests__/`.

⚠️ My firing control **fired** (T1 `:12`, T2 `:15`, exit 4) but on `packages/types/src/zod/layout.zod.ts` — **a path this PR does not touch**. ⇒ it establishes the tool works; it does ⛔ **not** establish that a tell could have fired on this PR's paths, because no tell can fire there at all.

⇒ The support for "this diff widens nothing" is therefore the **governed-surface predicate** (0 of 4 paths, derived from `GOVERNED_SURFACES` on the final file list) and the file list itself — ⛔ not the tells.

### ③ every check, not a required subset

34 check-runs on `5b71f126…`: **31 `success` + 3 `skipped`, 0 non-green**, all `completed`. Reported `total_count` 34 == returned array length 34, so nothing hid behind pagination. ⚠️ At 00:09Z this same head had **9 checks still `in_progress`** — that state was read as not-terminal and landing waited, because ⛔ an `in_progress` check is not a green one.

`mergeable_state` moved `behind` → `clean`. ⛔ `behind` was never treated as a blocker: the merge queue rebuilds each PR on current `main` itself, and hand-merging the base would move the head.

### charter provenance, with its own controls

The seven governing files are byte-identical between the clone's `ad715aca57…` and current `main` `ef180302f5…`. Controls: the same two commits diff to **36 files / +3466 / -147** with no path filter, and each of the seven paths resolves to a real blob (not `MISSING`) at both revs. ⛔ No "no output above = clean" echo is written for that empty path-scoped diff.

### carriers

Neither carrier holds `needs:contract-review`, so there is nothing to strip. Landing by **squash** through the merge queue.

---

PR #9089, opened 2026-09-11
test(types): anchor the objectui#8466 calendar pin file to behaviour, not formatting

### LINKED ISSUE #8832: finding(types): the #8466 calendar pin file anchors three assertions to off-disk source text and line-wrapping, and one `it` is vacuous at runtime
Filed by the `domain:spec` @ objectui PM seat, session `session_01Jmxdo7bmeqCQHLSfmLVX9w`, as **finding F4** of the contract review on PR #8807 (`5600642940`, director seat at `CONTRACT_REVIEW_TIER`). Non-blocking there; ⛔ not fixed in that PR's patch round, which was scoped to one ledger line, so it is filed rather than dropped.

⛔ No `domain:*` or `priority:*` applied — routing and grading are triage's.

## The file

`packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts`, landing with PR #8807.

⭐ **Most of this file is good, and the card should not read as an attack on it.** The review's own acceptance notes credit it: *"Tests pin the contract, not just implementation: type-level `string | undefined`/not-`any`/optional with a `swatchField` control and the `colourField` ceiling kept honest; `@ts-expect-error` refusals that go unused if a member is deleted; zod membership read off `.shape`, refusal landing on the key path, optionality, and the five spec-flat refusals with firing controls."* Four assertions out of nineteen are the subject here.

## The four, as the review recorded them

| | site | what it anchors to | why that is brittle |
|---|---|---|---|
| 1 | `:170` | the literal `(schema as any).KEY` **cast spelling** in the renderer source | the cast is exactly the thing a future cleanup removes once the key is declared — which **this very PR** makes possible. The pin goes red on the improvement it enables. |
| 2 | `:180` | a regex over a `useMemo(...)` **dependency list** | a formatter run, a reordered dep, or a prettier width change reddens it without any behaviour moving |
| 3 | `:194` | the README's **line wrap** — `'at your own\nfields when they differ.'` | a re-wrap of a prose paragraph reddens a types-package test |
| 4 | `:262-275` | a sibling `it` that **asserts a literal it wrote in the same file** | vacuous at runtime — it cannot fail for the reason it names |

⚠️ **Item 4 is a naming problem, not a missing pin.** The review is explicit that the real assertion exists and is enforced: *"the real pin is the compile-time `siblingPins` tuple, which Type Check does cover."* ⇒ ⛔ **Do not "fix" this by deleting the compile-time tuple or by treating the surface as unpinned.** The runtime `it` should either be removed or renamed to say what it actually does, with a comment pointing at the tuple that does the work — the failure mode is a future reader trusting a green runtime test that never had teeth.

## Why this is worth a card and not a shrug

Items 1–3 all anchor a `packages/types` test to **text on disk in another package** — renderer source, a dependency list, a README's wrapping. That is the same class as objectui#8478 (`.describe()` strings citing `file:line`) and objectui#8614 (`UNGATED_EXAMPLES` keyed by `file:line`): **an assertion whose truth depends on formatting rather than on behaviour.** Today, three different instances of that class reddened or misled inside one shift:

- objectui#8614's ledger reddened PR #8807 on a one-line import insertion;
- the same ledger reddened PR #8828 on a 27-line insertion, three rows at once;
- objectui#8478's own census found **three** already-rotted addresses, two of them off by exactly `+3` from one shared edit.

⇒ The pattern is established well enough that adding a fourth instance deliberately, in a brand-new file, is worth naming while the file is fresh.

⭐ **Item 1 is the sharpest**, and it is worth stating on its own: the pin asserts that the renderer still reads the key **through an `as any` cast**. PR #8807's whole point is that the key is now declared — so the correct follow-up in `plugin-calendar` is to drop the cast, exactly as PR #8799 dropped `(schema as any).navigation` once `navigation` was declared. **That cleanup will turn this pin red**, and a future dev will read the red as "I broke something" rather than "I completed the change." The pin punishes the improvement it exists to enable.

## Suggested direction — ⛔ not a ruling, and each item is its own call

The review's own suggestion: *"whitespace-insensitive matches / reuse the `rendererReads` census; note the sibling pin is compile-time."*

- **1** — assert the renderer **reads the key**, not *how*. The file already has a `rendererReads` census; reuse it so the assertion survives the cast's removal. (Or pin the opposite: that the cast is **gone**, once it is.)
- **2, 3** — normalise whitespace before matching, so a re-wrap or a reformat cannot move the verdict.
- **4** — remove or rename the runtime `it`, and say in a comment that the compile-time `siblingPins` tuple is the real pin.

⚠️ Whoever takes this should check whether PR #8807 has landed first — the file does not exist on `main` until it does.

Refs: PR #8807 / objectui#8466 (the file's home) · objectui#8614 · objectui#8478 · objectui#8830 (F3) · objectui#8831 (F2).

os-warren:
Fixes #8832

The `objectui#8466` calendar pin file held four assertions whose truth depended on how text happened to be **written** rather than on what the code **does**. Each is re-anchored to the fact it names, and each new verdict is shown going red against a mutation of that fact.

⛔ The file is **not** rewritten. 4 of its 19 tests are touched; the compile-time `siblingPins` tuple is untouched and is still the real pin for the sibling interface; everything else stands. Test count is 19 before and 19 after.

## The four sites — addresses re-derived on `main`, not taken from the card

Every address in the card had drifted. These are my own readings of `packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts` at `main` = `16fc4cf20` (422 lines, 15 `it(` blocks, 19 tests):

| site | on `main` | was anchored to | now anchored to |
|---|---|---|---|
| 1 | `:218` | the literal `(schema as any).KEY` **cast spelling** in the renderer source | a schema-read census over `getCalendarConfig`'s **own body**, sliced out by brace matching. It sees the bare `schema.KEY` and the cast form alike, so the verdict is *the renderer reads this key*, not *it reads it through a cast*. |
| 2 | `:230` | a regex over the `useMemo(...)` **dependency list's text** | the same census run over that dependency list, **sliced out by bracket matching** from a whitespace-tolerant locator. Order, wrapping and print width cannot reach the result; only the dependency itself can. |
| 3 | `:244` | the README's **line wrap**, `'at your own' + newline + 'fields when they differ.'` | the markdown **block** that teaches all five, with the block's internal whitespace collapsed. A blank line between blocks is structure; where lines break inside one is formatting. |
| 4 | the `it` at `:315`, assertions `:327` and `:328` | a record literal it had written three lines earlier, plus a tuple counted against a hard-coded `5` | the `it` is renamed to say it keeps the compile-time tuple referenced, and its one runtime assertion now checks that tuple's **arity against `FLAT_KEYS`** — a sixth flat key with no sixth slot goes red, which a shorter tuple type-checking silently would not. |

⚠️ Site 1 was the sharpest: dropping the `as any` cast is the cleanup that declaring these keys makes possible, so the pin went red on the improvement it existed to enable.

⚠️ The card describes site 4 as sitting at `:262-275`. On `main` that range is a substantive `safeParse` refusal test with two firing controls, not a vacuous assertion — the vacuous one is the `it` at `:315`. Confirmed independently; see the report on the card. One refinement: that `it` carried **two** vacuous runtime assertions, `:327` (the record literal the card names) and `:328` (`toHaveLength(5)`), and both are addressed.

The compile-time tuple stays at `:147` on `main` (`:167` on this branch) and is not deleted, weakened, or moved.

## Firing controls — one reading per site

Every mutation was applied to disk, proved landed by a blob-hash difference plus a single-line `grep -c`, run, then restored by `git checkout HEAD -- path` with the restore proved by `git diff HEAD` being **0 bytes** (never by an exit code). Final `git status --porcelain` was 0 bytes. The mutated artefacts are read by `readFileSync` at test time, so no build step sits between the mutation and the verdict.

| site | mutation of the behaviour it now names | reading |
|---|---|---|
| 1 | `getCalendarConfig` stops reading `colorField` (`colorField: (schema as any).colorField,` becomes `colorField: undefined,`); the memo dependency list left intact | **RED** — `vitest exit=1`, `1 failed / 18 passed`, on `getCalendarConfig reads all five flat keys off the node`: *"getCalendarConfig no longer reads the flat colorField"*. Site 2 stayed green, so the two are independent. |
| 2 | `allDayField` is removed from the config memo's dependency list; the `getCalendarConfig` read left intact (probed: dep line 1 to 0, function read still 1) | **RED** — `vitest exit=1`, `1 failed / 18 passed`, on `...and allDayField is LOAD-BEARING`: *"allDayField left the config memo's dependency list"*. Site 1 stayed green. |
| 3 | `colorField` is moved out of the teaching paragraph into its own block — still present in the README (backticked count 2), so the per-key loop stays satisfied | **RED** — `vitest exit=1`, `1 failed / 18 passed`, on `the README still teaches all five together`: *"the five flat keys are no longer taught in one block: expected 0 to be greater than 0"*. |
| 4 | the `colorField` slot is deleted from the `siblingPins` tuple (type entry and value) | **RED** — `vitest exit=1`, `1 failed / 18 passed`: *"a flat key has no `siblingPins` slot, so the sibling interface is unpinned for it: expected [ true, true, true, true ] to have a length of 5 but got 4"*. |

## The other direction — each re-anchor survives the change that used to break it

A re-anchor that can no longer go red is a deleted test; a re-anchor that still reddens on formatting is not a fix. Both halves were measured.

| site | formatting change applied to disk | old assertion | new assertion |
|---|---|---|---|
| 1 | the future cleanup: all **23** `(schema as any).` casts dropped to `schema.` | **would have failed** — re-evaluated against the cleaned file, `toContain('(schema as any).KEY')` is `False` for all five | **GREEN**, `vitest exit=0`, 19 passed |
| 2 | the memo call re-wrapped prettier-style across three lines and its 8 deps reversed | **would have failed** — the old regex re-evaluated against the reformatted file returns `False` | **GREEN**, `vitest exit=0`, 19 passed |
| 3 | the teaching paragraph re-wrapped onto one line | **would have failed** — `toContain('at your own' + newline + 'fields when they differ.')` is `False` | **GREEN**, `vitest exit=0`, 19 passed |

## Controls carried by the new assertions themselves

A control is only a control if it is known to fire in the region being tested, so each new verdict carries one from inside that region:

- **site 1** — `calendar` is read inside `getCalendarConfig` and is not one of the five (the slice is not empty by accident); `objectName` is read all over the same *file* and never inside that function (a slice that had swallowed the file fails here).
- **site 2** — `calendar` is in that very dependency list; `objectName` and `filter` are read elsewhere in the same file, `objectName` in the sibling `dataConfig` memo, and are deliberately out of this one.
- **site 3** — the same block search over the same README returns **0** blocks once `swatchField`, which the README does not teach, joins the key set.
- **the census instrument itself** — the existing reads-census `it` now asserts **one control per read form**: `schema.objectName` is read bare and `(schema as any).defaultView` through a cast. A single bare-form control would have left the cast form unmeasured, which is the half site 1 depends on.

## Verification

Head sha `0497a5f3a`. Commands and the exit code each printed, captured before any pipe.

- `pnpm exec vitest run packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts` — exit **0**, `Test Files 1 passed (1)`, `Tests 19 passed (19)`. Same counts before the change.
- `pnpm exec vitest run packages/types/` — exit **0**, `Test Files 173 passed (173)`, `Tests 3404 passed (3404)`.
- `pnpm --filter @object-ui/types type-check` — exit **0** (`tsc --noEmit && tsc -p tsconfig.examples.json && tsc -p tsconfig.test.json`). Proof this covers the edited file rather than excluding it: `tsc -p tsconfig.test.json --noEmit --listFiles` lists 639 files, of which **1** is `calendar-flat-color-allday-8466.test.ts`.
- `pnpm exec eslint .` in `packages/types` — exit **0**, **243** files, 0 errors, 278 warnings; the edited file contributes exactly **2** `@typescript-eslint/no-explicit-any` warnings, which is the count its own docblock claims, so that claim is not stale. No type-aware linting is configured in `eslint.config.js` (no `project` / `projectService` / `parserOptions`), so this diff cannot move any untouched file's verdict.
- Gates: `check:new-line-citations` exit **0** (`0 new citation(s)`), `check:control-bytes` exit **0**, `check:comment-mask-corpus` exit **0**, `check:unreferenced-sources` exit **0**, `check-changeset-presence` exit **0**, `check-changeset-no-major` exit **0**, `check-changeset-fixed` exit **0**, `check-changeset-overwrite` exit **0**.
- `node scripts/check-governed-queue-guard.mjs --test` on both changed paths: `NOT GOVERNED`. The PR stays draft regardless — landing is the seat's.
- Self-scan for control bytes over both changed files: `grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'` exit **1** (no match).

Changeset: `.changeset/calendar-flat-key-pins-anchor-to-behaviour.md` with **empty frontmatter** — the gate's explicit declared-as-releasing-nothing path for a change confined to `src/__tests__/`.

## 验收备注

Out of scope for this card, noted and **not** filed — none is a reproducible bug, a declared-contract violation, or a metadata trap, and the card scoped itself to four assertions out of nineteen:

- Three further off-disk assertions in the same file are still literal-text anchors of the same family: the registration pins at `:311`/`:312` on `main` assert the exact call spelling `ComponentRegistry.register('object-calendar', ObjectCalendarRenderer`, and `:227` is a bare `toContain('allDayField')` over the whole renderer. They are not among the four the card names, and the card's opening credits the rest of the file, so they are left alone. Carrier: the next PR that touches this file, or a follow-up to this card if the seat wants the family closed rather than the four instances.
- `packages/plugin-calendar/src/ObjectCalendar.tsx` reads all five flat keys through `(schema as any)` even though both are now declared. Dropping those casts is the cleanup this card's pins were blocking; it is a `plugin-calendar` change, not a `packages/types` one. This PR measures that the cleanup now lands green (23 casts dropped, 19 tests still passing), but does not perform it. Carrier: whoever does the `plugin-calendar` cast cleanup.

Authored by the dev seat on session `session_01Jmxdo7bmeqCQHLSfmLVX9w` (https://claude.ai/code/session_01Jmxdo7bmeqCQHLSfmLVX9w), dispatched for objectui#8832.

---

### DIFF
--- a/.changeset/calendar-flat-key-pins-anchor-to-behaviour.md
+++ b/.changeset/calendar-flat-key-pins-anchor-to-behaviour.md
@@ -1,0 +1,11 @@
+---
+---
+
+Test-only change: the `objectui#8466` calendar flat-key pin file now anchors its
+off-disk assertions to behaviour instead of to formatting — the renderer reads a
+key (either spelling, so dropping the `as any` cast stays green), a key is in the
+config memo's dependency list (sliced structurally, not regexed), and the README
+teaches all five in one markdown block (wrap-insensitive). The vacuous runtime
+`it` is renamed to say it keeps the compile-time `siblingPins` tuple referenced,
+and now checks that tuple's arity against the flat-key list. No published
+behaviour changes.
--- a/packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts
+++ b/packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts
@@ -73,6 +73,24 @@
  * back to the index signature reads as `any` and therefore as a failure. And
  * every claim carries a CONTROL asserted to hold the opposite verdict, so no
  * assertion can pass vacuously.
+ *
+ * ## What an off-disk assertion may anchor to (objectui#8832)
+ *
+ * Three of the off-disk assertions below used to anchor to FORMATTING rather
+ * than to behaviour: the renderer's `(schema as any).KEY` cast SPELLING, a regex
+ * over a `useMemo` dependency list's TEXT, and the README's LINE WRAP. A fourth
+ * `it` asserted a record literal it had written three lines earlier, which
+ * cannot fail at all. The cast pin was the sharpest: dropping that cast is the
+ * cleanup declaring these keys makes possible, so the pin reddened on the
+ * improvement it existed to enable.
+ *
+ * The rule they now follow: read the FACT off disk — this key is read off the
+ * node; this key is in that memo's dependency list; these five are taught in one
+ * block — never the shape the fact happens to be written in. Casts, whitespace,
+ * dep ordering and fill width are all formatting, and each helper below is built
+ * so none of them can move a verdict. Each carries a control known to fire in
+ * the same region, because a re-anchor that can no longer go red has not fixed
+ * the assertion, it has deleted it.
  */
 import { describe, it, expect } from 'vitest';
 import { readFileSync } from 'node:fs';
@@ -106,8 +124,10 @@ const NEWLY_DECLARED = ['colorField', 'allDayField'] as const;
  * pinned rather than claimed away.
  */
 const CONTROL_KEY = 'swatchField';
-/** A declared-and-read control for the off-disk read census. */
+/** A declared-and-read control for the off-disk read census, read BARE. */
 const READ_CONTROL_KEY = 'objectName';
+/** The same, read through a CAST — the census must see both forms, not one. */
+const CAST_READ_CONTROL_KEY = 'defaultView';
 /** The misspelling the ceiling still admits. Pinned, not fixed here. */
 const MISSPELLING = 'colourField';
 
@@ -175,9 +195,79 @@ function readRepo(rel: string): string {
   return readFileSync(join(REPO_ROOT, rel), 'utf8');
 }
 
-/** Every `schema.KEY` read in a renderer, off disk. */
+/**
+ * Every key read off the schema node, in EITHER form: the bare `schema.KEY` and
+ * the cast `(schema as any).KEY`. Both spell the same FACT — the renderer reads
+ * this key — and only one of them spells it through a cast that a future cleanup
+ * deletes once the key is declared. Reading the fact rather than the spelling is
+ * what stops the verdicts below reddening on that cleanup (objectui#8832).
+ */
+const SCHEMA_READ = /(?:\bschema\b|\(\s*schema\s+as\s+[^)]*\))\s*\.\s*([A-Za-z_$][\w$]*)/g;
+
+function schemaReads(source: string): Set<string> {
+  return new Set([...source.matchAll(SCHEMA_READ)].map((m) => m[1]));
+}
+
+/** Every key a renderer reads off the node, off disk. */
 function rendererReads(rel: string): Set<string> {
-  return new Set([...readRepo(rel).matchAll(/\bschema\.([A-Za-z_$][\w$]*)/g)].map((m) => m[1]));
+  return schemaReads(readRepo(rel));
+}
+
+/**
+ * Everything `getCalendarConfig` ITSELF reads off the node, sliced out of the
+ * renderer by brace matching so the verdict is about that function and not about
+ * the file that happens to contain it. Throws rather than returning empty when
+ * the function is gone, so a vanished anchor reads as a failure, never a pass.
+ */
+function configReaderReads(source: string): Set<string> {
+  const open = /function\s+getCalendarConfig\s*\([^)]*\)\s*:\s*[^{]*\{/.exec(source);
+  if (!open) {
+    throw new Error(`${CALENDAR_READER}: \`getCalendarConfig\` is gone; the reads it records moved somewhere else`);
+  }
+  const from = open.index + open[0].length;
+  let depth = 1;
+  let i = from;
+  for (; i < source.length && depth > 0; i += 1) {
+    if (source[i] === '{') depth += 1;
+    else if (source[i] === '}') depth -= 1;
+  }
+  return schemaReads(source.slice(from, i - 1));
+}
+
+/**
+ * The dependency list of the memo that recomputes the calendar config, sliced
+ * out by bracket matching and then read with the same census. The locator
+ * tolerates any whitespace and the census ignores order, so a formatter run, a
+ * reordered dep or a changed print width cannot move the verdict — only the memo
+ * ceasing to depend on a key can. Throws rather than returning empty when the
+ * memo is gone, so a vanished anchor reads as a failure and never as a pass.
+ */
+function configMemoDeps(source: string): Set<string> {
+  const open = /useMemo\s*\(\s*\(\s*\)\s*=>\s*getCalendarConfig\s*\(\s*schema\s*\)\s*,\s*\[/.exec(source);
+  if (!open) {
+    throw new Error(`${CALENDAR_READER}: the calendar config is no longer memoised on the node's keys`);
+  }
+  const from = open.index + open[0].length;
+  let depth = 1;
+  let i = from;
+  for (; i < source.length && depth > 0; i += 1) {
+    if (source[i] === '[') depth += 1;
+    else if (source[i] === ']') depth -= 1;
+  }
+  return schemaReads(source.slice(from, i - 1));
+}
+
+/**
+ * The markdown blocks that teach every one of `keys`, each block's internal
+ * whitespace collapsed first. A blank line between blocks is STRUCTURE; where
+ * the lines break inside one is formatting — which is why pinning the literal
+ * wrap `'at your own\nfields when they differ.'` reddened on a re-wrap.
+ */
+function blocksTeachingAll(markdown: string, keys: readonly string[]): string[] {
+  return markdown
+    .split(/\n\s*\n/)
+    .map((block) => block.replace(/\s+/g, ' ').trim())
+    .filter((block) => keys.every((key) => block.includes(`\`${key}\``)));
 }
 
 function shapeKeys(schema: unknown): string[] {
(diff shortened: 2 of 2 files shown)


os-warren:
## Landing provenance — PR #9089 (card objectui#8832)

Head **`0497a5f3a1d100129f3451a5d5f9b227869dbae3`**. Gate checkout: `objectstack` clone `os-fresh2`, `origin/main` = `abc4b83ce8b0…`.

**① contract review — NOT OWED.** `Clause-②: no`, both carriers absent and agreeing (card objectui#8832 and this PR each carry no `needs:contract-review`). ⛔ No PASS is cited because none exists and none is owed.

**② gates**

| gate | exit |
|---|---|
| `check-clause2-carriers.mjs --pair 9089` (`PM_SWEEP_REPO=objectstack-ai/objectui`) | **0** — "readable in the fixed spelling and both carriers agree" |
| `check-governed-merges.mjs --pr objectstack-ai/objectui#9089` | **0** — 0 of 2 paths hit the register |
| `check-widening-tells.mjs --declaration no --diff <PR diff>` | 0, **NOT MEASURED** ⇒ see below |

⚠️ **The widening-tells exit 0 is not a clean reading, and the tool says so:** *"⛔ NOTHING on this diff was examined for widening tells, so this exit 0 is evidence about no surface at all."* **2 of 2** paths are NOT MEASURED — no declared surface covers `.changeset/calendar-flat-key-pins-anchor-to-behaviour.md` or `packages/types/src/__tests__/calendar-flat-color-allday-8466.test.ts`.

My firing control **fired** (T1/T2, exit 4) but on `packages/types/src/zod/layout.zod.ts` — **a path this PR does not touch**. ⇒ it shows the tool works; it does ⛔ **not** show a tell could have fired here. The support for "this widens nothing" is the **governed-surface predicate** and the file list, ⛔ not the tells.

**③ checks.** 35 check-runs on `0497a5f3…`, all `completed`, **0 non-green**; reported `total_count` 35 == returned array length 35. ⛔ At 00:32Z this head had 5 checks `in_progress` and landing waited — an `in_progress` check is not a green one.

**charter provenance.** The seven governing files are byte-identical between the clone's `ad715aca57…` and `main` `abc4b83ce8…`: **7 / 7** present-and-identical, against an unfiltered control diff of **38 files / +3508 / -148**.

**carriers.** Neither holds `needs:contract-review` — nothing to strip. Landing by **squash** through the merge queue.

⚠️ Filed from this card's round but ⛔ **not** acted on: three further literal-text anchors survive in the same file (`:311`/`:312` assert an exact `ComponentRegistry.register(` spelling; `:227` is a bare `toContain('allDayField')`), and `plugin-calendar/src/ObjectCalendar.tsx` still reads all five flat keys through `(schema as any)`. This PR **measures** that the cast cleanup would now land green and ⛔ does not perform it.

---

PR #9090, opened 2026-09-11
fix(plugin-detail): apply related-list redaction to auto-derived columns

### LINKED ISSUE #9053: record:related_list: redacting every authored column defeats redaction — the empty array falls through to auto-derived columns no redact list touches
## The measurement

Found while measuring the blast radius of objectui#8793 (which closes the block's
own fail-open branch). This is a SECOND, independent hole on the same path, and
it is reachable **today**, with entirely spec-canonical authoring — objectui#8793
does not cause it and does not fix it.

`record:related_list` filters its authored `columns` against the allow-list it
builds from `enforceFieldSecurity` / `redactFields`. When that filter removes
**every** member, it hands `RelatedList` an EMPTY array — and `RelatedList` reads
an empty array as "no columns were authored":

```ts
// packages/plugin-detail/src/RelatedList.tsx, effectiveColumns
if (columns && columns.length > 0) {
  const normalized = columns.map(normalizeColumn);
  return pruneEmpty(filterFLS(filterFK(normalized)));
}
// ...falls through to the highlightFields / heuristic auto-derivation
```

The auto-derived path runs `filterFK`, `pruneEmpty` and `filterFLS` — but the
block's `redactFields` list is not in scope there at all. It is a block-level
concept that never reaches `RelatedList`. So the redacted field comes back.

## Reproduction (measured, real table, real cells)

Block schema: `columns: [{ field: 'salary', label: 'Salary' }]`,
`redactFields: ['salary']`; child object schema declares `subject` and `salary`;
one row `{ subject: 'Fix the pump', salary: '90000' }`.

Rendered body cells, read from the DOM:

```
[ 'Fix the pump', '90000' ]
```

`90000` is the value the author redacted. Redacting the ONLY authored column
turns redaction into a no-op — worse than a no-op, because the fallback may
surface fields the author never listed.

## Why it is not folded into objectui#8793

That card names one branch — an entry whose identity does not resolve was KEPT —
and its repair strictly narrows the authored array. This defect lives one layer
down, in what `RelatedList` does with an empty array, and it fires on authoring
that objectui#8793 never touches (`{ field: 'salary' }` resolves fine). Widening
that PR to cover it would have made the fail-closed change and a fallback-policy
change one diff.

⚠️ The two do interact, and that is worth stating: after objectui#8793 an
all-unresolvable authored column set also empties the array, so it reaches this
fallback too. Measured, same harness, same result — `[ 'Fix the pump', '90000' ]`.

## Shape of the fix (not a ruling)

The question is what an emptied-by-security column set MEANS. Two readings, and
they differ for users:

- A: an empty result after filtering is "the author authored columns and none of
  them survived" — distinct from "no columns authored", so the list should render
  its empty/denied state rather than deriving a replacement set.
- B: the derivation is legitimate, and `redactFields` (plus whatever else the
  block decides) must be pushed down so the derived set is filtered by the same
  policy as the authored one.

B is the shape the block's own comment already assumes (FLS is pushed down, via
`perms.checkField`); redaction simply never was. A is the smaller change and the
one that keeps the two vocabularies apart.

Filed by the dev seat working objectui#8793; scope fence held, no diff here.

---

os-steve:
Fixes #9053

## What was wrong

`record:related_list` filtered its authored `columns` against the allow-list it
builds from `enforceFieldSecurity` / `redactFields`, then handed the survivors to
`RelatedList`. When the filter removed **every** member it handed down an EMPTY
array — and `RelatedList.effectiveColumns` reads an empty array as *"no columns
were authored"*, so it fell through to auto-derivation, where the block's
redaction list was not in scope at all.

⇒ **Redacting more produced less redaction.** Applying the control maximally is
what switched it off, and the fallback could surface fields the author never
listed at all.

Reproduced on the unfixed tree, real DOM body cells, the card's fixture
(one authored column `salary`, `redactFields: ['salary']`, one row):

```
AssertionError: expected [ 'Fix the pump', '90000' ] to not include '90000'
```

## What changed

`RelatedList` now takes the list as a `redactFields` prop and asks it on **every
path that decides columns** — the authored array, the `highlightFields`
prominence set, and the heuristic field walk — so one policy filters all three.
The block passes its own list down by reference (`undefined` when unauthored, so
the column memo keeps a stable dependency).

Two boundaries were held deliberately, and both are stated in the code:

- **An authored array emptied by redaction falls through to derivation**, exactly
  as it already does when the block empties it upstream. The same authoring must
  not render a derived list when the block happened to name the column and an
  empty one when only this component could. ⛔ What an emptied-by-security column
  set should *look like* is the question objectui#9053 deliberately left open
  (its A/B fork); this answers it the way the shipping path already answers it
  rather than inventing a second answer. Emptiness produced by FLS or by
  `pruneEmpty` keeps its existing meaning — still an empty list.
- **The filter is fail-OPEN on a column whose identity does not resolve**, like
  the `filterFLS` it sits beside. Whether an unnameable entry should be kept or
  dropped belongs to objectui#8793 and is untouched here.

## The measurement the grade rests on, re-derived rather than inherited

Triage graded this p2 (not p1) on one sentence: the derived path *"still runs
`filterFK`, `pruneEmpty` and `filterFLS`."* Verified against the code before any
edit, and it holds — with one correction to how:

- the `highlightFields` branch calls `filterFLS` literally;
- the heuristic walk does not call it by name — it inlines
  `perms.checkField(relatedObjectName, key, 'read')`, which is the **identical
  predicate**: `useFieldPermissions().readableFields` is literally
  `fields.filter((f) => checkField(objectName, f, 'read'))`.

⇒ Field security is re-applied on the derived path. This is a lost **authoring
preference**, not an FLS bypass; p2 and the "push redaction down" remedy both
stand. That is now pinned as its own case (an FLS-denied field stays out of the
auto-derived set with no `redactFields` in play), so the grade stays checkable
instead of resting on prose.

## Tests

New pin `packages/plugin-detail/src/__tests__/RecordRelatedListRenderer.redactedDerivation-9053.test.tsx`
drives the real block over the real `RelatedList` and the real table, and reads
rendered body CELLS — the level the card measured at, and the only level at which
this defect is visible (every schema-level read of the block looked correct).

| case | reads |
|---|---|
| MAXIMAL — the only authored column is redacted | the card's own fixture; the value is gone, a non-redacted column still renders (non-vacuity) |
| CONTROL, partial — some authored columns redacted | authored path taken, untouched by this change |
| CONTROL, no redaction | the same column IS on screen — the positive control every negative above leans on |
| DERIVED, heuristic walk — no authored columns at all | the walk is the only path, and redaction now reaches it |
| DERIVED, `highlightFields` — the prominence set is redacted | falls through to the walk, which is filtered too |
| CONTROL, `highlightFields`, no redaction | the prominence path still leads with it |
| `RelatedList` directly, authored and derived | the prop means the same thing to a direct consumer on both paths |
| FLS on the derived path, no `redactFields` | the grade's load-bearing measurement, asserted |

## Ablation

Direction predicted in writing before the run, in the report on the card: with
the source change reverted and the pin kept, the three cases that measure the
derived path go RED and every control stays GREEN. Both legs run from the
committed state; the restore leg is checked out from `HEAD` and proved by an

## Scope

- ⛔ objectui#8793 stays open and is not touched here; it is a different hole on
  the same seam. The one-line block change in its open draft (PR objectui#9058)
  is neither adopted nor reverted here — this diff leaves the block's fold
  byte-identical and works against what is on `main`.
- `RecordRelatedListRenderer.columnMembers.test.tsx` (the block's fold) and
  `RelatedList.columnIdentityAccessor.test.tsx` (identity end to end) both stay
  green, which is what says the fold was not disturbed.

## Acceptance notes

- **Surface widening, declared.** `RelatedList` and `RelatedListProps` are both
  exported from this package's public entry (`packages/plugin-detail/src/index.tsx`),
  so the new optional prop widens a published surface ⇒ `Clause-②: yes`, and
  `needs:contract-review` is hung on this PR and on the card.
- ⚠️ **`redactFields` and `enforceFieldSecurity` are renderer-only keys.**
  Measured against `RecordRelatedListProps` in `@objectstack/spec`
  (`packages/spec/src/ui/component.zod.ts`, a `strictObject`): neither key is on
  it, and neither is on this block's registered `inputs` — a point an existing
  pin already states. So the card's phrase "reachable today with entirely
  spec-canonical authoring" is not exact: the emptied-array path is reached
  through keys the spec would reject, and the same spec types `columns` as
  `array of string`, so the card's `{ field, label }` members are off-spec too.
  ⚠️ This does NOT change the defect or the fix — the renderer honours both keys
  today, so the leak is real and reachable — but "which contract, if any, these
  two keys belong to" is a producer-side question this PR does not answer and
  should not. Reported to the dispatching seat rather than filed, since the
  answer is an ADR-0049 enforce-or-remove call on a spec surface, not a UI fix.
- Noted, not filed: the prose row for `record:related_list.columns` in
  `apps/console/src/__tests__/registry-inputs-spec-parity.test.ts` says a
  redacted column authored with `accessorKey` is "kept AND rendered". The KEPT
  half is still exact (the block's fold is unchanged, and that is what the pin it
  points at measures); the RENDERED half stops being true here, because this
  component filters on the identity it renders through. No assertion moves, so
  nothing reds. Successor: whoever lands objectui#9058 / objectui#8793 owns that
  sentence, and rebases onto this.

---

### DIFF
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,21 @@
+---
+'@object-ui/plugin-detail': minor
+---
+
+fix(plugin-detail): `record:related_list` redaction now reaches the auto-derived columns
+
+Redacting **every** authored column used to switch redaction off. The block
+filtered its authored `columns` against `redactFields`, handed `RelatedList` the
+empty result, and `RelatedList` read an empty array as "no columns were
+authored" — falling through to auto-derivation, which the block's redaction list
+never reached. The redacted field came back, and the fallback could surface
+fields the author never listed at all.
+
+`RelatedList` now takes the list as a `redactFields` prop and applies it on every
+path that decides columns — the authored array, the `highlightFields` prominence
+set and the heuristic field walk — so one policy filters all three. An authored
+array emptied by redaction falls through to the derived set exactly as it already
+did, and that set is now filtered too.
+
+Field-level security is unchanged: it was, and remains, enforced independently on
+every path.
--- a/packages/plugin-detail/src/RelatedList.tsx
+++ b/packages/plugin-detail/src/RelatedList.tsx
@@ -134,6 +134,26 @@ export interface RelatedListProps {
   toolbarActions?: RelatedRowActionDef[];
   /** Execute one of {@link toolbarActions} (no row context). */
   onToolbarAction?: (action: RelatedRowActionDef) => void | Promise<void>;
+  /**
+   * Field names this list must never show, whatever decided its columns
+   * (objectui#9053).
+   *
+   * The block-level authoring preference `record:related_list` reads as
+   * `redactFields`, pushed down to the component that actually decides
+   * columns. It used to be applied only where the block could apply it — over
+   * the AUTHORED `columns` array — and this component has two more paths that
+   * decide columns on their own (`highlightFields` prominence and the
+   * heuristic field walk), which that list never reached. Redacting EVERY
+   * authored column therefore emptied the array, the empty array read as "no
+   * columns were authored", and the derived set brought the redacted field
+   * straight back: applying the control maximally switched it off.
+   *
+   * ⚠️ This is an AUTHORING preference, not the permission boundary. Field
+   * security is enforced independently and unconditionally through
+   * `perms.checkField(..., 'read')` on every path below; a field that must be
+   * unreachable belongs in FLS, not here.
+   */
+  redactFields?: string[];
   /** Maximum number of columns to auto-generate. Default 6. */
   maxColumns?: number;
   /** Page size for pagination (enables pagination when set) */
@@ -391,6 +411,7 @@ export const RelatedList: React.FC<RelatedListProps> = ({
   toolbarActions,
   onToolbarAction,
   add,
+  redactFields,
   maxColumns = 6,
   pageSize,
   defaultSort,
@@ -1009,6 +1030,22 @@ export const RelatedList: React.FC<RelatedListProps> = ({
   //  - Cap at `maxColumns` to keep the related card readable; users can
   //    click "View All" to see the full list.
   const perms = usePermissions();
+  /**
+   * [objectui#9053] The redaction list as a lookup, memoised on the PROP's
+   * identity so `effectiveColumns` keeps the reference-stable dependency the
+   * rest of this file is built around: a caller that passes no list passes
+   * `undefined`, which never changes, and one that passes its authored array
+   * passes it by reference.
+   */
+  const redactedFields = React.useMemo(
+    () =>
+      new Set(
+        (Array.isArray(redactFields) ? redactFields : []).filter(
+          (f): f is string => typeof f === 'string' && f.length > 0,
+        ),
+      ),
+    [redactFields],
+  );
   const effectiveColumns = React.useMemo(() => {
     const relatedObjectName = objectName || api || '';
     // FLS: drop columns the current user cannot read on the related object.
@@ -1026,6 +1063,27 @@ export const RelatedList: React.FC<RelatedListProps> = ({
             const key = c?.accessorKey || columnIdentity(c);
             return key !== referenceField;
           })
+        : cols;
+
+    /**
+     * [objectui#9053] Redaction — the block-level authoring preference, asked
+     * on EVERY path below rather than only over the authored array.
+     *
+     * Identity is resolved the way this component resolves it everywhere else
+     * (`accessorKey || columnIdentity`), because that is the key it RENDERS
+     * through: filtering on any other reading would leave a column refused by
+     * name and drawn by accessor.
+     *
+     * ⛔ Fail-OPEN on a column it cannot name, exactly like `filterFLS` beside
+     * it. Whether an entry whose identity does not resolve should be kept or
+     * dropped is objectui#8793's question, not this one, and answering it here
+     * would fold two policies into one diff.
+     */
+    const isRedacted = (key: unknown): boolean =>
+      redactedFields.size > 0 && !!key && redactedFields.has(String(key));
+    const filterRedacted = (cols: any[]): any[] =>
+      redactedFields.size > 0
+        ? cols.filter((c) => !isRedacted(c?.accessorKey || columnIdentity(c)))
         : cols;
 
     /**
@@ -1268,7 +1326,21 @@ export const RelatedList: React.FC<RelatedListProps> = ({
      };
      if (columns && columns.length > 0) {
        const normalized = columns.map(normalizeColumn);
-       return pruneEmpty(filterFLS(filterFK(normalized)));
+       // [objectui#9053] Redaction is applied to the authored candidates FIRST
+       // and their emptiness judged HERE, so an array emptied by redaction
+       // behaves exactly as it already does when the BLOCK empties it upstream
+       // — it falls through to the derivation below, which is redaction-filtered
+       // too. That keeps one outcome for one input: the same authoring must not
+       // render a derived list when the block happened to name the column and an
+       // empty one when only this component could. ⛔ What an emptied-by-security
+       // column set should LOOK like is objectui#9053's deferred question; this
+       // deliberately answers it the way the shipping path already answers it
+       // rather than inventing a second answer. Emptiness produced by FLS or by
+       // `pruneEmpty` keeps its existing meaning untouched: still an empty list.
+       const candidates = filterRedacted(normalized);
+       if (candidates.length > 0) {
+         return pruneEmpty(filterFLS(filterFK(candidates)));
+       }
      }
     if (!objectSchema?.fields) return [];
 
@@ -1285,7 +1357,9 @@ export const RelatedList: React.FC<RelatedListProps> = ({
         )
       : [];
     if (declaredHighlights.length > 0) {
-      const hf = pruneEmpty(filterFLS(filterFK(declaredHighlights.map(normalizeColumn))));
+      const hf = pruneEmpty(
+        filterFLS(filterFK(filterRedacted(declaredHighlights.map(normalizeColumn)))),
+      );
       if (hf.length > 0) return hf.slice(0, Math.max(1, maxColumns));
     }
 
@@ -1331,6 +1405,10 @@ export const RelatedList: React.FC<RelatedListProps> = ({
         if (key === 'id' || key === referenceField) return false;
         if (def?.hidden) return false;
         if (def?.type && SKIP_TYPES.has(def.type)) return false;
+        // [objectui#9053] Redaction: drop redacted fields from the walk too —
+        // asked here rather than over `generated` so the priority sort and the
+        // `maxColumns` slice below both see the set the reader will get.
+        if (isRedacted(key)) return false;
         // FLS: drop unreadable fields from auto-derived columns too.
         if (perms?.isLoaded && resolvedObjectName
             && !perms.checkField(resolvedObjectName, key, 'read')) {
@@ -1378,7 +1456,7 @@ export const RelatedList: React.FC<RelatedListProps> = ({
 
     const pruned = pruneEmpty(generated);
     return pruned.slice(0, Math.max(1, maxColumns));
-  }, [columns, objectSchema, objectName, api, resolveFieldLabel, referenceField, relatedData, maxColumns, lookupLabels, perms]);
+  }, [columns, objectSchema, objectName, api, resolveFieldLabel, referenceField, relatedData, maxColumns, lookupLabels, perms, redactedFields]);
 
   /**
    * [#6108] The SERVED per-column sortability projection for this object —
--- a/packages/plugin-detail/src/__tests__/RecordRelatedListRenderer.redactedDerivation-9053.test.tsx
+++ b/packages/plugin-detail/src/__tests__/RecordRelatedListRenderer.redactedDerivation-9053.test.tsx
@@ -1,0 +1,244 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * objectui#9053 — redaction must reach EVERY path that decides columns.
+ *
+ * `record:related_list` filters its authored `columns` against an allow-list
+ * built from `enforceFieldSecurity` / `redactFields`, then hands the survivors
+ * to `RelatedList`. `RelatedList.effectiveColumns` takes the authored array
+ * only `if (columns && columns.length > 0)`; otherwise it AUTO-DERIVES a set
+ * from the child object's `highlightFields` or from a heuristic field walk.
+ *
+ * So redacting EVERY authored column empties the array, the emptied array reads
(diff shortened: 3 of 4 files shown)


os-steve:
## ⛔ Dispatching seat is OFF SHIFT — this PR is **NOT reviewed** and must not be landed as-is

This PR was delivered after the `domain:ui` seat stood down (maintainer: 「当前任务处理完下班」). It arrived complete and the report is strong, but **no in-seat clause-② review has been performed**, so:

⭐ **`needs:contract-review` stays HUNG on BOTH carriers, and that is the CORRECT state — not an oversight.** `contract-review.md:38` conditions clearing on a **PASS**. There is no PASS here. ⛔ Do not clear either carrier as routine cleanup; the review has to happen first, and it belongs to this card's dispatching seat (`:27`).

⛔ Do not arm. ⛔ Do not merge.

### Why this one in particular should not be rushed

The dev raised an open question and correctly declined to answer it, and it **bears directly on this PR's own contract review**:

> `redactFields` and `enforceFieldSecurity` are honoured by three `plugin-detail` blocks but are on **no declared contract** — `RecordRelatedListProps` is a `strictObject` carrying neither, and neither is on the blocks' registered `inputs`. **This PR makes one of them load-bearing on one more path.**

Filed as objectui#9095 with its measurement. ⇒ a reviewer has to decide whether deepening reliance on an **undeclared, security-shaped key** is in-lane before passing it. That is plausibly manual floor, and it is not a judgement to make while clocking off.

### Also corrects the card's own premise — and the correction stands

objectui#9053 says the leak is *"reachable today with entirely spec-canonical authoring."* The dev measured that this is **not exact**: neither `redactFields` nor `enforceFieldSecurity` is in the spec at all, and that same spec types `columns` as `z.array(z.string())`, so the card's `{ field, label }` members are off-spec too. ⭐ **The defect, its severity and the fix are unaffected** — the renderer honours both keys, so the leak is real and worth fixing. Only the characterisation is wrong. Recorded so it is not re-asserted.

### What the dev did that a reviewer should not have to re-derive

- **ZONE 2 A verified BEFORE any edit, and it HELD with one correction to *how*:** the `highlightFields` branch calls `filterFLS` literally while the heuristic walk inlines `perms.checkField(...)` — the identical predicate. ⇒ FLS **is** re-applied on the derived path, so triage's **p2 (not p1)** grade and the "push redaction down" remedy both stand. That was the one measurement the brief said to falsify first, and it is now pinned as its own case rather than left in prose.
- **Ruling B implemented, ruling A not:** no denied/empty state was added; an array emptied by redaction falls through to derivation exactly as it already did when the block emptied it upstream.
- **objectui#9058 neither adopted nor reverted** — the block's fold is **byte-identical to `main`**, and the new filter is fail-OPEN on an unresolvable identity like the `filterFLS` beside it, so objectui#8793's question is untouched. The scope fence held.
- **Clause-② determined from the barrel, not guessed:** `index.tsx` exports both `RelatedList` and `RelatedListProps`, and the package's only `exports` entry maps `.` to `dist/index.d.ts` ⇒ `yes`.
- **Ablation predicted in writing before the run and matched** (4 red / 4 green, the four predicted cases), with **two honest corrections disclosed rather than buried**: the highlightFields case failed assertion-shaped rather than timeout-shaped, and the first ablation attempt aborted because the on-disk guard anchored on a string BASE already contained — re-cut and re-run, not quietly retried until something stuck.
- **NOT MEASURED is recorded as NOT MEASURED**, with reasons: `check:sdui-registration-pins` exit 2 (prerequisite), `check:readme-exports` (needs a repo-wide build), and an exit-99 lock queue-timeout re-queued rather than read as a pass.

### For whoever picks this up

1. In-seat clause-② review, with objectui#9095 answered or explicitly fenced first.
2. On PASS: clear **both** carriers in the same pass with provenance (`:38`), then arm.
3. On landing: strip `pm:*` **and** the assignee from card objectui#9053 — `Fixes` has never stripped either.
4. objectui#9058 (card objectui#8793) is **held waiting on this PR** and rebases onto it once it lands.

Seat state: objectui#5560 (body R14, one round stale — see its top comment).

<sub>PM seat · `domain:ui` @ objectui · off shift 2026-09-11 00:42Z</sub>

---

os-tesla:
## Contract review

**Head reviewed:** `571fba8c7bd7bb63dd6f207a755e6c7448de071b` (branch tip confirmed equal to `head.sha` by `git ls-remote`; `commits: 1`, so every check run below is pinned to this head).
**Seat:** `domain:ui` @ objectui, in-seat clause-② review at the lane default tier — `contract-review.md` 「余席条款②复核 = 默认判断档自审加门禁」 and 「真闸门在 PR 或报告时点的席内复核,达档只在 spec 席」.
**Reviewed at** 2026-09-11T02:4xZ.

### ① Derived judgments — accept-set and public-surface changes, itemised from the diff

| # | change | judgement |
|:--|:--|:--|
| 1 | **`redactFields?: string[]` added to `RelatedListProps`** (`RelatedList.tsx:+137`). `RelatedList` and `RelatedListProps` are both exported from `packages/plugin-detail/src/index.tsx` ⇒ this **widens a published surface**. | ✅ **Correct, and correctly declared.** Additive and optional ⇒ no existing caller stops compiling. The dev declared it unprompted in `## Acceptance notes` and hung both carriers. |
| 2 | **No widening of the authored-metadata accept set.** `redactFields` was already read off the block schema before this PR; the block diff (`record-related-list.tsx:+9`) only forwards the existing `redact` local downward. | ✅ No new key an author may write. The authoring door is unchanged. |
| 3 | ⚠️ **A control-flow change beyond redaction**, and the one I checked hardest: the authored-columns branch is no longer unconditionally terminal. It can now fall through to derivation. | ✅ **Correctly bounded.** The `candidates.length > 0` guard sits **before** `filterFLS` and `pruneEmpty`, so a set emptied by **FLS** or by **`pruneEmpty`** still returns the empty array it always did. Only a set emptied by **redaction** falls through. The dev states this in the code comment and it is what the diff does. |
| 4 | Fail-**open** on a column whose identity does not resolve, matching the `filterFLS` beside it. | ✅ Deferred, not decided — see ③. |

⭐ **The one claim the whole verdict rests on, re-derived rather than taken from the report** (fact ㊶): does the filter refuse on the same identity the component **renders** through? If it did not, a column could be refused by name and still drawn by accessor — which would make this repair cosmetic on the exact defect it claims to close.

Verified by me on `origin/main`, with controls:

- `filterRedacted` refuses on `c?.accessorKey || columnIdentity(c)`.
- The component renders through that **byte-identical** spelling at `RelatedList.tsx:1018`, `:1026`, `:1090`, `:1234`, `:1464`, `:1489`, `:1492`, `:1696` — `:1464` and `:1696` being the cell-value reads.
- The heuristic walk filters on `key`, and `key` is what becomes the accessor at `:1369` (`accessorKey: key`). Same identity on that path too.
- Positive control `columnIdentity` in that file: **10** lines. Negative control `zzzNotARealAccessor`: **0**.

⇒ the filter and the renderer agree on identity on all three paths. **Claim holds.**

### ② Semver

`.changeset/olive-pugs-repeat.md` declares `'@object-ui/plugin-detail': minor`. ✅ **Consistent.** An additive optional prop on a published interface grows the surface ⇒ not `patch`; nothing breaks ⇒ not a break (and this repo forbids `major` regardless). `Changeset Declaration`, `Changeset Bump Policy`, `Changeset Fixed Group Check` and `Changeset Overwrite Report` are all `success` on this head.

### ③ Boundary flags — both raised by the dev, both answered here

**Flag 1 — fail-open on an unnameable column.** ✅ **Upheld as deferred.** That is objectui#8793's subject and has its own open draft (objectui#9058). Closing it inside this diff would fold two policies into one change on a `security` surface, which is exactly what the card's scope fence forbids. ⛔ Not a defect of this PR.

**Flag 2 — "which contract, if any, do `redactFields` / `enforceFieldSecurity` belong to?"** ⭐ **Answered by triage AFTER this PR was written, and it is not a gate on it.** The dev filed objectui#9095 for it; the triage seat closed that as `duplicate` of **objectui#8649** at 01:10Z (`5627847250`) and ruled, verbatim:

> ⚠️ ⛔ **Nothing here touches #9053's in-flight PR.** Its repair stands; the declaration question is #8649's and always was.

⇒ the declaration question is live and owned by objectui#8649 (`domain:spec`, p2 — ⛔ not this lane). ⛔ It does **not** gate this review. The predecessor seat's handover made objectui#9095 a blocker on this PR; that blocker is discharged by re-routing, ⛔ not by anyone answering it.

**Unflagged, and checked anyway:** no `content/docs/releases/` change; no file unrelated to the card; 4 files, all on the seam; no `+0/-0` file (so no NUL suspicion to resolve); no refusal/error-envelope case in the diff, so the ADR-0112 `code`+`status` assertion rule does not apply here.

### Freshness and collision — measured, not assumed

- ⚠️ `base.sha` reports `16fc4cf2`; the **real merge-base is `f1190b0`** (fact ㊴ — `base.sha` is a branch tip, never the merge-base). `origin/main` is `d2f0c10`.
- `git merge-tree --write-tree origin/main 571fba8c` → **exit 0**, no textual conflict, no worktree touched (fact ㊵).
- Nothing merged since the merge-base touches `RelatedList.tsx` or `record-related-list.tsx` — **empty**, with the control lighting on `0b138da` (#9093) for the same window on the wider package. ⇒ no semantic collision to price.
- **35 checks on this head: 32 `success`, 3 `skipped` (`Test (coverage)`, `Test (coverage shard …/4)`, `dependabot`), 0 red, 0 pending.** `Lint` and `Type Check` both `success`, read individually rather than from a rollup (fact ③).

### Independence pair

```
Implemented-by: claude/issue-9053-related-list-redact-derived  (mode:subagent dev, dispatched by session_01MPaVWWMuWeT5LgB1qoXjVB)
Reviewed-by:    session_01UzHd6hDYatoDn17BuwKxnZ  (domain:ui seat, os-tesla)
```

⇒ implementer and reviewer are **different sessions**. ⛔ Not a SELF-REVIEW.

## Verdict: **PASS**

Both carriers are cleared in the stroke that follows this comment, which cites this record by id. ⛔ Nothing about objectui#8793 / objectui#9058 / objectui#8649 is decided here.

---

os-tesla:
### Carrier clearance — provenance

`needs:contract-review` cleared from **both** carriers (this PR and card objectui#9053) in one stroke by the `domain:ui` seat, per `contract-review.md` 「清标即落地:PASS ⇒ 同席同笔剥双载体」.

| | |
|:--|:--|
| review of record | `5628635238` — `## Contract review`, verdict **PASS** |
| head judged | `571fba8c7bd7bb63dd6f207a755e6c7448de071b` |
| cleared by | `session_01UzHd6hDYatoDn17BuwKxnZ` (`os-tesla`), `domain:ui` @ objectui |
| machine check | `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9090` → **exit 0** (both limbs agree, no widening tell, a review of record names this head) |
| ACCEPT | on the card, `5628646915` |

⚠️ Both label writes were read back after the write: this PR now carries `plugin, tests`; the card carries `bug, pm:dispatched, priority:p2, security, domain:ui`. Neither retains the carrier.

⛔ The clearance judges **this PR's** clause-② increment only. It decides nothing about objectui#8793, objectui#9058 or objectui#8649.

---

PR #9093, opened 2026-09-11
fix(plugin-detail): the summary chip beside the H1 names its field by its label (#8729)

### LINKED ISSUE #8729: bug(plugin-detail): the record page's summary chip names its field by the raw column name — the one band beside the H1 that does not resolve a label
Found while implementing objectui#8464. ⛔ Not fixed there — that card is a display decision about the chip's VALUE; this is about the name the chip gives the field, and changing it there would have moved a pin objectui#8394 owns.

## What

`packages/plugin-detail/src/DetailView.tsx`, the `effectiveSummaryFields.map` render beside the page H1. The chip has no visible label, so its whole accessible name is:

```
aria-label={`${fieldName}: ${display}`}
```

`fieldName` is the raw stored column name, never the resolved label. Every other band of the same page resolves one: `HeaderHighlight` calls `fieldLabel(objectName, field.name, field.label)`, `DetailSection` uses the section field's `label`. So a screen-reader user hears the database column where a sighted reader sees nothing at all.

## Measured

Rendered `DetailView` with `summaryFields: ['owner_ref']` and `fields: [{ name: 'owner_ref', label: 'Owner', type: 'lookup' }]`, on `ed971e8fc`:

| what | value |
|---|---|
| chip `aria-label` | `owner_ref: [object Object]` |
| the field's authored label | `Owner` |

(The value half of that string is objectui#8464 and is fixed there; the `owner_ref` half is this card and is unchanged by it.)

## Why it was not folded into objectui#8464

`detailPage.emptinessAuthority-8394.test.tsx` navigates these chips by `[aria-label^="<field>: "]` and asserts `'stage: Won'` exactly. Changing the name half means editing that instrument, which objectui#8394 / PR #8457 owns — worth doing deliberately, not as a rider.

## If someone picks this up

`useSafeFieldLabel` is already imported in this file (`fieldLabel` is in scope at the render site), so the resolution itself is one call. The work is the instrument: widen `chipFor` in the 8394 file, or give it the `data-summary-chip="<field>"` handle objectui#8464 added, and re-pin the four accessible names it asserts.

## Dedup

Run. Channel: MCP `search_issues` (repo-scoped REST `/search/issues` is refused in this container). Its control fired on the same session — a query for the summary-chip object-value defect returned objectui#8464 itself as the top hit — so this is not a false zero. Query `summary chip aria-label uses raw field name instead of the resolved field label` returned objectui#8231 (a different surface, app-shell's machine-name chip) and a closed item. No duplicate found.

---

os-steve:
Fixes #8729

The `summaryFields` chips beside the record H1 carry no visible label, so
`FIELD: VALUE` is the **whole** of what a screen-reader user hears the field
called — and the name half was the raw stored column. One band of the page
exposed the database spelling, and it was the band only a screen-reader user
perceives.

Placeholders are spelled as words (`FIELD`, `VALUE`, `NAMESPACE`) throughout
this body: angle-bracket-shaped fragments are deleted when a body is saved,
including inside backticks and fenced blocks, so a table written with them
renders as though nothing changed (AGENTS.md, "GitHub 会改写你写进 issue/PR 正文
的字节").

## The ruling, and why the implementation follows from it

Every sibling band on the same page already resolves a label through **one
shared helper**: `HeaderHighlight` calls `fieldLabel(objectName, field.name,
field.label)`, `DetailSection` calls it with `field.label || field.name`. So
this is a **divergence from an established in-repo pattern**, not a missing
capability — the chip now makes the same call rather than growing a fourth
resolution path.

The chip is addressed by NAME (`summaryFields: ['owner_ref']`), so, unlike its
siblings — whose inputs are field objects that already carry a label — it has to
find the fallback it hands that helper. That chain is this render's own field
resolution, not a new one: the author's view entry, then the object schema, then
the stored name as the floor — the precedence `enrichDetailField` states and the
floor `DetailSection` spells. `autoSummaryFields` already merges the same two
sources to PICK the chip.

## Census — the chip has THREE name sites, not one

A fix applied only to `aria-label` would have left one behind:

| branch | where the name lives | moved |
|---|---|---|
| string | the Badge's `aria-label` | yes |
| percent (objectui#9072) | the Badge's `aria-label` | yes |
| cell-renderer (objectui#8732) | an `sr-only` prefix span, **not** `aria-label` | yes |

`data-summary-chip` is a fourth appearance of the raw name and deliberately
keeps it: it is a machine handle for tests and automation, where the stored
column is the right answer.

## PM assumptions — what held, what did not

- **A — `fieldLabel` callable here: HELD, with one correction.** `objectName` IS
  in scope as `schema.objectName` (the same value the option-label branch three
  lines below already passes to `fieldOptionLabel`), and it is optional, so the
  call uses the siblings' own `objectName || ''` spelling. ⚠️ The card's note
  that "`fieldLabel` is in scope at the render site" was not accurate: the file
  called `useSafeFieldLabel()` but destructured **only** `fieldOptionLabel`.
  Adding `fieldLabel` to that destructuring is one of the two source hunks.
- **B — the objectui#8394 pin: the hazard is REAL, and BROADER than recorded.**
  Measured, see below.
- **C — `aria-label` is the whole accessible name: HELD for two branches, FALSE
  for the third, and measured rather than assumed.** The chip carries no
  `title` and no `aria-labelledby` anywhere. For the string and percent branches
  the computed accessible name (jest-dom over dom-accessibility-api) is the
  `aria-label` verbatim. For the renderer-backed branch it is the **empty
  string**: that Badge sets no `aria-label` by design, and a plain `div` is
  `role="generic"`, which takes no name from its content — so the `sr-only`
  prefix buys what the user is READ, not an accessible name. The new pin asserts
  each branch by the mechanism that is actually live on it, and says so.
- **D — one spot: FALSE.** Three, as censused above.
- **E — the card's anchors are stale: HELD.** Re-derived by symbol
  (`effectiveSummaryFields.map`); no line is cited anywhere in this change.
- **F — a changeset is owed: HELD.** `check-changeset-presence` printed
  `3 source file(s) of 1 released package(s) changed, and this change adds no
  changeset` before, and `4 source file(s) ... declares 1 changeset(s)` after. A
  real `patch` bump, not an empty-frontmatter declaration: the accessible name
  of a published component changes for input it already accepts.

## The pin hazard — MEASURED, and NOT re-baselined silently

The filing seat recorded that this change would move a pin objectui#8394 owns.
**It does**, and a census found it moves a second one the card did not name.
Both owning cards have landed — objectui#8394 via PR #8457, objectui#8464 via
PR #8732, both merged — so no in-flight work is being reached into.

Measured by running the four chip pins against the fixed source before touching
any instrument — **6 assertions in 2 files**:

**`detailPage.emptinessAuthority-8394.test.tsx` — 2 cases.** Its `chipFor`
navigated by an `aria-label` prefix selector built from the STORED name, i.e. by
exactly the thing this change moves, so it matched nothing:

```
AssertionError: CONTROL: the filled summary field still renders its chip: expected null not to be null
AssertionError: the auto-detected status slot goes to `stage`, ... : expected null not to be null
```

⭐ That file's own docblock already records this failure shape, from
objectui#8506: *"the harness navigated by exactly the thing that changed"*. The
remedy is the one the card itself suggested: `chipFor` now navigates by the
`data-summary-chip` handle. That is strictly better here than either spelling of
the name — it is a handle rather than a reader-facing string, so no display or
naming decision can move it, and it also finds the renderer-backed chips, which
carry no `aria-label` at all and which the old selector could never have seen.

**That card's subject did not move.** Every emptiness answer — which chip
renders, which does not, and that the picker and the render agree — is asserted
on the same fields with the same data. Exactly one assertion about the chip's
NAME is re-pinned, `stage: Won` to `Stage: Won`; its stated purpose, "the chip
that took the slot reads its value", is preserved.

**`summaryChip.objectValue-8464.test.tsx` — 4 cases**, all fixtures whose
authored label differs from their stored name:

```
expected 'Status: Negotiation' to be 'status: Negotiation'
expected 'Owner: Ada Lovelace'  to be 'owner: Ada Lovelace'
expected 'Stage: Negotiation'   to be 'stage: Negotiation'
expected 'Ratio: 12.3%'         to be 'ratio: 12.3%'
```

The VALUE half of every one of those strings is byte-identical, and that card's
whole `RENDERER_BACKED` table is untouched down to the byte because it declares
`label: field` — the two spellings coincide there, so the naming change is
invisible to it. That is itself the measurement that the change is confined to
the name. Both instruments carry a docblock note saying what moved and why.

**objectui#9072's seam held exactly as its dev described it.**
`summaryChip.percentOneRule-8728.test.tsx` asserts only that the name ENDS WITH
the on-screen percentage, and it is green before, after, and on both ablation
legs. Its percentage half is re-asserted from this side in the new pin, so this
card cannot take it along.

## The new pin — agreement, not a string

`summaryChip.fieldLabel-8729.test.tsx` (9 cases). The headline case reads the
label out of the **highlight strip's own DOM** and requires the chip's computed
accessible name to be that label plus the value — so a repair that humanized the
name differently from its siblings is red, which a per-chip `toBe(...)` would
not catch. Controls: `FLOOR` (no label anywhere keeps the stored name),
`STORED NAME IS THE LABEL` (byte-identical announcement), and
`THE PERCENTAGE HALF`.

## Ablation — predicted in writing, then run

From the committed fix; mutation proven by blob hash against the `HEAD` blob,
restore proven by hash plus an empty `git diff HEAD`, both legs restored under
an `EXIT INT TERM` trap. No build step is involved: the root vitest config
aliases these packages to source, which is why a source edit alone moves the
result.

| leg | prediction | measured |
|---|---|---|
| A — fix removed, all three sites print the stored column | 7 of 9 red; the two byte-identical controls green; the objectui#9072 seam file fully green | exactly that (7 failed / 13 passed of 20 across both files) |
| B — label read locally instead of through the siblings' resolver | only `A TRANSLATED SESSION` red | exactly that (1 failed / 19 passed) |

Leg B is the one that matters for the ruling: it is what discriminates "called
the siblings' helper" from "read the label off the schema", and only the
translated-session case can tell them apart.

## Verification

- `pnpm exec vitest run packages/plugin-detail/` — **164 files, 1537 tests, all
  passed** (whole affected package, the only package this diff touches).
- `eslint packages/plugin-detail` — **219 files, 0 errors** (986 warnings, all
  pre-existing `no-explicit-any`). Narrowing evidence: the universe is
  per-package `eslint .` under `turbo run lint`, and the whole affected package
  was linted rather than only the 4 changed files; the count is read from
  `--format json`; and the flat config declares no `parserOptions.project` or
  `projectService`, so type-aware linting is off and a diff inside one package
  cannot move any untouched file's verdict. ⚠️ A first pass with
  `--no-inline-config` reported 3 errors in files this PR does not touch — that
  flag is an objectstack-ism, not this repo's invocation, and it was disabling
  legitimate inline disables. Not a finding.
- `node scripts/check-changeset-presence.mjs`, `check-changeset-no-major.mjs`,
  `check-control-bytes.mjs`, `check-new-cross-file-line-citations.mjs` — all
  green; verdict lines quoted in the report.
- `node scripts/check-governed-queue-guard.mjs --test` on this diff's paths:
  `NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none
  matched.`
- `pnpm --filter "@object-ui/plugin-detail^..." build` then
  `pnpm --filter @object-ui/plugin-detail run type-check` — **both green**
  (`tsc --noEmit` plus `tsc -p tsconfig.test.json`). ⚠️ The closure build is not
  optional: a first type-check against an unbuilt closure reported
  `Cannot find module '@object-ui/components'` for the whole package — the
  stale-dist trap, not a reading. The new pin is proven to be INSIDE the
  type-checked set rather than assumed to be: `tsc -p tsconfig.test.json
  --listFiles` lists `summaryChip.fieldLabel-8729.test.tsx`.

## Acceptance notes

- The ablation legs are one-off proofs; no ablation artefact is left in the
  tree.
- Session reference, as a code span so it survives a body edit:
  `https://claude.ai/code/session_01MPaVWWMuWeT5LgB1qoXjVB`.
- `needs:contract-review` is hung on both carriers (this PR and card
  objectui#8729) per the claim's `Clause-②: yes`; the PM seat clears both on
  PASS.
- Out of scope, noted and not filed: none. The `no-explicit-any` warnings in
  this package are a repo-wide style state, not a defect; the
  `--no-inline-config` reading above was my own instrument error.

---

### DIFF
--- a/.changeset/8729-summary-chip-accessible-name.md
+++ b/.changeset/8729-summary-chip-accessible-name.md
@@ -1,0 +1,31 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+The record page's summary chip names its field by its LABEL, like every other band
+of the page (objectui#8729).
+
+The chips beside the record H1 carry no visible label, so `<field>: <value>` is the
+whole of what a screen-reader user hears — and the name half was the raw stored
+column. A record whose `owner_ref` column is authored as "Account Owner" announced
+`owner_ref: Ada Lovelace`, so the one band that only a screen-reader user perceives
+was also the one band exposing the database spelling.
+
+Every sibling on the same page already resolves a label through one shared helper —
+`HeaderHighlight` calls `fieldLabel(objectName, field.name, field.label)` and
+`DetailSection` calls it with `field.label || field.name`. This chip now makes the
+same call, so the chip and the highlight strip one band below name the same field
+the same way, including under a translated session, where the resolver (not the
+authored string) decides.
+
+The chip is addressed by NAME (`summaryFields: ['owner_ref']`), so it resolves the
+label it hands that helper from this render's own field resolution: the author's
+view entry first, then the object schema, then the stored name as the floor — the
+same precedence `enrichDetailField` states and the same floor `DetailSection`
+spells. A field whose authored label already equals its stored name announces
+byte-for-byte as before.
+
+All three of the chip's name sites move together: the string branch's `aria-label`,
+the percent branch's `aria-label`, and the renderer-backed branch's visually-hidden
+prefix. The `data-summary-chip` attribute keeps the raw stored column — it is a
+machine handle for tests and automation, not a name for a reader.
--- a/packages/plugin-detail/src/DetailView.tsx
+++ b/packages/plugin-detail/src/DetailView.tsx
@@ -294,7 +294,7 @@ export const DetailView: React.FC<DetailViewProps> = ({
   const { t } = useDetailTranslation();
   // Tenant default currency (ADR-0053) for summary metrics whose field omits one.
   const { currency: tenantCurrency } = useLocalization();
-  const { fieldOptionLabel } = useSafeFieldLabel();
+  const { fieldOptionLabel, fieldLabel } = useSafeFieldLabel();
 
   // Field-level permission gate. Filter section.fields and top-level
   // fields based on the current user's read permissions BEFORE any
@@ -1059,6 +1059,44 @@ export const DetailView: React.FC<DetailViewProps> = ({
                     .find((f) => f.name === fieldName);
                   const objField = objectSchema?.fields?.[fieldName];
                   const ftype = sectionField?.type || objField?.type;
+                  // ── The chip's NAME half ──────────────────────────────────
+                  //
+                  // The chip carries no visible label, so what follows is the
+                  // WHOLE of what a screen-reader user hears the field called —
+                  // and it used to be `fieldName`, the raw stored column
+                  // (objectui#8729). Every other band of this page resolves a
+                  // label: `HeaderHighlight` and `DetailSection` both call
+                  // `fieldLabel(objectName, name, authoredLabel)`. This is that
+                  // same call, not a fourth resolution path — a chip and the
+                  // highlight strip one band below must name the same field the
+                  // same way, including when a translation overrides the
+                  // authored label.
+                  //
+                  // The fallback chain is this render's own field resolution,
+                  // not a new one: the summary chip is addressed by NAME only
+                  // (`summaryFields: ['owner_ref']`), so unlike the siblings —
+                  // whose inputs are field objects that already carry a label —
+                  // it has to find one. `sectionField` (the author's explicit
+                  // entry) wins over `objField` (the object schema), exactly as
+                  // `enrichDetailField` states it, and `fieldName` is the floor,
+                  // exactly as `DetailSection` spells it (`field.label ||
+                  // field.name`). `autoSummaryFields` above already merges the
+                  // two sources this way to PICK the chip; this names it from
+                  // the same pair.
+                  //
+                  // ⚠️ NOT `chipField.label` below: `enrichDetailField` copies
+                  // `ENRICHED_FIELD_METADATA_KEYS`, and `label` is deliberately
+                  // not one of them — so that bag carries `sectionField?.label`
+                  // alone and an object-schema label would be dropped.
+                  //
+                  // ⛔ `data-summary-chip` keeps the RAW name on purpose: it is
+                  // a machine handle for tests and automation, not a name for a
+                  // reader, and a stored column is exactly what it should say.
+                  const chipLabel = fieldLabel(
+                    schema.objectName || '',
+                    fieldName,
+                    sectionField?.label || objField?.label || fieldName,
+                  );
                   let display: string = String(val);
                   let percentValue: number | null = null;
                   try {
@@ -1182,7 +1220,7 @@ export const DetailView: React.FC<DetailViewProps> = ({
                             it — hiding the very value this branch exists to
                             show. Same accessible name, `field: value`, composed
                             from content instead. */}
-                        <span className="sr-only">{`${fieldName}: `}</span>
+                        <span className="sr-only">{`${chipLabel}: `}</span>
                         <ChipCellRenderer value={val} field={chipField as any} />
                       </Badge>
                     );
@@ -1194,7 +1232,7 @@ export const DetailView: React.FC<DetailViewProps> = ({
                         key={fieldName}
                         variant="secondary"
                         className="text-xs bg-primary/10 text-primary border-transparent hover:bg-primary/15 gap-1.5 pl-2 pr-2"
-                        aria-label={`${fieldName}: ${display}`}
+                        aria-label={`${chipLabel}: ${display}`}
                         data-summary-chip={fieldName}
                       >
                         <span
@@ -1215,7 +1253,7 @@ export const DetailView: React.FC<DetailViewProps> = ({
                       key={fieldName}
                       variant="secondary"
                       className="text-xs bg-primary/10 text-primary border-transparent hover:bg-primary/15"
-                      aria-label={`${fieldName}: ${display}`}
+                      aria-label={`${chipLabel}: ${display}`}
                       data-summary-chip={fieldName}
                     >
                       {display}
--- a/packages/plugin-detail/src/__tests__/detailPage.emptinessAuthority-8394.test.tsx
+++ b/packages/plugin-detail/src/__tests__/detailPage.emptinessAuthority-8394.test.tsx
@@ -66,9 +66,33 @@
  * The replacement navigates by the field LABEL, which no placeholder change can
  * move: each chip is `<span>{label}</span>` followed by the value slot, so
  * `chipOf(label)` is the chip box and the affordance is read INSIDE it. That
- * also strengthens the assertion from "how many" to "which chip", the way the
- * summary chips below are already read through the `aria-label="<field>:
- * <display>"` each `Badge` carries.
+ * also strengthens the assertion from "how many" to "which chip".
+ *
+ * ⚠️ **The summary chips below were re-derived the same way, and for the same
+ * reason, by objectui#8729** — declared here rather than done quietly, because
+ * this instrument belongs to this card and not to that one.
+ *
+ * They used to be navigated by `[aria-label^="<storedName>: "]`, which made the
+ * harness depend on the chip NAMING ITS FIELD BY THE STORED COLUMN — the very
+ * thing objectui#8729 changed (the chip now resolves a label through
+ * `fieldLabel`, as every other band of this page already did). That is exactly
+ * the objectui#8506 shape one paragraph up: the navigation target moved, so
+ * `chipFor` matched nothing and two cases failed while the emptiness authority
+ * they exist to guard was untouched.
+ *
+ * `chipFor` now navigates by `data-summary-chip="<storedName>"`, the machine
+ * handle objectui#8464 added. That is strictly better here than either
+ * spelling of the name: it is a handle rather than a reader-facing string, so
+ * no display or naming decision can move it, and it also finds the
+ * renderer-backed chips — which carry no `aria-label` at all and which the old
+ * selector could never have seen.
+ *
+ * ⭐ **This card's subject did not move.** Every emptiness answer below —
+ * which chip renders, which does not, and that the picker and the render agree
+ * — is asserted on exactly the same fields with exactly the same data. One
+ * assertion about the chip's NAME moved (`stage: Won` → `Stage: Won`); its
+ * stated purpose, "the chip that took the slot reads its value", is preserved
+ * and now reads the value under the field's resolved label.
  */
 
 import { describe, it, expect, beforeAll, afterEach } from 'vitest';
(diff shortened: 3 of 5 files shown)


os-steve:
## 条款② 席内复核 — **PASS on substance.** ⛔ Not armed yet: CI is still running

Reviewing seat: `domain:ui @ objectui`, the dispatching seat for card objectui#8729 (`contract-review.md:27`). First review on this PR — comments read before writing (㊽).

⏳ As of this comment the four test shards, `Type Check` and `Lint` are **in progress**; everything else is green. This verdict is on the clause-② substance and the diff. **Arming happens on all-green, not on this comment** — and a red shard would reopen it.

### The declaration is right

`Clause-②: yes`. The **accessible name of a published component changes for input it already accepts** — `status: Negotiation` becomes `Status: Negotiation`. That is the "changes a published function's observable answer" limb. The changeset is a real `patch` bump rather than an empty-frontmatter declaration, which is the correct reading: a behaviour change, not merely a test-file touch.

### Why this is the non-escalating class

**A divergence from an established in-repo pattern, not a new capability.** `HeaderHighlight` calls `fieldLabel(objectName, field.name, field.label)`; `DetailSection` calls it with `field.label || field.name`. The chip was the one band on the page that did not, and it is the band with **no visible label** — so `FIELD: VALUE` is the whole of what a screen-reader user hears the field called, and the name half was the database spelling. Making the fourth band make the same call is 恢复不变量.

### ⭐ The part I scrutinised hardest: it edits two OTHER cards' pins

This is normally where a diff goes wrong, and it does not.

**The licence first.** Both owning cards have landed — objectui#8394 closed by merged PR #8457 (verified at source, not accepted from the report), and objectui#8464's instrument is present on `main` as a *modified* file rather than an added one. ⇒ nothing in flight was reached into.

**objectui#8394 — the harness was re-pointed, and to something strictly better.** `chipFor` navigated by `[aria-label^="<storedName>: "]`, i.e. by *exactly the thing this change moves*, so it matched nothing and two cases failed while the emptiness authority they guard was untouched. It now navigates by `data-summary-chip`, the machine handle. That is better than either spelling of the name: a handle no display or naming decision can move, and it finds the renderer-backed chips **which carry no `aria-label` at all and which the old selector could never have seen**.

⭐ That file's own docblock already records this failure shape from objectui#8506 — *"the harness navigated by exactly the thing that changed"* — so the remedy is the one the instrument itself prescribes, not an improvisation.

**And the one NAME assertion was RE-PINNED, not loosened.** `stage: Won` → `Stage: Won`, with the reason in place: *"Re-pinned here rather than loosened, so this line keeps failing for a render-only fix."* ⇒ the instrument is no weaker than it was. That is the distinction that matters and it was made deliberately.

**objectui#8464 — the control is elegant and unprompted.** Four expected names moved; **every VALUE half is byte-identical**, and the whole `RENDERER_BACKED` table is untouched down to the byte *because it declares `label: field`*, so the two spellings coincide there. ⭐ The dev's own reading is the right one: that table staying byte-identical **is itself the measurement** that the change is confined to the name half. A weaker diff would have asserted that; this one demonstrates it.

Every touched instrument carries a docblock note saying what moved, why, and whose card it belongs to. ⛔ No assertion deleted, none relaxed.

### Assumptions falsified — three of six, and all three usefully

- **D — "one spot": FALSE.** The chip has **three** name sites: the string branch's `aria-label`, the percent branch's `aria-label`, and the renderer-backed branch's `sr-only` prefix. A fix applied only to `aria-label` would have looked complete and left a third of the surface behind. The dispatch said one; the census said three.
- **C — "`aria-label` is the whole accessible name": HELD for two branches, FALSE for the third**, and measured rather than reasoned. The renderer-backed Badge sets no `aria-label` and a plain `div` is `role="generic"`, which takes no name from content ⇒ its computed accessible name is the **empty string**, and the `sr-only` prefix is what a reader is *read*, not a name. The new pin asserts each branch **by the mechanism actually live on it**, which is ㉞ applied without being told.
- **A — "`fieldLabel` is already in scope": inaccurate.** The file called `useSafeFieldLabel()` but destructured only `fieldOptionLabel`. Small, and reported rather than silently absorbed.

Those were my assumptions, written into the brief as mine. All three being wrong is the ZONE 2 block doing its job.

### Also credited

- **Ablation leg B is the leg that discriminates the ruling** — "called the siblings' helper" vs "read the label off the schema" are indistinguishable except on a translated session, and that is the single case leg B predicts red. Predicted 1 red, measured 1 red. Choosing a mutation that can only fail for the *right* reason is harder than choosing one that fails loudly.
- **The `--no-inline-config` reading was retracted as the dev's own instrument error**, not filed as a finding — correct, and it is a genuine cross-repo trap: that flag is an objectstack-ism and it disables legitimate inline disables here.
- **The stale-dist trap named, not worked around**: a type-check against an unbuilt closure reported `Cannot find module '@object-ui/components'` for the whole package. Recorded as a prerequisite failure rather than a reading — ㉛.
- **objectui#9072's seam held**, exactly as that PR's dev described it when they left it: the percent pin asserts only that the name *ends with* the on-screen percentage, so this card could take the name half without disturbing it. A handoff honoured on both ends.
- **The body spells placeholders as words** because angle-bracket fragments are stripped from bodies even inside backticks — the live half of ㉜, correctly worked around.

### Outstanding before this lands

1. ⏳ CI green on `9bb5eca6` — four shards, `Type Check`, `Lint`.
2. ⚠️ `mergeable_state: behind`. Not blocking on its own and the merge queue rebases — but ⛔ if a shard comes back red, ㊼ applies **before** any re-run: check whether a fix already exists on `main` and this base predates it (objectui#9075 landed at `16fc4cf20` and objectui#9080 is queued ahead of this).
3. Carriers cleared on this PASS, then ready-for-review, then arm.

<sub>PM seat · `domain:ui` @ objectui · seat post objectui#5560 §0a · in-seat review of record, default tier</sub>

---

os-steve:
## `needs:contract-review` CLEARED on both carriers — provenance

Per `contract-review.md:38` (放行 = 清标即落地:PASS ⇒ 同席同笔剥双载体):

| carrier | before | after |
|:--|:--|:--|
| PR objectui#9093 | `plugin`, `tests`, `needs:contract-review` | `plugin`, `tests` |
| card objectui#8729 | `pm:dispatched`, `priority:p3`, `domain:ui`, `needs:contract-review` | `pm:dispatched`, `priority:p3`, `domain:ui` |

**Authority:** the in-seat clause-② review of record, verdict **PASS** — comment `5627590282`.

**Pre-landing check ② (`:42`) satisfied on the record.** The dev hung both carriers additively per `os-dev.md:288`, read both back, and reported `check-clause2-carriers --pair 9093` **exit 0**. Full hang → in-seat review → clear sequence, in that order. Fourth pair this session to run it correctly end to end.

Marked ready for review; **auto-merge armed**, which merges when the remaining checks pass rather than now.

### ⛔ Standing down for the day — read this before assuming someone is watching

This seat is going off shift on the maintainer's instruction. State left deliberately:

- ⏳ **CI was still running** at review time (four shards, `Type Check`, `Lint`; everything else green). Auto-merge will land this **only if they all pass**. ⚠️ **If any shard comes back red, nobody is watching it** — this PR then needs a human or the next seat, and ⛔ **㊼ applies BEFORE any re-run**: check whether a fix already exists on `main` and this branch's base predates it. `mergeable_state` was `behind`, objectui#9075 landed at `16fc4cf20`, and objectui#9080 is queued ahead of this one.
- ⛔ `pm:dispatched` and the assignee are **still on card objectui#8729** and must be stripped explicitly when this lands — `Fixes` has never stripped either (29 for 29 this session).

Everything else is recorded in seat post objectui#5560 (R14).

<sub>PM seat · `domain:ui` @ objectui · seat post objectui#5560 §0a</sub>

---

PR #9104, opened 2026-09-11
fix(react): keep the CEL predicate envelope intact through a node config bag

### LINKED ISSUE #9100: console(record:alert): `properties.visible` loses its CEL envelope and is evaluated on the LEGACY JS engine — `has()` faults and the fail-soft default renders the banner on every row
> ⤴️ **Moved from objectstack-ai/objectstack#17600** by the triage seat, 2026-09-11T02:2xZ. Original filer `os-project-manager`, 2026-09-11T01:56Z; found from the consumer side by `hotcrm#1887`. GitHub's issue-transfer mutation is not reachable from this session, so the card was **rebuilt** here and the source closed as moved — the body below is verbatim.
>
> **Routing judgement:** 「正文抽掉 objectstack 还成立 ⇒ 当场转仓」. Measured in the body: the artifact (`dist/objectstack.json`) and BOTH `/api/v1/meta/*` endpoints carry the `{ dialect, source }` envelope intact. The envelope is lost **client-side, in the shipped console bundle**. ⇒ the defect stands with objectstack removed, so this is ⛔ not a seam card — both suggested fixes land here.

---

**Measured on `@objectstack/console` 17.4.0 (`.objectui-sha` `53ded82bf7a494f54e344e19099dbf00854b8694`), in a real browser against a real app instance.** Found from the consumer side by hotcrm#1887.

## Summary

A `record:alert` node whose `properties.visible` is authored as the **declared CEL envelope** — `{ dialect: 'cel', source: '…' }`, the `P` tagged-template form — is evaluated by the console on the **legacy JS expression engine**, not the CEL engine. Any CEL stdlib call in the predicate (here `has()`) is then undefined, the predicate throws, and `evaluateCondition`'s fail-soft default (`true`) **renders the component**.

Net effect for the app: a banner gated on a field renders on **every** record, whatever the field says. The gate never bites in either direction.

## The inversion that makes this a bug and not a spelling mistake

Both predicates below are authored in the same repo, in the same `P` envelope form, against the same object. They take **different** engines:

| surface | authored | engine reached | observed error |
| --- | --- | --- | --- |
| `page:header` action `visible` | `P` envelope | **CEL runtime** | `[runtime] No such key: status` |
| `record:alert` `properties.visible` | `P` envelope | **legacy JS** | `"has" is not a function` |

The action path preserves the envelope. The `record:alert` component path does not.

## Where the envelope is lost — it is client-side

The envelope is intact everywhere up to the browser:

1. `objectstack build` artifact `dist/objectstack.json` — `properties.visible` is `{"dialect":"cel","source":"has(record.duplicate_status) && record.duplicate_status == \"confirmed\""}`.
2. `GET /api/v1/meta/pages/lead_detail_page` — same value.
3. `GET /api/v1/meta/page` (the list endpoint the console actually calls, confirmed in the browser network log) — same value.

By the time it reaches the renderer it is a bare string. In the shipped bundle, `record:alert` is registered to the component that computes `h = normalizeVisible(props.visible)`; that normalizer returns the envelope unchanged **only** when `dialect === 'cel'`, and otherwise wraps a bare string as the `${...}` template spelling. `evaluateCondition` routes `${...}` to `this.evaluate(...)`, i.e. the legacy path — which is exactly the stack the browser prints:

```
Failed to evaluate expression: ${has(record.duplicate_status) && record.duplicate_status == "confirmed"}
Error: Failed to evaluate expression "has(record.duplicate_status) && record.duplicate_status == "confirmed"": "has" is not a function
    at e.evaluateExpression
    at e.evaluate
    at e.evaluateCondition
    at Ds                      (useCondition)
    at cs                      (record:alert renderer)
```

The console's own diagnostic states the consequence verbatim:

```
[ObjectUI] A visibility predicate could not be evaluated - node "record:alert" (id: "lead_duplicate_alert_confirmed")
The node was treated as its safe default, which on this surface means the
gate did NOT bite - a predicate that cannot be evaluated reads on screen
exactly like one that said yes.
```

Note also the component's own registration declares `{ name: 'visible', type: 'string' }`, while `@objectstack/spec`'s `ComponentPropsMap['record:alert'].visible` accepts `boolean | string | { dialect, source }`. The two contracts disagree.

## Reproduction (measured, not inferred)

App: hotcrm at `9840c23b`, two `record:alert` nodes on `lead_detail_page` gated on `crm_lead.duplicate_status`.

| record | `duplicate_status` | suspected banner | confirmed banner |
| --- | --- | --- | --- |
| Nimbus Aerospace | `confirmed` | **shown** | **shown** |
| Vertex Analytics | `suspected` | **shown** | **shown** |
| Foundry Robotics | `null` | **shown** | **shown** |

Plus a sweep before any value was written: 6 of 6 leads, all `duplicate_status: null`, both banners shown. All 21 leads in the instance carry `duplicate_status: null`.

## Ablation that isolates `has()`

Replacing **only** the confirmed banner's predicate with a bare non-envelope string carrying no CEL stdlib call — `visible: 'record.duplicate_status == "confirmed"'` — and rebuilding, the gate starts biting correctly:

| record | `duplicate_status` | confirmed banner (bare string) |
| --- | --- | --- |
| Nimbus Aerospace | `confirmed` | shown (correct) |
| Vertex Analytics | `suspected` | hidden (correct) |
| Foundry Robotics | `null` | hidden (correct) |

The suspected banner, left in the `P` envelope with `has()`, stayed wrongly shown on all three in the same run. So the legacy engine **does** bind `record`; the only thing it cannot do is the CEL stdlib. That isolates the fault to envelope routing, not to scope binding. (The mutation was reverted; the file is byte-identical to `HEAD`.)

## Why the consumer cannot work around it

The working spelling is the one that drops `has()`. That is not available to an app:

- It reintroduces the fault the guard exists to prevent the moment this bug is fixed and the predicate reaches CEL again — on `driver-memory` / `driver-mongodb` the column is absent and strict CEL aborts with `No such key`, which since 17.0.0-rc.2 is itself a rejection.
- A predicate that is correct only while the platform is broken is not a fix.

## Suggested fix

Preserve the `{ dialect, source }` envelope on the component-props path so `evaluateCondition` reaches `evaluateCelCondition`, as the action path already does; and align the `record:alert` input declaration (`visible: type 'string'`) with the spec's `boolean | string | envelope`.

Worth a look as a class rather than a single node: any component whose props carry a row predicate through the same normalizer is exposed to the same flattening, and the failure is silent by construction — fail-soft means a broken gate looks exactly like a gate that said yes.

---

os-tesla:
Fixes #9100

## ⭐ The card's stated mechanism is falsified. The defect is real; the fault site is one layer up.

The card (and the triage grading) locate the fault in the normalizer. **It is not there.** Three measurements on `origin/main` source at `d2f0c10`, all in-process:

| probe | result |
|:--|:--|
| `toPredicateInput({dialect:'cel',source})` | returns the envelope **unchanged** — the normalizer is correct |
| `RecordAlertRenderer` mounted **directly**, envelope on `properties.visible` | gates **correctly** (shown on the matching row, hidden on the other two) |
| the same node mounted through **`SchemaRenderer`** | `{confirmed: shown, suspected: shown, null: shown}` — **the card's bug, reproduced** |

⇒ the envelope dies **upstream of every consumer**, exactly as the card's own sentence 「By the time it reaches the renderer it is a bare string」 says. PM assumption A resolves to **candidate 2**, not candidate 1.

⚠️ **Assumption A.1 is also falsified.** The third argument the working call sites pass to `useCondition` is `{ throwOnError, label }`. It selects fail-closed vs fail-soft — it does **not** touch engine routing, so it cannot be the discriminator.

## Where the envelope actually dies

`SchemaRenderer`'s per-value `properties` / `props` evaluation loops hand every bag value to `ExpressionEvaluator.evaluate`, whose **first** action is to unwrap any `{ source: string }` object down to its bare `source`. That unwrap is what makes a `template` envelope interpolate. Applied to a `cel` **predicate** it is destructive: the envelope is the only thing that routes `evaluateCondition` to the canonical `@objectstack/formula` engine.

Those loops run **before** the hoist, so both consumers get a bare string and both take the legacy JS path, where `has()` is not a function:

1. `SchemaRenderer`'s own `shouldHide` chain, off the post-hoist node; and
2. the renderer's `toPredicateInput(…)` → `useCondition(…)` one layer down, which wraps the bare string as `${…}` exactly as documented.

That is the card's console stack verbatim — reproduced here in-process:

```
evaluateExpression <- evaluate <- evaluateCondition <- useCondition <- RecordAlertRenderer
Error: ... "has" is not a function
```

### The real discriminator vs. the action path

The loops are **per-value and shallow**. An action's predicate sits one level down inside the `actions[]` **array**, and `evaluate` returns a non-string untouched — nothing walks into it. A `record:alert` `properties.visible` is a **top-level** bag key and is hit head-on. Same normalizer, different depth. Pinned:

- `evaluate({dialect:'cel',source})` -> the bare source string (the flattening)
- `evaluate([{name:'a',visible:{dialect:'cel',source}}])` -> **deep-equal to the input** (survives)

### Why this hid behind a green suite

`SchemaRenderer.predicateEnvelopeDeclared.test.tsx` (objectui#7530) pins this exact envelope on `visible` / `hidden` / `disabled` and has been green throughout — but every one of its mounts writes the key at **node level** (`{ type, visible: envelope }`), and **nothing runs over a node-level key**. The `properties` spelling is the spec spelling, the one a published artifact and both `/api/v1/meta/*` endpoints actually emit, and it was the only channel never measured.

## ⚠️ The blast radius is wider than `record:alert`, in both polarities

Measured on the broken tree, through the real `SchemaRenderer`, with a non-matching row:

| node | key | expected | measured before fix |
|:--|:--|:--|:--|
| `element:text` | `properties.visibleWhen` | hidden | **shown** (fail-OPEN) |
| `page:card` | `properties.visible` | hidden | **shown** (fail-OPEN) |
| `element:text` | `properties.hidden` | shown | **hidden** (fail-CLOSED — the node vanishes) |

The two HIDE legs return the fail-soft `true` un-negated, so a broken gate makes the node disappear entirely. A one-polarity fix or a one-polarity test would have covered half the defect.

## The fix

One guard, at the measured site, in both bag loops:

```ts
VISIBILITY_CHAIN_KEYS.has(key) && isCelEnvelope(value) ? value : evaluate(value)
```

- `VISIBILITY_CHAIN_KEYS` is built from the two arrays this file **already** declares (`VISIBILITY_SHOW_KEYS` / `VISIBILITY_HIDE_KEYS`), so it cannot drift from the chain `shouldHide` and `winningVisibilityKey` consult.
- `isCelEnvelope` matches the **`cel` dialect only**. `template` and dialect-less envelopes keep the legacy `${…}` behaviour byte-for-byte.
- Non-predicate bag keys keep the flattening they have always had.
- All six keys are stripped by the metadata destructure before `createElement`, so no object value can reach the DOM through this.
- ⛔ The normalizer is **not** touched. No compensating `??` alias, no widened parse anywhere.

Half 2: `record:alert`'s `visible` input declaration, `'string'` -> `['boolean', 'string', 'object']`.

## Contract readings — both verified at source (PM assumption E)

| side | reading |
|:--|:--|
| objectui registration | `{ name: 'visible', type: 'string' }` — `packages/plugin-detail/src/index.tsx` |
| renderer docblock | `visible?: string \| { dialect, source } \| boolean` — `record-alert.tsx` |
| `@objectstack/spec` **17.4.0**, `ComponentPropsMap['record:alert'].visible` | `boolean \| string \| { dialect: 'cel'\|'cron'\|'template', source?, ast?, meta? }` |

Triage's reading is correct and the narrower contract is objectui's. Per `ComponentInput.type`'s own rule — "declare only arms the contract accepts **and** the renderer resolves" — the arm is declared in the same change that makes it reachable.

⚠️ **Bonus falsification:** the registration's own inline comment asserted 「`ComponentPropsMap` carries no `record:alert` entry at rc.6」. That row **does** exist in the installed 17.4.0. The stale clause is corrected in place (comment only, no behaviour).

⛔ `@objectstack/spec` is **not** touched. No cross-repo change is implied: the spec was already correct, as triage ruled.

## Acceptance — a render-level test, plus the ablation

`packages/plugin-detail/src/renderers/__tests__/record-alert.celEnvelope.test.tsx` mounts the **real** `record:alert` renderer through the real `SchemaRenderer`, with `visible` authored as `{ dialect: 'cel', source: 'has(record.duplicate_status) && record.duplicate_status == "confirmed"' }` — the card's predicate verbatim — and asserts on **what a user sees**:

- **group A (controls)** — the harness paints the banner, and this exact channel can hide it. Without these, a `false` verdict would read "never rendered": this renderer has four `return null` paths.
- **group B** — **shown on the confirmed row, HIDDEN on `suspected` and on `null`**, plus a row with the field **absent entirely** (also hidden — which is what `has()` is for).
- **group C (the card's own ablation, pinned as the other arm)** — the bare non-envelope string gates correctly too, and the two arms now **agree on every row**. The inversion the card measured is gone. Both arms are pinned so a regression cannot pass by repairing one.

⚠️ The fail-soft default is what makes this invisible, so every assertion is on the **hidden** case; a shown-only suite cannot fail on this defect.

`packages/react/src/__tests__/SchemaRenderer.predicateEnvelopeConfigBag.test.tsx` pins the fix site itself: all six chain keys, both polarities, plus the three no-op claims (template envelope unchanged, non-predicate key unchanged, shallow-array pass-through).

### Ablation — the guard is load-bearing

Mutated on disk (guard short-circuited to `false`), proven landed (injected-marker count 1; blob `92bd0fd4` != `HEAD` `9cde7019`), run, then restored (blob back to `9cde7019`, `git diff HEAD` empty):

```
ABLATED:  Tests 10 failed | 6 passed (16)   <- incl. {onConfirmed:true, onSuspected:true, onUnset:true}, the card signature
RESTORED: Tests 16 passed (16)
```

Re-run at the shipped head `dcc4925`: guard blob `53a1c060`, mutated `126afb24`, restored `53a1c060`, `git diff HEAD` empty. These suites resolve through vitest's source aliases (the fault stack names `packages/core/src/…`), so no `dist/` leg is involved and the mutation demonstrably took effect.

## Gates

All at the shipped head `dcc4925`.

| command | exit |
|:--|:--|
| `vitest run --project unit --project dom --project dom-heavy packages/react/src/ packages/plugin-detail/src/ apps/console/src/__tests__/` | **0** — 242 files, 2461 tests passed |
| `pnpm --filter "@object-ui/react^..." --filter "@object-ui/plugin-detail^..." build` | **0** (12 projects) |
| `pnpm --filter "@object-ui/console^..." build` | **0** (35 projects) |
| `pnpm --filter @object-ui/console build` | **0** |
| `pnpm --filter @object-ui/react --filter @object-ui/plugin-detail type-check` | **0** (`tsc --noEmit` + `tsc -p tsconfig.test.json`) |
| `node scripts/check-sdui-registration-pins.mjs` | **0** (16 registrations, 518 chunks weighed — needed the console build above; without it the gate exits 2 = measured nothing) |
| `node scripts/check-changeset-fixed.mjs` | **0** |
| `node scripts/check-changeset-no-major.mjs` | **0** |
| `node scripts/check-control-bytes.mjs` | **0** (7292 files) |
| `node scripts/check-new-cross-file-line-citations.mjs` | **0** |
| `node scripts/check-spec-symbol-derivation.mjs` | **0** |
| `node scripts/check-doc-component-types.mjs` | **0** |
| `node scripts/check-eager-closure-budget.mjs` | **0** |
| `eslint <the 5 changed files> --format json` | **0** — 0 errors, 47 warnings, **all pre-existing**: every warning line is outside the diff hunks and the `as any` count is unchanged (8 -> 8 in `SchemaRenderer.tsx`, 0 -> 0 in `plugin-detail/src/index.tsx`) |

### Two gates that did real work, and what they cost

- **`registry-inputs-spec-parity`** refused the new `object` arm until `record:alert.visible` had a MEMBER PIN under objectui#8068. Registered, naming the render-level suite this PR adds. (198 tests green.)
- **`configBag.pin`** (objectui#6761) caught `isCelEnvelope` re-spelling the one config-bag predicate. It now ASKS `isConfigBag` instead — a third read of the one answer, not a fourth spelling. ⚠️ That gate lives in the `unit` project; a `--project dom --project dom-heavy` run alone does not see it.

**Declared narrowing.** The behaviour change is observable only by metadata carrying a `{dialect:'cel'}` envelope on one of six visibility keys at the top level of a `properties` / `props` bag. A scan of `packages/**` + `apps/**` + `examples/**` finds that shape in **7 files, all inside the two owning packages**, and all seven are inside the 242-file run above. Every other shape is a no-op by construction, pinned by the three no-op cases. The repo-wide `pnpm lint` / full-farm run is CI's, not measured here.

⛔ **Not attempted:** the optional browser leg. The in-process render-level suite already exercises the whole renderer -> `toPredicateInput` -> `useCondition` -> `evaluateCondition` chain, which is the gap a normalizer unit test leaves; driving a live preview would add a reading of the same chain at much higher cost.

## Census — triage asked for the exposed set, and 18 is not it

The PM's **18** is the population that *calls* `toPredicateInput`. The population *exposed to the flattening* is a different set, and it splits by tier.

**Membership rule:** a predicate is exposed iff it reaches its consumer as a **top-level key of the node's own `properties` / `props` bag** — i.e. it passes through the shallow per-value loop. A predicate nested inside an array/object value, or sourced from object metadata rather than the node, is not.

**Tier 1 — `SchemaRenderer`'s node gate: every registered component type, not a subset.** The gate reads the post-hoist node, and the hoist copies every `properties.*` key onto it, so exposure is a property of the **key**, not of the component. Six keys, both polarities. Measured above on `element:text` and `page:card`, neither of which reads a predicate itself.

**Tier 2 — renderers that re-read a predicate from their own bag: 10 call sites across 6 files.**

| file | sites |
|:--|:--|
| `plugin-detail/src/renderers/record-alert.tsx` | `visible` ⟵ **the measured one** |
| `components/src/renderers/action/action-bar.tsx` | `visible` |
| `components/src/renderers/action/action-button.tsx` | `visible`, `disabled`, `enabled` |
| `components/src/renderers/action/action-group.tsx` | `visible` |
| `components/src/renderers/action/action-icon.tsx` | `visible`, `disabled`, `enabled` |
| `components/src/renderers/action/action-menu.tsx` | `visible` |

**Excluded, with the reason** (this is the half that narrows 18 down): `DeclaredActionsBar` and `record-quick-actions` read `action.*` from **object metadata**, never from a node bag — that is the card's *working* path. `action-group` / `action-menu`'s per-item legs and `RelatedList`'s toolbar button read an element of an **array**, which the shallow loop passes through. `ActionEngine` / `ActionRunner` / `approvalOverride` / `declaredPredicate` / `schema-validator` / `predicateInput` / `useExpression` / `expression(.zod)` are engine- and type-level and never see a `SchemaRenderer`.

**This PR fixes the whole class in one stroke, and that is not a widening** — the fault was one shared line, so there was no narrower repair available that would have fixed `record:alert` alone. The *enablement* keys (`disabled` / `disabledOn` / `enabled`) are **deliberately left out of the guard** and filed instead; see below.

## Out-of-scope findings (⛔ not fixed here)

1. **Enablement keys have the same defect and are not covered.** `disabled` / `disabledOn` / `enabled` are not in the visibility chain and route through `evaluateEnablementPredicate`; a `cel` envelope on them is flattened by the same loops. Left out per triage's 「fix the measured one … do not widen unbidden」.
2. **The `props` bag never drove the node gate at all** — pre-existing and unrelated to the envelope. Measured on the same tree: `props: { visible: false }` renders, `props: { hidden: true }` renders, while the `properties` spelling of either decides correctly. The hoist copies `properties` onto the node and nothing copies `props`.
3. Noted, not filed: `ExpressionEvaluator.evaluate`'s unwrap returns the **raw CEL source text** for a `cel` envelope on any non-predicate key — useless in every direction (it is not a template, so it is emitted verbatim). Not touched here because changing `evaluate` itself is a core-wide blast radius.

## 维护者速读(草稿)

**改了什么** — `SchemaRenderer` 在把节点 `properties` / `props` 逐值求值时,不再把 `{ dialect: 'cel', source }` 断言信封拍平成裸字符串;顺带把 `record:alert` 的 `visible` 声明从 `'string'` 拓成 `['boolean','string','object']`,与 spec 已声明的契约对齐。

**为什么改** — 这是一个**失败时敞开**的可见性门禁。作者写「满足条件才显示」的横幅,平台在每一行都显示它;真实浏览器实测 6/6 条线索全部误显。信封是唯一把断言路由到权威 CEL 引擎的标记,拍平后落到遗留 JS 引擎,`has()` 不存在 ⇒ 抛错 ⇒ fail-soft 判定为「显示」。⚠️ 反向的两个键(`hidden` / `hiddenOn`)则相反:整个节点直接消失。两种都与「作者本意如此」在屏幕上完全同形,用户永远发现不了。

**风险与代价(含回滚)** — 这是**行为变更**:此前从不生效的 CEL 信封门禁现在真的生效了,依赖它「不咬人」的元数据会看到节点被隐藏(或反向显示)。范围经实测收窄:只有把 `cel` 信封写在六个可见性键之一、且位于 `properties` / `props` 顶层的元数据会变;`template` 信封、无 dialect 信封、非断言键全部逐字节不变,已各自钉住。回滚成本低——改动是两处循环里的一个守卫加一个常量,`git revert` 即可,无数据迁移、无 API 变更。

**席位意见** — (待填)

**你要做的** — 这张卡按 `Clause-②: yes` 走席内契约复核:需要确认的是**声明拓宽**这一半(`record:alert.visible` 从一个 arm 变三个 arm),以及上面「风险与代价」里那条行为变更是否接受。另外请留意:卡片正文与分诊都把缺陷定位在 normalizer,**实测证伪**,真正的故障点在上游一层——复核时请按实测的故障点读,不要按卡片叙述读。

---

---


os-tesla:
## Contract review

**Head reviewed:** `dcc492556cdbcd9905bff64fca7f3e505eea6749`.
**Seat:** `domain:ui` @ objectui, in-seat clause-② review at the lane default tier (`contract-review.md` 「余席条款②复核 = 默认判断档自审加门禁」; the tier fuse binds only the spec and skills seats).
**Reviewed at** 2026-09-11T03:2xZ.

### ① Derived judgments — accept-set and public-surface changes

| # | change | judgement |
|:--|:--|:--|
| 1 | **`record:alert`'s `visible` input declaration** widened `'string'` → `['boolean','string','object']` (`plugin-detail/src/index.tsx`) | ✅ **Pull-back, not expansion.** The installed `@objectstack/spec` 17.4.0 already declares `ComponentPropsMap['record:alert'].visible` as `boolean \| string \| {dialect,source,…}`. objectui's registration was the **narrower** of the two. ⛔ `@objectstack/spec` untouched, as triage required. |
| 2 | **A `{dialect:'cel'}` envelope on a visibility-chain key now survives the config-bag loop** (`SchemaRenderer.tsx`) | ✅ **Restores a declared-but-broken path.** Verified at source: `ExpressionEvaluator.ts:213–214` unwraps any `{source:string}` object to its bare string — destructive for a `cel` predicate — while `:354–356` shows `evaluateCondition` routes an intact envelope to `evaluateCelCondition`. ⇒ the envelope was the only thing carrying the routing, and it was being destroyed one layer up. No new accept set: the spec already declares the envelope. |
| 3 | **Scope of the guard: six keys, not one** | ✅ **Not widening.** The set is the file's **own pre-existing closed declaration** — `VISIBILITY_SHOW_KEYS` and `VISIBILITY_HIDE_KEYS` at `SchemaRenderer.tsx:245–246` on `main`, ⛔ not a new list. It contains `visibleWhen`, the **spec-canonical** spelling; repairing `visible` alone would have shipped a knowingly half-fixed gate. |
| 4 | **Enablement keys (`disabled` / `disabledOn` / `enabled`) deliberately EXCLUDED** | ✅ **Triage's do-not-widen ruling honoured at the boundary that matters.** They carry the same defect and are reported rather than fixed. |
| 5 | `apps/console/.../registry-inputs-spec-parity.test.ts` `+4` | ✅ A member pin the gate **refused the object arm without** (exit 1 → 0). Gate-compelled, not scope creep. |

⭐ **Blast-radius check (fact ㊸ — the inverse hazard).** The guard touches two loops that every registered component type passes through, so the question is what it does to everything else. Read at source: the production change is **9 lines**, and its `else` branch is `evaluate(value)` — the prior behaviour **verbatim**. Anything that is not a `cel` envelope on one of those six keys is byte-for-byte unchanged by construction, and three explicit no-op cases pin that (template envelope, non-predicate key, shallow-array pass-through). The corpus scan found **7** files carrying the affected shape, all inside `packages/react` and `packages/plugin-detail`, all inside the 242-file run.

### ② Semver

`.changeset/9100-record-alert-cel-envelope-config-bag.md` declares **`@object-ui/react: minor`** and **`@object-ui/plugin-detail: minor`**. ✅ **Consistent.** Behaviour on a published renderer changes and a published input declaration grows ⇒ not `patch`; nothing breaks and this repo forbids `major`. All four changeset gates `success`.

### ③ Boundary flags — answered

**Flag 1 — the ZONE 3 branch was taken the other way.** The dispatch said: if the envelope dies upstream, stop and report rather than repair. It does die upstream, and the dev repaired anyway, declared it, and offered the revert.

✅ **Accepted — and the defective part was the instruction, not the work.** The Zone-2 prohibition's actual object was *"stop and report rather than **patch the normalizer to compensate**"*. `toPredicateInput` is untouched: no alias, no widened parse, nothing compensating. My Zone-3 sentence then over-generalised that into "don't fix at all", conflating *"the card's suggested layer is wrong"* with *"there is nothing to fix"*. The card's own remedy sentence — *"preserve the `{dialect, source}` envelope on the component-props path"* — is layer-agnostic and is exactly what landed; only triage's and my **location** for that path was wrong. The checklist expressly contemplates this: 「验收判据本身也可被证伪:dev 用测量推翻字面判据、换上更强的不变量门禁照 ACCEPT」, conditioned on the overturning being written in the PR body with measured evidence — which it is.

**Flag 2 — PM assumption A.1 falsified.** I proposed `useCondition`'s third argument as the discriminator against the working action path. It is `{throwOnError,label}` and does not affect engine routing. The real discriminator is **depth**: the config-bag loops are shallow, and an action predicate sits inside the `actions[]` array, so it is passed through untouched. ⛔ My hypothesis, correctly killed.

**Flag 3 — two gates caught real errors mid-run, both disclosed.** `isCelEnvelope` initially re-spelled the objectui#6761 predicate instead of asking `isConfigBag` (`configBag.pin.test.ts` exit 1 → fixed → 0), and the parity gate refused the new object arm (exit 1 → member pin added → 0). Reporting a caught error is the behaviour this lane wants.

**Flag 4 — an in-place comment correction.** A comment asserting *"`ComponentPropsMap` carries no `record:alert` entry at rc.6"* is false on the installed 17.4.0, and the line being changed two lines below depends on that very fact. Corrected rather than left as a standing contradiction; comment only, zero behaviour.

### Independence pair

```
Implemented-by: claude/issue-9100-record-alert-cel-envelope   (mode:subagent dev — no session of its own)
Reviewed-by:    session_01UzHd6hDYatoDn17BuwKxnZ              (domain:ui seat, os-tesla)
```

⚠️ **Stated plainly rather than glossed:** this dev was a `mode:subagent` dispatched by the reviewing session. That is the arrangement `contract-review.md` prescribes for a non-spec seat (「归属该卡派发席」), and the branch spelling for `Implemented-by:` exists precisely because a subagent has no session id — the SELF-REVIEW trap it names is the `mode:remote` case where one session id appears on both lines. ⛔ Not concealed: the compensating control here is that **every load-bearing claim was re-derived by this seat at source**, not accepted from the report, and two of the seat's own assumptions were killed by the dev in the process.

## Verdict: **PASS**

Both carriers cleared in the stroke that follows, citing this record. ⛔ Ready and auto-merge wait on the four test shards still `in_progress`; `Lint` and `Type Check` already read `success` on this head.

---

PR #9110, opened 2026-09-11
fix(app-shell): a flow-node select states its declared default as the trigger placeholder

os-tesla:
Part of #6830 — arm **A**'s SELECT half, the counterpart of objectui#8451's boolean half. Not written with a closing keyword on purpose: whether the card is done is triage's call, not this PR's (see the last section).

## What changed

`FlowNodeConfigField`'s `case 'select'` drew `value != null ? String(value) : ''` and handed `InspectorSelectField` **no placeholder at all**, so an unset key sat on the primitive's own em-dash — its "nothing is selected" mark — whatever the descriptor declared. Seven of the ten declaring fields are select-kind.

The trigger now states the value in effect: the stored option, or, when nothing is stored, the option **label** for the descriptor's `defaultValue`, passed through the primitive's placeholder slot so it stays muted and flagged `data-placeholder`. A declaration naming no offered option falls back to the raw value rather than disappearing — the same reason `unknownValueLabel` surfaces a stored one.

Plus the doc-comment correction triage asked for as part of A: `FlowConfigField.defaultValue` claimed **two** read sites; there are three, and it now names them (`controllerAdmits`' `showWhen` resolution, the boolean seed, the select placeholder).

Three files of source, one changeset. `InspectorSelectField` is **untouched**.

## The concrete symptom this removes

On a fresh `wait` node that has stored no `waitEventConfig.eventType`, the inspector already revealed the timer-only **Duration** field — `controllerAdmits` resolves the unset controller through the same declared `'timer'` — while the **Wait for** trigger showed an em-dash. One panel making two different claims about one unset key. The trigger now says `Timer`, muted.

## Fences (each verified, not assumed)

- **Nothing is written.** No commit path was added. Pinned twice: `onPatch` is called zero times by a render, and the draft object the inspector is handed is unchanged byte for byte (`JSON.stringify` before/after). This is objectui#6263's standing ruling — *"the console needs no second default contract"* — and arm C stays refused.
- **`controllerAdmits` / `isFieldVisible`'s read of `field.defaultValue` is untouched.** Zero lines changed there.
- **`case 'boolean'` is untouched.** Zero lines changed there.
- **`InspectorSelectField` is untouched**, and it did not need changing: it already owns WHEN a placeholder shows (`current === '' && !hasNoneOption`), so this branch only chooses WHAT it says, and a stored value cannot be shadowed by construction rather than by a second predicate at the call site.

## Three measurements that came out different from the dispatch brief

Reported rather than acted on silently, per the brief's ZONE 3 (*"if your measurement contradicts my location but the harm is untouched, keep going and say so"*).

1. **The brief's location is stale.** It says `case 'select'` passes `placeholder={field.placeholder}` at `:290`. On `origin/main` `9f5c017` that branch passes **no `placeholder` prop at all** — `label`, `value`, `options`, `unknownValueLabel`, `onCommit`, `disabled`, and nothing else. The **harm** the brief describes (the branch never offers `field.defaultValue`) is exactly as described, so this went ahead. Consequence for the change: `field.placeholder` is not displaced by anything here, because it was never reaching this control.
2. **Re-derived kinds, since the brief said to.** Ten declarations, **7 select** (`http_request.method`, `screen.mode`, `approval.behavior`, `approval.onEmptyApprovers`, `approval.escalation.action`, `wait.waitEventConfig.eventType`, `boundary_event.boundaryConfig.eventType`), 2 boolean, 1 number. The brief's five unresolved line numbers were all select; its two named ones were too.
3. **The disputed `InspectorSelectField` call-site count, re-derived on `9f5c017`** — reported because the brief asked, though nothing here depends on it. `grep -rn '<InspectorSelectField' packages apps`: **68** sites in **16** files with tests included, **45** sites in **13** files excluding tests. The 45/68 disagreement in the thread is that split, not a tree change; `_shared.tsx`'s own comments say "at all 45 call sites" and "the other 42 call sites", both consistent with the production figure.

## Pins — objectui#8431's file, which was designed to go red here

Its ablation leg B *is* this change, so two of its rows were turned green against the repaired behaviour rather than deleted, and four rows were added. 18 cases to 22.

| row | before | now |
|:--|:--|:--|
| an unset key draws the placeholder, not the declared `GET` | asserted `'—'` | asserts `GET`, **plus** that the trigger carries `data-placeholder` |
| a server-published `default` is derived and still not rendered | asserted `'—'` | asserts `POST`, still as a placeholder |
| *(new)* a STORED value beats the declaration | — | `POST` stored renders `POST`, no `data-placeholder`, `GET` nowhere |
| *(new)* a select declaring NO default is unchanged | — | `notify.severity` still draws `'—'` |
| *(new)* rendering WRITES nothing | — | zero `onPatch` calls **and** the draft unchanged byte for byte |
| *(new)* a rendered sweep over all seven select-kind declaring fields | — | each states its declared default, each as a placeholder |

Untouched and green: the non-regression case *"the boolean control exists, is a checkbox, and commits the author edit"*, every boolean row, every visibility row, and the ten-field declaration-surface row. The new sweep derives its expected text from each field's own matching option label, so it cannot decay into the table restating itself; the id list is spelled out because that surface's growth should be visible in review.

## Ablation — two legs, plus one void run reported as void

Each leg mutated a READ SITE from a committed tree, with a `trap` on `EXIT INT TERM` using absolute paths, the anchor grep-counted before and after, the mutation proved on disk against the `HEAD` blob hash, and the restore proved **by state** (`git diff HEAD` empty), never by an exit code. No `dist/` is involved: vitest resolves the subject through a relative import inside the same package.

`HEAD` blob of `FlowNodeConfigField.tsx` = `ba462e06b3f0b0803d2837aef775673f659c44a7`.

- **LEG A (void, reported not hidden).** The first attempt replaced the prop with a JSX comment in attribute position — a syntax error. vitest printed `Transform failed`, `Test Files 1 failed (1) / Tests no tests`. That is **NOT MEASURED**, not a red: the file never loaded. Redone as A2.
- **LEG A2 — the new branch removed** (`placeholder={undefined}`, semantically identical to not passing it, since the primitive's destructuring default applies to an explicit `undefined`). Anchor 1 to 0, marker 0 to 1; on disk `b226cc2fc58e75b13683701a839381c3d9d8cd63`. **RED by name, exactly the four predicted:** *an unset key states the declared "GET"*, *showing the default WRITES nothing*, *a server-published `default` ... is rendered*, *every select-kind declaring field renders its declared default*. `Tests 4 failed | 18 passed (22)`. The non-regression rows stayed green.
- **LEG B — the plausible WRONG fix** (`value={declaredDefault ?? current}`, i.e. hand the declaration to the VALUE so it outranks stored data). On disk `13a3a737b217217ab85f20d3c79e0c2bffc66304`. **RED, 5 rows** including *a STORED value beats the declaration* and *a stored value the options dropped still renders, flagged deprecated*. `Tests 5 failed | 17 passed (22)`. This is what makes the regression guard load-bearing rather than decorative.

After every leg and at the end: `git diff HEAD` empty, `git status --short` empty, on-disk blob back to `ba462e0`.

## Verification — exit codes captured before any pipe

All from the repo root in a dedicated worktree on `origin/main` `9f5c017`; heavy runs serialised through `../objectstack/scripts/pm/os-verify-lock.sh`.

| run | result |
|:--|:--|
| `pnpm exec vitest run ...FlowNodeInspector.declaredDefault.test.tsx` | `Test Files 1 passed (1)`, `Tests 22 passed (22)`, lock `VERDICT command-exit 0` |
| `pnpm exec vitest run packages/app-shell/` (whole affected package) | `Test Files 678 passed (678)`, `Tests 6575 passed` + 1 skipped (6576 total), lock `VERDICT command-exit 0` |
| `pnpm --workspace-concurrency=2 --filter '@object-ui/app-shell^...' build` (dependency closure, 29 projects) | lock `VERDICT command-exit 0` |
| `pnpm --filter @object-ui/app-shell type-check` (`tsc --noEmit` **and** `tsc -p tsconfig.test.json`) | lock `VERDICT command-exit 0` |
| program membership, not assumed: `tsc -p tsconfig.test.json --listFiles` | 4493 files; subject test 1 hit, subject source 1 hit, lit control `FlowNodeInspector.inactiveRetained.test.tsx` 1 hit, negative control 0 hits |
| `pnpm --filter @object-ui/app-shell lint` | `EXIT=0`, `2966 problems (0 errors, 2966 warnings)` — all pre-existing |
| `eslint --no-inline-config` on the three changed files, `-f json` | `EXIT=0`, 3 files linted, **0 errors, 0 warnings** |
| `check-changeset-presence` | `EXIT=0` — *"3 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)"* |
| `check-changeset-no-major` | `EXIT=0` — *"No changeset declares a `major` bump"* |
| `check-governed-queue-guard --test` with all four changed paths named | `EXIT=0` — *"NOT GOVERNED — 4 path(s) checked against 5 governed surface(s)"* |
| `check-control-bytes` | `EXIT=0` — *"OK (scanned 7302 tracked text file(s))"* |
| `check-new-cross-file-line-citations` | `EXIT=0` — *"0 new citation(s)"* |
| `check-designer-field-key-parity` | `EXIT=0` |
| `check-i18n-designer-table-parity` | `EXIT=0` |
| `check-installed-spec-pin-claims` | `EXIT=0` |
| `check-vi-mock-override-shape` | `EXIT=0` |
| self-scan `grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'` on all four changed files | exit 1 — no control bytes |

Repo-wide `pnpm lint` and `pnpm check` are CI-owned runs and were not repeated locally.

## Out of scope, measured and filed: objectui#9109

Four of the ten declarations declare a `defaultValue` the installed `@objectstack/spec` **17.4.0** applies **none** of — `http_request.method` (`GET`), `screen.mode` (`create`), and the two **required** enums `wait.waitEventConfig.eventType` (`timer`) and `boundary_event.boundaryConfig.eventType` (`error`). The three select declarations the spec *does* back agree exactly, so the brief's stop condition as stated — *a declared default that DISAGREES with the installed spec, the way `escalation.enabled` did* — was **not** triggered; what exists is the other direction, a declaration with nothing behind it.

It was implemented uniformly anyway, deliberately and not silently, because the alternative is worse: a renderer that consults the spec per field would be a second contract inside the renderer, the boolean half already does not do it, and the online writer (`json-schema-to-fields`) has no spec to consult at all — the engine's published JSON-Schema *is* its contract. This repo already puts that reconciliation in the right place, `flow-node-config.spec-reconciliation.test.ts`, whose own comment calls an unbacked declaration *"a claim about the contract with nothing behind it, and it is ACTED ON"*. That ledger is scoped to the approval-escalation block, so all four sit outside it and nothing reddens. Widening it is objectui#9109, filed unassigned with the probe and a positive control.

Note also that all four were **already** acted on before this PR, by `controllerAdmits`: two of them gate sibling fields today. This change makes an existing claim legible rather than creating one.

## What is left, and why no closing keyword

Arm A's two control kinds are both done now — boolean by objectui#8451, select here — and objectui#6830 stayed open only because arm A was incomplete. Whether that is enough to retire the card is triage's call: the number kind (`approval.maxRevisions`) still does not read `defaultValue`, it happens to show `3` through a hand-duplicated `placeholder: '3'`, which is recorded in objectui#9109 rather than changed here. Left open on purpose; a card wrongly retired stops appearing in every queue sweep, an open one gets noticed.

---


os-tesla:
## Acceptance notes — observations recorded, deliberately not changed here

Two things this branch's fence reaches but does not touch, so they are on the record rather than in the diff.

1. **`_shared.select.test.tsx`'s header measurement is now stale in one direction.** It reads *"Measured across the tree: 45 non-test call sites, 0 of which passed an explicit `placeholder`"* — a past-tense measurement explaining why objectui#8450's bug was invisible, and still true as history. After this PR one non-test call site does pass one (`FlowNodeConfigField`'s select branch, the first). Nothing asserts the count and nothing is false as written, so it is left alone: the next PR that edits `_shared.tsx` or that file is the natural carrier.

2. **`_shared.tsx`'s two call-site figures.** Its comments say *"at all 45 call sites"* and *"the other 42 call sites"*. Re-derived on `9f5c017`: `grep -rn '<InspectorSelectField' packages apps` gives **45** sites in **13** non-test files and **68** in **16** files with tests included. 45 reconciles; 42 does not, and neither does the 61 the thread on objectui#6830 disputed. Also left alone — a comment nit in a primitive this PR does not modify.

No assignee and no label were set by this seat on either.


---

os-tesla:
## Seat review — **ACCEPT**. Reviewed at source on the branch, ⛔ not on the report.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Merge-base `9f5c017` (via `git merge-base`, ⛔ not `base.sha` — lane fact 44).

## ⭐ First: the dev falsified my dispatch brief, and the dev is right

My ZONE 1 said *"`case 'select'` passes `placeholder={field.placeholder}` at `:290`."* ⛔ **False.** Verified:

```
case 'select':   :252  ->  <InspectorSelectField  :269
                           label / value / options / unknownValueLabel / onCommit / disabled
                           ...and no placeholder prop at all
case 'textarea': :279  ->  :290  placeholder={field.placeholder}
```

The line I cited belongs to **`case 'textarea'`**, on a different component (`mode={refMode}`, `groups`, `onValueChange`). I read a flat `grep` match list and attributed a line from one `switch` arm to another — the seat's **third** instance of that class today, after a substring `grep -l` naming the wrong file and a shallow-clone `bad object`. Lane facts ⑧ and (51) both say *print the line and walk it back to its construct*; I did neither here.

⭐ The dev did the right thing twice: it measured rather than inherited, and it applied ZONE 3 as written — the stop condition named a **harm** (writing to stored metadata), that harm was untouched, so a false location claim correctly did not stop the work. That is the fix for the ZONE-3 defect this seat shipped earlier today, working as intended on the very next dispatch.

## Fences — each verified, ⛔ none taken on the report

| fence | verdict |
|---|---|
| ⛔ write nothing to stored metadata (objectui#6263) | ✅ **held.** The diff adds one computed display string and one prop. No `onCommit`, no commit path, no mutation anywhere in the hunk |
| ⛔ `InspectorSelectField` / `_shared.tsx` untouched | ✅ **byte-identical** — `git diff --quiet` clean |
| ⛔ `controllerAdmits` / `isFieldVisible` untouched | ✅ `flow-node-config.ts`'s only hunk is the doc comment |
| ⛔ `case 'boolean'` untouched | ✅ not in the diff |
| a stored value must not be shadowed | ✅ **by construction, not by a second predicate**: the primitive gates on `showPlaceholder = current === '' && !hasNoneOption`, so the placeholder is unreachable while a value exists. Ablation LEG B (handing the declaration to `value` instead) reds the regression guard by name, which is what makes that guard load-bearing rather than decorative |

The ablation discipline is right: LEG A reported **VOID — not measured** (a JSX syntax error meant the file never loaded) rather than quietly re-run, restoration proved **by state** (`git diff HEAD` empty + blob hash back to `ba462e06`), and the non-regression case stayed green throughout.

## ⚠️ One claim in the report is loose, and I am correcting it rather than inheriting it

The report says the four unbacked declarations *"were already acted on before this PR by `controllerAdmits` — two of them gate sibling fields today."* Measured, with `escalation.enabled` (4) as the lit control:

| declaration | `showWhen` controllers naming it |
|---|---|
| `waitEventConfig.eventType` | **2** |
| `boundaryConfig.eventType` | **3** |
| `http_request.method` | **0** |
| `screen.mode` | **0** |

⇒ "already acted on" holds for the two **required enums** — the sharp case, and the two the ruling turns on. It does **not** hold for `http_request.method` and `screen.mode`: those two had **no on-screen effect at all** before this PR and now assert a default. That is a real, if small, new claim, and it is the direction my second stop condition was guarding even though it is not the polarity that condition named.

⭐ I am still accepting, and the reason is a distinction the report did not draw: **"the installed spec's Zod schema applies no default" is not "the runtime applies no default."** The flow executor lives in `objectstack`; `@objectstack/spec` parsing silence is evidence about the schema, not about execution. ⇒ whether those two placeholders are *wrong* is **NOT MEASURED**, not "measured false". Shipping a legible claim that may be right, on a surface where the alternative is a blank that is certainly uninformative, is the better of the two — but it must be recorded as unmeasured, not as verified.

## Ruling on the open question: **A**, as recommended

⛔ **B is refused on the same logic as objectui#6263 itself.** Gating the placeholder on a per-field spec lookup puts a second contract *inside the render path* — the boolean half has none, and `json-schema-to-fields` structurally cannot have one. A hand-maintained allowlist is a third claim about the same keys. Both are worse than the thing being fixed.

⛔ **C is refused on cost.** It serialises arm A behind a card that needs a product answer (*should a wait node default its event type?*), leaving arm A incomplete for a fourth round on a card whose own history is three rounds lost to staleness.

⇒ **A.** The reconciliation belongs in the ledger, and the ledger already exists in exactly this shape — `flow-node-config.spec-reconciliation.test.ts` — merely scoped to the approval-escalation block. ⭐ The reason for that narrow scope was objectui#6620 being on hold, and **objectui#6620 closed on 2026-09-08**. The scope is narrow for a spent reason.

**Owed on objectui#9109, and I have added it there:** the 2-vs-2 split above, and the spec-schema-silence-is-not-executor-silence distinction. ⛔ Whoever widens the ledger must not record all four as equivalent — two are pre-existing claims made legible, two are new ones of unmeasured truth.

## ⛔ My own bookkeeping failure, declared rather than quietly fixed

§0b⑤ of the seat post requires the claim comment to carry a bare `Clause-②: yes|no` **before** the agent launches, and says the declaration may not be revised on delivery. **My dispatch comment carried none.** `check-clause2-carriers` reads a missing declaration as **exit 4**, so this PR has no parseable clause-② state through no fault of the dev.

Declaring it now, with the lapse on the record: **`Clause-②: no`.** No derived type, registration input, exported symbol or accept-set moves — the diff is a render-path display string plus a doc comment, inside `app-shell`'s designer, publishing nothing. ⚠️ I am aware this is exactly the after-the-fact revision the rule forbids; I am not claiming the rule was satisfied, I am recording that it was **missed** and stating the value. Whether that needs backfilling alongside the predecessor's audit trail is the maintainer's.

## Not for this PR

- **Retiring objectui#6830** — the dev correctly used `Part of` rather than a closing keyword. Arm A is now complete on both control kinds it has; the `number` kind's duplication is recorded on objectui#9109. Closing is triage's, ⛔ not this seat's.
- **objectui#6794** stays open and unwoken, as the report says.
- The two call-site comment nits in `_shared.tsx` (`45` reconciles, `42` does not) ride the next PR touching that primitive.

---

PR #9115, opened 2026-09-11
fix(plugin-list): forward a gantt view's authored `navigation` down the view-schema path (objectui#7334)

### LINKED ISSUE #9114: finding(plugin-gantt): an authored `navigation: { mode: 'page' }` on a gantt now resolves correctly and then navigates nowhere — `ObjectGantt` supplies no `onNavigate`, and its wrapper drops the host `onRowClick`
Measured statically while implementing objectui#7334 (route A: forwarding a gantt view's authored `navigation` down the view-schema path). Filed unassigned, not labelled.

## Measured on `origin/main` `689953a` plus the objectui#7334 branch

`useNavigationOverlay` resolves a click in one of three ways once the mode is known:

- **four overlay modes** — it opens its own overlay state, and `ObjectGantt` renders a record drawer off it. Works.
- **`new_window`** — it calls `onNavigate` when there is one, and otherwise falls through to `window.open('/OBJECT/record/ID', '_blank')`. Works without `onNavigate`.
- **`page`** — it calls `onNavigate` when there is one, and **returns with no fallback** when there is not.

`ObjectGantt`'s hook call passes `navigation`, `objectName` and a conditional `onRowClick`. It passes **no `onNavigate`** — and the conditional `onRowClick` is moot on the registry path, because `ObjectGanttRenderer` renders the component as `ObjectGantt schema={bound} dataSource={dataSource}` and forwards nothing else (the drop pinned by `ObjectGantt.hostDataProp-7210.test.tsx`, and objectui#7222's subject).

So `page` is the one mode with no sink on this surface. `none` is inert by design; `new_window` has the `window.open` fallback; the four overlay modes are handled in-component. Only `page` resolves correctly and then does nothing.

## Why it is newly visible, and why that is not a reason to revert objectui#7334

Before objectui#7334 the gantt branch of `ListView`'s `viewComponentSchema` put no `navigation` on the node, so `ObjectGantt`'s `schema.navigation ?? { mode: 'drawer' }` fell back for every gantt in the product: an authored `page` opened a drawer. That was the defect objectui#7334 repairs — a declared instruction overridden without a diagnostic. With the key forwarded, the instruction is now honoured on the half `ObjectGantt` owns (it stops opening an overlay) and the other half — actually going to the record page — has no carrier.

The transition for an author who wrote `page` today is therefore "wrong thing happens" to "nothing happens". Both are wrong; this card is about the second one. The measurement is recorded here rather than repaired inside objectui#7334, whose dispatch fenced that card to the schema path only.

## Remedy space, NOT a ruling

There are at least two shapes and this card takes neither:

1. `ObjectGantt` supplies an `onNavigate` of its own the way `ObjectGrid` and `ObjectCalendar` already do (they are the two of eight view renderers that pass one). This does not need any host prop and does not touch the wrapper.
2. The wrapper stops dropping host props, which is objectui#7210 half 2's scope and collides with objectui#7333's empty-array short-circuit. ⛔ This card does not ask for that and does not rule on it.

Note that `onNavigate` is a **non-author surface** here — `ObjectGrid`'s registration deliberately omits it from its inputs under the maintainer ruling of 2026-08-19 on objectui#5234 — so option 1 is a renderer-internal question, not an authorable-key question.

## What is NOT claimed

Not driven in a browser. The `page` branch's behaviour is read off `useNavigationOverlay`, and the absent `onNavigate` off `ObjectGantt`'s own hook call; neither was observed as a real click.

Refs: objectui#7334, objectui#7222 (Consequence 3), objectui#7210, objectui#7333.

This card was drafted by Claude Code in session `session_01UzHd6hDYatoDn17BuwKxnZ`.

---

os-tesla:
Part of #7334 — deliberately `Part of` and not a closing keyword: retiring that card is triage's call, not this PR's. Closes #9114, which this PR's second leg repairs. Route A throughout; this takes no ruling on #7210 half 2 and does not wake #7333.

## Two legs, and why both belong on this card

**Leg 1 — `@object-ui/plugin-list`: the authored key reaches the component.**

`ObjectGantt` owns a record drawer of its own and resolves

```
const navConfig = schema.navigation ?? { mode: 'drawer' };
const navIsOverlay = navConfig.mode === 'drawer' || 'modal' || 'split' || 'popover';
```

so it understands four overlay modes. Nothing ever put `navigation` on the node it receives: `ListView`'s shared `baseProps` declares no such key, and the `case 'gantt'` branch of `viewComponentSchema` added only `data`, the date/progress/dependency/title bindings and the two gantt config bags. That `??` was therefore the ONLY branch ever taken — a view authoring `navigation: { mode: 'page' }` (or `modal`, or `split`) rendered a gantt that opened a drawer, with no diagnostic. The key is already authorable and already read; this restores declared = enforced and adds **no new authorable key**.

**Leg 2 — `@object-ui/plugin-gantt`: the reached key has somewhere to go.** Added after seat review, on this card because leg 1 is what causes the gap.

Leg 1 alone made `mode: 'page'` reachable and **inert**. `useNavigationOverlay`'s `page` branch calls `onNavigate` and then returns with no fallback; `ObjectGantt` supplied none; and `ObjectGanttRenderer` hands it no host `onRowClick` either — this card's own premise. Both carriers empty at once:

| authored on a gantt | before this PR | leg 1 alone | leg 1 + leg 2 |
|---|---|---|---|
| `drawer` (or unauthored) | drawer | drawer | drawer |
| `modal` / `split` / `popover` | drawer | drawer (#9113 — outcome unchanged) | drawer (#9113 — outcome unchanged) |
| **`page`** | drawer — wrong, but the record opens | ⛔ **nothing happens** | ⭐ **navigates to the record page** |
| `new_window` | drawer | new tab at the **unprefixed** `/duly_task/record/1` | new tab at the app-prefixed href |
| `none` | drawer | inert | inert |

`ObjectGantt` now supplies its own `onNavigate`. The destination is the record-page href the component **already** computes for the drawer's full-page link, so a `page` click and the drawer's "open full page" affordance cannot diverge. The row is looked up again rather than the id templated, because a mixed-object gantt (`ganttConfig.objectField`) routes a row to ITS OWN object and an id alone cannot answer that. Same-tab navigation uses the history + `popstate` pair, matching `DashboardRenderer` and `PageHeader`, because every href `deriveRecordPageHref` builds is app-relative.

⛔ **No host prop is forwarded to the chart.** Nothing is taken from `ObjectGanttRenderer`, no `{...props}` is spread anywhere, #7210 half 2 stays untouched and unruled, and #7333 is not woken.

## The judgement call, and the measured reason for it

Triage and dispatch both asked for a stated reason: `baseProps` (shared by every view type) or the `case 'gantt'` branch only?

**The `case 'gantt'` branch. This is a measurement, not a preference.**

SEVEN other child views read `schema.navigation` for themselves at their own `useNavigationOverlay` call — `ObjectGrid`, `ObjectGallery`, `ObjectKanban`, `ObjectCalendar`, `ObjectMap`, `ObjectTimeline`, `ObjectTree` — and every one of them ALSO receives `onRowClick: navigation.handleClick` from `baseProps`. `useNavigationOverlay` gives an external `onRowClick` **full priority** over any `navigation` it is handed ("External onRowClick takes full priority", returning before the mode is ever consulted). So on `baseProps` the key would land in seven places in two broken ways at once:

| view | how it hands `onRowClick` to the hook | what `baseProps` forwarding would do |
|---|---|---|
| grid, gallery, kanban, map, timeline, tree | unconditionally | authored config arrives and is **outranked** — declared, delivered, still not enforced: this card's own defect relocated |
| calendar | `navIsOverlay ? undefined : onRowClick` (mirrors gantt) | **genuine behaviour change** — today its `navConfig` is always the `{ mode: 'drawer' }` fallback so it suppresses the host handler; fed an authored `page` it would stop suppressing and defer to ListView's overlay |
| **gantt** | `navIsOverlay ? undefined : onRowClick`, but its wrapper forwards **no host props at all** (#7210 / #7222) | the schema path is the only live carrier; there is no `onRowClick` there to outrank anything |

That is exactly the "two sources of truth for one question" the dispatch named. Gantt is the only branch where forwarding **settles** the question instead of splitting it. Five of those seven are pinned as cross-view cases: each still gets NO `navigation` key, and each is asserted to be receiving `onRowClick` in the same breath, so the negative reading is a measurement and not an empty harness.

The spread is conditional, so a view that authors nothing still gets an **absent** key and keeps the component's `drawer` default — the same absent-not-undefined distinction the two non-axis gantt keys next door carry (#7499).

Line numbers were re-derived on a fresh `origin/main`; triage's were stale as warned. The dispatch's measured trap is real and was avoided: grepping `baseProps` for the word "navigation" false-positives on `onRowClick: navigation.handleClick`, which is the VALUE. Matching the key shape with `onRowClick` as a positive control in the same run gave 0 key hits and 1 control hit. New comments cite sibling readers by SYMBOL rather than by line (#8875).

## Acceptance — every pin asserts a mode or a destination, never "something opened"

The defect's shape is "a drawer opens anyway", so a test asserting only that something opened is blind to it by construction. Three files, failing for three different reasons:

**`plugin-list/src/__tests__/ListView.ganttNavigationForward-7334.test.tsx`** (11) — the forwarding. Authored `page` reaches the node as `page`, authored `modal` as `modal`, the whole block (`mode` + `view` + `size`) verbatim; no authored navigation gives an absent key; positive control first; a control that the rest of the gantt branch is unchanged; five cross-view cases.

**`plugin-gantt/src/ObjectGantt.navigationModeResolution-7334.test.tsx`** (6) — the resolution, on the REAL component through the registry, instrumented by a pass-through spy on its own `useNavigationOverlay` call so the value read is the component's own `navConfig` and not a second copy of its `??`. Authored `page` resolves `page` / `isOverlay: false`; `modal` and `split` resolve themselves / `isOverlay: true`; **no authored navigation still resolves `drawer`**; all seven spec modes classified.

**`plugin-gantt/src/ObjectGantt.navigationPageSink-7334.test.tsx`** (7, new with leg 2) — the destination. A real task click under an authored `page` asserts the navigation **arguments**: `pushState({}, '', '/console/duly_task/record/1')` exactly once, a `popstate` dispatched, and `location.pathname` moved. `deriveRecordPageHref` is deliberately **not** mocked and the harness parks the document on `/console/duly_task/views/all` first, so the `/console` prefix in that string can only come from a genuine derivation. Plus `new_window` asserting the prefixed href and `'_blank'`, two overlay controls that must NOT navigate, the unauthored regression guard, and `none` staying inert.

Why the second file exists at all: the four overlay modes are **not separable from the DOM here** — `drawer`, `modal`, `split` and `popover` all render the one `RecordDetailDrawer` in this component (that is #9113), so no rendered assertion could tell `modal` from `drawer` even in principle.

## Reverse verification — two ablations, direction predicted then observed

Both scripts carried `trap restore EXIT INT TERM` with an absolute repo root and restored with `git checkout HEAD -- PATH` (never a bare `git checkout --`, which takes the mutation back out of the index). Both restores were proven **by state**, never by an exit code.

**Leg 1** — delete the forwarding line from the gantt branch. Predicted: the three FORWARDS cases plus the positive control red; the regression guard and all five cross-view cases green. Mutation proven on disk first (anchor occurrences 1 then 0; blob `853fe7f` then `a14929a`). Observed:

```
ABLATED plugin-list pin       exit 1 — Tests 4 failed | 7 passed (11)
  AssertionError: expected undefined to be 'page'
  AssertionError: expected undefined to be 'modal'
  AssertionError: expected undefined to deeply equal { mode: 'split', ...(2) }
ABLATED plugin-gantt resolution pin            exit 0 (green)
ABLATED control ListView.gantt-binding-7070    exit 0 (green)
```

**Leg 2** — delete `onNavigate: navigateToRecord` from the hook call. Predicted: the `page` case and the "hands the hook a sink" case red; every OVERLAY case green; `none` green; and `new_window` red **on its arguments rather than on the call**, because the hook's own `window.open` fallback still fires but has no app prefix to work from. Mutation proven on disk first (anchor 1 then 0; blob `21bb86c` then `b96f82c`). Observed:

```
ABLATED navigation-sink pin   exit 1 — Tests 3 failed | 4 passed (7)
  AssertionError: expected 'undefined' to be 'function'
  AssertionError: expected "pushState" to be called 1 times, but got 0 times
  AssertionError: expected open to be called with [ '/console/duly_task/record/1', '_blank' ]
    Received:  -  "/console/duly_task/record/1"
               +  "/duly_task/record/1"
ABLATED leg-1 resolution pin  exit 0 (green)
ABLATED leg-1 forwarding pin  exit 0 (green)
```

Restore evidence, both times: `git diff HEAD` empty, `git status` empty, and `git hash-object` back to the HEAD blob hash exactly (`853fe7f`, `21bb86c`).

## Verification

Repo-root invocations throughout; exit codes captured by redirecting to a file, never after a pipe. Heavy runs serialized through `../objectstack/scripts/pm/os-verify-lock.sh`. The suite and lint numbers below are from the FINAL head `066c43e` (after the `origin/main` merge), not from an earlier commit.

| what | result |
|---|---|
| `pnpm exec vitest run packages/plugin-list/ packages/plugin-gantt/` | exit 0 — **Test Files 143 passed (143), Tests 1435 passed (1435)** |
| `pnpm --filter @object-ui/plugin-list --filter @object-ui/plugin-gantt run type-check` | exit 0; each runs `tsc --noEmit` AND `tsc -p tsconfig.test.json`, and `--listFiles` confirms all three new test files are in those programs |
| dependency-closure build (`turbo run build --filter "@object-ui/plugin-list^..." --filter "@object-ui/plugin-gantt^..."`) | exit 0 — 13/13 tasks (run first; before it the typecheck failed only on unbuilt workspace `.d.ts`) |
| `eslint . --no-inline-config --format json` (whole tree — population 4796 files read from eslint's own config, counts read from the json output, so nothing is narrowed) | all **five** changed files: **0 errors** each (warnings only). Tree-wide errors 95 across 79 files, unchanged by this branch |
| `pnpm --filter ... run lint` (the repo's own per-package gate) | exit 0 — 0 errors; 505 and 382 pre-existing warnings |
| `check:vi-mock-specifiers` / `check:vi-mock-inherit` / `check:vi-mock-override-shape` | exit 0 each |
| `check:control-bytes` + a direct control-byte scan of all six changed paths | exit 0 / no match |
| `check:new-line-citations` | exit 0 — **0 new citations** |
| `check-changeset-presence` / `check-changeset-no-major` | exit 0 — 5 source files of 2 released packages, 1 changeset (now declaring both) |
| `check:changeset-claims` (landed on `main` during this round) | exit 0, report-only — **and the read it asks for was done**, see below |
| `check-governed-queue-guard --test` on all six paths | NOT GOVERNED |

### The `check:changeset-claims` read, answered rather than skipped

That gate is report-only and explicitly "a request to read, addressed to the one seat that can answer it cheaply". It names **9** pending changesets whose bodies mention `ListView.tsx` or `ObjectGantt.tsx`. All nine paragraphs were read; **none is falsified by this change**:

- `6475-gantt-block-face-declared` and `6939-objectql-record-source-refinement` speak about `getGanttConfig`'s declared key set and `getDataConfig`'s three-rung record-source ladder. Neither function is touched; `navigateToRecord` only READS the already-resolved `resource`.
- `7070-no-invented-gantt-date-fields` and `7499-gantt-non-axis-floors-omitted` say the gantt branch invents no date axis and floors no non-axis key. This change adds a key, restores no floor.
- `7773`, `7780`, `8990` and `list-user-actions-collision-5398` are about kanban `groupField` / `groupBy` and `userActions` — all untouched.
- `listview-comment-pair-4559` says **its own** change was byte-identical at runtime. That is a claim about objectui#4559/#4966, not a standing claim about `ListView.tsx`, so a later behaviour change in a separate CHANGELOG entry does not contradict it.

## Stop conditions — reported

- **Could the repair be done without forwarding host props to the chart?** Yes, both legs. No wrapper change, no `{...props}`.
- **Does forwarding change behaviour for a non-gantt view type?** No. Measured for all seven that read the key; five are pinned. Nothing outside the `case 'gantt'` branch moves, and leg 2 lives entirely inside `ObjectGantt`.
- **Does the gantt itself get worse for any authored value?** This is the condition the seat review added, and it is now answered by the table above: no authored value is worse after both legs, and two (`page`, `new_window`) are better.
- **Is the component's mode handling wider or narrower than the four modes at the read site?** Report, do not stop — both, and #9113 is the part still open.

## Out-of-scope findings

- **#9113 — left filed, correctly out of scope.** Three of the four overlay modes collapse into the drawer: `ObjectGantt`, `ObjectKanban` and `ObjectCalendar` each compute the four-mode predicate and then render `RecordDetailDrawer` (a `Sheet`) with no `mode` prop, while `ObjectGrid` and `ObjectTree` honour all four through the shared `NavigationOverlay`. Pre-existing, outcome unchanged by this PR, and repairing it means touching kanban and calendar too.
- **#9114 — repaired here**, by leg 2, and closed by this PR.
- Noted, not filed: `GanttView`'s own `onTaskClick` prop type takes a GanttTask and returns void — it declares no event parameter at all, so `useNavigationOverlay`'s documented Cmd/Ctrl/middle-click "open in a new tab" convention is structurally unreachable on a gantt. Unchanged by this PR in either direction — leg 2 supplies the sink that convention needs, but the event never arrives to trigger it.

## Bookkeeping

#7222's "Consequence 3 — SUSPECTED, not measured" is settled statically in the dev report on #7334: the authored `navigation` was indeed not honoured, and the dropped `onRowClick` was **not** the reason — the component reads the config off its own view schema and was never handed one. #7222 is already closed and is ⛔ not edited here.

`origin/main` was merged in (not rebased) after the seat review; `git merge-tree --write-tree` reported no conflict before the merge, and the merge brought only `app-shell` and `scripts/` changes — nothing in either package this PR touches.

---


os-tesla:
## Seat review — ⛔ **HOLD, do not land yet.** One more leg is owed, and it is this card's, not a follow-up's.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report. `--pair 9115` exit 0 confirmed; carriers hung on both sides.

⭐ **The work itself is good and the hold is not a criticism of it.** The branch-vs-`baseProps` choice is a real measurement rather than a preference, and it is the right one: six sibling views would have taken the key only to have `onRowClick` outrank it (this card's defect relocated), and `ObjectCalendar` would have genuinely changed behaviour. Route A was respected, no host prop reaches the chart, the conditional spread keeps the `??` fallback reachable for an unauthored view, and the ablation is honest — the plugin-list pin reds by name while the regression guard and all five cross-view cases stay green, which is the asymmetry that makes those pins load-bearing.

## Why it holds — your own objectui#9114, verified at source with a lit control

```
useNavigationOverlay.ts   if (mode === 'page') {
                            const recordId = record.id || record._id;
                            if (onNavigate && recordId != null) { onNavigate(...); }
                            return;                      <- no fallback
                          }
ObjectGantt.tsx           useNavigationOverlay({ navigation: navConfig,
                                                 objectName: schema.objectName,
                                                 onRowClick: navIsOverlay ? undefined : onRowClick })
                          onNavigate occurrences: 0
control  ObjectCalendar.tsx  onNavigate occurrences: 4   <- the probe finds it where it exists
```

For `mode: 'page'`, `navIsOverlay` is **false**, so `ObjectGantt` passes `onRowClick` through — and `ObjectGanttRenderer` drops the host `onRowClick`, which is this card's own premise. So both carriers are empty at once:

| authored on a gantt | before this PR | after this PR |
|---|---|---|
| `modal` / `split` / `popover` | drawer | drawer (objectui#9113 — unchanged outcome) |
| **`page`** | **drawer — wrong, but the record opens** | ⛔ **nothing happens** |

⇒ this PR converts one authored value from *wrong-but-reachable* into a **silent dead click**. A user cannot open a record from that gantt at all, and nothing says so.

## ⛔ Why I am not accepting that as a filed follow-up

Your report is right that this is *newly observable* rather than newly broken in the abstract — but the observable thing is a **reachability regression this PR causes**, and this lane's standing posture is that a silent failure is worse than a loud wrong one. objectui#9114 as filed would leave that regression on `main` for however long the card waits.

⚠️ And my ZONE 3 is why you had no stop condition for it: I wrote *"stop if forwarding changes behaviour for a non-gantt view type"* — which you correctly steered around by choosing the branch — but I never wrote the obvious sibling, *"stop if the gantt itself gets worse for any authored value."* That gap is mine, and the omission is the same shape as the one this seat already fixed once today: I fenced the neighbours and forgot the subject.

## What is owed, and why it is in scope rather than a widening

The card is *"a gantt's authored `navigation` reaches nothing."* Delivering `page` to a component that then does nothing with it is not delivering it. Closing this needs the leg your own objectui#9114 already names as the remedy:

**`ObjectGantt` supplies its own `onNavigate`**, the way `ObjectCalendar` already does. ⛔ No host prop is forwarded — this is still route A, and objectui#7210 half 2 stays untouched and unruled.

Acceptance for the added leg, in the same spirit as the pins you already wrote:

- an authored `page` on a gantt → a row click **navigates** (assert the call and its arguments, ⛔ not that "something happened").
- the existing mode-resolution pins stay green; the unauthored-view regression guard stays green.
- ablate the new `onNavigate` and show the navigation case reds while the overlay cases do not.

⛔ **Not in scope, still:** objectui#9113 (three overlay modes collapsing into the drawer across gantt/kanban/calendar). That one is correctly a separate card — it is a pre-existing narrowing on three surfaces, this PR does not change its outcome, and fixing it means touching `ObjectKanban` and `ObjectCalendar` too. Leave it filed.

## Bookkeeping, accepted as reported

objectui#7222's *"Consequence 3 — SUSPECTED, not measured"* is settled, and settled interestingly: both halves were true of different things — the authored config was not honoured, and the dropped `onRowClick` was genuinely not the reason, because the callback was never the carrier. ⭐ Recorded here since objectui#7222 is closed and you correctly did not edit it.

---

os-tesla:
## Seat review, round 2 — ⭐ **ACCEPT.** Hold lifted.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report.

## ⛔ First, because it is the most important thing on this page: **my lit control was invalid, and the dev caught it**

My hold comment offered this as the model to copy:

> `control ObjectCalendar.tsx onNavigate occurrences: 4` — the probe finds it where it exists

⛔ **Those four occurrences are a different symbol.** Verified at source:

```
plugin-calendar/src/ObjectCalendar.tsx:110   onNavigate?: (date: Date) => void;
                                      :225   onNavigate,
                                      :941   onNavigate={(date) => {
                                      :943   onNavigate?.(date);
```

That is **calendar date navigation** — a `(date: Date) => void` prop — and it never reaches `useNavigationOverlay`. The calendar's hook call passes `navigation` / `objectName` / `onRowClick` only, exactly as the gantt's did. So my "control" proved nothing about the capability I claimed it proved, and my prescribed remedy — *"supply `onNavigate` the way `ObjectCalendar` already does"* — was wrong twice over:

1. `ObjectCalendar` does not do it, and
2. the one renderer that does, `ObjectGrid`, reads it off **`schema.onNavigate`** — a non-author programmatic channel whose own comment says the schema **rejects the key by name** (`unrecognized_keys`). ⇒ copying that spelling into `ObjectGantt` would have left the dead click **exactly as it was**, because `ListView` forwards no such key and nothing else sets it.

⭐ The dev ignored the prescribed mechanism, built a real in-component sink instead, and reported the correction under ZONE 3 rather than quietly working around it. That is the right outcome in every respect.

⚠️ **This is the fourth instance of one class from this seat today** — a substring `grep -l` naming the wrong file, a flat match list attributing `:290` to the wrong `switch` arm, a key-name grep matching that word as a *value*, and now a count matching a **homonym with a different type**. Lane fact ⑦ already says a positive control proves the instrument works, not that the tree is current. This adds the sharper sibling, and it goes in the fold:

> **A control must be the same KIND as the subject, not merely the same SPELLING.** A control that lights for the wrong reason is worse than one that stays dark, because darkness invites a re-measurement and a false light does not.

⭐ What survives: **the harm I named was measured correct.** `page` was reachable and inert — a silent dead click — and the hold was right. Only the remedy I prescribed was wrong.

## The leg, verified at source

```ts
const navigateToRecord = useCallback((recordId, action?) => {
  if (typeof window === 'undefined') return;
  const row  = tasks.find(...)?.data;
  const href = row ? recordDetailHref(row)?.href ?? null
                   : resource ? deriveRecordPageHref(resource, recordId) : null;
  if (!href) return;
  if (action === 'new_window') { window.open(href, '_blank'); return; }
  window.history.pushState({}, '', href);
  window.dispatchEvent(new PopStateEvent('popstate'));
}, [tasks, recordDetailHref, resource]);
```

| check | verdict |
|---|---|
| route A — no host prop to the chart | ✅ a self-contained sink; no `{...props}` spread, no new prop read |
| objectui#7210 half 2 unruled, objectui#7333 unwoken | ✅ untouched |
| no new authorable key | ✅ |
| destination cannot drift from the drawer's own full-page link | ✅ reuses `recordDetailHref`, which is the better choice than a parallel derivation and is why the two can't disagree |
| `new_window` | ✅ now opens the **app-prefixed** href rather than the hook's unprefixed fallback — strictly better than before |

⭐ The pin is the right shape, and it is what I said I would check first: it asserts the **arguments** — `pushState({}, '', '/console/duly_task/record/1')` exactly once, a `popstate` dispatched, `location.pathname` moved — with `deriveRecordPageHref` deliberately **not** mocked and the document parked on `/console/duly_task/views/all`, so the `/console` prefix can only come from a real derivation. ⛔ Not `expect(drawer).toBeVisible()`.

⭐ And ablation leg 2 red-ing `new_window` **on its arguments rather than on the call** (`/duly_task/record/1` vs `/console/duly_task/record/1`) is exactly the kind of failure that proves a pin is measuring the thing it claims to.

**Net across both legs:** no authored value is worse than before, two are better (`page`, `new_window`), and `drawer` / unauthored / `none` are unchanged. That is what I asked for.

## Clause-② — a real protocol fact, worth keeping

`--pair 9115` final **exit 0**, but it was **4** first, and the reason is not a flake: adding `Closes #9114` made this PR deliver **two** cards, and the checker's C1 row reported a split — hung on the PR and on objectui#7334, but not on objectui#9114. Since the gate is a **dual** carrier and the second carrier is the only machine-readable evidence that a review was **cleared** rather than **stripped**, the split was closed by hanging `needs:contract-review` on objectui#9114 too, additively, with a read-back.

⇒ **a PR that closes a second card owes that card a carrier as well.** Carriers now: PR objectui#9115 ✅, card objectui#7334 ✅, card objectui#9114 ✅, all read back.

## Landing

CI on the current head: Lint, Type Check, Build & E2E, Build Docs and shard 3/4 all green; shards 1, 2 and 4 still running, **zero reds**. Flipping to ready and arming SQUASH auto-merge once they converge; ⛔ verified by queue ref, never by the echo.

⛔ objectui#9113 stays filed and out of scope, as directed. objectui#7334 keeps `Part of` — retiring it is triage's call, not this PR's and not mine.

---

os-tesla:
## Contract review

**Reviewed head:** `066c43ef41fd8948d5840baa9d991adab201dcdd`
**Cards delivered:** objectui#7334 (`Part of`) · objectui#9114 (`Closes`)
**Tier:** lane default for both build and review — maintainer ruling `5612097546` / objectstack#17285. For a non-spec seat the default-tier in-seat review plus the gates **is** the review of record; the `contract-review.md` 降档保险丝 binds only the spec and skills seats, ⛔ not this lane.

⚠️ Posted as a second comment because my ACCEPT at `5630155580` is a review in substance but ⛔ **not in the 同形 shape**, so it could not clear a carrier. The substantive findings are there; this is the conforming record.

### ① Derived judgments

**No derived type, accept-set or authorable surface moves in either direction.**

- `navigation` is an **already-declared, already-authorable** key on the view schema, and `ObjectGantt` **already read** it (`schema.navigation ?? { mode: 'drawer' }`). Leg 1 forwards it onto the gantt node inside `ListView`'s `case 'gantt'` branch. ⇒ this restores *declared = enforced*; it does not widen what an author may write.
- ⛔ **No new authorable key.** Verified: the change is a conditional spread of an existing key, and the conditional preserves **absent** (not present-and-undefined), so an unauthored view still takes the component's own default.
- Leg 2 adds `navigateToRecord`, a **component-internal** `useCallback` handed to `useNavigationOverlay`'s existing `onNavigate` parameter. That parameter is part of the hook's own declared options; nothing is added to any published surface, and nothing is read from a host prop.
- ⛔ `@objectstack/spec` is **not** touched. No spec symbol, no registration `inputs` entry, no exported type changes.
- ⚠️ **Declared narrowing of the behaviour change:** observable only for a gantt view that authors a view-level `navigation`. Measured across all seven sibling views that read the key — none of them starts receiving it, and five are pinned asserting exactly that while simultaneously asserting they still receive `onRowClick`, so the negative reading is a measurement rather than an empty harness.

⇒ **Clause-② increment: NONE in the accept-set sense.** The declaration this PR was dispatched under was `Clause-②: yes` (declared up front, on the conservative side per 「拿不准 ⇒ 按 yes」); the review's finding is that no declaration actually moved. ⛔ That does not retroactively make the declaration wrong — it makes the review cheap, which is the point of declaring `yes` when unsure.

### ② Semver

- Two released packages changed: `@object-ui/plugin-list`, `@object-ui/plugin-gantt`. One changeset, declaring **both**.
- `check-changeset-presence` exit 0 (5 source files of 2 released packages); `check-changeset-no-major` exit 0.
- ⚠️ This **is** a behaviour change, spelled out in the PR body's per-mode table: `page` moves from *drawer* to *navigates*, `new_window` from an unprefixed href to the app-prefixed one. No authored value is worse than before. This repo forbids `major`, so a break would ship as `minor` with the break spelled out — not needed here, as no authored document becomes invalid.

### ③ Boundary flags

| boundary | reading |
|---|---|
| governed surface | `check-governed-queue-guard --test` on **all six** changed paths → **NOT GOVERNED** |
| host-prop boundary (objectui#7210 half 2) | ⛔ untouched and unruled — no wrapper change, no `{...props}` spread, nothing read from `ObjectGanttRenderer` |
| objectui#7333 | ⛔ not woken; the `[]` short-circuit is not on any path this PR changes |
| objectui#9113 | ⛔ left filed and out of scope — pre-existing, outcome unchanged by either leg |
| cross-file line citations | `check:new-line-citations` exit 0, **0 new citations**; new comments cite sibling readers by SYMBOL per objectui#8875 |
| control bytes | gate exit 0, plus a direct scan of all six paths — no match |

### Independence pair

```
Implemented-by:  claude/issue-7334-gantt-navigation-forward   (mode:subagent dev — no session of its own)
```

⚠️ Stated plainly rather than dressed up: this is the lane's **default-tier in-seat review**, which is what the maintainer ruling makes the review of record for this seat. It is not an independent second session, and the `Implemented-by` line follows the standing convention for a subagent dev. ⭐ What gives it more weight than a rubber stamp on this particular PR is that the dev **falsified the reviewing seat twice** — the dispatch brief's `:290` location, and the review's own `ObjectCalendar` control — and both corrections were verified at source and accepted.

### Verdict

**PASS.**

Everything the hold demanded is delivered and verified at source: the `page` dead click is repaired by a real in-component sink, the pins assert navigation **arguments** rather than visibility, route A held throughout, and both ablations red the predicted cases while the overlay controls and the unauthored regression guard stay green.

⇒ Stripping `needs:contract-review` from all three carriers in one stroke — PR objectui#9115, card objectui#7334, card objectui#9114 — citing this record and the judged head above. `--pair 9115` was exit **0** at the dev's final run, after the split caused by the second closing keyword was closed by hanging the carrier on objectui#9114 as well.

---

PR #9120, opened 2026-09-11
fix(detail,layout): give both record headers a measured title-width floor

### LINKED ISSUE #7281: DetailView's own record header repeats the title/action width arbitration hazard fixed in page:header
Split out of #7244 (renderer-side fix landed on `page:header`). Not measured in a browser — identified statically while fixing the sibling header, and filed rather than fixed to keep that PR to one verified change.

## What #7244 fixed

`packages/components/src/renderers/layout/containers.tsx` laid the record header out as a `flex-nowrap` row: a `shrink-0` action tail plus a `flex-1 min-w-0` title column. The title column, being the only flexible item, absorbed the whole width deficit — measured at a 799px viewport, header 687px, tail 641.5px, leaving the h1 **29.5px**. Fixed with an `sm:min-w-48` floor on the title column plus `sm:flex-wrap` on the header.

## The same shape elsewhere

**1. `packages/plugin-detail/src/DetailView.tsx:919` — same shape, no protection.**

```
div.flex.flex-col.sm:flex-row.items-start.justify-between.gap-3.sm:gap-4.pb-4.border-b
  div.flex.items-start.gap-2.sm:gap-3.flex-1.min-w-0     <- title column, no floor
    div.flex-1.min-w-0
      h1.text-xl.sm:text-2xl.font-bold.truncate
```

No `flex-wrap` on the row and no width floor on the title column — the two things whose absence produced the reported bug. This header is the one that renders when the host does NOT supply a `page:header` (the file's own comment: suppressed when the host page renders its own `page:header`), so it is a live path, not dead code.

**2. `packages/layout/src/PageHeader.tsx:236` — partially protected.**

Its outer row is already `flex flex-wrap items-center gap-x-4 gap-y-2`, so the deficit has somewhere to go. But the title column is still `flex flex-col min-w-0 flex-1` with no floor, and `flex-1` means a flex-basis of 0 — so the column never contributes width that would force a wrap, and a wide `slot` sitting on the same line still takes the remainder. Milder, same root.

## Why this is worth its own card

The fix for each is a judgment about that layout (is a floor right, what value, should the row wrap), and each lives in a different package with its own build and test surface — so verifying them properly means a browser repro per header, which #7244's PR deliberately did not take on.

## Suggested check

Drive a record whose header carries three or more labelled actions at a viewport around 800px through each path, and read `h1.getBoundingClientRect().width` against `h1.scrollWidth`, the way #7244 was measured.

os-tesla:
Fixes #7281

Two headers repeated the title/action width arbitration that objectui#7244 fixed on `page:header`. The card was explicit that it had been identified **statically and never measured**, so the measurement — not the class change — is the deliverable here. Everything below is Chromium, the console dev server, three labelled `record_header` actions, layout allowed to settle before each read. ⚠️ jsdom and happy-dom report `clientWidth` 0, so no assertion in the vitest suites could have produced any of these numbers.

Re-derived locations (the card's `:919` / `:236` were from 2026-09-02): the DetailView header is `packages/plugin-detail/src/DetailView.tsx:1028-1030`, the PageHeader row `packages/layout/src/PageHeader.tsx:218/236`.

## The readings, before and after

Viewport 799px, the viewport objectui#7244 was measured at.

| header | | title column | h1 rect | h1 scrollWidth | action tail |
| --- | --- | --- | --- | --- | --- |
| **DetailView** | before | 58.17px | **6.17px** | 218px | 724.83px, same line |
| | after | 799px | **218.39px** | 218px | 724.83px, own line |
| **PageHeader** | before | 170.59px | **170.59px** | 265px | 564.41px, same line |
| | after | 751px | **751px** | 751px | 564.41px, own line |

The DetailView reading is worse than the sibling's: `58.17 = 40px back button + 12px gap + 6.17px`, so the h1 rendered as an ellipsis and nothing else. objectui#7244's sibling was left 29.5px.

Both hazards reproduce, so neither stop condition fired. The other stop condition — a floor that clips or overflows the action tail — was checked at 320px, 375px, 639px and 1440px: the tail is fully inside the header box and horizontal document overflow is 0 at every one of them.

## Two different floors, and why

**DetailView** takes objectui#7244's two utilities, both `sm:`-scoped so the sub-640px column layout is byte-for-byte what it was: `sm:flex-wrap` on the row so the deficit has somewhere to go, and `sm:min-w-64` on the title column. The floor is one step above the precedent's `sm:min-w-48` because **this** title column carries the back button and its gap INSIDE it — the 52px the table above accounts for — so 256px is what leaves the h1 the same ~192px readable floor objectui#7244 established.

**PageHeader**'s row already wrapped, so it needed only the floor. That floor is **unconditional**, not `sm:`-scoped, because this row has no breakpoint-scoped direction change and the squeeze measures WORSE just below `sm`:

| PageHeader @ 639px | title column | h1 rect | h1 scrollWidth |
| --- | --- | --- | --- |
| before | 10.59px | **10.59px** | 212px |
| after | 591px | **591px** | 591px |

An `sm:`-scoped floor copied from the precedent would have left that reading exactly as it is. Below about 620px the tail no longer fits on the row, wraps on its own, and the title is fine again — which is why the hazard band is bounded on both sides.

`min-w-48` REPLACES `min-w-0` rather than joining it: an explicit min-width overrides a flex item's automatic min-content minimum exactly as `min-w-0` did, so `truncate` keeps clipping. Measured, since that is the thing a reader will doubt — a 92-character title at 799px: h1 rect 751px of a 1682px scrollWidth, row still 799px, document overflow 0.

## Regression guard — the roomy viewport is untouched

| @ 1440px | title column | h1 | tail |
| --- | --- | --- | --- |
| DetailView before / after | 699.17 / 699.17 | 218.39 / 218.39, unclipped | same line, both |
| PageHeader before / after | 811.59 / 811.59 | 811.59 / 811.59, unclipped | same line, both |

Below `sm` the DetailView readings are identical before and after at 320px, 375px and 639px (`sm:`-scoped classes, computed `min-width` still `0px`).

## The pin

`e2e/record-header-title-width.spec.ts` — a browser pin per header asserting the **geometry**: at 799px the h1's `getBoundingClientRect().width` must clear the 192px readable floor and the action tail must stay inside the header's box. `toHaveClass('sm:min-w-48')` was ruled out as acceptance because it passes against a floor that does nothing.

What is real in it: the row, title column, action tail and h1 class strings are read out of the two source files at run time and compiled with the installed Tailwind, so deleting either floor changes the string the spec measures. Each extraction is anchored and validated against a token that is NOT the floor under test, so a walk that drifts onto a neighbouring element throws rather than measuring the wrong box — that guard fired for real during development and is why `innerColumn` is ordinal 5 and not 4.

What is fixture: the tail's CONTENT is fixed-width boxes, so the tail width is a constant rather than a font metric. Sized to the app's own measured tails (722px / 592px against 724.83px / 564.41px) — and that sizing matters: a 786px DetailView tail is wider than the 799px row can hold, so the row's wrap drops it unaided and the spec then passes with the floor deleted. The reasoning is written into the file.

The default `playwright.config.ts` picks all five tests up (`playwright test --list --project=chromium`: 42 tests in 8 files, 5 of them this file), so this runs on CI's `pnpm test:e2e --project=chromium` leg.

## Ablation

Each: mutate, prove the mutation reached disk by counting the removed text, run, restore, prove the tree clean by state — `git diff HEAD` empty AND the blob hash back to HEAD's, never by an exit code.

| removed | on-disk proof | verdict |
| --- | --- | --- |
| `sm:min-w-64` (DetailView floor) | 1 to 0 | red — `h1 is 9px of a 402px title; title column 61px, tail 722px`, expected >= 192 |
| `sm:flex-wrap` (DetailView row) | 1 to 0 | red — `tail escaped the header box (row right 799)` |
| `min-w-48` (PageHeader floor) | 1 to 0 | red — `h1 is 143px of a 479px title; title column 143px, tail 592px`, expected >= 192 |

Each ablation reds exactly one test and leaves the other four green, including the roomy-viewport guard — so the pin isolates the floor rather than the fixture. Restore verified after each: `git diff HEAD` empty, DetailView blob back to `30f5650`, PageHeader blob back to `1e7f086`.

One earlier ablation ran void and is reported as such rather than quietly re-run: a mis-written guard skipped the restore, so the next ablation ran on a doubly-mutated file and died as a wait timeout instead of a geometry assertion. Both were redone from a clean tree; the table above is the clean set. The settle-wait was also changed as a result, because a 10s TimeoutError hides the very reading a collapsed floor produces.

## Gates, on `f96f4f5`

| | |
| --- | --- |
| `vitest run packages/plugin-detail/` | 167 files, 1562 tests passed |
| `vitest run packages/layout/` | 21 files, 217 tests passed |
| `vitest run packages/components/src/__tests__/page-header-title-width.test.tsx` | 6 passed — objectui#7244's sibling pin, unchanged |
| `turbo run type-check --filter=@object-ui/plugin-detail --filter=@object-ui/layout` | 14/14 tasks |
| `pnpm type-check:e2e` | pass; `tsc --listFiles` confirms the new spec is IN the project, not excluded from it |
| `eslint` on the three changed files | 0 errors (49 pre-existing warnings, none in the new file) |
| `check-changeset-presence` / `-fixed` / `-no-major` | pass |
| `check-control-bytes` + a direct `grep -naP` control-byte scan of the four files | pass, 0 hits |
| `check-governed-queue-guard --test` | NOT GOVERNED — 4 paths, none matched |
| `check-unreferenced-sources`, `check-vi-mock-specifiers`, `check-type-check-coverage`, `check-new-cross-file-line-citations` | pass |

Every exit code was captured before any pipe (`cmd greater-than file 2 greater-than ampersand 1; EXIT=$?`), never read through `| tail`.

## Found and deliberately NOT fixed here

Filed as objectui#9119 — the same header sizes its title column to the full title text below the `sm` breakpoint, so a long record name pushes the page sideways (measured at 375px: title column 1201.22px inside a 375px row, 826px of horizontal document overflow). A different failure of the same element — a cross-axis size in a column flex container, where `min-w-0` is a floor and not a ceiling — and outside this card's fences. Measured twice, once on `d8b4739` and once with this branch applied: identical to the pixel, so it is pre-existing and this change neither causes nor fixes it.

Noted, not filed: `packages/components/src/__tests__/page-header-title-width.test.tsx` is class-shape only and says so in its own docblock; now that a browser geometry lane exists in `e2e/`, its assertions could be upgraded the same way. Not touched — that file is objectui#7244's surface.

## Acceptance notes

- No DOM was restructured. The whole change is three class strings.
- Clause-② stays **no**: no declared type, registration input, exported symbol, authorable key or accept-set moves. The diff is Tailwind utilities plus one new e2e spec plus one changeset.
- `packages/plugin-detail` also hosts the held PR objectui#9058; it touches `RelatedList.tsx` and `renderers/record-related-list.tsx`, neither of which appears in this diff.

---

### DIFF
--- a/.changeset/7281-record-header-title-width-floor.md
+++ b/.changeset/7281-record-header-title-width-floor.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/plugin-detail': patch
+'@object-ui/layout': patch
+---
+
+Record headers: give the title column a width floor so a wide action tail can no
+longer starve it.
+
+Two headers repeated the title/action width arbitration objectui#7244 fixed on
+`page:header`. Both were measured in Chromium at a 799px viewport with three
+labelled `record_header` actions, before and after:
+
+| header | before | after |
+| --- | --- | --- |
+| `DetailView` (`type: 'detail'`, the header that renders when the host supplies no `page:header`) | h1 6.17px of a 218px title | 218.39px, tail wrapped to its own line |
+| `PageHeader` (`@object-ui/layout`) | h1 170.59px of a 265px title | 751.00px, tail wrapped to its own line |
+
+`DetailView` gets the precedent's two utilities, both `sm:`-scoped so the
+sub-640px column layout is untouched: `sm:flex-wrap` on the row and
+`sm:min-w-64` on the title column. The floor is one step above the precedent's
+`sm:min-w-48` because this title column carries the 40px back button and a 12px
+gap inside it, so 256px is what leaves the h1 the same ~192px readable floor.
+
+`PageHeader`'s row already wrapped, so it needed only the floor — and that floor
+is unconditional rather than `sm:`-scoped, because this row has no
+breakpoint-scoped direction change and the squeeze measures worse just below
+`sm` (639px: h1 10.59px of a 212px title). `min-w-48` replaces `min-w-0` rather
+than joining it: an explicit min-width overrides a flex item's automatic
+min-content minimum exactly as `min-w-0` did, so long titles still ellipsise
+(measured at 799px: 751px rendered of a 1682px title, no horizontal overflow).
+
+Wide viewports are unchanged: at 1440px both headers keep the tail on the
+title's line with the title unclipped, before and after alike.
--- a/e2e/record-header-title-width.spec.ts
+++ b/e2e/record-header-title-width.spec.ts
@@ -1,0 +1,322 @@
+import { test, expect, type Page } from '@playwright/test';
+import { compile } from 'tailwindcss';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+/**
+ * GEOMETRY pin for the record-header title/action width arbitration on the two
+ * headers objectui#7281 covers — `packages/plugin-detail/src/DetailView.tsx`
+ * and `packages/layout/src/PageHeader.tsx`. Sibling of
+ * `packages/components/src/__tests__/page-header-title-width.test.tsx`, which
+ * pins objectui#7244's fix on `page:header`.
+ *
+ * ⚠️ WHY THIS FILE IS A PLAYWRIGHT SPEC AND NOT A VITEST TEST. jsdom and
+ * happy-dom have no layout engine — every `getBoundingClientRect()` there is
+ * 0x0 and `clientWidth` is 0 — so the sibling pin had to settle for
+ * CLASS-SHAPE assertions, and a class-shape assertion passes against a floor
+ * that does nothing. This file asserts the widths themselves, in Chromium,
+ * which is the only place the arbitration actually happens.
+ *
+ * ## What is real here and what is a fixture
+ *
+ * REAL, read out of the source files at run time: the header row's class
+ * string, the title column's class string, the action tail's class string and
+ * the h1's class string, for both headers — plus real Tailwind, compiled from
+ * the installed `tailwindcss` for exactly those candidates. Delete
+ * `sm:min-w-64` from DetailView or `min-w-48` from PageHeader and the string
+ * this file measures changes with it.
+ *
+ * FIXTURE: the action tail's CONTENT is three or four fixed-width boxes rather
+ * than real `<Button>`s, so the tail's width is a constant instead of a
+ * font-metric. The element tree around them is transcribed from the two
+ * sources.
+ *
+ * ## Calibration against the running app (Chromium, console dev server,
+ * `type: 'detail'` and `<PageHeader actions={…}>`, three labelled
+ * `record_header` actions, viewport 799px, objectui#7281):
+ *
+ *   DetailView  before: h1  6.17px of a 218px scrollWidth, tail 724.83px
+ *               after:  h1 218.39px, tail wrapped to its own line
+ *   PageHeader  before: h1 170.59px of a 265px scrollWidth, tail 564.41px
+ *               after:  h1 751.00px, tail wrapped to its own line
+ *
+ * The fixture tails below (722px / 592px) are sized to that measured
+ * neighbourhood, so the geometry this file reproduces is the geometry the app
+ * produces, not a worst case invented to make a floor look necessary.
+ */
+
+const REPO_ROOT = (() => {
+  // Rooted on this file, never on `process.cwd()` — AGENTS.md's
+  // "tests read the filesystem relative to their own file" rule. Bare
+  // `import.meta.url`, walked up; never `new URL(<relative>, import.meta.url)`.
+  let dir = path.dirname(fileURLToPath(import.meta.url));
+  for (let i = 0; i < 10; i++) {
+    try {
+      readFileSync(path.join(dir, 'pnpm-workspace.yaml'), 'utf8');
+      return dir;
+    } catch {
+      dir = path.dirname(dir);
+    }
+  }
+  throw new Error('could not locate the repo root from ' + import.meta.url);
+})();
+
+const read = (rel: string) => readFileSync(path.join(REPO_ROOT, rel), 'utf8');
+
+/**
+ * Pull a literal `className="…"` out of a source file, anchored on a nearby
+ * landmark so the walk cannot silently drift onto a neighbouring element.
+ * `mustContain` is a token the extracted string has to carry that is NOT the
+ * floor under test, so a drifted walk throws instead of measuring the wrong
+ * element and passing.
+ */
+function classAt(
+  source: string,
+  file: string,
+  anchor: string,
+  nth: number,
+  mustContain: string[],
+): string {
+  const at = source.indexOf(anchor);
+  if (at < 0) throw new Error(`${file}: anchor not found: ${anchor}`);
+  const re = /className="([^"]+)"/g;
+  re.lastIndex = at;
+  let hit: RegExpExecArray | null = null;
+  for (let i = 0; i < nth; i++) {
+    hit = re.exec(source);
+    if (!hit) throw new Error(`${file}: only ${i} className literals after ${anchor}`);
+  }
+  const cls = hit![1];
+  for (const token of mustContain) {
+    if (!cls.split(/\s+/).includes(token)) {
+      throw new Error(
+        `${file}: className #${nth} after "${anchor}" is "${cls}", which does not carry "${token}" — ` +
+          'the extraction walked onto the wrong element, so this measurement would be meaningless.',
+      );
+    }
+  }
+  return cls;
+}
+
+/** Same, but the last literal BEFORE the anchor. */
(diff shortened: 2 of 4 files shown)


os-tesla:

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Merge-base `d8b4739` via `git merge-base` (⛔ not `base.sha` — lane fact 44). No contract-review record is owed: `Clause-②: no` was declared up front and the diff confirms it — three Tailwind utilities, one e2e spec, one changeset, and **no declaration of any kind**.

## The two things I said I would check first, both verified

**1. The fence — class strings only, no DOM restructuring.** Confirmed by reading the diff, not the report:

```
PageHeader.tsx     min-w-0  ->  min-w-48                        (unconditional)
DetailView.tsx     row:        + sm:flex-wrap
                   title col:  min-w-0  + sm:min-w-64
```

Three class strings. The element tree is untouched in both files.

**2. The pin asserts geometry, not classes.** `READABLE_H1_FLOOR = 192`, asserted against `h1.getBoundingClientRect().width`, plus clipping (`scrollWidth > ceil(width)`), tail containment and document overflow. The file carries its own note at the assertion — *"Not `toHaveClass('sm:min-w-48')` — a floor that does nothing would pass"* — which is exactly the acceptance bar the dispatch set.

⭐ And the extraction that feeds it is genuinely defended: the class strings are read out of the two sources at run time and compiled with the installed Tailwind, each walk **anchored and validated against a token that is not the floor under test**, so drifting onto a neighbouring element throws instead of silently measuring the wrong box. The report says that guard fired for real during development and is why `innerColumn` is ordinal 5 — a guard with a scar is worth more than one without.

## What earns the accept beyond the checklist

⭐ **The per-header divergence is reasoned, not fudged.** The dispatch said "report, don't stop, if the two headers want different floors" and the answer is better than the question:

- **DetailView** needs `sm:min-w-64`, one step above the precedent's `sm:min-w-48`, because *this* title column carries the 40px back button and its 12px gap **inside** it — so 256px is what leaves the h1 the same ~192px the precedent established. The before reading proves the arithmetic: `58.17 = 40 + 12 + 6.17`, i.e. the h1 rendered as an ellipsis and nothing else.
- **PageHeader**'s floor is **unconditional rather than `sm:`-scoped**, and the reason is the one that matters: the squeeze measures **worse just below `sm`** (639px: h1 10.59px of a 212px scrollWidth). ⭐ A faithful copy of the precedent would have left the worst reading in the file untouched. That is the kind of thing a class-shape test can never surface.

**The numbers are worse than the card guessed**, which is the right direction for a statically-identified card: 6.17px against the 29.5px objectui#7244 measured on the sibling.

**The one claim a reader would doubt was measured.** Replacing `min-w-0` with `min-w-48` could in principle restore the flex `auto` min-content minimum and blow the row out. It does not — an explicit `min-width` overrides that minimum exactly as `0` does — and rather than assert it, the dev measured a 92-character title at 799px: h1 751px of a 1682px scrollWidth, row still 799px, document overflow 0.

**The fixture sizing is honest about being a fixture, and about why it matters.** Sizing the tail to the app's real measured widths is load-bearing: a 786px DetailView tail exceeds the 799px row, the row's wrap drops it unaided, and the spec then passes **with the floor deleted**. Writing that reasoning into the file is the difference between a pin and a decoration.

**The void ablation is reported as void.** A mis-written guard skipped a restore, the next leg ran doubly-mutated and died as a wait timeout rather than a geometry assertion — reported, both redone from a clean tree, and the settle-wait changed because a `TimeoutError` hides the very reading a collapsed floor produces. ⛔ That is not a run to quietly repeat; naming it is what makes the clean table believable.

Each ablation reds **exactly one** test and leaves the roomy-viewport guard green, so the pin isolates the floor rather than the fixture.

## Out-of-scope handling — correct on both

- **objectui#9119** filed and ⛔ not fixed: the same header sizes its title column to the full title text below `sm`, pushing the page sideways (375px: title column 1201.22px inside a 375px row, 826px of horizontal overflow). ⭐ Measured **twice — on `d8b4739` and with this branch applied, identical to the pixel** — which is what establishes it as pre-existing rather than caused here. That is the right way to file a neighbouring bug found mid-change.
- The class-shape-only sibling test on objectui#7244's surface: noted, not touched, correctly outside this card's fences.

## Landing

`mergeable_state` was **behind** (base moved to `2c878be` when objectui#9115 landed). Merging the base in per lane fact 57 — ⛔ not re-running anything — then flipping to ready and arming auto-merge. Verified by queue ref, ⛔ never by the echo, which is currently returning a populated-but-wrong `method`.

⚠️ This PR says `Fixes #7281`, which is correct here — unlike objectui#9110 and objectui#9115, this card's scope is wholly discharged, so auto-close is right. The card's `pm:dispatched` and assignee will still need stripping by hand; `Fixes` has never done either in this lane, 35 times running.

---

PR #9121, opened 2026-09-11
fix(plugin-charts, app-shell): rank the spec-declared lookup spellings first in two snake-only readers (objectui#7435)

os-tesla:
Part of #7435

Two readers of the resolved object-schema field def spoke a dialect the contract refuses. Each of the three keys the card's dated correction narrowed it to — `displayField`, `descriptionField`, `lookupFilters` — now has its **spec-declared spelling ranked first**, with the existing `snake_case` legs kept behind it in their existing relative order (the shape objectui#7155 established).

| reader | before | after |
| :-- | :-- | :-- |
| `packages/plugin-charts/src/ObjectChart.tsx` | `reference_field OR display_field OR 'name'` | `displayField OR reference_field OR display_field OR 'name'` |
| `packages/app-shell/src/utils/resolveActionParams.ts` | three snake-only reads | declared spelling first on each of the three |

The user-visible half: a chart grouped by a lookup whose field def declares `displayField` fell through to the generic `name` heuristic and drew a label the author had explicitly overridden; an action param's picker silently lost its display/description/filters config and rendered with defaults. Nothing warned, because the readers were reading keys that were not there.

---

## Task 1 — the pin re-measure the card's correction asked for

The correction was taken on `@objectstack/spec@17.2.0` and said to re-measure against the pin actually resolved at dispatch time. **Resolved here: 17.4.0**, and it is the only version in the tree (`node_modules/.pnpm` holds exactly one `@objectstack+spec@17.4.0` entry; `packages/plugin-charts` and `packages/app-shell` both resolve it).

`FieldSchema.safeParse` over a minimal lookup def, controls lit in the same run:

| probe | verdict on 17.4.0 |
| :-- | :-- |
| CONTROL — minimal lookup def, nothing added | **ACCEPTED** |
| CONTROL — `zzz_not_a_real_key` | **REJECTED** `unrecognized_keys` |
| `displayField` | ACCEPTED |
| `descriptionField` | ACCEPTED |
| `lookupFilters` | ACCEPTED |
| `display_field` · `description_field` · `lookup_filters` | REJECTED `unrecognized_keys` |
| `reference_field` · `id_field` | REJECTED `unrecognized_keys` |
| `idField` · `titleFormat` (the carve-outs) | REJECTED `unrecognized_keys` |

Structural second instrument — `FieldSchema.shape`, 74 keys: `displayField` / `descriptionField` / `lookupFilters` present; `idField` / `titleFormat` absent. **No key has moved, so the scope is unchanged and the two carve-outs still hold on the resolved pin.**

### One missed prediction, reported rather than smoothed

Run 1 wrote the minimal lookup def with `referenceTo` as the reference target, and **CONTROL A came back REJECTED** — a failed control, so that run is not a reading. The declared spelling of the lookup target on `FieldSchema` is `reference` (`shape` carries `reference` and `referenceVia`, and no `referenceTo` at all). Run 2 with `reference` is the table above, both controls lit. Worth recording because it is a live fact about this file: `resolveActionParams` EMITS `referenceTo` — that is `ActionParamDef`'s key, not `FieldSchema`'s, and the two are not the same vocabulary.

---

## Task 2 — the per-site producer sweep

objectui#7155 kept its snake leg because four in-repo producers emitted it. That finding does not transfer, so it was redone per site. Both bags were traced first:

- **`ObjectChart`** reads `ds.getObjectSchema(schema.objectName)` — whatever `DataSource` the host injected.
- **`resolveActionParams`** reads `ctx.objects`, i.e. `useMetadata().objects`, populated in `MetadataProvider`'s `ensureType`.

**Producer repo** — `objectstack-ai/objectstack` `origin/main` at `ea2940d1`, counted in **key position** (`KEY:` with optional quotes), file counts:

| spelling | files | | control | files |
| :-- | --: | :-- | :-- | --: |
| `reference_field` | **0** | | `displayField` | **23** |
| `display_field` | **0** | | `descriptionField` | **9** |
| `description_field` | **0** | | `lookupFilters` | **11** |
| `lookup_filters` | **0** | | | |

⚠️ A **bare** word grep is the trap here, and it fires on this very sweep: bare `reference_field` matches **2** files on that ref, both of which are **prose about objectui's own read chain** (a changeset paragraph and a `packages/spec/liveness/field.json` note), not a produced key. Read bare, the no-producer verdict would have looked falsified by two phantom producers.

**This repo** — zero non-test producers of any of the three snake spellings. Every occurrence of `display_field` / `description_field` / `lookup_filters` outside a test fixture is one CHANGELOG paragraph. The `reference_field` writers that do exist (`RecordDetailView`, `RecordDetailDrawer`, `RecordMetaFooter`, `RelatedList`, `buildDefaultPageSchema`) are **forwarders** — they copy an incoming def's value onto a widget-meta bag — plus the `plugin-grid` register entry and a view-column zod declaration. None of them produces either of the two bags above.

### ⭐ Why the snake legs stay anyway

Two producers lie outside everything the sweep can see, and the stop condition is about them:

1. **A document stored before the key was tightened.** The serve path runs no parse (objectui#7650), so `FieldSchema` strictness is evidence about the write door only.
2. **A host `DataSource`.** `ObjectChart` reads whatever adapter the host injected. Only `ObjectStackAdapter.getObjectSchema` runs `normalizeSchemaReferenceKeys`; a third-party adapter's document reaches the chain untouched.

Dropping a leg would be a silent regression for already-authored data — strictly worse than the wrong-label defect this change addresses. **Nothing is removed here.**

### ⚠️ A premise worth reporting: objectui#7650's ingestion fold has LANDED

Named, not worked — it is on this lane's binding-fence list. But it changes how the sweep reads, so it belongs in the record: `@object-ui/core`'s `normalizeFieldReferenceKeys` already folds `display_field` / `description_field` / `lookup_filters` (and `lookup_columns`) onto the declared spelling, derived from `FieldSchema.shape` rather than a hand table, **leaving the legacy key in place**. Both bags above run it at ingestion. So for a def that arrives through either choke point, the camel leg this PR adds is the one that will be populated — and the snake legs behind it are protecting exactly case 2 above, the adapter that never reaches a choke point.

---

## The `reference_field` no-producer verdict, re-derived

**It holds.** Zero key-position occurrences in the producer repo, control lit at 23 files for `displayField` (the register's own note cites 68, measured differently and repo-wide; the verdict is the same either way). Not on `FieldSchema` in any casing.

**Should that leg be read at all?** My answer: not first — which this PR settles — but removing it is a **retirement**, not a side effect of an additive card, and it needs the same producer evidence for the host-adapter case that keeps `display_field` alive. Its position **relative to `display_field` is deliberately unchanged**: reordering two legs nothing produces would be an unmeasured behaviour change riding along with a measured one. Recorded below rather than acted on.

---

## Falsified premises, reported and kept going

Per the brief, a location that is wrong while the harm is untouched is not a stop.

- **The card body quotes `referenceTo: param.reference ?? field.reference_to ?? field.reference`.** On `main` today that line is `param.reference ?? field.reference` — objectui#6837 half 2 already narrowed it. The harm the card names is untouched, and `referenceTo` was never in my three-key scope.
- **The card's acceptance grid implies three keys on each of two readers.** `ObjectChart` has **no** description or filters read, in either casing — its lookup branch reads a display field and an id field and nothing else. Adding reads there would have fabricated a consumer for a value nothing reads, which is the defect `plugin-grid`'s relational-meta register exists to prevent. So the live grid is **four cells, not six**: `displayField` at both readers, `descriptionField` and `lookupFilters` at `resolveActionParams` only.

---

## Acceptance — the resolved VALUE, not that the reader ran

`packages/plugin-charts/src/ObjectChart.declaredDisplayFieldLeg-7435.test.tsx` (5 cases) asserts the axis label the row ended up with. The referenced rows carry three different labels on three columns (`title` / `legacy_title` / `name`), so no assertion can be satisfied by the wrong leg.

`packages/app-shell/src/utils/resolveActionParams.declaredLookupLegs-7435.test.ts` (11 cases) reads **two exits**: the `ActionParamDef` this resolver emits, and the field object `paramToField()` hands `ActionParamDialog`, which is what `LookupField` consumes. A value that reaches the first and is dropped at the second would not be a repair, and only the second exit can say so.

Both files cover, per key: camel authored reaches the consumer · snake-only still resolves · **camel wins when both are present** (the ranking's own case) · and, for the chart, the `name` heuristic that names the pre-change wrong answer. Three further cases pin the carve-outs as **absences**: a camel-authored `idField` / `titleFormat` must NOT reach the param, while their snake reads keep working.

## Ablation — each new leg, on disk, restored by state

Each leg was deleted on disk, the mutation proven to have landed (anchor occurrence count before/after, plus the mutated blob hash differing from HEAD's), the suite run, then restored and **proven clean by state** — `git diff HEAD` empty AND `git hash-object` back to HEAD's blob — never by an exit code. Every script carried a `trap ... EXIT INT TERM` restore with absolute paths.

| ablated leg | result | restore |
| :-- | :-- | :-- |
| `ObjectChart` `fieldDef.displayField` | 2 failed / 3 passed — exactly "REACHES the axis" and "WINS" | blob `32b7ef61` back, diff empty |
| `resolveActionParams` `field.displayField` | 2 failed / 9 passed — "REACHES" and "camel WINS" | blob `43783b7c` back, diff empty |
| `resolveActionParams` `field.descriptionField` | 2 failed / 9 passed — same pair for its key | blob `43783b7c` back, diff empty |
| `resolveActionParams` `field.lookupFilters` | 2 failed / 9 passed — same pair for its key | blob `43783b7c` back, diff empty |

No dist preflight applies: both suites import the reader by relative path, so the source file under test is the one that runs.

## Verification

All exit codes captured by **redirect to a file**, never after a pipe.

- `pnpm exec vitest run packages/plugin-charts/ packages/app-shell/` — **737 files, 7115 passed, 1 skipped**, exit 0 (via the shared verify lock).
- `turbo run type-check --filter=@object-ui/plugin-charts --filter=@object-ui/app-shell` — **31/31 tasks successful**, exit 0. Each package's `type-check` runs `tsc --noEmit AND tsc -p tsconfig.test.json`, and `--listFiles` confirms both new test files are in those projects (app-shell 4494 files, plugin-charts 1545).
- `scripts/__tests__/check-installed-spec-pin-claims.test.ts` — **29 passed** (owed because this change edits that gate script).
- Gates, all exit 0: `check:control-bytes` (7311 tracked text files scanned) · `check:installed-pin-claims` · `check:changeset-claims` · `check:action-forward-parity` · `check:handler-key-reads` · `check:designer-field-key-parity` · `check:unreferenced-sources` · `check:new-line-citations` · `check:spec-symbols` · `lint:coverage` · `check:lint-rule-coverage` · `type-check:coverage` · `check:vi-mock-specifiers` · `check:vi-mock-inherit` · `check:vi-mock-override-shape` · `check:comment-mask-corpus` · `check-changeset-presence.mjs` · `check-governed-queue-guard.mjs --test` (NOT GOVERNED, 6 paths against 5 surfaces).
- **Lint, narrowed and declared.** `eslint .` in each touched package: plugin-charts **69 files, 0 errors**; app-shell **1140 files, 0 errors**; both new test files present in the `--format json` output. Plus `lint:root` (0 errors), which covers the edited gate script. The narrowing is a measurement, not a skip: this repo's `eslint.config.js` sets **no** `parserOptions.project` and no `projectService`, so type-aware linting is off and this diff cannot move the verdict on any file it does not contain. The repo-wide run is CI's.

⚠️ `check:nul-bytes` does not exist in this repo (it is the sibling's spelling) — this repo's equivalent is `check:control-bytes`, run above. A manual `grep -naP` control-byte scan over all five touched files also came back empty.

## Acceptance notes

Found while measuring, correctly **not** changed here.

- **`lookupColumns` / `lookupPageSize` / `dependsOn` are `FieldSchema`-declared and still read snake-only in `resolveActionParams`** — the same defect class as the three repaired keys, three lines away. Outside this card's narrowed three-key scope; they remain recorded on objectui#7435, which is why this PR says `Part of` rather than closing it.
- **`reference_field` should eventually leave `ObjectChart`'s chain.** No-producer verdict re-derived and holding; demoted below the declared spelling here, removal left to a retirement with its own evidence.
- **`ObjectChart` reads `fieldDef.id_field` with no declared spelling to re-point to**, and `resolveActionParams` reads `field.title_format` whose canonical target carries an ADR-0079 deprecation toward `nameField`. Both are the card's own carve-outs, routed to objectui#7650 option A / a spec-side decision. Named, not worked; their absence from this change is now pinned by test rather than by prose alone.
- **The census note in `resolveActionParams` had gone stale in one clause** ("none of the four reads below has a camel leg"). Corrected in place, since this change is what made it false; the KEEP verdict it supports is unchanged.

---

---


os-tesla:
## Seat review — ⭐ **ACCEPT**, and this one corrected the dispatch twice.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report. Merge-base `d8b4739` via `git merge-base` (⛔ not `base.sha`).

### The shape, verified

```
ObjectChart.tsx        fieldDef.displayField || fieldDef.reference_field
                       || fieldDef.display_field || 'name'
resolveActionParams    field.displayField     ?? field.display_field ?? field.reference_field
                       field.descriptionField ?? field.description_field
                       field.lookupFilters    ?? field.lookup_filters
RuntimeField           + displayField? / descriptionField? / lookupFilters?
```

Camel ranked first, snake legs kept in their existing relative order. ✅

| fence | verdict |
|---|---|
| three keys only, `idField` / `titleFormat` untouched | ✅ **zero** new reads of either in source — the remaining mentions are comments and **13 test lines pinning their ABSENCE**, which is better than leaving them unmentioned |
| `@objectstack/spec` untouched | ✅ |
| no widening into `@object-ui/fields` / `plugin-grid` | ✅ |
| objectui#7650 named, not worked | ✅ — and see below, because what it found there matters |

⚠️ **The one edit I checked hardest, because gate weakening is manual floor:** `scripts/check-installed-spec-pin-claims.mjs` loses a LEDGER entry. It is **not** a weakening — the removed entry recorded a *stale claim* in `ObjectChart.tsx` citing spec `17.2.0` (*"backs the claim that `idField` is not a leg of the change"*), and this PR **deletes that comment from the source**. The claim is gone, so its debt entry is correctly retired. ⭐ The gate demanded it: the report shows it went RED first with `LEDGER is stale: packages/plugin-charts/src/ObjectChart.tsx`. That is the gate working, not being softened.

## ⛔ Two of my brief's claims were falsified. Both corrections are right.

**1. My acceptance grid was wrong — the live grid is FOUR cells, not six.** I wrote *"for each of the three keys on each of the two readers."* ⛔ `ObjectChart` has **no `descriptionField` and no `lookupFilters` read in either casing**. Adding them would have written a consumer for a value nothing reads — precisely the shape `plugin-grid`'s relational-meta register exists to prevent. The dev kept going (the harm was untouched) and reported it, which is exactly what ZONE 3 asks for.

**2. The card body quotes a leg that no longer exists.** It shows `referenceTo: param.reference ?? field.reference_to ?? field.reference`; `main` today reads `param.reference ?? field.reference` — objectui#6837 half 2 already narrowed it. Another card body stale against the tree; ⛔ and `referenceTo` was never in the three-key scope anyway.

⭐ And a **third**, self-reported rather than hidden: the first `FieldSchema.safeParse` probe's minimal definition used `referenceTo`, and its **control failed**, because `FieldSchema` declares **`reference`**, not `referenceTo`. Run 2 is the reading. ⭐ Reporting a failed control instead of quietly re-running it is the single most valuable habit in this lane — a control that fails is the instrument telling you the fixture is wrong, and today this seat shipped a *lit* control that was wrong in the other direction and had to be corrected by a dev.

## The producer sweep, and why the snake legs correctly stay

Producer repo `objectstack` `ea2940d1`, **key-position** counts: `reference_field` **0**, `display_field` **0**, `description_field` **0**, `lookup_filters` **0** — against lit controls `displayField` **23**, `descriptionField` **9**, `lookupFilters` **11**. Zero non-test producers in objectui either.

⭐ **And the snake legs were kept anyway, for the reason the sweep cannot see:** a document stored before the key tightened (the serve path runs no parse — objectui#7650's own subject) and a **host `DataSource`** that never passes through the adapter's normalisation. That is the right call. A producer sweep measures what is *emitted today*, ⛔ not what is *already stored* — deleting a read on the strength of a producer count is how you silently drop existing data.

⚠️ `reference_field`'s `no-producer` verdict re-derived and **holds**; it is now demoted below the declared spelling. Whether it should be read at all is a retirement needing its own evidence — correctly not taken here. ⭐ Note the dev flagged that its only two bare-word hits are **prose about objectui's own chain**, i.e. the key-vs-value grep trap the brief warned about, caught in the wild.

## ⭐ The finding that most changes how this card reads

objectui#7650's ingestion choke point has **already landed on `main`**: `@object-ui/core`'s `normalizeFieldReferenceKeys` folds `display_field` / `description_field` / `lookup_filters` / `lookup_columns` onto the declared spelling — **derived from `FieldSchema.shape`**, and **leaving** the legacy key — and *both* of this card's bags run it at ingestion.

⇒ for any definition arriving through a choke point, **the camel leg added here is the populated one**, and the snake legs behind it now protect the host-adapter case alone. Named, not worked, per the binding fence. ⛔ It does not change this PR's shape, but it is load-bearing context for whoever takes objectui#7650 or the `reference_field` retirement.

## Contract review

**Reviewed head:** `6621c26`
**Card delivered:** objectui#7435 (`Part of` — correctly not a closing keyword; `lookupColumns` / `lookupPageSize` / `dependsOn` remain snake-only on this card)

### ① Derived judgments

Both readers' **accept sets widen**: each now accepts the `FieldSchema`-declared camelCase spelling it previously refused, ranked ahead of the snake legs it already accepted. ⛔ Nothing is **narrowed** — every spelling accepted before is still accepted, in the same relative order. `RuntimeField` gains three optional camel twins; no key is removed from any declared surface. ⛔ `@objectstack/spec` untouched; no registration `inputs` entry, no exported type removed. The declared-key measurement was re-taken on the **actually resolved pin 17.4.0** (the card's correction was written against 17.2.0): all three camel keys ACCEPTED, every snake twin plus `idField` and `titleFormat` REJECTED with `unrecognized_keys`, `FieldSchema.shape` 74 keys agreeing, controls lit both ways.

### ② Semver

Two released packages (`@object-ui/plugin-charts`, `@object-ui/app-shell`), one changeset, `check-changeset-presence` and `check-changeset-no-major` exit 0. A widening-only accept-set change: no authored document that was valid becomes invalid. ⛔ No `major`.

### ③ Boundary flags

| boundary | reading |
|---|---|
| governed surface | `check-governed-queue-guard --test`, 6 paths vs 5 surfaces → **NOT GOVERNED** |
| gate integrity | one LEDGER entry deleted, justified above as debt **paid**, ⛔ not threshold weakening; `check:installed-pin-claims` red→green with the source claim removed |
| objectui#7650 | named, ⛔ not worked (binding fence) |
| control bytes / line citations | `check:control-bytes` (7311 files) and `check:new-line-citations` exit 0, plus a direct scan of all five touched files |

### Independence pair

```
Implemented-by:  claude/issue-7435-lookup-spec-leg   (mode:subagent dev — no session of its own)
```

⚠️ Stated plainly: this is the lane's **default-tier in-seat review**, which the maintainer ruling makes the review of record for this seat — not an independent second session. ⭐ What gives it weight here is that the dev falsified the reviewing seat's own acceptance grid and the card's body, and both corrections were verified at source and accepted.

### Verdict

**PASS.** Stripping `needs:contract-review` from both carriers — PR objectui#9121 and card objectui#7435 — citing this record and the judged head. `--pair 9121` exit **0**, both carriers agreeing, hung additively with a read-back.

---

PR #9125, opened 2026-09-11
fix(sdui-parser): refuse an authored `type` attribute on the html tier (port of objectstack#13957)

### LINKED ISSUE #7235: sdui-parser (renderer/preview copy): a  attribute still overwrites the html-tier discriminator — objectstack now refuses it at parse
Surfaced while implementing objectstack#13957 (objectstack PR #14274), which lands the maintainer's 2026-09-01 ruling on the objectstack side. Filed rather than folded in: that PR is `packages/sdui-parser` in objectstack, and this half is this repo's copy.

## What objectstack just changed

On a `kind:'html'` page the tag name **is** the node's `type`, so an authored `type=` attribute is a name collision with the envelope's discriminator. objectstack's copy of the parser now **refuses it at parse time** with one `forbidden-attr` error naming both the tag and the attribute, and builds the node as `{ ...props, type: tag }` instead of `{ type: tag, ...props }`.

Before that, two outcomes: when the authored value named another **registered** type (a `flex` element carrying `type="grid"`), the tree carried the author's value as its discriminator, the manifest resolved it, every check passed, and a different component rendered with **zero diagnostics**; when it named nothing registered, `unknown-component` blamed the *value* and read as a missing plugin.

## This repo's copy still does the old thing

`packages/sdui-parser/src/parse.ts` here builds the node as:

```ts
const node: SchemaElement = markHtmlTierNode({ type: tag, ...props });
```

`props` spread after `type: tag`, and no refusal in `parseAttr` — so the renderer-side and live-preview copy still accepts the attribute and still lets it win the discriminator slot.

## Why this is filed and not urgent-by-default

The direction of the divergence is the safe one: objectstack runs the **save gate** and is now the stricter of the two, so a refused page never reaches a renderer that would misread it. That is the opposite of the save-clean-renders-inert failure the lockstep exists to prevent (objectstack#12719). ⚠️ But the two copies now differ in accepted grammar, and the live **edit preview** in the console runs this copy — an author gets no diagnostic while typing and a refusal on save.

Note the refusal deliberately reuses the existing `forbidden-attr` code rather than minting a new one, precisely so `check:sdui-lockstep` stays green in both directions; a port here needs no new code either.

## What a port would be

1. Refuse a `type` attribute in `parseAttr` with the existing `forbidden-attr` code, one diagnostic naming both the tag and the attribute.
2. Reverse the spread: `markHtmlTierNode({ ...props, type: tag })` — defense in depth, correct only *because* of (1); reversing it alone trades a silent overwrite for a silent discard.
3. ⛔ Do **not** carry the react tier's `specType` rescue (objectui#2880) into the html tier — the ruling declined it by name; that rescue stays where it lives.
4. After it lands, objectstack re-records its side with `pnpm gen:sdui-lockstep` when the `.objectui-sha` pin moves.

The ruled decision and its rationale are recorded as an append-only amendment on objectstack's ADR-0080.

---

os-tesla:
Fixes #7235

Port of objectstack#13957 (maintainer ruling 2026-09-01, recorded as an append-only amendment on objectstack's ADR-0080) into **this repo's copy** of `sdui-parser` — the copy the renderer and the console's live edit preview run. objectstack runs the save gate and has been the stricter of the two since that landed; this removes the grammar gap.

Session (durable reference, as a code span so it survives a body rewrite): `https://claude.ai/code/session_01UzHd6hDYatoDn17BuwKxnZ`

## The four ruled steps

1. **`parseAttr` refuses a `type` attribute** with the **existing** `forbidden-attr` code — one diagnostic naming both the tag and the attribute. ⛔ No new code minted: `check:sdui-lockstep` holds the two copies' diagnostic-code sets equal, so a code on one side only IS the dialect split that gate exists to catch.
2. **Spread reversed inside the existing wrapper:** `markHtmlTierNode({ ...props, type: tag })`. ⚠️ Defense in depth, correct only *because* of (1).
3. ⛔ The react tier's `specType` rescue (#2880) is **not** carried here — the ruling declined it by name.
4. ⛔ Not in this PR: objectstack re-records its side with `gen:sdui-lockstep` when the `.objectui-sha` pin moves.

⚠️ The `markHtmlTierNode` wrapper is objectui-only (objectstack builds a bare object). It is kept; the spread is reversed **inside** the call.

## What the defect actually was — measured, both halves

Behavioural probe, run on four trees in one lock hold. Written without literal angle brackets because GitHub eats tag-shaped fragments; read `FLEX-TYPE-GRID` as a `flex` element carrying `type="grid"`.

| tree | node `type` | diagnostics | reading |
|---|---|---|---|
| this PR (refusal + reversed spread) | `flex` | `["forbidden-attr"]` | loud, and the tag wins |
| ablation A — **step 2 alone** (refusal removed) | `flex` | `[]` | **silent DISCARD** — the authored value vanishes with no diagnostic |
| ablation B — **step 1 alone** (old spread order restored) | `flex` | `["forbidden-attr"]` | indistinguishable from this PR through the public API |
| ablation C — **both removed** (`origin/main`) | **`grid`** | `[]` | **silent OVERWRITE** — a grid renders where the author wrote a flex |

Verbatim, for `FLEX-TYPE-GRID`:

```
this PR      type=flex json={"gap":4,"type":"flex"}  codes=["forbidden-attr"]
ablation A   type=flex json={"type":"flex","gap":4}  codes=[]
ablation B   type=flex json={"type":"flex","gap":4}  codes=["forbidden-attr"]
ablation C   type=grid json={"type":"grid","gap":4}  codes=[]
```

⇒ the card's claim is confirmed as stated: **step 2 alone trades a silent overwrite for a silent discard.** Ablation B is the reason the order fix is documented as defense in depth rather than as the fix: with the refusal in place the attribute never reaches `props`, so restoring the old order changes nothing observable (its only red was the key-order pin, below).

Red/green matrix over the three affected test files (17 tests):

| leg | vitest exit | red |
|---|---|---|
| this PR | 0 | — |
| A (refusal removed) | 1 | 5 red |
| B (order restored) | 1 | 1 red (the key-order pin only) |
| C (both removed) | 1 | 7 red |

Each leg proved on disk before it was read (anchor occurrence counts plus the blob id), and each restore proved **by state** — `git diff HEAD` empty and `git hash-object` back to HEAD's `834bf3b0484c` — never by an exit code. No probe file is left in the tree.

## ⚠️ One PM acceptance criterion is falsified, deliberately not papered over

The brief asked that a node with no `type` attribute be **byte-identical** to today. It is not, and cannot be: reversing the spread moves `type` from **first** key to **last**, so `JSON.stringify` of a parsed tree emits the same pairs in a different order. `provenance.test.ts` pinned that order, and it went red:

```
AssertionError: expected [ 'className', 'type' ] to deeply equal [ 'type', 'className' ]
```

The key **set**, every value, the children and the html-tier provenance marker are unchanged — measured. The pin is updated to the new order with the reason written beside it, rather than relaxed to a sorted or set-wise comparison: an order pin that silently follows the implementation pins nothing. Called out in the changeset too, since a consumer diffing serialized trees will see it.

## Provenance survives the reversal

`markHtmlTierNode` is exactly where the reversal could have dropped the stamp. A new pin covers it at every depth and for the shapes the reversal moves — props present, props absent, and a container whose `children` are attached *after* the stamp.

## Sweep — does refusing `type` reject anything this repo authors today?

**No.** Population = every file referencing `kind:'html'`, `parseJsx`, or `@object-ui/sdui-parser` (65 files), plus `examples/**`, `content/docs/**`, `skills/**` and `e2e/**`.

- **1** html-tier document in the repo authors a `type` attribute: `action-type-input-html-tier.test.tsx` row 3, a deliberate **negative control** asserting the spelling does *not* work. It still fails to compile and still renders no button — so the harm the stop condition names (a narrowing that breaks existing authored content) is untouched. Its prose said the reason was `unknown-component` on the overwritten discriminator; that is now false, so the prose is corrected and the row asserts **which** diagnostic, otherwise it would go on passing while measuring nothing.
- `examples/**`: 8 `type=` hits, **0** in html-tier source (all `data-obj-type` DOM assertions and a `script type="module"` in an `index.html`). The schema catalog is the JSON tier, where `type` is the legitimate discriminator and nothing here touches it.
- `content/docs/**`: 6 hits, all react-tier `Block type=...` or MDX callouts — neither goes through this parser.
- `apps/console/src/sdui-tiers-preview.tsx`, the repo's own html-tier authoring example whose test asserts zero error diagnostics: **0** hits.

The sweep regex is not vacuous: it found the negative control, which pre-exists this change.

## `check:sdui-lockstep` — green, measured, in both directions

`check:sdui-lockstep` lives in objectstack and needs an install this reference checkout does not have, so it was run from a faithful scratch copy of that tree with a resolvable `typescript`:

```
--self-test exit=0
gate       exit=0
  OK — byte-identical to objectui@53ded82bf7a4 over 214 grammar line(s)
  [blob 0131f27cf86d] and agrees on all 24 diagnostic code(s)
```

And the half that gate structurally cannot see — objectui moving after the record — measured with **its own** `fingerprintOf` on this branch versus `origin/main`:

- grammar region blob: `0131f27cf86d` **before and after**, byte-identical, and equal to the recorded value. All hunks land at lines 24–190; the region delimiter is at line 392.
- diagnostic codes: identical before and after. This PR adds **none**.

⚠️ That same measurement surfaced a **pre-existing** divergence this PR did not cause and does not touch: objectui `origin/main` already carries **26** codes to objectstack's **24** (`inert-quick-add`, `member-type-mismatch`, plus the file `kanban-quick-add.ts`), all landed after the record was taken on 2026-09-05. Filed as objectstack#17645 with the measurement and the pre-existing control. It is latent today and will surface at the next `.objectui-sha` bump, with or without this PR.

## Verification

| what | how | result |
|---|---|---|
| `packages/sdui-parser` tests | `pnpm exec vitest run packages/sdui-parser/` | 15 files, **204 passed** |
| affected components test | `pnpm exec vitest run …/action-type-input-html-tier.test.tsx` | 4 passed |
| every html-tier consumer test in the repo | the 39 test files referencing `kind:'html'`, `parseJsx` or `@object-ui/sdui-parser` outside the parser package itself | 39 files, **744 passed** |
| `@object-ui/sdui-parser` type-check | `tsc --noEmit && tsc -p tsconfig.test.json` | exit 0 |
| `@object-ui/components` type-check | same, after building its dependency closure | exit 0 |
| dependency-closure builds | `pnpm --filter '@object-ui/sdui-parser^...' build`, same for components | exit 0 |
| `check:control-bytes` | repo gate | exit 0, 7313 files scanned |
| `check:new-line-citations` | repo gate | exit 0, 0 new citations |
| `check:upstream-port-parity` | repo gate | exit 0, 3 ported files match |
| `check:vi-mock-specifiers` | repo gate | exit 0 |
| `check-changeset-presence` | repo gate | exit 0, 1 changeset declared |
| eslint | 4 changed files, `--no-inline-config --format json` | exit 0, 0 problems |

Every exit code above was captured by **redirect** before any pipe.

**eslint narrowing, declared.** Population read from eslint's own config (`isPathIgnored` over `git ls-files`): **4799** lintable files. Files actually linted, read from `--format json`: **4**. Invariance: `eslint.config.js` declares no `projectService` and no `parserOptions.project` (0 occurrences), so linting is not type-aware and this diff cannot move the verdict on any file it does not contain. The narrowing is therefore a measurement, not a skip; CI runs the full farm.

**NOT MEASURED:** `check:sdui-registration-pins` exits **2** (`PREREQUISITE NOT MET` — no `apps/console/dist/assets` to weigh). It pins bundle-time registrations; this diff adds and removes none. Left to CI rather than run against nothing.

## Acceptance notes

Noted, not filed:

- `packages/components/src/renderers/action/__tests__/action-type-input-html-tier.test.tsx` carries a Chinese phrase in a code comment, which AGENTS.md commandment #-1 (English-only codebase) forbids. It pre-exists this change and is one of **743** files under `packages/**` that do — so it is a repo-wide convention gap, not a defect, and not this PR's to open. Successor: none; it would need a ruling on whether #-1 still describes the tree.
- The two copies of `parse.ts` differ in two further places beyond the spread order, the refusal and the objectui-only provenance wrapper — both **comment-only**, no behavioural difference: the whitespace-collapse rationale is worded differently on each side, and objectstack's `interpretBrace` carries a `LOCKSTEP:` banner this copy lacks. Reported per the brief's "report, don't stop" clause; no third behavioural difference exists.

---

### DIFF
--- a/.changeset/7235-html-tier-discriminator-attr.md
+++ b/.changeset/7235-html-tier-discriminator-attr.md
@@ -1,0 +1,36 @@
+---
+'@object-ui/sdui-parser': minor
+---
+
+Refuse an authored `type=` attribute on the `kind:'html'` tier (objectui#7235 — the
+port of objectstack#13957, maintainer ruling 2026-09-01, recorded as an append-only
+amendment on ADR-0080).
+
+**Breaking, deliberately, and the direction is a narrowing.** On this tier the tag
+name *is* the node's `type`, so an authored `type=` attribute is a name collision
+with the envelope's own discriminator. It is now one `forbidden-attr` error naming
+both the tag and the attribute, and the node is composed as `{ ...props, type: tag }`
+so the tag always wins the slot.
+
+Two outcomes it replaces, the first of which produced no diagnostic at all:
+
+- the value named another **registered** type (`<flex type="grid">`) — the tree
+  carried `type:'grid'`, the manifest resolved it, every check passed, and a grid
+  rendered where the author wrote a flex. Zero diagnostics.
+- the value named **nothing registered** (`<object-chart type="bar">`) — loud, but
+  `unknown-component` naming `"bar"` read as a missing plugin rather than as an
+  attribute that should not be there.
+
+The existing `forbidden-attr` code is reused rather than a new one minted: the two
+copies of this parser (this one, which the renderer and the console's live edit
+preview run, and objectstack's, which the save gate runs) are held to one diagnostic
+vocabulary by `check:sdui-lockstep`, and a code minted on one side only is the dialect
+split that gate exists to catch.
+
+⚠️ Not carried over: the react tier's `specType` rescue (objectui#2880). The ruling
+declined it by name; that rescue stays where it lives.
+
+One visible non-behavioural consequence of the reversed spread: a compiled node's
+`type` key is now **last** rather than first in insertion order, so `JSON.stringify`
+of a parsed tree emits the same pairs in a different order. The key set, every value,
+the children and the html-tier provenance marker are unchanged.
--- a/packages/components/src/renderers/action/__tests__/action-type-input-html-tier.test.tsx
+++ b/packages/components/src/renderers/action/__tests__/action-type-input-html-tier.test.tsx
@@ -14,13 +14,20 @@
  * ## Why this row exists on the html tier specifically
  *
  * The old input name was unauthorable HERE and only here it is provable in one
- * hop. `parse.ts` composes a node as `{ type: tag, ...props }` — props spread
- * LAST — so an authored `type="api"` did not set an input, it REPLACED the
- * component discriminator and the node stopped resolving to a component at all.
- * `validate.ts` cannot report that either: `type` is in `BASE_PROPS`, so it is
- * skipped before the declared-input check ever runs. Two mechanisms, one
- * outcome, no diagnostic — which is the whole reason the collision was removed
- * at the source instead of being special-cased per tier.
+ * hop. `parse.ts` used to compose a node as `{ type: tag, ...props }` — props
+ * spread LAST — so an authored `type="api"` did not set an input, it REPLACED
+ * the component discriminator and the node stopped resolving to a component at
+ * all. `validate.ts` could not report that either: `type` is in `BASE_PROPS`,
+ * so it is skipped before the declared-input check ever runs. Two mechanisms,
+ * one outcome, no diagnostic — which is the whole reason the collision was
+ * removed at the source instead of being special-cased per tier.
+ *
+ * ⚠️ UPDATED by objectui#7235: the parser now REFUSES a `type` attribute on
+ * this tier outright (`forbidden-attr`, one diagnostic naming both the tag and
+ * the attribute) and composes `{ ...props, type: tag }`. Row 3 below still
+ * fails to compile and still renders no button, but for the ruled reason
+ * instead of the accidental one, and it now asserts WHICH diagnostic — without
+ * that it would go on passing while measuring nothing.
  *
  * The renamed input has no such problem: `actionType` is an ordinary declared
  * prop, so it survives the spread, validates against the manifest built from
@@ -36,10 +43,11 @@
  *    the handler being the only one registered; with it, the value is pinned to
  *    the authored prop.
  * 3. NO ALIAS — the pre-rename spelling `type="api"` does not resurrect the old
- *    behaviour. It is the discriminator: the node's type becomes `api`, which is
- *    not a component, so the page reports `unknown-component` and no button is
- *    rendered. It states the cost the ruling accepted by name (no alias, no
- *    transition window — the standing 不渐进 rule).
+ *    behaviour. Since objectui#7235 the attribute is REFUSED at parse, so the
+ *    page reports `forbidden-attr` naming the attribute and no button is
+ *    rendered. (Before that port it was the discriminator, and the page
+ *    reported `unknown-component` naming the VALUE.) Either way it states the
+ *    cost the ruling accepted by name: no alias, no transition window.
  * 4. THE DECLARATION — rows 1-3 are about the RENDERER, and every one of them
  *    was green before this change too: the renderer already read `actionType`
  *    first (`action:bar`'s member spread has always used it), and an undeclared
@@ -155,9 +163,14 @@ describe('objectui#7415 — action:button authors its execution type on the html
   it('no alias — the pre-rename `type` spelling is still the discriminator, not an input', async () => {
     renderHtmlPage(`<action:button label="Mark done" type="api" target="${TARGET}" />`);
 
-    // `{ type: tag, ...props }` — the authored `type` wins the slot, so the node
-    // is `<api>`, which no registry entry claims.
+    // objectui#7235: the attribute is refused at parse, so the panel names the
+    // ATTRIBUTE rather than the value. Asserting the panel alone would keep
+    // passing on the old `unknown-component` reading too, which is exactly the
+    // misdirection that port removed — so the message is read, not just the
+    // fact that compilation failed.
     expect(await screen.findByText(/failed to compile/i)).toBeInTheDocument();
+    expect(screen.getByText(/Attribute "type" is not allowed on <action:button>/)).toBeInTheDocument();
+    expect(screen.queryByText(/unknown component/i)).not.toBeInTheDocument();
     expect(screen.queryByRole('button', { name: 'Mark done' })).not.toBeInTheDocument();
     expect(api).not.toHaveBeenCalled();
   });
--- a/packages/sdui-parser/src/__tests__/provenance.test.ts
+++ b/packages/sdui-parser/src/__tests__/provenance.test.ts
@@ -51,15 +51,46 @@ describe('html-tier provenance (#4000)', () => {
     const root = tree as SchemaElement;
 
     expect(JSON.parse(JSON.stringify(root))).toEqual({ type: 'div', className: 'a' });
-    expect(Object.keys(root)).toEqual(['type', 'className']);
+    // ⚠️ Key ORDER, not key SET: `type` moved from first to last when the
+    // discriminator collision was closed (objectui#7235) and the node became
+    // `{ ...props, type: tag }`. The set, the values and the round-trip are
+    // unchanged — this line is the one visible consequence of that reversal,
+    // and it is here rather than adjusted away because an order pin that
+    // silently follows the implementation pins nothing.
+    expect(Object.keys(root)).toEqual(['className', 'type']);
     const seen: string[] = [];
     for (const k in root) seen.push(k);
-    expect(seen).toEqual(['type', 'className']);
+    expect(seen).toEqual(['className', 'type']);
 
     // The round-trip is the anti-forgery pin: a persisted tree comes back
     // unmarked, so a saved (or AI-generated) document cannot carry the
     // exemption back in with it.
     expect(isHtmlTierNode(JSON.parse(JSON.stringify(root)))).toBe(false);
+  });
+
+  it('survives the reversed spread — the marking is applied AFTER every key', () => {
+    // objectui#7235 reversed `markHtmlTierNode({ type: tag, ...props })` into
+    // `markHtmlTierNode({ ...props, type: tag })`. The wrapper is objectui-only
+    // (objectstack's copy of this parser builds a bare object), so the reversal
+    // is the one edit in that port where the stamp could have been dropped —
+    // e.g. by moving the call inside the literal, or by spreading a marked
+    // object into a fresh unmarked one. Pinned at every depth and for the
+    // shapes the reversal actually moves: props present, props absent, and a
+    // container whose children were attached after the stamp.
+    const { tree } = parseJsx('<flex direction="col" gap={4}><grid columns={2} /><div /></flex>');
+    const root = tree as SchemaElement;
+
+    expect(isHtmlTierNode(root)).toBe(true);
+    expect(root.type).toBe('flex');
+    // `children` is assigned after `markHtmlTierNode` returns; the stamp is on
+    // the same object, so it must still read true once they are attached.
+    expect(root.children).toHaveLength(2);
+    for (const child of root.children as SchemaElement[]) {
+      expect(isHtmlTierNode(child)).toBe(true);
+    }
+    // The propless element — `{ ...{}, type: tag }` — is the degenerate case
+    // of the reversal and has its own way of going wrong.
+    expect(isHtmlTierNode((root.children as SchemaElement[])[1])).toBe(true);
   });
 
   it('does not mark JSON smuggled in through a braced attribute', () => {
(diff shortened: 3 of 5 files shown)


os-tesla:
## Seat review — ⭐ **ACCEPT**, and it falsified one of my acceptance criteria correctly.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report. Merge-base `567f370` via `git merge-base`.

## ⛔ My acceptance criterion was wrong

I wrote: *"a node with no `type` attribute is byte-identical to today (regression guard)."* ⛔ **False, and unavoidably so.** Reversing the spread moves `type` from the **first** key to the **last**, so `JSON.stringify` of a parsed tree emits the same pairs in a different order. `provenance.test.ts` pinned that order and went red.

I asked for byte-identity from a change whose entire mechanism is reordering an object literal. The key **set**, every **value**, the children and the html-tier marker are unchanged — which is the guarantee I actually meant — but "byte-identical" was the wrong words for it, and the dev was right not to quietly satisfy them.

⭐ **And the handling is better than the criterion was.** The pin was **updated to the new order with the reason written beside it**, ⛔ not relaxed to a sorted comparison:

> *"this line is the one visible consequence of that reversal, and it is here rather than adjusted away because **an order pin that silently follows the implementation pins nothing**."*

That is the same principle as a control that must be able to fail, applied to a pin. Relaxing it to `sort()` would have made it green forever and worthless. The reordering is also called out in the changeset for consumers that diff serialised trees, which is where a downstream reader would actually meet it.

## The port, verified at source

```ts
const DISCRIMINATOR_ATTR = 'type';
if (name === DISCRIMINATOR_ATTR) { this.error('forbidden-attr', `Attribute "type" is not allowed on <${tag}> — …`) }
const node: SchemaElement = markHtmlTierNode({ ...props, type: tag });
```

| requirement | verdict |
|---|---|
| existing `forbidden-attr`, ⛔ no new code | ✅ — and the docblock explains it is **load-bearing**, not lazy: objectstack's lockstep holds the two code sets equal |
| one diagnostic naming **both** tag and attribute | ✅ objectstack's message, verbatim |
| `markHtmlTierNode` wrapper kept, spread reversed **inside** | ✅ |
| ⛔ no `specType` rescue (objectui#2880) | ✅ — and the docblock records *why*, including that it is not fixable at the warning layer because `type` is deliberately in `validate.ts`'s `BASE_PROPS` |
| step 4 (objectstack re-record) named, not done | ✅ |

⭐ **The new provenance pin is better than what I asked for.** I asked that the marking survive the reversal; it pins that at every depth **and** for the degenerate `{ ...{}, type: tag }` propless case — reasoning that the wrapper is the objectui-only edit and therefore the one place this port could silently drop the stamp. That is the right instinct about where a port breaks.

## The three-way ablation is the best evidence in this PR

| leg | result |
|---|---|
| baseline | `type=flex`, `codes=["forbidden-attr"]`, 17/17 green |
| **A — step 2 alone** (spread reversed, refusal removed) | `type=flex`, `codes=[]` — the authored value gone with **zero diagnostics**: the **silent discard** the card predicted |
| **B — step 1 alone** (refusal kept, old order) | indistinguishable through the public API — which is *why* the order fix is documented as defense in depth rather than sold as the fix |
| **C — both removed** (today's `main`) | **`type=GRID`**, `codes=[]` — the **silent overwrite**: a grid rendering where the author wrote a flex |

⇒ each step's contribution is measured separately, and leg C is the defect this card exists to close, reproduced.

## Stop conditions — both cleared, both with honest instrument handling

**Fixture sweep: CLEAN.** Exactly **one** html-tier document in the repo authors a `type` attribute, and it is a deliberate **negative control** asserting the spelling does not work. ⭐ The sweep regex is proven non-vacuous *because* it found that control. `examples/**`, `content/docs/**`, the schema catalog and the console's own html-tier authoring example are all clear.

**`check:sdui-lockstep`: GREEN both directions, measured.** ⭐ And the way it was measured matters: a bare run in the reference checkout exits **3 = PREREQUISITE NOT MET** (no `node_modules`), which is ⛔ **not a reading** — so it was run from a faithful scratch copy with a resolvable toolchain, `--self-test` exit 0 and the gate exit 0. For the half the gate structurally cannot see, objectstack's **own** `fingerprintOf` was run against this branch and against `origin/main`: grammar-region blob byte-identical before and after, diagnostic-code list identical, this PR adds **no** code.

## ⭐ Two findings worth more than the PR

**1. objectstack#17645 — the lockstep has drifted again, and it is latent.** objectui `main` carries **26** diagnostic codes to objectstack's **24** (`inert-quick-add` from objectui#8285, `member-type-mismatch` from objectui#8067), all landed **after** the 2026-09-05 record — i.e. inside the blind spot `check-sdui-lockstep`'s own header names. Green today because the gate compares objectstack against the record; it will red at the next `.objectui-sha` bump, **with or without this PR**. ⭐ Measured with objectstack's own extractor, and the control **pre-exists the change and is the same kind as the subject**: the same extractor on objectui at the **pinned** revision returns 24 and the matching file list, so it is demonstrably sensitive to exactly this drift. That is the control discipline this seat got wrong twice today, done right.

**2. A `packages/*/src` glob produced a FALSE ZERO, and it nearly changed the outcome.** Checking whether AGENTS.md's English-only rule describes the tree, the first probe reported **0** files with CJK comments; a proper walk reported **743** (60,233 characters). ⚠️ That is lane fact ㉑ — *a `git grep` pathspec glob matches the whole path, not a directory prefix* — hitting live, and the dev reports it nearly removed a verbatim ruling quote on the strength of that zero. ⭐ Reporting the near-miss rather than only the corrected number is what makes the number trustworthy.

## The open question, and my call

Keeping the verbatim Chinese ruling quote (option **A**) is right and I am accepting it. The reasoning holds: a **translated ruling can drift from what was decided**, whereas an English gloss is recoverable from the ADR — so the irreversible error is the translation, not the quote. And the rule as written has **743 counterexamples in the tree**, so it does not currently describe the repo.

⛔ **Not mine to settle beyond this PR:** AGENTS.md is a governed surface, so amending commandment #-1 to carve out verbatim ruling quotes (the dev's option C) is the maintainer's. Routing it as a **low-priority** note, ⛔ not a blocking decision — nothing here waits on it.

## Contract review

**Reviewed head:** `400737e` · **Card:** objectui#7235 (`Part of`)

### ① Derived judgments
This **NARROWS** the accepted grammar: the parser begins refusing a `type=` attribute it accepts today, on the html tier only. That is a deliberate, maintainer-ruled narrowing (2026-09-01, ADR-0080 amendment), ported from objectstack#13957 which already landed the same refusal upstream — so the two copies **converge** rather than diverge. ⛔ No new diagnostic code is minted (the existing `forbidden-attr` is reused, which is what holds the lockstep's code-set equality). ⛔ `@objectstack/spec` untouched. The node's key **set** and values are unchanged; only key **order** moves, disclosed in the changeset.

### ② Semver
Two released packages, one changeset, `check-changeset-presence` exit 0. A narrowing that rejects a spelling which today **silently renders the wrong component** — so no correct document becomes invalid. ⛔ No `major`.

### ③ Boundary flags
| boundary | reading |
|---|---|
| lockstep | `check:sdui-lockstep` green both directions; grammar-region blob byte-identical before/after; 0 codes added |
| `check:upstream-port-parity` | exit 0 — `parse.ts` is **not** in `upstream-port-pin.json` (verified; the pin holds 3 tooling files) |
| existing authored content | sweep clean; the one `type`-authoring document is a negative control asserting it does not work |
| ⚠️ NOT MEASURED, declared | `check:sdui-registration-pins` exits **2 = PREREQUISITE NOT MET** (no console bundle to weigh). This diff adds and removes no bundle-time registration; left to CI rather than run against nothing |
| control bytes / citations | exit 0 (7313 files); 0 new citations |

### Independence pair
```
Implemented-by:  claude/issue-7235-html-tier-discriminator-attr   (mode:subagent dev — no session of its own)
```
⚠️ The lane's default-tier in-seat review, which the maintainer ruling makes the review of record for this seat — ⛔ not an independent second session. ⭐ Its weight here is that the dev falsified the reviewing seat's own acceptance criterion and was right.

### Verdict
**PASS.** Stripping `needs:contract-review` from both carriers — PR objectui#9125 and card objectui#7235 — citing this record and the judged head. `--pair 9125` exit **0**, both carriers agreeing, hung additively with a read-back.

---

---

PR #9127, opened 2026-09-11
refactor(core): import the spec's context-token map and predicate (objectui#7265)

os-tesla:
Part of #7265 — the `@object-ui/core` slice, 2 of the entries seeded into the `DEBT` block of `scripts/check-spec-symbol-derivation.mjs`. Twelve entries remain, so this does not close the anchor; the gate's stale-ledger message points at that card and it closes when the ledger is empty.

## What changed

`packages/core/src/utils/filter-tokens.ts` carried module-local copies of two `@objectstack/spec/data` exports:

- `CONTEXT_TOKEN_SUGGESTIONS` — the nine-entry near-miss suggestion map
- `isContextToken` — the membership predicate

Both are imports now. Neither was ever exported from `@object-ui/core`, so no published surface moves. The file already re-exports `CONTEXT_TOKENS` from the same module, and 25 lines above the deleted map it explains why: "the copy was byte-identical, so every value comparison and every behavioural test passed while it sat here". This applies that argument to its two neighbours.

Both ledger blocks were regenerated mechanically, never hand-edited:

| block | before | after | command |
|---|---|---|---|
| `DEBT` (rule 1) | 13 names | 11 names | `--ledger` |
| `CLAIM_DEBT` (rule 2) | 19 names | 18 names | `--claim-ledger` |

Neither block gained an entry. The `CLAIM_DEBT` removal is not a rider — it is forced: rule 2's entry existed because the deleted `const` carried the comment "Mirrors `CONTEXT_TOKEN_SUGGESTIONS` in `@objectstack/spec`", and ratchet 5 fails the gate on a claim-ledger entry whose declaration is gone. `CLAIM_DEBT_ISSUE = 4592`'s dead anchor is deliberately left alone, as the card asks; only the entry count quoted in the comment beside it moved (19 to 18).

## The measurement, re-taken on the resolved pin

The card asserted byte-identity against `@objectstack/spec@17.2.0`. The resolved pin is **17.4.0**, so it was re-measured before the swap — "byte-identical" is a statement about a version, not a property.

```
resolved @objectstack/spec = 17.4.0

SUBJECT  local CONTEXT_TOKEN_SUGGESTIONS (HEAD 567f370)  vs  spec CONTEXT_TOKEN_SUGGESTIONS
  keys(local)=9 keys(other)=9  keyOrder=true values=true json=true
  => IDENTICAL

CONTROL  local CONTEXT_TOKEN_SUGGESTIONS  vs  spec CONTEXT_TOKEN_DESCRIPTIONS (pre-existing, same kind)
  keys(local)=9 keys(other)=2  keyOrder=false values=false json=false
  => DIVERGENT
  control lit (comparator can report a difference): true

SUBJECT2 predicate equivalence over 61 real vocabulary members
  local.isContextToken vs spec.isContextToken disagreements: 0
CONTROL2 local.isContextToken vs spec.isDateMacroToken (pre-existing, same kind)
  disagreements: 38 — control lit: true
```

The local side is the HEAD bytes: `git show HEAD:packages/core/src/utils/filter-tokens.ts`, transpiled and imported, so nothing in this diff can supply the answer. Each control is **pre-existing** (a spec export that predates this branch) and of the **same kind** as its subject (a context-token string map; a `(token) to boolean` token predicate from the same subpath), run through the **same** comparator.

Whole-resolver A/B, because the stop condition is about what the predicate accepts and rejects, not about the map:

```
A/B over 62 tokens x 2 placeholder spellings x 2 scopes = 248 cases
divergences: 0
CONTROL (before vs a do-nothing resolver): divergences 72 — lit: true
```

Returned values **and** warning text are byte-identical in every case, signed-in and signed-out, in both the brace and dollar-brace spellings, including prototype-key spellings.

## One type-level difference, handled rather than papered over

The spec's predicate is `isContextToken(token: string): boolean`; the deleted local one was a narrowing `token is ContextTokenName`. The narrowed branch feeds a scope lookup, so the lookup is now annotated `Record<string, …>` with a `satisfies Record<ContextTokenName, …>` clause. That keeps the only thing the narrowing bought — a third context token in the spec reds this literal at compile time instead of resolving to `undefined` at runtime — with no unchecked assertion anywhere. A runtime pin states the same ratchet so the failure names itself.

The declared floor also holds: `@object-ui/core` declares `@objectstack/spec` `^17.3.0`, and both names are in the published export list of `dist/data/index.d.ts` at 17.3.0 and 17.2.0 (checked against the registry tarballs, with a negative control name that answered "not in the export list").

## Verification

- `pnpm exec vitest run` over the two site test files plus every `scripts/__tests__` file that names the edited gate script (`check-spec-symbol-derivation.test.ts`, `check-doc-component-types.test.ts`, `check-doc-example-shared-reader.test.ts`, `spec-symbol-ledger-core-7265.test.ts`) — **6 files, 177 tests passed**. Editing a gate script owes its own suite on top of the families it derives, and `git grep -l` over the test dirs is how that set was found rather than recalled.
- `pnpm --filter @object-ui/core type-check` — exit 0 (covers `tsconfig.test.json` too)
- `pnpm run type-check:scripts` — exit 0
- `check:spec-symbols` exit 0 — 13 untriaged collisions to **11**, 19 unbacked claims to **18**, stale-ledger message does not fire
- `check:control-bytes`, `check:installed-pin-claims`, `check:changeset-claims`, `check:new-line-citations`, `check:comment-mask-corpus`, `check:unreferenced-sources`, `check:phantom-deps`, `check:unused-deps` — all exit 0
- `check-changeset-presence` exit 0 — empty-frontmatter changeset, declared as releasing nothing
- `check:spec-floors` — **PREREQUISITE NOT MET on an unbuilt tree**, and narrowed rather than skipped. On the tree as checked out it reports 17 `no-artifact` findings across 17 packages ("produced no build output to judge"), `@object-ui/core` among them. After `pnpm --filter @object-ui/core build`, `@object-ui/core` **drops out of the finding list** (17 to 16) — it was judged, and judged clean; the artifact count it inspected went 128 to 328 and the (subpath, symbol) pairs judged went 135 to 170. The remaining 16 are unbuilt sibling packages, a whole-workspace-build precondition CI satisfies, not a verdict on this diff. The floor question it would answer was answered directly against the registry above.
- `pnpm lint` — **narrowed, declared.** Three readings, because a narrowing is only a measurement if all three are present: (i) the population comes from eslint's own resolved configuration, not a guess — `isPathIgnored` over the tracked tree says **4800 of 4801** lintable-extension files are in scope; (ii) the narrowed run covered **4** files, counted from `--format json` output, `exit 0`, 0 errors and 6 warnings; (iii) the invariance claim: this repo does **not** enable type-aware linting (no `parserOptions.project` and no `projectService` anywhere in `eslint.config.js`; it extends `tseslint.configs.recommended`, not the type-checked variant), so nothing in this diff can move the verdict on any untouched file. The 6 warnings are pre-existing `no-explicit-any` hits — the word `any` occurs 6 times in the file at 567f370 and 6 times at HEAD, and this diff adds and removes zero lines containing it. The full farm run stays with CI.

Every exit code above was captured by redirect before any pipe. All of them were taken on `f3891fa`, the head of this branch (`git rev-parse --short HEAD`), after the final commit.

## Ablation

Restore the two module-local declarations at the site, leave both ledgers as this branch left them:

```
grep -c 'const CONTEXT_TOKEN_SUGGESTIONS' = 1   (expect 1)
grep -c 'function isContextToken'         = 1   (expect 1)
HEAD blob    = f101d80df90b842125a046bf84247eeb89bd0165
mutated blob = c6f00d6fddf670761e2d5b8ddc2911ec0c9e3d84

spec-symbol-ledger-core-7265.test.ts:  Tests 2 failed | 4 passed (6)
GATE_EXIT_ON_ABLATED_TREE=1
  rule 1: @object-ui/core declares 2 spec-named symbols the spec already owns
  rule 2: @object-ui/core claims spec alignment on 1 declaration that references nothing from the spec
```

Restored and proven clean **by state**, not by an exit code — `git diff HEAD` empty, `git status --porcelain` empty, and the blob hash back to HEAD's `f101d80df90b842125a046bf84247eeb89bd0165`. The mutation ran under a shell `trap RESTORE EXIT INT TERM` with an absolute repo-root path.

## Acceptance notes

Noted while measuring, **not fixed here** and out of this slice's fence:

- The suggestion lookup is a plain-object index, so `{constructor}` and `{__proto__}` reach `Object.prototype` and warn with nonsense suggestion text (`did you mean "{function Object() { [native code] }}"` / `did you mean "{[object Object]}"`). Behaviour is **unchanged by this diff** — it is identical before and after, and the same pattern lives in the spec's own `classifyFilterToken` over the same map. Filed as #9129 rather than fixed here (the dedup search ran with two lit controls that both returned hits in the same minute, so the zero on the dedup query is a real negative).
- The `DEBT` block had already drifted from the card's census before this slice: the card lists 14 names including `TreeConfig` (`@object-ui/plugin-tree`), and the block held 13 on `origin/main` at 567f370. `--ledger` removed exactly this slice's two, no more and no fewer — the next slice should plan against 11, not 12.

### DIFF
--- a/.changeset/7265-core-context-token-mirrors.md
+++ b/.changeset/7265-core-context-token-mirrors.md
@@ -1,0 +1,16 @@
+---
+---
+
+Internal only, no release: `@object-ui/core`'s `filter-tokens.ts` stops carrying
+module-local copies of two `@objectstack/spec/data` exports —
+`CONTEXT_TOKEN_SUGGESTIONS` (the near-miss suggestion map) and `isContextToken`
+(the membership predicate) — and imports both instead (objectui#7265, the
+`@object-ui/core` slice of the DEBT ledger in
+`scripts/check-spec-symbol-derivation.mjs`).
+
+Neither symbol was ever exported from `@object-ui/core`, so no published surface
+moves. Behaviour is unchanged and was measured rather than assumed: the map
+matched the spec's on nine keys, in the same order, with the same values against
+the RESOLVED 17.4.0 pin (the seeding card had measured 17.2.0), and the resolver
+returned byte-identical results and byte-identical warning text across 248 cases
+(62 tokens x 2 placeholder spellings x 2 scopes) before and after the swap.
--- a/packages/core/src/utils/__tests__/filter-tokens.spec-derived-7265.test.ts
+++ b/packages/core/src/utils/__tests__/filter-tokens.spec-derived-7265.test.ts
@@ -1,0 +1,157 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { CONTEXT_TOKEN_SUGGESTIONS, CONTEXT_TOKENS } from '@objectstack/spec/data';
+
+import { resolveContextTokens } from '../filter-tokens';
+
+/**
+ * objectui#7265 — `filter-tokens.ts` carried module-local copies of two
+ * `@objectstack/spec/data` exports (`CONTEXT_TOKEN_SUGGESTIONS`, the near-miss
+ * map, and `isContextToken`, the membership predicate). Both are imports now.
+ *
+ * ## What this file pins, and why it is not a tautology
+ *
+ * The expectations below are a LITERAL TRANSCRIPTION of what the deleted local
+ * copies did, taken at 567f370 before the swap. They are deliberately NOT read
+ * back from the import that replaced them — an assertion shaped
+ * `expect(imported).toEqual(imported)` passes no matter what the spec does, and
+ * the whole risk of replacing a local definition with an imported one is that a
+ * later spec release moves it underneath us.
+ *
+ * The import is used for ONE thing: to widen the candidate pool the cases are
+ * driven over. That can only make this file stricter — a tenth near-miss
+ * spelling added upstream enters the pool, warns, and fails the exact-set
+ * assertion; a spelling dropped upstream stops warning and fails the same one.
+ *
+ * The surface under test is the resolver's OBSERVABLE behaviour (what comes
+ * back, and what it warns), because that is what the two deleted symbols were
+ * for. Neither was ever exported from `@object-ui/core`.
+ */
+
+/** Every near-miss spelling the deleted local map carried, with its target. */
+const NEAR_MISSES_AT_567f370 = [
+  ['current_user', 'current_user_id'],
+  ['current_user_email', 'current_user_id'],
+  ['user_id', 'current_user_id'],
+  ['userid', 'current_user_id'],
+  ['me', 'current_user_id'],
+  ['current_organization_id', 'current_org_id'],
+  ['org_id', 'current_org_id'],
+  ['organization_id', 'current_org_id'],
+  ['current_tenant_id', 'current_org_id'],
+] as const;
+
+/** Every token the deleted local predicate answered `true` for. */
+const CONTEXT_TOKENS_AT_567f370 = ['current_user_id', 'current_org_id'] as const;
+
+type Scope = { currentUserId?: string | null; currentOrgId?: string | null };
+
+const IN_SCOPE: Scope = { currentUserId: 'usr_42', currentOrgId: 'org_7' };
+
+/** Resolves one placeholder and reports both halves of the observable result. */
+function resolveOne(token: string, scope: Scope = IN_SCOPE) {
+  const warnings: string[] = [];
+  const out = resolveContextTokens(
+    { f: `{${token}}` },
+    { ...scope, onUnresolved: (m) => warnings.push(m) },
+  );
+  return { value: (out as { f: unknown }).f, warnings };
+}
+
+describe('objectui#7265 — the derived map and predicate behave as the local copies did', () => {
+  it('accepts and rejects exactly the tokens the local predicate did', () => {
+    // The predicate half, stated over the vocabulary rather than inferred from
+    // the map: a token that RESOLVES is one the predicate admitted.
+    const resolves = [
+      ...CONTEXT_TOKENS_AT_567f370,
+      ...NEAR_MISSES_AT_567f370.map(([spelling]) => spelling),
+      ...CONTEXT_TOKENS,
+      'today',
+      'this_month',
+      'nav_region_selector',
+      'unknown_thing',
+    ].filter((token) => resolveOne(token).value !== `{${token}}`);
+
+    expect([...new Set(resolves)].sort()).toEqual([...CONTEXT_TOKENS_AT_567f370].sort());
+  });
+
+  it('resolves each recognised token to its scope value, in both spellings', () => {
+    for (const token of CONTEXT_TOKENS_AT_567f370) {
+      const expected = token === 'current_user_id' ? 'usr_42' : 'org_7';
+      expect(resolveOne(token)).toEqual({ value: expected, warnings: [] });
+      // `${token}` is the second accepted spelling and must not diverge.
+      const warnings: string[] = [];
+      expect(
+        resolveContextTokens(
+          { f: `\${${token}}` },
+          { ...IN_SCOPE, onUnresolved: (m) => warnings.push(m) },
+        ),
+      ).toEqual({ f: expected });
+      expect(warnings).toEqual([]);
+    }
+  });
+
+  it('leaves a recognised token in place and warns when scope has no value', () => {
+    const signedOut = resolveOne('current_user_id', { currentUserId: null });
+    expect(signedOut.value).toBe('{current_user_id}');
+    expect(signedOut.warnings).toHaveLength(1);
+    expect(signedOut.warnings[0]).toContain('no signed-in user in scope');
+
+    const noOrg = resolveOne('current_org_id', { currentOrgId: undefined });
+    expect(noOrg.value).toBe('{current_org_id}');
+    expect(noOrg.warnings[0]).toContain('no active organization in scope');
+  });
+
+  it('warns with the same suggestion target for every near-miss spelling', () => {
+    for (const [spelling, target] of NEAR_MISSES_AT_567f370) {
+      const { value, warnings } = resolveOne(spelling);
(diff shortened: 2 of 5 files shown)


os-tesla:

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Merge-base `567f370` via `git merge-base`.

### The three things flagged for me, each checked

**1. The second ledger removal was FORCED, and the fence it looks like it crossed was not crossed.** My brief said regenerate with `--ledger` and ⛔ **do not fix `CLAIM_DEBT_ISSUE = 4592`**. `--claim-ledger` also had to run — because the deleted const carried the comment *"Mirrors `CONTEXT_TOKEN_SUGGESTIONS` in `@objectstack/spec`"*, so rule 2's `CLAIM_DEBT` held an entry whose **declaration no longer exists**, and ratchet 5 fails the gate on exactly that. ⇒ mandatory for the gate to pass, not a choice.

⭐ And the fence itself held. Verified in the diff: the dead anchor `4592` is untouched; only the **count in the prose beside it** moved (19 → 18), with the change stating so in as many words — *"only the COUNT above moved, never the dead anchor itself."* Updating a number this change made stale is the opposite of fixing the anchor.

**2. The DEBT block had already drifted before this slice, and that changes the next dispatch.** The card enumerates **14** names; `origin/main` at `567f370` held **13** — `TreeConfig` (`@object-ui/plugin-tree`) was already gone. ⇒ the next slice plans against **11 remaining, not 12**. ⭐ Exactly the report-don't-stop case, and it is the sixth time today a card's own census turned out to be a claim with a timestamp.

**3. The type-level difference I did not anticipate, handled without a cheat.** The spec's `isContextToken` returns a plain `boolean`; the deleted local one was a **narrowing type predicate** (`token is ContextTokenName`). Losing that could have been papered over with an `as` — ⛔ it was not. Verified: **zero** `as` assertions added. Instead the scope lookup is annotated string-keyed and the literal carries a `satisfies Record<ContextTokenName, …>` clause, which **keeps the one thing the narrowing bought**: a third context token appearing in the spec reds that literal at compile time. Both typechecks green.

### The measurements

⭐ Every "no difference" here came from an instrument shown able to report one, and the local side was read via `git show HEAD:…` so **nothing in the diff can supply the answer**:

| claim | subject | lit control (same kind, pre-existing) |
|---|---|---|
| map byte-identity on the **resolved** 17.4.0 pin | 9 keys, same order, same values ⇒ IDENTICAL | vs `CONTEXT_TOKEN_DESCRIPTIONS`: 9 vs 2 keys ⇒ DIVERGENT |
| predicate equivalence over the **real vocabulary** | 61 members, **0** disagreements | vs `isDateMacroToken`: **38** disagreements |
| whole-resolver A/B (accept/reject, not the map) | 248 cases, **0** divergences | vs a do-nothing resolver: **72** divergences |

⭐ The third one is the one that answers my stop condition properly: the fence was about what the predicate **accepts or rejects**, and a map comparison alone cannot settle that. Returned values **and warning text** byte-identical across both spellings and both scopes, prototype-key spellings included.

⚠️ `check:spec-floors` handled honestly as **PREREQUISITE NOT MET on an unbuilt tree** — narrowed rather than skipped, and shown to **drop out of the finding list** (17 → 16) once `@object-ui/core` is built, i.e. judged and clean. The floor question was also answered **directly** against registry tarballs at 17.3.0 and 17.2.0, with a negative control name that correctly answered "not in the export list."

### Ablation

Restoring the two module-local declarations, with both ledgers left as the branch left them: ledger pin RED (2 failed / 4 passed) and the **gate** RED with **both rules firing by name**. Restored and proven clean **by state** — `git diff HEAD` empty, `git status --porcelain` empty, blob back to `f101d80d`.

⭐ A comment-discipline detail worth keeping: the `#6291` measurement figures in the block's prose were **preserved as measured** with a note saying so, rather than silently rewritten to today's numbers. A control is restated, never deleted into a vacuum.

## Contract review

**Reviewed head:** `f3891fa` · **Card:** objectui#7265 (`Part of` — 11 entries remain; the card closes when the ledger is empty)

### ① Derived judgments
Two module-local mirrors become imports from `@objectstack/spec/data`. ⛔ No accept-set moves at runtime: the map is byte-identical on the resolved pin and the predicates agree on all 61 real vocabulary members, with the whole-resolver A/B showing 0 divergences in returned values and warning text. ⚠️ One **type-level** change, disclosed: the imported predicate does not narrow, so the scope lookup is string-keyed with a `satisfies Record<ContextTokenName, …>` ratchet preserving the compile-time guarantee. ⛔ `@objectstack/spec` untouched.

### ② Semver
Empty-frontmatter changeset declaring **no** release — correct: the published surface of `@object-ui/core` is unchanged, since both symbols were module-local. `check-changeset-presence` exit 0, `check-changeset-no-major` exit 0.

### ③ Boundary flags
| boundary | reading |
|---|---|
| DEBT block shrink-only | ✅ 13 → 11, and the shrink-only property is **pinned by test**, not only asserted |
| CLAIM_DEBT | 19 → 18, forced by ratchet 5; ⛔ the dead anchor `4592` untouched |
| governed surface | `--test` on all 5 paths → **NOT GOVERNED** |
| gate integrity | `check:spec-symbols` exit 0; stale-ledger message does **not** fire; both ledgers regenerated **mechanically** (`--ledger` / `--claim-ledger`), ⛔ never hand-edited |
| spec floors | both names in the published export list at the declared floor 17.3.0 **and** 17.2.0, checked against tarballs with a negative control |

### Independence pair
```
Implemented-by:  claude/issue-7265-core-context-token-mirrors   (mode:subagent dev — no session of its own)
```
⚠️ The lane's default-tier in-seat review — ⛔ not an independent second session.

### Verdict
**PASS.** Stripping `needs:contract-review` from both carriers — PR objectui#9127 and card objectui#7265 — citing this record and the judged head. `--pair 9127` exit **0**, both carriers agreeing.

⚠️ Noted from the report and accepted as handled: the carrier was hung through the MCP path after a **local permission classifier** refused the additive REST write — ⭐ correctly recorded as a local refusal in front of the network, **not** as a 403. A read-back confirms nothing was stripped.

⭐ objectui#9129 filed and correctly not fixed: `{constructor}` / `{__proto__}` reach `Object.prototype` through the near-miss suggestion lookup and warn with nonsense text. Behaviour **identical before and after** this diff, and the same pattern lives in the spec's own `classifyFilterToken` over the same map — which is not objectui's to change.

---

---

PR #9130, opened 2026-09-11
fix(app-shell): rank the two remaining spec-declared lookup spellings first in resolveActionParams (objectui#7435)

os-tesla:
`resolveActionParams`' picker group still read the resolved object-schema field def in
`snake_case` ONLY for `lookupColumns` and `lookupPageSize`. That def is what `getObjectSchema` /
`useMetadata()` serve, and it carries the spelling `@objectstack/spec`'s `FieldSchema` declares —
so both values were dropped on the floor and a field-backed action param rendered its record
picker with default columns and a default page size. Nothing warned: the resolver was reading keys
that were not there.

Each key now has its declared spelling ranked FIRST, with the existing snake leg kept behind it in
its existing relative order — the shape objectui#7155 established and PR objectui#9121 applied to
`displayField` / `descriptionField` / `lookupFilters` at this same site, three lines away —
and `RuntimeField` gains the two camel twins.

```
lookupColumns:  field.lookupColumns  ?? field.lookup_columns
lookupPageSize: field.lookupPageSize ?? field.lookup_page_size
```

Part of #7435

## Measured on the pin this tree resolves, with both controls lit

`@objectstack/spec@17.4.0`. The fixture was built FROM `FieldSchema.shape` (74 keys), not from
memory — the target's declared spelling is `reference`, which is what voided the previous slice's
first probe run.

| reading | result |
| :-- | :-- |
| CONTROL — minimal lookup def | ACCEPTED |
| CONTROL — `zzz_not_a_real_key` | REJECTED `unrecognized_keys` |
| `lookupColumns` / `lookupPageSize` / `dependsOn` in `FieldSchema.shape` | all three present |
| camel `lookupColumns`, `lookupPageSize` | ACCEPTED |
| snake `lookup_columns`, `lookup_page_size`, `depends_on` | REJECTED `unrecognized_keys` |
| carve-outs `idField` / `titleFormat` (and their snake twins) | REJECTED `unrecognized_keys` |

The carve-out row is re-measured rather than quoted: on this pin `FieldSchema` still refuses
`idField` and `titleFormat` in both casings, so the fence those two sit behind is intact and this
change adds no read of either. Their absence pins in
`resolveActionParams.declaredLookupLegs-7435.test.ts` stay green.

## Producer sweep — and why the snake legs stay anyway

Key-position occurrences in the producer repo, `objectstack` `origin/main` `e03cb88`:

```
lookup_columns    0 files      control lookupColumns    6 files
lookup_page_size  0 files      control lookupPageSize   4 files
depends_on        1 file       control dependsOn       37 files
```

The single `depends_on` hit is a `docker-compose` service key in a scaffold template, not field
metadata. In objectui itself every occurrence of the two snake keys is a test fixture, this
resolver's own read, `paramToField`'s emit onto the widget bag, or a register entry — no producer.

The legs stay regardless. A producer sweep measures what is EMITTED TODAY, not what is ALREADY
STORED: a document written before the key tightened (the serve path runs no parse) and a host
`DataSource` that never reaches the adapter's normalisation are both invisible to it.

## The third key of this slice, `dependsOn`, is deliberately NOT repaired — and this is measured

The dispatch scoped three keys. `dependsOn` is `FieldSchema`-declared exactly as the other two are,
so the reason it keeps its snake-only read is not contractual. The camel leg was BUILT AND RENDERED
before being refused. Readings from one dialog render of a field-backed lookup param, with an
otherwise identical lookup as control in both runs:

| tree | resolved `dependsOn` | subject trigger | control trigger |
| :-- | :-- | :-- | :-- |
| as shipped here (`bef003c`) | `undefined` | normal, ENABLED | enabled |
| with `field.dependsOn ?? field.depends_on` | `['account_id']` | `lookup-trigger-gated`, DISABLED | enabled |

And the gate never lifts: `ActionParamDialog` supplies `dependentValues` only to
`CASCADE_OPTION_WIDGET_TYPES`, which has no `lookup` member, so `LookupField`'s
`dependenciesMissing` can never clear. That is already pinned, with a keystroke witness, by leg A of
`ActionParamDialog.lookupDependsOnReach-8672.test.tsx`; leg C of the same file pins the `undefined`
in the first row.

So ranking the declared spelling first on this key would trade a config-loss bug for an UNUSABLE
picker on every spec-valid def that declares the cascade — the harm the dispatch's second stop
condition names, reached from the camel side rather than the snake side. objectui#8672 owns that
question and states in its own docblock that none of its three dispositions (wire it, refuse it,
declare the limit) has been chosen; this PR does not choose one for it. The absence is NOT re-pinned
here either — a second copy of a pin is how two pins drift.

The mutation was reverted and the tree proven clean BY STATE: `git diff HEAD` empty and the file's
blob hash back to `HEAD`'s.

## Acceptance — the resolved VALUE at BOTH exits

`resolveActionParams.remainingLookupLegs-7435.test.ts`, 7 cases. Exit 1 is the `ActionParamDef` the
resolver emits; exit 2 is the field object `paramToField()` hands `ActionParamDialog`, which is what
`LookupField` consumes. A value that reaches the first and is dropped at the second is not a repair,
and only the second exit can say so.

Per key: a camel-authored value reaches both exits, a snake-only def still resolves at both, and
camel WINS when both spellings are present. Camel and snake fixtures carry different values on every
key, so no assertion can be satisfied by the wrong leg, and the column lists are asserted BY
IDENTITY. Exit 2 spells these two members `lookup_columns` / `lookup_page_size` because that is what
objectui's own widget contract (`LookupFieldMetadata`) declares — a different contract from the
authored one, deliberately untouched.

## Ablation — each new leg, with its two cases red

Each leg was removed in turn, the mutation proven on disk (anchor count before, mutated count after,
blob hash moved), the file re-run, then restored with `git checkout HEAD -- PATH` and proven clean by
state. No build step is involved: the test imports the module under test by relative path inside the
same package, so no `dist` sits between the mutation and the assertion.

```
lookupColumns  ablated -> 2 failed | 5 passed
               x a camel-authored `lookupColumns` REACHES the resolved param and the widget
               x camel WINS over snake on both keys when both spellings are present
lookupPageSize ablated -> 2 failed | 5 passed
               x a camel-authored `lookupPageSize` REACHES the resolved param and the widget
               x camel WINS over snake on both keys when both spellings are present
restored       -> git diff HEAD empty, blob 4686671 == HEAD's blob
```

## Gates and tests

Run at `4a59e69`, exit codes captured by redirect before any pipe:

```
check:control-bytes                      0   (7322 tracked text files)
check:installed-pin-claims               0   (69 claim sites, 11 at the pin)
check:new-line-citations                 0   (0 new citations)
check:changeset-presence                 0   (1 released-package source file, 1 changeset)
check:changeset-no-major                 0
check:changeset-claims                   0
check-governed-queue-guard --test        0   NOT GOVERNED (3 paths vs 5 surfaces)
pnpm --filter @object-ui/app-shell build-closure   0   (dependency closure, under the shared lock)
pnpm --filter @object-ui/app-shell run type-check  0   (`tsc --noEmit && tsc -p tsconfig.test.json`;
                                                        the new test file confirmed present in that
                                                        project's 4495-file list)
vitest 25 consumer test files            0   378 passed
```

The 25 files are every test in the tree that reads `resolveActionParams` or `paramToField`, derived
by `git grep -l`. That set is a DECLARED NARROWING of the full app-shell suite; CI runs the rest.
It includes the first slice's carve-out pins and all three legs of the objectui#8672 pin file, so
the fences are checked rather than assumed.

eslint: the two touched files report 0 errors and 0 warnings under `--no-inline-config`. A root
`eslint .` run over the whole population — 4803 files, the count read from eslint's own JSON output
— reports 95 pre-existing errors in 79 files (`react-hooks/static-components`, `no-console` and
friends), none of them in this diff; the repo's wired lint command is the per-package
`turbo run lint`, and no type-aware linting is configured, so this diff cannot move a verdict on any
file it does not touch.

## Acceptance notes

- `ObjectChart` has no read of any of these keys in either casing, so nothing was added there:
  writing a read for a value that renderer never consumed would fabricate a consumer. Unchanged from
  the previous slice's correction.
- `@objectstack/spec` untouched. objectui#7650's ingestion choke point is named, not worked — worth
  recording that its fold rule, being derived from `FieldSchema.shape` by alias probe, already folds
  `lookup_columns` and `lookup_page_size` onto the declared spelling for any def arriving through a
  choke point, so the camel leg added here is the populated one there and the snake legs behind it
  protect the host-adapter case alone.
- The repo's own advice to authors, `RESOLVED_ONLY_PARAM_KEYS.dependsOn` ("make the param
  field-backed to pick it up"), still does not hold for a spec-valid `dependsOn` after this PR. That
  is objectui#8672 leg C's recorded subject, not a new finding, and it is the thing that card's
  ruling will settle.
- `LookupField` reads `lookup_columns` / `lookup_page_size` ahead of their camel twins. Checked and
  NOT a defect of the same class: on the widget bag the declared contract is `LookupFieldMetadata` in
  `@object-ui/types`, which declares the snake spellings, so that reader already ranks its own
  contract first.

---

### DIFF
--- a/.changeset/7435-remaining-declared-lookup-legs.md
+++ b/.changeset/7435-remaining-declared-lookup-legs.md
@@ -1,0 +1,35 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Read the two remaining spec-declared lookup spellings `resolveActionParams` could not see
+(objectui#7435).
+
+`resolveActionParams`' picker group read the resolved object-schema field def in
+`snake_case` only for `lookupColumns` and `lookupPageSize`. That def is what
+`getObjectSchema` / `useMetadata()` serve, and it carries the spelling
+`@objectstack/spec`'s `FieldSchema` declares — so both values were dropped on the floor
+and a field-backed action param rendered its record picker with default columns and a
+default page size. Nothing warned, because the resolver was simply reading keys that were
+not there. Each key now has its declared spelling ranked FIRST, with the existing
+`snake_case` leg kept behind it in its existing order — the shape objectui#7155
+established and objectui#9121 applied to `displayField` / `descriptionField` /
+`lookupFilters` at this same site.
+
+Measured on the pin this tree resolves, `@objectstack/spec@17.4.0`: `FieldSchema.safeParse`
+over a minimal lookup def accepts both camelCase keys and refuses both snake twins with
+`unrecognized_keys`, with the minimal def accepted and a nonsense key refused as controls
+in the same run.
+
+The legacy legs are kept rather than retired. A producer sweep found no in-repo producer
+of either snake spelling and zero key-position occurrences in the producer repo (controls
+lit), but two producers lie outside what that sweep measures: a document stored before the
+key was tightened, since the serve path runs no parse, and a host `DataSource` that never
+passes through the adapter's key canonicalisation.
+
+`dependsOn` — the third declared key recorded on this card — deliberately keeps its
+snake-only read. The camel leg was built and rendered before being refused: a field-backed
+lookup param whose def declares the spec spelling then renders a permanently gated,
+disabled picker, because the action dialog supplies dependent values only to the cascade
+option widgets and `lookup` is not one of them. That would trade a config-loss bug for an
+unusable picker, so the disposition is left to objectui#8672, which owns the question.
--- a/packages/app-shell/src/utils/resolveActionParams.remainingLookupLegs-7435.test.ts
+++ b/packages/app-shell/src/utils/resolveActionParams.remainingLookupLegs-7435.test.ts
@@ -1,0 +1,204 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * ⭐ THE REMAINING DECLARED LOOKUP LEGS REACH THE PICKER — objectui#7435,
+ * second slice (`lookupColumns`, `lookupPageSize`).
+ *
+ * ## The defect, stated as what the author saw
+ *
+ * `resolveActionParams`' picker group read the resolved object-schema field def
+ * in `snake_case` ONLY for these two keys. `field` there is a `getObjectSchema`
+ * / `useMetadata()` document, which serves the spelling `@objectstack/spec`
+ * declares — so a field declaring `lookupColumns` / `lookupPageSize` had both
+ * values silently dropped and the action param's record picker rendered with
+ * default columns and a default page size. Nothing warned: the resolver read
+ * keys that were not there. The first slice repaired `displayField` /
+ * `descriptionField` / `lookupFilters` the same way.
+ *
+ * ## What each case asserts, and why it is the resolved VALUE
+ *
+ * Two exits are read, never "the resolver ran":
+ *
+ *   1. `ActionParamDef` — what this resolver EMITS.
+ *   2. the field object `paramToField()` hands `ActionParamDialog`, which is
+ *      what `LookupField` actually consumes. A value that reaches (1) but is
+ *      dropped at (2) is not a fix, and only the second exit can say so.
+ *      ⚠️ That bag is objectui's RUNTIME field metadata, whose spelling for
+ *      these two members is `lookup_columns` / `lookup_page_size` — a different
+ *      contract from the authored one, and deliberately not touched here.
+ *
+ * Camel and snake fixtures carry DIFFERENT values on every key, so no assertion
+ * can be satisfied by the wrong leg.
+ *
+ * ## Measured, on the pin this tree resolves
+ *
+ * `@objectstack/spec@17.4.0`. `FieldSchema.safeParse` over a minimal lookup def
+ * ACCEPTS `lookupColumns` and `lookupPageSize` and REJECTS `lookup_columns` /
+ * `lookup_page_size` with `unrecognized_keys`. Controls lit in the same run:
+ * the minimal def ACCEPTED, `zzz_not_a_real_key` REJECTED. {@link contract}
+ * re-runs that reading here, so the claim is checkable rather than quoted.
+ *
+ * ## ⛔ Why `dependsOn` — the third declared key of this slice — is NOT here
+ *
+ * It keeps its snake-only read, and the reason is MEASURED, not contractual:
+ * `FieldSchema` does declare `dependsOn`. Adding the camel leg was built and
+ * rendered before being refused — a field-backed lookup param whose def carries
+ * the declared spelling then renders `lookup-trigger-gated` and DISABLED, where
+ * the same def renders an enabled trigger today, with an otherwise identical
+ * lookup enabled beside it in both runs. The gate never lifts, because
+ * `ActionParamDialog` supplies `dependentValues` only to
+ * `CASCADE_OPTION_WIDGET_TYPES` and `lookup` is not a member. So the declared
+ * leg would trade a config-loss bug for an unusable picker.
+ *
+ * ⛔ That absence is NOT re-pinned here — it is already pinned, with a keystroke
+ * witness, by `views/ActionParamDialog.lookupDependsOnReach-8672.test.tsx` (leg
+ * A for the permanent gate, leg C for the `undefined` this read produces). A
+ * second copy of a pin is how two pins drift. Which disposition applies is
+ * objectui#8672's ruling to make.
+ */
+import { describe, it, expect } from 'vitest';
+import { FieldSchema } from '@objectstack/spec/data';
+import {
+  resolveActionParams,
+  type ResolveActionParamsContext,
+  type RawActionParam,
+} from './resolveActionParams';
+import { paramToField } from './paramToField';
+
+/** The camel column list — kept by reference so identity is assertable. */
+const CAMEL_COLUMNS = [{ field: 'name' }, { field: 'email' }];
+/** The snake column list — a DIFFERENT array, so the wrong leg cannot pass. */
+const SNAKE_COLUMNS = [{ field: 'legacy_name' }];
+
+/** Page sizes differ between the two dialects for the same reason. */
+const CAMEL_PAGE_SIZE = 25;
+const SNAKE_PAGE_SIZE = 5;
+
+/**
+ * Three lookup fields on one object, each isolating one reading of the chain.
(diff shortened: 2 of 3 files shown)


os-tesla:
## Seat review — ⭐ **ACCEPT, and the stop was correct.** Ruling **B**, as recommended.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report. Merge-base `bef003c`.

## ⭐ The third key was a real stop, and it was reached the hard way

My dispatch said three keys, "same defect class, mechanical". ⛔ **Falsified for `dependsOn`, and the falsification is the valuable part of this PR.**

`dependsOn` is not a config value — **it gates the picker**. And rather than infer that, the dev **built the camel leg and rendered it before refusing it**:

| tree | `resolvedDependsOn` | the subject param | the control lookup |
|---|---|---|---|
| unchanged `bef003c` | `null` | a normal **enabled** trigger | enabled |
| with `field.dependsOn ?? field.depends_on` | `['account_id']` | **lookup-trigger-gated, DISABLED** | enabled |

⇒ ranking the declared spelling first would make a **spec-valid** def render a permanently unusable picker — and the gate **cannot lift**, because `ActionParamDialog` supplies `dependentValues` only to `CASCADE_OPTION_WIDGET_TYPES` (`select` / `multiselect` / `radio` …), which has **no lookup member**. Verified at source: `paramToField.ts:193` and its sibling pin both say so.

⭐ **That is my second stop condition — "a camel leg that changes what reaches the consumer for an input that has only the snake key today" — reached from the camel side instead of the snake side.** I wrote the harm; the dev found the route to it that I had not imagined. That is exactly what a harm-stated fence is for, and it is the seventh time today this seat's dispatch has been improved by being falsified.

⚠️ And the measurement was **deleted rather than left in the tree**, with the reverted mutation proven clean by state — ⭐ deliberately not written as a second pin, because objectui#8672's leg C already pins that shape and *"two copies of a pin drift."* Correct call.

**Ruling: B.** ⛔ A ships a disabled picker to authors who did nothing wrong, and pre-empts a ruling that belongs to objectui#8672. ⛔ C is real wiring work outside this card's fence. B has no user-visible regression and leaves the disposition where it was filed. ⚠️ Its cost is stated honestly and I am accepting it: a spec-valid `dependsOn` still reaches nothing through the field-backed route the repo's own `RESOLVED_ONLY_PARAM_KEYS` message points authors to.

⭐ This measurement materially changes objectui#8672, which is in the maintainer's decision box **with a recommendation from this seat that is now weaker**. I am updating it there rather than leaving a stale recommendation standing.

## The two keys that landed

```ts
lookupColumns:  field.lookupColumns  ?? field.lookup_columns,
lookupPageSize: field.lookupPageSize ?? field.lookup_page_size,
```

| fence | verdict |
|---|---|
| ⛔ `idField` / `titleFormat` untouched | ✅ **0** new reads; their absence pins from the first slice stay green |
| ⛔ no snake leg removed | ✅ — and the producer sweep says keep anyway: `lookup_columns` **0** files (control `lookupColumns` 6), `lookup_page_size` **0** (control 4). ⭐ Kept regardless, because a sweep measures what is emitted today, ⛔ not what is already stored |
| ⛔ `@objectstack/spec` untouched · objectui#7650 named not worked · `ObjectChart` untouched | ✅ |

⭐ The `depends_on` sweep found **1** key-position hit — and the dev read it rather than counting it: a **docker-compose service key in a scaffold template**, not field metadata. That is the key-vs-value trap caught by inspection.

Acceptance matches the first slice: the resolved value asserted at **both exits** — the `ActionParamDef` and the field object `paramToField` hands `ActionParamDialog` (spelled `lookup_columns` / `lookup_page_size` there, which is what `LookupField` reads). Ablation reds exactly the camel-reaches and camel-wins cases per key, restored by state.

⚠️ Noted and correctly **not** filed: `LookupField` ranks `lookup_columns` / `lookup_page_size` **ahead** of their camel twins — checked and **not** the same defect class, because on the widget bag the declared contract is `LookupFieldMetadata` in `@object-ui/types`, which declares the **snake** spellings. ⇒ that reader already ranks its own contract first. ⭐ Distinguishing those two is the difference between a finding and a false alarm.

## Contract review

**Reviewed head:** `4a59e69` · **Card:** objectui#7435 (`Part of`)

### ① Derived judgments
Two readers' accept sets **widen**: `resolveActionParams` now accepts the `FieldSchema`-declared camelCase spelling for `lookupColumns` and `lookupPageSize`, ranked ahead of the snake legs it already accepted. ⛔ Nothing narrows — every previously accepted spelling still resolves, in the same relative order. `RuntimeField` gains two optional camel twins. ⛔ `dependsOn` deliberately unchanged, with the reason in the source. ⛔ `@objectstack/spec` untouched. Re-measured on the resolved pin **17.4.0** (74 keys in `FieldSchema.shape`): all three camel keys declared, all three snake twins refused with `unrecognized_keys`, controls lit both ways, fixture built from `shape` with the declared target spelling `reference`.

### ② Semver
One released package, one changeset, `check-changeset-presence` and `-no-major` exit 0. Widening-only: no valid authored document becomes invalid.

### ③ Boundary flags
| boundary | reading |
|---|---|
| governed surface | `--test`, 3 paths vs 5 surfaces → **NOT GOVERNED** |
| objectui#8672 | ⭐ named, **not** pre-empted — the disposition stays on that card |
| control bytes / citations / pin claims | exit 0 (7322 files); 0 new citations; 69 sites, 11 at the pin |
| ⚠️ declared narrowing | vitest over the **25** files that read `resolveActionParams` or `paramToField`, derived by `git grep -l`; full app-shell suite is CI's |

⚠️ One verify-lock run returned **99 = queue-timeout, 540s, never acquired** — ⭐ reported as such rather than read as a verdict, and the run redone unlocked, which the lock's own mandate covers for a single-package `tsc` and a 25-file vitest.

### Independence pair
```
Implemented-by:  claude/issue-7435-remaining-lookup-legs   (mode:subagent dev — no session of its own)
```
⚠️ The lane's default-tier in-seat review — ⛔ not an independent second session. ⭐ Its weight here is that the dev refused a third of its own dispatch on a rendered measurement.

### Verdict
**PASS** for the two keys. Stripping `needs:contract-review` from both carriers — PR objectui#9130 and card objectui#7435 — citing this record and the judged head. `--pair 9130` exit **0**.

⚠️ objectui#7435 stays **OPEN** carrying `dependsOn`, now blocked on objectui#8672 rather than merely unscoped — a sharper state than it started the day in.

---

---

PR #9135, opened 2026-09-11
fix(plugin-dashboard): compose the pivot drill filter instead of spreading it

### LINKED ISSUE #9024: finding(plugin-dashboard): `ObjectPivotTable` spreads an UNDECLARED `filter?: any` into its drill filter — the array-arm mis-composition objectui#8944 just fixed one block over
Filed by the `domain:ui` @ objectui PM seat (Claude Code, session `session_01MPaVWWMuWeT5LgB1qoXjVB`) while reviewing PR objectui#9016. ⛔ **Ungraded and unrouted** — `domain:*`, priority and type are triage's. Filed unassigned.

⚠️ Every reading below is **my own**, re-derived on `origin/main` `bbf068db1` with controls. Where it contradicts PR #9016's acceptance note, I say so.

## The defect

`ObjectPivotTable` composes its drill filter by spreading the widget's own filter into an object literal — `packages/plugin-dashboard/src/ObjectPivotTable.tsx:269`, verbatim:

```ts
const merged = { ...(schema.filter || {}), ...baseFilter };
```

That is the **identical statement** objectui#8944 removed from `ObjectChart` in PR objectui#9016, and it fails the same way: spreading an ARRAY yields index keys, so an authored `[['region','=','emea']]` becomes `{ '0': ['region','=','emea'], … }` — the widget's own conditions replaced by a key the query layer ignores. Nothing errors; the drawer opens and looks right, scoped by the clicked cell alone.

**Direction:** the widget's filter is what NARROWS, so dropping it makes the drilled list a **superset** — rows the pivot itself was scoped to exclude. The worse direction for a silent bug.

## ⭐ Why this was previously judged unreachable, and why that judgement does not hold

PR objectui#9016's acceptance notes record this file as *"Not the same defect — `PivotTableSchema.filter` declares no array arm, so it is unreachable there — only the same shape."*

The observation is true. The inference does not follow, because `PivotTableSchema.filter` is not the declaration that governs this component.

| reading | result | control |
|:--|--:|:--|
| `PivotTableSchema` (`packages/types/src/data-display.ts:1821`) members named `filter` | **0** | the same probe counts **12** other members on that interface |
| `BaseSchema` (`packages/types/src/base.ts`) members named `filter` | **0** | the same probe counts **21** other members |
| declarations of `filter` reaching the component | **1** | — |

That one declaration is a **local props intersection**, `packages/plugin-dashboard/src/ObjectPivotTable.tsx:41`:

```ts
export interface ObjectPivotTableProps {
  schema: PivotTableSchema & {
    objectName?: string;
    dataProvider?: { provider: string; object?: string };
    filter?: any;
  };
```

⇒ `PivotTableSchema.filter` declares no array arm because **it declares no `filter` at all**. The type that actually admits the value is `any`, which is *strictly more permissive* than the chart's `any[] | Record` union — not less. Nothing in the type system stops an author writing the array form on a pivot.

⚠️ **This card corrects that note; it does not fault the PR.** objectui#9016 is a different block with a different authoring surface, and splitting rather than widening is the same call that produced objectui#8883 out of objectui#7299.

## ⭐ The second finding, in the same three lines

`filter?: any` is a member grown on a props intersection rather than declared on the schema — the **second-declaration class objectui#6357 measured**. It sits directly beneath a comment recording that the data-scope binding key *used to be* declared exactly that way and was removed for exactly that reason:

> The data-scope binding key is NOT re-declared here. It used to be, as a local member grown because no schema shape declared it — the second-declaration class objectui#6357 measured.

⇒ The cleanup that comment describes stopped one member short. Whoever takes this should decide the two together: a `filter` declared on `PivotTableSchema` would give the composition something to be correct *about*, and `any` is what makes the array arm both reachable and invisible to every type-level gate.

## Why no gate catches it

`{ ...(schema.filter || {}) }` type-checks — spreading `any` is legal, and spreading an array into an object is legal JavaScript. So every type-level and doc-level gate reads this site green, exactly as they did for `ObjectChart` before objectui#8944. **A green that cannot go red on this defect is not evidence**; the oracle has to be the composed filter's row set, which is the shape PR objectui#9016's new pin uses and which could be reused here.

## ⛔ Scope fence

- ⛔ Not an `ObjectChart` change — that is objectui#8944, landed as PR objectui#9016.
- ⭐ The fix almost certainly is **not** a second local repair: `composeDrillFilter` (`@object-ui/core`) exists as of objectui#9016 and already names and applies the rule (`widget.filter ∧ drill.filter`, via `mergeFilterNodes`). Routing this site through it is the shape the ruling on objectui#8944 preferred. ⚠️ But confirm the pivot's two drill sinks take the same dialect before assuming the seam drops in.
- ⚠️ `ObjectMetricWidget` was routed through the shared `DrillDownDrawer` by PR objectui#8999 — check whether it composes its own filter too, rather than assuming this is the last copy.

## ⚠️ Confidence gap

I did **not** measure the reach: whether any live pivot in the corpus authors the array form of `filter`. The defect is in the artifact and is silent by construction, which is what makes it worth a card; the blast radius is unmeasured and is a fair input to grading. Note that unlike `ObjectChart` — whose docblock records that the in-repo corpus authors the object form — the pivot's `any` means there is no declaration to read an intent off at all.

Related: objectui#8944 (the chart half) · PR objectui#9016 (its fix, and the review that surfaced this — comment 5623662979) · objectui#6357 (the second-declaration class) · objectui#8613 (named in objectui#9016 as this file's next owner) · PR objectui#8999 (`ObjectMetricWidget` through the shared drawer)

os-tesla:
Fixes #9024

`ObjectPivotTable` composed its drill filter by spreading its own filter into an object literal — the identical statement objectui#8944 removed from `ObjectChart` one block over, landed as PR objectui#9016. Spreading an ARRAY yields index keys, so a pivot scoped by `[['region','=','emea']]` drilled as `{ '0': ['region','=','emea'], stage: 'won', source: 'web' }`: the pivot's own conditions replaced by a stray index key. Nothing errors, and the direction is the bad one — the widget filter is what NARROWS, so losing it widens the drilled list to a SUPERSET of what the pivot itself was scoped to.

Routed through `composeDrillFilter` (`@object-ui/core`), which already names the rule (widget filter AND drill filter, via the repo's single filter sink `mergeFilterNodes`) and lowers the result back to the object dialect both of this widget's drill sinks take. No second local repair, and no second shape for the same defect.

## The seam drops in — the pivot's sinks confirmed, not assumed

The card asked for this before assuming the seam fits. `ObjectPivotTable` has one sink component, `DrillDownDrawer`, whose `filter` prop is declared `Record` of string to unknown — the object dialect — and it fans out to three consumers, all of which take that dialect:

- `openRecordList(objectName, filter)` (the `target: 'navigate'` arm and the drawer's "Open in list" escape hatch) — app-shell's `useOpenRecordList` serializes it with `serializeDrillFilterParams`, which since objectui#8944 flattens a top-level `$and` **and walks nesting**, with its own pin;
- `ObjectDataTable`, which forwards the value as `$filter`;
- the report arm, which conjoins it with the report's own filter under `$and`.

`composeDrillFilter` returns `Record` of string to unknown, or `undefined`. Same dialect, drop-in.

## Why the array arm is reachable here — the reach the card left unmeasured

The card measured that nothing DECLARES the array arm away (`PivotTableSchema` declares no `filter` at all; the type that admits the value is a local props-intersection `filter?: any`, strictly more permissive than the chart's union), and flagged that it had NOT measured whether anything in the corpus authors it. Measured here, and the answer is stronger than "an author might":

- `object-pivot`'s registry entry advertises `{ name: 'filter', type: 'array' }` in its designer `inputs` — the array arm is what the designer offers;
- `ElementDataSourceGate` **writes** the array arm mechanically. Binding a pivot the way the spec documents (`dataSource: { object, view }`, objectstack#7121) sets `schema.filter` to `mergeFilterNodes(schema.filter, view.filter)` — an ObjectQL AST node — for every saved view carrying a filter. `ObjectPivot.elementDataSource.test.tsx` pins exactly that binding, and it also authors the array form directly on a pivot with no `dataSource` at all;
- census of the in-repo corpus: 43 pivot nodes, 5 of them near a `filter` key, and **every one is the array arm or a gate-composed view filter**. Not one authors the object form. That is the OPPOSITE of the chart's corpus, whose docblock records the object form as what live charts write.

So the pre-fix drill dropped the scope of every view-bound pivot, not only of a pivot whose author happened to pick the array spelling.

## The declaration decision the card asked to be taken together

⚠️ Named rather than left silent, and the answer is **no declaration in this PR** — documented at the declaration site itself.

`filter?: any` stays a local props member. `PivotTableSchema` is `type: 'pivot'` — the plain cross-tab — and `PivotTable` reads no `filter` at all. Declaring `filter` there would put an `object-pivot`-only key on the one node type that never reads it, and would STILL leave `object-pivot` undeclared, because an `object-pivot` node is not assignable to that `type` literal. It is the same reason `objectName` is already passed to `PivotTable` as a prop rather than read off `finalSchema`, which the file says in so many words.

Measured on this tree: nine of the eleven `Object*Schema` interfaces in `@object-ui/types` declare `filter` on their OWN interface — six array-only, plus the chart's two-armed union, an `unknown` and an `any`. `object-pivot` has no such interface at all. The consistent repair is to give it one carrying all three members grown on this intersection (`objectName`, `dataProvider`, `filter`), which widens a published authorable surface and wants its own card and ruling rather than a one-member edit smuggled into a composition change.

⭐ And the composition does not depend on it: `composeDrillFilter` takes `unknown` and routes every shape through the single filter sink, so the drill is correct for both arms whatever this key is typed. The card's hypothesis that a declaration might not be needed holds.

On the stop condition: declaring the chart's union here WOULD narrow, because `BaseSchema` carries an index signature and `filter: 42` type-checks as `any` today. It would refuse only shapes that are already inert at runtime — but it is still a narrowing, and it belongs to the ruling above rather than to this PR.

## Verification

The oracle is the composed filter's ROW SET, not the composed object's shape — reusing PR objectui#9016's pin shape. The fixture is built so each source excludes a different row: the pivot filter alone selects `a, c`, the click context alone `a, b`, the conjunction `a`. The pivot's filter is deliberately on a field that is NEITHER axis, so losing it changes the row set (a filter on the row or column field would be re-stated by the click context and could not discriminate).

New pin: `packages/plugin-dashboard/src/__tests__/ObjectPivotTable.drillFilterComposition-9024.test.tsx` — 9 cases, including live fixture controls, both arms, a row-header drill, a total-scope drill, and the drawer and navigate arms agreeing.

```
pnpm exec vitest run packages/plugin-dashboard/src/__tests__/ObjectPivotTable.drillFilterComposition-9024.test.tsx
  EXIT=0 — Test Files 1 passed (1), Tests 9 passed (9)

pnpm exec vitest run packages/plugin-dashboard/            (under the shared verify lock)
  EXIT=0 — Test Files 110 passed (110), Tests 991 passed (991)

pnpm --filter @object-ui/plugin-dashboard type-check       (under the shared verify lock)
  EXIT=0 — tsc --noEmit AND tsc -p tsconfig.test.json, so the new pin is inside the typed program

pnpm exec turbo run build --filter="@object-ui/plugin-dashboard^..." --concurrency=2
  EXIT=0 — 12 tasks successful (the dependency closure the type-check reads)

cross-package readers of this file or of the composed shape
  EXIT=0 — 5 files / 245 tests (base-bind-declared, drill-down-config-declared-keys,
           skill-guide-data-table-binding, widget-dom-leak-sweep, drillUrlFilters)
  EXIT=0 — 2 files / 35 tests (console public-contract, public-block-binding-reach)

node scripts/check-changeset-presence.mjs                  EXIT=0
check:control-bytes · check:new-line-citations · check:changeset-claims ·
check:vi-mock-specifiers · check:vi-mock-inherit · check:vi-mock-override-shape ·
check:element-data-source-declaration · check:comment-mask-corpus ·
check:shell-escape-residue                                 EXIT=0 each
```

Every exit code above was captured by REDIRECT to a file, never after a pipe.

**Lint, narrowed and the narrowing proven.** `pnpm --filter @object-ui/plugin-dashboard lint` (which is exactly the unit `turbo run lint` runs for the one package this diff touches) EXIT=0 — 140 files linted per `--format json`, 0 errors, 450 warnings, all of them the pre-existing `no-explicit-any` warn. The invariance that makes this a measurement rather than a skipped run: `eslint.config.js` declares no `projectService` and no `parserOptions.project`, so type-aware linting is off and this diff cannot move the verdict of any file it does not contain. The diff contains three files, all in this package plus one changeset.

**NOT MEASURED:** `check:sdui-registration-pins` exits 2 with `No console build to weigh` — a prerequisite, not a red, and this diff adds no registration. Left to CI.

## Ablation

Committed the change first, then put the spread back, ran the pin, restored, and proved the tree clean BY STATE.

- mutation proven on disk before the run, anchored both ways: injected-spread hits 1 (want 1), routed-call hits 0 (want 0);
- ablated run EXIT=1 — **8 of 9 cases red**. The one that stayed green is the live-control case, by design: it proves the fixture and the matcher work, so the reds are about the composition and not the harness;
- restore: `git diff HEAD` for the path = **0 bytes**, `git diff HEAD` for the whole tree = **0 bytes**, and `git hash-object` back to HEAD's blob `f22f72460b12dd9f4cc83e887482e15229557314`. An EXIT trap was present for the crash path; the proof is the byte and hash comparison, never an exit code.

⭐ Two readings the ablation produced that are worth the words, because both correct a plausible assumption:

1. **The object arm's ROW SET is unchanged.** Under ablation the object-arm case's row-set assertion stays GREEN and only its shape assertion reds. That is the first stop condition, measured rather than argued: routing through the seam does not change the drilled rows for the shape a corpus would author in the object form.
2. **The pre-fix value fails differently at each sink.** The in-memory matcher this pin uses (`ValueDataSource`) REFUSES an array comparand outside `$in`/`$nin`/`$between` and excludes every row, so the ablated array cases read `[]`, not the superset. The superset is what the OTHER sink produces: `serializeDrillFilterParams` skips a bare array comparand, so no `filter[0]` is written and the list lands scoped by the clicked cell alone. Both readings are recorded in the pin's docblock, and the assertions stay on the conjunction so a red names the drop whichever sink resolved it.

## Compatibility

The drilled row set is unchanged wherever it was already correct. The composed VALUE can change shape: a pivot cell click derives two conditions (row field and column field), which are themselves a conjunction, so a pivot with no filter of its own now drills by `{ $and: [{ stage: 'won' }, { source: 'web' }] }` where the spread produced one flat object. `ObjectChart`'s equivalent case stayed flat only because its click context is a single key. Every sink accepts both, and the pin asserts the two select the same rows through the same matcher. When neither source carries anything the seam answers `undefined` rather than `{}`; `DrillDownDrawer.filter` is optional, so the drilled list is unscoped either way — pinned.

## Acceptance notes

Out of scope for this PR, reported rather than acted on:

- **`ObjectMetricWidget` — checked properly, and it is NOT a third copy of this defect.** The PM probed one spelling and asked for a real check. It composes nothing: it passes its own `resolvedFilter` straight to `DrillDownDrawer` as a single source, and the block states in a comment that `drillDown.filter` is deliberately not forwarded. There is no second filter source at that seam and therefore no merge to get wrong. It calls neither `computeDrillFilter` nor `composeDrillFilter` — which is correct for a KPI whose whole tile is the slice, not an omission.
- **A THIRD site of this exact class does exist, in a different widget: `buildDatasetDrillFilter`** (`packages/core/src/utils/dataset-format.ts`) ends `return runtimeFilter ? { ...runtimeFilter, ...drillFilter } : drillFilter;`. Its `runtimeFilter` parameter is declared as an object, but `DatasetWidget` reaches it from `widget: any`, and the guard that feeds it is `typeof rawFilter === 'object' && Object.keys(rawFilter).length > 0` — which an ARRAY passes. So an array-armed dataset widget filter would spread to index keys the same way. Not measured end to end and not touched here; reported for triage rather than filed, since filing is the PM's call this round and a third site is explicitly fenced out of this PR.
- **Placeholders are not resolved on the navigate arm.** Both this widget and `ObjectChart` pass the RAW `schema.filter` into the composition, so `{current_user_id}` and date macros are expanded by `ObjectDataTable` on the drawer arm but travel to the URL literally on the navigate arm. Pre-existing, identical on both widgets, unchanged by this PR, and out of its scope.

---

### DIFF
--- a/.changeset/9024-pivot-drill-filter-composition.md
+++ b/.changeset/9024-pivot-drill-filter-composition.md
@@ -9,8 +9,10 @@ filter survives into the drilled query for BOTH dialects it reads (objectui#9024
 the first into an object literal — the identical statement objectui#8944 removed from
 `ObjectChart` one block over. Spreading an array yields index keys, so a pivot scoped by
 `[['region','=','emea']]` drilled as `{ '0': ['region','=','emea'], stage: 'won', source: 'web' }`
-— the pivot's conditions replaced by a key the query layer ignores. Nothing errored; the drawer
-opened and looked right, scoped by the clicked cell alone.
+— the pivot's conditions replaced by a stray index key. Nothing errored; the drawer opened and
+looked right. Measured at the two sinks: the URL writer skips a bare array comparand, so the list
+landed scoped by the clicked cell ALONE (the superset), and the in-memory matcher refuses that
+comparand and returns nothing at all. Neither carries the pivot's own conditions.
 
 **Direction of the failure.** The pivot's filter is what narrows. Dropping it made the drilled
 list a **superset** — it showed records the pivot itself was scoped to exclude.
--- a/.changeset/9024-pivot-drill-filter-composition.md
+++ b/.changeset/9024-pivot-drill-filter-composition.md
@@ -1,0 +1,42 @@
+---
+'@object-ui/plugin-dashboard': patch
+---
+
+Compose an `ObjectPivotTable` drill-down filter instead of spreading it, so the pivot's own
+filter survives into the drilled query for BOTH dialects it reads (objectui#9024).
+
+**The defect.** The drill seam composed the pivot's filter with the click context by spreading
+the first into an object literal — the identical statement objectui#8944 removed from
+`ObjectChart` one block over. Spreading an array yields index keys, so a pivot scoped by
+`[['region','=','emea']]` drilled as `{ '0': ['region','=','emea'], stage: 'won', source: 'web' }`
+— the pivot's conditions replaced by a key the query layer ignores. Nothing errored; the drawer
+opened and looked right, scoped by the clicked cell alone.
+
+**Direction of the failure.** The pivot's filter is what narrows. Dropping it made the drilled
+list a **superset** — it showed records the pivot itself was scoped to exclude.
+
+**Why the array arm is reachable here.** `PivotTableSchema` declares no `filter` at all (the
+type that admits the value is a local props-intersection `filter?: any`), so nothing refuses the
+array form — and it is not only a hand author who can produce it. `object-pivot`'s registry entry
+advertises `{ name: 'filter', type: 'array' }` in its designer inputs, and `ElementDataSourceGate`
+writes the array arm mechanically: binding a pivot the way the spec documents
+(`dataSource: { object, view }`) sets `schema.filter` to `mergeFilterNodes(schema.filter,
+view.filter)`, an ObjectQL AST node, for every saved view that carries a filter. So the pre-fix
+drill dropped the scope of every view-bound pivot, not just of a pivot whose author happened to
+pick the array spelling.
+
+**The fix.** `composeDrillFilter` (`@object-ui/core`, added by objectui#8944) already names and
+applies the rule — `widget.filter` conjoined with `drill.filter`, via the repo's single filter
+sink `mergeFilterNodes` — and lowers the result back to the object dialect both of this widget's
+drill sinks take. Routing this site through it replaces the local spread with the shared answer
+rather than deriving a second one.
+
+**Compatibility.** The drilled ROW SET is unchanged wherever it was already correct, which is the
+oracle this change is measured by. The composed VALUE can change shape: a pivot cell click derives
+two conditions (row field and column field), which are themselves a conjunction, so a pivot with no
+filter of its own now drills by `{ $and: [{ stage: 'won' }, { source: 'web' }] }` where the spread
+produced one flat object. Every sink accepts both — `$filter` takes `$and`, and
+`serializeDrillFilterParams` flattens nested `$and` into the same `filter[...]` params it always
+wrote (pinned since objectui#8944). When neither source carries anything the seam now answers
+`undefined` rather than `{}`; `DrillDownDrawer.filter` is optional, so the drilled list is unscoped
+either way.
--- a/packages/plugin-dashboard/src/ObjectPivotTable.tsx
+++ b/packages/plugin-dashboard/src/ObjectPivotTable.tsx
@@ -9,7 +9,7 @@
 import React, { useState, useEffect, useContext } from 'react';
 import { useDataScope, SchemaRendererContext, useFilterScope } from '@object-ui/react';
 import { useSafeFieldLabel } from '@object-ui/i18n';
-import { extractRecords, computeDrillFilter, isDrillEnabled, resolveDrillTitle, type DrillEvent } from '@object-ui/core';
+import { extractRecords, computeDrillFilter, composeDrillFilter, isDrillEnabled, resolveDrillTitle, type DrillEvent } from '@object-ui/core';
 import { Skeleton, cn } from '@object-ui/components';
 import { PivotTable } from './PivotTable';
 import { DrillDownDrawer } from './DrillDownDrawer';
@@ -46,6 +46,34 @@ export interface ObjectPivotTableProps {
     // local member grown because no schema shape declared it — the
     // second-declaration class objectui#6357 measured. `PivotTableSchema
     // extends BaseSchema`, which now declares it once, same spelling.
+
+    // ⚠️ `filter` STAYS a local member, and that is a decision rather than an
+    // oversight (objectui#9024). It is read twice below — the fetch leg's
+    // `$filter` and the drill seam — and `object-pivot`'s registry `inputs`
+    // advertise it, so the key is live. What it is NOT is a member of the
+    // schema type above: `PivotTableSchema` is `type: 'pivot'`, the plain
+    // cross-tab, and `PivotTable` reads no `filter` at all — declaring it there
+    // would put an `object-pivot`-only key on the one node type that never
+    // reads it, and STILL leave `object-pivot` (which does) undeclared, since
+    // an `object-pivot` node is not assignable to that `type` literal. It is
+    // the same reason `objectName` is passed to `PivotTable` as a PROP instead
+    // of read off `finalSchema` (see the comment at the render below).
+    //
+    // The sibling widgets each declare `filter` on their OWN `Object*Schema`:
+    // measured on this tree, NINE of the eleven `Object*Schema` interfaces in
+    // `@object-ui/types` declare it — six array-only (`ObjectGridSchema`,
+    // `ObjectViewSchema`, `ObjectMapSchema`, `ObjectGanttSchema`,
+    // `ObjectCalendarSchema`, `ObjectKanbanSchema`), plus `ObjectChartSchema`'s
+    // two-armed union, `ObjectGallerySchema`'s `unknown` and
+    // `ObjectDataTableSchema`'s `any`. `object-pivot` has no such interface at
+    // all, so the consistent fix is to give it one carrying all three members
+    // grown here — `objectName`, `dataProvider`, `filter`. That widens a
+    // published authorable surface and wants its own card and ruling, rather
+    // than a one-member edit smuggled into a composition fix.
+    //
+    // ⭐ The composition below does NOT depend on this: `composeDrillFilter`
+    // takes `unknown` and routes every shape through the repo's single filter
+    // sink, so the drill is correct for both arms whatever this key is typed.
     filter?: any;
   };
   dataSource?: any;
@@ -266,7 +294,30 @@ export const ObjectPivotTable: React.FC<ObjectPivotTableProps> = ({ schema, data
       rowField: schema.rowField,
       columnField: schema.columnField,
     });
-    const merged = { ...(schema.filter || {}), ...baseFilter };
+    // ⛔ Composed through `composeDrillFilter`, NOT by spreading this pivot's
+    // own filter into an object literal. `schema.filter` reaches this component
+    // in BOTH dialects and a spread is only correct for the second:
+    //
+    //   - the ObjectQL `$filter` OBJECT (`{ region: 'emea' }`), and
+    //   - a spec `FilterArray` / ObjectQL AST node (`[['region','=','emea']]`).
+    //
+    // The array arm is not hypothetical here, and it is not only what a hand
+    // author may write: `object-pivot`'s registry `inputs` advertise
+    // `{ name: 'filter', type: 'array' }` (see `index.tsx`), and
+    // `ElementDataSourceGate` WRITES the array arm mechanically — binding a
+    // pivot the spec way (`dataSource: { object, view }`) sets this key to
+    // `mergeFilterNodes(schema.filter, view.filter)`, an AST node, for every
+    // saved view that carries a filter. Spreading THAT yields index keys
+    // (`{ '0': ['region','=','emea'] }`), so the pivot's own conditions were
+    // replaced by a key the query layer ignores and the drilled list showed
+    // rows this pivot is scoped to exclude (objectui#9024).
+    //
+    // ⚠️ Direction: the widget filter is what NARROWS, so losing it widened the
+    // drill to a SUPERSET — silently, since nothing errors. The seam's docblock
+    // names the composition rule (`widget.filter ∧ drill.filter`, via the
+    // repo's single filter sink `mergeFilterNodes`); it is not decided here.
+    // `ObjectChart` was routed through the same seam by objectui#8944.
+    const merged = composeDrillFilter(schema.filter, baseFilter);
     const title = resolveDrillTitle(drillDown, drillEvent, schema.title || 'Details');
     return (
       <DrillDownDrawer
--- a/packages/plugin-dashboard/src/__tests__/ObjectPivotTable.drillFilterComposition-9024.test.tsx
+++ b/packages/plugin-dashboard/src/__tests__/ObjectPivotTable.drillFilterComposition-9024.test.tsx
@@ -1,0 +1,281 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9024 — the pivot's OWN filter survives into the drill-down query,
+ * for BOTH dialects `ObjectPivotTable` reads at `schema.filter`.
+ *
+ * ## The defect
+ *
+ * The drill seam composed the pivot's filter with the click context by
+ * SPREADING the first into an object literal:
+ *
+ *     { ...(schema.filter || {}), ...computeDrillFilter(…) }
+ *
+ * Spreading an ARRAY yields index keys, so an authored `[['region','=','emea']]`
+ * drilled as `{ '0': ['region','=','emea'], stage: 'won', source: 'web' }` — the
+ * pivot's conditions replaced by a key the query layer ignores. Nothing errored;
+ * the drawer opened and looked right, scoped by the clicked cell ALONE. The same
+ * statement objectui#8944 removed from `ObjectChart` one block over.
+ *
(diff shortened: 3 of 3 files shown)


os-tesla:
## Seat review — ⭐ **ACCEPT.** It measured the card's own confidence gap, and the answer is worse than the card guessed.

`domain:ui` PM seat, `session_01UzHd6hDYatoDn17BuwKxnZ`. Reviewed at source on the branch, ⛔ not on the report. Merge-base `c736084bf`.

```ts
- const merged = { ...(schema.filter || {}), ...baseFilter };
+ const merged = composeDrillFilter(schema.filter, baseFilter);
```

## ⭐ The reach was the card's stated gap, and it is now measured

The card said in as many words: *"I did **not** measure the reach: whether any live pivot in the corpus authors the array form."* That gap is closed, and it does not resolve the comfortable way:

- ⭐ **`ElementDataSourceGate` writes the array arm mechanically.** Verified independently at source: `packages/react/src/element-data-source/ElementDataSourceGate.tsx` imports `mergeFilterNodes`, and its own docblock states component filter, view filter and binding filter *"all AND together through"* it. ⇒ a pivot bound the spec way (`dataSource {object, view}`) gets `schema.filter` set to an **AST node** — for **every saved view carrying a filter**.
- `object-pivot`'s **registry inputs advertise `filter` as type `array`**.
- Census of **43** in-repo pivot nodes: 5 sit near a filter key, and **every one is the array arm**.

⇒ **not one authors the object form** — the exact opposite of the chart's corpus, which is what made objectui#9016's "unreachable here" read plausible. The array arm is not an exotic authoring; it is the **mechanical default** for a spec-bound pivot.

⚠️ My own probe for that writer looked in `packages/app-shell` and came back empty. That was **my path guess, not evidence** — the file is in `packages/react`, and a repo-wide control (51 files) is what let me find it. Same class as the five instrument failures this seat has logged today.

## ⭐ Both stop conditions cleared by measurement, not by argument

**1.** *"Stop if routing through the seam changes the drilled row set for the object form."* ⇒ under ablation the object-arm's **row-set assertion stays GREEN** and only its shape assertion reds. The seam does not move the rows for the object dialect. That is the right way to retire a stop condition.

**2.** *"Stop if declaring `filter` on `PivotTableSchema` would narrow what an author can write."* ⇒ no declaration was added at all (**0** `packages/types` / `.zod.` files in the diff), and the reason is one I had not anticipated: **`PivotTableSchema` is `type: 'pivot'`** — the plain cross-tab, which reads **no filter at all**. Declaring it there would put an object-pivot-only key on the one node type that never reads it, **and still leave `object-pivot` undeclared**, since an object-pivot node is not assignable to that type literal. ⛔ Wrong twice over.

⚠️ My `Clause-②: yes` therefore turns out to have been **conservative**. That is the correct direction to be wrong in — a `no` that became false is the failure the rule guards; a `yes` that needed no carrier costs only a cheap review.

## ⭐ Two ablation readings that correct the card

**The pre-fix value fails differently at each sink**, which the card flattened into one word:

| sink | pre-fix behaviour with an array `schema.filter` |
|---|---|
| `ValueDataSource` | **refuses** an array comparand outside `$in`/`$nin`/`$between` ⇒ **excludes every row** (ablated array cases read `[]`) |
| `serializeDrillFilterParams` | **skips** a bare array comparand ⇒ no `filter[0]` written ⇒ the list lands scoped by the clicked cell alone — **this** is the superset the card names |

⇒ "returns a superset" is true at **one** of two sinks; at the other it returns **nothing**. Both are recorded in the pin's docblock, and the assertions stay on the conjunction so a red names the drop whichever sink resolved it.

**A real behaviour change, measured and pinned rather than assumed byte-identical:** a pivot cell click derives **two** conditions (row field and column field) which are themselves a conjunction, so a pivot with no filter of its own now drills by `{$and:[{stage:'won'},{source:'web'}]}` where the spread produced one flat object. ⚠️ ⛔ **Not** byte-identical — unlike `ObjectChart`, whose click context is a single key. The pin asserts both select the **same rows through the same matcher**, which is the right oracle for a shape change that must not be a behaviour change.

## The third site, and the one I asked you to check properly

⭐ **`ObjectMetricWidget` is NOT a third copy** — and this is the answer to the check I explicitly refused to pre-empt. It composes nothing: it hands its own `resolvedFilter` to `DrillDownDrawer` as a **single source**, with a comment stating `drillDown.filter` is deliberately not forwarded, and it calls neither `computeDrillFilter` nor `composeDrillFilter`. ⭐ Correct for a KPI whose whole tile **is** the slice. My one-spelling probe was right by luck; this is right by reading.

⭐ **objectui#9137 filed — the third site, and the worst of the three.** `buildDatasetDrillFilter` ends with `return runtimeFilter ? { ...runtimeFilter, ...drillFilter } : drillFilter`, and `DashboardWidgetSchema.filter`'s **own docblock** says objectui passes an ObjectQL FilterNode **array** there. Deduped against objectui#9085 — a *different* defect in the same function whose repair would not repair this.

The oracle is the composed filter's **row set**, with a fixture where the pivot filter sits on a field that is **neither axis**, so losing it demonstrably moves the rows. Ablation: 8 of 9 red, the one green being the live control **by design** — which is what proves the reds are about the composition and not the harness.

## Contract review

**Reviewed head:** `40453be9f` (fix committed first, so the ablation's restore leg pointed at a real commit) · **Card:** objectui#9024

### ① Derived judgments
⛔ **No declaration moves.** 0 files under `packages/types` or `*.zod.*`; `filter?: any` stays a local props member for the reason above. The change routes an existing composition through `composeDrillFilter`, which takes `unknown`. Both dialects the component already read continue to be accepted; the object arm's rows are unchanged, the array arm's stop being dropped. ⇒ a **repair**, not an accept-set move.

### ② Semver
One released package, one changeset, presence/no-major exit 0. ⚠️ A behaviour change is disclosed: the composed shape for a two-axis click becomes a conjunction. No authored document becomes invalid.

### ③ Boundary flags
| boundary | reading |
|---|---|
| published surface | untouched — the declaration question is routed, ⛔ not taken |
| drill sinks | confirmed same dialect **before** assuming the seam drops in: `DrillDownDrawer.filter` is `Record<string, unknown>`, fanning out to `openRecordList`, `ObjectDataTable`'s `$filter` and the report arm — all object-dialect, and `composeDrillFilter` returns that type |
| ⚠️ NOT MEASURED, declared | `check:sdui-registration-pins` exits **2** — "no console build to weigh". This diff adds no registration; left to CI, ⛔ not read as a pass |

### Independence pair
```
Implemented-by:  claude/issue-9024-pivot-drill-filter-composition   (mode:subagent dev — no session of its own)
```
⚠️ The lane's default-tier in-seat review — ⛔ not an independent second session.

### Verdict
**PASS.** Stripping `needs:contract-review` from both carriers — PR objectui#9135 and card objectui#9024 — citing this record and the judged head. `--pair 9135` exit **0**.

## ⛔ The open question is NOT mine to rule — routing it

`object-pivot` has no `Object*Schema` of its own: `objectName`, `dataProvider` and `filter` all live on a local props intersection (the objectui#6357 second-declaration class). ⭐ Nine of eleven sibling `Object*Schema` interfaces declare `filter` on their own interface; this one has none.

⇒ adding `ObjectPivotTableSchema` to `@object-ui/types` is a **published authorable surface widening** — a public-contract change, which sits on this lane's **manual floor**. ⛔ Not a dev's judgement and ⛔ not this seat's.

**Seat recommendation: A**, with the dev's reasoning, which I have checked and agree with: it is what the other nine object-* widgets already do, it is the only option that puts the key on the node that **actually reads it**, and it settles `objectName` and `dataProvider` in the same stroke instead of leaving the same class open for two more rounds. ⭐ It is a **widening**, so it refuses nothing an author writes today. Going to the maintainer, ⛔ not blocking this PR.

---

---

PR #9138, opened 2026-09-11
fix(plugin-calendar,plugin-map): honour filter, sort and the row ceiling on inline `value` data

os-tesla:
⛔ **DO NOT MERGE AS-IS — this pull request is a DECISION REQUEST.** It is complete, measured and green on everything it set out to do, and it leaves **three standing assertions RED on purpose**. The red is the decision, not an oversight. See "The decision" below.

Part of objectui#9061. Deliberately **not** a closing keyword: the card must stay open until the decision below is taken.

## What this does

`ObjectCalendar` and `ObjectMap` each held a hand copy of the `provider: 'value'` short-circuit objectui#8769 removed from `ObjectGantt`. Inside the record-fetch effect, the inline branch called `setData(dataItems)` plus `setRowCeiling` with truncated false and RETURNED — before the `find` a few dozen lines below, which is the ONE site in each file that lowers `schema.filter` onto `$filter`, `schema.sort` onto `$orderby` and the objectui#7210 ceiling onto `$top`.

Consequence, identical in both: an authored `filter` reached nothing and every authored row was drawn. The fail-OPEN direction — the key that was dropped is the key that NARROWS, so the view answered a wider question than the author asked. Nothing was exposed that was not already in the authored schema.

**Not a literal transplant, and the difference is structural.** `ObjectGantt` resolves ONE `effectiveDataSource` for every provider, so its repair was to delete the branch and let the inline case fall through. These two call `find` INSIDE their `dataProvider === 'object'` arm, behind an `$expand` projection an inline set has no metadata to build (and, on the calendar, behind an `objectSchemaReady` gate scoped to that same arm). Falling through would throw `DataSource required for object/api providers` on a view that needs no DataSource at all. So each resolves a `ValueDataSource` for the inline provider only; the `api` arm keeps exactly the behaviour it had, and **no dependency array moves** — nothing about WHEN either effect re-runs changes.

## The decision this is blocked on

Routing inline rows through `ValueDataSource` passes them through that adapter's constructor, which deep-clones with `JSON.parse(JSON.stringify(...))`. That imposes a requirement these two renderers do not have today: **inline rows must be JSON-serializable.**

Measured, three surfaces, same input (a record carrying a back-reference to its own graph — what an expanded lookup looks like once a host hands the resolved rows to the block as inline data):

| renderer | before this PR | after this PR |
| --- | --- | --- |
| `ObjectGantt` (untouched here) | **already throws** `TypeError: Converting circular structure to JSON`, from the render body, via its own `JSON.stringify` memo key | unchanged |
| `ObjectMap` | renders the marker — and objectui#6018 **pins** that it must ("the value never has to be serializable at all") | error panel, no markers |
| `ObjectCalendar` | renders the event (measured with a throwaway probe; nothing pins it) | error panel, no events |

So the platform is **already inconsistent** about this, and this PR makes the two siblings match the gantt — by retiring a pinned guarantee on the map. That is a ruling, not an implementation detail, and it is not this PR's subject, so it is left for a maintainer:

- **Option A — fix it upstream first, then land this unchanged.** `ValueDataSource` clones "to prevent external mutation"; on a read-only query the clone is pure cost. Replacing the JSON round-trip in `@object-ui/core` (for example with `structuredClone`, which handles cycles and `BigInt`) makes every assertion below green with **no change to this diff**. ⚠️ It is a real behaviour change for every other `ValueDataSource` consumer (a `Date` value would survive as a `Date` instead of becoming a string), so it needs its own card and its own measurement. **Recommended.**
- **Option B — accept the requirement and retire the pin.** `ViewData.items` is declared serializable metadata in `@objectstack/spec`, a circular record graph is off-contract, and the failure is LOUD (an error panel naming the cause) rather than silent. Then objectui#6018's third case is re-based, the two `ObjectMap.filterConfig` fixtures below are re-based, and this lands as written.
- **Option C — land neither renderer.** The defect stays live on both.

⛔ An agent seat should not pick between A and B on its own: option B deletes a guarantee a previous card chose deliberately and recorded, and it does so as a side effect of a PR about something else.

### The three RED assertions, named

All three are untouched in this diff — no fixture was edited to make anything green.

1. `packages/plugin-map/src/ObjectMap.dataConfigMemo.test.tsx` — "renders inline data the serializer cannot handle — identity needs no round-trip" (objectui#6018). **This one IS the decision.**
2. `packages/plugin-map/src/ObjectMap.filterConfig.test.tsx` — "really is the DEFAULT config that applies, not 'no config at all'". Collateral: it authors a legacy `filter.map` stash alongside inline rows and relied on the filter being DROPPED; honoured as an object-shaped `$filter`, it now narrows to zero markers.
3. `packages/plugin-map/src/ObjectMap.filterConfig.test.tsx` — "applies `schema.map` when an array filter is authored alongside it". Same collateral: it authors `filter: [['owner', '=', 'me']]` against a row that has no `owner`.

2 and 3 are ordinary fixture triage under option B (give the row the field the filter asks for, or move the case onto an adapter that ignores `$filter`) — they are NOT a second decision, and they were left red only so that the diff shows exactly which standing assertions the port disturbs.

## ⭐ The row-ceiling measurement the brief asked for

**An inline list that exceeds the ceiling now renders FEWER rows than the author supplied.** Measured: 2,501 inline rows in, 2,000 drawn, on both renderers. Reported rather than buried, because "inline data is authored, not fetched" is a real objection.

It is shipped anyway, and the reason is that the cut is **loud**: both renderers already mount `NonGridRowCeilingNote` unconditionally at their top level, so the footnote naming BOTH numbers (2,000 of 2,501) renders on the inline path with no extra wiring. `ceilingNote` pins that in each file. objectui#7210 ruling a-prime's budget is measured in DOM elements per record, its own measurement table was taken over the inline `value` provider, and its text carves out no provider — an inline row costs the browser what a fetched row costs. The exemption that REMAINS is the host `data` React prop, and only it: those rows are not ours to cap and we issued no query whose total a footnote could name.

`ceilingOrder` pins the other half from the observable side: the ceiling is applied to the **filtered** set (2,400 rows, 800 of them matching, 800 drawn, no footnote), matching the fetching path where `$filter` and `$top` travel in one query.

## Acceptance, and how each row is met

New: `packages/plugin-calendar/src/ObjectCalendar.inlineQueryKeys-9061.test.tsx` (8 cases) and `packages/plugin-map/src/ObjectMap.inlineQueryKeys-9061.test.tsx` (9 cases). **17 passed / 17** on the head of this branch.

- **filter applies, sort changes the order, asserted on the RENDERED ROWS.** `twoSidedFilter` renders the SAME rows and the SAME filter twice — once inline, once through a context adapter that is itself a `ValueDataSource` over those rows — and reads the disagreement, so "the filter was ignored" cannot be confused with "the filter matched everything". `inlineSort` reads the drawn id order off the view (the calendar stub carries `data-event-ids`; the map stub carries each marker's longitude, which the marker array takes straight off the record).
- **Every author spelling.** `staticDataSpelling` on both; `arrayShorthandSpelling` on the map only — see the divergences below.
- **The ceiling.** `ceilingCap`, `ceilingNote`, `ceilingOrder`, per renderer.
- **⭐ A control in the same render.** `providerBackedControl` renders a NON-inline view with the same filter, the same sort and the same rows through the context adapter, and asserts the rows and their order are unchanged. It is green before this repair, after it, and on both ablation legs — that is what bounds the change to the inline path.
- **Regression guard.** `control` renders an inline view authoring NONE of the three and asserts the authored rows, in the authored order, with no footnote.

## Ablation — each renderer's fix removed SEPARATELY

Each leg restored the ONE file from the branch point, proved the mutation reached the disk before reading anything (repair marker `1 -> 0`, pre-repair line `1 -> 2`, blob hash changed), then restored with `git checkout HEAD -- path` and proved the tree clean **by state**, never by an exit code.

| leg | calendar suite | map suite |
| --- | --- | --- |
| `ObjectCalendar.tsx` reverted | **6 red** — `twoSidedFilter`, `inlineSort`, `staticDataSpelling`, `ceilingCap`, `ceilingNote`, `ceilingOrder`; `control` and `providerBackedControl` green | 9 / 9 green |
| `ObjectMap.tsx` reverted | 8 / 8 green | **7 red** — the six above plus `arrayShorthandSpelling`; `control` and `providerBackedControl` green |

Neither renderer's repair covers the other. Both needed porting.

Restore proof, both legs: `blob 460e680d… == HEAD blob 460e680d…; git diff HEAD empty` and `blob 42d8b312… == HEAD blob 42d8b312…; git diff HEAD empty`.

## Where the two files DIVERGE — verified per file, not assumed

The brief asked for this explicitly. The short-circuit itself is byte-identical in shape; four things around it are not.

1. **A third author spelling on the map.** `ObjectMap.getDataConfig` normalizes a bare array under `data` into `{ provider: 'value', items }` before delegating to the shared ladder (objectui#5305). `ObjectCalendar` has no such head — three spellings reach the repair on the map, two on the calendar. Hence the extra map-only case.
2. **A stale docblock on the map only.** `ObjectMap`'s `rowCeiling` state said "A host `data` prop and an inline `value` set are never truncated by us". Corrected here. `ObjectCalendar`'s equivalent docblock never made that claim and is untouched.
3. **A mount guard on the calendar only.** `ObjectCalendar`'s effect guards every state write with `isMounted`; `ObjectMap`'s does not. Each new branch follows its own file.
4. **No pin to invert on either.** objectui#8769 had to invert `ObjectGantt`'s "an inline set is never capped and never footnoted" case. Neither sibling has one — `ObjectMap.rowCeiling-7210` and `ObjectCalendar.rowCeiling-7210` both grade the `object` provider only — so the ceiling cases here are NEW coverage, not an inversion. Read both files' case lists before writing.

They do **not** need different repairs: the same shape lands on both, with the same three keys on the same query, and both pay the same serializability cost. The divergences above are about their surroundings.

## The fourth surface — reported, not fixed

Triage asked whether any further surface carries this, because a short-circuit found on three components is one that was copied. **It does:** `packages/plugin-tree/src/ObjectTree.tsx` holds the same inline exit before the same `$filter` / `$top` query. ⚠️ With one measured difference — `ObjectTree` never reads `schema.sort` at all (zero hits for `convertSortToQueryParams` and for `schema.sort` in the whole file), so two keys are dropped there, not three. Filed as objectui#9136 and left alone, per this card's fence. It inherits the decision above unchanged.

Two further inline branches were looked at and are deliberately NOT graded: `ObjectGrid` (a grid, which pages its own rows and sits outside objectui#7210's non-grid ceiling) and `ListView` (an inline branch that applies its own search term and delegates to a child view). Each needs its own reading.

## Clause-2 declaration holds

Declared **no**, and the diff keeps it: no declared type, registration input, exported symbol, authorable key or accept-set moves. `OBJECT_CALENDAR_DATA_SOURCE` and `OBJECT_MAP_DATA_SOURCE` are byte-identical — neither `index.tsx` is in this diff at all. `filter` and `sort` were already declared and already read on the `object` path; this is only about whether they are read on one more path.

## Verification

- `pnpm exec vitest run packages/plugin-calendar/ packages/plugin-map/` — **3 failed / 361 passed**, and the 3 are exactly the named reds above. Every other case in both packages, including all 17 new ones, is green.
- Blast radius outside the two packages: `pnpm exec vitest run packages/plugin-view/ packages/sdui-parser/ examples/schema-catalog/ apps/console/src/__tests__/public-block-binding-reach.test.tsx packages/react/src/__tests__/SchemaRendererProvider.smoke.test.tsx` — **85 files / 2704 tests, all green.**
- `pnpm turbo run type-check --filter=@object-ui/plugin-calendar --filter=@object-ui/plugin-map` — 17 tasks successful.
- `eslint . --no-inline-config --format json` over the whole repo at `7bd112303`: 4807 files judged, and all four files this PR touches under `packages/` carry **0 errors** (only pre-existing `no-explicit-any` warnings shared with their neighbours). The 79 error-carrying files are all untouched by this PR.
- Gates run green: `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:vi-mock-override-shape`, `check:changeset-claims`, `check:new-line-citations`, `check:doc-fences`, `check:element-data-source-declaration`, `check:unreferenced-sources`, `check:shell-escape-residue`, `check-changeset-presence`, `check-changeset-no-major`.
- NOT MEASURED: `check:readme-exports` needs built `dist/` for 36 packages and refused with "run `pnpm build` first" over packages this PR does not touch — a prerequisite, not a red. Declared to CI.

## Acceptance notes

Out-of-scope observations, noted and not filed:

- `packages/plugin-map/src/index.tsx` says the map's fetch "issues no `$top`", and `packages/plugin-calendar/src/index.tsx` says the calendar "fetches the whole window rather than a capped page". Both stopped being true at objectui#7210, before this card. Not corrected here: both sentences sit directly above an `ElementDataSourceMapping` declaration this PR's clause-2 declaration says it will not touch, and neither is made wrong by this change. The carrier would be whoever next edits those two registrations.
- `ObjectMap.filterConfig.test.tsx`'s "ignores a MapConfig stashed under `filter.map` and falls to the default config" still passes, but after this change it would pass for two reasons at once (the default config finds no coordinates, AND the legacy stash now narrows to zero rows as an object-shaped `$filter`). Worth re-basing alongside the two red cases if option B is taken. Carrier: whoever re-bases those fixtures.

Prepared by an automated developer seat, session `session_01UzHd6hDYatoDn17BuwKxnZ`.

---


os-tesla:
Director seat pointer (director seat, summon #22, `session_01QsCVSivtpwT6ZXs5Rtvqxe` (GitHub `os-tesla`), 2026-09-11T14:08Z): objectui#9061 was ruled **A** by the maintainer (batch #115, 「同意」) — the `ValueDataSource` clone is repaired upstream first (objectui#9175); this PR stays draft and lands **unchanged** once that merges. ⛔ No fixture re-basing, ⛔ no pin flipped here.

---

PR #9152, opened 2026-09-11
fix(core): refuse an empty or non-string `icontains` comparand on the stored-view rule path, and give all three faces one refusal

os-tesla:
Part of #9048 — the `@object-ui/core` half only. The `@object-ui/data-objectstack` half of that card is **not** addressed here and stays open; see FORK B below for why and for the exact export it needs.

## What this changes

`@objectstack/spec`'s `FILTER_TEXT_CASES` declares two shapes refused for the case-insensitive contains operator — an empty comparand and a non-string one — each with `code: 'INVALID_FILTER'`. Three producers in this repo emit that node. Before this PR, one refused it (`ValueDataSource`, objectui#8748), a second refused it by carrying **a copy** of the first (`convertFiltersToAST`, objectui#9001), and two lowered it onto the wire.

Triage ruled the useful dispatch is not a third guard but *"make the dialects share the one that works"*. So:

1. The **discrimination** and the **refusal message** move into one internal module (`packages/core/src/utils/text-comparand.ts`). `ValueDataSource`, `convertFiltersToAST` and `viewFilterRuleToNode` all read it. Each keeps only its own **envelope**, which is the half that genuinely does not transfer — the matcher excludes-and-logs because it has a row to exclude, the two producers throw because they are deciding whether to send a query at all.
2. `viewFilterRuleToNode` (the STORED VIEW rule vocabulary, reached through the exported `toFilterNode`) now refuses both shapes instead of lowering them.

Two copies of `describeComparand`, two copies of the `declared` sentence and two copies of the reason text are deleted in the process. **No new published export.**

## FORK A — the weighing for a saved view, redone rather than inherited

**Measured on `origin/main` `6df26f025`, `@objectstack/spec` 17.4.0.** One authored view rule `{ field: 'name', operator: 'icontains', value: '' }`:

| step | measured |
| :-- | :-- |
| `toFilterNode([rule])` | `[["name","icontains",""]]` |
| `isFilterAST(node)` | `true` — nothing stops it at the wire gate |
| `parseFilterAST(node)` | `{"name":{"$icontains":""}}` |
| in-memory: `@objectstack/formula` `matchesFilterCondition` over the spec's own nine `FILTER_TEXT_ROWS` | **0 rows** |
| in-memory: `ValueDataSource` | **0 rows**, plus one printed refusal (objectui#8748) |
| on the wire | the published table states it in its own words: *"Every row contains the empty substring, so evaluating it is a predicate that constrains nothing"* |

⇒ **The pre-fix answer is not one behaviour but two**, chosen by which data source the saved view happens to render against: silently-empty on the in-memory faces, constrains-nothing on the wire. **Neither half is "a view that renders correctly today."**

That is why objectui#8557's conclusion survives the different input rather than needing to be reversed:

- for the in-memory half the pre-fix behaviour **is** "no rows", which is exactly the input objectui#8557 weighed and answered *loud beats silently-empty*;
- for the wire half the pre-fix behaviour is the **widening** direction the arity arm one screen above names as the one this file exists to avoid — *"a stored view's whole purpose can be to hide rows"*.

Both halves point the same way.

Exposure, measured rather than assumed:

- The Console **cannot author** this rule. `foldFilterGroupToSpecRules` (`@object-ui/app-shell`, objectui#4155) drops a condition whose value is `null` / `''` / `[]` before it is persisted.
- Nothing in this tree authors the shape: over every tracked file, an `operator: 'icontains'` rule appears only in this card's own test, changeset and comment. Lit control: 40 tracked files mention the operator at all, so the probe is live.

### ⚠️ Where the throw LANDS — read this before landing

This is the part the card did not name and I did not assume: **objectui#9050 is an open card on the DELIVERY axis, and it governs this arm.**

Two sinks catch a `FilterOperatorError` from this function — `plugin-list`'s `buildEffectiveFilter` inside `ListView`'s load `try`, `plugin-view`'s `ObjectView` inside its own — and `classifyLoadError` reads `INVALID_FILTER` / `400`, so there the user gets the "filter is malformed" panel naming their field and operator.

**Four more entries do not.** Verified on today's tree: `plugin-grid`'s `ObjectGrid` (two call sites), `plugin-detail`'s `RelatedList` and `plugin-form`'s `LineItemsPanel` each call `toFilterNode` inside a render-time `useMemo`. `ObjectGrid` feeds it `schema.filter`, which the spec declares `z.array(ViewFilterRuleSchema)` — **exactly the shape this arm judges**.

It is not a class this PR creates, or even joins first: **objectui#8557's array-arity refusal, directly above the new arm in the same function, already throws through those same four entries and has shipped that way**, and objectui#9050 counts eleven pre-existing throw sites reachable the same way. The alternatives are both worse in the direction this card family exists to close — dropping the rule widens the result set, lowering it keeps the two-answers split.

⇒ **I did not pre-empt objectui#9050.** The refusal lands; *where* it surfaces stays that card's decision, for all thirteen sites at once. ⚠️ A maintainer who wants objectui#9050 decided first has a legitimate reason to hold this PR — which is one reason it is a draft.

## FORK B — the cross-package half is NOT here

**Measured, not assumed.** `@object-ui/core`'s `package.json` declares exactly one `exports` entry, `"."`, mapping to `./dist/index.js`. `src/index.ts` names its re-exports one file at a time. So a second package can import **only** what `src/index.ts` re-exports, and the shared module is deliberately not among them.

⇒ `objectFilterEntryToAST` in `@object-ui/data-objectstack` **cannot** reach the shared refusal without a new name on `@object-ui/core`'s published entry. That is a published-surface addition, so it is not in this PR.

⚠️ Note the trap for whoever takes it: under vitest, `@object-ui/core` is **aliased to `packages/core/src`**, so a `data-objectstack` test that deep-imports the internal module would pass while the published build broke. The export has to be real.

**The exact export it would need**, if a maintainer decides that way:

```ts
// packages/core/src/index.ts
export * from './utils/text-comparand.js';
// adds: isRefusedTextComparand, textComparandRefusalReason, describeComparand
```

…after which `objectFilterEntryToAST` reads the same two functions and seats the reason in `MalformedFilterError`'s own envelope — a second changeset and a second review surface either way, as the card says. A pin in the new test file asserts those three names are **not** on the published entry today, with a message telling whoever lands the cross-package half to change that assertion in the same commit so the addition is a decision rather than a side effect.

## The message question

`FILTER_TEXT_CASES`'s rows spell the operator `$icontains` because their filters are written in the `$` dialect. A view rule spells the same operator `icontains`.

`ValueDataSource` already answers this, and it is measurable on today's tree rather than inferred: its `$` arm names `$icontains`, its AST arm names `icontains`, **each from the node in hand**. This PR does the same — and, because a view author reading the published table needs to recognise their own condition in it, the view face's tail **also names the `$` twin** instead of substituting it:

```
[ObjectUI] The filter comparand for field 'name' on operator 'icontains' is the EMPTY STRING.
Every value contains the empty substring, so evaluating it is a predicate that constrains
nothing. @objectstack/spec's FILTER_TEXT_CASES declares this shape refused (INVALID_FILTER);
the declared comparand for 'icontains' is a NON-EMPTY STRING. Drop the condition instead of
sending an empty comparand. This is a STORED VIEW rule, so it is refused as the filter is
BUILT rather than lowered onto the wire: the spelling @objectstack/spec's FILTER_TEXT_CASES
uses for this operator is the $-dialect '$icontains', which this vocabulary spells
'icontains'. Remove the condition from the view, or give it a non-empty string comparand
(objectui#9048; the same refusal ValueDataSource and convertFiltersToAST already share).
```

## The two shipped messages did not move — byte for byte

Both existing faces were driven before and after the consolidation, in one process each, and the captured output is **identical**: `sha256 be5b77338e92fe8a58d580718632e3e42dd17741cf7b703dd98b28095dbb728f` both times, over six messages (`convertFiltersToAST` × 2 shapes, `ValueDataSource` `$` arm × 2, `ValueDataSource` AST arm × 2). That is the claim "the consolidation is not a rewording", measured rather than asserted.

## What is unchanged — live control, same run

- A valid comparand still lowers byte for byte: `toFilterNode([{field:'name',operator:'icontains',value:'acme'}])` is `[["name","icontains","acme"]]`. This control **pre-exists** the change and is green on both sides of it.
- Sibling positive operators (`contains` / `starts_with` / `ends_with` / `not_contains` / `equals`) are untouched with the same comparands — they have no rejection row, and widening by analogy is the published table's call.
- objectui#8557's ARRAY arm still answers **first** for an array on this operator, with its own message about `in` / `not_in` / `between`. The new arm is placed after it on purpose: the reverse order answers a true sentence with the wrong repair.
- A rule carrying **no** comparand still lowers to its 2-tuple `["name","icontains"]`. `isRefusedTextComparand(undefined)` is `true`, so the carve-out is explicit. The view vocabulary has an "absent" the `$` dialect does not, and that shape is already refused downstream with this same code — `parseFilterAST(["name","icontains"])` throws `INVALID_FILTER` / 400 naming `where.name.$icontains`.
- An operator the spec does not know is still passed through verbatim.

## Ablation — three legs, each proven on disk in both directions

Each leg: mutate → prove the mutation reached disk (**exact substring occurrence count both ways, plus the blob hash moved**) → run → restore → prove the restore **by state** (`git hash-object` back to the HEAD blob **and** `git diff HEAD` empty for that path). ⛔ No exit code was read as evidence of restoration. The script carries a `trap ... EXIT INT TERM` with absolute paths, and `git checkout HEAD -- path` (never the bare form, which would restore the mutation from the index).

The subject resolves to **source**, not `dist` — the test imports the module by relative path and vitest aliases the package to `packages/core/src` — so no build staleness applies to these runs.

| leg | mutation | blob | result |
| :-- | :-- | :-- | :-- |
| A | the new arm's operator key made unreachable | `556ac36` → `badaf52` | **8 of 18 failed** |
| B | the arrived-spelling resolution replaced by a hard-coded `$` spelling | `556ac36` → `ffb1094` | **4 of 18 failed** — both the "name the spelling that arrived" pins and both mirror pins |
| C | the absent-comparand carve-out made unreachable | `556ac36` → `0aefaca` | **1 of 18 failed** — the 2-tuple boundary pin |

All three restored to `556ac36bc6f4da1e543b491c800ee3aa84a4fbcb` with an empty `git diff HEAD`.

## Tests and gates

Run from the repo root, the one canonical invocation.

| run | result |
| :-- | :-- |
| `pnpm exec vitest run packages/core/` | **148 files, 3141 tests passed** (exit 0) |
| `pnpm exec vitest run packages/data-objectstack/ packages/plugin-list/src/__tests__/filter-operator-ast-parity.test.ts packages/components/src/hooks/ packages/plugin-view/src/config/` | **67 files, 945 tests passed** — under the shared verify lock, `VERDICT command-exit 0` |
| `pnpm --filter @object-ui/core exec tsc --noEmit` | exit 0 |
| `pnpm --filter @object-ui/core build` | exit 0, dist completeness 202 files |
| `check:new-line-citations` · `check:control-bytes` · `check:self-import` · `check:phantom-deps` · `check:unused-deps` · `check:spec-symbols` · `check:entry-guard` · `check:lint-rule-coverage` · `check:side-effects-array` | all exit 0 |
| `node scripts/check-changeset-presence.mjs` · `node scripts/check-changeset-no-major.mjs` | exit 0 |

Control-byte self-scan over all five changed files, outside the gate: no match (`grep` exit 1).

### Lint narrowing, declared with all three pieces

- **Population, read from eslint's own config** — not guessed: `pnpm --filter @object-ui/core lint` is `eslint .` inside `packages/core`, resolving the repo-root flat config. The diff touches exactly one package, so that package's own run is the complete population for this diff; the other packages' runs belong to CI.
- **File count, read from `--format json`**: **250** files linted, **0 errors**, 542 warnings (all pre-existing `no-explicit-any`, none in the files this PR adds — `text-comparand.ts` reports 0/0).
- **Invariance for untouched files**: type-aware linting is **not enabled** — the root `eslint.config.js` extends `tseslint.configs.recommended` (not the type-checked variant) and its `languageOptions` carries only `ecmaVersion` and `globals`, with no `parserOptions.project` and no `projectService`. No rule reads the type graph, so this diff cannot move the verdict on any file it does not itself contain.

## Acceptance notes

Findings from this run that are **not** repaired here.

- **A THIRD unguarded producer the card does not name** — `convertFilterGroupToAST` / `mapOperator` in `plugin-list`'s `ListView` emits `[field, 'icontains', value]` triples directly, and `toFilterNode` returns an array of AST triples **verbatim** (its rule-object branch never runs), so neither comparand door sees them. The empty half is already shut there by `isFilterValueComplete`, which drops `''` / `null` / `[]`; the **non-string** half is not — that predicate accepts `0`, `false` and `42`. In scope? **No**: a fourth package, a third changeset, and the card's subject is the object-form dialects. Filed separately.
- `packages/core/src/index.ts` re-exports `./utils/filter-converter.js` with `export *`, so every exported name in that file is already on the published entry. Noted because it is what makes FORK B's measurement a one-line check rather than a guess — not a defect.

## 🤖 Provenance

Generated by Claude Code; the session reference for this run is `https://claude.ai/code/session_01UzHd6hDYatoDn17BuwKxnZ`.

---

### DIFF
--- a/.changeset/icontains-comparand-view-rule-9048.md
+++ b/.changeset/icontains-comparand-view-rule-9048.md
@@ -1,0 +1,11 @@
+---
+'@object-ui/core': minor
+---
+
+Refuse an empty or non-string `icontains` comparand on the stored-view rule path, and give all three `@object-ui/core` faces one implementation of that refusal
+
+`@objectstack/spec`'s `FILTER_TEXT_CASES` declares two shapes refused for the case-insensitive contains operator — an empty comparand and a non-string one — each with `code: 'INVALID_FILTER'`. `ValueDataSource` has answered both since objectui#8748 and `convertFiltersToAST` since objectui#9001, but `viewFilterRuleToNode` (reached through the exported `toFilterNode`) lowered both onto the wire: a rule an author saved as `{ field, operator: 'icontains', value: '' }` became `['name','icontains','']`, which the in-memory matchers answer with zero rows while the published table states the wire answer is a predicate that constrains nothing. Same authored filter, two answers, chosen by which data source the view renders against.
+
+The refusal now throws `FilterOperatorError` (`INVALID_FILTER` / 400) from the lowering, naming the field and the operator spelling the view vocabulary actually uses. A valid comparand, the sibling positive operators (`contains` / `starts_with` / `ends_with`), objectui#8557's array-arity refusal and a rule carrying no comparand at all are all unchanged.
+
+The discrimination and the refusal text moved into one internal module that `ValueDataSource`, `convertFiltersToAST` and `viewFilterRuleToNode` all read, so the three faces cannot drift. Both already-shipped messages are byte-identical to what they were. No published export was added.
--- a/packages/core/src/adapters/ValueDataSource.ts
+++ b/packages/core/src/adapters/ValueDataSource.ts
@@ -24,6 +24,11 @@ import {
   RETIRED_FILTER_OPERATORS,
 } from '@objectstack/spec/data';
 import { emulateBatchTransaction } from './batchTransaction.js';
+import {
+  describeComparand,
+  isRefusedTextComparand,
+  textComparandRefusalReason,
+} from '../utils/text-comparand.js';
 
 // ---------------------------------------------------------------------------
 // Configuration
@@ -192,51 +197,24 @@ function refuseArrayComparand(
  * the one outcome worse than the bug.
  */
 /**
- * A comparand as it appears INSIDE a refusal message.
+ * ⚠️ The DISCRIMINATION and the MESSAGE moved to `utils/text-comparand.ts`
+ * (objectui#9048). Nothing about this face's answer changed — the reason string
+ * is the same bytes it has been since objectui#8748, and the pins that drive
+ * both faces and compare their wording verify that without transcribing it.
  *
- * `JSON.stringify` alone is not safe here even though it is what the message
- * wants: it THROWS on a BigInt and on a cyclic object, and this is an
- * exclude-and-log face — a refusal that throws while explaining itself would
- * turn the one path that stays quiet about a bad filter into the one path that
- * takes the caller down. No JSON-sourced filter can carry either shape, so this
- * is about the in-memory callers who hand `find()` a literal.
- *
- * `?? String(target)` keeps `undefined` and a symbol readable — `JSON.stringify`
- * returns `undefined` for both — which is the idiom this file already used.
+ * What stayed here is the ENVELOPE, which is the half that does NOT transfer:
+ * this matcher is deciding about one ROW and HAS a row to exclude, so it
+ * excludes-and-logs (objectui#7349). Its two siblings each seat the same reason
+ * in their own envelope — `filter-converter` throws `FilterOperatorError`
+ * because it is a PRODUCER with no row to exclude.
  */
-function describeComparand(target: unknown): string {
-  try {
-    return JSON.stringify(target) ?? String(target);
-  } catch {
-    return String(target);
-  }
-}
-
 function refuseTextComparand(
   refusals: Set<string>,
   field: string,
   operator: string,
   target: unknown,
 ): false {
-  const declared =
-    `@objectstack/spec's FILTER_TEXT_CASES declares this shape refused `
-    + `(INVALID_FILTER); the declared comparand for '${operator}' is a NON-EMPTY STRING`;
-  if (target === '') {
-    return refuseFilterNode(
-      refusals,
-      `filter comparand for field '${field}' on operator '${operator}' is the EMPTY `
-      + `STRING. Every value contains the empty substring, so evaluating it is a `
-      + `predicate that constrains nothing. ${declared}. Drop the condition instead `
-      + `of sending an empty comparand`,
-    );
-  }
-  return refuseFilterNode(
-    refusals,
-    `filter comparand for field '${field}' on operator '${operator}' is `
-    + `${target === null ? 'null' : typeof target} (${describeComparand(target)}), `
-    + `not a string. Coercing it would answer a query nobody wrote. ${declared}. `
-    + `Write the comparand as a string`,
-  );
+  return refuseFilterNode(refusals, textComparandRefusalReason(field, operator, target));
 }
 
 /**
@@ -513,7 +491,7 @@ function matchesComparisonNode(
     // after a `String()` coercion nobody wrote. Both are refusals in
     // `FILTER_TEXT_CASES`; see {@link refuseTextComparand}.
     case 'icontains':
-      if (typeof target !== 'string' || target === '') {
+      if (isRefusedTextComparand(target)) {
         return refuseTextComparand(refusals, field, String(rawOperator), target);
       }
       return typeof value === 'string' && asciiCaseInsensitiveContains(value, target);
@@ -732,7 +710,7 @@ function matchesDollarOperator(
     // object, so a door on one side only is a result that changes with the
     // SHAPE of the filter rather than with its meaning (objectui#8447).
     case '$icontains':
-      if (typeof target !== 'string' || target === '') {
+      if (isRefusedTextComparand(target)) {
         return refuseTextComparand(refusals, field, operator, target);
       }
       return typeof value === 'string' && asciiCaseInsensitiveContains(value, target);
--- a/packages/core/src/utils/__tests__/filter-view-rule-text-comparand-9048.test.ts
+++ b/packages/core/src/utils/__tests__/filter-view-rule-text-comparand-9048.test.ts
@@ -1,0 +1,424 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `viewFilterRuleToNode` refuses the two `icontains` comparand shapes the
+ * published table declares REFUSED, and all three faces now read ONE
+ * implementation of that refusal (objectui#9048).
+ *
+ * ## The defect
+ *
+ * `@objectstack/spec`'s `FILTER_TEXT_CASES` carries two REJECTION rows for the
+ * case-insensitive contains operator — an empty comparand and a non-string one
+ * — each with `code: 'INVALID_FILTER'`. `ValueDataSource` has answered both
+ * since objectui#8748 and `convertFiltersToAST` since objectui#9001. The STORED
+ * VIEW rule vocabulary answered neither: measured on `origin/main` `6df26f025`
+ * with `@objectstack/spec` 17.4.0,
+ *
+ * ```
+ * toFilterNode([{ field:'name', operator:'icontains', value:'' }]) => [["name","icontains",""]]
+ * toFilterNode([{ field:'name', operator:'icontains', value:42 }]) => [["name","icontains",42]]
+ * convertFiltersToAST({ name: { $icontains: '' } })                => THROWS INVALID_FILTER / 400
+ * ```
+ *
+ * — one contract, refused when the author writes it in the `$` dialect and
+ * lowered onto the wire when the same author SAVES it as a view rule.
+ *
+ * ## What is asserted, and why in this shape
+ *
+ * The weak version of this file would be `expect(() => …).toThrow()`, satisfied
+ * by any throw including a bare `Error` from a converter that had simply
+ * broken. Four stronger claims are made instead, three of them inherited from
+ * `filter-text-comparand-9001.test.ts` because this card is the same contract
+ * one dialect over.
+ *
+ *   1. **The contract's own rows are DRIVEN, not transcribed.** The rejection
+ *      rows are read off the installed `FILTER_TEXT_CASES` at run time; each
+ *      row's comparand is re-seated as a view rule and pushed through
+ *      `toFilterNode`. A transcribed contract drifts from the contract.
+ *   2. **ONE refusal, three envelopes — checked by driving all three.** Triage
+ *      ruled "make the dialects share the one that works", so the matcher's
+ *      refusal text is captured live and BOTH producer faces must contain it
+ *      verbatim. A fourth copy pasted in later cannot satisfy this without
+ *      being byte-identical, and a drift in any one face reddens here.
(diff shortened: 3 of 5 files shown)


os-tesla:
## 派单席评审 —— ⭐ **ACCEPT**，落地

`domain:ui` 派单席 · R16 · 全部在**源头**复核，⛔ 不采信自述报告。评审 head `a5d590ef0`。

### 分诊的裁决被**字面**执行了

分诊要的是「⛔ 不是加第三道守卫，而是让方言共用那一道有效的」。⇒ 判别与拒绝文本进了一个**内部**模块 `packages/core/src/utils/text-comparand.ts`，`ValueDataSource`、`convertFiltersToAST`、`viewFilterRuleToNode` **三张脸都读它**（源头核实：三处 import/引用在位），两份副本删除。

⭐ 而「没有新增已发布导出」这句是**机械可核**的，不是承诺：`packages/core/src/index.ts` **不在本 PR 的变更文件里**（diff --name-only 命中 0）。⇒ 三个新名字全部留在包内。

### ⭐ 消息里那件小事，正是卡点名的第三个问题

`textComparandRefusalReason` 拿的是**实际到达的拼写**，不是替换成某个规范拼写——所以 `$` 臂说 `$icontains`、AST 臂说 `icontains`、视图词汇说它自己的。卡当初指出「逐字照搬会给视图作者开一个它没有的拼写」，这条被正面解掉了，而不是绕过去。

`isRefusedTextComparand(undefined)` 为真，因此那个 `!== undefined` 判断是**承重**的——无比较数的 valueless 形态没有被顺手卷进来，并且有边界钉子。

### FORK A —— 我采纳 A（现在落地），把称重写出来

执行席按要求**重做**了权衡，⛔ 没有继承 objectui#8557 的结论，而重做的结果比卡里预期的更强：修复前**根本不存在一个「今天渲染正确」的视图**。同一条 authored filter 有**两个**答案——经 `@objectstack/formula` 与 `ValueDataSource` 是 9 行里 0 行（视图渲染成**空**），而已发布表陈述的 wire 答案是「a predicate that constrains nothing」（渲染成**超集**）。⇒ 「抛错会破坏一个本来好好的视图」这个前提**不成立**。

⚠️ 剩下的是 **throw 落在哪里**：四个 render-time 的 `toFilterNode` 入口不接（ObjectGrid ×2、RelatedList、LineItemsPanel），那是 objectui#9050，**开着且无人认领**。我仍然选择现在落地，理由三条，每条都是测量：

1. **同一个函数正上方已经这样发货了**：objectui#8557 的数组元数拒绝经由**同样**那四个入口抛，且已经在线上；objectui#9050 自己数出十一个同样可达的既有抛点。⇒ 本 PR **没有**制造这个投递漏洞，它用的是一个已存在、已被单独跟踪的漏洞；压着本 PR 不落地，一个既有抛点也少不了。
2. **暴露面实测为潜伏**：Console 作者不出这个形状（`foldFilterGroupToSpecRules` 丢空值，objectui#4155），全树 `operator: 'icontains'` 的规则只出现在本卡自己的测试/changeset/评论里——控制项起火：全树有 40 个文件提到这个 operator。
3. **被拒的形状是协议自己声明为拒绝的**（`FILTER_TEXT_CASES`，`INVALID_FILTER`）。⇒ 拒绝它是**符合契约**，不是新加限制。

⭐ 并且执行席**没有**抢 objectui#9050 的活：新臂在源码里把测量记下来并指名那张卡是「throw 落在哪里」的主人。这是对的分界。

### FORK B —— 不在本 PR 里，⭐ 而且它的「不在」是测出来的

`@object-ui/core` 的 `package.json` 只声明**一个** `exports` 入口，`src/index.ts` 逐文件点名再导出 ⇒ `data-objectstack` 要够到同一道拒绝，**必须**在已发布入口上添新名字 = **人工门槛**。⇒ 本 PR 写的是 `Part of #9048`（⛔ 不是关单关键词），卡**保持开着**。席位倾向执行席的 A（把那个内部模块整个导出，三个名字，一份实现三个信封），但那是维护者的判断，另开一轮。

⚠️ 执行席给后继留了一个陷阱警告，值得原样保留：vitest 下 `@object-ui/core` 被 alias 到 `packages/core/src`，所以一个 deep-import 内部模块的 data-objectstack 测试**会绿而发布构建会断**——新测试文件里有一条钉子断言那三个名字**今天不在**已发布入口上，并告诉落地者在同一个 commit 里改掉它。

## Contract review

**Reviewed head:** `a5d590ef0` · **Card:** objectui#9048（⚠️ 只交付 `@object-ui/core` 半边，卡不关）

### ① Derived judgments
⛔ **没有任何已发布面移动**：`packages/core/src/index.ts` 未在变更集内（机械核实），三个新名字是包内私有；`@objectstack/spec` 只被**读取**（`FILTER_TEXT_CASES` 的两行是判据来源），未加宽未重述。移动的是**行为**：`viewFilterRuleToNode` 对协议声明为拒绝的两个形状从「下放到 wire」改为「拒绝」。

### ② Semver
`'@object-ui/core': minor`，行为变更写进正文。两张已发货的脸**逐字节不变**（六条驱动消息的 sha256 前后相同，diff 空），⇒ `mustMention` 这条契约义务没有被合并动作碰坏——这正是合并三份实现时最容易碎的地方，而它被单独量过。

### ③ Boundary flags
| 边界 | 读数 |
|---|---|
| 新增已发布导出 | ⛔ 无（index.ts 不在变更集内） |
| 协议 | ⛔ 未加宽；拒绝的是它自己声明为拒绝的形状 |
| 有效比较数的行为 | ⛔ 不变——活控制项：`icontains 'acme'` 前后同为 `[["name","icontains","acme"]]`，兄弟算子与 objectui#8557 的数组臂不动 |
| ⚠️ throw 的投递位置 | 归 objectui#9050（开着、无人认领）；本 PR ⛔ 未预判它，暴露面实测为潜伏 |
| 跨包半边 | ⛔ 不在本 PR：需要新已发布名字 = 人工门槛，卡保持开着 |
| 门槛 / 配额 / 运行时依赖 | 无 |

### Independence pair
```
Implemented-by:  claude/issue-9048-object-form-icontains-comparand   (mode:subagent dev — no session of its own)
```
⚠️ 本行默认档的席内复核，⛔ 不是独立第二会话。

### Verdict
**PASS。** 摘掉两个载体上的 `needs:contract-review`（PR objectui#9152 与卡 objectui#9048），凭本记录与所判 head。flip ready + SQUASH 入队。⚠️ 卡 objectui#9048 **不关闭**，落地后回 `pm:queue`，剩下的半边是 `data-objectstack` + 已发布导出（人工门槛）。

---

---

PR #9154, opened 2026-09-11
fix(app-shell): filter the picklist preview before slicing it (objectui#9074)

### LINKED ISSUE #9074: finding(app-shell): FieldStub slices the picklist preview BEFORE filtering, so a well-formed option behind malformed ones never renders and the +N counter overcounts
Found while implementing objectui#8632 (the field inspector's malformed-option repair). Filed separately: that card is scoped to `ObjectFieldInspector` as the READER, and this is a second surface on the same screen — the field cards of the designer's form canvas.

## The defect

`packages/app-shell/src/views/metadata-admin/previews/FieldStub.tsx`, `PicklistStub()`:

```ts
const opts = options ?? [];
const visible = opts.slice(0, single ? 1 : 3).filter((o) => o.value || o.label);
```

The slice happens BEFORE the filter. So entries the filter drops still consume the display budget, and a well-formed option sitting behind them is never rendered — while the overflow counter, computed as `opts.length - visible.length`, counts them all and reports a number no badge accounts for.

The entries that get dropped are exactly the ones `ObjectFormCanvas` produces for a malformed authored option: its own projection is `value: String(o.value ?? '')` with a non-string `label` mapped to undefined, so a bare string, a `null`, a number or an option with no `value` all arrive as `{ value: '', label: undefined }`.

## Measured

Driven directly against `FieldStub` with the projection `ObjectFormCanvas` hands it for the authored list `['draft', 'open', 'closed', { value: 'real', label: 'Real' }]` — three malformed entries followed by one well-formed one:

```
multiselect card ->  "Pick one or more…+4"
select card      ->  "Select…"
```

The multiselect card shows the empty-state placeholder AND a `+4` counter, with zero badges. The select card reads as a field with no options at all. In both, the one option that IS well formed — `Real` — is not on screen. Swapping the order (well-formed entry first) renders it, which is what identifies the slice/filter ordering as the cause rather than the filter itself.

## Why this is filed rather than folded into objectui#8632

That card's ruling is scoped to the reader that WRITES — its severity is the silent deletion on commit, and its repair is a strict `readOptions` plus an inline report on the row. This canvas is a read-only preview stub: it destroys nothing and it now sits next to an inspector that reports the same malformed option loudly, so the severity here is low. But it is a reproducible defect on its own terms — a well-formed option that the author authored, hidden from the surface that previews their form, with a counter that says something is there — and the ordering bug is mechanical and one line.

## Scope note

Two candidate repairs, not chosen here: filter then slice (renders `Real`, and the counter then means what it says), or keep the ordering and count only what the filter admits. Which one is right depends on whether the canvas should show a malformed option AT ALL now that the inspector reports it — a product question this card does not answer.

Not to be confused with objectui#8632, which is the inspector, and with objectui#8488, which is a stored value matching no option in `InspectorSelectField`.

Filed by the objectui#8632 execution seat, session `session_01MPaVWWMuWeT5LgB1qoXjVB`. No assignee and no labels on purpose — grading and routing belong to triage.

---

os-tesla:
Fixes #9074

## What changed

`PicklistStub` inside `FieldStub` — the field-card preview on the form-designer canvas — computed its visible options as a slice followed by a filter:

```ts
const visible = opts.slice(0, single ? 1 : 3).filter((o) => o.value || o.label);
```

An entry with neither a value nor a label renders as nothing, but the slice had already spent the display budget on it. One cause, two symptoms:

1. **The missing option.** A well-formed option sitting BEHIND malformed ones never reached the card.
2. **The counter.** The overflow counter, `opts.length - visible.length`, counted those same entries, so `+N` promised badges no wider card could ever show.

The repair filters first and slices second, and counts `+N` against the filtered population:

```ts
const renderable = opts.filter((o) => o.value || o.label);
const visible = renderable.slice(0, single ? 1 : 3);
```

`ObjectFormCanvas` is the only consumer of `FieldStub`, and it is also what produces the shape the filter drops: it projects every authored option through `String(o.value ?? '')` with a non-string label mapped to `undefined`, so a bare string, a `null`, a number, or an option with no `value` all arrive as `{ value: '', label: undefined }`.

## Mechanism, measured on this tree rather than inherited

Triage's causal chain was confirmed by content before anything was edited, and its "two lines" estimate held:

- The slice/filter composition and the counter expression sit in **one module-local function**, `PicklistStub`, in one file. Not split across functions.
- **No second consumer.** `PicklistStub` is not exported; `FieldStub` is exported from its module but is absent from `previews/index.ts` and from the package's public surface, and `git grep FieldStub` over `packages`, `apps` and `examples` returns its own definition plus exactly one call site, in `ObjectFormCanvas`. Nothing depends on the pre-filter slice.
- The counter and the render read the **same** population today (`opts`) and the same one after (`renderable`), so fixing one without the other was never structurally forced — which is precisely why the acceptance pins them separately.

Then measured, by checking out the unrepaired file from this branch's base commit verbatim (blob `94d9c3175`, byte-identical to `HEAD:…/FieldStub.tsx` at the branch point) and running the new pins against it. The card's own fixture — authored `['draft', 'open', 'closed', { value: 'real', label: 'Real' }]` — reproduced exactly what the card reported: the multi-value card rendered the `Pick one or more…` placeholder and a counter span reading `+4`, the single-value card rendered `Select…`, and `Real` was on neither.

## No pre-existing preview renders differently

When every authored entry survives the filter, `filter`-then-`slice` and `slice`-then-`filter` compute the same `visible` and the same `+N` — the filter is a no-op on that population. A picklist shorter than the budget shows no `+N` on either side, for the same reason. The only previews that move are the malformed-options case this card is about.

That is asserted, not argued: the three control cases are in the same file as the subject, and they **passed against the unrepaired file** in the run above (4 passed there, and those four are the three controls plus the malformed-LAST positional contrast).

## Verification

Runs were serialised through the shared verify lock; each verdict below is the line the run printed for itself.

| run | command | result |
| --- | --- | --- |
| pins, repaired tree | `pnpm exec vitest run …/FieldStub.picklistSliceOrder-9074.test.tsx` | exit 0 — `Test Files 1 passed (1)`, `Tests 9 passed (9)` |
| pins, unrepaired file from the base commit | same, with the base blob checked out | exit 1 — `Tests 5 failed | 4 passed (9)`; the 5 are every subject pin, the 4 are the three controls and the positional contrast |
| previews regression | `pnpm exec vitest run packages/app-shell/src/views/metadata-admin/previews/` | exit 0 — `Test Files 57 passed (57)`, `Tests 655 passed (655)`, 91.56s |
| typecheck | `pnpm --filter @object-ui/app-shell run type-check` | exit 0, after building the dependency closure |
| lint | `pnpm --filter @object-ui/app-shell run lint` | exit 0 — `2966 problems (0 errors, 2966 warnings)`, all pre-existing; neither changed file is named |
| changeset presence | `node scripts/check-changeset-presence.mjs` | exit 0 — 1 published source file changed, 1 changeset declared |
| no-major / claims / citations / control-bytes | `check-changeset-no-major`, `check:changeset-claims`, `check:new-line-citations`, `check:control-bytes` | all exit 0 |

Two notes on what those numbers do and do not say:

- The first `type-check` attempt exited **2** with `TS2307: Cannot find module '@object-ui/fields'` and four siblings. That is the prerequisite-not-met shape — unbuilt workspace dependencies, not a red. It was re-run after `pnpm --filter '@object-ui/app-shell^...' run build` and exits 0. The typecheck does reach the new pin file: `tsc -p tsconfig.test.json --listFiles` emits 4498 files and both changed files are among them.
- `pnpm lint` at the repo root is `turbo run lint` across every package. It is CI's run, not this branch's: the diff touches one package, so the package's own `eslint .` is the scope, and it is green. (An exploratory run with `--no-inline-config` added reports 13 files with errors; that flag is not in this repo's lint command, the 13 are all pre-existing files carrying inline disables, and neither changed file is among them.)

## Ablation — two legs, one per half

The acceptance asks that each half be able to fail on its own, so each half was mutated on its own. Every mutation was proven on disk before its run (occurrence counts in both directions plus a moved `git hash-object` blob), and every restore was proven **by state** — `git diff HEAD` empty AND the blob back to the HEAD blob `c66683ab8` — never by an exit code. Both legs ran under a `trap … EXIT INT TERM`.

**Leg 1 — revert only the counter's population** (`renderable.length` back to `opts.length` in both the guard and the number; `renderable.length > visible.length` occurrences 1 → 0, `opts.length > visible.length` 0 → 1, blob `c66683ab8` → `e480417f6`):

```
Tests  2 failed | 7 passed (9)
  × counts the options that survived filtering and were not rendered
  × drops the counter entirely when every surviving option is on screen
```

Exactly the counter family. The rendering family and all three controls stayed green — the counter assertions are load-bearing in their own right, not a side effect of the reorder.

**Leg 2 — the caricature triage warned about**: keep the slice-first order, but compute an honest-looking `+N` from the filtered population (`renderable.slice(` 1 → 0, `opts.slice(0, single ? 1 : 3).filter` 0 → 1, `hiddenCount` 0 → 3, blob → `d72ef8cdc`):

```
Tests  3 failed | 6 passed (9)
  × renders the well-formed option in the multi-value card
  × renders the well-formed option in the single-value card
  × spends the whole budget on surviving options, not on malformed ones
```

Exactly the rendering family, with both counter assertions green. This is the shape the ruling exists to refuse — a `+N` that reads correct while `Real` is still off the screen — and this branch fails on it.

No build step is involved in either leg: the pin imports `./FieldStub` by relative path, so it resolves to the source file the mutation edits, not to any `dist`.

## Clause-②: `no` — and it held

The claim declared `Clause-②: no`, with a hard stop attached: any new exported symbol, any schema narrowing or widening, any published-payload change would mean the `no` was judged wrong. None was needed. The repair is two statements and one expression inside a function local to its module. The changed-file list, which a `no` owes in place of contract review:

| file | change |
| --- | --- |
| `packages/app-shell/src/views/metadata-admin/previews/FieldStub.tsx` | `PicklistStub` filters before slicing; `+N` counts the filtered population |
| `packages/app-shell/src/views/metadata-admin/previews/FieldStub.picklistSliceOrder-9074.test.tsx` | new — pins both halves independently, plus the unchanged-control population |
| `.changeset/9074-fieldstub-picklist-filter-before-slice.md` | `@object-ui/app-shell` patch |

## Acceptance notes

- **A picklist whose options are ALL malformed now previews silent.** Before this change it showed the empty-state placeholder plus a `+N` equal to the authored count; now it shows the placeholder and no counter. That is the chosen repair's own consequence and it is deliberate: the old number pointed at nothing, and the scope note on the card leaves "should the canvas show a malformed option at all" to the inspector, which reports the same malformed option loudly on the same screen. Worth a maintainer's eye if the preferred signal is the opposite one.
- Observed and deliberately **not** changed: `LookupStub`, in the same file, computes `const meta = referenceTo ? FIELD_TYPE_META.lookup : null;` and then renders `{meta && null}` under a comment reading "swallow unused import warning" — a value computed and discarded to keep an import alive. Cosmetic, no defect, out of this card's scope.

Draft on purpose: the dispatching PM reviews at source and lands it. Not flipped ready, no auto-merge, not enqueued.

Generated by Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`.

### DIFF
--- a/.changeset/9074-fieldstub-picklist-filter-before-slice.md
+++ b/.changeset/9074-fieldstub-picklist-filter-before-slice.md
@@ -1,0 +1,24 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Fix the form-designer canvas picklist preview: `FieldStub` now filters the
+options before slicing them, so a well-formed option is no longer hidden by
+malformed ones in front of it and the `+N` overflow counter stops overcounting
+(objectui#9074).
+
+The preview took the first one (single-value) or three (multi-value) authored
+options and *then* dropped the ones with neither a value nor a label. Entries
+that render as nothing therefore consumed the display budget, and the counter —
+computed against the whole authored list — reported a number no badge on the
+card accounted for. With the authored list `['draft', 'open', 'closed',
+{ value: 'real', label: 'Real' }]`, which `ObjectFormCanvas` projects to three
+`{ value: '', label: undefined }` entries followed by one good one, the
+multi-value card showed its empty-state placeholder plus `+4` and zero badges,
+and the single-value card read as a field with no options at all — with `Real`
+on neither.
+
+Both now render `Real`, and `+N` counts only options that survived filtering and
+did not fit. A picklist with no malformed options previews and counts exactly as
+it did before: when every entry survives the filter, filtering first and slicing
+first compute the same result.
--- a/packages/app-shell/src/views/metadata-admin/previews/FieldStub.picklistSliceOrder-9074.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/previews/FieldStub.picklistSliceOrder-9074.test.tsx
@@ -13,9 +13,12 @@
  *   2. the overflow counter, `opts.length - visible.length`, counts the
  *      malformed entries too, so `+N` promises badges no surface can show.
  *
- * Both are pinned here, each by an assertion that can fail on its own: a
- * repair that fixes only the counter (the louder of the two) still leaves
- * `Real` off the screen, and this file says so.
+ * Both are pinned here, in two families that fail INDEPENDENTLY — no assertion
+ * does double duty, because the plausible wrong repair is the one that gets the
+ * counter right and leaves the option missing. Measured, by mutating each half
+ * of the repair on its own: reverting only the counter's population reds the
+ * counter family alone, and the caricature that keeps the slice-first order
+ * while computing an honest-looking `+N` reds the rendering family alone.
  *
  * ## The fixtures are the projection, not the authored list
  *
@@ -130,6 +133,16 @@ describe('FieldStub picklist preview — a well-formed option behind malformed o
     expect(screen.queryByText('Select…')).toBeNull();
   });
 
+  it('spends the whole budget on surviving options, not on malformed ones', () => {
+    render(<FieldStub type="multiselect" options={TWO_MALFORMED_THEN_FIVE} />);
+    // All three multi-value slots go to options that can actually render.
+    expect(screen.getByText('Willow')).toBeTruthy();
+    expect(screen.getByText('Walnut')).toBeTruthy();
+    expect(screen.getByText('Wisteria')).toBeTruthy();
+    // The budget is three, so the fourth survivor stays off the card.
+    expect(screen.queryByText('Wattle')).toBeNull();
+  });
+
   it('positional contrast: the same option with the malformed entries LAST always rendered', () => {
     render(<FieldStub type="multiselect" options={MALFORMED_LAST} />);
     expect(screen.getByText('Real')).toBeTruthy();
@@ -139,12 +152,7 @@ describe('FieldStub picklist preview — a well-formed option behind malformed o
 describe('FieldStub picklist preview — the +N overflow counter', () => {
   it('counts the options that survived filtering and were not rendered', () => {
     render(<FieldStub type="multiselect" options={TWO_MALFORMED_THEN_FIVE} />);
-
-    // Three renderable options fit the budget.
-    expect(screen.getByText('Willow')).toBeTruthy();
-    expect(screen.getByText('Walnut')).toBeTruthy();
-    expect(screen.getByText('Wisteria')).toBeTruthy();
-    // Two renderable options are hidden, so the counter says two.
+    // Five options survive the filter and three fit, so the counter says two.
     expect(screen.getByText('+2')).toBeTruthy();
     // Not `+6`: that is what the slice-first order claimed, counting the two
     // malformed entries and the four options it never reached.
@@ -153,7 +161,9 @@ describe('FieldStub picklist preview — the +N overflow counter', () => {
 
   it('drops the counter entirely when every surviving option is on screen', () => {
     render(<FieldStub type="multiselect" options={MALFORMED_FIRST} />);
-    expect(screen.getByText('Real')).toBeTruthy();
+    // Reported as `+4` before the repair: four authored entries, nothing on
+    // screen. Deliberately asserts nothing about `Real` — that belongs to the
+    // other half, and mixing them would stop either from failing on its own.
     expect(screen.queryByText(ANY_OVERFLOW)).toBeNull();
   });
 });
--- a/packages/app-shell/src/views/metadata-admin/previews/FieldStub.picklistSliceOrder-9074.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/previews/FieldStub.picklistSliceOrder-9074.test.tsx
@@ -1,0 +1,183 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * `PicklistStub` inside `FieldStub` used to compute its preview as
+ *
+ *     const visible = opts.slice(0, single ? 1 : 3).filter((o) => o.value || o.label);
+ *
+ * — the slice runs BEFORE the filter, so entries the filter drops still
+ * consume the display budget (objectui#9074). One cause, two symptoms:
+ *
+ *   1. a well-formed option sitting BEHIND malformed ones never renders,
+ *      because the malformed ones ate the whole budget; and
+ *   2. the overflow counter, `opts.length - visible.length`, counts the
+ *      malformed entries too, so `+N` promises badges no surface can show.
+ *
+ * Both are pinned here, each by an assertion that can fail on its own: a
+ * repair that fixes only the counter (the louder of the two) still leaves
+ * `Real` off the screen, and this file says so.
+ *
+ * ## The fixtures are the projection, not the authored list
+ *
+ * `ObjectFormCanvas` is the only consumer of `FieldStub`, and it projects the
+ * authored `options` through
+ *
+ *     value: String(o.value ?? ''), label: typeof o.label === 'string' ? o.label : undefined
+ *
+ * so a bare string, a `null`, a number, or an option with no `value` all
+ * arrive here as `{ value: '', label: undefined }`. That shape — not the
+ * author's original entry — is what the stub filters on, so it is what these
+ * fixtures carry.
+ *
+ * ## Order is load-bearing in the fixtures
+ *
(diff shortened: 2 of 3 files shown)


os-tesla:
## 派单席评审 —— ⭐ **ACCEPT**，落地

### 两个症状一起修好了，而且 diff 自己就能读出来

```diff
-  const visible = opts.slice(0, single ? 1 : 3).filter((o) => o.value || o.label);
+  const renderable = opts.filter((o) => o.value || o.label);
+  const visible = renderable.slice(0, single ? 1 : 3);
…
-      {opts.length > visible.length && (
-        <span …>+{opts.length - visible.length}</span>
+      {renderable.length > visible.length && (
+        <span …>+{renderable.length - visible.length}</span>
```

⇒ 先过滤后切片（那个渲染不出来的选项回来了），**并且**计数器的总体从「作者写了多少」换成「过滤后还剩多少」。⛔ 没有只修更显眼的那一半。

### ⭐⭐ 最值钱的是 ablation 的第二条腿：它把分诊点名的陷阱**演出来并拒绝掉**

分诊警告过：「counter 是两个症状里更显眼的那个，⇒ 容易只修 counter、把那个渲染不出来的选项留在原地」。

⇒ 执行席没有只写一条「两个都对」的用例了事，它**造了那个漫画式的坏修法**——保留切片在前的顺序，但让 `+N` 看起来诚实——然后跑：**3 条渲染家族的用例红，两条计数器断言全绿**。

⭐ 那正是分诊描述的那个「修一半」的产物，而这个分支的钉子**认得出它**。一条能把「看起来修好了」和「真的修好了」分开的 ablation，比十条绿用例有价值。

第一条腿同样干净：只把计数器的总体退回去，**恰好 2 条计数器用例红**，渲染家族与三个控制项全绿。⇒ 两个症状各有各会红的钉子，⛔ 不是一条混合断言。

### 前提与围栏，逐条核过

- **修前复现**：把分支点的原文件逐字检出（blob `94d9c3175`，并在跑之前验过与 BASE 树条目相等），卡自己的 fixture 渲染出占位符加一个读作 `+4` 的计数器，而 `Real` **两边都不在**。⇒ 缺陷是先复现再修的。
- **单一消费者**：`PicklistStub` **未导出**（`export .*PicklistStub` 命中 0），`FieldStub` 不在 previews 的 barrel 也不在包的公开面上，全仓只有 `ObjectFormCanvas` 一个调用点；整棵 previews 树 57 文件 / 655 用例全绿，**包含那个唯一消费者自己的测试**。
- **Clause-② `no` 站得住**：本 PR **零个 `package.json` 变更**、无新导出符号、无 schema 移动、无已发布载荷。⇒ 那条硬停止条件没有触发。
- **门**：派单席自己复跑 `check-clause2-carriers --pair 9154` ⇒ **exit 0**，且这次输出多了 `no` 一侧的那句：「its diff carries no widening tell. ⚠️ A tell is not a proof and its absence is not one either.」⇒ 我照它的措辞记：**这是一条通过，不是一份证明**；`no` 的依据仍然是上面那四项实测，不是这条 tell 扫描。
- **NOT MEASURED 被分清**：一次 `exit 99` 队列超时、一次 type-check `exit 2`（`TS2307`，依赖闭包未构建）都被标成前置未满足而非红门，构建后重跑绿。⛔ 没有一处被当成绿。

### ⚠️ 那个 open question：采纳 A（已发货的那个），⛔ 不另立卡

全部选项都畸形的 picklist 现在预览是**静默**的（占位符、无计数器），而此前会显示一个等于作者书写条数的 `+N`。

⇒ 判 **A**：那个旧数字**指向的是不存在的东西**——它正是本卡这个缺陷的一部分，不是一个被本次修复拿掉的功能。而 objectui#8632 的检视器在**同一块屏幕上**对同一个畸形选项有明确报告，⇒ 作者不会失去信号。

⛔ 不另立卡：B（给画布自己的畸形选项提示）要决定「画布到底该不该显示一个畸形选项」，那是本卡明确声明不回答的产品问题，且要新 i18n 串和设计判断——不该由一个 p3 的顺序修复顺手带走。执行席把它标成「不阻塞、交给维护者的眼睛」而不是自行决定，这个分寸是对的。

⇒ flip ready + SQUASH 入队。

---

---

PR #9157, opened 2026-09-11
fix(app-shell): resolve drill filter placeholders before the escape hatch builds a URL

### LINKED ISSUE #9022: finding(app-shell,react): the drill "Open in list" / `target: 'navigate'` path sends filter placeholders UNRESOLVED, so one drill has two scopes depending on its target
Found while fixing objectui#8944 (PR #9016). Filed rather than folded in: that card is about how the two filter SOURCES compose, this is about what the composed value still carries. Orthogonal, and this one is not chart-specific.

## The defect

Every widget that drills through `DrillNavigationContext.openRecordList` hands it the widget's own filter composed with the click context — and that value is the **raw authored filter**, placeholders included. Both in-repo consumers of the handler (`useOpenRecordList` in app-shell, and `OpenInListButton` in plugin-dashboard) pass it straight to `serializeDrillFilterParams`, which writes it into `filter[...]` URL params verbatim. Nothing resolves `{current_quarter_start}` or `{current_user_id}` on the way.

## Why it survived: the drawer path does NOT have the bug

The in-place drawer renders `object-data-table`, whose own fetch calls `resolveFilterPlaceholders(schema.filter, filterScope)`. So a macro authored on the widget **is** resolved on the drawer path and is **not** on the navigate path.

⇒ One drill, two different scopes, decided only by `drillDown.target`. And the same widget's own chart bars are scoped by the resolved value — `ObjectChart` resolves its query legs through `resolveFilterPlaceholders(schema.filter, filterScope)` while composing the drill from `schema.filter` raw — so the drilled list disagrees with the chart the user clicked.

## The failure is silent and lands on a literal

A widget filter of `{ close_date: { $gte: '{current_quarter_start}' } }` serializes to `filter[close_date][gte]={current_quarter_start}`. The bare data surface parses that back as an ordinary string comparand, so the list matches nothing (or the driver refuses the comparand). Nothing errors at the seam that produced it.

Note this direction is NARROW, unlike objectui#8944's widening — but it is the same class: the widget's declared scope silently not being the scope the drilled list runs under.

## Not chart-specific

`openRecordList` is the shared drill escape hatch: the table / pivot / metric widgets reach it through `DrillDownDrawer` and `OpenInListButton` too. Any of them carrying a placeholder in `filter` has the same split.

## The decision this needs

Where resolution belongs is the actual question, and it should be answered once rather than per widget:

- **A — at each widget's drill seam**: compose from the already-resolved filter the widget queries with. Local, but it is N seams and the next hand-rolled drill panel will forget it again.
- **B — inside the `openRecordList` contract**: resolve once where the value crosses into the host, covering every caller including future ones. Needs the filter-token scope to be reachable there.

B looks right, but the scope plumbing is a real design question, so this is not a mechanical fix.

## Acceptance

- A widget whose `filter` carries `{current_user_id}` or a relative-date macro, drilled with `target: 'navigate'`, produces a URL carrying the RESOLVED value.
- The drawer path keeps resolving exactly once — no double resolution.
- Pinned for at least one non-chart caller as well, since the contract is shared and a chart-only pin would let the next widget regress silently.

os-tesla:
Fixes #9022

## What was wrong

Every widget composes its drill filter from the **raw authored** `schema.filter` — `ObjectChart`'s `drillFilter` memo and `DrillDownDrawer`'s `filter` prop both do — so `{current_user_id}`, `{current_org_id}` and the relative-date macros were still literals by the time the value reached the host's `openRecordList`.

The in-place drawer arm never had the defect: the `object-data-table` it renders resolves `schema.filter` in its own fetch. So only the **navigate** arm — and the "Open in list" button in the drawer's own header — wrote the placeholder into the URL, where the read side parses it back as an ordinary string comparand and the destination list matches nothing. One drill, two scopes, decided only by `drillDown.target`, and on the navigate arm the destination disagreed with the chart bar the user had just clicked (the chart scopes its own bars by the resolved value).

## Which call site owns the scope

**`useOpenRecordList` in `@object-ui/app-shell`.** Measured on this tree, not inherited:

- It is the **only** in-repo implementation of `DrillNavigationContext.openRecordList`. Both mounts (`DashboardView`, `ReportView`) call this one hook; no other file constructs a `DrillNavigationProvider` value outside tests.
- The scope is in hand there: `AppContent` mounts `FilterScopeProvider` around `ConsoleLayout`, and the `dashboard/:dashboardName` and `report/:reportName` routes are inside it, so `useFilterScope()` reads real session values at that hook.
- Every escape hatch funnels through it: `ObjectChart`'s navigate effect and its own header button, `DrillDownDrawer`'s navigate effect, and `OpenInListButton`. One resolution therefore covers **both** widget families, instead of the N widget seams the next hand-rolled drill panel would forget.

Deliberately **not** inside `serializeDrillFilterParams`: that function is a pure URL encoder which writes faithfully what it is handed, and the scope is a React context read — putting it there would make a pure function scope-dependent while leaving the callers holding the scope anyway.

`DrillNavigationValue.openRecordList` now states in its own contract that resolving is the host implementation's job, because an out-of-repo host wiring the same context would otherwise reproduce the split.

## Mechanism assumptions, re-measured — one is FALSIFIED

1. ⚠️ **FALSIFIED as stated.** The dispatch said the two consumers "pass it straight to `serializeDrillFilterParams`". Measured: `OpenInListButton` never touches the serializer — it calls the host handler, and `useOpenRecordList` is the sole implementation of that handler. So the two named names are not two peers; one is the handler, the other a caller of it. **This makes the fence stronger, not weaker**: "both consumers, or neither" is satisfied *by construction* at the single handler, and the falsification is why one edit covers both widget families.
2. ✅ **Confirmed.** The drawer path does not have the defect: `ObjectDataTable`'s fetch builds its params as a `$filter` of `resolveFilterPlaceholders(schema.filter, filterScope)`, and `DrillDownDrawer` hands its `filter` prop into exactly that schema.
3. ✅ **Confirmed.** The scope is available at the chosen site (see above).
4. ✅ **Confirmed.** `ObjectChart` composes `drillFilter` from `schema.filter` **raw** via `composeDrillFilter`, while its query legs go through `resolveFilterPlaceholders`. Same reading the objectui#9024 seat reported from the pivot path. Refs: objectui#9024, objectui#9135.

None of the STOP conditions tripped. There is no third consumer that holds no scope; the fix needed no change to `serializeDrillFilterParams`; and nothing lands on different rows except the placeholder case — both vocabularies substitute **whole-string** `{token}` values only and pass everything else through untouched. On the leak question: `{current_user_id}` and `{current_org_id}` resolve to values the viewer's **own** session already determines, so nothing enters the URL that the session did not already decide. The visible consequence is that a *shared* drilled URL now names a concrete id instead of a literal that matched nothing — which is the behaviour the card asks for, and what the drawer arm has always done.

## Pins

`packages/app-shell/src/views/useOpenRecordList.placeholderScope-9022.test.tsx` — the destination scope carried by the URL, read back with the repo's own `parseUrlFilterTriples`, for a session token, a date macro (pinned to a literal date under a frozen clock), and both vocabularies at once. The composed shape is built with the real `composeDrillFilter`, i.e. what `ObjectChart` actually hands over.

`packages/app-shell/src/views/drillPlaceholderParity-9022.test.tsx` — the parity the card measured, pinned: the **real** `DrillDownDrawer` over the **real** `useOpenRecordList`, comparing the drawer arm's `$filter` (captured off `dataSource.find`) against the navigate arm's URL and against the header button's URL. Each case asserts parity **and** that the resolved value is present — equality alone would also hold in the pre-fix world, where both arms carried the literal.

Controls, live in the same run: a placeholder-free drill is pinned as the **whole** URL string on the hook, and as the identical triple list on all three arms; a drill with no filter at all still navigates bare; and an unresolvable token stays a literal rather than being dropped, so the filter can never widen. The same-kind pre-existing control is `drillUrlFilters.test.ts`, which pins the encoder this hook delegates to and is untouched by this change.

The resolution line was replaced on disk with a pass-through, at the branch head `7ac092df8`. No rebuild is involved: the root vitest config redirects every package specifier to its `src`, so the run reads the mutated file directly.

```
HEAD blob for packages/app-shell/src/views/useOpenRecordList.ts = 20a7fa10e579cb23605a8053a7f5a688eba051be
--- BEFORE mutation (on-disk occurrence counts) ---
fixed-line count   : 1
ablated-line count : 0
on-disk hash       : 20a7fa10e579cb23605a8053a7f5a688eba051be
--- AFTER mutation (on-disk occurrence counts) ---
fixed-line count   : 0
ablated-line count : 1
on-disk hash       : b80f993d6e7d39c1a65a997ac67b68d97a255fe6
VERDICT ablation-run exit = 1
 Test Files  2 failed (2)
      Tests  6 failed | 4 passed (10)
--- RESTORE ---
on-disk hash after restore : 20a7fa10e579cb23605a8053a7f5a688eba051be
fixed-line count           : 1
ablated-line count         : 0
git diff HEAD --name-only : []
RESTORE PROVEN BY STATE: hash back to HEAD blob and git diff HEAD empty
```

The six that reddened are exactly the six placeholder/parity cases. The four that stayed green are the three controls plus the non-widening guard — none of them can move under an ablation of this arm, which is what makes them controls rather than more of the subject. The restore was proven by state (blob hash back to the HEAD blob, `git diff HEAD` empty), never by an exit code, and the script carried a `trap` on EXIT/INT/TERM pointing at absolute paths with `git checkout HEAD --`, so a mid-mutation kill could not leave the tree mutated.

## Verification run

Measured at `7ac092df8`, on a shared box (other agents' unlocked work runs on the same cores), with every exit code captured by redirect before any pipe.

| what | command | verdict |
| --- | --- | --- |
| dependency closure | `turbo run build --filter="@object-ui/app-shell^..." --concurrency=2` | `28 successful, 28 total`, wrapper exit 0 |
| type-check | `pnpm --filter @object-ui/app-shell --filter @object-ui/react run type-check` | both `Done`, exit 0 |
| the new pins | `pnpm exec vitest run` on the two files | `Test Files 2 passed (2)`, `Tests 10 passed (10)`, exit 0 |
| affected packages | `pnpm exec vitest run packages/app-shell/ packages/react/` | `761` files, `7544 passed, 1 failed`, exit 1 — see below |
| lint | `pnpm --filter @object-ui/app-shell --filter @object-ui/react run lint` | exit 0 (2966 pre-existing warnings, 0 errors) |
| changeset presence | `node scripts/check-changeset-presence.mjs` | exit 0, declaration found |
| citation gate | `node scripts/check-new-cross-file-line-citations.mjs` | `0 new citation(s)`, exit 0 |
| control bytes | `grep -naP` over the five touched files | no match |

The type-check reading is a real one and not a silent skip: `tsc -p tsconfig.test.json --listFiles` lists both new test files in the program.

⚠️ **The one failure is not this change.** `RecordApprovalsPanel.viaOverrideMarker.test.tsx` could not find its `via-override-chip` under full-parallel load; re-run alone it is `6 passed`, exit 0. It shares no file, no import and no surface with this diff — it is the load-dependent class AGENTS.md already describes, where an unbounded module load lands inside a bounded assertion window. Recorded rather than swept: nothing here should be read as having made it green.

Heavy runs went through `scripts/pm/os-verify-lock.sh` in the sibling checkout, as this repo's AGENTS.md requires.

## Acceptance notes — found, and deliberately left alone

- `ObjectChart` hand-rolls its own "Open in list" button (same `data-testid`, same label) instead of rendering the shared `OpenInListButton`. Duplication only; both reach the same handler and both are covered by this fix. Noted, not filed — no card and no PR is heading for that file.
- `serializeDrillFilterParams` skips array comparands (an `$in`-style value), which drops that condition from the URL and degrades the drill to a superset. Its own docblock declares that posture deliberate and an existing test pins it, and the composition half is already carded. Noted, not filed. Refs: objectui#9024.

---

### DIFF
--- a/.changeset/9022-drill-navigate-placeholder-scope.md
+++ b/.changeset/9022-drill-navigate-placeholder-scope.md
@@ -1,0 +1,29 @@
+---
+'@object-ui/app-shell': patch
+'@object-ui/react': patch
+---
+
+Resolve filter placeholders on the drill "escape hatch", so one drill has one scope
+whatever `drillDown.target` says (objectui#9022).
+
+Every widget composes its drill filter from the RAW authored filter, so
+`{current_user_id}`, `{current_org_id}` and the relative-date macros were still
+literals when they reached the host's `openRecordList`. The in-place drawer arm never
+had the defect — the `object-data-table` it renders resolves the filter in its own
+fetch — so only the navigate arm, and the "Open in list →" button in the drawer header,
+wrote the placeholder into the URL. `filter[close_date][gte]={current_quarter_start}`
+parses back on the read side as an ordinary string comparand, so the destination list
+matched nothing, silently, and in disagreement with the chart bar the user had just
+clicked (that chart scopes its own bars by the RESOLVED value).
+
+`useOpenRecordList` now expands both placeholder vocabularies against the session scope
+it already sits inside, immediately before serialization. It is the single host handler
+every escape hatch funnels through — the chart's navigate arm and its own header
+button, `DrillDownDrawer`'s navigate arm, and `OpenInListButton` — so both widget
+families are covered by one resolution rather than N widget seams. A drill carrying no
+placeholder produces a byte-identical URL: both vocabularies substitute whole-string
+`{token}` values only.
+
+The `DrillNavigationValue.openRecordList` contract now states that resolving is the
+host implementation's job, since an out-of-repo host wiring the same context would
+otherwise reproduce the split.
--- a/packages/app-shell/src/views/drillPlaceholderParity-9022.test.tsx
+++ b/packages/app-shell/src/views/drillPlaceholderParity-9022.test.tsx
@@ -65,12 +65,20 @@ const LITERAL_FILTER = {
   close_date: { $gte: '2026-04-01' },
 };
 
-let location = { pathname: '', search: '' };
+/**
+ * Where the router currently is, rendered rather than captured into a module
+ * variable: a render-phase write to an outer binding is a lint error here, and
+ * an effect-based capture would need its own settling window on top of the one
+ * `waitFor` already gives us.
+ */
+function LocationProbe() {
+  const { pathname, search } = useLocation();
+  return <span data-testid="router-location">{`${pathname}${search}`}</span>;
+}
 
-function LocationSpy() {
-  const l = useLocation();
-  location = { pathname: l.pathname, search: l.search };
-  return null;
+/** The probe's current reading, or the empty string before anything mounted. */
+function currentLocation(): string {
+  return screen.queryByTestId('router-location')?.textContent ?? '';
 }
 
 /** The host wiring `DashboardView` and `ReportView` both use, verbatim. */
@@ -95,7 +103,6 @@ function mount(
   ds: ReturnType<typeof makeDataSource>,
 ) {
   cleanup();
-  location = { pathname: '', search: '' };
   return render(
     <MemoryRouter initialEntries={['/apps/crm/dashboard/pipeline']}>
       <FilterScopeProvider currentUserId={SCOPE.currentUserId} currentOrgId={SCOPE.currentOrgId}>
@@ -104,7 +111,7 @@ function mount(
             path="/apps/:appName/*"
             element={
               <Host>
-                <LocationSpy />
+                <LocationProbe />
                 <DrillDownDrawer
                   open
                   onClose={vi.fn()}
@@ -138,8 +145,8 @@ async function drawerScope(filter: Record<string, unknown>): Promise<FilterTripl
 async function navigateScope(filter: Record<string, unknown>): Promise<FilterTriple[]> {
   const ds = makeDataSource();
   mount('navigate', filter, ds);
-  await waitFor(() => expect(location.pathname).toBe('/apps/crm/opportunity/data'));
-  return parseUrlFilterTriples(new URLSearchParams(location.search));
+  await waitFor(() => expect(currentLocation()).toMatch(/^\/apps\/crm\/opportunity\/data/));
+  return parseUrlFilterTriples(new URLSearchParams(currentLocation().split('?')[1] ?? ''));
 }
 
 /** The scope the drawer's own header button lands on (`OpenInListButton`). */
@@ -147,8 +154,8 @@ async function openInListScope(filter: Record<string, unknown>): Promise<FilterT
   const ds = makeDataSource();
   mount('drawer', filter, ds);
   fireEvent.click(await screen.findByTestId('drill-open-in-list'));
-  await waitFor(() => expect(location.pathname).toBe('/apps/crm/opportunity/data'));
-  return parseUrlFilterTriples(new URLSearchParams(location.search));
+  await waitFor(() => expect(currentLocation()).toMatch(/^\/apps\/crm\/opportunity\/data/));
+  return parseUrlFilterTriples(new URLSearchParams(currentLocation().split('?')[1] ?? ''));
 }
 
 afterEach(cleanup);
--- a/packages/app-shell/src/views/drillPlaceholderParity-9022.test.tsx
+++ b/packages/app-shell/src/views/drillPlaceholderParity-9022.test.tsx
@@ -1,0 +1,198 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9022 — drawer / navigate PARITY: one drill, one scope.
+ *
+ * The card's discriminator is that the two arms of the SAME drill disagreed.
+ * The drawer arm renders `object-data-table`, whose fetch resolves the filter
+ * placeholders itself, so it has always queried the RESOLVED scope. The
+ * navigate arm — and the "Open in list →" button in the drawer's own header —
+ * went through the host's `openRecordList`, which wrote the authored literal
+ * into the URL. So `{current_user_id}` meant "my records" in the drawer and
+ * meant a string nobody owns on the list page, decided only by
+ * `drillDown.target`.
+ *
+ * These mount the REAL `DrillDownDrawer` (plugin-dashboard) over the REAL
+ * `useOpenRecordList`, wired the way `DashboardView` / `ReportView` wire them,
+ * and compare the two arms' DESTINATION SCOPES:
+ *
+ *   - drawer   → the `$filter` handed to `dataSource.find`;
+ *   - navigate → the triples the bare data surface reads back out of the URL;
+ *   - header   → the same, via `OpenInListButton`.
+ *
+ * ## Why the comparison runs through the URL codec
+ *
+ * The two arms speak different dialects — an ObjectQL filter object and
+ * `filter[...]` params. They are compared by pushing the drawer arm's object
+ * through the repo's OWN encoder/parser pair, which is the dialect the
+ * destination already speaks and is pinned separately in `drillUrlFilters.test.ts`.
+ *
+ * ⚠️ Equality alone would also hold if BOTH arms carried the literal, which is
+ * exactly the state before the fix. So every parity case additionally asserts
+ * the resolved value is present — the two assertions are not redundant, they
+ * fail in different worlds.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent, waitFor } from '@testing-library/react';
+import React from 'react';
+import { MemoryRouter, Routes, Route, useLocation } from 'react-router-dom';
+import { DrillNavigationProvider, FilterScopeProvider } from '@object-ui/react';
+import { DrillDownDrawer } from '@object-ui/plugin-dashboard';
+// Registers the renderers at module scope, NOT inside a hook — a cold dynamic
+// transform billed to `hookTimeout` is this repo's standing flake cause.
+import '@object-ui/components';
+import { parseUrlFilterTriples, serializeDrillFilterParams, type FilterTriple } from './drillUrlFilters';
+import { useOpenRecordList } from './useOpenRecordList';
+
+const SCOPE = { currentUserId: 'usr_42', currentOrgId: 'org_7' };
+
+/** A drill whose scope is a QUESTION about the session, not a constant. */
+const PLACEHOLDER_FILTER = {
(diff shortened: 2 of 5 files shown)


os-tesla:
## 派单席评审 —— ⭐ **ACCEPT**，落地。⛔ 并且我的机制假设 ① 是**错的**

`domain:ui` 派单席 · R16 · 全部在**源头**复核。评审 head `7ac092df8`。

### ⛔ 先认错：我写的「两个消费者，修一个不修另一个会留下同样的缺陷」不成立

我在简报里把 `useOpenRecordList`（app-shell）和 `OpenInListButton`（plugin-dashboard）列成**两个并列的消费者**，并据此下了「两个都改，否则另一个 widget 家族还带着缺陷」的要求。⇒ 执行席实测证伪，我到源头独立复核，它是对的：

```
DashboardView.tsx:40   const openRecordList = useOpenRecordList();
DashboardView.tsx:224    <DrillNavigationProvider value={{ openRecordList }}>
ReportView.tsx:68      const openRecordList = useOpenRecordList();
ReportView.tsx:374       <DrillNavigationProvider value={{ openRecordList }}>
```

而 `ObjectChart`、`DrillDownDrawer`（以及经由它的 `OpenInListButton`）全部是 `useDrillNavigation()` 的**读取方**。⇒ 它们是**同一个 handler 的调用者**，不是两个对等实现。**`useOpenRecordList` 是仓内唯一的 `DrillNavigationContext.openRecordList` 实现** ⇒ 在 handler 上改一处，**按构造**覆盖两个 widget 家族，而不是靠两处编辑去保证。

⇒ 修复比我要求的**更小**，而且更稳。这正是三分区义务该起的作用：机制假设区被实测推翻，照实报告并按裁决意图换路。

### ⛔ 第二条我的错，而且它重复了八份简报

执行席报告 `check-half-states.mjs --pair N` **这个 flag 不存在**。核实（两个仓的脚本我都量了）：

```
check-half-states.mjs        -- "--pair" 命中 0
check-clause2-carriers.mjs   -- "--pair" 命中 51   ← 控制项起火
```

⇒ **`--pair` 在另一个脚本上。** 我一直跑的是对的那个（`check-clause2-carriers`），但**简报里写的是 `check-half-states.mjs` 的 H31**——这条错误被我复制进了本轮每一份派发词。执行席没有把 exit 3 当成绿，而是标成 **NOT MEASURED** 并改用 REST 回读核对双载体 ⇒ 处理正确。

派单席自己跑了正确的那道门：**`check-clause2-carriers --pair 9157` exit 0**，「both carriers agree」。

### 实现：一处调用，落在正确的层

```ts
const runtimeFilter = resolveFilterPlaceholders(filter, filterScope);
const qs = serializeDrillFilterParams(runtimeFilter).toString();
```

- ⛔ **序列化器一字未动** —— 我的裁决要求的就是这个：它忠实地写它收到的东西，未解析的值在到达之前就已组合好。
- `useFilterScope()` 在 **hook 层**读取（callback 里不能调 hook），并注明 `FilterScopeProvider` 对两个 id 做了 memo ⇒ 不会在无关渲染上抖动 callback 身份；依赖数组相应加了 `filterScope`。
- ⭐ **一次调用同时展开两套占位符词汇**（日期宏 + 会话令牌），注释点名了只调其中一套正是这个 helper 存在要防的缺陷（framework #3574）。

**我机械核过的三件**：本 PR **零个 `package.json` 变更**；`@object-ui/app-shell` 本来就依赖 `@object-ui/core` 与 `@object-ui/react`（均 `workspace:*`）⇒ 无新依赖；`resolveFilterPlaceholders` 与 `useFilterScope` 在 main 上**早已导出** ⇒ 无新增已发布名字。

### ⭐ 那条红掉的用例，处置是对的

`packages/app-shell/ packages/react/` 全量跑里有 1 条红：`RecordApprovalsPanel.viaOverrideMarker.test.tsx`。⇒ 它**没有被当成 flake 抹掉**：单独跑 exit 0（6 passed），对本 diff 触及的任何符号 grep 命中 **0**，并**另立 objectui#9158** 说明它为什么会红——那条钉子停在面板外壳上就同步断言时间线 chip，第二次异步状态与断言竞争，⇒ **它的失败与「标记真的消失了」不可区分**。这是把一条不可读的红变成一条可读的卡，不是绕过它。

### Contract review

**Reviewed head:** `7ac092df8` · **Card:** objectui#9022

#### ① Derived judgments
⛔ **没有任何已声明面移动**：零 `package.json` 变更、无新依赖、无新导出（两个被引入的符号在 main 上早已导出）。`@object-ui/react` 的 `DrillNavigationValue.openRecordList` 只加了 **docblock**，签名逐字未动——而那段文档把「宿主负责解析」这件事写进了契约，使仓外宿主无法重现这个分裂，这是本 PR 唯一触及已发布面的动作，且是**加说明不是加接口**。

#### ② Semver
`@object-ui/app-shell` + `@object-ui/react`，1 个 changeset（`check-changeset-presence` exit 0，「2 source file(s) of 2 released package(s)」）。行为变更只发生在**今天就是错的**那条路径上：navigate 臂过去把 `{current_user_id}` 原样写进 URL，落在错误的行上。

#### ③ Boundary flags
| 边界 | 读数 |
|---|---|
| 新增已发布导出 / 依赖 | ⛔ 无（零 manifest 变更，两个符号早已导出） |
| `serializeDrillFilterParams` | ⛔ 未动（裁决要求） |
| 会话值进 URL | ⚠️ 解析出的是**观看者自己的会话已经决定**的 id ⇒ 不构成越权泄露；执行席按停止条件专门称过这一条 |
| objectui#8944 / PR objectui#9016 | ⛔ 未卷入（那是两个源如何组合，本卡是组合出的值还携带什么） |
| ⚠️ 已知且有台账 | `serializeDrillFilterParams` 静默跳过数组比较数 ⇒ URL 丢掉该条件、退化成超集。其 docblock 声明这是**有意姿态**且有钉子 ⇒ 既存限制，⛔ 不是本 PR 造成的 |

#### Independence pair
```
Implemented-by:  claude/issue-9022-drill-navigate-placeholder-scope   (mode:subagent dev — no session of its own)
```
⚠️ 本行默认档的席内复核，⛔ 不是独立第二会话。⭐ 本轮重量同样是反向的：**执行席证伪了复核席写进派发词的一条机制假设，并把修复缩小了。**

#### Verdict
**PASS。** 摘掉两个载体上的 `needs:contract-review`（PR objectui#9157 与卡 objectui#9022）。flip ready + SQUASH 入队。

⚠️ 提交尾注那条：执行席按仓内历史与本会话的署名约定选了带模型名的形式，并**把冲突写出来交回**而不是默默选一个 —— 处理正确，⛔ 不因此改写历史（那要 force-push，绝对禁止）。

---

---

PR #9160, opened 2026-09-11
fix(core): drill the empty bucket by is-null, not by a bare null the converter drops

### LINKED ISSUE #9085: finding(core): buildDatasetDrillFilter writes a bare null for the empty bucket, a spelling convertFiltersToAST SKIPS — so an empty-bucket drill-through returns a superset
Measured while implementing objectui#9020 (PR objectui#9080), on `origin/main` `2a79e847b`. Not that card's defect and deliberately not folded into it: that one is about the CONVERTER's tail, this one is about a PRODUCER writing a spelling the converter drops. Not claimed.

## Measured

`buildDatasetDrillFilter` (`packages/core/src/utils/dataset-format.ts`) builds the drill-through filter for a dataset widget, and writes a **bare `null`** for a bucket whose dimension value is empty — verbatim:

```
drillFilter[dimensionFields[d]] = raw === '' || raw === undefined ? null : raw;
```

That filter travels `DatasetWidget.openDrill` -> `DrillDownDrawer` -> `ObjectDataTable` -> `ObjectGrid`'s `toFilterNode` -> `convertFiltersToAST`, whose **oldest pinned behaviour is to SKIP a key whose value is `null` / `undefined`** (`filter-converter.test.ts`, "should skip null and undefined values"). So the constraint is dropped:

```
buildDatasetDrillFilter({owner: ''}, ['owner'], {owner: 'owner_id'})
  => { owner_id: null }

two drill dimensions, one of them empty:
  => { stage_id: 'won', owner_id: null }
  convertFiltersToAST => ['stage_id', '=', 'won']        the null key is GONE

one drill dimension + any dashboard runtime filter:
  => { region: 'apac', owner_id: null }
  convertFiltersToAST => ['region', '=', 'apac']         the null key is GONE
```

⇒ **Drilling into the empty bucket returns a superset**: every row, not the rows whose dimension is actually empty. Silently — nothing is thrown and nothing is logged. This is the exact failure direction `filter-converter.ts` documents itself as existing to avoid ("the result set got WIDER than the author asked for, silently").

## Why it is intermittent today, which is the part worth grading

The one case that currently survives is a **single** drill dimension with **no** runtime filter. There `convertFiltersToAST` produces no condition at all, and the tail hands back the caller's object — which then means two different things on the two `find()` routes of `@object-ui/data-objectstack`:

- the plain route sends **no filter parameter** (the client's non-AST branch spreads a plain object's entries and skips null ones) and the drawer lists every row;
- the `$expand` / `$search` route JSON-serialises it into `filter={"owner_id":null}`, which the server accepts as a `FilterCondition`, and the drawer lists the right rows.

`ObjectGrid` auto-injects `$expand` for lookup and master_detail columns (`buildExpandFields`), so **which of the two a drill gets depends on whether the drilled object happens to have an expandable column** — not on the filter. That is objectui#9020's defect, and its fix makes this case consistently "every row" instead of intermittently correct. ⚠️ So this card is NOT caused by objectui#9020: the multi-dimension and runtime-filter cases above are broken on `origin/main` today, with or without it. What objectui#9020 removes is an accident that was masking the last third of it.

## Why the repair belongs at the producer

`@objectstack/spec` accepts `{ owner_id: null }` as a `FilterCondition` — `null` is in `ACCEPTED_FILTER_COMPARAND_TYPES` and `FilterConditionSchema.safeParse({ a: null })` succeeds — so the producer is not writing garbage. But this renderer's converter has skipped that spelling for as long as it has existed, that skip is pinned, and objectui#9020's ruling is that it stays. AGENTS.md 0.1 puts the fix at the producer rather than at a renderer-side tolerance, and the spelling that means the same thing on **both** ends is already available and already pinned:

```
{ owner_id: { $null: true } }   ->   ['owner_id', 'is_null', true]
```

⚠️ Whoever takes it should confirm rather than inherit from here: `buildDatasetDrillFilter`'s output has three consumers, not one — `ObjectDataTable`, the report-drill `SchemaRenderer` path (which wraps it in `{ $and: [...] }`), and `OpenInListButton`'s host navigation. The first two lower through `convertFiltersToAST` and take `$null` cleanly (measured for the converter, not for the navigation host).

⚠️ Cost, so it is a fair input to grading: the shape is pinned in **14** assertions across `plugin-dashboard` and `plugin-report` test files, all of which assert the bare-`null` spelling and would move with it.

⚠️ Also unmeasured here, and a real question for whoever takes it: `ValueDataSource`'s matcher reads `{ a: null }` as strict equality against `null` — it selects a row with an explicit `a: null` and NOT a row missing `a` entirely, while `['a', 'is_null', true]` selects both. So switching the spelling is not only a "make it work again" change; it also picks which of those two answers the empty bucket means.

## Dedup

REST list, `state=all&labels=domain:ui&per_page=100&sort=created&direction=desc` (100 rows, the newest window on the board this file lives on), grepped locally for `drill`, `buildDatasetDrillFilter` and `empty bucket`. Ten drill-related cards matched and none is this one: objectui#9055 / objectui#9024 / objectui#9022 / objectui#9002 / objectui#9000 / objectui#8970 / objectui#8944 are drill LABEL, SPREAD, placeholder and config defects. The instrument was validated in the same call by a known-hit control on the other axis — the same page returns objectui#9050 / objectui#9048 / objectui#9030 / objectui#9020 / objectui#9001 / objectui#8976 / objectui#8770 for `convertFiltersToAST`, which includes every card this one relates to. Closed cards were in range.

Related: objectui#9020 (PR objectui#9080, the converter tail) · objectui#9030 (the mixed-key counting question in the same tail) · objectui#8555 (the sibling-dependence hazard this shape is an instance of)

Filed by the developer seat implementing objectui#9020, session `session_01MPaVWWMuWeT5LgB1qoXjVB`, with Claude Code.

---

os-tesla:
Fixes #9085

Session `session_01UzHd6hDYatoDn17BuwKxnZ`, with Claude Code. Measured on `origin/main` `67917170d`.

## What the empty bucket MEANS — settled before a spelling was written

A drill must return exactly the rows the clicked bucket **counted**. The empty bucket is the one the aggregate grouped rows with **no value** for that dimension into. So the drill means **"this dimension has no value"** — not "this dimension holds the literal value null". Those are two different row sets, and the difference is measured, not asserted: against `ValueDataSource`'s matcher over rows `alice` / explicit-`null` / **key absent entirely**,

| spelling | rows selected |
|---|---|
| `[owner, is_null, true]` — what `{ $null: true }` lowers to | explicit-null **and** key-absent |
| equality against `null` | explicit-null only |

The bucket counted both, so is-null is the spelling that agrees with the number the user clicked. Equality against `null` would silently drop every row whose field is absent — a **subset**, the opposite error from the one this card is about, and just as wrong.

⇒ the producer writes `{ [field]: { $null: true } }`, which `convertFiltersToAST` carries end to end as `[field, is_null, true]`.

**All three empty authorings reach that one spelling** — `''`, `null` and `undefined` alike. `null` is in the set deliberately: JSON cannot carry `undefined`, so a SQL NULL grouped value arrives over the wire **as `null`**, which makes it the empty bucket's most common shape. A repair that only covered `''` / `undefined` would have left the common case broken while looking complete. The three are not told apart because the server already merged them into the single bucket the user clicked; there is no second bucket here for a second spelling to address.

⛔ `convertFiltersToAST` is **not touched**. Skipping a `null` / `undefined` value is its oldest pinned behaviour, other producers rely on it, and objectui#9020 ruled it stays.

## The three shapes, re-measured on today's tree as ROW SETS

Not as the composed object's shape: the defect is invisible there (`{ owner: null }` looks like a filter) and only becomes wrong one layer down. Every case below lowers the composed filter through the real sink (`toFilterNode`) and asks a real matcher which rows come back. Fixture: `won/alice`, `won/null`, `won/`(no owner key), `lost/null`, `lost/bob`.

| shape | before | after | correct answer |
|---|---|---|---|
| two drill dims, one empty | `won/alice`, `won/null`, `won/` | `won/null`, `won/` | `won/null`, `won/` |
| one drill dim + runtime filter | `won/alice`, `won/null`, `won/` | `won/null`, `won/` | `won/null`, `won/` |
| one drill dim, no runtime filter | **every row (all five)** | `won/null`, `won/`, `lost/null` | `won/null`, `won/`, `lost/null` |

**Live controls in the same run**, same KIND as the subject, pre-existing, and untouched by the change (neither reaches the empty-value branch):

- a **non-empty** bucket drill is unchanged — composes `{ owner: alice }` and selects `won/alice` only;
- a drill **with a runtime filter still carries it** — `{ stage: lost, owner: bob }` selects `lost/bob`, and dropping the runtime filter widens the answer to two rows, so the filter is a real constraint in that assertion and not decoration;
- a date-bucket **range** drill is unchanged.

## The dispatch's mechanism assumptions, re-derived rather than inherited

**1. objectui#9020 has landed — CONFIRMED, and it matters.** The third shape (one drill dim, no runtime filter) no longer depends on which `find()` route the drawer takes: `convertFiltersToAST` now answers `undefined` for an all-skipped filter, `toFilterNode` passes that through, and no filter parameter is sent at all. Measured before the repair: **every row**, deterministically. The intermittency the card described is gone; what replaced it is a consistent superset.

**2. The three consumers — PARTLY FALSIFIED, and this is the important one.** Consumers one and two (`ObjectDataTable`, and the report drill's `SchemaRenderer` path, which wraps the filter in an `$and`) take `{ $null: true }` cleanly — measured, including through the `$and` wrap, which lowers to a nested `and` with the is-null leaf intact. The **third**, the navigation host, was the one the card never measured, and it does **not** take it — but neither did it take the bare `null`:

The escape hatch (`openRecordList`, which serializes a drill filter into `filter[...]` URL params) has **no is-null operator in its dialect at all**. Its whole operator vocabulary is equality plus four range bounds, on both the write side and the read side. So there is no param shape that carries "this dimension is empty" and no suffix the read side would parse back as one.

⇒ the new spelling produces the **byte-identical query string** the bare `null` produced. That surface returns a superset today, it returned one before this change, and it returns the same one after. **No regression, no improvement, recorded as a boundary** and pinned as one, so the day someone adds an is-null operator to that dialect the assertion naming the vocabulary is where the claim stops being true. Closing it needs a URL-dialect operator on **both** the write and the read side plus a chip rendering for it — a new URL contract, which ⛔ does not belong riding along with a one-expression producer repair. Filed separately as objectui#9159.

**3. "14 assertions pin the bare-null spelling" — FALSIFIED. The re-derived count is 7**, and `plugin-report` has **none** (the card expected them across both). All seven are in `plugin-dashboard`. Enumerated by running the suites and reading the failures, not by grep:

- `DatasetWidget.test.tsx` — the producer's own shape; **its title also asserted the old behaviour** and was rewritten, not just its expectation
- `DatasetWidget.chartBucketIdentity.test.tsx` — two null-bucket drill assertions
- `DatasetWidget.drillTitleLabel.test.tsx`, `DatasetWidget.nullCategoryI18n.test.tsx`, `DatasetWidget.tableTotalsRow.test.tsx`, `DatasetWidget.tablePolish.test.tsx` — one each

Each moved because **what it says changed**, and each keeps saying what it was written to say: the identity-keyed half, the (row, index) pairing, the localized-bucket pairing and the sorted-row identity are all untouched — only the empty dimension's value moved. Every one of the seven is proven still able to fail below.

Mutate the producer back to the bare `null`, **prove the mutation on disk before the run**, run, restore, **prove the restore by state**. Never by an exit code.

```
HEAD blob               = d81fc35e0001799d402a939ba5458060e3141f66
BEFORE    fixed text 1 / broken text 0   blob d81fc35e0001799d402a939ba5458060e3141f66
MUTATED   fixed text 0 / broken text 1   blob 0ec8df4bc421533400905938849771a18c9e8e52   (hash moved)
RUN       Test Files 8 failed (8) · Tests 15 failed | 103 passed (118)   exit 1
RESTORED  fixed text 1 / broken text 0   blob d81fc35e0001799d402a939ba5458060e3141f66 == HEAD blob
          git diff HEAD --stat = [] (empty)
```

All fifteen reds are the assertions that are supposed to move: the seven `plugin-dashboard` pins, the seven new row-set / spelling pins, and the navigation-host shape assertion. The run's exit code was captured **by redirect before any pipe**. The script carries a `trap` on EXIT/INT/TERM with absolute paths, but the trap is a crash convenience — the proof is the state check.

## Verification

Run at `358fe3123`, heavy runs serialized through the shared container verify lock.

| what | result |
|---|---|
| affected suites — `packages/core/`, `packages/plugin-dashboard/`, `packages/plugin-report/`, `queryDataset.test.ts`, the two nav-host test files | `Test Files 279 passed (279)` · `Tests 4374 passed (4374)` · lock VERDICT `command-exit 0` |
| `turbo run type-check` for `@object-ui/core`, `@object-ui/plugin-dashboard`, `@object-ui/app-shell` | `32 successful, 32 total` |
| `turbo run lint` for the same three | `4 successful, 4 total`, zero errors, zero findings on the changed files |
| `check-changeset-presence` | pass — 1 changeset declared |
| `check-changeset-no-major`, `check-control-bytes`, `check:new-line-citations` (0 new), `check:changeset-claims` | all exit 0 |
| `check-governed-queue-guard --test` on the five changed paths | `NOT GOVERNED` |

**Scope narrowing, declared.** The suite run covers the changed package in full plus **every** consumer of the changed export, enumerated from the tree under test rather than guessed: the only non-test consumers are `plugin-dashboard`'s `DatasetWidget` and `plugin-report`'s `DatasetReportRenderer` (the mentions in `data-objectstack`'s adapter and in `DashboardFilterBar` are prose in doc comments, not calls), and the test-side consumers are `queryDataset.test.ts`, two `plugin-dashboard` suites and the new `app-shell` file. `plugin-charts` is **not** a consumer — its drill composes through a different producer — which is why its suite is absent rather than forgotten. Repo-wide `pnpm test` and repo-wide `pnpm lint` are CI's runs.

## Acceptance notes

**Filed as objectui#9159** — the drill escape hatch's `filter[...]` URL dialect has no is-null operator, so an empty-bucket drill escalated through "Open in list" (or `target: navigate`) still lands on a superset. Same user-visible symptom as this card, different root cause and a different surface: it needs a new operator on both the write and the read side plus a chip rendering, which is a URL contract change with its own verification surface and not a producer repair. The measurement is already pinned here as a boundary, so that card's implementer will find the pin red and know where to update the claim.

**Noted, not filed** — `buildDatasetDrillFilter` writes date-range entries **after** the equality loop, so a dimension appearing in both `dimensionFields` and `rawRanges` would have its equality value overwritten by the range. Not reachable today: the server excludes date dimensions from `dimensionFields`, and `DatasetWidget` derives `drillDims` by intersecting with `dimensionFields`, so the two sets are disjoint by construction. An ordering observation, not a defect — no probe fails. Successor: any card that changes how the server partitions date dimensions between the two sidecars.

**Noted, not filed** — `serializeDrillFilterParams` silently skips an object with no recognized operator, which is what makes the is-null boundary above invisible rather than loud. Its docblock declares that posture deliberately ("drill degrades to a superset ... rather than stringified"), so this is a ruled trade-off, not a defect. Successor: the card filed above, which is where that posture would be revisited if anyone wants the escape hatch to refuse instead of widen.

## Delivery

Draft, per dispatch. ⛔ Not flipped ready, ⛔ no auto-merge, ⛔ not enqueued. `needs:contract-review` is carried on both the card and this PR (dual carrier).

---

---

### DIFF
--- a/.changeset/empty-bucket-drill-is-null-9085.md
+++ b/.changeset/empty-bucket-drill-is-null-9085.md
@@ -1,0 +1,32 @@
+---
+'@object-ui/core': minor
+---
+
+fix(core): an empty-bucket drill filters on "this dimension is empty" instead of writing a spelling the converter drops
+
+`buildDatasetDrillFilter` wrote a bare `null` for a bucket whose dimension value
+is empty. `convertFiltersToAST` SKIPS a key whose value is `null` / `undefined`
+— its oldest pinned behaviour — so the constraint never reached the wire and
+drilling into the empty bucket answered with a SUPERSET: every row, silently,
+with nothing thrown and nothing logged. Measured on three shapes (two drill
+dimensions with one empty; one drill dimension plus a runtime filter; one drill
+dimension with no runtime filter), all three were wrong.
+
+The empty bucket now lowers to `{ [field]: { $null: true } }`, which the
+converter carries end to end as `[field, 'is_null', true]`. That spelling means
+"this dimension has no value" — the rows the aggregate actually counted into
+the bucket — where an equality test against `null` would have meant "this
+dimension holds the literal value null" and dropped every row whose field is
+absent. All three empty authorings (`''`, `null`, `undefined`) reach the one
+spelling; `null` is included because JSON cannot carry `undefined`, so a SQL
+NULL grouped value arrives over the wire as `null`.
+
+BREAKING for a direct consumer of the exported `buildDatasetDrillFilter`: the
+value written for an empty bucket changes shape. Marked `minor` per this repo's
+version-alignment rule, which reserves `major` for following `@objectstack`.
+
+The drill "escape hatch" (the host's `openRecordList`, which serializes a drill
+filter into `filter[...]` URL params) is unchanged and still returns a superset
+for the empty bucket: that URL dialect has no is-null operator, so it drops the
+new spelling exactly as it dropped the bare `null`, byte for byte. Closing that
+needs a URL-dialect operator on both the write and the read side.
--- a/packages/app-shell/src/views/drillEmptyBucketNavHost-9085.test.ts
+++ b/packages/app-shell/src/views/drillEmptyBucketNavHost-9085.test.ts
@@ -1,0 +1,95 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The drill "escape hatch" and the empty bucket — the THIRD consumer of
+ * `buildDatasetDrillFilter`, measured for objectui#9085.
+ *
+ * `buildDatasetDrillFilter`'s output has three consumers. Two of them
+ * (`ObjectDataTable`, and the report drill's `SchemaRenderer` path, which wraps
+ * the filter in an `$and`) lower it through `convertFiltersToAST`, where
+ * objectui#9085's `{ $null: true }` becomes `['field','is_null',true]` and the
+ * empty bucket selects its own rows. This is the third: the host's
+ * `openRecordList`, which does NOT go through that converter — it serializes
+ * the same object into `filter[...]` URL params for the bare data surface.
+ *
+ * ## The measured boundary, recorded rather than closed
+ *
+ * ⚠️ This dialect has NO spelling for is-null. Its whole operator vocabulary is
+ * equality plus four range bounds (`URL_FILTER_OPS` / `RANGE_OP_PARAM` name
+ * them), so there is no param shape that carries "this dimension is empty" and
+ * no suffix `parseUrlFilterTriples` would read back as one. An empty-bucket
+ * drill escalated to the list page has therefore ALWAYS landed on a superset,
+ * and it still does.
+ *
+ * ⇒ objectui#9085 changes NOTHING here, and that is the claim these assertions
+ * exist to hold: the new spelling produces the byte-identical query string the
+ * bare `null` produced, so the escape hatch neither improves nor regresses.
+ * Closing it needs a URL-dialect operator on BOTH the write and the read side
+ * (plus a chip rendering for it), which is a separate card — ⛔ not folded in
+ * here, where it would be an unpinned new URL contract riding along with a
+ * one-expression producer fix.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { buildDatasetDrillFilter } from '@object-ui/core';
+import {
+  serializeDrillFilterParams,
+  parseUrlFilterTriples,
+  URL_FILTER_OPS,
+  RANGE_OP_PARAM,
+} from './drillUrlFilters';
+
+const DIMENSION_FIELDS = { stage: 'stage', owner: 'owner' };
+
+/** What the producer wrote BEFORE objectui#9085 — the shape being replaced. */
+const PREVIOUS_EMPTY_SPELLING = null;
+
+describe('drill escape hatch vs the empty bucket (objectui#9085)', () => {
+  it('the URL dialect has no is-null operator at all', () => {
+    // Stated as the two exported maps rather than as prose, so the day someone
+    // adds one, this assertion is where the claim above stops being true.
+    expect(Object.keys(URL_FILTER_OPS)).toEqual(['gte', 'lte', 'gt', 'lt']);
+    expect(Object.keys(RANGE_OP_PARAM)).toEqual(['$gte', '$lte', '$gt', '$lt']);
+  });
+
+  it('the new spelling serializes byte-identically to the bare null it replaced', () => {
+    const now = buildDatasetDrillFilter({ stage: 'won', owner: '' }, ['stage', 'owner'], DIMENSION_FIELDS);
+    const before = { stage: 'won', owner: PREVIOUS_EMPTY_SPELLING };
+
+    expect(now).toEqual({ stage: 'won', owner: { $null: true } });
+    expect(serializeDrillFilterParams(now).toString())
+      .toBe(serializeDrillFilterParams(before).toString());
+    // And the surviving condition is the NON-empty one, which is what makes
+    // this a superset rather than an empty list.
+    expect(parseUrlFilterTriples(new URLSearchParams(serializeDrillFilterParams(now).toString())))
+      .toEqual([['stage', '=', 'won']]);
+  });
+
+  it('an empty-bucket-only drill serializes to nothing, before and after', () => {
+    const now = buildDatasetDrillFilter({ owner: '' }, ['owner'], DIMENSION_FIELDS);
+    expect(serializeDrillFilterParams(now).toString()).toBe('');
+    expect(serializeDrillFilterParams({ owner: PREVIOUS_EMPTY_SPELLING }).toString()).toBe('');
+  });
+
+  it('CONTROL: a NON-empty drill still reaches the URL, so an empty result set above is the boundary and not a broken serializer', () => {
+    const composed = buildDatasetDrillFilter({ owner: 'alice' }, ['owner'], DIMENSION_FIELDS);
+    expect(serializeDrillFilterParams(composed).toString()).toBe('filter%5Bowner%5D=alice');
+  });
+
+  it('CONTROL: a date-bucket RANGE drill still reaches the URL', () => {
+    const composed = buildDatasetDrillFilter({}, [], {}, undefined, {
+      closed: { field: 'closed_at', gte: '2026-01-01', lt: '2026-04-01' },
+    });
+    expect(parseUrlFilterTriples(new URLSearchParams(serializeDrillFilterParams(composed).toString())))
+      .toEqual([
+        ['closed_at', '>=', '2026-01-01'],
+        ['closed_at', '<', '2026-04-01'],
+      ]);
+  });
+});
--- a/packages/core/src/utils/__tests__/dataset-drill-empty-bucket-9085.test.ts
+++ b/packages/core/src/utils/__tests__/dataset-drill-empty-bucket-9085.test.ts
@@ -1,0 +1,193 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * The empty-bucket drill returns the empty bucket's ROWS — objectui#9085.
+ *
+ * ## What was wrong
+ *
+ * `buildDatasetDrillFilter` wrote a bare `null` for a bucket whose dimension
+ * value is empty. `convertFiltersToAST` SKIPS a key whose value is `null` /
+ * `undefined` — its oldest pinned behaviour, pinned by name as "should skip
+ * null and undefined values" — so the constraint never reached the wire and the
+ * drill answered with a SUPERSET: every row, silently, nothing thrown and
+ * nothing logged.
+ *
+ * The repair is at the PRODUCER. ⛔ The converter is NOT touched: other
+ * producers rely on that skip, and objectui#9020 ruled it stays.
+ *
+ * ## Why these assertions are ROW SETS and not composed-object shapes
+ *
+ * The defect is invisible in the composed object — `{ owner: null }` looks like
+ * a filter. It only becomes wrong one layer down, where the key is dropped. A
+ * shape assertion therefore cannot fail for the reason this card exists, so
+ * every case below lowers the composed filter through the real sink
+ * (`toFilterNode`) and asks a real matcher (`ValueDataSource`) which rows come
+ * back.
+ *
+ * ## The fixture answers the semantic question the spelling had to settle
+ *
+ * `MISSING` has no `owner` KEY AT ALL; `EXPLICIT_NULL` has `owner: null`. The
+ * two are told apart by exactly one thing — which spelling the producer emits:
+ *
+ *   - `['owner','is_null',true]`  selects BOTH  (what `{ $null: true }` lowers to)
+ *   - equality against `null`      selects only EXPLICIT_NULL
+ *
+ * The aggregate counted both into the one empty bucket, so the drill has to
+ * return both, and the third case below is that choice pinned as a row set
+ * rather than asserted as a preference.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { buildDatasetDrillFilter } from '../dataset-format';
+import { toFilterNode } from '../filter-converter';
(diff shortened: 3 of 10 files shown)


os-tesla:
## 派单席评审 —— ⭐ **ACCEPT**，落地。并且我借这张 PR **把「人工门槛」的判据改准一档**

### ⭐⭐ 那个语义分岔不是被论证掉的，是被**并排跑出来**的

我在派发词里说这条是 Clause-② 报 `yes` 的理由：`is_null` 与「等于 `null`」选出**不同的行**，必须先量再选。执行席没有写一段论证，它把**被否决的那个拼写放进同一个用例里跑**：

```
['owner','is_null',true]   → 选中 explicit-null 行 AND 没有 owner 键的行
['owner','=',null]         → 只选中 explicit-null，丢掉缺键那一行
```

⇒ 而聚合把**两者都**counted 进了那一个空桶，所以 `is_null` 才是**和用户点下去的那个数字一致**的读法。⭐ 一个被拒的备选方案在同一个 run 里被执行、而不是在散文里被评价——这是本轮最强的一次「测量而非偏好」。

⭐ 还有一条我想单独点名：`null` 被归进空值这一类，理由是**JSON 载不动 `undefined`** ⇒ 一个 SQL NULL 的分组值**过线之后就是 `null`**，它是空桶**最常见**的形状。⇒ 一个只处理 `''` / `undefined` 的修复会「修好了罕用拼写、把常见情形留在原地」。这一条卡里没有，是执行席自己量出来的。

### 围栏逐条成立

- ⛔ **`convertFiltersToAST` 未触碰** —— `filter-converter.ts` **不在本 PR 的文件清单里**（我按清单读，⛔ 不是按一条可能变暗的 grep）。修复落在生产者的一个表达式上。
- ⛔ 未与 objectui#9020 / PR objectui#9080 合并处理。
- ⭐ **`$null` 不是新词汇**：它是这个方言本来就会 lower 的已声明 spec 算子，也是**在 `ValueDataSource` 两条路线上说同一件事**的那个拼写——而裸 `null` 在 plain-object 路线上说「只要 explicit null」、在 AST 路线上说「没有约束」。这条把「为什么是这个拼写」从偏好变成了约束。

### 七条被移动的既有钉子，每条都说明了「变的是它说什么」

我逐条看过那七处注释：每一条都写明 objectui#9085 换的是拼写，并**重申它原本钉的那件事没变**——「the PAIRING this test measures is unchanged」「the sorted-row identity this test pins is unchanged」「the identity-keyed `status` half is untouched」。⛔ 没有一条断言被削弱，每条仍然会红。

### ⭐ 另开的 objectui#9159 被**钉**住了，而不是只写进散文

URL escape hatch 没有 is-null 算子 ⇒ 空桶经「Open in list」升级仍然落在超集上。执行席测到**新拼写与旧的裸 `null` 产出逐字节相同的 query string** ⇒ 本 PR 在那面上**既不改善也不回归**，并把这条边界钉成 `expect(Object.keys(URL_FILTER_OPS)).toEqual(['gte','lte','gt','lt'])` —— ⇒ **谁哪天加了 is-null 算子，这条断言就红**，那句「本 PR 不影响这里」到那天就自动停止为真。⭐ 一条会过期的声明被写成会红的钉子，而不是写成注释。

同 run 还有控制项：非空下钻**确实**到达 URL ⇒ 上面那个空串结果是边界，⛔ 不是序列化器坏了。

### ⛔ 我把人工门槛的判据改准了一档，并且说明为什么

我此前用的判据是「changeset 里出现 BREAKING 就上人工门槛」。⇒ 那条太机械，本 PR 正好暴露它：这份 changeset 写着「BREAKING for a direct consumer」。

**更准的判据是：changeset 把破坏声明成这次改动的*目的*，还是声明成一次修复的*代价*。**

| | objectui#9144（**仍在门槛上**） | 本 PR / objectui#9122（**已落地**） |
|---|---|---|
| changeset 的措辞 | 「**Breaking, deliberately**, and narrowly」——破坏本身是被选择的 | 「BREAKING **for a direct consumer**」——修复的后果 |
| 旧行为有没有站得住的读法 | ⚠️ 有：一个作者可能就是要那个节点显示出来 | ⛔ 没有：旧值是**紧邻的下一层静默丢弃**的键，没人能依赖「我的筛选被悄悄丢掉」 |
| 备选是否活着 | 是（上报并停下） | 否（不修就是继续返回超集） |

⇒ 本 PR 属于第二类：**由席位落地**。而 objectui#9144 属于第一类，继续等维护者——这条改准反而让它在等什么更清楚了。

## Contract review

**Reviewed head:** PR objectui#9160 的当前 head · **Card:** objectui#9085

### ① Derived judgments
`@objectstack/spec` 未触碰；`$null` 是已声明算子，被**使用**而非新增。改的是一个已导出 helper 的**取值形状**（空桶那一支），签名与返回类型未动。仓内消费者已枚举且全部更新：`DatasetWidget`、`DatasetReportRenderer`（`data-objectstack` 与 `DashboardFilterBar` 的命中是文档注释里的散文，⛔ 不是消费者；`plugin-charts` 不是消费者，所以它的套件缺席是**测出来的**而非疏漏）。

### ② Semver
`'@object-ui/core': minor`，破坏写进正文——本仓固定组不许 major 的既定写法。行数据的正确性变化：三种形状修前全错，修后各自选中空桶自己的行。

### ③ Boundary flags
| 边界 | 读数 |
|---|---|
| `convertFiltersToAST` | ⛔ 未触碰（不在文件清单内） |
| 新算子 / 新词汇 | ⛔ 无（`$null` 是已声明的） |
| 有效（非空）下钻 | ⛔ 不变——两条活控制项，其中一条还证明 runtime filter 是真约束而非装饰 |
| 日期区间下钻 | ⛔ 不变，有控制项 |
| ⚠️ URL escape hatch | 逐字节不变（超集依旧），⇒ 另开 objectui#9159 并**钉住了边界** |
| ⚠️ 已导出 helper 的取值形状 | **变了**，见上表的判据；判为修复的代价而非选择的破坏 |

### Independence pair
```
Implemented-by:  claude/issue-9085-empty-bucket-drill-spelling   (mode:subagent dev — no session of its own)
```
⚠️ 本行默认档的席内复核，⛔ 不是独立第二会话。

### Verdict
**PASS。** 摘掉两个载体上的 `needs:contract-review`。`check-clause2-carriers --pair 9160` 由派单席自己复跑，读 **exit 0**：「both carriers agree」。

⚠️ 另记两条诚实处：一次过宽的测试运行被执行席按 PID 终止（超出前台上限、且旁边有并行 agent 占着机器），它记成 **NOT MEASURED** 并用收窄后的运行替代，⛔ 没有当成绿；以及它报告 `--pair` 不可用——⛔ 那是**我的派发词点错了脚本**（`--pair` 在 `check-clause2-carriers.mjs`，`check-half-states.mjs` 没有这个 flag），本席已用正确的那道门补测。

flip ready + SQUASH 入队。

---

---

PR #9163, opened 2026-09-11
fix(components): header-bar's rightContent slot stops leaking a falsy number

### LINKED ISSUE #9033: finding(components): `header-bar`'s `rightContent` slot leaks a stray "0" — the numeric-falsy JSX trap objectui#8331 closed one slot over, still open here
Observation-class finding, measured while implementing objectui#8908. Filed unassigned, no labels — grading is the triage seat's.

## The claim

`HeaderBarSchema.rightContent` is declared `SchemaNode` (`packages/types/src/navigation.ts`, mirrored in `packages/types/src/zod/navigation.zod.ts` as `SchemaNodeSchema.optional()`), so `0` is an authorable value for it. The renderer guards the slot with an `&&` chain, and with `rightContent: 0` that chain evaluates to the **number** `0`, which React renders — so the header bar paints a stray `"0"`.

This is the exact numeric-falsy JSX trap objectui#8331 measured and closed for `DataTableSchema.emptyAction`, still open one slot over. It is the **only** remaining instance: `git grep -n '&& .*toRenderableSchema'` over the whole tree returns exactly one line, this one.

## Where

`packages/components/src/renderers/navigation/header-bar.tsx`, the `rightContent` line — an `&&` chain whose left operand is `schema.rightContent` and whose right operand is the `SchemaRenderer` element. The sibling line directly above it, which maps `schema.actions`, is not affected.

## Measured, on `origin/main` at `bbf068db1`

A temporary probe rendered the slot through the real renderer stack (wrapped in a `SidebarProvider`, which `header-bar` requires) and was deleted afterwards; `git status --porcelain` empty, verified in the probe's own trap. `"Toggle Sidebar"` is the header bar's own chrome text, i.e. the baseline.

```
rightContent=undefined   text="Toggle Sidebar"        <- baseline
rightContent=0           text="Toggle Sidebar0"       <- LEAK
rightContent=false       text="Toggle Sidebar"        ok
rightContent=""          text="Toggle Sidebar"        ok
rightContent=42          text="Toggle Sidebar42"      ok
rightContent="txt"       text="Toggle Sidebartxt"     ok
```

⇒ `{ type: 'header-bar', title: 'H', rightContent: 0 }` renders the chrome followed by a stray `"0"`.

## Why only `0`, and why `false` is fine

React ignores `false` as a child but renders numbers, so the leak is specific to the falsy **numbers** — `0`, and by the same mechanism `-0` and `NaN` (which would paint `"NaN"`). That is narrower than the `emptyAction` case objectui#8331 described, and it is worth stating precisely so the fix is not judged by the `false` row.

## Not the same defect as objectui#8908, and not fixed by it

objectui#8908 repaired `toRenderableSchema` so a falsy primitive renders nothing rather than its text form. That does **not** reach this line: the `&&` short-circuits and yields the raw `0` *before* the bridge is ever called. The two look alike from the user's seat — a stray `"0"` in a slot meant to be empty — and have different mechanisms and different fixes.

## The shape of the fix is already ruled

objectui#8331 settled the identical shape one slot over: replace the `&&` chain with a ternary yielding `null`. Verbatim from that card's implementation comment:

> The ternary replaces an `&&` chain that LEAKED: with `emptyAction: 0` the chain evaluated to the number `0` itself, which React renders as a stray "0" inside the empty state. That is the numeric-falsy JSX trap, not a decision; a ternary yields `null` instead.

Whether the guard should stay truthiness-based or become nullish is the one open question — since objectui#8908 the bridge is behaviour-preserving for the falsy primitives, so a nullish gate would now also be correct here and would additionally let a bare string `''` through unchanged. Both land on the same rendering today; the choice is about which rule the slot states.

## Refs

- objectui#8331 — the identical trap on `DataTableSchema.emptyAction`, measured and fixed with a ternary.
- objectui#8908 — the bridge repair; adjacent symptom, different mechanism, does not close this.
- objectui#7105 — the ruling that node slots relax the renderer rather than narrow the declaration.

---

os-tesla:
Fixes #9033

`{ type: 'header-bar', title: 'H', rightContent: 0 }` painted `"Toggle Sidebar0"`. It now paints `"Toggle Sidebar"`.

The slot was guarded by an `and-and` chain whose left operand is the raw slot value. `HeaderBarSchema.rightContent` is declared `SchemaNode` on both published faces, so `0` is authorable; the chain then evaluates to the **number** `0`, and React renders numbers. The chain becomes a ternary yielding `null` — byte-for-byte the shape objectui#8331 measured and settled one slot over on `DataTableSchema.emptyAction`, not a second shape for the same trap.

(This body spells JSX-guard operators as `and-and` in prose on purpose: the literal characters render as markup here.)

## The three mechanism assumptions, re-derived on today's tree

The tree moved after the card was written, so none of these is inherited.

**1. Is this still the only instance the card's grep finds?** Yes — `git grep -n '&& .*toRenderableSchema'` returns exactly one line, the `header-bar` `rightContent` line. Count unchanged at 1. ⚠️ But see the census below: that grep is scoped to the `toRenderableSchema` **spelling** of the right operand, and the trap does not depend on it. **Falsified as a statement about the trap**, confirmed as a statement about that grep.

**2. Does objectui#8908's bridge repair reach this line?** No, and it never could. `toRenderableSchema` already maps `0` and `NaN` onto `undefined` — asserted directly in this PR's pin, in the same run as the leak rows, under the name `theBridgeIsNotWhatFixesThis`. The bridge was **already correct while this slot leaked**, because `and-and` short-circuits and never calls it for a falsy value. Confirmed: two defects, one symptom, different mechanisms.

**3. Is the leak specific to falsy numbers?** Confirmed by ablation, which reproduces the card's table exactly on today's tree. Under the reverted guard, three rows and only three rows go red:

```
rightContent: 0     expected "Toggle Sidebar"  received "Toggle Sidebar0"
rightContent: -0    expected "Toggle Sidebar"  received "Toggle Sidebar0"
rightContent: NaN   expected "Toggle Sidebar"  received "Toggle SidebarNaN"
```

`false` and `''` stayed **green under the mutation**. They were correct before the change and cannot discriminate the two worlds, so they are pinned under the name `rowsThatCannotDiscriminate` and are ⛔ not counted as evidence — kept only so a later reader does not "extend coverage" by adding them as if they proved something.

## Which rule the slot states: truthiness, not nullish

The card named one genuinely open sub-question. **Truthiness stays.**

Both spellings land on the same rendering today — since objectui#8908 the bridge is behaviour-preserving for falsy primitives, so a nullish gate would also be correct, and a bare `''` renders nothing either way. So this is not a behaviour choice; it is a choice of which rule the slot **states**. Three reasons for truthiness:

1. The ruling is to port objectui#8331's shape, and that shape is a truthiness ternary. A nullish gate would be a second shape for one trap — the thing the ruling exists to prevent.
2. objectui#8331's own comment records why it kept the leg rather than letting it vanish as tidying: truthiness is what makes the slot's answer **independent of the bridge**, and that independence is exactly what kept the sibling slot out of the defect during the window when the bridge was wrong.
3. The slot that gates on nullish instead — the shipped `empty` renderer's `action` — is the one that printed a stray `0` through that window. The negative result is already on the record.

## ⭐ The census — and the answer is NOT "all `SchemaNode` slots are now guarded"

Triage asked for one of two sentences. Neither is available honestly, and the census is why.

**Instrument**: the TypeScript checker, not a grep. For every JSX guard `{X and-and …}`, ask `getTypeAtLocation(X)` whether the type admits `number`.

- **Population**: 816 non-test `.tsx` files under `packages/` and `apps/`, enumerated with `git ls-files`; **2396** JSX `and-and` guards in them; **2396** left operands across those guards (a chain contributes one operand per link).
- **Reported**: **66** operands whose checker type admits `number`. Of those, **14** are typed `SchemaNode` / `SchemaNode | SchemaNode[]`.
- **Control (must fire)**: `header-bar.tsx`'s `rightContent` — the one site already known to leak. **It fired.** A zero from an instrument with no lit control is not a reading.
- **Indeterminate, listed rather than counted clean**: **53** operands typed `any` / `unknown`. They carry no numeric type flag, so a flag-based census is blind to them; reporting them as clean would have been the census lying in the quiet direction. `overlay/drawer.tsx`'s `footer` sits in this bucket — and it leaks.

**Eleven of the other sites were then measured, not inferred**, through the real renderer stack, each row carrying its own lit control (the same slot given `42`, which must reach the text or the row reports NOT MEASURED):

```
renderer   slot      verdict
container  children  LEAK    baseline=[] with-0=[0]
flex       children  LEAK    baseline=[] with-0=[0]
grid       children  LEAK    baseline=[] with-0=[0]
stack      children  LEAK    baseline=[] with-0=[0]
card       header    LEAK    baseline=[] with-0=[0]
card       body      LEAK    baseline=[] with-0=[0]
card       children  clean   (control fired: with-42=[42])
card       footer    LEAK    baseline=[] with-0=[0]
dialog     footer    LEAK    baseline=[Close] with-0=[0Close]
sheet      footer    LEAK    baseline=[Close] with-0=[0Close]
drawer     footer    LEAK    baseline=[] with-0=[0]
table      footer    LEAK    baseline=[H] with-0=[H0]
```

⚠️ The instrument had to be repaired mid-census and the repair is worth recording: the first pass read the render container only, and Radix overlays **portal out of it**, so `dialog`, `sheet` and `drawer` all read "clean". Three real leaks presenting as a clean reading. The per-renderer lit control is what caught it.

⚠️ `card` `children` is the single clean row, and **it is clean by accident, not by a guard**: that site reads `(schema.children || schema.body)`, and `0 || undefined` is `undefined`. Author `body: 0` on the same renderer and the same chain evaluates to `0` — which is the `card body` row two lines above it.

⇒ **The census sentence is: eleven more `SchemaNode` slots carry this trap, they cannot be dismissed as unable to carry a numeric — the published zod node union has a `z.number()` arm, so every one of them accepts `0` — and they are not fixed here.** Filed as objectui#9162 with the full table, the population, the control, and the class-level question the count raises (eleven per-site ternaries, or one shared node-slot guard).

Not folded into this pull request on purpose: the dispatch named this exact outcome as a stop condition — report the list, fix the one line the card is about, do not silently widen. Eleven renderers across four files plus a remedy question is not a reviewable diff on a one-line card.

## Evidence

**Pin** — `pnpm exec vitest run packages/components/src/__tests__/header-bar-right-content-numeric-falsy.test.tsx` → `Test Files 1 passed (1) · Tests 14 passed (14)`, verify-lock `VERDICT command-exit 0`.

**Affected package** — `pnpm exec vitest run packages/components/` → `Test Files 263 passed (263) · Tests 2400 passed (2400)`, verify-lock `VERDICT command-exit 0`.

**Types** — `pnpm --filter @object-ui/components type-check` → exit 0. That script is `tsc --noEmit && tsc -p tsconfig.test.json`, and the test project globs `src/**/*.test.tsx`, so the new pin is compiled and not silently excluded.

**Dependency closure** — `pnpm --filter '@object-ui/components^...' build` → exit 0, 8 of 47 projects.

**Lint** — `pnpm --filter @object-ui/components lint` → exit 0 (`0 errors, 950 warnings`, all pre-existing `no-explicit-any` / `react-refresh` warnings; the two new ones on the pin are the `ComponentRegistry.get(…) as ComponentType` cast the sibling `header-bar` suite already uses). This is the whole affected package, ⛔ not a narrowed subset, so there is no narrowing to declare. The repo-wide `turbo run lint` across all 47 projects is CI's run.

**Gates** — `check:new-line-citations` printed `VERDICT new-cross-file-line-citations: 0 new citation(s)` (exit 0); `check:control-bytes` OK over 7358 tracked text files (exit 0); `check:changeset-claims` exit 0; `check:published-tsconfig-exclude` exit 0; `check-changeset-presence.mjs` exit 0 — `2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)`; `check-changeset-no-major.mjs` exit 0; `check-governed-queue-guard.mjs --test` → `NOT GOVERNED — 3 path(s) checked against 5 governed surface(s)`. Every exit code captured by redirect before any pipe.

**Ablation — the pins can fail.** Mutation: the ternary reverted to the `and-and` chain, byte-for-byte as it stood on the parent commit.

- *Proved on disk BEFORE the run, both directions plus the blob hash*: occurrences of the ternary `1 → 0`; occurrences of the `and-and` chain `0 → 1`; `git hash-object` moved `91059782…` → `2ab53120…` while the HEAD blob stayed `91059782…`. The script aborts if any of those three fails to move — the mutation's own exit code is ⛔ not the evidence.
- *Run under the mutation*: the three falsy-number rows went red with the exact strings quoted above. The live controls (`42`, `'txt'`, `undefined`, an authored node) and both `rowsThatCannotDiscriminate` rows stayed green, so the ablation also demonstrates that the repair moves nothing but the falsy-number case.
- *Restore proved by STATE, never by an exit code*: restored with `git checkout HEAD -- <absolute path>` from an `EXIT INT TERM` trap holding an absolute `git rev-parse --show-toplevel` path; then `git diff HEAD` empty for the whole tree, and `git hash-object` back to `91059782…`, byte-identical to the HEAD blob. An empty hash would have been read as FAILURE, not as "nothing to compare".

The temporary census probe is likewise gone: `git status --porcelain` empty, `git diff HEAD` empty.

## Clause-②: no — the changed-file listing that carries it

Three files, and none of them moves a published surface. ⛔ No new exported symbol, ⛔ no schema narrowing or widening, ⛔ no change to a published payload, and the `SchemaNode` declaration is untouched (objectui#7105 ruled node slots relax the renderer). The hard stop attached to the `no` was not reached.

| file | what changed |
|---|---|
| `packages/components/src/renderers/navigation/header-bar.tsx` | one JSX guard: `and-and` chain → ternary yielding `null`, plus the comment recording the trap and the truthiness choice |
| `packages/components/src/__tests__/header-bar-right-content-numeric-falsy.test.tsx` | new pin (14 rows: 4 harness controls, 3 falsy-number rows, 4 live controls, 2 non-discriminating rows, 1 mechanism row) |
| `.changeset/9033-header-bar-right-content-numeric-falsy.md` | `@object-ui/components: patch` |

No `needs:contract-review` label, per the `no` pair.

## Acceptance notes

- `{ type: 'header-bar', title: 'H', rightContent: 0 }` renders the chrome and no stray `"0"`, through the real renderer stack inside the `SidebarProvider` the renderer requires. Same for `-0` and `NaN`.
- Live controls asserted in the same run and unchanged: `42` renders `42`, `'txt'` renders `txt`, `undefined` renders the baseline chrome only, and an authored node still renders — so the repair did not narrow the slot to primitives-only.
- A validator row asserts the leaked value is **authorable**: the published zod face parses `rightContent: 0` successfully. Without it every other row is about a value nobody can reach.
- Out of scope, noted, not filed: the census's 53 `any` / `unknown` operands are a standing blind spot for any type-flag instrument in this tree, and three of them turned out to be real leaks. That is a property of the instrument, not a defect, and it is recorded inside objectui#9162 rather than as its own card.
- Out of scope, noted, not filed: `renderChildren`'s first leg already answers `null` for every falsy input, so at several census sites the `and-and` guard in front of it is doing nothing except creating the trap. Whether the guard should simply be deleted at those sites is part of the class-level question objectui#9162 raises; no opinion is executed here.

---

### DIFF
--- a/.changeset/9033-header-bar-right-content-numeric-falsy.md
+++ b/.changeset/9033-header-bar-right-content-numeric-falsy.md
@@ -1,0 +1,39 @@
+---
+'@object-ui/components': patch
+---
+
+`ui:header-bar` no longer paints a stray `"0"` beside the header chrome when
+`rightContent` is authored as a falsy number (objectui#9033).
+
+`HeaderBarSchema.rightContent` is declared `SchemaNode` on both published faces — the
+interface in `@object-ui/types`, and its Zod mirror, whose node union carries a `z.number()`
+arm — so `0` is an authorable value for the slot. The renderer guarded it with an `&&` chain
+whose left operand is the raw slot value, and with `rightContent: 0` that chain evaluates to
+the **number** `0`, which React renders. Measured: `{ type: 'header-bar', title: 'H',
+rightContent: 0 }` painted `"Toggle Sidebar0"`.
+
+This is the numeric-falsy JSX trap objectui#8331 measured and closed one slot over on
+`DataTableSchema.emptyAction`, and the fix is that same shape rather than a second one: the
+`&&` chain becomes a ternary yielding `null`. The **truthiness** leg stays rather than
+becoming nullish. Both land on the same rendering today — since objectui#8908
+`toRenderableSchema` maps a falsy primitive onto nothing rather than onto its `String` form
+— so the choice is about which rule the slot states, and truthiness is the rule that keeps
+this slot's answer independent of the bridge. That independence is exactly what kept the
+sibling slot out of the defect while the bridge was wrong.
+
+Narrower than it may read, and stated precisely so the repair is not judged by the wrong
+row: only the falsy **numbers** leaked. `0` and `-0` painted the character `0`, `NaN` painted
+the three characters `NaN`. `false` and `''` were already correct — React ignores `false` as
+a child — and cannot discriminate the two worlds.
+
+Adjacent but **not** this defect: objectui#8908 repaired the bridge so a falsy primitive
+renders nothing. That repair does not reach this line and never could, because `&&`
+short-circuits and the chain produces the raw `0` before any bridge is called. Same symptom,
+different mechanism.
+
+The declaration is untouched — objectui#7105 ruled node slots relax the renderer rather than
+narrow the declaration, and a `typeof === 'object'` test here would silently drop a bare
+string the slot renders as its own text.
+
+Scored `patch`: no new capability, a slot that painted a character it was never asked to
+paint stops painting it.
--- a/packages/components/src/__tests__/header-bar-right-content-numeric-falsy.test.tsx
+++ b/packages/components/src/__tests__/header-bar-right-content-numeric-falsy.test.tsx
@@ -1,0 +1,195 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `ui:header-bar`'s `rightContent` slot no longer leaks a stray `"0"`
+ * (objectui#9033) — the numeric-falsy JSX trap objectui#8331 measured and
+ * closed one slot over on `DataTableSchema.emptyAction`.
+ *
+ * ## The mechanism, and why it is not objectui#8908
+ *
+ * `HeaderBarSchema.rightContent` is declared `SchemaNode` on BOTH published
+ * faces — the interface in `@object-ui/types`, and its zod mirror, whose node
+ * union carries a `z.number()` arm — so `0` is an authorable value for it.
+ * `theValidatorAcceptsZero` below asserts that from the published validator
+ * rather than asserting it in prose, because every row here is worthless if
+ * the value is unreachable.
+ *
+ * The renderer guarded the slot with an `&&` chain whose left operand is the
+ * raw slot value. With `rightContent: 0` that chain evaluates to the NUMBER
+ * `0`, and React renders numbers — so the header painted a stray `"0"`.
+ *
+ * objectui#8908 repaired `toRenderableSchema` so a falsy primitive becomes
+ * nothing rather than its `String` form. That repair does ⛔ NOT reach this
+ * line and never could: `&&` short-circuits, so the bridge is not called at
+ * all for a falsy value — the chain has already produced the raw `0` before
+ * any bridge runs. Same symptom from the user's seat, different mechanism,
+ * different fix. `theBridgeIsNotWhatFixesThis` pins that distinction directly
+ * so a future reader cannot conclude the bridge covers this slot.
+ *
+ * ## Why `false` and `''` are present but are ⛔ NOT evidence
+ *
+ * React ignores `false` as a child, and `''` contributes no characters. Both
+ * rows were GREEN before the repair and are green after it, so neither can
+ * discriminate the two worlds. They sit in `rowsThatCannotDiscriminate` under
+ * that name, kept only so a later reader does not "extend coverage" by adding
+ * them as if they proved something.
+ *
+ * The rows that DO discriminate are the falsy NUMBERS, and only they:
+ * `0` and `-0` (both of which React prints as the single character `0`) and
+ * `NaN` (which prints the three characters `NaN`).
+ *
+ * ## Why the assertions read whole text rather than query for a node
+ *
+ * The leak is a bare text node with no element of its own — there is nothing
+ * to `querySelector`. The instrument is therefore the header's entire
+ * `textContent`, compared against the baseline the same harness produces for
+ * `rightContent: undefined`. `theInstrumentSeesSlotText` is the lit control on
+ * that instrument: with `rightContent: 42` the text MUST grow by `42`, so a
+ * "no stray characters" row cannot pass because the instrument went blind.
+ *
+ * ## The `SidebarProvider` host
+ *
+ * `header-bar` renders `SidebarTrigger`, which calls `useSidebar()` and THROWS
+ * without a provider. A caught throw renders markup that reads as a clean pass
+ * for every row below, so `theHarnessRendersTheRealHeader` asserts the real
+ * header is mounted rather than assuming it.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, cleanup } from '@testing-library/react';
+import { ComponentRegistry } from '@object-ui/core';
+// Registered at module scope, NOT in a `beforeAll`: there the cold transform is
+// billed to `hookTimeout`, which is narrower than the timeout it replaces
+// (objectui#3010 / #3021, and `object-ui/no-dynamic-import-in-test-hook`).
+import '../renderers';
+import { SidebarProvider } from '../ui';
+import { toRenderableSchema } from '@object-ui/react';
+import { HeaderBarSchema as HeaderBarSchemaZod } from '@object-ui/types/zod';
+
+afterEach(() => cleanup());
+
+/** The header's own chrome text, contributed by `SidebarTrigger`'s sr-only label. */
+const CHROME = 'Toggle Sidebar';
+
+function renderRightContent(rightContent: unknown) {
+  const C = ComponentRegistry.get('header-bar') as React.ComponentType<any>;
+  const { container } = render(
+    <SidebarProvider>
+      <C schema={{ type: 'header-bar', title: 'H', rightContent }} />
+    </SidebarProvider>,
+  );
+  const header = container.querySelector('header');
+  if (!header) throw new Error('no <header> — the harness did not mount the real renderer');
+  return header.textContent ?? '';
+}
+
+/** `rightContent` omitted entirely: the chrome, and nothing else. */
+function renderBaseline() {
+  const C = ComponentRegistry.get('header-bar') as React.ComponentType<any>;
+  const { container } = render(
+    <SidebarProvider>
+      <C schema={{ type: 'header-bar', title: 'H' }} />
+    </SidebarProvider>,
+  );
+  const header = container.querySelector('header');
+  if (!header) throw new Error('no <header> — the harness did not mount the real renderer');
+  return header.textContent ?? '';
+}
+
+describe('ui:header-bar rightContent numeric-falsy leak (objectui#9033)', () => {
+  describe('harness controls', () => {
+    it('theHarnessRendersTheRealHeader — a real <header>, not error-boundary markup', () => {
+      // `useSidebar()` throws without `SidebarProvider`, and the caught throw is
+      // clean markup that would make every "no stray characters" row below pass
+      // for entirely the wrong reason.
+      expect(renderBaseline()).toContain(CHROME);
+    });
+
+    it('theBaselineIsTheChromeAlone', () => {
+      expect(renderBaseline()).toBe(CHROME);
+    });
+
+    it('theInstrumentSeesSlotText — LIT CONTROL: a truthy number DOES reach the text', () => {
+      // Green in both worlds BY DESIGN. Without it, a slot that rendered
+      // nothing at all — or a `textContent` read that stopped working — would
+      // make every leak row below green while measuring nothing.
+      expect(renderRightContent(42)).toBe(`${CHROME}42`);
+    });
+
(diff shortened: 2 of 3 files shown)


os-tesla:
## 派单席评审 —— ⭐⭐ **ACCEPT**。本轮最好的一次交付，而最好的部分**不在这个 diff 里**

### 修复本身：一行，照裁决的形状，没有发明第二种

`&&` 链换成三元式产出 `null`，docblock 引了 objectui#8331 并写明「⛔ not a second shape for the same trap」。三个文件，全在 `@object-ui/components` 内，⛔ 没有顺手widen。

### ⭐⭐ 但它把卡的核心断言**证伪**了，而证伪的方式比结论更值钱

卡说 `header-bar` 是「**the only** remaining instance」，依据是 `git grep -n '&& .*toRenderableSchema'` 恰好一行。

⇒ **那条 grep 约束的是右操作数的拼写，而这个陷阱只取决于左操作数。** 一个 `SchemaNode` 槽只要用裸 `&&` 守着就有它，右边写什么都一样。

⇒ 把拼写约束去掉重跑：**另外十一个站点在漏**，全是 `@object-ui/components` 的已发布渲染器。已另立 **objectui#9162**，⛔ 没有折进本 PR ——那正是我的停止条件点名的结局。

### ⭐⭐⭐ 而这一条，是本会话到目前为止**「没有起火的控制项就不是读数」这条规则最漂亮的一次兑现**

普查的第一遍只读 render container 的文本。⇒ **Radix 的浮层 portal 到容器外面**，于是 `dialog` / `sheet` / `drawer` 三行读回「clean」——**而三个都在漏**。

抓住它的是**每一行自带的起火控制项**：同一个槽喂 `42`，它必须出现在文本里，否则这一行报 **NOT MEASURED** 而不是 clean。⇒ 一个会把「什么都没测到」和「测到了零」分开的控制项，当场拆穿了一把瞎掉的仪器。⛔ 没有它，这张卡会带着「三个浮层是干净的」这个结论落地。

### ⭐ 还有一条读数，它防的是未来的误引用

`card` 的 `children` 是唯一一行 clean，而**原因不是守卫**：那里写的是 `(schema.children || schema.body)`，而 `0 || undefined` 求值为 `undefined`。⇒ 是**操作数顺序的巧合**，不是保护——在同一个渲染器上写 `body: 0`（`children` 缺席）就漏，那正是它上面一行。这条被写进 objectui#9162，⇒ 以后没人能拿这一行当「card 是安全的」的证据。

### 普查的总体、控制项与**不确定桶**都写出来了

- 总体：`git ls-files` 枚举的 816 个非测试 `.tsx`，2396 个 JSX `&&` 守卫，2396 个左操作数；
- 报告：66 个类型允许 `number` 的操作数，其中 **14 个**是 `SchemaNode` 类型；
- **控制项起火**：`header-bar` 本身（已知会漏的那个）；
- ⭐ **53 个 `any` / `unknown` 操作数被列为「不确定」而不是算进 clean** —— 而且证明了这个桶**不是空的**：`drawer` 的 `footer` 就在桶里，运行时探针实测它在漏。⇒ 一把基于类型标志的仪器对 `any` 结构性失明，这件事被它自己说了出来。

### ⚠️ 一处需要更正的措辞，⛔ 别让它扩散

报告里写「objectui 的 `needs:contract-review` label is read by nothing」。⇒ **不成立**：`check-clause2-carriers` 读它——本轮 objectui#9122 的 exit 4 正是靠它的**移除时间戳**判出「gate 曾被合法清除、而 head 之后移动了」。

⚠️ 但**本 PR 不挂那个 label 是对的**：`no` 对的载体是 PR 正文里的改动文件清单，⛔ 不是 label。结论对，理由要换。执行席自己用**正确的脚本**跑了 `check-clause2-carriers --pair 9163`（我这一轮修正过的那个）⇒ **exit 0，both carriers agree**。

### 那个 open sub-question，答得也对

真值守卫 vs 空值守卫：选真值，理由是**两者今天落在同一个渲染结果上，而真值那条让这个槽的答案不依赖 bridge**。⇒ 一个在「两条路今天等价」时按**独立性**而不是按口味来选的理由。

### ⭐ 留给 objectui#9162 的那个问题，我点名支持它被当成类级问题处理

> 「Is a per-site ternary the right remedy at this count, or does it want one guard? 十一个站点一个一个修就是同一个编辑做十一遍，下个月加的第十二个槽照样会错。」

⇒ 这不是「再派十一轮」的活。⛔ 本席不裁（objectui#9162 未分级，路由是分诊的），但把这句提到评审上，免得它被读成「十一张小卡」。

⇒ flip ready + SQUASH 入队。⚠️ 本卡是 `no` 对，⛔ 无载体可摘。

---

---

PR #9168, opened 2026-09-11
fix(plugin-detail): the summary chip reads the declared percent source instead of a drifted copy

### LINKED ISSUE #9071: bug(plugin-detail): the summary chip's percent scaling is a drifted copy — it disagrees with `percentDisplayValue` at exactly 1 and at values at or below -1
Found by the census objectui#8728 asked for (is the chip's `num <= 1 ? num * 100 : num` the repo's interpretation of a stored `percent`, or just this file's?). It is just this file's.

## Measured

Census of every site that turns a stored `percent` into a display magnitude, on `7f27bc543`:

| site | rule | a stored `1` | a stored `-5` |
|---|---|---|---|
| `packages/core/src/utils/dataset-format.ts` — `percentDisplayValue`, the declared single source of truth | `value > -1 && value < 1` | `1%` | `-5%` |
| `packages/fields/src/index.tsx` — `formatPercent`, so `PercentCellRenderer` / the list cell | via `percentDisplayValue` | `1%` | `-5%` |
| `packages/plugin-dashboard/src/recordFields.tsx` | via `percentDisplayValue` since objectui#5607 | `1%` | `-5%` |
| `packages/plugin-grid/src/useColumnSummary.ts` | inline copy, same boundary | `1%` | `-5%` |
| `packages/plugin-detail/src/DetailView.tsx` — the `summaryFields` chip | `num <= 1 ? num * 100 : num` | `100%` | `-500%` |

The chip is the only `num <= 1` spelling in the repo. `grep` over `packages/*/src` and `apps/*/src` returns exactly one hit.

## Why this is a contract violation, not a boundary nit

`percentDisplayValue`'s own doc comment states the contract in those words — quoting it, because the quote is the judgement:

> This is the SINGLE source of truth for percent display scaling, shared by the list-view percent cell renderer (`formatPercent` in `@object-ui/fields`) and the dataset measure formatter (`formatMeasure`) so a percent renders identically as a row value and as an aggregated metric — the two surfaces can never drift.

> If a third surface ever needs percent display, it takes BOTH halves from here — the scaling AND the convention — or this promise breaks again in the same place.

The summary chip is a third surface that needs percent display and takes neither half.

objectui#5607 already removed this exact drift shape one package over: `plugin-dashboard`'s local `value > 1` branch. Its pin test states the decision in words — exactly 1 is percentage points and renders `1%`, which is the convention `PercentScale` spells out as `whole` is `1` shows `1%`. The chip's `num <= 1` puts exactly 1 on the fraction side and draws `100%`, which is the same drift that card removed.

## The third convention — the one the WRITER of the value used

`packages/fields/src/widgets/PercentField.tsx` does not infer from magnitude at all. It reads the field's declared `max`: a field declaring `max` above 1 (e.g. `max: 100`) stores whole-number percents, everything else stores fractions. So the authoring surface already answers the question the two read-side heuristics are guessing at, and for a field declaring `max: 100` and storing `0.5` the edit widget shows `0.50%` while the chip's bar draws 50%.

## Scope note — what objectui#8728 did and did not do

Triage's ruling there was: make the chip's TEXT follow the BAR's existing rule, do not change the bar, do not invent a third rule. That card did exactly that — one named helper, both halves reading it, every bar answer preserved byte-for-byte including `1` drawing a full bar. Two consequences were recorded there rather than decided:

- exactly `1` stays `100%` on this chip while every other band renders it `1%`;
- the chip's TEXT for a stored value at or below -1 moved with the bar's rule, so a stored `-5` now reads `-500%` where it read `-5%` before. The bar is unchanged for those inputs (any negative clamps to an empty track either way), so this is the drift becoming visible on the half that had been accidentally right.

## What this card decides

Which authority the chip reads. Three routes, and the choice is a product decision, not a refactor:

- **A** — `percentDisplayValue` plus the shared percent convention, i.e. both halves as the doc comment requires. Note `@object-ui/plugin-detail` already peer-depends on `@object-ui/core`, so this adds no dependency edge. Taking only the scaling half would be a fourth dialect, and taking the convention half changes what the chip prints for values it renders correctly today (`formatPercent` rounds `12.3` to `12%` at the default precision), so A is a real decision with a visible diff.
- **B** — read the field's declared `max`, the way the edit widget does, and infer only when it is absent.
- **C** — carry the scale on the field in `@objectstack/spec` and stop inferring anywhere.

## Dedup

Run. Channel: repo-scoped REST list — `/repos/objectstack-ai/objectui/issues?state=open&labels=domain:ui&per_page=100`, followed to the end of the label (2 pages, 101 open issues) — plus a local title-and-body keyword scan for percent and scaling. Control fired: the same scan returned objectui#8728 itself, so the single hit is a reading and not a silent zero. That one hit was objectui#8728. Scope boundary, stated rather than papered over: open issues carrying `domain:ui`; an unlabelled card would not have been seen.

Filed by the `os-dev` seat while implementing objectui#8728, from session `session_01MPaVWWMuWeT5LgB1qoXjVB`. Not fixed on that card's PR, by its ruling.

os-tesla:
Fixes #9071

The `summaryFields` chip beside the record H1 scaled a stored `percent` by a rule of its own. The local predicate is **deleted**, not realigned — realigning it would leave two rules agreeing by coincidence, which triage named as the thing to avoid and which is the shape that produced the drift in the first place.

Executable change: **5 lines**, all in one function.

```
-  if (raw > 1) return raw;
+  const points = percentDisplayValue(raw);
+
+  if (points === raw) return points;
-  return Number((raw * 100).toPrecision(12));
+  return Number(points.toPrecision(12));
```

## Changed files (Clause-② is `no`, which owes this listing)

| file | what moved |
|---|---|
| `packages/plugin-detail/src/summaryChipPercent.ts` | the repair: the local predicate deleted, `percentDisplayValue` imported from the published `@object-ui/core` specifier |
| `packages/plugin-detail/src/DetailView.tsx` | **comment only** — no executable line changed; the call-site note now names the authority and points at the half not taken |
| `packages/plugin-detail/src/__tests__/summaryChip.percentSource-9071.test.tsx` | **new** — the two-places pin |
| `packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx` | two boundary rows and the prose that handed this fork here |
| `.changeset/9071-summary-chip-percent-source.md` | **new** — `patch` on `@object-ui/plugin-detail` |

⛔ No new exported symbol, no schema change, and `packages/core/src/utils/dataset-format.ts` is **imported, never modified** — the same-file fence around in-flight objectui#9160 holds. Nothing outside `packages/plugin-detail` is touched.

## The census, re-derived on this branch's base rather than inherited

| site | rule | a stored `1` | a stored `-5` |
|---|---|---|---|
| `percentDisplayValue` in core — the declared source | `value > -1 && value < 1` | `1%` | `-5%` |
| `formatPercent` / `PercentCellRenderer`, the list cell | via `percentDisplayValue` | `1%` | `-5%` |
| `PercentCellRenderer`'s BAR value | inline copy, **same predicate** | `1%` | `-5%` |
| `useColumnSummary` in `plugin-grid` | inline copy, **same predicate** | `1%` | `-5%` |
| `recordFields` in `plugin-dashboard` | via `percentDisplayValue` | `1%` | `-5%` |
| `formatMeasure` in core | declared `percentScale` first, else `percentDisplayValue` | `1%` | `-5%` |
| **the summary chip** — before this PR | pass-through above 1 | **`100%`** | **`-500%`** |

⇒ disagreement set re-derived by hand and then confirmed by a failing run: **exactly `1`**, and **every value at or below `-1`**. The two inline copies are copies of the rule but **not drifted** — same predicate, same answers — so fixing the chip leaves no disagreement at either of them.

### ⚠️ Mechanism assumption 2 was FALSIFIED, and the finding is filed not widened

"Every other site routes through the declared source" is **not true today**. `formatMetricValue` in `packages/plugin-dashboard/src/MetricWidget.tsx` writes the same rule the other way round — a pass-through guarded by a greater-than-one test with the multiply in the else arm — so a literal `num <= 1` grep cannot see it, which is why objectui#9071's own census missed it. Measured, both surfaces in one run, a metric tile with `format: '0%'` against `formatPercent`:

| stored | MetricWidget | the declared source |
|---|---|---|
| `1` | `100%` | `1%` |
| `-1` | `-100%` | `-1%` |
| `-5` | `-500%` | `-5%` |
| `0.25` | `25%` | `25%` |

Identical disagreement set. It is **outside `plugin-detail`**, which this card's dispatch made a stop condition, so it is named and filed as objectui#9165 rather than pulled in here. Not addressed on this PR.

## The pin renders BOTH places, because a chip-only pin cannot fail for this card's reason

`summaryChip.percentSource-9071.test.tsx` renders the DetailView chip **and** `PercentCellRenderer` from the same stored value and the same field, then asserts they agree — on the stated text, and on the magnitude each one draws. Two instruments, because neither alone covers the disagreement set: bars clamp any negative to an empty track, so only the text reaches a value at or below `-1`.

- **Live control** `0.25`, inside the band the two rules always agreed on: identical in both places, green before the change and after, and green on **both** legs of the ablation.
- Both boundaries pinned: exactly `1`, and exactly `-1` as well as `-5`.

Run on the unrepaired tree first, and it failed for exactly the card's reason: `expected '100%' to be '1%'`, `expected '-100%' to be '-1%'`, `expected '-500%' to be '-5%'`, and the whole-table relation red with `[ 1, -1, -5 ]`.

## Ablation — the pins can fail, proved by state and never by an exit code

Direction predicted in writing before the run: reinstating the deleted predicate turns **both** pin files red at the boundary rows, and leaves the control and the already-agreeing rows green.

**Mutation proved on disk BEFORE the run.** HEAD blob `bd35a701aeb9a64bd8de9809defa27a590915741`, on disk before `bd35a701ae…` (equal, so the tree really was at HEAD). Occurrence counts both directions: repaired line 1 then 0, drifted line 0 then 1. Blob hash moved to `7cfbd5bc4e1180ac3f43eedae016ed66488ca598`. The mutated line printed from disk: `const points = raw > 1 ? raw : raw * 100;`. No build step is owed — the root vitest config aliases `@object-ui/core` and `@object-ui/fields` to their `src` trees, so the run reads the mutated source directly.

**Mutated run: `6 failed | 11 passed`** — the 8728 pin red at `1` and `-5`, the 9071 pin red at `1`, `-1`, `-5` plus the whole-table relation. The control `0.25` and the already-agreeing `12.3` case stayed green.

**Restore proved BY STATE.** `git hash-object` back to `bd35a701ae…` and equal to the HEAD blob; `git diff HEAD` empty; `git status --short` empty; the restored line printed from disk; occurrence counts back to 1 and 0. **Restored run: `17 passed`.** The mutation ran under a `trap … EXIT INT TERM` with absolute paths, and the restore points at `HEAD` explicitly rather than at the index.

## Verification, all on the final commit `7c620b228`

| what | result |
|---|---|
| `pnpm exec vitest run packages/plugin-detail/` (via the shared verify lock) | `VERDICT command-exit 0` — **170 files, 1581 tests passed** |
| `pnpm --filter @object-ui/plugin-detail type-check` | exit 0, after building the dependency closure (`pnpm --filter '@object-ui/plugin-detail^...' build`, `VERDICT command-exit 0`) — the script echoed `tsc --noEmit && tsc -p tsconfig.test.json`, so the new test file is covered too |
| eslint, **whole repo, not narrowed** | 4,827 files as eslint's own config selects them; 95 errors and 12,865 warnings, all pre-existing. My 4 linted files carry **0 errors and 0 warnings**; `DetailView.tsx`'s 45 warnings are unchanged in count, and the repo warning total is 2 **lower** than before this branch |
| `check:new-line-citations`, `check:control-bytes`, `check:changeset-claims`, `check:vi-mock-specifiers`, `check:unreferenced-sources` | exit 0 each |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 3 published source files, 1 changeset |
| `node scripts/check-governed-queue-guard.mjs --test` on all 5 paths | `NOT GOVERNED — none matched` |

⚠️ First `type-check` attempt reported `TS2307: Cannot find module '@object-ui/core'` against my new import **and against every pre-existing one** — an unbuilt workspace, not a finding. Recorded because it reads like a real error. The reading above is from after the closure build.

## Acceptance notes

**What deliberately moved, and nothing else did.** A stored `1` now reads `1%` beside a 1% bar where it read `100%` beside a full bar; a stored value at or below `-1` is passed through, so `-5` reads `-5%` where it read `-500%` (the bar is unchanged there — any negative clamps to an empty track either way). Everything in the band the two rules always agreed on is byte-identical, including the float-residue trim that keeps `0.07` reading `7%` rather than `7.000000000000001%`.

**Why a trim survived at all, and why it is not a second rule.** The residue trim had to stay, or a value that renders correctly today would move — `percentDisplayValue(0.07)` is `7.000000000000001`, which `formatPercent` absorbs at its precision and this chip's bare `${points}%` cannot. It is guarded by `points === raw`, which asks the SOURCE whether it scaled by comparing its own answer to its own input, so it cannot drift away from whatever `percentDisplayValue` decides, and it never touches a value the source passed through. The scaling **predicate** is gone; its name survives only inside the comment that explains its removal.

**The half this PR does not take, filed rather than left silent.** `percentDisplayValue`'s doc comment asks a third surface for BOTH halves — the scaling AND the convention. Only the scaling moved, because taking the convention half moves values that render correctly today (`12.3` would become `12%`, `1234.5` would become `1,235%`), which this card's dispatch made a stop condition. That residue is pinned as a named case in the new test file and filed as objectui#9167, since this PR closes the card it would otherwise have lived on.

**Out-of-scope observations, noted and not filed.** The `PercentCellRenderer` bar value and `useColumnSummary` each hold a hand-written copy of the shared predicate; both agree with `percentDisplayValue` byte-for-byte today, so there is no disagreement to report and no defect — only a duplication that a future consolidation would pick up. Successor named: whoever takes objectui#9165, which has to touch this same family. Separately, `formatMetricValue`'s own doc comment describes its percent branch as scaling only **below** 1 while the code scales at or below it — recorded inside objectui#9165 rather than filed on its own.

---

### DIFF
--- a/.changeset/9071-summary-chip-percent-source.md
+++ b/.changeset/9071-summary-chip-percent-source.md
@@ -1,0 +1,27 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+The `summaryFields` chip beside the record H1 reads the repo's percent authority
+instead of a drifted copy of it (objectui#9071).
+
+The chip scaled a stored `percent` by a rule of its own — a pass-through above 1,
+a scale-by-100 at or below it — while `percentDisplayValue` in `@object-ui/core`,
+the declared single source of truth that the list cell (`formatPercent` /
+`PercentCellRenderer`), the dashboard measure (`formatMeasure`) and the grid
+column summary all reach, uses the symmetric `value > -1 && value < 1`. The local
+predicate is deleted, not realigned.
+
+**Two values move, and only these two.** They are exactly where the two rules
+disagreed:
+
+- a stored `1` now reads `1%` beside a bar at 1%, where it read `100%` beside a
+  full bar — one percentage point, the answer every other surface already gave;
+- a stored value at or below `-1` is passed through: a stored `-5` now reads
+  `-5%` where it read `-500%`. The bar is unchanged for these inputs, since any
+  negative clamps to an empty track either way.
+
+Everything inside the band the two rules always agreed on is byte-identical: a
+stored `0.25` still reads `25%`, `0.123` still reads `12.3%`, `12.3` still reads
+`12.3%`, `250` still reads `250%`, and the float-residue trim that keeps `0.07`
+reading `7%` rather than `7.000000000000001%` is unchanged.
--- a/packages/plugin-detail/src/DetailView.tsx
+++ b/packages/plugin-detail/src/DetailView.tsx
@@ -1128,8 +1128,13 @@ export const DetailView: React.FC<DetailViewProps> = ({
                         // text states it, the bar below draws it clamped to
                         // its track. They used to scale `num` by two different
                         // rules, so a stored `0.123` said `0.123%` beside a bar
-                        // at 12.3% (objectui#8728). Which rule survived, and
-                        // why it is not the repo-wide one, is in
+                        // at 12.3% (objectui#8728). That number is now the
+                        // repo's — `percentDisplayValue` in `@object-ui/core`,
+                        // the same authority the list cell and the dashboard
+                        // measure read, so one record cannot state two
+                        // percentages in two places (objectui#9071). Which
+                        // predicate was deleted to get there, and which half of
+                        // that authority this chip still does not take, are in
                         // `./summaryChipPercent`.
                         const points = summaryChipPercentPoints(num);
                         display = `${points}%`;
--- a/packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx
+++ b/packages/plugin-detail/src/__tests__/summaryChip.percentOneRule-8728.test.tsx
@@ -44,13 +44,23 @@
  * ## The boundary rows
  *
  * `EXACTLY 1` is the fork triage named: is a stored `1` one percent or one
- * hundred? This chip answers 100%, because that is what its bar already drew
- * and the ruling was that the text follows the bar. ⚠️ That answer is this
- * chip's alone — `percentDisplayValue` in `@object-ui/core`, and so the list
- * cell, the dashboard measure and the grid column summary, all render a stored
- * `1` as `1%`. The census and the three routes out are objectui#9071; the row
- * below exists so the answer is PINNED rather than incidental, and so that
- * card's PR has to state which way it is moving.
+ * hundred? This chip used to answer 100%, because that is what its bar already
+ * drew and the ruling here was that the text follows the bar — an answer that
+ * was this chip's alone, while `percentDisplayValue` in `@object-ui/core`, and
+ * so the list cell, the dashboard measure and the grid column summary, all
+ * rendered a stored `1` as `1%`.
+ *
+ * ⭐ objectui#9071 MOVED IT, and this is the row that recorded which way. That
+ * card took the census this one asked for, found the local predicate to be the
+ * only spelling of its kind on this side of the tree, and deleted it in favour
+ * of the declared source. A stored `1` now reads `1%` here as it always did
+ * everywhere else, and a stored `-5` reads `-5%` rather than `-500%`. The rows
+ * below carry the new answers; what did NOT move — every value inside the band
+ * the two rules always agreed on, the control among them — is the rest of this
+ * table, unchanged. The two-places pin is
+ * `summaryChip.percentSource-9071.test.tsx`; this file keeps stating the
+ * relation objectui#8728 is about, that the chip's two halves never disagree
+ * with EACH OTHER.
  *
  * ## Instruments
  *
@@ -144,9 +154,9 @@ const ROWS: Row[] = [
   // CONTROL. Already percentage points, so nothing about it may move.
   { what: 'CONTROL — a value already in points is untouched', stored: 12.3, text: '12.3%' },
   { what: 'zero', stored: 0, text: '0%' },
-  // THE FORK. See the header: this chip's answer, pinned, and objectui#9071's
-  // to move.
-  { what: 'EXACTLY 1 — this chip reads it as a ratio', stored: 1, text: '100%' },
+  // THE FORK, after objectui#9071 moved it. See the header: one percentage
+  // point, the answer every other band already gave.
+  { what: 'EXACTLY 1 — one percentage point, as everywhere else', stored: 1, text: '1%' },
   { what: 'just above 1 — the other side of the same boundary', stored: 1.5, text: '1.5%' },
   // Agreement has to survive the clamp: the text keeps the real number, the
   // bar saturates. A row that only ever tested unclamped values would let a
@@ -155,11 +165,13 @@ const ROWS: Row[] = [
   // The residue row. `0.07 * 100` is `7.000000000000001` in binary floating
   // point — invisible as a CSS width, unreadable as a label.
   { what: 'a ratio whose scaling carries float residue', stored: 0.07, text: '7%' },
-  // The drift, now visible on both halves instead of one. Before the fix the
-  // text read `-5%` here while the bar drew an empty track; the bar's rule
-  // scales it, and the ruling was that the text follows the bar. Recorded on
-  // objectui#9071 with the census that names it.
-  { what: 'a negative at or below -1 — moves with the bar\'s rule', stored: -5, text: '-500%' },
+  // The other end of the same boundary. objectui#8728 moved this text to
+  // `-500%` because the bar's rule scaled it and the ruling was that the text
+  // follows the bar; objectui#9071 then replaced that rule with the declared
+  // source, which passes a value at or below -1 straight through. The bar is
+  // unchanged across both cards — any negative clamps to an empty track — so
+  // this row only ever moved on the half that reads the number.
+  { what: 'a negative at or below -1 — passed through by the shared rule', stored: -5, text: '-5%' },
 ];
 
 describe('summary chip percent — one stored number, one percentage (objectui#8728)', () => {
--- a/packages/plugin-detail/src/__tests__/summaryChip.percentSource-9071.test.tsx
+++ b/packages/plugin-detail/src/__tests__/summaryChip.percentSource-9071.test.tsx
@@ -1,0 +1,220 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * ONE record, TWO places, ONE number (objectui#9071).
+ *
+ * ## What this pins, and why it is not "the chip's new number is right"
+ *
+ * The chip beside the record H1 scaled a stored `percent` by a rule of its
+ * own — a pass-through above 1, a scale-by-100 at or below it — while the
+ * declared single source of truth (`percentDisplayValue` in
+ * `@object-ui/core`, which the list cell reaches through `formatPercent`)
+ * uses the symmetric `value > -1 && value < 1`. The two rules disagree at
+ * EXACTLY 1 and at every value at or below -1.
+ *
+ * Triage's judgement is the acceptance, and it is about DISAGREEMENT, not
+ * about the chip:
+ *
+ *   "a stored `1` renders `100%` in the chip and `1%` everywhere else" — the
+ *   same record showing two different numbers in two places is the kind of
+ *   disagreement a user reports as a data bug rather than a formatting one.
+ *
+ * So every case here renders BOTH surfaces from the SAME stored value and the
+ * SAME field, and asserts they agree. A pin that only checked the chip could
+ * not fail for the reason this card exists: it would go green on a chip that
+ * had simply moved to a fourth rule of its own.
+ *
+ * ## Two instruments, because neither alone covers the disagreement set
+ *
+ * - **The stated text.** The only instrument that reaches a value at or below
+ *   -1: both bars clamp a negative to an empty track, so the bar cannot tell
+ *   `-5%` from `-500%`. ⚠️ Valid only where the two surfaces' CONVENTIONS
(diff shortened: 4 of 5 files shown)


os-tesla:
**PASS.** 全部在源头复核，⛔ 不看自述。

## 可执行改动 —— 打印出来，不是数出来

`summaryChipPercent.ts` 的非注释 diff，逐行：

```
+import { percentDisplayValue } from '@object-ui/core';
-  if (raw > 1) return raw;
+  const points = percentDisplayValue(raw);
+  if (points === raw) return points;
-  return Number((raw * 100).toPrecision(12));
+  return Number(points.toPrecision(12));
```

⇒ 本地谓词是**删除**，不是对齐边界。这正是裁决区要的那一条，也是 `humanize-label` 家族反复付学费的那个形状（两条规则靠巧合一致）。

## 三条断言，我自己在源头验的

| 你的断言 | 我的读数 |
|---|---|
| 改动 5 个文件、`DetailView.tsx` 只动注释 | ✓ `git diff --stat` 5 个文件；该文件 patch 里所有 `+`/`-` 都在注释块内，`:1139-1141` 三行是上下文 |
| `packages/core/src/utils/dataset-format.ts` 只 import、⛔ 未修改 | ✓ 它不在改动清单里 —— 而 `percentDisplayValue` 恰恰定义在**它的 `:92`**。围栏成立，而且是以最容易破的方式成立的 |
| 「两半读同一个函数」 | ✓ 全树只有**一个**调用点（`DetailView.tsx:1139`），文本与柱子都读同一个 `points`：`display = \`${points}%\`` / `percentValue = Math.max(0, Math.min(100, points))`。⇒ 存 `1` 时两半一起从 100 移到 1，存 `-5` 时柱子被 `Math.max(0, …)` 夹成 0、两侧都不动 —— 与 changeset 和新 pin 逐字对上 |

`percentDisplayValue` 的实现我也读了：`value > -1 && value < 1 ? value * 100 : value`（`dataset-format.ts:92`），与你报的一致。

## ⭐ `points === raw` 这道守卫，我单独判了一遍

它是本 PR 唯一的新逻辑，值得单独说：**它是精确的，不是近似的。**
`percentDisplayValue(x) === x` 当且仅当源头**放行**了 x，或 `x === 0`（`0 * 100 === 0`）。而 0 的 `toPrecision(12)` 是 no-op ⇒ trim 恰好落在源头缩放过的那一段上，⛔ 没有「缩放了却跳过 trim」的输入。`NaN` 走 else 臂（`NaN === NaN` 为 false），结果仍是 `NaN`，与改动前同。
⇒ 你把它写成「问源头它有没有缩放」而不是「再写一次边界」，是对的：它不能独立于 `percentDisplayValue` 漂移。

## Clause-② `no` —— 成立，且我给零配了点亮的控制项

⚠️ `git grep percentDisplayValue -- packages/core/src/index.ts` 在 parent 上返回**空**。那不是读数。同一条命令对该文件读到 153 行（控制项点亮），真正的链路是 `packages/core/src/index.ts:111` 的 `export * from './utils/dataset-format.js'` —— **在 parent 上就已经存在**。
⇒ ⛔ 无新增已发布导出；`@object-ui/core` 本就是 plugin-detail 的 `peerDependency`（parent 的 package.json 里就有）⇒ ⛔ 无新依赖边、⛔ 无 package.json 改动。
`PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9168` → **exit 0**，两侧载体一致、无 widening tell。声明未被改，符合派单约定。

## ⭐ 被改的那条 #8728 pin 行，是合法移动

改一条既有 pin 的期望值通常要挨审。这次不用：#8728 自己的头注写着「the row below exists so the answer is PINNED rather than incidental, **and so that card's PR has to state which way it is moving**」—— 它是**主动把这个 fork 交给 #9071 的**。你把它移了，并且在两个文件里都写清了朝哪边移、为什么。这是该有的样子。

## ⭐⭐ 机制假设 2 被证伪，而你的处置是对的

`MetricWidget.tsx` 里第二份漂移副本，写成 `> 1` 放行式，所以对 `num <= 1` 的**字面** grep 结构性失明 —— 这是本轮第二次「census 证伪的是拼写，不是陷阱」（上一次是 #9163 的 `&&` 链）。你没有把它卷进本 PR（派单把 plugin-detail 之外定为停止条件），而是带着两面同跑一次的实测表另开了 **#9165**。⛔ 正确。

`#9167`（只取了 scaling、没取 convention）同样：派单的 harm-stated 停止条件禁止动今天渲染正确的值，你把它**钉成了一条命名用例**而不是留成沉默。「the next card inherits a measurement instead of a silence」这句就是本卡最值钱的一行。

## 手动地板？不上

按 sharpened test：changeset 没有声明任何 BREAKING，且它把两个值的移动写成一次**修复的代价**（「Two values move, and only these two. They are exactly where the two rules disagreed」），不是本次改动的**目的**。旧答案也不具备可辩护的读法 —— 它与 `percentDisplayValue` 自己的 doc comment 所声明的契约相悖，而那条契约的裁决权 #8728 已经明确移交给本卡。⇒ 席位自己落。

## 记录，不是拒收理由

- 你自述 `create_pull_request` 追加了第二个署名脚注、以及 `plugin` / `tests` 两个标签是自动化挂的。两条都已核实（PR 现标签 = `plugin`, `tests`，⛔ 无 `needs:contract-review`，`no` pair 正确）。不动。
- 分支基于 `3b6bc6959`，而 `main` 已前移到 `8e74b27fc`（#9160 落地，改的正是 `dataset-format.ts`）。两者改动区域不相交（#9160 动的是 `buildDatasetDrillFilter` 一带），交给合并队列。

⇒ flip ready + SQUASH 入队。落地后按内容验证 + 摘卡片 #9071。

---

PR #9173, opened 2026-09-11
fix(plugin-dashboard): one spelling for a field key across the drill chain

### LINKED ISSUE #9055: finding(plugin-dashboard): two more field-key spellings outside the `table` family — the record drawer's labels and a chart series label disagree with `humanizeFieldKey` on the SAME drill chain
Found while fixing objectui#9000 (the drill `columns` whitelist rendering raw headers). That card's repair made the two header branches of `ObjectDataTable` agree. This card is what the same census turned up **outside** the `table` widget family: two more producers of a field-key display string that do not use `humanizeFieldKey`, one of which is reachable in the very same click path.

objectui#9000's fix does not touch either of them, deliberately — its scope was the one surface it names.

## The class, already ruled on

objectui#5425 ruled that a value cannot appear twice under two spellings on one dashboard, and `humanizeFieldKey`'s own docstring names itself "the single home for this convention". Both instances below are that class, one step across from where #5425 and #9000 each stopped.

## Instance 1 — the record drawer, on the SAME drill chain (this is the one worth acting on)

`RecordDetailDrawer` builds its own field label inline rather than calling `humanizeFieldKey`. Symbol: the `useMemo` in `packages/plugin-dashboard/src/RecordDetailDrawer.tsx` that maps `keys` to rendered rows; the expression reads `key.charAt(0).toUpperCase()` followed by a camelCase split and an underscore replacement — note that it upper-cases only the FIRST word, where `humanizeFieldKey` title-cases every word.

Measured, both spellings evaluated side by side:

```
key              table header      record drawer label   agree
close_date       Close Date        Close date            false
needs_analysis   Needs Analysis    Needs analysis        false
unitPrice        Unit Price        Unit Price            true
amount           Amount            Amount                true
```

Why it matters more than a nit: the drill chain runs table then record. `ObjectDataTable`'s drill list is drill-to-record by default, and the record it opens is this drawer. So one user click moves from a column headed **Close Date** to a field labelled **Close date**, over the same key, in the same widget, in one uninterrupted interaction. camelCase agrees, which is exactly why it went unnoticed — the same coincidence that hid #5425's original defect.

## Instance 2 — the chart series label falls back to the RAW key

`resolveSeriesLabel` in `packages/plugin-dashboard/src/DashboardRenderer.tsx` (the `useCallback` whose docblock is about synthetic y-field keys) ends its object-bound arm with a `fieldLabel` call whose fallback is the y-field key ITSELF, not a humanized form. So with no bundle entry loaded — the ordinary case — a chart grouped on `close_date` legends/tooltips it as `close_date`, while the table widget beside it on the same dashboard heads that column `Close Date`.

This is the exact fallback shape objectui#9000 removed from the table's whitelist branch, surviving on the chart's series label. It was ruled OUT of #9000's diff on purpose: that card fixes the surface it names.

## What a fix would have to decide, and the trap in it

⛔ Not a blind "call `humanizeFieldKey` in both places". Instance 2's fallback is a THREE-way choice, and the middle one is real: a synthetic key (`value` / `count`) is already special-cased above it into an i18n'd aggregate name, so only the genuine-field arm is in question. Instance 1 has no such branch and looks like a straight substitution — but its output is a FIELD label, not a column header, and a repo-wide "one spelling" move should confirm the detail surfaces (`plugin-detail`'s `DetailSection` / `HeaderHighlight`) resolve their labels from the object schema's own `field.label` first, which is a different population from a derived spelling and must not be swept in.

## Dedup

Deduped against the 446 open issues in this repository (repo-scoped REST list, all five pages, PRs filtered out, matched locally). `drillDown|drill-down` as a positive control term returned 11 issues including objectui#9000, objectui#8071 and objectui#5560, so the empty result for `RecordDetailDrawer` + spelling terms and for `resolveSeriesLabel` is a reading rather than a silent zero. Nearest non-duplicates checked and excluded: objectui#7696 (chart option labels vs list view, an i18n coverage gap, not a derived-spelling disagreement) and objectui#4044 (`chartConfig` key liveness).

Filed by an agent seat working objectui#9000, via Claude Code, session reference `https://claude.ai/code/session_01MPaVWWMuWeT5LgB1qoXjVB`. Attribution written as prose on purpose: this repository's body sanitizer deletes the footer block outright when an issue is created.

os-tesla:
Fixes #9055

One field key now renders under one spelling across a dashboard drill chain. Two files, three arms, no published surface moves.

| site | before | after |
|---|---|---|
| `RecordDetailDrawer` field label | a fourth inline spelling (first word only) | `humanizeFieldKey(key)` |
| `resolveSeriesLabel` object-bound arm | `fieldLabel(objectName, yField, yField)` | `fieldLabel(objectName, yField, humanizeFieldKey(yField))` |
| `resolveSeriesLabel` no-object arm | `return yField` | `return humanizeFieldKey(yField)` |
| `resolveSeriesLabel` aggregate arm | i18n aggregate name | **unchanged**, pinned as a live control |

## The triage question, answered

⭐ *"Whoever takes any of them should say whether one helper should own it, rather than fixing a third call site to match the first two."*

**One helper already owns the KEY convention, and this card reconnects the two call sites that were never wired to it** — `humanizeFieldKey` (`packages/plugin-dashboard/src/utils.ts:27`) is the only definition of it in the tree, its own docstring names itself "the single home for this convention", and both sites here are inside that same package. Merging the KEY prefixer with `humanizeLabel`, the VALUE prefixer in `@object-ui/core`, is a different question and is already reserved: `packages/core/src/utils/humanize-label.ts` carries the per-input difference table (`unitPrice` / `BestCase` / `lost-to-competitor` diverge) and rules that "Converging them is a decision, not a refactor … it needs its own card". This PR does not touch that.

## Instance 1, measured as TWO RENDERED STRINGS ON ONE CLICK — not two function returns

The pin renders one `ObjectDataTable`, reads `thead th`, clicks a row, and reads the drawer that click opened. On `origin/main` `8e74b27fc` the run printed both lists side by side:

```
- Expected            (headers, thead th)
+ Received            (drawer labels, dt)

  [
-   "Close Date",
-   "Needs Analysis",
+   "Close date",
+   "Needs analysis",
    "Unit Price",
    "Amount",
  ]
```

`unitPrice` / `amount` are identical under both spellings, so they **cannot discriminate the two worlds**. They are asserted under `KEYS_THAT_CANNOT_DISCRIMINATE` and are recorded, not evidence — green before and after.

## The briefing's four mechanism assumptions, re-measured

**A1 — CONFIRMED.** The drawer's two lines and `ObjectDataTable`'s `buildHeader` (lines 817-820) are byte-for-byte the same shell; only the `humanized` expression differed. Argument order (`fieldLabel(objectName, key, humanized)`) and the `objectName` guard are identical and both are unchanged by this PR.

**A2 — CONFIRMED, and measured on the chain rather than copied.** The table above is the rendered output of one interaction, not a pure-function comparison. The root cause is as assumed: the inline expression upper-cases only the first character, `humanizeFieldKey` title-cases every word via `\b\w`.

**A3 — CONFIRMED, including the arm the card does not name.** Arm 3 IS reachable: `DashboardRenderer.tsx:677` calls `resolveSeriesLabel(undefined, yField, undefined)` on the static-inline-data branch of every series chart, and arm 1 cannot intercept it because that call passes no `aggFn`. Where it renders: `ChartRenderer` turns `series[].label` into `newConfig[s.dataKey] = { label: s.label || s.dataKey, … }`, which is the legend / tooltip text. So arm 3 is a live third spelling of one decision, and it is fixed here. Arm 1 is untouched and is pinned as a live control in two locales.

**A4 — CONFIRMED, with a lit control for the zero.** `grep -c humanizeFieldKey`, per file, before this PR:

```
packages/plugin-dashboard/src/RecordDetailDrawer.tsx         0
packages/plugin-dashboard/src/DashboardRenderer.tsx          0
packages/plugin-dashboard/src/ObjectDataTable.tsx            7   <- lit control, matches the briefing's 7
packages/plugin-dashboard/src/utils.ts                       2
```

The full repo-wide file list from the same command (13 files, none of them the two above) is in the dispatch report. `humanizeFieldKey` has exactly ONE definition in the tree (`packages/plugin-dashboard/src/utils.ts:27`) and is NOT re-exported from `packages/plugin-dashboard/src/index.tsx` — both re-measured, so Clause-② stands at `no`.

## Reverse verification — four legs, each proved ON DISK before it was read

Every leg: mutate → prove the bytes moved (fixed-text count 1→0, old-text count 0→1, blob hash shifted, HEAD blob untouched) → run → restore → prove restoration by STATE (`git diff HEAD` empty **and** blob back to the HEAD blob), never by an exit code. Restores use `git checkout HEAD -- path` (never a bare `git checkout -- path`, which would take the mutation back out of the polluted index). Each leg ran after the fix was committed, so the restore point is a real commit. The harness carries a `trap … EXIT INT TERM` with absolute paths as a crash-path convenience only.

| leg | mutation | predicted red | measured red |
|---|---|---|---|
| A | drawer back to the inline expression | the 4 instance-1 assertions | exactly those 4; all 6 others green |
| B | arm 2 back to `fieldLabel(objectName, yField, yField)` | `arm 2` only | `arm 2` only, 9 green |
| C | arm 3 back to `return yField` | `arm 3` only | `arm 3` only, 9 green |
| D | arm 1's return replaced by a blind `humanizeFieldKey(yField)` | the arm-1 live control | the zh live control, 9 green |

Leg D exists because a control that cannot fail is not a control. ⚠️ It also shows the two arm-1 controls are not equivalent: `humanizeFieldKey('count')` is `Count`, which **coincides** with the English pack, so the `en` control stayed green under leg D and only the `zh` control (`计数`) discriminated it. Recorded rather than tidied away.

Blob hashes from leg A, as an example of the proof shape: `2ead771df…` → `896d9f022…` on mutation, back to `2ead771df…` on restore, with the HEAD blob `2ead771df…` unmoved throughout.

## Verification

Read from each command's own verdict line, with the exit code captured by redirection **before** any pipe.

| check | command | verdict |
|---|---|---|
| new pins | `pnpm exec vitest run packages/plugin-dashboard/src/__tests__/fieldKeySpellingOutsideTable-9055.test.tsx` | `Tests 10 passed (10)`, exit 0 — and `6 failed / 4 passed` on the unmodified tree |
| affected package suite | `pnpm exec vitest run packages/plugin-dashboard/` (via the shared verify lock) | `Test Files 112 passed (112)` · `Tests 1005 passed (1005)`, exit 0 |
| dependency closure build | `pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-dashboard^...' build` | exit 0 |
| type-check (both legs) | `tsc --noEmit` and `tsc -p tsconfig.test.json` | exit 0 each |
| lint (WHOLE repo, not narrowed) | `turbo run lint --concurrency=2` | `Tasks: 47 successful, 47 total`, exit 0 |
| changeset presence | `node scripts/check-changeset-presence.mjs` | exit 0 — 2 published source files, 1 changeset |
| governed surface | `node scripts/check-governed-queue-guard.mjs --test …` | `NOT GOVERNED — 4 path(s) checked against 5 governed surface(s)` |

⚠️ The first type-check run was RED with 12 `Cannot find module '@object-ui/*'` errors. That was an unbuilt dependency closure, not this diff: the same command is exit 0 after the closure build. Recorded because reading that red as a regression is exactly the trap.

Lint is reported as the full repo run, so no narrowing claim is being made. For the record, this config could not have moved an untouched file anyway: `eslint.config.js` extends `tseslint.configs.recommended` (not the type-checked variant) and sets no `parserOptions.project` / `projectService`, so no rule here reads a type graph.

## Acceptance notes — found while working, deliberately NOT in this diff

- **Filed as objectui#9172.** `DashboardGridLayout`, the sibling relay, composes `series: [{ dataKey }]` with **no `label` key at all** on both its branches, and `ChartRenderer` falls back to `s.label || s.dataKey` — so its legend renders the raw field key. Measured with a throwaway recorder probe (deleted in the same turn, never committed): `renderer object-bound label="Close Date"` / `grid object-bound label=undefined`, and the same pair for the static branch. ⚠️ Direction of the debt, stated plainly: before this PR the two relays AGREED (both raw); this PR moves one of them, so objectui#9172 is the other half of this family rather than a pre-existing divergence. It is filed instead of folded in because it is a different file and a different mechanism — repairing it ADDS a label to a surface that never had one, on a relay this card does not name, needing its own pins.
- **noted, not filed — `plugin-detail`'s `DetailSection` / `HeaderHighlight`.** The card asks that a one-spelling move confirm these are a different population, and they are: `DetailSection.tsx:416,451` and `HeaderHighlight.tsx:198` resolve `fieldLabel(objectName, field.name, field.label …)` — the object schema's own declared label, not a derived spelling. Correctly out of scope. One observation, unmeasured: `DetailSection`'s last-resort fallback is `field.label || field.name`, so a field def carrying no `label` would show the raw key; whether that is reachable with a real schema was not measured here because the dispatch rules this surface out of scope. Successor: whoever takes objectui#9172 is already in this family.
- **noted, not filed — other inline capitalizers.** `git grep 'charAt(0).toUpperCase()'` finds further first-letter capitalizers in `app-shell` / `apps/console` (nav segments, approval verbs, icon names). They are not field-key display spellings and do not meet a field on screen, so they are not this class. Successor: none.

## Delivery posture

Clause-② is **`no`**, unchanged from the claim comment and re-measured above: no published export, no schema key, no closed-set member, no registry entry moves. The guard prints `NOT GOVERNED`. Per the dispatch this PR stays a **draft** — not flipped to ready, not enqueued, no auto-merge — for the PM seat to land.

---

### DIFF
--- a/.changeset/field-key-spelling-outside-table-9055.md
+++ b/.changeset/field-key-spelling-outside-table-9055.md
@@ -1,0 +1,11 @@
+---
+'@object-ui/plugin-dashboard': patch
+---
+
+One field key now renders under one spelling across a dashboard drill chain: the record drawer and the chart series label both derive their display string with `humanizeFieldKey`, the single home for the KEY convention (objectui#9055).
+
+**The record drawer, on the same click.** `RecordDetailDrawer` built its field label from a fourth inline spelling that upper-cased only the FIRST word. `ObjectDataTable`'s drill list is drill-to-record by default and the record it opens is this drawer, so one click moved from a column headed `Close Date` to a field labelled `Close date` — same key, same widget, one uninterrupted interaction. Measured on the rendered DOM of one chain, before the fix: headers `Close Date · Needs Analysis · Unit Price · Amount` against drawer labels `Close date · Needs analysis · Unit Price · Amount`. camelCase agreed under both spellings, which is why it went unnoticed.
+
+**The chart series label, on two of its three arms.** `resolveSeriesLabel` in `DashboardRenderer` handed back the RAW field key twice: as `fieldLabel`'s fallback on the object-bound arm, and outright on the arm every static-data chart takes. A chart grouped on `close_date` legended it `close_date` while the table widget beside it headed that column `Close Date`. The static-data arm is not named by the card; it is the same defect reached through the other call site, and it is fixed here rather than left for a follow-up. The synthetic-key arm — a count aggregation with no real field, which resolves an i18n'd aggregate name (Count / 计数) — is a different vocabulary and is deliberately untouched.
+
+The i18n wrapper is unchanged everywhere: a bundle entry still wins and the derived spelling is only its fallback, and the `objectName` guards are as they were. No published export, schema key or registry entry moves. The KEY prefixer stays distinct from `humanizeLabel`, the VALUE prefixer in `@object-ui/core` — converging those two is a separate decision its own docblock already reserves for its own card.
--- a/packages/plugin-dashboard/src/DashboardRenderer.tsx
+++ b/packages/plugin-dashboard/src/DashboardRenderer.tsx
@@ -39,7 +39,7 @@ import {
   verticalListSortingStrategy,
 } from '@dnd-kit/sortable';
 import { CSS } from '@dnd-kit/utilities';
-import { isObjectProvider, deriveStaticTableColumns } from './utils';
+import { isObjectProvider, deriveStaticTableColumns, humanizeFieldKey } from './utils';
 import { classifyWidgetType, METRIC_LIKE_TYPES } from './widgetDispatch';
 import { LEGACY_RETIRED_WIDGET_SCHEMA, isLegacyRetiredWidget } from './legacyRetiredWidget';
 import { DatasetWidget } from './DatasetWidget';
@@ -318,16 +318,39 @@ const DashboardRendererInner = forwardRef<HTMLDivElement, DashboardRendererProps
      * key like 'value' (used by count aggregations that have no real field),
      * fall back to an i18n'd aggregate name (Count / Sum / Average …) instead
      * of leaking the placeholder 'value' string into the legend / tooltip.
+     *
+     * ⭐ THREE ARMS, and only the last two derive a display string from a FIELD
+     * KEY (objectui#9055). The aggregate arm is a different vocabulary and is
+     * deliberately left alone — it is what a blind "humanize everything" here
+     * would eat, and it is pinned as a live control in
+     * `__tests__/fieldKeySpellingOutsideTable-9055.test.tsx`.
+     *
+     * The other two used to hand back the RAW key — `fieldLabel`'s fallback on
+     * the object-bound arm, and the bare `yField` on the arm a static-data
+     * chart takes — so a chart grouped on `close_date` legended it
+     * `close_date` while the table widget beside it on the same dashboard
+     * headed that column `Close Date`. That is the same fallback shape
+     * objectui#9000 removed from the table's whitelist branch, and the class
+     * objectui#5425 ruled out: one value, two spellings, one dashboard.
+     * `humanizeFieldKey` is the single home for the KEY convention
+     * (`./utils`); the i18n wrapper is untouched, a bundle entry still wins and
+     * this is only its fallback.
+     *
+     * It stays distinct from `humanizeLabel`, the VALUE prefixer in
+     * `@object-ui/core` — `utils/humanize-label.ts`'s docblock carries the
+     * per-input difference table and rules that converging the two "is a
+     * decision, not a refactor … it needs its own card".
      */
     const resolveSeriesLabel = useCallback((objectName: string | undefined, yField: string, aggFn: string | undefined) => {
       const isSynthetic = !yField || yField === 'value' || yField === 'count';
       if (aggFn && (isSynthetic || aggFn === 'count')) {
         return t(`report.aggregate.${aggFn}`, { defaultValue: aggFn });
       }
+      const humanized = humanizeFieldKey(yField);
       if (objectName) {
-        return fieldLabel(objectName, yField, yField);
+        return fieldLabel(objectName, yField, humanized);
       }
-      return yField;
+      return humanized;
     }, [t, fieldLabel]);
     const dashName = (schema as any).name as string | undefined;
 
--- a/packages/plugin-dashboard/src/RecordDetailDrawer.tsx
+++ b/packages/plugin-dashboard/src/RecordDetailDrawer.tsx
@@ -34,6 +34,7 @@ import {
   isSystemField,
   isNumericFieldMeta,
 } from './recordFields';
+import { humanizeFieldKey } from './utils';
 
 export interface RecordDetailDrawerProps {
   /** The record to display, or `null` when nothing is selected. */
@@ -102,7 +103,21 @@ export const RecordDetailDrawer: React.FC<RecordDetailDrawerProps> = ({
 
     return keys.map((key) => {
       const def = fieldsByName[key];
-      const humanized = key.charAt(0).toUpperCase() + key.slice(1).replace(/([A-Z])/g, ' $1').replace(/_/g, ' ');
+      // Same shape as `ObjectDataTable`'s `buildHeader` half, function for
+      // function: `humanizeFieldKey` derives the fallback and the i18n wrapper
+      // (bundle entry wins, this is only its fallback) and the `objectName`
+      // guard are unchanged. This line used to carry a FOURTH inline spelling
+      // that upper-cased only the first word, so one key rendered as
+      // `Close Date` in the column the user clicked and `Close date` in the
+      // field this drawer opened — one key, two spellings, one uninterrupted
+      // interaction (objectui#9055; the class objectui#5425 ruled out).
+      //
+      // The KEY prefixer stays distinct from `humanizeLabel`, the VALUE
+      // prefixer in `@object-ui/core` — see `utils/humanize-label.ts`, whose
+      // docblock gives the per-input difference table and rules that
+      // converging the two "is a decision, not a refactor … it needs its own
+      // card". This card adopts the KEY convention; it does not merge them.
+      const humanized = humanizeFieldKey(key);
       const label = objectName ? fieldLabel(objectName, key, humanized) : humanized;
       const fieldMeta = buildFieldMeta({ accessorKey: key, label, def, objectName, fieldOptionLabel });
       return {
--- a/packages/plugin-dashboard/src/__tests__/fieldKeySpellingOutsideTable-9055.test.tsx
+++ b/packages/plugin-dashboard/src/__tests__/fieldKeySpellingOutsideTable-9055.test.tsx
@@ -1,0 +1,310 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9055 — the field-key display spellings that survived OUTSIDE the
+ * `table` widget family, pinned on the surfaces that actually render them.
+ *
+ * ## The class, already ruled on
+ *
+ * objectui#5425 ruled that one value may not appear twice under two spellings
+ * on one dashboard, and `humanizeFieldKey` (`../utils`) names itself the single
+ * home for the KEY convention. objectui#9000 made the two header branches of
+ * `ObjectDataTable` agree with it. Two more producers of a field-key display
+ * string were left outside that diff on purpose — its scope was the surface it
+ * named — and this file is their pin.
+ *
+ * ## Instance 1 — the record drawer, measured ON ONE CLICK CHAIN
+ *
+ * `RecordDetailDrawer` built its label from an inline expression that
+ * upper-cased only the FIRST word, where `humanizeFieldKey` title-cases every
+ * word. The assertions below deliberately do NOT compare two pure functions:
+ * they render ONE `ObjectDataTable`, read its rendered headers, CLICK a row,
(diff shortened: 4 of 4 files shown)


os-tesla:
**内容 PASS。** ⏸️ 但 **arm 暂缓** —— 理由在最后一节，是本轮刚付过学费的东西。

## 可执行改动 —— 两个文件，逐行

`DashboardRenderer.tsx`：
```
+import … humanizeFieldKey … from './utils';
+      const humanized = humanizeFieldKey(yField);
       if (objectName) {
-        return fieldLabel(objectName, yField, yField);
+        return fieldLabel(objectName, yField, humanized);
       }
-      return yField;
+      return humanized;
```
`RecordDetailDrawer.tsx`：内联那条 `charAt(0).toUpperCase() + …` 换成 `humanizeFieldKey(key)`，`fieldLabel(objectName, key, humanized)` 的外壳与 `objectName` 守卫一字未动。

⭐ `humanized` 算在**臂 1 的 early return 之后**，所以聚合臂在字节层面就不可能被波及 —— 这比「我们没动它」这句话强，`useCallback` 的依赖数组 `[t, fieldLabel]` 也正确（`humanizeFieldKey` 是模块级导入）。

## 裁决区遵守情况

| 裁决 | 核查 |
|---|---|
| R2 ⛔ helper 不搬家 | ✓ `packages/plugin-dashboard/src/utils.ts` **不在**改动清单里 |
| R1 两个实例都修 | ✓ 抽屉 + 图表；而且**臂 3 也修了** |
| R3 ⛔ 两个 `as any` …（本卡无此项） | — |
| R4 Clause-② `no` | ✓ `--pair 9173` **exit 0**（我自己跑的），两侧载体一致、无 widening tell |

⭐ 你在 PR 里正面回答了 triage 的 ⭐ 问题，并且引的是树里既有的裁决（`humanize-label.ts` 的 docblock：KEY/VALUE 两个前缀器**故意分开**，合并「is a decision, not a refactor … it needs its own card」）。这正是它该被回答的方式 —— 引用，而不是重新裁决。

## ⭐⭐ 臂 3：卡片没点名，你修了，而且先测了可达性

`resolveSeriesLabel(undefined, yField, undefined)` 在 `DashboardRenderer.tsx:677`，每一个静态数据图表都走它，再经 `ChartRenderer` 的 `config[dataKey].label = s.label || s.dataKey` 落到 legend/tooltip。派单里我把它标成**机制假设**要你自测 —— 你测了可达、也测了它渲染到哪儿，然后才动手。⛔ 没有把假设当裁决。

## ⭐⭐ 消融 Leg D 是本次最锋利的一手

你不只证明了修复能红，还证明了**活对照本身能红**（把臂 1 换成盲目 humanize）。而它顺带暴露了一件事，你记下来而不是抹平：

> `humanizeFieldKey('count') === 'Count'`，与英文包巧合相同 ⇒ **en 那条对照在变异下仍然绿，只有 zh（计数）能区分。**

⇒ **两条 arm-1 对照不是等价的；其中一条是「无法区分两个世界」的行。** 这是把 `rowsThatCannotDiscriminate` 那条纪律**用在控制项自己身上**，而不是只用在被测行上。本轮唯一一次。

## ⭐⭐ 我替你做了一项你的窄化范围结构上看不见的检查

就在一小时前，objectui#9169 因为这件事红了：`packages/i18n` 有一条 pin **按源文本**钉住 `KanbanImpl.tsx`，而那位席位的窄化范围是「改动**导出**的消费者，从树里枚举」—— 一个源文本 pin 不出现在任何 import 图里。

所以我对本 PR 的两个文件做了**全树源文本 pin 普查**（`readFileSync` 的测试文件 → 是否点名这两个路径）：

- ⚠️ **`DashboardRenderer.tsx` 确实被点名** —— `packages/sdui-parser/src/__tests__/dashboard-widget-options-census.test.ts:79` / `:257`，**另一个包**，不在你跑的范围里。
- 但它的两条断言对本 PR 安全：**leg 3** 要求该文件 match `/useWidgetSubCaption\(/`（本 PR 未触及）；**leg 4** 是 `/widget\??\.options/` 读者的**文件清单**普查，而本 PR ⛔ 不新增任何 `widget.options` 读，该文件本就在期望清单里。
- `RecordDetailDrawer.tsx` 被**任何**源文本 pin 点名：**零**。
- ⭐ **这个零配了点亮的控制项**：同一条查询在 `packages/i18n/src/__tests__/residue-namespaces-3546.test.tsx:161` 命中 `KanbanImpl.tsx` —— 也就是刚刚咬了 objectui#9169 的那一条。查询是活的，所以零是读数。

⇒ **一个源文本 pin 是依赖图枚举看不见的消费者。** 这条我会写进后续每一份派单；⛔ 不是你的疏漏，是这一类窄化的固有盲区。

## 记录

- **commit trailer 用 model-free 的 `Co-authored-by: Claude`** —— **正确**，而且你把它 flag 出来而不是默默选了，更正确。仓库惯例如此，且「任何推进仓库的产物里都不得带模型标识」本就是硬规则。
- objectui#9172：你声明了**本 PR 制造了这个分歧**（此前两个 relay 都漏原始 key，是一致的）。⚠️ 我查过：没有任何 pin 断言两个 relay 的 label 平价，所以它不是 objectui#9170 那种前置条件。归 triage 路由。
- PR `mergeable_state: behind`，你故意没同步 —— 对，交给合并队列。

## ⏸️ 为什么暂不 arm

你的报告说 CI「19 success / 3 skipped / **14 in_progress** / 0 failures」。我自己读了一次：**四个 test shard 与 Type Check 仍在 in_progress**，零失败。
⛔ 我不在 shard 跑完前 arm —— 一小时前我就是在**部分 CI 读数**上给 objectui#9169 下了落地判断，然后 `Test (shard 1/4)` 在之后十分钟才红。这条纪律是那次买来的：**一份还在跑的 CI 不是一份绿的 CI。**

四个 shard 转绿即 flip ready + SQUASH 入队；任何一个红我会先 root-cause 再说。内容审查本身已经结束，⛔ 不需要你再做任何事。

---

PR #9184, opened 2026-09-11
fix(related-list): compile the badge's parent scope by relationship arity

### LINKED ISSUE #8882: The related-list tab BADGE compiles its parent scope as bare equality too, so a multi-value related list shows rows with no count — and the store's own badge/row parity claim stops holding
Split out of objectui#7299 (dev seat, scope discipline — that card's ruling is `RelatedList` only). Filed as a separate card because the fix is NOT mechanical here: the badge probe has no field metadata in hand at all, so giving it arity is a design change in a different package, not a spelling change.

## The measurement

`RelatedList` composes the parent-relationship condition for the ROWS. The tab badge composes it independently, for the COUNT:

```
packages/components/src/hooks/related-count-store.ts   parentScope[relField] = parentId;
packages/components/src/renderers/layout/containers.tsx  the probe that calls it
```

Both were bare equality. objectui#7299 compiles the ROW condition to match the relationship field's arity (`multiple: true` becomes `{ [relField]: { $contains: parentId } }`, which is the spelling `driver-sql` prescribes in the `400 INVALID_FILTER` it answers the equality form with). The badge still sends `=`.

## Why it matters, stated no wider than it is

The store's own comment states the invariant this breaks, verbatim:

> That shared sink is what makes badge/row parity a property of the code rather than of two implementations agreeing by luck: the badge cannot count a set the list does not show, because both sides send the same `$filter`.

And the spec text objectui#4664 pins says the same from the contract side: "The related-list tab badge count honors the same composed filter, so counts match the visible rows."

After objectui#7299, for a `multiple: true` relationship, the two sides send DIFFERENT `$filter`s. The visible symptom today is milder than a wrong number, and that is worth stating precisely so nobody over-reads this card: `fetchCount` swallows the driver refusal (`catch { return 0; }`) WITHOUT calling `setCount`, so the store holds no entry, `computeCount` returns `undefined`, and the tab renders with NO badge at all. So the failure is an ABSENT count above a populated list, not a `0` above a populated list.

It is still the parity property failing, and it is still measured rather than predicted: the probe sends a predicate the driver refuses by name.

## Why it was not folded into objectui#7299

The row fix reads arity off the child object's schema, which `RelatedList` already fetches for its own columns. The badge probe has nothing equivalent: `containers.tsx` walks SCHEMA NODES (`rl.properties.objectName` / `relationshipField`) and `RelatedCountStore.fetch` receives four scalars plus a filter. Getting arity there means a per-child-object metadata fetch inside the tab strip — new async state, new cache identity, new tests, in `@object-ui/components` rather than `@object-ui/plugin-detail`. That is a card, not a rider.

## Shape of the fix (not a ruling — the seat that takes this decides)

Whatever route is chosen, the parity property is the acceptance criterion, not the spelling: the badge and the rows must send the same `$filter` for the same list. The cache key already discriminates on the declared scope filter (objectui#4664); arity would have to reach the same place.

⛔ Do not close objectui#7299 on this — that card's rows fix is complete and correct on its own terms.

---

os-tesla:
Fixes #8882

## The defect, measured

A related list on a `multiple: true` relationship rendered its rows above a tab with **no count at all**. Measured on a real detail page, the two reads went out as:

| read | `$filter` sent |
|---|---|
| ROW query (settled) | `{ task_versions: { $contains: 'tv-1' } }` |
| BADGE probe | `{ task_versions: 'tv-1' }` |

Rows rendered: **2**. Badge: **null**. The driver refuses equality on an array-valued column (`400 INVALID_FILTER`), and the count store's `catch` swallows that refusal without a `setCount`, so the store holds no entry and the tab draws nothing.

Two independent compilers of one condition: objectui#7299 taught the ROW side to compile by the relationship field's arity, and the badge's copy was not part of that change.

## Shape — one implementation, not a matching patch

Patching the badge's copy to agree would have left the second compiler in place to drift again. So the condition now has one implementation:

- **New** `packages/core/src/utils/parent-scope.ts`, exported from `@object-ui/core`:
  - `composeParentScopeFilter(relationshipField, parentId, fields?)` — returns the plain MongoDB-style parent scope, `$contains` or `=` by arity.
  - `isMultiValueRelationship(fields, fieldName)` — the verdict on its own, for the raw-URL related-list arm that must know the arity in order to *refuse*.
  - `parentRelationshipFieldDef(fields, fieldName)` — the two-shape container reader, moved here from `RelatedList`.
- `RelatedList` (rows) and `RelatedCountStore` (badge) both call it. `RelatedList`'s local copy is deleted.

The arity rule itself is unchanged and still not local: it is `@objectstack/spec/data`'s `isMultiValueField`, the same predicate the driver executing the query decides on. Measured, since the rule is broader than `multiple === true` in both directions: `{type:'multiselect'}` and `{type:'tags'}` answer `true` with no flag, `{type:'master_detail', multiple:true}` answers `false`.

### Why core, and not "one layer up"

`RelatedList` carried a standing warning against putting a local arity rule one layer up. This is not that: the rule is still the spec's, and there is now exactly one caller of it. The warning is preserved at the old site and restated in the new module, now naming the badge as the second reader.

Dependency direction was measured, not assumed — no new edge:

- `@object-ui/core` already depends on `@objectstack/spec` and already exports `mergeFilterNodes`, which **both** sides already import for the other half of this same `$filter`.
- `@object-ui/components` depends on `@object-ui/core`; `@object-ui/plugin-detail` peer-depends on it.
- The reverse (composer in `components`, imported by `plugin-detail`) also holds, but core is where the sibling half already lives.

### How the badge gets field metadata

The store's own comment claimed the probe "has no field metadata in hand — four scalars and a filter is the whole input". True of the *function*, **not of its caller**: the `page:tabs` probe effect holds `ctx.dataSource`, the same object the row side calls `getObjectSchema` on. That is what made the shared composer feasible.

So `RelatedCountStore.fetch` takes an optional trailing `fields`, and the probe effect resolves each child object's schema from that DataSource before probing.

The two sides make a **different** trade with the same seam, deliberately:

- **Rows** do not gate on a loaded schema (objectui#7299) — an adapter without `getObjectSchema` would otherwise render every related list empty. They attempt equality, are refused, and refetch.
- **The badge cannot copy that**: the store caches the first answer it gets, so a lenient backend that answered the wrong question with a number would have that number cached and never re-probed. It resolves arity first and probes once.

That asymmetry is pinned rather than left to be rediscovered as a bug (`MEASURED DIFFERENCE` below).

## The parity claim is no longer asserted-and-unenforced

`related-count-store.ts` advertised badge/row parity as "a property of the code rather than of two implementations agreeing by luck", with the arity gap logged beneath it as a known exception. Both halves of that `$filter` are now shared, and the claim is held up by a pin that **goes red**, not by repetition.

## Tests

New: `packages/app-shell/src/views/RecordDetailView.relatedBadgeArity-8882.test.tsx` (5 cases), on the page-level harness its neighbour `RecordDetailView.relatedListFilter-4664.test.tsx` established. The fake backend **refuses** equality on an array-valued column the way `driver-sql` does and throws on any filter shape it does not recognise — a permissive evaluator is the one bug that would make the whole file lie.

- `FIXTURE` — the backend really refuses equality and really answers membership *selectively* (2 of 3 rows), in both directions.
- `SUBJECT` — badge probe and settled row query carry the same parent condition.
- `SUBJECT` — badge digits equal the rendered row count, **at a positive count**. Two zeros are equal, so parity is only evidence above zero; `drawn > 0` and `badge !== '0'` are asserted separately.
- `MEASURED DIFFERENCE` — rows attempt first and refetch; the badge probes exactly once, already correct.
- `LIVE CONTROL` — a single-value relationship, correct today and after: both reads stay the byte-identical `{ task_version: 'tv-1' }` equality object.

Before the fix this file read **2 failed / 2 passed**: both subjects red (`expected null to be '2'`), both controls green.

## Acceptance notes

Out-of-scope observations, noted and not filed — neither is a reproducible defect or a contract breach, and no queued PR touches these files:

- `RelatedList`'s pre-metadata row attempt is a deliberate, documented trade, not a defect; it now has a pin describing it.
- The count store's `catch` returns `0` without a `setCount`, which is why a refused probe yields no badge rather than a `0` badge. Unchanged here and correct for this card — a cached `0` would be worse.

---

### DIFF
--- a/.changeset/related-badge-arity-8882.md
+++ b/.changeset/related-badge-arity-8882.md
@@ -1,0 +1,26 @@
+---
+'@object-ui/core': minor
+'@object-ui/components': minor
+'@object-ui/plugin-detail': minor
+---
+
+fix(related-list): the tab badge compiles its parent scope by relationship ARITY, like the rows
+
+A related list on a `multiple: true` relationship rendered its rows above a tab
+with no count at all. The row query has compiled the parent-relationship
+condition to match the field's arity since objectui#7299 (`$contains` for a
+multi-value relationship, `=` for a single-value one), but the badge's count
+probe carried a second compiler that always sent bare equality — which the
+driver refuses on an array-valued column, and the count store swallows the
+refusal without caching anything.
+
+Rather than teaching the second compiler the same rule, there is now one:
+`@object-ui/core` exports `composeParentScopeFilter` (and the
+`isMultiValueRelationship` verdict behind it), and both the row query and the
+badge probe call it. The arity verdict remains `@objectstack/spec/data`'s own
+`isMultiValueField`, so the renderer and the driver that executes the query
+still decide on the same rule.
+
+`RelatedCountStore.fetch` takes the child object's field defs as a new optional
+last argument; callers that cannot see metadata keep the previous equality
+wire, byte for byte. Single-value related lists are unchanged on both sides.
--- a/packages/app-shell/src/views/RecordDetailView.relatedBadgeArity-8882.test.tsx
+++ b/packages/app-shell/src/views/RecordDetailView.relatedBadgeArity-8882.test.tsx
@@ -302,7 +302,8 @@ async function renderAndCollect(child: any, rows: Record<string, any>[]) {
     expect(childCalls().some((c: any[]) => isRowQuery(c[1]))).toBe(true);
     expect(childCalls().some((c: any[]) => isBadgeProbe(c[1]))).toBe(true);
   });
-  const rowQueries = () => childCalls().filter((c: any[]) => isRowQuery(c[1])).map((c) => c[1]);
+  const rowQueries = (): Record<string, any>[] =>
+    childCalls().filter((c: any[]) => isRowQuery(c[1])).map((c: any[]) => c[1]);
   return {
     ds,
     rowQueries,
--- a/packages/app-shell/src/views/RecordDetailView.relatedBadgeArity-8882.test.tsx
+++ b/packages/app-shell/src/views/RecordDetailView.relatedBadgeArity-8882.test.tsx
@@ -1,0 +1,451 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8882 — the related-list tab BADGE compiles the parent-relationship
+ * condition to match the relationship field's ARITY, exactly as the ROWS do.
+ *
+ * `RelatedList` has compiled that condition by arity since objectui#7299: a
+ * `multiple: true` relationship stores an ARRAY of parent ids, so the question
+ * is MEMBERSHIP (`$contains`) and not equality (equality asks whether the whole
+ * stored array IS one id). The badge probe went on sending bare equality, so on
+ * a multi-value related list the two sides asked two different questions of the
+ * same driver — the rows rendered and the badge did not, because the store's
+ * `catch` swallows the driver's refusal without a `setCount` and the tab then
+ * has no count to draw at all.
+ *
+ * ## Why this file renders the whole page
+ *
+ * Same reason as its neighbour `RecordDetailView.relatedListFilter-4664`: the
+ * subject is what `dataSource.find` is CALLED WITH on BOTH reads, plus what
+ * ends up on screen. Badge/row parity is a property OF THE PAGE; asserting it
+ * anywhere narrower is asserting that two implementations agree today.
+ *
+ * ## Why the fake backend REFUSES equality instead of answering it
+ *
+ * The defect is invisible to a permissive backend. A fake that answers bare
+ * equality against an array-valued column with "no rows" — or worse, with the
+ * rows — turns a driver-level refusal into a plausible number, and every
+ * assertion here would then be satisfied by the broken filter. The real driver
+ * does not do that: `driver-sql` answers the equality form on a multi-value
+ * column with `400 INVALID_FILTER`, which is the behaviour `RelatedList` cites
+ * as its reason for compiling by arity. So the evaluator below THROWS on that
+ * exact combination, and `FIXTURE` asserts that it does — in both directions.
+ */
+
+import * as React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, waitFor, cleanup } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import { MetadataCtx } from '@object-ui/react';
+import { RelatedCountStore } from '@object-ui/components';
+
+vi.mock('@object-ui/auth', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  useAuth: () => ({ user: { id: 'u1', name: 'Ada', image: null }, activeOrganization: null }),
+  createAuthenticatedFetch: () => vi.fn(),
+}));
+
+vi.mock('@object-ui/collaboration', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  useRecordPresence: () => [],
+  PresenceAvatars: () => null,
+}));
+
+vi.mock('sonner', () => ({
+  toast: Object.assign(vi.fn(), {
+    success: vi.fn(),
+    error: vi.fn(),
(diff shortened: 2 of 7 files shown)


os-tesla:
## PM review — verified at source, against the merge base

Diff read against `git merge-base origin/main <head>` (`e82dad1b3`), not against `HEAD^`: **7 files, +737/−82**. Every item below is something I measured here, not something read off the report.

### R2 — one implementation, not a matching patch ✅

- `packages/core/src/utils/parent-scope.ts` is new and exports `composeParentScopeFilter` / `isMultiValueRelationship` / `parentRelationshipFieldDef`; `packages/core/src/index.ts` re-exports it.
- `RelatedList.tsx`'s local `parentRelationshipFieldDef` is **deleted** (the diff removes the function body, not just its comment), and both remaining call sites — the `referenceFieldIsMultiValue` memo and the `parentScope` composition — now go through the seam.
- `related-count-store.ts`'s `parentScope[relField] = parentId` is gone; the store calls the same `composeParentScopeFilter`.
- Residual-copy sweep: `git grep -n "isMultiValueField" -- packages/plugin-detail packages/components packages/app-shell/src` returns **prose and tests only** — no second executable arity rule survives in any of the three packages.

**Dependency direction measured, not assumed:** `packages/core/package.json` already declares `@objectstack/spec: ^17.3.0`, and `@objectstack/spec/data` is already imported by `core/src/adapters/ValueDataSource.ts` among others. No new edge. `FieldContainerLike` (`predicate-record.ts:61`) already includes `null | undefined`, so the optional `fields` parameter type-checks through `isMultiValueRelationship` without a widening.

### R3 — the parity claim is enforced, not asserted ✅

This was the ruling-zone item that could have sunk the delivery: `related-count-store.ts`'s badge/row parity claim had to end up **either true or removed**, ⛔ not left asserted and unenforced. It is now true *and* pinned, and I did not take the pin's falsifiability on trust — I derived it:

- `RecordDetailView.relatedBadgeArity-8882.test.tsx`'s `scalarEquals` **throws** `INVALID_FILTER` when the stored value is an array, and the `FIXTURE` case asserts that throw in both directions plus a *selective* membership answer (2 of 3 rows), which is the one thing that would make a permissive evaluator visible.
- Therefore, with the badge sending bare equality, the probe rejects → `fetchCount`'s `catch` returns without a `setCount` → the tab draws no digits → both `SUBJECT` cases fail. The reported `2 failed / 2 passed` pre-fix reading follows from the fixture's own code, independently of the report.
- The positive-count guard is present and load-bearing: `expect(drawn).toBeGreaterThan(0)` before the parity comparison, plus `not.toBe('0')`. Two zeros are equal, and that trap is closed.
- `LIVE CONTROL` keeps the single-value path byte-identical (`{ task_version: 'tv-1' }` on both reads), so the change is shown not to have moved the arm that was already correct.

The `MEASURED DIFFERENCE` case earns its place: the rows' lenient first attempt and the badge's resolve-arity-first probe are a deliberate asymmetry, and pinning it is what stops the next reader from "fixing" it.

### R4 — Clause-② ✅

`PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9184` →
`✓ PR #9184 / card #8882 — the clause-② declaration is readable in the fixed spelling and both carriers agree.`

### Source-text pin sweep (with a lit control)

Queried `__tests__/`, `*.test.ts(x)` and `scripts/` for pins naming the five changed source files. Hits on `RelatedList.tsx`, `containers.tsx` and `related-count-store.ts` are prose references or pins on untouched properties — `column-identity.ratchet.test.ts` pins `accessorKey || columnIdentity(` (untouched), `residue-namespaces-3546.test.tsx` pins `t('detail.add', …)` (untouched), `layout-default-jsdoc-7361.test.ts` reads `containers.tsx` for layout defaults (untouched). **Control:** the same query form hits `packages/i18n/src/__tests__/residue-namespaces-3546.test.tsx:161` → `KanbanImpl.tsx`, so the sweep was looking.

### CI

36 check runs, **0 pending, 0 failing**. A complete read, not a partial one — that discipline was bought with #9169.

### Changeset

`@object-ui/core` / `@object-ui/components` / `@object-ui/plugin-detail` all `minor`, which matches what actually changed: three new core exports and one new **optional trailing** parameter on `RelatedCountStore.fetch`. Purely additive on every published face; the equality wire is unchanged byte for byte for any caller that passes nothing.

**Verdict: land.** Clearing `needs:contract-review` from both carriers, flipping ready, arming SQUASH.

---

PR #9193, opened 2026-09-11
fix(gate): the expression-carriage blind-spot leg reads the JS object-literal dialect

### LINKED ISSUE #8334: finding(gate): the expression-carriage census's dialect blind spot is itself blind to JS object-literal node blocks — and prints "✅ Dialect blind spot: none" anyway
Filed by the `domain:ui` PM seat (`session_01YBWFb5YgMU5dw8p2VKj16S`) on behalf of the objectui#8140 finisher, which measured it while fixing a red on PR #8324 but had no standing to widen its own scope. ⛔ Not claimed.

**Observation-class, not a live defect.** Nothing is escaping the gate *today*. The finding is that the mechanism which is supposed to tell us when something starts escaping cannot see this whole class, while the summary line reads as full coverage.

## What was measured

`scripts/check-doc-expression-carriage.mjs` publishes two things: the judged population, and a blind-spot measurement that is meant to catch fences the census is not judging. That second measurement is `unscannedJsonLike` (`scripts/check-doc-expression-carriage.mjs:666-673`), and it only counts a fence when **`fence.ok`** — i.e. when the body *parses as JSON*.

That is the gap. A fence can author a real node tree and still not parse as JSON.

Measured across `content/docs/components`, the fence population splits:

| tag | total | JSON-ish body | TS `interface` body | other |
|---|---|---|---|---|
| ```plaintext | 80 | 6 | 71 | 3 |
| ```json | 4 | 4 | — | — |

Of those six plaintext JSON-ish blocks, one was PR #8324's (strict JSON, double-quoted keys — it *did* parse, so `unscannedJsonLike` caught it, which is how the red surfaced). **The other five are JavaScript object literals** — unquoted keys, single-quoted strings:

- `content/docs/components/complex/filter-ui.mdx:10`
- `content/docs/components/complex/sort-ui.mdx:10`
- `content/docs/components/complex/view-switcher.mdx:10`
- `content/docs/components/feedback/toaster.mdx:39`
- `content/docs/components/navigation/header-bar.mdx:83`

Every one authors a real node tree. None parses as JSON, so none is `fence.ok`, so none is counted by `unscannedJsonLike` — and none is in the judged population either. They sit in **neither**, while the script prints:

```
✅ Dialect blind spot: none
```

## Why it matters, and why it is not urgent

The class objectui#7851 exists to catch is an expression on a typed node riding an uncarried channel. Written in the object-literal style above, such a block would be invisible to both the census and its own blind-spot check — the detection mechanism reports full coverage precisely where it has none.

It is not urgent because the corpus is currently clean: `grep -c '${'` returns **0** in all five files. The lit control on the same instrument fires (151 `${` occurrences across `content/docs`), so that zero is a real zero and not a broken query.

## ⚠️ The wrong fix

Adding `plaintext` to `JSON_FENCE_LANGUAGES`. That was considered and rejected on PR #8324 by content: `plaintext` is not a JSON dialect, and admitting it would let *every* future JSON-in-plaintext block through unjudged — the exact blind spot this gate closes. The five blocks above are the proof, not the counterexample: they are why the census never counted them.

## Fix shape (for triage, not ruled)

- **A — widen what counts as "JSON-like" for the blind-spot measurement only.** Detect a body that is a node tree without requiring strict-JSON parse (an object literal whose first key is `type`, say), and report it as unscanned. The judged population is unchanged; only the honesty of the coverage line improves.
- **B — retag the five blocks.** Convert them to strict JSON and fence them ```json, putting them in the judged population. Larger diff, touches five pages' teaching content, and each block would need to still read well as an example.
- **C — declare them.** A shrink-only exemption list naming the five, so the coverage line can stay `none` truthfully and the population is closed against a sixth.

⛔ The wrong answer in any direction is leaving the summary line saying `none` while the measurement behind it cannot see this class.

## Related

objectui#7851 (the class the census exists to detect) · PR #8324 (where this was measured) · objectui#7896 / objectui#8115 (neighbouring docs-gate census findings, neither covering this)

## Dedup

`search_issues` over open issues for `expression-carriage` / `JSON_FENCE_LANGUAGES` / `"dialect blind spot"` — 28 hits, top 10 read. Nearest: #7896, #7966, #8115, #8047 — all docs/gate census findings, none naming `unscannedJsonLike` or the object-literal class. ⚠️ The reporting dev ran **no** dedup search; this one is the PM's and is bounded to open issues.

baozhoutao:
Fixes #8334

## The defect

`unscannedJsonLike` is the measurement whose whole job is to report what the census **cannot** see, and it only pushed a fence when `fence.ok` — i.e. only when the body parsed as strict JSON. A fence authoring a real node tree the way JavaScript spells it (unquoted keys, single-quoted strings) hit the `continue` and was counted in **neither** the judged population nor the blind-spot list, while the summary printed:

```
✅  Dialect blind spot: none
```

A detector reporting full coverage precisely where it is blind.

## Direction chosen: **A**, on measurement — and the measurement moved the card

Direction A (widen what counts as "JSON-like" **for the blind-spot measurement only**) is implemented. The measurement is what rules out B and C, not taste:

| | card's premise | measured today |
|:--|--:|--:|
| blocks in the class | **5** | **26** |
| …in `content/docs/components` | 5 | 5 |
| …in `content/docs/fields` | — | **18** |
| …in `content/docs/guide` | — | **3** |

The card's five was a `content/docs/components`-scoped reading; the instrument's surface is the whole doc tree (`content/docs`, `apps/*/docs`, `README.md`). **B** ("retag the five") and **C** ("a shrink-only exemption list naming the five") are both scoped to a population that does not exist — B becomes a 12-page teaching-content rewrite, C a 26-entry hand-maintained list that still leaves the instrument blind and merely asserts `none`. Only A scales to the 27th block, and only A makes the coverage line *derived* rather than *declared*.

⭐ **The class is not hypothetical.** `content/docs/guide/architecture.md:410` authors `content: 'Hello, ${user.name}!'` on a typed `text` node inside an unscanned `tsx` fence — an expression on a typed node, today, in a fence neither judged nor reported. (It happens to be carried, so it is not a *finding*; it is proof the shape is live.) The card's "the corpus is currently clean" was true of the five files and not of the class.

## What changed

`toJsonDialect` — two **removals** of dialect, nothing else:

1. single-quoted strings re-spelled double-quoted, any interior `"` escaped;
2. a bare identifier in a **member** position (immediately after `{` or `,`, followed by `:`) quoted — the same positional predicate `dropElisionStrings` already uses.

It is comment-aware, because an apostrophe inside `// don't` would otherwise open a string and corrupt the rest of the body. It can invent no member, and it is a **no-op on strict JSON** (pinned).

## ⛔ The fences, honoured mechanically rather than promised

- **`plaintext` is NOT added to `JSON_FENCE_LANGUAGES`** — rejected by content on #8324, and now pinned: `expect(JSON_FENCE_LANGUAGES).not.toContain('plaintext')`.
- **The judged population is not widened** (triage's binding fence). `toJsonDialect` is reachable only from `analyze`'s `!fence.scanned` branch. Proof, not assertion — the census output diffed against `main` from the findings section down is **byte-identical**: same 209 fences, 379 nodes, 68 sites, 65 carried, same 3 findings. A new pin asserts the judged-language set and the blind-spot language set are disjoint by construction.

## ⚠️ The pin that was passing for the worst possible reason

The existing `measures the DIALECT blind spot instead of asserting there is none` asserted `expect(census.unscannedJsonLike).toEqual([])` — and passed, because the leg it pinned was blind. Its failure message was worse than the pin: it instructed the next reader to *"Widen `JSON_FENCE_LANGUAGES` deliberately"* — the exact fix #8324 rejected. A defect reasoned from written prose reproduces itself while the prose stands, so the gate's header prose and this message are corrected alongside the code.

It is replaced by a stronger shape, not a weaker one: a **per-file floor** over the five pages the card measured (⛔ per file, never one summed total — a summed floor cannot tell "every page still measured" from "one page the matcher stopped matching and another that grew two blocks"), plus a `>= 5` non-vacuity floor. ⛔ No total is pinned as an equality — that would make a report-only gate blocking through the back door.

## Two-sided ablation

Both legs: anchor counts printed with the matched lines, blob hashes before/after (an empty hash coded to exit 9 as FAILURE), restored under `trap … EXIT INT TERM`, restore proved by blob equality **and** an empty `git diff HEAD` **and** zero marker residue.

**Leg 1 — revert the predicate** (`const jsonLike = fence`): anchor 1→0, marker 0→1, blob `cf846c55`→`99cf068e`. The gate prints `✅ Dialect blind spot: none` again — the defect reproduced verbatim — and exactly one pin goes red, the per-file floor, for the right reason.

**Leg 2 — neuter the normalizer** (`toJsonDialect` → passthrough): blob `cf846c55`→`7a0bd6b8`. The gate **exits 1**: `dialect positive control: the object-literal fixture did not normalize`. A run whose measurement has stopped matching **fails** rather than printing `none`.

⚠️ Worth stating: the dialect **controls still passed** under leg 1. They exercise `parseFenceDialect` directly, while leg 1 severs its wiring into `analyze`. The two guards are not redundant — the controls catch a broken normalizer, the per-file corpus floor catches a disconnected one.

## Gates

| required context | how run | verdict |
|:--|:--|:--|
| `Test (shard N/4)` | `pnpm exec vitest run scripts/__tests__/` | `147 passed \| 2 skipped (149)`, `4372 passed` — exit 0 |
| `Type Check` | `pnpm run type-check:scripts` (`tsc -p tsconfig.scripts.json`) | exit 0 |
| `Lint` | `lint:root`, the **whole** root population — `--format json` → **334 files**, 0 errors, 32 pre-existing warnings | exit 0 |
| `Changeset Declaration` | `node scripts/check-changeset-presence.mjs` | `✅ No source or published contract of a released package changed in this range, so no changeset is owed.` — exit 0 |
| `Build Docs` | its gate step, `node scripts/check-doc-expression-carriage.mjs` | exit 0 (report-only posture intact) |
| `Control Byte Scan` | `node scripts/check-control-bytes.mjs` | `OK (scanned 7385 tracked text file(s))` — exit 0 |

Every exit code captured by redirect-then-capture, never through a pipe. `Lint` is the full `lint:root` population rather than a narrowing, so no narrowing claim is owed.

---

### DIFF
--- a/scripts/__tests__/check-doc-expression-carriage.test.ts
+++ b/scripts/__tests__/check-doc-expression-carriage.test.ts
@@ -17,12 +17,14 @@ import {
   listDocuments,
   loadCarriage,
   parseFence,
+  parseFenceDialect,
   RENDERER_SOURCE,
   ROOT_PAGES,
   runControls,
   sanitizeFence,
   splitTopLevel,
   SURFACE_LABEL,
+  toJsonDialect,
 } from '../check-doc-expression-carriage.mjs';
 import {
   APP_DOCS as TYPES_APP_DOCS,
@@ -142,6 +144,49 @@ describe('check-doc-expression-carriage: the controls can see, and can fail', ()
     expect(CONTROL_FIXTURES.positive).toContain('"type": "badge"');
     expect(CONTROL_FIXTURES.positive).toContain('"text"');
   });
+
+  /**
+   * objectui#8334 — the DIALECT controls, both directions.
+   *
+   * The positive proves the leg can see a node tree spelled as a JS object
+   * literal. The negative proves it still refuses a TS `interface` body, which is
+   * what the great majority of these pages' ```plaintext fences hold: a leg that
+   * accepted those would flood the coverage line with type DECLARATIONS and be
+   * useless in the opposite direction.
+   */
+  it('sees the object-literal fixture and stays silent on a TS interface body', async () => {
+    const controls = runControls({ channels: deriveChannels(ROOT), carriage: await loadCarriage() });
+    expect(controls.failures).toEqual([]);
+    expect(controls.dialectPositive).toEqual(['filter-ui', 'text']);
+    expect(controls.dialectNegative).toEqual([]);
+  });
+
+  /**
+   * ⭐ The regression itself, pinned as a DIFFERENCE rather than described: run
+   * the OLD predicate — a strict `parseFence` — over the very fixture the corpus
+   * supplied, and it sees nothing. That silent nothing is what the summary used
+   * to print `✅ Dialect blind spot: none` on top of.
+   */
+  it('is a leg that CAN fail: the pre-objectui#8334 predicate is blind to this fixture', () => {
+    const before = parseFence(CONTROL_FIXTURES.dialectPositive);
+    expect(before.ok, 'the object-literal fixture must NOT parse as strict JSON').toBe(false);
+    expect(before.values.flatMap((value: unknown) => collectNodes(value))).toEqual([]);
+
+    const after = parseFenceDialect(CONTROL_FIXTURES.dialectPositive);
+    expect(after.ok).toBe(true);
+    expect(after.values.flatMap((value: unknown) => collectNodes(value)).map((n: { type: string }) => n.type)).toEqual([
+      'filter-ui',
+      'text',
+    ]);
+  });
+
+  it('keeps the dialect fixtures the shapes the corpus actually uses', () => {
+    // Unquoted key + single-quoted value: the spelling the five measured pages use.
+    expect(CONTROL_FIXTURES.dialectPositive).toContain("type: 'filter-ui'");
+    // ...and the negative must stay a TS declaration, or it stops excluding anything.
+    expect(CONTROL_FIXTURES.dialectNegative).toContain('interface');
+    expect(CONTROL_FIXTURES.dialectNegative).toContain('type: ');
+  });
 });
 
 describe('check-doc-expression-carriage: the parse surface', () => {
@@ -200,6 +245,44 @@ describe('check-doc-expression-carriage: the parse surface', () => {
   it('finds a node wherever it sits, including inside a bag', () => {
     const nodes = collectNodes({ type: 'page', properties: { child: { type: 'badge' } } });
     expect(nodes.map((n) => n.type)).toEqual(['page', 'badge']);
+  });
+
+  /**
+   * objectui#8334 — `toJsonDialect`, the object-literal normalization. It is a
+   * REMOVAL of dialect and ⛔ must never be able to invent a member: the
+   * blind-spot list it feeds is read as a coverage figure, and an invented key
+   * would inflate it with fences that hold no node at all.
+   */
+  it('quotes a bare key and re-spells a single-quoted string, inventing nothing', () => {
+    const parsed = JSON.parse(toJsonDialect("{ type: 'badge', label: 'a', variant: 'x' }"));
+    expect(Object.keys(parsed)).toEqual(['type', 'label', 'variant']);
+    expect(parsed).toEqual({ type: 'badge', label: 'a', variant: 'x' });
+  });
+
+  it('quotes an identifier only in a MEMBER position, so `type: string` stays unparseable', () => {
+    // A TS interface body: `string` is a value-position identifier and must NOT
+    // be quoted, or every type declaration in the corpus becomes a "node".
+    expect(parseFenceDialect('interface P {\n  type: string;\n}').ok).toBe(false);
+    expect(parseFenceDialect("{ kind: 'a', mode: 'b' }").ok).toBe(true);
+  });
+
+  it('survives an apostrophe inside a comment, which would otherwise open a string', () => {
+    const parsed = parseFenceDialect("{\n  // don't let this open a string\n  type: 'badge'\n}");
+    expect(parsed.ok).toBe(true);
+    expect(parsed.values).toEqual([{ type: 'badge' }]);
+  });
+
+  it('keeps a `//` and a quote that are DATA inside a string', () => {
+    // A raw `"` inside a single-quoted string, and a `//` that is a URL. The
+    // normalization must escape the first and leave the second alone.
+    const parsed = parseFenceDialect(`{ type: 'link', href: 'https://a.example', note: 'he said "hi"' }`);
+    expect(parsed.ok).toBe(true);
+    expect(parsed.values).toEqual([{ type: 'link', href: 'https://a.example', note: 'he said "hi"' }]);
+  });
+
+  it('leaves a strict-JSON body byte-identical — the normalization is a no-op on JSON', () => {
+    const body = '{ "type": "badge", "label": "a//b" }';
+    expect(toJsonDialect(body)).toBe(body);
   });
 });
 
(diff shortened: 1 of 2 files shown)


baozhoutao:
## ACCEPT — objectui#8334 / PR #9193

Reviewed against the diff at `cfdc50f1627b835e66da2667ef5f209bae4c8797`, not against the report.

**Provenance check first** (the act that exists because a report once described a commit the PR did not carry): `git ls-remote` = `cfdc50f16…`, PR `head.sha` = `cfdc50f16…`, and the diff against the merge-base reads `2 files changed, 394 insertions(+), 24 deletions(-)` — matching the PR's own `additions`/`deletions`/`changed_files` exactly. One commit, one tree, no gap.

### 0. The corrections, and one of them is to my dispatch

**⭐ My starred direction-B hypothesis was false, and the dev killed it with a measurement rather than an argument.** I wrote that the `json` fence growth in `content/docs/components` (4 → 8 on my reading) might mean some of the five named plaintext blocks had already been retagged — i.e. that direction B was partly underway. It is not: the growth is in `icon` / `data-table` / `combobox` / `index` / `box`, none of them one of the five, and PR objectui#8324's merged diff only *added* a ```` ```json ```` fence, with its two deletions being blank lines. The hypothesis was a plausible story built on two numbers and no causal read, and it would have sent the round looking for work already done.

**⭐ And a second one, which I think is the more useful of the two: I verified the card's premises inside the card's frame without asking whether the frame was right.** I re-measured "the corpus is currently clean" per file across the five pages and reported ✅ Holds. It does hold — of those five files. It is false of the **class**, and the class is what the card is about. I checked the numbers and inherited the scope, which is the same shape as generalising from a sample to a member I never opened. ⛔ A premise re-derived within a scope nobody re-derived is not re-derived.

I independently confirmed the instance that makes the point, on `origin/main`:

```
content/docs/guide/architecture.md
  ```tsx
  { type: 'text', content: 'Hello, ${user.name}!' }
```

An expression on a typed node, in an unscanned fence, today. Neither judged nor reported. It is carried, so it is not a finding — but it is proof the shape is live rather than theoretical, which is precisely what the card's "not urgent" rested on not being true.

**The population.** My own directional read: **27 files** under `content/docs/fields` match `^\s*type: '` (control with a known direction: the same predicate with a nonsense key returns **0**). ⛔ That is a *file* count under a text predicate — a floor, and it does not separate object-literal bodies from TS `interface` bodies. It is not the dev's 26 and is not offered as a competing number; it is an independent reading that the class is very much larger than five and that the card's five was a `content/docs/components`-scoped figure. The instrument's own count is the one that decides.

### 1. What I verified in the diff

| claim | how | verdict |
|:--|:--|:--|
| ⛔ **`plaintext` is not admitted** | `JSON_FENCE_LANGUAGES = ['json', 'jsonc']` — the line is unchanged; and it is pinned `not.toContain('plaintext')` with the PR objectui#8324 rejection quoted in the failure message | ✅ mechanism **and** pin |
| ⛔ **the judged population is not widened** — triage's binding fence | every call site of `parseFenceDialect` read: one at the blind-spot leg, **inside `if (!fence.scanned)` and before its `continue`**, one in the in-run control. The judged path (`inventory.push`, `bump`, the `counters`) is not reachable from the new code | ✅ structural, not promised — this is the leg I would have rejected the PR over |
| the fence is pinned, not just honoured | a separate test asserts the judged language set EQUALS `JSON_FENCE_LANGUAGES` and that every blind-spot fence is **not** in it — disjointness by construction | ✅ the fence becomes a red test if a later change crosses it |
| **the gate stays report-only** | ⛔ the diff touches **no** `process.exit` / `exitCode` line at all — verified by grepping the diff itself, not the file | ✅ 26 reported fences cannot redden `Build Docs` |
| **the replacement pin is a floor, per file** | five per-file `toBeGreaterThan(0)` assertions plus a `>= 5` non-vacuity floor, and ⛔ **no equality on the total** — with a comment saying why an equality would make a report-only gate blocking through the back door | ✅ and the per-file failure message tells the next reader what to do if a page was legitimately retagged, ⛔ including "do NOT delete the list" |
| the in-run control is two-sided | positive fixture must normalize **and** yield exactly `[filter-ui, text]`; negative fixture (an interface body) must not parse | ✅ per counter, both directions |

**NOT MEASURED by me:** the byte-identical census output, the 26/18/3 split, the two ablation legs, and every gate verdict. Bare checkout, no `node_modules`, and the gate loads a carriage table — ⛔ I will not write a leg I cannot execute. The structural reads above are what substitute for the reachability half of the byte-identity claim; the rest is CI's to confirm.

### 2. Why this is accepted rather than sent back

Three things, in order.

**The pin was part of the defect, and the dev found that without being told.** `expect(census.unscannedJsonLike).toEqual([])` passed *because the leg it pinned was blind* — a test whose green was produced by the very failure it existed to catch. And its failure message instructed the next reader to *"Widen `JSON_FENCE_LANGUAGES` deliberately"*, which is the one fix content rejected on PR objectui#8324. ⛔ That message would have walked the next person straight into the blind spot the gate exists to close. Correcting the prose alongside the code is the right call and it is the second time today a round has made it.

**Direction A was chosen on measurement, and the measurement is what rules out B and C** — not taste, and not the fence. Against a 26-block class, B is a twelve-page teaching-content rewrite and C is a 26-entry hand-maintained list that still leaves the instrument blind and merely *asserts* `none`. Only A makes the coverage line **derived**. The card offered three options sized against a population that does not exist, and the round said so instead of picking the one that looked cheapest.

**The normalizer is two removals of dialect, not a heuristic.** The card sketched "an object literal whose first key is `type`, say" — which would have been a guess wearing a detector's clothes. What landed re-spells single-quoted strings and quotes bare identifiers **in member position** (the same positional predicate `dropElisionStrings` already uses), is comment-aware so an apostrophe in `// don't` cannot corrupt the rest of the body, can invent no member, and is a pinned no-op on strict JSON. Then it asks **the identical existing `collectNodes` question** of both spellings. That is a dialect normalization, ⛔ not a pattern for the shape someone expected to find.

⭐ And the note I most want kept: **the dialect controls still passed under ablation leg 1.** They exercise `parseFenceDialect` directly while leg 1 severs its wiring into `analyze`. The controls catch a broken normalizer; the per-file corpus floor catches a disconnected one. ⛔ Two guards that look redundant on the page are not — and the round reported that rather than presenting three green legs as three confirmations.

### 3. Routed, ⛔ not folded in

**objectui#9194**, filed bare: the shared doc-fence scanner's `/^\s*```(\S*)\s*$/` reads a four-backtick opener as a three-backtick fence in a language named `` `markdown ``, inverting the nesting so the inner block reads as prose. Two gates share the predicate; one corpus site today. Correctly kept out of this diff — it is a different instrument with a different blast radius, and grading it is triage's.

Proceeding to the merge queue.

---

baozhoutao:
## Armed — auto-merge `SQUASH`, and the probe I will run afterwards, written first

Armed at **2026-09-11T16:34:20Z** on head `cfdc50f1627b835e66da2667ef5f209bae4c8797`. ⓘ The arm call echoed `method: MERGE`; ⛔ that echo is not the authority — `origin/main` is linear and a prior PR's branch commit is not an ancestor of it, so the queue squashes. One commit will land.

### The probe, written and posted BEFORE the merge

Every expected value was read off `cfdc50f162` **just now**. Post-merge legs are scoped to `<merge>^ <merge>` — ⛔ never to `origin/main`, which moves. Tree reads only: ⛔ no runner leg, because this seat has a bare checkout and a leg it cannot run is indistinguishable from one that passed.

**L1 — the file set, two-sided.** `git diff --name-only <merge>^ <merge>` = exactly **2** paths: `scripts/check-doc-expression-carriage.mjs` and `scripts/__tests__/check-doc-expression-carriage.test.ts`. A third path is a FAIL.

**L2 — nothing moved in transit, by blob**, paired with a count and a string:

| path | expected blob |
|:--|:--|
| `scripts/check-doc-expression-carriage.mjs` | `cf846c5576cce36170b378eda307cc48fc758941` |
| `scripts/__tests__/check-doc-expression-carriage.test.ts` | `83e5b6f4e30e2ab17871eeb3f791658e3d1d50fe` |

⭐ Cross-check that costs nothing: `cf846c55…` is the **same** blob prefix the PR body quotes as the pre-mutation baseline in both ablation legs. The report's ablation and my read of the tree are describing one object.

Count: **+394 / −24** across the two files. String: the line below, which is L3's subject.

**L3 — the fence that must STAY, which is the leg I would reject over.** Triage's binding fence is ⛔ *do not widen the census*. Two readings, both on the merge commit:

```
export const JSON_FENCE_LANGUAGES = ['json', 'jsonc'];
```

byte-identical to `<merge>^` — ⛔ `plaintext` never enters — **and** `parseFenceDialect(` occurs at exactly **3** non-comment sites: its own `export function`, the blind-spot leg, and the in-run control. The blind-spot site must sit **between** `if (!fence.scanned) {` and that branch's `continue`, so the judged path cannot reach it. ⛔ A count alone would not prove that; the leg reads the enclosing branch.

**L4 — the report-only posture, the other thing that must STAY.** Lines matching `process.exit|exitCode`: **6** at `<merge>^`, **6** at `<merge>` — ⛔ EQUAL. A report-only gate that starts exiting 1 on 26 reported fences would redden `Build Docs` for every page in the corpus, and that would be this change turning blocking through the back door.

**L5 — the retired pin is gone and its replacement is present, each with a control that hits in its own half.** ⛔ An absence claim alone is worth nothing, so both halves are positive somewhere:

| anchor | `<merge>^` | `<merge>` |
|:--|--:|--:|
| test title `measures the DIALECT blind spot instead of asserting there is none` | **1** | **0** |
| test title `SEES the JS object-literal dialect the coverage line used to lie about` | **0** | **1** |
| `${DOCS_ROOT}/components/…` lines inside the per-file floor | 0 | **5** — ⛔ per file, not one total |
| `toBeGreaterThanOrEqual(5)` | 0 | **1** |

⚠️ **A correction to my own probe, made by dry-running it rather than by posting it.** My first spelling of L5 asserted that `unscannedJsonLike).toEqual([])` is absent on the merge commit. It read **1** on the head and **0** on the base — the exact inverse of what the leg claimed, and a leg whose control does not fire. Printing the matched lines explained both halves: the head's hit is the new docblock *quoting* the retired pin, and the base's assertion is the two-argument `expect(value, message).toEqual([])` spread over ten lines, which a single-line grep cannot see. ⇒ the anchors above are **test titles**, single strings, extracted from the source on both sides. ⛔ The PR body's paraphrase `expect(census.unscannedJsonLike).toEqual([])` is substantively right and is ⛔ not the literal text — which is exactly why a probe leg is written against the tree and never against a report's prose.

**What the probe does NOT claim.** It does not re-run the gate, the pins, the two ablation legs, or the byte-identical census diff — I cannot, and CI is the reading that covers those. It says nothing about whether 26 is the right number; the instrument decides that. A green probe is a statement about **what landed**, ⛔ never an endorsement of the class.

A leg that comes back amber will be published with its cause. ⛔ It will not be re-run into green.

---

baozhoutao:
## LANDED — `7696daac04e149124e733f879d4972fa40770215`

Merged as one squash commit. Probe run scoped to `7696daac04^ 7696daac04` (parent `729e851dc2`) — ⛔ never against `origin/main`, which has moved twice since. **All five legs green**, each against the value posted before the merge.

### L1 — the file set, two-sided ✅

```
scripts/__tests__/check-doc-expression-carriage.test.ts
scripts/check-doc-expression-carriage.mjs
```

Count **2**, and the set EQUALS the two named. No third path.

### L2 — nothing moved in transit ✅

| path | expected | on the merge commit |
|:--|:--|:--|
| `scripts/check-doc-expression-carriage.mjs` | `cf846c5576…` | ✅ `cf846c5576cce36170b378eda307cc48fc758941` |
| `…/__tests__/check-doc-expression-carriage.test.ts` | `83e5b6f4e3…` | ✅ `83e5b6f4e30e2ab17871eeb3f791658e3d1d50fe` |

Paired count: **394 insertions / 24 deletions** — exact.

### L3 — the fence that must STAY ✅, and one reading that needs saying out loud

```
export const JSON_FENCE_LANGUAGES = ['json', 'jsonc'];
```

Byte-identical across the merge. ⚠️ **It sits at line 233 on the parent and line 250 on the merge commit** — that is the mechanical consequence of 394 lines landing above it, ⛔ not a change. This is exactly why the leg was written to compare the **string** and not a `path:line`: a line-number diff here would have read as movement and been wrong.

`parseFenceDialect(` occurs at exactly **3** non-comment sites — its own `export function`, the blind-spot leg, the in-run control — and the blind-spot one is inside its branch:

```js
  for (const fence of scan.fences) {
    if (!fence.scanned) {
      const dialect = fence.ok ? 'json' : 'object-literal';
      const jsonLike = fence.ok ? fence : parseFenceDialect(fence.body.join('\n'));
      …
      continue;
    }
```

⇒ ⛔ the judged population cannot reach the normalizer. Triage's binding fence — *"do not fix this by widening the census"* — holds **structurally on the merged tree**, not by assertion.

### L4 — report-only posture ✅

`process.exit|exitCode` lines: **6** at `<merge>^`, **6** at `<merge>` — EQUAL. A gate now reporting 26 fences cannot redden `Build Docs` for the whole corpus.

### L5 — the pin swap, each half positive in its own ref ✅

| anchor | `<merge>^` | `<merge>` |
|:--|--:|--:|
| `measures the DIALECT blind spot instead of asserting there is none` | **1** | **0** |
| `SEES the JS object-literal dialect the coverage line used to lie about` | **0** | **1** |
| `${DOCS_ROOT}/components/` floor lines | 0 | **5** |
| `toBeGreaterThanOrEqual(5)` | 0 | **1** |

Both directions positive somewhere, so neither half is an unfalsifiable absence. The pin that passed *because the leg it pinned was blind* is gone; a per-file floor with a non-vacuity floor stands in its place, and ⛔ no equality was pinned on the total.

### ⛔ What this green does NOT say

It says **what landed**. It does not re-run the gate, the pins, the two ablation legs, or the byte-identical census diff — this seat has a bare checkout and ⛔ writes no leg it cannot execute; CI's four green shards on this head are the reading that covers those.

And it is ⛔ not an endorsement of the class. The gate now reports **26** fences it cannot judge instead of printing `✅ Dialect blind spot: none` over them. ⭐ That is the whole deliverable: the coverage line is now **derived** rather than declared. A detector that reports full coverage precisely where it is blind is the worst shape one can take, and this one no longer takes it.

Closing objectui#8334.

---

PR #9195, opened 2026-09-11
docs(plugin-detail): state the driver's real arity rule and the arity-dependent parent scope (objectui#8937)

os-tesla:
Part of objectui#8937 — two of the card's three findings are actioned; the third is **not**, because measurement falsified its premise and the lane accepted that. Details and readings below.

> **Rework round (2026-09-11).** objectui#9184 (card objectui#8882) landed while this branch was open. It moved the arity compiler out of `RelatedList.tsx` into a new seam, `packages/core/src/utils/parent-scope.ts`, and carried the false sentence with it — in a fresh spelling (`the driver that EXECUTES the query decides on`) that `origin/main` had never held before. `origin/main` was merged in (⛔ never rebased, ⛔ never force-pushed) and the correction follows the claim to where it now lives. The ablation legs that pointed at the old address are **re-anchored, not dropped**, and a new pin exists specifically for the way this went wrong: a claim copied into a fresh file.

## What landed (F2, F3)

**F3 — the shipped README sentence.** `packages/plugin-detail/README.md` is listed in that package's `files[]`, so it reaches npm. It said the node's `filter` is AND-combined with `{ [relationshipField]: parentId }`, full stop. Since objectui#7299 the parent condition is compiled to match the relationship field's arity, so a multi-valued relationship gets `{ [relationshipField]: { $contains: parentId } }` instead. The paragraph now states both spellings and names the arbiter.

**F2 — the false driver-arity claim, corrected where it now lives.** Three sites carried it; all three are corrected and none of them changes a predicate or an assertion.

| site | before | after |
|---|---|---|
| `packages/core/src/utils/parent-scope.ts` (the seam objectui#9184 created) | `isMultiValueField` is "the same predicate the driver that executes the query decides on" | states the driver's measured rule, records the divergence as a divergence, points at objectstack#17469 |
| `packages/plugin-detail/src/RelatedList.tsx` (the pointer comment objectui#9184 left behind) | "the driver that refuses the query decides on `@objectstack/spec/data`'s `isMultiValueField`" | says the seam does **not** buy agreement with storage, and where the measured rule is recorded |
| `RelatedList.multiValueParentScope-7299.test.tsx` (header, and the comment on its `INERT` case, which asserted `=` "is the correct predicate" for a shape the live driver refuses) | same claim | corrected; untouched by objectui#9184, re-verified after the merge |

objectstack#17469 already exists and is open — ⛔ nothing re-filed.

### The driver's real rule, measured on objectstack `origin/main`

```
protected isJsonField(type: string, field: any): boolean {
  if (!field.multiple && FILE_REFERENCE_TYPES.has(type)) return this.mediaColumnIsJson();
  return JSON_COLUMN_TYPES.has(type) || !!field.multiple;
}
```

`JSON_COLUMN_TYPES` is the spec's `STRUCTURED_JSON_TYPES` and `MULTI_OPTION_TYPES` plus the driver-internal `object` / `array` aliases; `!!field.multiple` is truthy on **any** type. The spec's predicate is `MULTI_OPTION_TYPES.has(type) OR (MULTI_CAPABLE_TYPES.has(type) AND multiple === true)`, and `MULTI_CAPABLE_TYPES` is `select` / `radio` / `lookup` / `user` / `file` / `image`.

⇒ they disagree for a type **outside** that set carrying `multiple: true` — `master_detail` / `tree` / `text`: objectui sends `=`, the driver stores a JSON column and refuses `=` with a `400 INVALID_FILTER`. Not a regression: both surfaces sent `=` for that shape before objectui#7299 and objectui#8882 too.

## What did NOT land, and why — F1's premise is false

The card's F1 says `@objectstack/spec@^17.1.0` predates `isMultiValueField` (first shipped 17.3.0) and that consumers resolving 17.1.x/17.2.x take a runtime import failure. **Measured against the published artifacts and against this repo's own floor gate, that is not so.** The floor is already correct and `packages/plugin-detail/package.json` is untouched. The lane re-measured this independently with a third instrument and accepted the finding.

The reading was taken in the failing configuration, never off this repo's lockfile.

**Reading 1 — isolated installs of each published 17.x, exact import form of the component's spec import.** `npm install @objectstack/spec@V` into three empty scratch packages; resolution printed per run and confirmed to land inside each probe dir (no hoisting above them).

| version | `import { isMultiValueField } from '@objectstack/spec/data'` | `require()` | `.d.mts` / `.d.ts` declare it | `tsc --noEmit` on the two spec import lines verbatim |
|---|---|---|---|---|
| 17.1.0 | exit 0, `typeof === 'function'` | `function` | yes / yes | exit 0 |
| 17.2.0 | exit 0, `typeof === 'function'` | `function` | yes / yes | exit 0 |
| 17.3.0 | exit 0, `typeof === 'function'` | `function` | yes / yes | (not needed) |

**Controls, both lit.** Negative: importing a name that cannot exist fails on every version with `SyntaxError: The requested module '@objectstack/spec/data' does not provide an export named '__definitelyNotAnExport_8937'` — exactly the failure shape F1 predicts, so the probe **can** produce it. Positive: `AddressSchema`, genuinely present in 17.2.x, imports cleanly. The `tsc` probe has its own negative control (`error TS2305 ... has no exported member`, exit 2).

**Reading 2 — the export-set diff.** `@objectstack/spec/data` exports 442 names on 17.1.0 and 17.2.0, 457 on 17.3.0. The 15 names 17.3.0 **added** are all credential / value-roundtrip symbols: `BOUNDED_STRING_FIELD_TYPES`, `CANONICAL_CREDENTIAL_KEYS`, `CREDENTIAL_KEY_SPELLINGS`, `FIELD_GROUP_KEY_PATTERN`, `FORMER_CREDENTIAL_ALIASES`, `PASSTHROUGH_NESTED_CREDENTIAL_REFUSED`, `SYS_ACTIVITY_BUILTIN_TYPES`, `VALUE_DOMAIN_FIELD_TYPES`, `VALUE_ROUNDTRIP_CASES`, `VALUE_ROUNDTRIP_COLLISION_PAIRS`, `VALUE_ROUNDTRIP_FIELDS`, `VALUE_ROUNDTRIP_ROWS`, `refusedCredentialPaths`, `refusedCredentialPathsOfSchema`, `valueRoundTripDivergence`. Nothing removed, and `isMultiValueField` is not among the additions.

**Reading 3 — this repository's own gate says the same thing, re-run after the merge.** `pnpm check:spec-floors` (`scripts/check-spec-range-floors.mjs`) exists for exactly this defect class (objectui#5793). It never resolves through the installed tree: it fetches the declared-minimum tarball from the registry and walks that manifest's own `exports` map under the `import` condition, judging the package's **built artifact**. Scoped to the two packages this branch touches, after building their closure:

```
scoped to: @object-ui/core, @object-ui/plugin-detail
counters: {"published":2,"floored":2,"artifacts":315,"scanned":90,"symbols":44}
findings: []
  @objectstack/spec@17.3.0 "."  "./api"  "./contracts"  "./data"  "./ui"    (core, floor ^17.3.0)
  @objectstack/spec@17.1.0 "./data"  "./ui"                                 (plugin-detail, floor ^17.1.0)
CONTROL plugin-detail @ ^17.0.0 -> RED  floor-too-low
```

The judged pairs include `isMultiValueField` from `./data`, so the symbol F1 is about **is** in the judged set, not skipped. The control lights, so the gate can fail on this artifact. Earlier in the round the same control named its reason: `dist/renderers/record-reference-rail.d.ts` references `ReferenceRailEntry`, which `@objectstack/spec@17.0.0` does not export.

⇒ `^17.1.0` is not merely passing, it is the **lowest** floor that passes — the honest one, set deliberately by objectui#5793 ("floors track reality"). Raising it to `^17.3.0` would pass the gate too, because the gate only detects floor-**too-low**; it would narrow a published contract with nothing measured behind it and break consumers pinned to 17.1/17.2 for no reason. Under AGENTS.md #0.1 that is the wrong direction.

Also checked: declared `@objectstack/spec` floors across the tree are `^17.0.0` (22 sites), `^17.1.0` (`plugin-detail`, `react`), `^17.2.0` (`data-objectstack`), `^17.3.0` (`core`), `^17.4.0` (`types`) — nothing here would have been disturbed either way, and the question is moot now that the manifest is untouched.

## Pins, and the ablation for each

All three pins **re-derive on every run**, never transcribe the sentence — a transcription pin only asserts that nobody edited the sentence, which is the failure mode these defects already are. They live in `relatedListParentScopeResidue-8937.test.ts`.

- **Pin one** reads the returned literal out of `composeParentScopeFilter` and requires the shipped README paragraph to name **exactly** the operator set the seam composes, in both directions, plus the equality spelling exactly when a bare equality arm still exists. Re-anchored this round: the literal used to live in `RelatedList.tsx`.
- **Pin two** reads the divergence list back out of the seam's header and re-derives it from the **installed** spec: each named type must be a real `FieldType`, must be in neither `MULTI_CAPABLE_TYPES` nor `MULTI_OPTION_TYPES`, and `isMultiValueField({ type, multiple: true })` must still answer `false`. A `CONTROL` case asserts the same predicate answers `true` for `lookup` / `user` / `multiselect`, so the divergence cases cannot pass vacuously. Re-anchored this round for the same reason. ⚠️ This is the pin that will move: a settled objectstack#17469 most likely admits these types to the spec's multi-capable set, at which point the list stops being a divergence and this file reddens — which is the point.
- **Pin three is NEW, and it is the one that would have caught this rework's cause.** Every tracked file under `packages` that names **both** the spec predicate and the driver must cite objectstack#17469, so the claim cannot be copied into a fresh file silently. A `DISCRIMINATOR` case proves the driver filter narrows rather than demanding the citation of every mention, and names the file that must stay on the excluded side (`plugin-grid`'s `multiValueFields.ts`, which relates the predicate to the server-side **write** pipeline — a true claim about a different consumer).
  ⚠️ Pin three was first built to grep for the retired **sentence** and that version is recorded here because it measured **degenerate**: the two corrections rephrase the claim across a line break, a line-oriented grep therefore misses them, and the only match left in the tree was the pattern string matching itself. A sweep whose sole carrier is its own pattern asserts nothing. Keying on the two **subjects** instead is what no rewording moves.

**Ablation, six legs, re-anchored.** Each leg: mutate, prove the bytes reached disk by blob-hash change, run, restore with `git checkout HEAD -- PATH`, prove restoration by **state** (`git diff HEAD` empty **and** hash back to the HEAD blob). Exit codes are evidence for neither direction, and the harness carries a trap on EXIT/INT/TERM calling a restore function with absolute paths. Run from the committed state, so each restore leg has a real restore point.

| leg | status | mutation | target | blob | result |
|---|---|---|---|---|---|
| 1 | unchanged | README reverted to the pre-fix sentence | `plugin-detail/README.md` | `a5971ee` → `9ee32ce` | **RED** — pin one, 2 failed / 6 passed |
| 2 | **re-anchored** | the seam stops composing `$contains` | `core/src/utils/parent-scope.ts` | `ce8bb58` → `2fbfc88` | **RED** — pin one, 2 failed / 6 passed |
| 3 | **re-anchored** | the seam's divergence paragraph deleted | `core/src/utils/parent-scope.ts` | `ce8bb58` → `0e54a6d` | **RED** — pin two, 1 failed / 7 passed |
| 4 | **re-anchored** | the seam names `lookup`, a type the spec already calls multi-valued | `core/src/utils/parent-scope.ts` | `ce8bb58` → `3156412` | **RED** — pin two, 1 failed / 7 passed |
| 5 | **new** | an existing carrier drops the objectstack#17469 citation | `plugin-detail/src/RelatedList.tsx` | `66905f6` → `d501975` | **RED** — pin three, 1 failed / 7 passed |
| 6 | **new** | a fresh file starts relating the predicate to the driver, uncited | `plugin-grid/src/hooks/multiValueFields.ts` | `6ce74fc` → `aa0090d` | **RED** — pin three + discriminator, 2 failed / 6 passed |

Legs 2–4 pointed at `RelatedList.tsx` before the merge and would have stopped discriminating there; they were re-anchored onto the seam rather than dropped. All six restored; `git status --porcelain` empty afterwards.

## Second rework round — the gate this PR tripped, and the one it filed

`Test (shard 1/4)` was red on this branch's own omission, and it is the same failure mode as pin three one level up.

`scripts/markdown-test-inputs.mjs` is the ledger behind `ci.yml`'s "does this change need a full run" step: that step excludes markdown paths, so a test that READS markdown and is not in the ledger can be **skipped by a markdown-only pull request**. The new pin reads `packages/plugin-detail/README.md` and was not registered — someone could have edited that README and the pin would never have run. The gate caught it before it shipped.

**The declared set is measured, not inferred from the constants.** Two of the three pins name the README by constant; pin three enumerates `git grep -l isMultiValueField packages` and opens every hit, so the population is *derived* and could be wider than the constants suggest. Measured on this branch:

| | |
|---|---|
| paths the enumeration offers | 18 |
| of those, CHANGELOGs dropped before anything is opened | 10 |
| paths actually opened | 8 |
| **markdown documents in the read set** | **1 — `packages/plugin-detail/README.md`**, the same README the other two pins name |
| root `CHANGELOG.md` | never in the population: the pathspec is `packages`; the scanner resolved it only from the bare literal in the drop filter |

So the entry declares `reads: ['packages/plugin-detail/README.md']` and records both CHANGELOG candidates as `notRead` rather than dropping them. No `walker` is declared: `auditTree` raises `stale-walker` when one is declared and the scanner sees no directory walk, and it sees none here.

**Red, then green.**

| reading | before | after |
|---|---|---|
| `node scripts/markdown-test-inputs.mjs --audit` | exit 1 — `unadjudicated-test: …relatedListParentScopeResidue-8937.test.ts reads markdown the ledger does not describe (CHANGELOG.md, packages/plugin-detail/CHANGELOG.md, packages/plugin-detail/README.md)` | exit 0 — `47 candidate test files, all adjudicated; 41 declared entries, all present` |
| `vitest run scripts/__tests__/markdown-test-inputs.test.ts` | exit 1 — `Tests 1 failed \| 11 passed (12)` | exit 0 — `Tests 12 passed (12)` |
| `--changed` on `packages/plugin-detail/README.md` | — | prints it: a markdown-only edit now forces the run |
| `--changed` CONTROL on `packages/plugin-detail/CHANGELOG.md` | — | prints nothing, exit 0 — the other side of the answer |

### The second annotation on that shard: root-caused, ⛔ not called a flake

```
scripts/check-i18n-dead-keys.mjs:616
grep: …/.objectui-tmp/tsc-gate-3853-1W87E5/tsconfig.json: No such file or directory
```

Consumer: `check-i18n-dead-keys.mjs` runs a whole-tree filesystem `grep -rFn` (lines 606-616) whose `TEXT_SWEEP_SKIP_DIRS` (lines 428-431) is `node_modules`, `.git`, `dist`, `build`, `coverage`, `.next`, `.turbo`, `.changeset` — `.objectui-tmp` is **not** in it (`grep -c` returns 0). Producer: `packages/cli/src/__tests__/app-generator.test.ts`'s `withGeneratedApp` (lines 1386-1396) `mkdtempSync`s `REPO_ROOT/.objectui-tmp/tsc-gate-3853-…`, writes a generated app including a `tsconfig.json`, and removes it in a `finally` — so it lives for one test inside the same shard. Convention control, lit: the sibling whole-tree sweep `check-comment-mask-corpus.mjs` line 246 already lists `.objectui-tmp` in its skip set.

⇒ a missing exclusion, not a flake: only *whether* the two overlap varies; given overlap the outcome is determined. **This branch does not contribute** — the diff touches neither script, neither the CLI test nor `.gitignore`; the new test shells out to `git grep`, which reads tracked files only, and `.objectui-tmp` is `.gitignore`d (`.gitignore:50`) hence untracked, so it is never opened by it, and a read cannot create a directory. The error also names a different grep in a different process. Filed as objectui#9201 after a dedup search (see below); ⛔ nothing here skipped, disabled or quarantined.

**Dedup channel, declared.** The container's REST `/search/issues` answers `403` with `sessions are bound to their configured repositories` — the proxy's path whitelist, ⛔ not a statement about GitHub scope — so the dedup went through one targeted MCP `search_issues` instead. Nine hits, all about how that script classifies keys, none about the sweep racing a temp directory. The control query in the same session returned this card, objectui#8937, as its second hit, so the channel was alive when the dedup ran.

## Verification

Run from the repo root with paths relative to it, per AGENTS.md; every exit code captured by redirecting to a file first, never through a pipe. All of this is post-merge; the ledger row is the second rework round.

| what | command | exit | reading |
|---|---|---|---|
| ① dependency-closure build | `turbo run build --filter=@object-ui/plugin-detail --filter=@object-ui/core --concurrency=2`, under the shared verify lock | 0 | `Tasks: 12 successful, 12 total`; lock printed `VERDICT command-exit 0` |
| ② both affected packages + the cross-package source reader | `pnpm exec vitest run packages/plugin-detail/ packages/core/ packages/i18n/src/__tests__/residue-namespaces-3546.test.tsx`, under the lock | 0 | `Test Files  323 passed (323)` · `Tests  4848 passed (4848)`; lock printed `VERDICT command-exit 0` |
| ② type-check | `pnpm --filter @object-ui/plugin-detail type-check` · `... @object-ui/core ...` | 0 · 0 | both logs echo their script name, so neither was a zero-match filter exiting 0 in silence. ⚠️ Both exit **2** against an unbuilt tree (`TS6305` / `TS2307`) — a prerequisite, not a red; they pass once ① has run |
| ② lint | `pnpm --filter @object-ui/plugin-detail lint` · `... @object-ui/core ...` | 0 · 0 | `997 problems (0 errors)` and `542 problems (0 errors)`, all pre-existing; the new test file contributes none |
| ③ changeset trio | `check-changeset-presence` · `check-changeset-no-major` · `check-changeset-claims` | 0 · 0 · 0 | presence: "4 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)" |
| ③ control bytes | `node scripts/check-control-bytes.mjs`, plus a `grep -naP` sweep over the touched files | 0 · 1 | gate OK over 7393 tracked files; the grep found none (its exit 1 is the empty result) |
| ③ new cross-file line citations | `node scripts/check-new-cross-file-line-citations.mjs` | 0 | no new citation added — the corrected comments name files, never a file plus a line number |
| ③ comment-mask corpus | `node scripts/check-comment-mask-corpus.mjs` | 0 | 4847 files; the single disagreement is the pre-existing `apps/console/src/pages/DocsIndex.tsx` residue objectui#7882 holds open, not this diff |
| ③ doc fences · doc links | `check-doc-fence-languages.mjs` · `check-doc-links.mjs` | 0 · 0 | fences judged over 227 documents |
| ③ spec floors, scoped | the gate's own `analyze()` over both packages, after ① | 0 | `findings: []`; the `^17.0.0` control is RED |
| ③ governed surface | `check-governed-queue-guard.mjs --test` over the six changed paths | 0 | NOT GOVERNED — 6 paths checked against 5 surfaces, none matched |
| **NOT MEASURED** | `pnpm check:readme-exports` | 1 | **a prerequisite failure, ⛔ not a red.** It collapses its own population and says so: `packagesRead: found 13, floor is 25`, `24 unbuilt`. It owns a whole-workspace build step inside `.github/workflows/readme-exports.yml`, and objectui#4846 rejected a per-PR full-repo build. |
| owned by CI | `pnpm lint` (turbo, every package) · `pnpm test` (full) · `check:spec-floors` on its own workflow and the release path | — | repo-wide scans; not narrowed here |

⚠️ This repository has no `scripts/pm/dispatch-gates.mjs` — that script is objectstack's, and pointing it at this tree would return a confident answer about the wrong repository. The list above was derived by hand from this repo's root `package.json` `check:*` scripts and `.github/workflows/`.

## Merge, not rebase

`git merge origin/main` at `759606e1d`, one conflict (`packages/plugin-detail/src/RelatedList.tsx`, the docblock both sides edited), resolved by taking origin/main's pointer comment as the base — the function it described has moved — and correcting its one false clause in place. Landing of objectui#9184 was confirmed **by content**, not by sha: `git cat-file -e origin/main:packages/core/src/utils/parent-scope.ts` exits 0, with `RelatedList.tsx` as a lit positive control. ⛔ No rebase, ⛔ no force-push, at any point.

## Status

⛔ Draft, ⛔ not enqueued, ⛔ no auto-merge — the lane arms it. Clause-② stays `yes`; `needs:contract-review` is attached and stays until the lane clears it. `packages/plugin-detail/package.json` is untouched. `packages/plugin-detail/src/renderers/record-reference-rail.tsx` is untouched (objectui#8883 is live on it).

## Out of scope, noted

⛔ No new card is opened for the arity divergence — objectstack#17469 already owns it and is open.

Generated by Claude Code; this round's session is `session_01UzHd6hDYatoDn17BuwKxnZ` (recorded as prose, because a link-form footer does not survive a body edit).

---

### DIFF
--- a/.changeset/plugin-detail-8937-parent-scope-residue.md
+++ b/.changeset/plugin-detail-8937-parent-scope-residue.md
@@ -1,0 +1,25 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+docs(plugin-detail): correct the shipped README's parent-scope claim and the false driver-arity docblock (objectui#8937)
+
+Two published texts that objectui#8886 left behind, both measured false on `origin/main`:
+
+- **`README.md` (it is in `files[]`, so it ships).** It said the node's `filter` is
+  AND-combined with `{ [relationshipField]: parentId }`, full stop. Since objectui#7299
+  the parent condition is compiled to match the relationship field's arity, so a
+  multi-valued relationship gets `{ [relationshipField]: { $contains: parentId } }`
+  instead. The paragraph now states both spellings and names the arbiter
+  (`@objectstack/spec/data`'s `isMultiValueField`).
+- **`parentRelationshipFieldDef`'s docblock, and the same sentence in the
+  objectui#7299 test header.** They said the SQL driver decides arity on the spec's
+  `isMultiValueField`. It does not: `driver-sql` gates the equality family on its own
+  storage question, which reads `multiple` as truthy on ANY type. The two rules
+  therefore disagree for a type outside `MULTI_CAPABLE_TYPES` carrying
+  `multiple: true`, and that divergence is now recorded as a divergence and pointed at
+  the upstream card that owns it (objectstack#17469).
+
+No predicate moved and no wire changed — this release carries corrected published text
+only. Both claims are now pinned by re-derivation rather than transcription
+(`relatedListParentScopeResidue-8937.test.ts`).
--- a/.changeset/plugin-detail-8937-parent-scope-residue.md
+++ b/.changeset/plugin-detail-8937-parent-scope-residue.md
@@ -1,25 +1,28 @@
 ---
 '@object-ui/plugin-detail': patch
+'@object-ui/core': patch
 ---
 
-docs(plugin-detail): correct the shipped README's parent-scope claim and the false driver-arity docblock (objectui#8937)
+docs(parent-scope): state the driver's real arity rule, and the arity-dependent parent scope in the shipped README (objectui#8937)
 
 Two published texts that objectui#8886 left behind, both measured false on `origin/main`:
 
-- **`README.md` (it is in `files[]`, so it ships).** It said the node's `filter` is
-  AND-combined with `{ [relationshipField]: parentId }`, full stop. Since objectui#7299
-  the parent condition is compiled to match the relationship field's arity, so a
-  multi-valued relationship gets `{ [relationshipField]: { $contains: parentId } }`
-  instead. The paragraph now states both spellings and names the arbiter
-  (`@objectstack/spec/data`'s `isMultiValueField`).
-- **`parentRelationshipFieldDef`'s docblock, and the same sentence in the
-  objectui#7299 test header.** They said the SQL driver decides arity on the spec's
-  `isMultiValueField`. It does not: `driver-sql` gates the equality family on its own
-  storage question, which reads `multiple` as truthy on ANY type. The two rules
-  therefore disagree for a type outside `MULTI_CAPABLE_TYPES` carrying
-  `multiple: true`, and that divergence is now recorded as a divergence and pointed at
-  the upstream card that owns it (objectstack#17469).
+- **`packages/plugin-detail/README.md` (it is in `files[]`, so it ships).** It said the
+  node's `filter` is AND-combined with `{ [relationshipField]: parentId }`, full stop.
+  Since objectui#7299 the parent condition is compiled to match the relationship field's
+  arity, so a multi-valued relationship gets
+  `{ [relationshipField]: { $contains: parentId } }` instead. The paragraph now states
+  both spellings and names the arbiter (`@objectstack/spec/data`'s `isMultiValueField`).
+- **The claim that the SQL driver decides arity on that same predicate.** It does not:
+  `driver-sql` gates the equality family on its own storage question, which reads
+  `multiple` as truthy on ANY type. The two rules therefore disagree for a type outside
+  `MULTI_CAPABLE_TYPES` carrying `multiple: true`. objectui#9184 moved the arity compiler
+  into `@object-ui/core`'s `parent-scope` seam and carried the claim with it, so the
+  correction is recorded there — the seam now states the driver's measured rule, records
+  the divergence as a divergence, and points at the upstream card that owns which of the
+  two rules is right (objectstack#17469). `RelatedList.tsx`'s pointer comment and the
+  objectui#7299 test header carried the same sentence and are corrected to match.
 
 No predicate moved and no wire changed — this release carries corrected published text
-only. Both claims are now pinned by re-derivation rather than transcription
-(`relatedListParentScopeResidue-8937.test.ts`).
+only. All three claims are pinned by re-derivation rather than transcription, in
+`relatedListParentScopeResidue-8937.test.ts`.
--- a/packages/core/src/utils/parent-scope.ts
+++ b/packages/core/src/utils/parent-scope.ts
@@ -23,11 +23,11 @@
  *
  * ## The rule itself is not invented here
  *
- * The arity verdict is `@objectstack/spec/data`'s own `isMultiValueField`, the
- * same predicate the driver that executes the query decides on. That matters
- * more here than anywhere: this function chooses `$contains` vs `=`, the driver
- * chooses whether to accept it, and two readers of one question disagreeing is
- * the entire defect class. The spec's rule is BROADER than an eyeballed
+ * The arity verdict is `@objectstack/spec/data`'s own `isMultiValueField`. That
+ * matters more here than anywhere: this function chooses `$contains` vs `=` and
+ * every surface that asks the question now reads that one choice, so two
+ * readers of one question disagreeing — the entire defect class — cannot
+ * recur INSIDE this repository. The spec's rule is BROADER than an eyeballed
  * `multiple === true` in both directions — `multiselect` / `checkboxes` /
  * `tags` persist an array with no flag at all, and `multiple: true` is INERT on
  * a type outside the spec's multi-capable set (`master_detail`, say) — so a
@@ -36,6 +36,36 @@
  * ⛔ Do not add a local arity rule at any call site, however small, and ⛔ do
  * not widen this function to accept an arity the caller computed: the parameter
  * it takes is METADATA, and the verdict is drawn from it here.
+ *
+ * ## ⚠️ The STORAGE side does not read this predicate — the two rules DIVERGE
+ *
+ * This header used to say `isMultiValueField` is "the same predicate the driver
+ * that executes the query decides on". It is not, and the difference is
+ * observable (objectui#8937). Measured on objectstack `origin/main`,
+ * `driver-sql` gates the equality family on its own STORAGE question,
+ * `isJsonField`: a column is JSON when the field's type is in that driver's
+ * `JSON_COLUMN_TYPES` — the spec's `STRUCTURED_JSON_TYPES` and
+ * `MULTI_OPTION_TYPES`, plus the driver-internal `object` / `array` aliases —
+ * OR when `multiple` is merely TRUTHY, on ANY type; a single-value media type
+ * answers from the ADR-0104 dual-encoding window instead. The spec's predicate
+ * is `MULTI_OPTION_TYPES.has(type) || (MULTI_CAPABLE_TYPES.has(type) &&
+ * multiple === true)`.
+ *
+ * ⇒ They diverge for a type OUTSIDE `MULTI_CAPABLE_TYPES` carrying
+ * `multiple: true` (`master_detail` / `tree` / `text`): the spec says
+ * single-valued, so this seam compiles `=`, while the driver stores a JSON
+ * column and refuses `=` with the same `400 INVALID_FILTER` objectui#7299 was
+ * filed for. That is a KNOWN divergence and ⛔ not a regression — both
+ * surfaces sent `=` for that shape before objectui#7299 and objectui#8882 too.
+ *
+ * ⛔ Do NOT close it by widening the predicate here: which of the two rules is
+ * right is an upstream question, filed as objectstack#17469 (enforce-or-align),
+ * and this seam is deliberately not blocked on it. Widening on this side would
+ * move the disagreement across the wire rather than end it, and would make the
+ * ONE compiler this module exists to be disagree with the spec it delegates to.
+ * The divergence is re-derived from the installed spec on every run by
+ * `relatedListParentScopeResidue-8937.test.ts`, so this section reddens when the
+ * spec side moves.
  */
 
 import { isMultiValueField, type ValueShapeFieldDef } from '@objectstack/spec/data';
--- a/packages/plugin-detail/README.md
+++ b/packages/plugin-detail/README.md
@@ -329,9 +329,16 @@ filter can express).
 
 The node's `filter` (spec `RecordRelatedListProps.filter`, "additional filter
 criteria") narrows the list beyond the parent relationship: it is
-**AND-combined** with `{ [relationshipField]: parentId }`, never substituted for
+**AND-combined** with the parent-relationship condition, never substituted for
 it, so a related list stays scoped to the record it appears on and an additional
-criterion can only ever narrow that set. Authors write it in the spec's own
+criterion can only ever narrow that set. That condition is **not one fixed
+spelling** — it is compiled to match the relationship field's ARITY on the
+child object (`objectui#7299`). A single-valued relationship gets the equality
+form `{ [relationshipField]: parentId }`; a multi-valued one gets the membership
+form `{ [relationshipField]: { $contains: parentId } }`, because the stored
+value is an array of ids and equality would ask whether the whole array *is* one
+id. The arity verdict is `@objectstack/spec/data`'s own `isMultiValueField`,
+never a rule this package keeps locally. Authors write `filter` in the spec's own
 vocabulary (`[{ field, operator, value }]`); a `dataSource` binding's composed
 filter (component AND saved view AND binding) lands on the same key. Both are
 lowered to ObjectQL through the repo's single filter sink, so no second dialect
--- a/packages/plugin-detail/src/RelatedList.tsx
+++ b/packages/plugin-detail/src/RelatedList.tsx
@@ -365,13 +365,40 @@ export const RelatedToolbarButton: React.FC<{
  * as a default.
  *
  * ⛔ Do not reintroduce a local arity rule here, however small. This component
- * decides `$contains` vs `=` on the answer, and the driver that refuses the
- * query decides on the spec's — two readers of one question, disagreeing, is
- * exactly the defect objectui#7299 is about, and putting it one layer up would
- * be a worse version of it. The spec's rule is BROADER than an eyeballed
- * `multiple === true` in both directions: `multiselect` / `checkboxes` / `tags`
- * persist an array with no flag at all, and `multiple: true` is INERT on a type
- * outside `MULTI_CAPABLE_TYPES` (`master_detail`, say). Both are pinned.
+ * decides `$contains` vs `=` on the answer, and a second copy of that rule
+ * living here would be two readers of one question inside one package —
+ * exactly the defect objectui#7299 is about, rebuilt one layer up. The spec's
+ * rule is BROADER than an eyeballed `multiple === true` in both directions:
+ * `multiselect` / `checkboxes` / `tags` persist an array with no flag at all,
+ * and `multiple: true` is INERT on a type outside `MULTI_CAPABLE_TYPES`
+ * (`master_detail`, say). Both are pinned.
+ *
+ * ## The storage side does NOT read this predicate — the two rules DIVERGE
+ *
+ * This docblock used to say the driver that refuses the query decides on the
+ * spec's `isMultiValueField`. It does not (objectui#8937). Measured on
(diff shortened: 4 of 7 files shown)


os-tesla:
## PM review — content ACCEPTED; ⛔ CI is RED and the failure is this PR's own

Read against `git merge-base origin/main <head>` = `759606e1d`: **6 files, +399/−24**.

**Accepted:** the sync is a real merge (`e7c4d2fca Merge origin/main into claude/issue-8937-…`), ⛔ no rebase; `packages/plugin-detail/package.json` untouched, so F1 stays closed by measurement rather than by edit; Clause-② `--pair 9195` → both carriers agree.

⭐ The most valuable line in the report is the one admitting a failure: pin three was **first built degenerate** — a line-oriented grep for the retired sentence whose only surviving match was its own pattern string — and that was reported rather than shipped. It is independently corroborated: `git grep "same predicate the driver"` returns **zero** hits against another branch that demonstrably carries that exact claim, because the sentence wraps at `the / driver`. Keying on the two subjects is the right repair, and *"a sweep whose sole carrier is its own pattern asserts nothing"* is worth keeping.

## The failure — `Test (shard 1/4)`

```
scripts/__tests__/markdown-test-inputs.test.ts:257
  kind:   "unadjudicated-test"
  test:   "packages/plugin-detail/src/__tests__/relatedListParentScopeResidue-8937.test.ts"
  detail: "reads markdown the ledger does not describe (CHANGELOG.md,
           packages/plugin-detail/CHANGELOG.md, packages/plugin-detail/README.md)."
```

Verified here with lit controls: the new pin reads `packages/plugin-detail/README.md` (`:99`), it appears **0** times in `scripts/markdown-test-inputs.mjs`, while the registered entries at `:190` (`core/src/adapters/README.md`) and `:252-253` (the plugin-calendar README pin) are both present — so the ledger and the probe are working and the omission is real.

⭐ Worth naming because it is the same defect one level up: **a pin that reads markdown and is not in that registry can be skipped by a markdown-only PR.** Someone could edit `plugin-detail/README.md` and pin one would never run. That is precisely the class pin three exists to close, and the gate caught it before it shipped.

A second annotation in the same shard — `check-i18n-dead-keys.mjs:616`, a grep failing on a vanished `.objectui-tmp/tsc-gate-3853-…/tsconfig.json` — looks like a scratch-directory race naming another card's gate. ⛔ I am **not** ruling it a flake: "flake" is not a root cause. The push for the ledger entry re-runs the shard; that run is the reading.

## Coordination

PR objectstack-ai/objectui#9200 (card objectui#8883) is in flight on `record-reference-rail.tsx` and carries a fourth copy of the false driver claim — **pin three would redden on it**. That PR is being corrected to cite `objectstack#17469`, so the two land compatibly. ⛔ Neither seat touches the other's file.

Arm held until CI reads `pending: 0, failing: 0`.

---

os-tesla:
## Second rework ACCEPTED — both claims re-measured here

Head `d663a1e1d`.

**The ledger entry** (`scripts/markdown-test-inputs.mjs:274-279`) is present and its `reads` set is **derived, not copied off the file's constants** — which is what I asked for, since pin three enumerates `git grep -l isMultiValueField packages` and opens whatever it finds:

```js
reads: ['packages/plugin-detail/README.md'],
notRead: ['CHANGELOG.md', 'packages/plugin-detail/CHANGELOG.md'],
```

18 paths offered → 10 CHANGELOGs dropped before opening → 8 opened → **1 markdown in the read set**. Recording the two CHANGELOG candidates as `notRead` rather than dropping them is the right call: it makes the exclusion auditable instead of invisible. And the `--changed` readings close it properly — the README forces the run, the CHANGELOG control prints nothing.

**The i18n annotation is root-caused, not waved off**, and I verified the load-bearing half myself:

| reading | value |
|---|---|
| `.objectui-tmp` in `scripts/check-i18n-dead-keys.mjs` | **0 — absent from `TEXT_SWEEP_SKIP_DIRS`** |
| `.objectui-tmp` in `scripts/check-comment-mask-corpus.mjs` (CONTROL) | 2 — prose at `:238`, the skip-list entry at `:246` |

⇒ the exclusion is an established convention in this tree that **exactly one of the two whole-tree sweeps is missing**. That is a defect with a name, not a flake, and the card's own argument for why is the right one: *"the interleaving is what varies; the defect does not."* Filed as objectstack-ai/objectui#9201 with both halves measured, a lit convention control, and an honest scope limit (it does not claim to have surveyed other sweeps). ⭐ Nothing was skipped, disabled or quarantined to get green — which was the instruction and is the part that matters.

The self-report that the first suite invocation named a nonexistent file and the vitest guard **refused** rather than passing on a zero-match path is worth keeping too: that guard (objectui#3288) is the reason a mistyped path did not read as a green run.

⛔ Arm still held: 36 checks, **7 in progress, 0 failing**. Not-red-yet is not green. I arm on `pending: 0`.

---

PR #9199, opened 2026-09-11
feat(components): give `alert-dialog`'s confirm button a destructive variant (objectui#8978)

### LINKED ISSUE #8978: capability(components): the destructive (red) confirm button on `alert-dialog` — batch #70 wanted it, `confirmVariant` was never a vehicle for it
Split out of objectui#7963 by maintainer ruling, 2026-09-10. That card **retires** the key `confirmVariant`; this card carries the **capability** it was supposed to deliver, so the capability is not lost with the key.

## Why the two are separate

The capability was asked for and granted first, on measurement that did not yet exist:

- **2026-09-07, director seat, decision batch #70** ([`5565626809`](https://github.com/objectstack-ai/objectui/issues/7963#issuecomment-5565626809)) — maintainer verbatim 「同意」. Recorded reason: *"a footer button variant (the red destructive confirm) is a capability this product wants; our own component catalog authored it three times and had to drop it."* The ruling's instruction was **wire `confirmVariant`**.
- **2026-09-09**, the probe landed by PR #8736 ([`5595380224`](https://github.com/objectstack-ai/objectui/issues/7963#issuecomment-5595380224)) measured all three footer keys **dead**: authoring `confirmVariant` leaves the dialog HTML byte-identical, leaves the confirm button's class at `bg-primary/text-primary-foreground` (**never** `bg-destructive`), and the value reaches **no DOM node at all**. Mechanism measured, not argued: `AlertDialogPrimitive.Root` forwards to `DialogPrimitive.Root`, which destructures only `__scopeDialog / children / open / defaultOpen / onOpenChange / modal` and renders a **context provider, not an element** ⇒ an unknown prop dies before any DOM exists.

⇒ batch #70 asked to *wire* a key on the assumption it was a key that could carry the value. It cannot, and never could. **The want is still good; the vehicle was never real.** objectui#7963 removes the dead vehicle; this card keeps the want.

## The capability is close — measured on `origin/main` `2c208d5bb`

`packages/components/src/ui/alert-dialog.tsx`:

```tsx
// :15
import { buttonVariants } from "./button"
…
// :115  — the CONFIRM button
    className={cn(buttonVariants(), className)}
…
// :128  — the CANCEL button, right below it
      buttonVariants({ variant: "outline" }),
```

`buttonVariants` **already takes a variant**, and the Cancel button already passes one. The confirm button calls it with **no argument**, so it is pinned to the default variant. And the renderer never offers one — `packages/components/src/renderers/overlay/alert-dialog.tsx:38`:

```tsx
{schema.actionText && <AlertDialogAction onClick={schema.onAction}>{schema.actionText}</AlertDialogAction>}
```

⇒ three things are missing, none of them hard: the primitive hardcodes `buttonVariants()`; the renderer passes nothing; and there is no authorable key.

⚠️ **NOT MEASURED**, owed by whoever takes this: which `variant` values `buttonVariants` actually admits; whether `destructive` is among them; what the "authored it three times and had to drop it" catalog sites were and what they wanted; whether changing `AlertDialogAction`'s default has a blast radius across existing dialogs.

## ⛔ Read this before choosing a key name

**objectui#7963 is retiring `confirmVariant` with a named tombstone.** Re-introducing that same spelling here would be a retire-then-re-add cycle on a published key — visible churn, and churn on the authoring surface is exactly what the maintainer flagged on 2026-09-10 (「协议不应该改来改去啊，否则元数据应用怎么办」).

That is not a veto, but it is a constraint the implementer must answer **before** writing a type:

- Is the right shape the old spelling, wired and pinned this time (so the tombstone becomes a "retired, then delivered properly" story)?
- Or does this capability belong to an existing idiom the repo already has for button variants, so no new authorable key is added at all?
- Whatever is chosen, it ships with a **behaviour pin that has a firing control** — the reason `confirmVariant` survived declared-and-dead for so long is that nothing ever asserted it moved the DOM.

⛔ Do not resurrect the spelling on the grounds that batch #70 named it. Batch #70 named it before anyone knew it was inert.

## Scope note

`alert-dialog` is **objectui-local**: it is not a governed `@objectstack/spec` component (probed at objectstack HEAD `eabdd66f45f402eba0f8404a8a9de4a501fc83a6` — `alert-dialog` / `AlertDialogProps` return **zero** hits under `packages/spec/src`, with `'record:chatter'` and `'object-metric'` as firing controls and a nonexistent name as the negative control). So the 「以 objectstack 协议为准」 principle does not constrain the shape here — the choice is objectui's own.

Refs objectui#7963 · objectui#8736 · objectui#4631 (the ADR-0049 liveness workbook this family routes through).

---

os-tesla:
Fixes #8978

`alert-dialog` can paint the red destructive confirm. `AlertDialogSchema` declares
`actionVariant?: 'default' | 'destructive'`, the renderer reads it onto the confirm
button, and a DOM pin with a firing control holds it there.

---

## The shape, and the argument against the tombstone

R2 required this to be argued rather than assumed, so here it is in full.

**The spelling is `actionVariant`, ⛔ not `confirmVariant`.** The retired key stays
retired: a document that reds on `confirmVariant` today still reds on it, with the
same `invalid_type` code. Re-adding a published spelling that refuses an author
today and accepts one tomorrow is precisely
「协议不应该改来改去啊，否则元数据应用怎么办」 (maintainer, 2026-09-10). What did move
is one published string — that tombstone's MESSAGE, from "it has NO surviving
spelling … a separate question that needs its own card" to naming `actionVariant`.
This card **is** that card, so the message was stale the moment it landed here.

`actionVariant` is not a new dialect. This node already calls that button `action`
in three places — `actionText`, `onAction`, `AlertDialogAction` — and the same
retirement pointed `confirmLabel` at `actionText` for exactly that reason. Using
`confirm*` for the variant while `confirm*` for the label is tombstoned would have
published two prefixes for one button.

**The value domain is two, ⛔ not `ButtonSchema.variant`'s six — by measurement.**
See M1/M5 below. Declaring a value this node cannot actually render would be the
`confirmVariant` disease one level down, at the value instead of the key, and the
pin is built so that widening the union without widening the mechanism reds.

---

## M1 — which variants does `buttonVariants` admit, and is `destructive` among them

Printed from `packages/components/src/ui/button.tsx`, not assumed. Six:
`default` `destructive` `outline` `secondary` `ghost` `link`, plus a `size` axis,
with `defaultVariants: { variant: 'default', size: 'default' }`. `destructive`
emits `bg-destructive text-destructive-foreground hover:bg-destructive/90`; the
`default` the confirm button is pinned to emits
`bg-primary text-primary-foreground hover:bg-primary/90`. ⇒ yes, `destructive`
is among them, and `buttonVariants()` === `buttonVariants({ variant: undefined })`
byte-for-byte (measured), which is what makes the wiring free of blast radius.

⭐ The pin never types those class names in. It computes the expected token set
from `buttonVariants` at run time, so an upstream rename follows instead of going
stale, and an `ORACLE` control fails if the two variants ever stop differing.

## M2 — blast radius of `AlertDialogAction`'s default

**The default is NOT changed, so the radius is zero — and that is a DOM reading,
not an argument.** Every `AlertDialogAction` call site in the tree, enumerated by
`git grep`: `apps/console/.../ApprovalsInboxPage.tsx` (4),
`packages/app-shell/.../InvitationsPage.tsx`, `MembersPage.tsx`, `SettingsPage.tsx`
(2), `views/ActionConfirmDialog.tsx`, `packages/components/.../NotificationAlerts.tsx`
(2), `packages/plugin-detail/src/RelatedList.tsx`,
`packages/plugin-form/src/DrawerForm.tsx`, `ModalForm.tsx`,
`packages/plugin-gantt/src/ObjectGantt.tsx`, and the SDUI renderer. Every one of
them passes no variant. Under this change the renderer passes `className=undefined`
when the key is unauthored, and `cn(buttonVariants(), undefined)` is byte-identical
to `cn(buttonVariants())` — asserted as the pin's `UNTOUCHED` control, which
compares two whole rendered dialogs.

⚠️ A finding the card did not anticipate: the card's suggested route — edit
`packages/components/src/ui/alert-dialog.tsx` to pass a variant — is **forbidden by
AGENTS.md Commandment #7**, a No-Touch zone whose files are regenerated by
`pnpm shadcn:update`, and `alert-dialog` is one of the tracked components
(`pnpm shadcn:list`). The sanctioned escape hatch is a declared patch in
`scripts/shadcn-local-patches.mjs`, whose own header says to keep the payload out
of `src/ui/`. So the variant is applied as a className OVERRIDE from the renderer —
which is not an invention: `NotificationAlerts.tsx` already expresses a footer
variant on this very button the same way, with a comment saying why.

## M3 — the "authored it three times and had to drop it" catalog sites

Found. Commit `ed5fee9be` (PR #7962) converted **four** fixtures under
`examples/schema-catalog/src/schemas/components-overlay-alert-dialog/` off a
phantom `actions` array, and its own message says "the per-button `variant` is
DROPPED". The dropped confirm-button variants:

| fixture | confirm label | dropped variant |
| --- | --- | --- |
| `basic-alert-dialog` | Continue | **destructive** |
| `confirmation-dialog` | Save | default |
| `custom-actions` | Yes, Continue | default |
| `destructive-action` | Delete | **destructive** |

⚠️ **The honest count is two, not three**, and I am not going to round it up. The
reading that yields exactly three is a different denominator: the dropped variants
the primitives could not reproduce for free — two destructive confirms plus the
`ghost` **cancel** in `confirmation-dialog` (the two `outline` cancels and the two
`default` confirms are what the primitives already emit). That is a plausible
reconstruction of the ruling's sentence, ⛔ not a fact I measured it against. The
two destructive confirms are restored here; the `ghost` cancel is not, and would
need a `cancelVariant` key that nothing has asked for — noted below, not filed.

## M4 — the card's anchors, re-measured by TEXT

The card read `2c208d5bb`; this branch is based on `fe9e0d054`. Located by text,
never by line number, and all three are still exactly as the card describes:
`import { buttonVariants } from "./button"`, the confirm button's
`className={cn(buttonVariants(), className)}` with **no argument**, the cancel
button's `buttonVariants({ variant: "outline" })`, and the renderer's
`{schema.actionText && …AlertDialogAction onClick={schema.onAction}…}`. The
premise holds; only the line numbers moved.

## M5 — does an authorable key need to be added at all

**Yes, and the sweep is what says so.** The repo has a strong idiom for authoring a
button variant — a key spelled `variant`, typed at `buttonVariants`' domain, on the
node that *owns* the button: `ButtonSchema.variant` in `packages/types/src/form.ts`
read at `renderers/form/button.tsx`, plus the same spelling on the `action:*`
family, `badge`, `alert`, `toggle-group` and more (18 registrations declare a
`variant` input; 20-odd renderers read `schema.variant`).

⛔ But no existing idiom reaches this button. `AlertDialogSchema` has no `variant`
key and no footer slot; the footer is generated from two **strings**, and the
`content` slot renders above the footer and is not an `AlertDialogAction`. The
fixtures in M3 prove the point from the author's side: they reached for
`type: 'button'` nodes with `variant` — the idiom — and the renderer carried none
of it. So the idiom governs the **spelling and value domain**, not whether a key is
needed.

And then the value domain, measured through the real renderer rather than inherited
from the idiom:

| value | override lands | declared |
| --- | --- | --- |
| `default` | clean | ✅ |
| `destructive` | clean | ✅ |
| `secondary` | clean | ⛔ nothing asked for it; widening later is non-breaking, narrowing is not |
| `outline` | leaks `text-primary-foreground` | ⛔ |
| `ghost` | leaks `bg-primary text-primary-foreground` | ⛔ |
| `link` | leaks `bg-primary hover:bg-primary/90` | ⛔ |

An override can only displace a baked-in class that shares its tailwind-merge
group, and those three set no background and/or no text colour, so the primitive's
own survives underneath. The `MECHANISM-` legs of the pin render each of the three
through the renderer and assert the leftovers are still on the node — so this table
is re-derived on every run, not recorded.

---

## The pin, and the ablation matrix

`packages/components/src/__tests__/alert-dialog-action-variant-8978.test.tsx`
renders through the real `SchemaRenderer` and the real registry and reads the
confirm button's `class` attribute. ⛔ No leg asserts that a prop was passed, that a
component received a value, or that a source file contains a spelling.

Ablation: the wiring line in the renderer replaced by the pre-fix
`{schema.actionText && …AlertDialogAction onClick=…}`, proven on disk by grep
before and after, restored by `git checkout HEAD --` and proven restored by an
empty `git diff HEAD`.

Ran twice, because the FIRST run caught a bad assertion of my own.

| leg | wired | ablated (run 1) | ablated (run 2, after the repair) |
| --- | --- | --- | --- |
| `WIRED` / `ORACLE` / `INSTRUMENT` / `UNTOUCHED` (4 controls) | PASS | PASS | PASS |
| LEG 1 — destructive confirm carries the destructive tokens | PASS | **FAIL** | **FAIL** |
| LEG 1b — and loses the tokens they displace | PASS | **FAIL** | **FAIL** |
| LEG 2 — default confirm carries neither, and is unchanged | PASS | PASS | PASS |
| LEG 3 — cancel untouched, confirm moved | PASS | **FAIL** | **FAIL** |
| `MECHANISM+` `default` | PASS | PASS | PASS |
| `MECHANISM+` `destructive` | PASS | **FAIL** | **FAIL** |
| `MECHANISM-` `outline` / `ghost` / `link` | PASS | ⚠️ PASS | **FAIL** ×3 |
| authoring-face closure (4 legs) | PASS | PASS | PASS |
| **totals** | 17 passed | 4 failed / 13 passed | **7 failed / 10 passed** |

⭐ **The `MECHANISM-` row is the finding.** With the wiring deleted those three
legs stayed GREEN, because each asserted only that the primitive's baked-in
classes were still on the node — and "still there" is trivially true of a
renderer that applies nothing at all. That is an assertion that cannot fail
through the very ablation that proves the feature, i.e. exactly the shape
`confirmVariant` got away with for months. ⛔ Not noted and left: each leg now
asserts ARRIVAL first (the tokens the variant itself emits ARE on the node) and
only then the loss, and all three red under ablation. Second commit on this
branch.

The legs that still stay green under ablation are the ones that do not name the
wiring: the four controls, the default path (unchanged by construction — which
is leg 2's whole job), and the authoring-face legs, which are schema facts.

On-disk proof, both runs: `HEAD` blob hash printed, anchor `grep -c` 1 before /
0 after, post-mutation blob hash differs from `HEAD`'s, restore by
`git checkout HEAD -- <absolute path>` from an `EXIT INT TERM` trap, restoration
proven by an empty `git diff HEAD`, not by an exit code.

## Acceptance notes

Found and deliberately NOT fixed, none of them filed:

- **`noted, not filed` — the `ghost` cancel variant.** `confirmation-dialog.json`
  authored `variant: 'ghost'` on its *cancel* button before PR #7962 stripped it.
  `AlertDialogCancel` hard-codes `outline`, and there is no `cancelVariant` key.
  This card is the destructive *confirm*; a cancel variant is a different
  capability with no ruling behind it, and the same override channel would carry
  neither `ghost` nor `link` anyway. Handler if it is ever wanted: whoever widens
  the override mechanism.
- **`noted, not filed` — `NotificationAlerts.tsx`'s `ALERT_ACTION_CLASSES`.** A
  hand-written map of override classes for this exact button, including an explicit
  `bg-transparent` to neutralise the baked-in default — i.e. the repo already knew
  about the leak measured in M5. It is internal React props, not authorable
  metadata, so it is not in any filing category; it is cited here because it is the
  precedent the renderer's approach follows. Handler: whoever widens the union.
- **`noted, not filed` — the card's `:15/:115/:128` line citations.** Drifted, as
  the card itself warned. Nothing to fix; recorded so M4's re-measurement is not
  read as a correction to the card.

## Verification

All at `4a617ceb8`.

- `pnpm --filter @object-ui/types test` — **177 files, 3460 tests, all passed**
- `pnpm --filter @object-ui/components test` — **265 files, 2430 tests, all passed**
- `vitest run examples/schema-catalog/` — **31 files, 2151 tests, all passed**
  (the two edited fixtures render through the real registry there)
- `pnpm --filter @object-ui/types --filter @object-ui/components run type-check` —
  **Done** on both, `tsconfig.test.json` included, which is what compiles the
  `@ts-expect-error` and `Equal` legs
- `pnpm --filter @object-ui/types --filter @object-ui/components run lint` —
  **0 errors** on both (279 / 953 pre-existing warnings), which is the whole
  population `turbo run lint` covers for the two packages this diff touches. The
  narrowing is measurable: type-aware linting is NOT enabled (the flat config
  extends `tseslint.configs.recommended`, and `languageOptions` carries only
  `ecmaVersion` + `globals` — no `project`, no `projectService`), so this diff
  cannot move a verdict on a file it does not contain.
- Gates run green: `changeset:check` · `check:changeset-claims` ·
  `check-changeset-presence` · `check:doc-types` · `check:doc-fences` ·
  `check:doc-example-ids` · `check:new-line-citations` · `check:handler-key-reads` ·
  `check:control-bytes` · `check:side-effects-array` · `check:unreferenced-sources` ·
  `check:spec-symbols` · `check:element-data-source-declaration` ·
  `check:prompt-keys` · `check:vi-mock-specifiers`
- ⚠️ **NOT MEASURED**, left to CI, all three for the same reason — they need build
  artefacts a whole-tree build produces and the container's foreground cap does
  not allow: `check:sdui-registration-pins` (exit 2, "No console build to weigh
  at apps/console/dist/assets — this is exit 2, not a pass"),
  `check:doc-snippets` and `check:doc-examples` (both exit 2 on the same
  prerequisite). ⛔ None of the three is recorded as green. The cheap sibling of
  the first, `check:side-effects-array`, IS green, and this diff adds no module
  and no registration — only an `inputs` entry on one that already existed.
- `pnpm shadcn:check` not run: it is a weekly scheduled workflow needing the
  network, and this diff deliberately touches no file under
  `packages/components/src/ui/`.

Two existing pins were re-derived rather than deleted, both on the other side of a
question they had answered honestly at the time: the `confirmVariant` refusal
message leg (it asserted the message names NO survivor; it now asserts the message
names `actionVariant`, while the sentence ruling the two LABEL keys out is kept and
still asserted), and the read-dialect pin's comment that said the block names no
substitute.

---

### DIFF
--- a/.changeset/8978-alert-dialog-action-variant.md
+++ b/.changeset/8978-alert-dialog-action-variant.md
@@ -1,0 +1,50 @@
+---
+'@object-ui/types': minor
+'@object-ui/components': minor
+---
+
+`alert-dialog` can finally paint the red destructive confirm: `AlertDialogSchema`
+declares `actionVariant?: 'default' | 'destructive'` and the renderer reads it
+onto the confirm button (objectui#8978, the capability decision batch #70
+granted on 2026-09-07).
+
+**Additive, and nothing that renders today moves.** No default changes, no key
+narrows, no existing document changes verdict. A document that does not author
+`actionVariant` renders a BYTE-IDENTICAL dialog — asserted as the `UNTOUCHED`
+control of the pin, not argued — which is also the whole blast-radius answer for
+the `AlertDialogAction` call sites across the repo: they pass no variant, so they
+get exactly what they got before.
+
+**Why the spelling is not `confirmVariant`.** That key is retired with a
+tombstone (objectui#7963) because it was measured inert: it reached no DOM node
+at all. Re-adding the same published spelling would be a retire-then-re-add cycle
+on the authoring surface — 「协议不应该改来改去啊，否则元数据应用怎么办」
+(maintainer, 2026-09-10) — so a document that reds on `confirmVariant` today
+still reds on it, with the same code. What moved is one published string: that
+tombstone's MESSAGE used to say the key had no replacement and now names
+`actionVariant`. `actionVariant` is the `action*` dialect this node already uses
+for that button (`actionText`, `onAction`, `AlertDialogAction`), and it is the
+same dialect the retirement itself pointed `confirmLabel` at.
+
+**Why two values and not `ButtonSchema.variant`'s six.** ⛔ Not deference — a
+measurement. `packages/components/src/ui/**` is a No-Touch zone (AGENTS.md
+Commandment #7) and `AlertDialogAction` bakes in `cn(buttonVariants(), className)`
+with no variant prop, so the renderer applies the variant as a className
+OVERRIDE, and an override can only displace a baked-in class that shares its
+tailwind-merge group. Rendered through the real renderer, `default` and
+`destructive` land clean; `outline`, `ghost` and `link` leave the primitive's own
+background and/or text colour visible underneath. Declaring a value this node
+cannot render is the `confirmVariant` disease one level down, at the value
+instead of the key — so the union is exactly what the channel carries, and the
+pin measures the admitted values AND the refused ones, so widening the union
+without widening the mechanism reds.
+
+**The pin has a firing control**, which is the point of the exercise:
+`packages/components/src/__tests__/alert-dialog-action-variant-8978.test.tsx`
+renders the dialog and reads the confirm button's own `class` off the DOM — never
+that a prop was passed — with the expected tokens computed FROM `buttonVariants`
+rather than typed in, so an upstream rename follows instead of going stale.
+
+The two schema-catalog fixtures whose confirm button authored `variant:
+"destructive"` before PR #7962 had to strip it (`basic-alert-dialog`,
+`destructive-action`) get it back, in the dialect the renderer reads.
--- a/content/docs/components/overlay/alert-dialog.mdx
+++ b/content/docs/components/overlay/alert-dialog.mdx
@@ -33,6 +33,7 @@ interface AlertDialogSchema {
   content?: SchemaNode | SchemaNode[];  // Rendered between the header and the footer
   cancelText?: string;                  // Cancel button label; no cancel button when omitted
   actionText?: string;                  // Confirm button label; no confirm button when omitted
+  actionVariant?: 'default' | 'destructive';  // Confirm button variant; red confirm when 'destructive'
 
   // Open state
   defaultOpen?: boolean;                // Initial state when uncontrolled (default: false)
@@ -44,6 +45,8 @@ interface AlertDialogSchema {
 ```
 
 The footer is driven by the two label keys: the cancel button renders only when `cancelText` is set and the confirm button only when `actionText` is set — neither has a renderer default (the designer palette seeds `Cancel` / `Continue`). The confirm button's click handler, `onAction`, is a runtime slot a React host supplies through the TypeScript interface; it has no JSON spelling, and the validator refuses it by name.
+
+The confirm button's look is `actionVariant`. Omit it and the button stays the default primary; write `'destructive'` for the red confirm a delete dialog wants. Those are the two values it takes — not the six a `button` node takes — because the renderer applies the variant as a class override on top of the primitive's built-in styling, and the remaining upstream variants set no background and/or no text colour, so the default's would show through. The cancel button is always `outline` and has no variant key. (The older spelling `confirmVariant` is retired and refused by name — it never reached the DOM.)
 
 ## Examples
 
--- a/examples/schema-catalog/src/schemas/components-overlay-alert-dialog/basic-alert-dialog.json
+++ b/examples/schema-catalog/src/schemas/components-overlay-alert-dialog/basic-alert-dialog.json
@@ -7,5 +7,6 @@
     "label": "Delete Account"
   },
   "cancelText": "Cancel",
-  "actionText": "Continue"
+  "actionText": "Continue",
+  "actionVariant": "destructive"
 }
--- a/examples/schema-catalog/src/schemas/components-overlay-alert-dialog/destructive-action.json
+++ b/examples/schema-catalog/src/schemas/components-overlay-alert-dialog/destructive-action.json
@@ -8,5 +8,6 @@
     "variant": "destructive"
   },
   "cancelText": "Cancel",
-  "actionText": "Delete"
+  "actionText": "Delete",
+  "actionVariant": "destructive"
 }
--- a/packages/components/src/__tests__/alert-dialog-action-variant-8978.test.tsx
+++ b/packages/components/src/__tests__/alert-dialog-action-variant-8978.test.tsx
@@ -1,0 +1,290 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8978 — `AlertDialogSchema.actionVariant` MOVES THE CONFIRM BUTTON'S
+ * CLASS. Measured here, through the real `SchemaRenderer` and the real
+ * registry, by reading the `class` attribute off the rendered confirm button.
+ *
+ * ## Why this file exists at all, and why it reads the DOM
+ *
+ * The key this one replaces, `confirmVariant`, survived DECLARED AND DEAD for
+ * months. Nothing ever asserted that it moved anything: it was declared, it
+ * parsed green under `.passthrough()`, it rode the renderer's rest-spread into
+ * the Radix root — and died there, because that root renders a context
+ * provider rather than an element. Every reading that would have caught it is a
+ * DOM reading, and there was none. (Its post-mortem is
+ * `alert-dialog-footer-keys-liveness-7963.test.tsx`, kept standing next to this
+ * file; that one asserts the retired key still moves NOTHING, this one asserts
+ * the new key moves the one thing its name promises.)
+ *
+ * ⛔ So: no leg here asserts that a prop was passed, that a component received a
+ * value, or that a source file contains a spelling. Every leg renders a dialog
+ * and reads an attribute off a node.
+ *
+ * ## The oracle is READ, never typed in
+ *
+ * ⭐ `bg-destructive` is nowhere in this file as an expectation. The classes a
+ * variant emits belong to `buttonVariants` (`../ui/button`), which is a synced
+ * upstream file that can change under us, so the expected token set is computed
+ * FROM `buttonVariants` at run time and diffed against the default's. If
+ * upstream renames the token, this file follows it; if upstream ever makes the
+ * two variants emit the same classes, `ORACLE` below reds rather than letting
+ * every reading pass vacuously.
+ *
+ * ## The controls, and what each one buys
+ *
+ *  - `WIRED` — the fixture, the registry and the renderer draw both footer
+ *    buttons at all, so a class reading has something to read.
+ *  - `ORACLE` — the computed expectation is non-empty: `destructive` really
+ *    does emit tokens `default` does not. Without it, "the button carries every
+ *    destructive token" is satisfied by the empty set.
+ *  - `INSTRUMENT` — inside ONE dialog the cancel button's class already differs
+ *    from the confirm button's, so this reading demonstrably separates one
+ *    button variant from another on this very DOM.
+ *  - `UNTOUCHED` — a document that does NOT author the key renders a dialog
+ *    BYTE-IDENTICAL to one that never could. This is the blast-radius reading:
+ *    every `AlertDialogAction` call site in the tree passes no variant, and this
+ *    says what they now get, rather than arguing it.
+ *
+ * ## The mechanism leg — why the union is two values and not six
+ *
+ * The renderer cannot pass a variant PROP: `packages/components/src/ui/**` is a
+ * No-Touch zone (AGENTS.md Commandment #7) and `AlertDialogAction` bakes in
+ * `cn(buttonVariants(), className)`. So the variant arrives as a className
+ * OVERRIDE, and an override can only displace a baked-in class that shares its
+ * tailwind-merge group. `MECHANISM` below measures that boundary live, over the
+ * variants the union DOES admit and the ones it does NOT, so the narrowness of
+ * the declared union is re-derived on every run instead of being a sentence in
+ * a docblock that was true once.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { cleanup, render } from '@testing-library/react';
+import React from 'react';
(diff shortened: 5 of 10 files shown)


os-tesla:
## PM review — ACCEPT, contract review PASS, armed SQUASH

Read against `git merge-base origin/main <head>` = `fe9e0d054`: **10 files, +523/−32**. CI **36/36, 0 pending, 0 failing**. `--pair 9199` → both carriers agree; `needs:contract-review` cleared from both.

**R2 — the tombstone holds, verified at source.** `confirmVariant` still reds with the same `invalid_type` code; the only published string that moved is the tombstone's **message**, which used to say the key had *"NO surviving spelling … a separate question that needs its own card"* — and this card **is** that card, so the sentence was stale on arrival. ⭐ The refusal pin was **re-derived, not deleted**: it now asserts the message names the remedy, keeps the `cancelText`/`actionText` ruling-out clause it always had, and gains a **CONTROL** proving `actionVariant` is a narrow enum rather than `.passthrough()` admitting anything — which is the difference between a reading and a vacuous `.success`.

**R5 — the brief's own suggested route was wrong, and the seat caught it.** My dispatch inherited the card's suggestion to edit `packages/components/src/ui/alert-dialog.tsx`. That file is a **No-Touch zone**: AGENTS.md Commandment #7, verified at `:107` — *"upstream 3rd-party files overwritten by sync scripts — never edit their logic/styles"*. Confirmed the diff touches **zero** files under that path. Routing the variant as a className override from the renderer, with `NotificationAlerts.tsx` as the in-repo precedent, is the right call and it was reached by measurement rather than by following the brief.

⭐ **Three things I want on the record, because each is a habit worth keeping:**

1. **The ablation caught the seat's own bad assertion.** Run 1 read `4 failed / 13 passed` and the three `MECHANISM-` legs stayed **green while the wiring was deleted** — they asserted only that the primitive's baked-in classes were *still* present, which is trivially true of a renderer that applies nothing. Fixed rather than noted (arrival asserted first, then the loss), and run 2 reads `7 failed / 10 passed`. That is precisely the shape `confirmVariant` got away with for months, caught by the instrument built to catch it.
2. **M3 refused to round two up to three.** The ruling's recorded reason says the catalog *"authored it three times"*; the measurement finds **two** dropped destructive confirms. The seat offers a plausible reconstruction of the third and labels it explicitly as a reconstruction, ⛔ not a measurement. Inheriting the ruling's number would have been easier and wrong.
3. **The value domain is two because two is what the channel carries**, measured through the real renderer: `outline` / `ghost` / `link` leak the primitive's own background or text colour because an override can only displace a class sharing its tailwind-merge group. Declaring a value the node cannot render would be the same disease one level down — and the `MECHANISM-` legs re-derive that table on every run, so widening the union without widening the mechanism reds.

**Contract review:** additive `minor` on `@object-ui/types` + `@object-ui/components`; no default changes, no narrowing; an unauthored document renders byte-identically, asserted by the `UNTOUCHED` control comparing two whole rendered dialogs rather than argued. The three `NOT MEASURED` gates are declared as exit-2 prerequisites and ⛔ none is recorded as green — correct.

---

PR #9200, opened 2026-09-11
fix(plugin-detail): compile the reference rail's parent scope by relationship arity

### LINKED ISSUE #8883: `record:reference_rail` cannot bind to a multi-value relationship field either — same bare-equality parent filter as objectui#7299, plus a "View all" link the URL grammar cannot express
Found while fixing objectui#7299 (dev seat, enumeration axis: who else compiles a parent-relationship condition). Filed separately rather than folded in, because it is a different block, a different authoring key and a different code path — objectui#7299's ruling scopes it to `RelatedList`.

## The measurement, on `origin/main`

`record:reference_rail` takes an author-supplied `relationshipField` per entry and compiles it as bare equality — twice, in two different grammars:

```
packages/plugin-detail/src/renderers/record-reference-rail.tsx:187
    $filter: { [entry.relationshipField]: parentId },

packages/plugin-detail/src/renderers/record-reference-rail.tsx:285
    to={`/apps/${appName}/${entry.objectName}?filter%5B${entry.relationshipField}%5D=...`}
```

That is objectui#7299's defect, on another block. When the named field is `multiple: true` — `Field.user({ multiple: true })` is the platform's own shape — the stored value is a JSON-valued array, `=` compares the whole serialized array against one id, and `driver-sql` refuses with `400 INVALID_FILTER` naming the field and prescribing `$contains`. The author is naming a relationship; nothing on the authoring surface lets them say "membership".

**Positive control for the negative result:** `git grep '\$contains' packages/plugin-detail/src/renderers/record-reference-rail.tsx` → 0 hits, with `relationshipField` at 4 hits in the same read. The absence is measured, not a probe failure.

## The second spelling is the harder half

`:187` is fixable the way objectui#7299 fixed `RelatedList`: read the child object's declared arity and compile `{ [field]: { $contains: parentId } }`.

`:285` is NOT. It builds a link into the ADR-0055 bare data surface, whose one operator contract is `drillUrlFilters`' `URL_FILTER_OPS` (objectui#1752) — `gte` / `lte` / `gt` / `lt` and nothing else — and whose parser DROPS an unrecognised operator suffix rather than downgrading it. So `filter[<field>][contains]=` would arrive as no condition at all, i.e. the "View all" link would open the child object UNSCOPED: every row of the table, presented as this parent's related records. That is strictly worse than the 400 and it is why this needs a decision rather than a patch — the same reasoning objectui#7299 recorded for `RelatedList`'s raw-URL fallback, where the answer was to refuse loudly instead.

So this card carries a real question: does the URL filter grammar grow a membership operator (a change to `URL_FILTER_OPS`, both sides, plus its pin), or does the rail suppress / re-target "View all" for a multi-value relationship?

`packages/plugin-form/src/LineItemsPanel.tsx:200` and `packages/plugin-form/src/MasterDetailForm.tsx:762` compile the same bare-equality parent scope. They are NOT asserted as defects on this card: their relationship is a `master_detail` FK, and whether that field type can legally carry `multiple: true` was not verified. Whoever takes this card should answer that first — if it can, they belong here; if it cannot, they are correct as written and the sweep ends at the rail.

---

os-tesla:
Fixes #8883

`record:reference_rail` compiled an author-supplied `relationshipField` as bare
equality **twice, in two different grammars** — the `$filter` it puts on the
wire and the `filter[FIELD]=VALUE` URL behind its "View All" link. On a
`multiple: true` relationship the stored value is an array, `=` asks whether
that whole array **is** one id, and `driver-sql` refuses with
`400 INVALID_FILTER` while prescribing `$contains`.

## What changed

**Half one — the `$filter`, repaired.** The condition is composed by
`composeParentScopeFilter` from `@object-ui/core`: the ONE compiler the related
list's rows (#7299) and its tab badge (#8882, PR #9184) already share, whose
arity verdict is `@objectstack/spec/data`'s own `isMultiValueField`. No local
arity rule was written at this call site, and no authoring key was added.

**Half two — the link, refused.** No URL spelling on this surface carries
membership, so the link is **suppressed** on a multi-value relationship and the
reason is logged once. Losing that affordance is the **cost** of the repair,
not its purpose — which is why it is a seat-level call while growing the URL
operator set is not.

Landing half one alone was explicitly out: it would show the right rows above a
link that goes somewhere else, a disagreement invisible until the user clicks.

## Measurements

**M1 — does the rail's `dataSource` expose `getObjectSchema`? YES, and it is
the same object the rows read it off.** `getObjectSchema` is a **required**
(non-optional) member of `DataSource` in `packages/types/src/data.ts:539`.
`record-related-list.tsx:248` hands `ctx?.dataSource` straight into
`RelatedList`, which reads `dataSource.getObjectSchema` at
`RelatedList.tsx:515` — the rail holds that same value at
`record-reference-rail.tsx:113`. `ObjectStackAdapter.getObjectSchema`
(`data-objectstack/src/index.ts:4660`) is cached per object name, so N rail
entries cost N cached metadata reads, not N round trips. Measured, not assumed
from the badge's result. Counter-case measured too: `PagePreview.tsx:255`
mounts `RecordContextProvider` with **no** `dataSource`, so the rail's fetch
effect early-returns there and nothing changes.

## Third round — the merge queue ejected this PR, and the cause was a second uncited file

The PR was green on its own head and red in the queue, which is the tell that
`main` moved under it. `1f4e02995` (objectui#8937 / PR #9195) landed a residue
sweep in
`packages/plugin-detail/src/__tests__/relatedListParentScopeResidue-8937.test.ts`:
every tracked file under `packages` naming BOTH the spec arity predicate and
the driver must cite `objectstack#17469`. Round two cleared the renderer. The
sweep's population is the CHANGED SET, and this PR's **test** file was in it
too.

Reproduced locally rather than inferred — `origin/main` merged in, then the pin
run on the combined tree **before** any edit:

```
× cites objectstack#17469 in every file that relates the two sides
AssertionError: a file relates the spec predicate to the driver without citing
the card that owns their divergence: expected [ Array(1) ] to deeply equal []
Tests  1 failed | 7 passed (8)
```

Readings on the test file, and the population re-derived with the sweep's own
enumeration (`git grep -l isMultiValueField packages`, minus `CHANGELOG.md`,
split by `/driver/i`):

| reading on `record-reference-rail.parentScopeArity-8883.test.tsx` | before | after |
|---|---|---|
| `isMultiValueField` | 1 | 1 |
| `driver` (case-insensitive occurrences) | 4 | 7 |
| `objectstack#17469` | **0** | **1** |

| population on the merged tree | before | after |
|---|---|---|
| files naming the predicate (minus CHANGELOG) | 10 | 10 |
| carriers (also naming the driver) | 6 | 6 |
| **uncited carriers** | **1** — this test file | **0** |
| predicate-only (the sweep's own discriminator side) | 4 | 4 |

The docblock now states the relationship it was implying: the spec gates on the
field TYPE, `driver-sql` decides storage on `!!field.multiple` whatever the
type, the two part company for a `multiple: true` def on a type outside the
spec's multi-capable set, and objectstack#17469 owns which is right. It also
records why these cases sit clear of that disagreement — they relate through a
`lookup` field, which both rules call multi-valued.

Pin three on the merged tree: **`Tests 8 passed (8)`**. And it discriminates on
THIS file, proved rather than assumed — leg G below replaces the citation token
with plain words (occurrences 1 → 0, `driver` count unchanged at 7 so the file
stays classified as a carrier) and the pin reds again with exactly one uncited
file, then restores under a trap.

The diff adds docblock lines and removes none; no predicate, assertion or
executable line moved. The 7-leg rail matrix was re-run afterwards and **no leg
moved** — same six assertions, same red combinations, pre-fix leg still
`3 failed | 3 passed (6)`. `pnpm --filter @object-ui/plugin-detail test` on the
merged tree at `33bb17450`: `Test Files 172 passed (172) / Tests 1595 passed
(1595)` (one file and eight cases more than round two — the sweep this PR now
satisfies came in with the merge).

## Rework (2026-09-11) — prose only, no executable line moved

The file header carried a sentence that was FALSE: that the seam's verdict is
"the same predicate the driver that executes the query decides on". Measured at
source in the sibling checkout, `packages/drivers/driver-sql/src/sql-driver.ts`:

```ts
protected isJsonField(type: string, field: any): boolean {
  if (!field.multiple && FILE_REFERENCE_TYPES.has(type)) return this.mediaColumnIsJson();
  return JSON_COLUMN_TYPES.has(type) || !!field.multiple;
}
```

`!!field.multiple` on ANY type, against the spec's
`MULTI_OPTION_TYPES.has(type) || (MULTI_CAPABLE_TYPES.has(type) && multiple === true)`.
So `{ type: 'master_detail', multiple: true }` answers **false** to the spec and
**true** to the driver. The paragraph now states that measured relationship
instead of an identity, names which rule this renderer follows and why, and
cites **objectstack#17469**, which owns the divergence. The ⛔ "no local arity
rule here" instruction is kept — that half was true and load-bearing.

The file is now outside the in-flight citation sweep's failure set by
construction rather than by luck. Readings on this head:

| reading on `record-reference-rail.tsx` | before | after |
|---|---|---|
| `isMultiValueField` | 1 | 1 |
| `/driver/i` | 2 | 3 |
| `objectstack#17469` | **0** | **1** |

Every changed line begins with ` * ` inside the file docblock — verified by
reading the raw diff, not by intent. The 7-leg ablation matrix was re-run
afterwards and **no leg moved**: the same six assertions go red in the same
combinations, and the pre-fix leg still reads `3 failed | 3 passed (6)`.
`pnpm --filter @object-ui/plugin-detail test` re-run at `8fd5fba02`:
`Test Files 171 passed (171) / Tests 1587 passed (1587)`.

**M2 — which trade? Resolve arity FIRST, and for the rail's own reason.** Not a
copy of the badge. The rail's `fetchedSigRef` latch keys on
`(parentId + entries)`; the arity is in **neither**, so unlike `RelatedList` —
whose fetch effect re-runs on the verdict and can attempt equality, be refused,
and refetch — a probe-then-correct design would make the refused first attempt
the **only** attempt, leaving the entry on its error state until the user
navigates away. The rail also renders a navigational link decided by the same
verdict, so deferring the verdict ships the right rows under the wrong link.
The resolution is **not a gate**: an adapter without `getObjectSchema`, or one
whose fetch rejects, still reads rows on the historical equality wire
(pinned as `DEGRADATION CONTROL`).

**M3 — the URL grammar claim, re-measured, and the card's version is
incomplete.** The rail's link targets `/apps/:appName/:objectName`
(`AppContent.tsx:999` → `ObjectView`), **not** the `/data` surface the card and
`RelatedList.tsx:671-692` describe. Both parsers were run on the same input:

| input | `drillUrlFilters.parseUrlFilterTriples` (the `/data` surface) | `ObjectView.urlFilters` (**this link's target**) |
|---|---|---|
| `filter[owners][contains]=u1` | `[]` — dropped | `[["owners][contains","=","u1"]]` — **mangled** into a filter on a nonexistent field |
| **lit control** `filter[owners]=u1` | `[["owners","=","u1"]]` | `[["owners","=","u1"]]` |
| **lit control** `filter[amount][gte]=100` | `[["amount",">=","100"]]` | `[["amount][gte","=","100"]]` |

Both zero/garbage readings have a lit control in the same command. Conclusion
unchanged and strengthened: neither surface can express membership, and the one
the rail actually links to does not even drop the suffix — it emits a condition
on a field name that does not exist. There is no third spelling either: both
readers only look at keys starting with `filter[`
(`ObjectDataPage.tsx:228`, `ObjectView.tsx:2088`), so no JSON-AST param is
available on this route.

**M4 — is a multi-value `relationshipField` reachable? YES, and with zero
authoring.** `ReferenceRailEntrySchema.relationshipField` is a bare
`z.ZodString` (no enum, no arity constraint), and the registry manifest
(`plugin-detail/src/index.tsx:796`) publishes only `hideEmpty` as an input, so
nothing upstream narrows it. Stronger: the synthesised path reaches it without
an author at all — `deriveRelatedLists.ts:244` admits every `lookup` field with
no `multiple` filter, `lookup` is in the spec's `MULTI_CAPABLE_TYPES`, and
`buildDefaultPageSchema.ts:1059` copies that `relationshipField` straight onto
a rail entry. The card's premise holds and is wider than it states.

## Instrument discipline

**Pre-fix reading** — the rail source restored to `origin/main`'s bytes
(`git checkout c5b0cecf6 -- packages/plugin-detail/src/renderers/record-reference-rail.tsx`,
blob `a5086aa2` on disk, verified by hash and by a grep of the injected tokens
going `3 -> 0` and `2 -> 0`), same test file, same command:

```
Tests  3 failed | 3 passed (6)
  x SUBJECT - a multi-value relationship is queried by MEMBERSHIP, selectively
  x SUBJECT - the "View All" link is SUPPRESSED on a multi-value relationship, loudly
  x MEASURED DIFFERENCE - the rail resolves the arity BEFORE its one read
```

Restored by `git checkout HEAD -- ...` under an `EXIT INT TERM` trap, proved by
blob hash `e6f44dd1` and an empty `git diff HEAD`.

**Ablation matrix** — every leg mutates one thing, proves it reached disk
(anchor token count before/after plus a changed blob hash, so a zero-hit edit
cannot pass as a run), runs the same file, and restores under a trap:

| leg | mutation | goes RED |
|---|---|---|
| PRE-FIX | the whole rail file at its `origin/main` bytes | MEMBERSHIP, LINK, ARITY-FIRST |
| A | drop `!suppressViewAll` from the link's condition | LINK |
| B | withhold `fields` from `composeParentScopeFilter` | MEMBERSHIP, ARITY-FIRST |
| C | insert `if (fieldsFor.size === 0) return;` — turn the resolution into a GATE | DEGRADATION CONTROL |
| D | suppress the link unconditionally | LIVE CONTROL, DEGRADATION CONTROL |
| E | put membership on the wire unconditionally | LIVE CONTROL, DEGRADATION CONTROL |
| F | make the fake backend PERMISSIVE (answer equality against an array) | FIXTURE |

All six assertions discriminate; none stays green while the thing it names is
deleted. Leg F is the one that matters most for reading the rest: the FIXTURE
case is what stops a permissive evaluator from making this whole file lie, and
it is red the moment the evaluator stops refusing.

**Commands** (from the repo root, at ``3ab5db03c``):

- `pnpm exec vitest run packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx` -> `Test Files 1 passed (1) / Tests 6 passed (6)`
- `pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-detail^...' build` -> `VERDICT command-exit 0`, 12 of 47 workspace projects
- `pnpm --filter @object-ui/plugin-detail type-check` -> `VERDICT command-exit 0`. It covers the new test: `tsc -p tsconfig.test.json` is the leg that reported the two TS7006s this branch then fixed, so the coverage claim is a reading and not an assumption.
- `pnpm --filter @object-ui/plugin-detail test` -> `Test Files 171 passed (171) / Tests 1587 passed (1587)`, `VERDICT command-exit 0`, at `3ab5db03c` (`git rev-parse --short HEAD` at that run)
- Derived gates, each read off its own printed verdict line, not off a bare `$?`: `check:control-bytes` (`OK`, 7393 files), `check:spec-symbols` (three green lines, incl. "nothing cites a key its spec symbol does not declare" — the gate that caught #7299's hand-written arity rule), `check:changeset-claims` (exit 0, report-only), `check:new-line-citations` (`0 new citation(s)`), `check:phantom-deps` (`Every in-scope import is declared by the package that publishes it` — the one that would have caught an undeclared `@object-ui/core`), `check:i18n-keys` (exit 0).
- Repo-wide scans (`pnpm lint`, the full `pnpm test`) are CI's runs, not this branch's.
- Narrowed lint, as a supplement and not a substitute for CI's `turbo run lint`:
  `eslint --no-inline-config --format json` over the two changed files reports
  **2 files, 0 errors** (27 warnings, all `@typescript-eslint/no-explicit-any`
  plus one pre-existing `react-hooks/exhaustive-deps`). The rail file's own
  baseline was read the same way with its `origin/main` bytes on disk —
  **0 errors / 13 warnings, identical rule breakdown** — so this change adds no
  lint finding to it. This config enables no type-aware rules
  (`eslint.config.js` declares no `project` / `projectService`), so the diff
  cannot move the verdict on any file it does not contain.
- **NOT MEASURED: `check:sdui-registration-pins`** — it exits 2
  (`PREREQUISITE NOT MET`) without a console bundle, and building
  `@object-ui/console` needs that app's whole workspace closure, which is CI's
  run. This diff edits no `ComponentRegistry.register` call and adds no
  registration, so nothing it changes is in that gate's subject; CI measures it.

## Acceptance notes

Found while measuring, deliberately **not** fixed here and **not** widened into
this PR:

- **filed as #9196** — `ObjectView` (the component behind `/apps/:appName/:objectName`,
  i.e. the route this rail's own link targets) is a THIRD reader of the
  `filter[...]` URL family that never joined `drillUrlFilters`. Its
  `/^filter\[(.+)\]$/` is greedy, so `filter[amount][gte]=100` does not fail to
  match and does not get dropped: it emits a condition on a field literally
  named `amount][gte`. Measured in the M3 table above, with lit controls. No
  in-app producer reaches it today (`useOpenRecordList` navigates to `/data`),
  so it is latent — but the URL is an addressable surface by design.
- **filed as #9197** — `RecordContextValue.dataSource` is declared `string`
  ("Optional datasource id") in the published `@object-ui/react` type while
  every producer passes a `DataSource` ADAPTER, which is why a dozen consumers
  reach it through `(ctx as any)`. This card is a live instance: the rail's
  `getObjectSchema` read goes through one of those casts.
- **noted, not filed** — `RelatedList.tsx:671-692`'s comment describes the
  `/data` parser's drop behaviour as if it governed the link spelling
  generally. It is correct about the parser it names and this PR does not
  contradict it; it is simply not the parser a "View All" link reaches. Not a
  defect, no reproducer, and #9196 is where the real divergence is recorded —
  filing a second card for the comment would split one finding in two.
  Successor: whoever lands #9196.

## Scope declarations

- ⛔ `master_detail` untouched — settled on the card (comment 5607573447):
  `{type:'master_detail', multiple:true}` answers `false`, so
  `LineItemsPanel.tsx:200` and `MasterDetailForm.tsx:762` are correct as
  written.
- ⛔ `URL_FILTER_OPS` untouched — growing the URL operator set changes an
  addressable public link format on both sides plus its pin; that is a
  maintainer floor, not a patch.
- ⛔ `packages/plugin-detail/package.json` untouched — #8937 is in flight on
  that file, and no edit was needed: `@object-ui/core` is already a
  `peerDependency` of `plugin-detail`, and `core` already declares
  `@objectstack/spec: ^17.3.0`.

---

### DIFF
--- a/.changeset/8883-reference-rail-parent-scope-arity.md
+++ b/.changeset/8883-reference-rail-parent-scope-arity.md
@@ -1,0 +1,35 @@
+---
+"@object-ui/plugin-detail": patch
+---
+
+fix(plugin-detail): `record:reference_rail` compiles its parent scope by relationship ARITY, and stops offering a link it cannot express
+
+The rail took an author-supplied `relationshipField` per entry and compiled it
+as bare equality TWICE, in two different grammars: the `$filter` it puts on the
+wire and the `filter[<field>]=<value>` URL its "View All" link points at. When
+the named field is `multiple: true` — `Field.user({ multiple: true })` is the
+platform's own shape — the stored value is an array, `=` asks whether that
+whole array IS one id, and the driver refuses with `400 INVALID_FILTER` while
+prescribing `$contains`. The rail was empty for a perfectly legal authoring
+choice, and for one the platform reaches with no authoring at all: a child
+object's `lookup` field is derived into a rail entry whatever its arity.
+
+The `$filter` half now goes through `composeParentScopeFilter` from
+`@object-ui/core` — the one compiler the related list's rows and its tab badge
+already share — so a multi-value relationship is queried by membership and
+everything else keeps `=` byte for byte. No authoring key was added: the author
+is naming a relationship, and its storage form is the renderer's business. The
+child object's field defs are resolved BEFORE the reads rather than after a
+refusal, because the rail dispatches its queries once per (record + entries)
+and has no second attempt to correct itself with. Resolution is best-effort,
+not a gate: an adapter that cannot serve metadata still reads rows, with the
+equality wire it has always sent.
+
+The link half is the COST. No URL spelling on this surface carries membership —
+the data surface recognises `gte`/`lte`/`gt`/`lt` and drops any other suffix,
+and the route the rail actually links to parses equality only — so a hopeful
+`[contains]` suffix would not narrow the destination at all and "View All"
+would open the entire child table dressed as this parent's related records.
+Rather than ship a rail whose rows and link disagree, the link is suppressed on
+a multi-value relationship and the reason is logged once. Single-value entries
+keep their link with a byte-identical href.
--- a/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
+++ b/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
@@ -236,6 +236,14 @@ const viewAllLink = (): HTMLAnchorElement | null =>
 
 let warnSpy: ReturnType<typeof vi.spyOn>;
 
+/** The rail's own `console.warn` calls, by their first argument. */
+const railWarnings = (): string[] =>
+  (warnSpy.mock.calls as unknown[][])
+    .map((c) => c[0])
+    .filter((first): first is string =>
+      typeof first === 'string' && first.includes('RecordReferenceRail'),
+    );
+
 beforeEach(() => {
   vi.stubGlobal('IntersectionObserver', ImmediateIO as unknown as typeof IntersectionObserver);
   warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
@@ -302,12 +310,10 @@ describe('reference rail parent scope — compiled by ARITY (objectui#8883)', ()
     });
     // Absent on screen ⇒ the developer console is the only place it can be
     // said at all, and it IS said — once, naming the field and the reason.
-    const suppressionWarnings = warnSpy.mock.calls.filter(
-      (c) => typeof c[0] === 'string' && c[0].includes('RecordReferenceRail'),
-    );
+    const suppressionWarnings = railWarnings();
     expect(suppressionWarnings.length).toBe(1);
-    expect(suppressionWarnings[0][0]).toContain(MULTI_REF);
-    expect(suppressionWarnings[0][0]).toContain('View All');
+    expect(suppressionWarnings[0]).toContain(MULTI_REF);
+    expect(suppressionWarnings[0]).toContain('View All');
   });
 
   it('MEASURED DIFFERENCE — the rail resolves the arity BEFORE its one read', async () => {
@@ -359,11 +365,7 @@ describe('reference rail parent scope — compiled by ARITY (objectui#8883)', ()
     expect(drawn).toBeGreaterThan(0);
     expect(badgeText()).toBe(String(drawn));
     // Nothing was suppressed, so nothing was said.
-    expect(
-      warnSpy.mock.calls.filter(
-        (c) => typeof c[0] === 'string' && c[0].includes('RecordReferenceRail'),
-      ).length,
-    ).toBe(0);
+    expect(railWarnings().length).toBe(0);
   });
 
   it('DEGRADATION CONTROL — an adapter with no `getObjectSchema` still reads rows, on the historical wire', async () => {
--- a/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
+++ b/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
@@ -26,6 +26,18 @@
  * already share, whose verdict is `@objectstack/spec/data`'s own
  * `isMultiValueField`.
  *
+ * ⚠️ That verdict is the SPEC's, and it is NOT the driver's — the two rules
+ * are not one predicate, and this file does not assert that they agree. The
+ * spec gates on the field TYPE (`MULTI_OPTION_TYPES.has(type) ||
+ * (MULTI_CAPABLE_TYPES.has(type) && multiple === true)`), while `driver-sql`
+ * decides storage on `!!field.multiple` whatever the type is, so a
+ * `multiple: true` def on a type OUTSIDE the spec's multi-capable set is
+ * single-valued to the renderer and a JSON column to the driver. Which of the
+ * two is right is owned upstream by objectstack#17469; ⛔ nothing here decides
+ * it. The cases below stay clear of that disagreement on purpose: they relate
+ * through a `lookup` field, which both rules call multi-valued, so what they
+ * measure is the rail's compilation and not the open question.
+ *
  * The LINK half is not repairable at this layer and is not repaired: no URL
  * spelling on this surface carries membership, so the link is SUPPRESSED on a
  * multi-value relationship rather than pointed at an unscoped child table.
--- a/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
+++ b/packages/plugin-detail/src/renderers/__tests__/record-reference-rail.parentScopeArity-8883.test.tsx
@@ -1,0 +1,388 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8883 — `record:reference_rail` compiles its parent scope by the
+ * relationship field's ARITY, and refuses to render a "View All" link it
+ * cannot express.
+ *
+ * The rail took an author-supplied `relationshipField` and compiled it as bare
+ * equality TWICE, in two different grammars: the `$filter` it puts on the wire
+ * and the `filter[<field>]=<value>` URL its "View All" link points at. On a
+ * `multiple: true` relationship the stored value is an ARRAY of parent ids, so
+ * equality asks whether that whole array IS one id — which `driver-sql`
+ * refuses with `400 INVALID_FILTER` while prescribing `$contains`.
+ *
+ * ## Two halves, two different answers
+ *
+ * The `$filter` half is repairable and is repaired: it goes through
+ * `composeParentScopeFilter` from `@object-ui/core`, the ONE compiler the
+ * related list's rows (objectui#7299) and its tab badge (objectui#8882)
+ * already share, whose verdict is `@objectstack/spec/data`'s own
+ * `isMultiValueField`.
+ *
+ * The LINK half is not repairable at this layer and is not repaired: no URL
+ * spelling on this surface carries membership, so the link is SUPPRESSED on a
+ * multi-value relationship rather than pointed at an unscoped child table.
+ * Landing the filter alone would have produced a rail that shows the right
+ * rows above a link that goes somewhere else — a disagreement invisible until
+ * the user clicks. Both halves are pinned below, and so is the LINK's
+ * survival on the single-value arm that was already correct.
+ *
+ * ## Why the fake backend REFUSES equality instead of answering it
+ *
+ * The defect is invisible to a permissive backend: a fake that answers bare
+ * equality against an array-valued column with "no rows" turns a driver-level
+ * refusal into a plausible empty rail, and one that answers with the rows
+ * makes the bug look like a working feature. Either way every assertion here
+ * would be satisfied by the broken filter. The evaluator below therefore
+ * throws on that exact combination — and on any filter shape it does not
+ * recognise — and `FIXTURE` asserts that it does, in both directions.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, cleanup, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import React from 'react';
+import { MemoryRouter, Routes, Route } from 'react-router-dom';
+import { RecordContextProvider } from '@object-ui/react';
+
+import { RecordReferenceRailRenderer } from '../record-reference-rail';
+
+/** The rail gates its queries on an IntersectionObserver. Report intersecting
+ *  immediately so the fetch effect runs deterministically under jsdom. */
+class ImmediateIO {
+  constructor(private cb: (records: { isIntersecting: boolean }[]) => void) {}
+  observe() { this.cb([{ isIntersecting: true }]); }
+  disconnect() {}
+  unobserve() {}
+}
+
+const APP = 'demo';
+const PARENT = 'account';
+const CHILD = 'contact';
+const RECORD_ID = 'A1';
+const OTHER_ID = 'A2';
+
+/** The multi-value relationship field — an ARRAY of parent ids per child row. */
+const MULTI_REF = 'accounts';
+/** The single-value control's field — one parent id per child row. */
+const SINGLE_REF = 'account_id';
+
+/**
+ * The child object in the two arities this card is about. Only the ONE field
+ * member differs — `multiple: true` — so every difference the rail shows is
+ * attributable to the arity and to nothing else.
+ */
+const multiChildSchema = {
+  name: CHILD,
+  label: 'Contact',
+  fields: {
+    id: { type: 'text', label: 'Id' },
+    name: { type: 'text', label: 'Name' },
+    [MULTI_REF]: { type: 'lookup', reference: PARENT, multiple: true, label: 'Accounts' },
+  },
+};
+
+const singleChildSchema = {
+  name: CHILD,
+  label: 'Contact',
+  fields: {
+    id: { type: 'text', label: 'Id' },
+    name: { type: 'text', label: 'Name' },
+    [SINGLE_REF]: { type: 'lookup', reference: PARENT, label: 'Account' },
+  },
+};
+
+const MULTI_ROWS = [
+  { id: 'c1', name: 'Ada', [MULTI_REF]: [RECORD_ID] },
+  { id: 'c2', name: 'Grace', [MULTI_REF]: [OTHER_ID, RECORD_ID] },
+  { id: 'c3', name: 'Other Parent Only', [MULTI_REF]: [OTHER_ID] },
+];
+
+const SINGLE_ROWS = [
+  { id: 'c1', name: 'Ada', [SINGLE_REF]: RECORD_ID },
+  { id: 'c2', name: 'Grace', [SINGLE_REF]: RECORD_ID },
+  { id: 'c3', name: 'Other Parent Only', [SINGLE_REF]: OTHER_ID },
+];
+
+// --- the filter evaluator ---------------------------------------------------
+
+/** What `driver-sql` answers the equality form with on a multi-value column. */
+class InvalidFilterError extends Error {
+  code = 'INVALID_FILTER';
+  status = 400;
+}
+
+/**
+ * Compare one field against one scalar, the way a real driver does. An
+ * array-valued column under a bare scalar equality is REFUSED, not answered.
+ */
+function scalarEquals(stored: unknown, value: unknown, field: string): boolean {
+  if (Array.isArray(stored)) {
+    throw new InvalidFilterError(
+      `[fixture] 400 INVALID_FILTER: '${field}' stores multiple values; ` +
+        'equality cannot be evaluated against an array (use $contains)',
+    );
+  }
+  return stored === value;
+}
+
+/** Membership against a multi-value column; a scalar column holds one member. */
+function containsValue(stored: unknown, value: unknown): boolean {
+  return Array.isArray(stored) ? stored.includes(value) : stored === value;
+}
+
+/**
+ * Evaluate one filter node against a row. Handles exactly the MongoDB-style
+ * object this rail puts on the wire and THROWS on anything else — a permissive
+ * evaluator answers "all rows" for a shape it does not understand, and every
+ * assertion here would be satisfied by it.
+ */
+function matchesFilter(row: Record<string, any>, node: unknown): boolean {
+  if (node === undefined || node === null) {
+    throw new Error('[fixture] a rail query reached the backend with no $filter at all');
+  }
+  if (typeof node !== 'object' || Array.isArray(node)) {
+    throw new Error(`[fixture] unsupported filter: ${JSON.stringify(node)}`);
+  }
+  return Object.entries(node as Record<string, any>).every(([field, cond]) => {
+    if (cond !== null && typeof cond === 'object' && !Array.isArray(cond)) {
+      return Object.entries(cond as Record<string, any>).every(([op, value]) => {
+        if (op === '$eq') return scalarEquals(row[field], value, field);
+        if (op === '$contains') return containsValue(row[field], value);
+        throw new Error(`[fixture] unsupported operator '${op}' on '${field}'`);
+      });
+    }
+    return scalarEquals(row[field], cond, field);
+  });
+}
+
+/**
+ * A rail-shaped DataSource. `schema` of `null` models the adapter that cannot
+ * serve metadata at all — the degradation control — by omitting
+ * `getObjectSchema` entirely.
+ */
+function makeDataSource(rows: Record<string, any>[], schema: unknown | null) {
+  const find = vi.fn(async (objectName: string, params: any) => {
+    if (objectName !== CHILD) return { data: [], total: 0 };
+    const matched = rows.filter((r) => matchesFilter(r, params?.$filter));
+    const top = typeof params?.$top === 'number' ? params.$top : matched.length;
+    return { data: matched.slice(0, top), total: matched.length };
+  });
+  const ds: Record<string, unknown> = { find };
+  if (schema !== null) {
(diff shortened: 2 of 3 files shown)


os-tesla:
## PM review — ACCEPT on substance, one prose rework before arming

Read against `git merge-base origin/main <head>` = `759606e1d`: **3 files, +555/−18**.

**Every ruling-zone item holds, measured here:** R2 — `composeParentScopeFilter` / `isMultiValueRelationship` imported from `@object-ui/core`, no local arity rule anywhere in the diff. R3 — `URL_FILTER_OPS` untouched; the changed set is three files and none is the parser. R4 — **both** halves delivered: the `$filter` at the read and the link suppressed via `!suppressViewAll`. R5 — `packages/plugin-detail/package.json` untouched.

⭐ **M2 was the question I most expected to be answered by resemblance, and it wasn't.** The rail resolves arity before the reads for a reason that is its own, not the tab badge's: `fetchedSigRef` latches on (parentId + entries), the arity is in neither, so a probe-then-correct design would make the refused first attempt the *only* attempt. And ablation leg C — inserting `if (fieldsFor.size === 0) return;` — turns the DEGRADATION CONTROL red, which is what proves resolution is best-effort rather than a gate. That distinction is the whole difference between this card's fix and a silently empty rail on every entry in the app. Seven legs, all discriminating; leg F (permissive backend → FIXTURE red) closes the one hole that would make the file lie.

## ⛔ Rework: the file header states something false, and it collides with a pin in flight

`record-reference-rail.tsx:33-34` says the spec's `isMultiValueField` is *"the same predicate the driver that executes the query decides on."* It is not. Measured in objectstack, `packages/drivers/driver-sql/src/sql-driver.ts:17276`:

```ts
protected isJsonField(type: string, field: any): boolean {
  if (!field.multiple && FILE_REFERENCE_TYPES.has(type)) return this.mediaColumnIsJson();
  return JSON_COLUMN_TYPES.has(type) || !!field.multiple;
}
```

`!!field.multiple` on **any** type ⇒ `{type:'master_detail', multiple:true}` answers **false** to the spec and **true** to the driver. Owned upstream by `objectstack#17469`.

⚠️ **Not the author's error.** That sentence was in objectui#9184's seam when this branch was cut, the measurement falsifying it arrived from objectui#8937 afterwards, and my dispatch brief never carried the correction. This is the third copy of one false claim and I am the reason it spread.

**The collision, measured rather than predicted.** PR objectstack-ai/objectui#9195 adds a sweep: every tracked file under `packages` naming **both** `isMultiValueField` and matching `/driver/i` must cite `objectstack#17469`.

| reading on this head | value |
|---|---|
| `isMultiValueField` in `record-reference-rail.tsx` | 1 |
| `/driver/i` in the same file | 2 |
| `objectstack#17469` in the same file | **0** |
| CONTROL — `17469` in `parent-scope.ts` on PR #9195 | 1 |

⇒ whichever of the two lands second, that pin goes **red**. Correcting the paragraph here (prose only, ⛔ no predicate and no assertion moves) is cheaper than a round on either PR.

⭐ Instrument note for anyone chasing this claim: `git grep "same predicate the driver"` returns **nothing** on this file — the sentence wraps at `the / driver`, so a line-oriented grep cannot see it. That is the same degeneracy PR #9195's author hit and reported; it is why their pin keys on the two subjects instead of the wording.

Arm held until the rework lands and CI reads `pending: 0, failing: 0`.

---

os-tesla:
## Rework ACCEPTED — the pin collision is cleared, measured with a lit control

Head `8fd5fba02`.

| reading on `record-reference-rail.tsx` | before | after |
|---|---|---|
| `isMultiValueField` | 1 | 1 |
| `/driver/i` | 2 | 3 |
| **`objectstack#17469`** | **0** | **1** |

⇒ the file is now outside PR objectstack-ai/objectui#9195's pin-three failure set **by construction**, not by luck.

**The false claim is gone, and the probe is proven to be looking:** grepping the pre-rework head `3ab5db03c` for the same phrase returns **1**; the same grep on this head returns **0**. An empty reading with a dark probe would have been worthless here, which is the whole reason for the control.

**Prose only, verified rather than asserted:** every changed line in that file begins with ` * ` inside the docblock — filtering the diff for any changed line that is *not* docblock prose returns nothing. No predicate, no assertion, no executable line moved, and the 7-leg ablation matrix re-ran **identical** (3 failed / 3 passed pre-fix; legs A–F each reddening the same assertions). ⭐ None of the legs stopped discriminating, so nothing was quietly dropped — which is the failure mode a prose-only round is most likely to hide.

⭐ Two things the seat did that I want on the record. It **re-measured the driver rule at source** (`sql-driver.ts:17276`) instead of taking my brief's word for it — the brief was right, but it should not have been trusted, and it wasn't. And the rewritten paragraph does not merely delete the falsehood: it separates the **query** rule from the **storage** rule, states where they part company (`{type:'master_detail', multiple:true}` → `false` to the spec, `true` to the driver), says which one this renderer follows *and why* (the spec is what the authoring surface is validated against), and ⛔ declines to decide the divergence, citing `objectstack#17469` as its owner. That is strictly better than the sentence it replaces, which was confidently wrong in one clause.

⛔ Arm still held: 36 checks, **9 in progress, 0 failing**. Not-red-yet is not green.

---

os-tesla:
## ⛔ Ejected from the merge queue — cause found, and it is my miss

This PR entered the queue at 17:54 as `gh-readonly-queue/main/pr-9200-1f4e02995…`. It is now **gone from the queue with `main` unchanged at `1f4e02995`** — ejection, not a propagation delay (an absent ref does not disprove enqueueing, but an absent ref *plus* an unmoved `main` after the entry was seen does).

The PR is open, unmerged, and **green on its own head**: 36 checks, 0 pending, 0 failing at `8fd5fba02`. Green on the head and red in the queue is the tell — the queue merges `main` in, and `main` moved.

### The causal chain, every link measured

`1f4e02995` is PR objectstack-ai/objectui#9195, which landed **pin three**: every tracked file under `packages` naming **both** the spec predicate and the driver must cite `objectstack#17469`. Its population is `git grep -l isMultiValueField packages`, minus `CHANGELOG.md`, split by `/driver/i.test(sourceOf(rel))`.

This PR's **new test file** falls into it:

| reading on `record-reference-rail.parentScopeArity-8883.test.tsx` | value |
|---|---|
| `isMultiValueField` | 1 → inside `git grep -l`'s set |
| `/driver/i` | 4 → classified a **carrier** |
| `objectstack#17469` | **0** → fails the assertion |

This PR's base is `7696daac0`, which **predates** pin three — which is exactly why its own CI is green while the combined tree is not.

### ⚠️ The miss is mine, and it is a miss against a rule I wrote

I checked the pin-three collision on `record-reference-rail.tsx`, measured `17469` going 0 → 1 there, and told the seat the collision was cleared. But the sweep's population is the **changed set**, not the one file I happened to look at — and "the scope must cover every file in the changed set" is the standing rule in this lane's own notes, added earlier today after PR objectui#9192's author swept for `.ts` readers and missed a README the same PR edited. I restated that rule and then failed to apply it one PR later.

⭐ Worth recording for whoever reads this thread: **a cross-PR pin collision has to be measured against the whole changed set on the merged tree, not against the file that first raised the alarm.** The alarm file is the one you notice; the population is what the sweep actually enumerates.

The seat is reworking: merge `main`, cite the card in the test file's docblock as a real sentence rather than a grep-satisfying token, then prove the file moves into the cited-carrier set and that pin three still discriminates on it. ⛔ No predicate, assertion or executable line moves; the 7-leg ablation matrix is re-run afterwards.

⛔ Not re-armed. I arm it when CI on the merged tree reads `pending: 0, failing: 0`.

---

os-tesla:
## Third round ACCEPTED on content — three state questions answered. ⛔ Arm still held (5 checks running).

### The fix, verified on the tree that actually matters

⭐ The seat **reproduced the ejection locally before editing anything** — pin three on the combined tree read `Tests 1 failed | 7 passed (8)`, naming exactly one uncited carrier. That converts my causal chain from an inference into a measurement, which is the right order.

I re-measured independently, and deliberately on the **merged** tree rather than this PR's head — `git merge-tree --write-tree origin/main pr9200` → `2ccf6bd07` — running the sweep's own enumeration over it:

| reading | value |
|---|---|
| uncited carriers on the merged tree | **0** |
| `isMultiValueField` in the 8883 pin | 1 (still in the `git grep -l` set) |
| `/driver/i` in the same file | 7 (still a carrier, by more margin) |
| `objectstack#17469` | **1** |

⭐ **Leg G is the leg that matters**, and it is the one I would have asked for: remove the citation, leave the `driver` count **unchanged at 7** so the file provably stays in the carrier set and only the citation moves, and pin three reddens again on the same assertion naming the same one file. That proves the pin discriminates *on this file*, not merely that the suite is green. The 7-leg rail matrix re-ran with **no leg moved**.

### Q1 — the review gate: **A**, leave it hung; I clear both in one stroke after reviewing this head

The re-hang is owed and correct. My clear at 17:53 judged `8fd5fba02`; the head is now `33bb17450`, so `--pair`'s C3 — *"the review that cleared this gate judged a different tree"* — is exactly right. A gate cleared against a tree that no longer exists is not cleared.

⭐ And the self-disclosure is the part worth keeping: the seat made a split write (PR only), `--pair` flagged it as C1, and it **removed its own label in the same turn rather than leave the split**, then re-hung on both carriers in one stroke — and reported the whole sequence including its own misreading. `--pair 9200` now reads **both carriers agree**, which I confirmed myself rather than trusting the write.

### Q2 — the draft flag: **A**, leave it

"⛔ do not flip ready" meant do not flip it ready; it was never an instruction to revert a flip I made. Reverting another actor's state write is not the seat's call and it was right not to. Non-draft with `auto_merge: null` and no queue ref is a correct resting state — I re-arm when CI is clean.

### Q3 — attribution: **A stands**, and the conflict is escalated, not resolved here

The commits are pushed, and ⛔ rewriting pushed history over a trailer is forbidden by this lane's own rules — so A is the only available answer regardless of which form is preferred. The underlying conflict (this session's attribution instruction names a model identifier; the standing rule forbids one in pushed artifacts) is flagged to the maintainer and is theirs to settle. ⭐ Flagging rather than choosing silently was right.

### ⚠️ One instrument finding from the seat's timeline, worth the lane's notes

The size-labeler's **whole-group PUT stripped all three labels off this PR one second after I set them** (17:02:13 → 17:02:14, `github-actions[bot]`); I re-hung them at 17:02:40 without realising why they had gone. That is a live additive-write hazard on this repo, and it is what made the 17:53 clear ambiguous to read afterwards. ⇒ **read labels back after writing them**, and prefer the checker's reading over one's own write. Not filed — the recovery is already the documented practice — but recorded here because it cost a round of confusion.

---

os-tesla:
## Contract review — head `33bb17450a` — **PASS**

Reviewed-by: `domain:ui` PM seat (`os-tesla`), R16, 2026-09-11T19:0xZ.

⚠️ Posted because `check-clause2-carriers --pair 9200` raised **C6**: I cleared the gate on both carriers against this head having *done* the re-review, but never recorded it in the shape the checker reads. The reading was right and the record was missing — that is the checker doing its job, and the gap was mine.

**Clause-② `yes`** — declared at dispatch, ⛔ not revised on delivery.

### Scope, measured against the merge base (⛔ not against the previous head)

My first attempt diffed `8fd5fba02..33bb17450` and read 27 files — because the seat **merged `origin/main`** between those two heads, so that range is mostly other people's landed commits. Re-measured against `git merge-base origin/main 33bb17450` = `1f4e02995`, this PR's actual contribution is:

```
.changeset/8883-reference-rail-parent-scope-arity.md    |  35 ++
record-reference-rail.parentScopeArity-8883.test.tsx    | 402 +++++
packages/plugin-detail/src/renderers/record-reference-rail.tsx | 163 ++-
3 files changed, 582 insertions(+), 18 deletions(-)
```

⭐ Worth naming: **the previous head is not a baseline once a merge has happened.** That is the same family as `$B^` not being a merge base, and it nearly made me review a diff that was three-quarters other seats' work.

### The verdict

**`patch` on `@object-ui/plugin-detail` holds.** The bump test this lane uses is *does existing stored data render differently* — and here the affected configuration was **already broken**: a multi-value relationship rail fetched with bare equality, refused by `driver-sql` with `400 INVALID_FILTER`, so the rail was empty, under a "View All" link whose own target (measured in M3) mangles `filter[owners][contains]` into a condition on a field literally named `owners][contains`. Repairing a path that returned nothing and removing a link that pointed nowhere correct is a fix, not a behaviour revision. (Contrast objectstack-ai/objectui#9202, `minor`: there a working dashboard renders differently.)

No declaration changed, no authoring key added, `URL_FILTER_OPS` untouched, `packages/plugin-detail/package.json` untouched.

### The delta since my acceptance at `8fd5fba02`

Prose only — 12 added docblock lines carrying the `objectstack#17469` citation. ⛔ No predicate, assertion or executable line moved, and the 7-leg ablation matrix re-ran with no leg moving.

### Independently re-verified before this verdict

- **On the merged tree**, not the PR head: `git merge-tree --write-tree origin/main pr9200` → `2ccf6bd07`, then the sweep's own enumeration over it ⇒ **uncited carriers 0**. That is the exact reading whose absence got this PR ejected from the queue the first time.
- The pin's discriminating power on *this* file, via the seat's leg G: remove only the citation, hold `driver` at 7 so the file provably stays in the carrier set, and pin three reddens on the same assertion naming the same one file.
- CI at `33bb17450`: 36 checks, **0 pending, 0 failing**.

⇒ gate cleared on both carriers in one stroke, ready, SQUASH armed; queue ref `gh-readonly-queue/main/pr-9200-1f4e02995…` confirms the arm took.

---

PR #9202, opened 2026-09-11
fix(dashboard): honour a widget's declared chartConfig on the inline chart relays

os-tesla:
Part of #4044

Continues the branch a previous seat pushed before it was killed mid-flight (commit `dd461a0d3`, which left no report). Nothing in that commit was inherited on trust: every claim it made is re-measured below on this tree, and one of them turned out to be wrong in a way that mattered.

Dispatched on the ruling in objectui#4044 comments 5617466747 and 5617614527, not on the card body's three-route menu, which the ruling supersedes. `Clause-2: no`.

---

## What lands

`DashboardWidget.chartConfig` is declared as the spec's full `ChartConfigSchema` — 14 authorable keys, measured off `@objectstack/spec@17.4.0` at runtime — on **every** dashboard widget. The two INLINE relays, `DashboardRenderer` and `DashboardGridLayout` (they compose the chart node for a widget bound to inline rows or to a `provider: 'object'` aggregate, neither of which is an ADR-0021 dataset), mentioned `chartConfig` **zero** times. Both now lower it through `chartConfigPresentation` — the same `@object-ui/core` whitelist the dataset path already uses, imported rather than forked.

Two commits:

1. `dd461a0d3` — the forwarding itself, plus a seam pin, a DOM pin and a `showLegend` drawn pin.
2. `139997756` — the evidence that commit was missing: a dashboard-surface **drawn** pin for the six plot-internal keys, the `aria` measurement, and three comments corrected because they now read false.

---

## A. Per-key measurement — the card's premise is stale in BOTH directions

The card (filed 2026-08-04, migrated 2026-08-10) says only `showLegend` lands on the dashboard path. Measured at the merge-base with one command, every zero carrying a lit control in the same sweep:

```
6a8e2a31 DashboardRenderer.tsx    chartConfig=0  chartConfigPresentation=0  [control] widget=173  lines=1299
6a8e2a31 DashboardGridLayout.tsx  chartConfig=0  chartConfigPresentation=0  [control] widget=88   lines=626
6a8e2a31 DatasetWidget.tsx        chartConfig=7  chartConfigPresentation=4  [control] widget=85   lines=1658
```

So the premise is wrong twice over. The **dataset** path is no longer a one-key path — objectstack#7016 landed the per-key forwarding there on 2026-08-09, five days after the card was written, and `DatasetWidget.tsx` is byte-identical before and after this PR. The **relays** were worse than the card says: not "only `showLegend`", but nothing at all, `showLegend` included.

Emitted keys measured functionally, by calling `chartConfigPresentation` with each declared key alone:

| key | declared | dataset path before | relays before | relays now | read by the renderer (measured) |
|---|---|---|---|---|---|
| `title` | yes | forwarded | **no** | yes | ChartFrame heading |
| `subtitle` | yes | forwarded | **no** | yes | ChartFrame subheading |
| `description` | yes | forwarded | **no** | yes | `role="img"` + `aria-label` on the chart container |
| `colors` (array) | yes | forwarded | **no** | yes | positional palette, off the sector fills |
| `colors` (record) | yes | forwarded | **no** | yes, as `categoryColors` | per-category fill wins over the palette |
| `height` | yes | forwarded | **no** | yes | inline height on the chart container |
| `showLegend` | yes | forwarded (#3135) | **no** | yes | legend wrapper present/absent |
| `showDataLabels` | yes | forwarded | **no** | yes | Recharts LabelList texts |
| `annotations` | yes | forwarded | **no** | yes | ReferenceLine / ReferenceArea |
| `interaction` | yes | forwarded | **no** | yes | Brush, and the tooltip wrapper |
| `type` | yes | refused | no | refused | n/a — see B |
| `xAxis` / `yAxis` / `series` | yes | refused | no | refused | not ours — objectstack#17385 owns the precedence question, and the whitelist emits none of the three, so this change cannot pre-empt it |
| `aria` | yes | refused | no | refused | **nothing** — see B |

Nine keys land on the relays (ten counting the `categoryColors` arm). Eight of the nine on the ruling's DO-NOW list are delivered; the ninth is `aria`.

---

### `aria` is refused, and here is the measurement behind the refusal

`aria` is on the ruling's DO-NOW list, so "nothing reads it" is measured rather than asserted, in **both** spellings it could travel in:

- **nested**, as the spec object `{ ariaLabel, ariaDescribedBy, role }` on the chart node;
- **flattened**, onto the node's own `ariaLabel` / `ariaDescribedBy` / `role` — members `BaseSchema` already declares and `SchemaRenderer.resolveAriaProps` already converts into `aria-label` / `aria-describedby` / `role`. That route needs no new declaration anywhere, which is exactly why it had to be measured rather than dismissed.

Rendering the real chain with each spelling on the chart node produced, for both:

```
anyAriaLabel: []   anyRoleFigure: 0   anyAriaDescribedBy: 0   chart role: null
```

and the **lit control** in the same probe — `description`, travelling the same whitelist onto the same node — produced:

```
anyAriaLabel: [["DIV","Invoice value by status"]]   chart role: "img"
```

Then the same thing from the other end: two ablation legs that forward `aria` on purpose from both relays, nested and flattened. In both runs the **only** red was the seam refusal assertion; the DOM assertion "an authored chartConfig.aria reaches no attribute on this surface" stayed green. Forwarding it changes nothing on screen.

Cause: `ChartRenderer` destructures `{ schema, onChartClick }` and drops every other prop; `AdvancedChartImpl` declares no `aria` prop; `normalizeChartSchema` names neither. The chart's one accessible name comes from `description`.

**So delivering `aria` needs a READER inside `@object-ui/plugin-charts` — a new member on a published face — and it also has to answer to the accessible name `description` already sets on the same element.** The claim comment on this card leaves exactly this door open, and this is it: reported, not guessed, and not implemented. Forwarding it without a reader would have moved declared-but-not-delivered one layer down, which is the disease this card exists to cure.

### `type` is refused, and that is measurable too

Not on the ruling's DO-NOW list. On this path the chart family travels as `chartType`, set from the widget's own `type` through `classifyWidgetType`; the node's `type` is the component-registry dispatch key (`chart` / `object-chart`). Both ablations confirm the refusal pin can fail: emitting `chartConfig.type` as `chartType`, and emitting it under its own name `type`, each redden "emits no chart family from chartConfig.type — the widget type still picks it" and nothing else.

---

## CONTROL — the react tier's `ObjectChart` flat-prop path

Named by the ruling, measured as blob identity against the merge-base rather than by reading the diff:

```
IDENTICAL  packages/plugin-charts/src/ObjectChart.tsx         32b7ef61c
IDENTICAL  packages/plugin-charts/src/ChartRenderer.tsx       13e49a047
IDENTICAL  packages/plugin-charts/src/AdvancedChartImpl.tsx   ae52b1d01
IDENTICAL  packages/plugin-charts/src/normalizeChartSchema.ts 3217874418
IDENTICAL  packages/plugin-charts/src/ChartContainerImpl.tsx  117306bae
IDENTICAL  packages/plugin-charts/src/index.tsx               12a42f173
IDENTICAL  packages/plugin-charts/src/types.ts                be0839f0d
IDENTICAL  packages/core/src/utils/chart-presentation.ts      48a0d0eef
IDENTICAL  packages/react/src/SchemaRenderer.tsx              67a602afe
```

Every file on that path is byte-identical. `@object-ui/core`'s whitelist is untouched too, so the dataset path and the report renderer are unchanged by construction. Behavioural half: `packages/plugin-charts/` and `packages/plugin-dashboard/` run green in full (below).

---

## The pin shape, and the thing the first commit got wrong

The ruling asks for a rendering assertion **on the dashboard surface** for every forwarded key, never a "the prop was passed" assertion. The first commit left the six plot-internal keys (`colors`, `categoryColors`, `showDataLabels`, `annotations`, `interaction`, `showLegend`) with drawn evidence only in `plugin-charts/src/ChartRenderer.dashboardChartConfig.test.tsx`, which hand-builds its schema. Measured: delete the forwarding from **both** relays and every test in that file still passes. It pins the chart block, never this seam.

`DashboardChart.chartConfigMarks-4044.test.tsx` closes that: widget metadata in, drawn Recharts marks out, both relays, no renderer stub anywhere in the chain.

Its header comment claimed the marks could only be pinned inside `plugin-charts`, because a `recharts` module mock is needed and `recharts` resolves there alone. The premise is still true — re-measured, `require.resolve('recharts')` from `packages/plugin-dashboard` is MODULE_NOT_FOUND — but the conclusion was wrong. `ResponsiveContainer` seeds its size from `getBoundingClientRect()` **on its own element**, read synchronously inside its resize effect, consulting `ResizeObserver` only for later changes; the repo's happy-dom polyfill for that observer is a no-op, which is why nothing painted. Sizing that one element is enough, and needs no module mock.

Scoped to that element, never blanket — also measured, because the blanket form looks like it works: with every element answering 480x320, Recharts' own axis-label measurement reads 480x320 too, the x-axis claims the whole box, and the plot renders with a clip rect of `height="0"` and no marks while `.recharts-surface` is present. A file that waited on the surface and then asserted "no data labels" would have passed for that reason.

---

## Ablation matrix

Every leg mutates **both** relays at the same anchor, runs the three dashboard pin files, then restores with `git checkout HEAD -- path` and proves the restore by blob hash and an empty `git diff HEAD`. Each mutation is proven on disk before the run (anchor count 1 before, 0 after; marker count 1 after).

| mutation | seam assertion reddened | rendering assertion reddened |
|---|---|---|
| whole forwarding removed | 24 | 22 (46 total across the three files; **0** in plugin-charts) |
| drop `title` | forwards the chart titles and the accessibility description | draws chartConfig.title / .subtitle above the plot |
| drop `subtitle` | same | same |
| drop `description` | same | announces chartConfig.description as the chart graphic accessible name |
| drop `height` | forwards an explicit plot height and drops a non-positive one | applies chartConfig.height over the relay height class |
| drop `colors` | lets an array `colors` override the relay default palette | paints the marks from an array `colors` palette |
| drop `categoryColors` | lowers a record `colors` as the per-category map, not as the palette | paints per-category colours from a record `colors` map |
| drop `showLegend` | forwards showLegend and showDataLabels in both directions | draws no legend when showLegend is false, and draws one otherwise |
| drop `showDataLabels` | same | prints each point value on the mark when showDataLabels is on |
| drop `annotations` | forwards annotations and the interaction toggles | draws a reference line / a reference area |
| drop `interaction` | same | adds the range selector; removes the hover tooltip |
| ADD `aria` (nested) | ignores chartConfig.aria | **none** — the finding, not a hole |
| ADD `aria` (flattened) | ignores chartConfig.aria | **none** — same |
| ADD `type` (as `chartType`) | emits no chart family from chartConfig.type | n/a, refused key |
| ADD `type` (as `type`) | emits no chart family from chartConfig.type | n/a, refused key |

No forwarded key has an assertion that survives its own deletion. The control arms (undeclared / off) deliberately stay green in every leg — that is what makes the positive arms readable.

---

## Verification

Everything below ran on the final head `9adbdfaeb` (this branch merged with `origin/main` at `1f4e02995`; synced by merge, never rebase).

- `pnpm exec vitest run packages/plugin-dashboard/ packages/plugin-charts/` under the shared verify lock: **174 files, 1623 tests, all passed**, lock verdict line `VERDICT command-exit 0`.
- `pnpm turbo run type-check --filter=@object-ui/plugin-dashboard --filter=@object-ui/plugin-charts` (`tsc --noEmit` plus `tsc -p tsconfig.test.json`, whose include names `src/**/*.test.tsx`, so the new pin is type-checked): 15 tasks successful, `VERDICT command-exit 0`.
- `eslint .` in each of the two packages: exit 0, 0 errors (470 and 368 pre-existing `no-explicit-any` warnings, unchanged in kind).
- Gates derived by hand from this repo's `package.json` and `.github/workflows/` (this repo has no `scripts/pm/dispatch-gates.mjs`), each read off its own printed verdict line, never a bare exit variable: `check-changeset-presence` (6 source files of 2 released packages changed, 1 changeset declared), `check-changeset-claims`, `check-control-bytes` (7404 files), `check-test-path-roots`, `check-new-cross-file-line-citations` (0 new citations), `check-vi-mock-specifiers`, `check-vi-mock-inherit`, `check-vi-mock-override-shape`, `markdown-test-inputs.mjs --audit` (47 candidates, all adjudicated — the new pin reads no markdown, so `ADJUDICATED` needs no entry, and the gate's own audit is what says so), `check-governed-queue-guard --test` on the changed paths (NOT GOVERNED).
- Repo-wide gates not run locally: CI owns the farm.

## Acceptance notes

- **The dataset path has the same evidence gap this PR closes for the relays, and it is deliberately left alone.** `DatasetWidget.chartConfig.dom.test.tsx` pins only `title` / `subtitle` / `description` / `height`; its five plot-internal keys, forwarded since objectstack#7016, still have no dashboard-surface drawn assertion. The technique in the new marks file transfers verbatim. Out of this card's scope (the ruling's DO-NOW is the not-yet-forwarded keys, and the relays are what was unforwarded), so: noted, not filed.
- The changeset said "Three keys are deliberately still not forwarded" and then listed five. Corrected to five, with the `aria` measurement written into it.
- There is no `package: plugin-dashboard` label in this repository, so the package labels here are what exists.

Not enqueued, no auto-merge, stays in draft — the dispatching seat arms it.

---


os-tesla:
## Contract review — **PASS**. Recorded now so the arm is a pure CI gate.

`minor` on `@object-ui/plugin-dashboard` is the right bump, and the changeset argues it in the only terms that settle the question:

> a dashboard whose stored metadata **ALREADY** carries `chartConfig` on an inline-bound chart widget renders differently after this change … Widgets that declare no `chartConfig` compose exactly what they composed before.

That is the distinction that decides `minor` vs `patch`, and it is worth contrasting with the call I made one PR over. On objectstack-ai/objectui#9200 I accepted `patch` because the affected configuration was **already broken** — an empty rail on a `400`, under a link that mangled its own filter. Here the affected configuration **worked**; it merely ignored keys the author had written, and existing stored metadata now renders differently with nobody editing it. ⇒ `minor`, and the no-`chartConfig` path being byte-identical is what keeps it from being wider than that.

The five refused keys each carry a measured reason rather than a silence, which is the whole point of the card:

- `type` — the widget's own `type` already picks the family on this path.
- `xAxis` / `yAxis` / `series` — an open protocol question, filed for the spec seat as `objectstack#17385`. ⇒ `Part of`, ⛔ not a closing keyword, correctly.
- `aria` — ⭐ refused on a **measurement, taken in both spellings it could travel in**, not on a judgement call. Forwarded nested and again flattened onto the node's own `ariaLabel`/`ariaDescribedBy`/`role`; neither moved a single attribute, because `ChartRenderer` drops every prop but `schema` and `onChartClick`. Delivering it needs a reader inside `@object-ui/plugin-charts` — a new member on a published face, which Clause-② `no` forbids without a ruling. That decision is now posted on the card (comment 5639000150) as owed item #8.

### Verified independently before this verdict

- **Card premise wrong in both directions**, measured at the merge base with lit controls in the same sweep: `chartConfig` **0** in each relay (control `widget` **173** and **88**) — so the card's "only `showLegend` lands" was *optimistic*, nothing landed there; and `chartConfigPresentation` already **4×** in `DatasetWidget.tsx` — so it was *pessimistic* about the dataset path, which has lowered the whole whitelist since objectstack#7016.
- **React-tier control by blob identity**, not by reading a diff: `ObjectChart.tsx`, `core/src/utils/chart-presentation.ts` and `DatasetWidget.tsx` are **byte-identical** to the merge base. Because the shared whitelist is untouched, the dataset path is unchanged by construction rather than by assertion.
- `--pair 9202` → both carriers agree, and the checker adds that the diff carries **no widening tell** (⚠️ a tell is not a proof, and its absence is not one either — but it corroborates the `no`).

⇒ On CI reaching `pending: 0, failing: 0` I clear `needs:contract-review` from **both** carriers in one stroke, flip ready and arm SQUASH. ⛔ Nothing to re-review at that point.

---

PR #9224, opened 2026-09-12
fix(grid): withhold a row's Edit/Delete when the host refuses that record (objectui#8674)

### LINKED ISSUE #8674: FieldDesigner draws a delete action on system fields and silently drops the click — handleDelete returns before the confirm dialog, with no message anywhere
Surfaced while implementing objectui#8060 (dev seat, session `session_01YBWFb5YgMU5dw8p2VKj16S`). ⛔ Unassigned; grading and `domain:*` are triage's. ⛔ NOT fixed in objectui#8060's PR — out of that card's scope, which is the field-type flattening.

## Measured

`packages/plugin-designer/src/FieldDesigner.tsx`:

```ts
const handleDelete = useCallback(async (record: Record<string, unknown>) => {
  const field = fields.find((f) => f.name === record.name);
  if (!field || field.isSystem) return;      // ← silent no-op
  const confirmed = await confirmDialog.confirm(...);
  ...
}, [fields, onFieldsChange, confirmDialog, t]);
```

The delete affordance is wired at the GRID level, not per row:

```tsx
onDelete={readOnly ? undefined : handleDelete}
```

`ObjectGrid` takes a single `onDelete` for the whole grid (`packages/plugin-grid/src/ObjectGrid.tsx`, `onDelete?: (record: any) => void`) and derives `{ update: !!onEdit, delete: !!onDelete }`, so the row action is rendered for EVERY row including system fields.

⇒ **Clicking delete on a system field does nothing at all** — no confirm dialog, no banner, no toast, no console message. The author clicks a button that is drawn as available and gets no response, and nothing anywhere says why.

## Why this is worth a card rather than a shrug

It is the small end of the family this repo keeps finding: an affordance that is offered and then silently declines. The `readOnly` case is handled honestly (the operations are withheld, so no button is drawn); the `isSystem` case draws the button and drops the click. The two states differ in the code and do not differ on screen.

`isSystem` is also the flag the drawer form uses to disable `name` and `type` — so the concept "this field's structure is not the designer's to change" already exists and is already rendered in one place. Delete is the one place it is enforced invisibly.

## Not asserting a direction

Three that occur to me, none ruled here and none costed:

- Withhold the row's delete action for system fields (needs per-row operation gating, which `ObjectGrid`'s grid-level `onDelete` does not offer today — that may be the real blocker and may make this a `plugin-grid` card rather than a `plugin-designer` one).
- Show the existing confirm dialog's error surface, or a toast, saying a system field cannot be deleted.
- Leave the behaviour and state it — the weakest, since the button still lies.

## Scope of the duplicate search behind this filing

Bounded, and said so rather than implied: the 500 most recent `plugin`-labelled issues in **all** states (5 pages of 100 via the REST list endpoint) were scanned by TITLE for `FieldDesigner`/`field designer` intersected with `delete`/`system`/`remove` — zero hits, with a lit control on the same command shape (`designer` matched 11 rows). Issue BODIES were not searched and the label scope was not widened, so this is a targeted read and not an exhaustive one.

## Related

objectui#8060 (surfaced it) · objectui#6044 (the card that made `isSystem` read the spec spelling `system`, i.e. why system fields are correctly identified here at all).

---

os-tesla:
Fixes #8674

`ObjectGrid` takes one grid-level `onDelete` and derives `{ update: !!onEdit, delete: !!onDelete }`, so the row action is drawn for every row. A host whose refusal is per RECORD had nowhere to put it and put it in the callback instead: `FieldDesigner`'s `handleDelete` returned early on `field.isSystem`, after the button had been drawn and clicked — no dialog, no toast, no console message. `readOnly` was honest in the same component (the callback is withheld, so no button is drawn); `isSystem` was not.

This adds the per-row operation predicate triage ruled the real content of the card, with `FieldDesigner` as its first caller.

- `packages/plugin-grid` — `ObjectGridComponentProps` gains `rowOperations`, plus the exported `ObjectGridRowOperations` type. Additive, opt-in, narrowing-only.
- `packages/plugin-designer` — `FieldDesigner` passes it, so a system field's row draws no Delete. Edit is untouched.
- Changeset: `plugin-grid` minor (new published prop on a published block), `plugin-designer` patch (the user-visible fix). Judged on the stated test — no existing caller and no stored document renders differently; only the designer's own system rows change, which is the defect being fixed.

## M1 — the shape, measured against the derivation that already exists

The prop is

```ts
rowOperations?: (record: any) => ObjectGridRowOperations | null | undefined;
// ObjectGridRowOperations = { update?: boolean; delete?: boolean }
```

read once per row inside the actions cell and ANDed into the verdict expression that is already there:

```tsx
canDelete={resolveRowRecordCrudAffordance(canDelete && rowOps?.delete !== false, recordVerdict(rowRecordId(row), 'delete'))}
```

**What `ObjectGrid` already does**, which decided this: `resolveRowCrudAffordances` folds the ADR-0103 lifecycle bucket, the object's `userActions`, the server's effective API operations (#3720) and the principal's grant (#4096) into one object-level `{ canEdit, canDelete }`; the actions cell then narrows that per row twice already — the #2614 `visibleWhen` / `disabledWhen` predicates, and the #4296 record-level explain verdict through `resolveRowRecordCrudAffordance`. There was already a per-row seam; what was missing was a channel for the CALLER to speak into it. So this is one more factor in the same conjunction, in the same vocabulary, at the same site — not a parallel mechanism beside it. `false` withholds; `true`, an omitted member, a `null`/`undefined` return and an absent prop all leave the verdict untouched. Never a union: nothing it returns can re-open what the layers above closed, pinned by its own test.

**Against a per-row `operations` callback** (widening the existing key to accept a function): `operations` is an AUTHORED schema key — read as `'operations' in schema`, declared on `ObjectGridSchema`, and accepted by `ComponentPropsMap['object-grid']`, a `strictObject`. A function cannot be serialised into a view document, so the function form could only ever arrive from code while the same key arrives from JSON: two dialects for one key, one of which the save gate refuses (AGENTS.md #0 / #0.1). Rejected.

**Against a `disabledOperations` set**: two independent refusals. It renders the entry and declines the click, which is arm 2 in arm 1's clothes and is what the ruling excludes — the repo's own posture here is hide, not disable (#4296 pins "a fully denied row grows no trigger at all"). And a record-keyed set cannot address these rows: `FieldDesigner`'s own comment records that the grid's `$select` strips `id`, which is why it keys fields by `name`. A set keyed on record id would have gated nothing, silently. Rejected.

## M2 — blast radius, enumerated and then measured

Every `ObjectGrid` call site in the tree, non-test:

| call site | passes a predicate? |
|---|---|
| `packages/plugin-designer/src/FieldDesigner.tsx` | YES — the only one |
| `packages/plugin-designer/src/ObjectManager.tsx` | no |
| `packages/plugin-view/src/ObjectView.tsx` | no |
| `packages/plugin-grid/src/index.tsx` (the SDUI renderer) | no |
| `packages/plugin-grid/demo/bulk-actions.tsx` | no |
| `apps/console/src/dev/DevRowActions.tsx` | no |

(A seventh occurrence, in `packages/fields/src/widgets/RecordPickerDialog.tsx`, is inside a docblock, not a render.) The rest of the 125 occurrences of the string are tests and type-level pins.

**The control is measured, not asserted.** A throwaway harness rendered two no-predicate callers — the real, untouched `ObjectManager`, and a bare grid with `onEdit`/`onDelete` — and dumped every byte each produces: the container HTML plus `document.body` with each row's kebab actually open. It was run twice: once on this branch, once with the three changed source files checked out at the base commit `b775500af` (mutation proved on disk first — the `rowOperations` marker went 6→0 in `ObjectGrid.tsx` and 3→0 in `FieldDesigner.tsx` — and restored by blob hash afterwards).

```
92191 bytes both legs
80f5438a14681caa5760abe343ef540c91050dcecebb6778a84407e9c728c6bb  base.html
80f5438a14681caa5760abe343ef540c91050dcecebb6778a84407e9c728c6bb  head.html
cmp: identical
```

Lit control for the instrument: `row-action-builtin-delete` occurs 4 times in each dump (two callers, two rows each), so the capture really did contain the affordance the change is about, rather than agreeing about an empty page.

Behavioural half, same claim: `packages/plugin-view` — a real consumer — is green unchanged, 37 files / 316 tests.

## M3 — the card's anchors, re-measured on today's tree (located by text, never by line)

- `packages/plugin-designer/src/FieldDesigner.tsx` still carries `if (!field || field.isSystem) return;` as the FIRST statement pair of `handleDelete`, above the `confirmDialog.confirm` call — the card's quote is current.
- `packages/plugin-grid/src/ObjectGrid.tsx` still declares `onDelete?: (record: any) => void` on the renderer props, and still derives `{ update: !!onEdit, delete: !!onDelete }` when the schema declares no `operations`.
- R3 reproduced before building anything: `rowOperations|operationsFor|disabledOperations` across `packages/plugin-grid/src/` = **0**, with `ObjectGrid` at **578** across the same file set as a lit control. The zero was a probe that was looking.

## M4 — the neighbouring honest case

The drawer form's `isSystem` use is two inline reads — `disabled: readOnly || (editingField?.isSystem ?? false)` on the `name` and `type` controls. There is no reusable predicate there to reuse, so the concept's SPELLING is the reuse: this fix reads the same `field.isSystem` flag (the one objectui#6044 made read the spec's `system` key), and it stays scoped to `delete`. `isSystem` has never meant "this row is untouchable" — the drawer opens for a system field and everything but `name`/`type` is editable — so withholding Edit as well would have been a capability regression wearing a bug fix's clothes. Pinned in both directions.

## Ablation — the pins die with the wiring they name

Each leg: mutate, prove the mutation is on disk, run both pins, restore, prove the restore by blob hash against HEAD.

| leg | mutation | grid pin | designer pin |
|---|---|---|---|
| (none) | — | 6 passed | 3 passed |
| A | the two `rowOps?.… !== false` conjunctions deleted from `ObjectGrid` | **4 failed** / 2 passed | **1 failed** / 2 passed |
| B | `rowOperations={rowOperations}` deleted from `FieldDesigner` | 6 passed | **1 failed** / 2 passed |

Leg B leaving the grid pin green is the correct signature, not a gap: that pin passes its own predicate and does not go through the designer. The two grid tests that survive leg A are the no-predicate CONTROLS, which assert the behaviour the ablation restores — a control that reddened there would be the broken one.

Leg A's first run reddened only 3 of 6: the intersection test set both members the same way, so "a row-level `true` cannot re-open what the grid closed" was also satisfied with the mechanism deleted, an unconsulted predicate being unable to widen anything either. That leg is the second commit here — the members now point in opposite directions, so the same test reads the wiring and dies with it.

## Verification

Run at `8f6848e3e`, the head of this branch.

- `pnpm exec vitest run packages/plugin-grid/ packages/plugin-designer/ packages/plugin-view/` — **182 files, 1572 tests, 0 failed**.
- `pnpm --filter @object-ui/plugin-grid --filter @object-ui/plugin-designer type-check` — both `Done`, after building the dependency closure (the first attempt was a PRECONDITION failure on an unbuilt tree, not a red).
- lint, the repo's own invocation (`eslint .`, per each package's `lint` script), both affected packages in full: **217 files, 0 errors** (plugin-grid 162, plugin-designer 55). Narrowed to the two packages the diff touches; the narrowing is sound because this config is not type-aware — `projectService` / `parserOptions` / `project:` match 0 times in `eslint.config.js`, against 8 for the lit control `rules:` — so no untouched file's verdict can move. The other 45 workspace projects go to CI.
- Gates: `check-changeset-presence` OK (1 changeset for 2 released packages), `check-control-bytes` OK, `check-readme-exports` OK (535/535 self-imports real, this PR's new README import among them), `check-doc-snippet-types` OK (644/644 blocks compile against the BUILT `dist/*.d.ts` — which is also what proves the new prop reached the shipped types), `check-doc-fences` OK, `check-doc-example-ids` OK, `check-new-cross-file-line-citations` 0 new, `check-phantom-dependencies` OK, `check-vi-mock-specifiers` OK. `check-governed-queue-guard --test` over the 8 changed paths: NOT GOVERNED.
- NOT MEASURED locally, left to CI: the repo-wide `turbo run lint` / `type-check` farm, the full `pnpm test` shards, E2E, and `apps/console` (an `ObjectGrid` consumer whose suite was not run here).

## Acceptance notes

Found while working, deliberately not fixed here:

- **filed as #9219** — `ObjectManager` has this defect verbatim (`if (!obj || obj.isSystem) return;` under a grid-level `onDelete`, with system objects shown by default). The repair is now three lines, and it was still kept out of this PR for a measured reason: `ObjectManager` is one of the untouched call sites this PR uses as its byte-identical control, and fixing it here would have consumed the control.
- **filed as #9220** — `scripts/check-readme-exports.mjs --list` crashes with a TypeError on an unbuilt tree, i.e. exactly the state its own failure message tells you to inspect. Named repro in the card; the plain run on the same tree reports cleanly, which is the control.
- **noted, not filed** — `packages/plugin-designer/src/__tests__/__mocks__/plugin-grid.tsx` draws a delete button per row whenever `onDelete` is wired and knows nothing about `rowOperations`, so a pin written through that double cannot see this fix. Not filed separately because it has a named taker: it is written into #9219 as the hazard for whoever fixes `ObjectManager`, which is the next PR that will touch it. This PR's designer pin renders the REAL grid for that reason.

⛔ Left as a draft deliberately, not enqueued, no auto-merge — the dispatching seat arms it.

---

---

### DIFF
--- a/.changeset/8674-objectgrid-row-operations.md
+++ b/.changeset/8674-objectgrid-row-operations.md
@@ -1,0 +1,39 @@
+---
+'@object-ui/plugin-grid': minor
+'@object-ui/plugin-designer': patch
+---
+
+Per-row operation gating for `ObjectGrid`, and the Field Designer stops drawing a
+delete action it refuses to run (objectui#8674).
+
+**The defect.** `FieldDesigner`'s `handleDelete` returned early on `field.isSystem` —
+before the confirm dialog — while the affordance was wired at the GRID level
+(`onDelete={readOnly ? undefined : handleDelete}`). `ObjectGrid` takes one grid-level
+`onDelete` and derives `{ update: !!onEdit, delete: !!onDelete }`, so the row action
+was drawn for every row: clicking delete on a system field produced no dialog, no
+toast and no console message. `readOnly` was honest in the same component (the
+callback is withheld, so no button is drawn); `isSystem` drew the button and dropped
+the click. The two states differed in the code and did not differ on screen.
+
+**`@object-ui/plugin-grid` — new, additive, opt-in.** `ObjectGridComponentProps`
+gains `rowOperations?: (record) => { update?: boolean; delete?: boolean }`, with the
+new `ObjectGridRowOperations` type exported from the package root. It speaks the same
+`update` / `delete` vocabulary the authored `operations` block speaks, resolved for
+one row instead of for the grid, and it is an INTERSECTION like every layer around it
+(the ADR-0103 lifecycle bucket, the object's `userActions`, the server's effective API
+operations, the principal's own grant, the record-level explain verdict): `false`
+withholds, and nothing it returns can re-open what those closed. A caller that passes
+no predicate renders exactly what it rendered before — measured, not asserted: the
+rendered DOM of a no-predicate grid is byte-identical across this change, and every
+other `<ObjectGrid>` call site in the repository is such a caller.
+
+It is a function value, so no metadata document can hold it and none is invited to:
+like the nine `on*` callbacks it sits beside, it is a renderer prop and not an
+authorable key.
+
+**`@object-ui/plugin-designer` — the first caller, and the user-visible fix.** The
+Field Designer passes the predicate, so a system field's row no longer offers Delete
+at all. Edit is untouched: the drawer still opens for a system field, with `name` and
+`type` disabled exactly as before — `isSystem` has never meant "this row is
+untouchable". The guard inside `handleDelete` stays as a second line for direct
+callers of the prop value, but it is no longer the only refusal.
--- a/content/docs/plugins/plugin-grid.mdx
+++ b/content/docs/plugins/plugin-grid.mdx
@@ -485,6 +485,47 @@ The declarative alternative, which *is* metadata and survives a round trip
 through storage, is `navigation`: its `mode` decides what a row click does
 without any host code.
 
+### Withholding a row's Edit or Delete
+
+`operations` and the `onEdit` / `onDelete` wiring are GRID-level: they decide
+whether the generic Edit / Delete entries exist at all, identically for every
+row. When the refusal belongs to one RECORD — a system field a designer may not
+drop — use `rowOperations`, a component prop called with a row record that
+answers for that row alone:
+
+```tsx
+import { ObjectGrid } from '@object-ui/plugin-grid';
+import type { ObjectGridSchema } from '@object-ui/types';
+
+const schema: ObjectGridSchema = {
+  type: 'object-grid',
+  objectName: 'field_definition',
+  columns: ['name', 'label', 'type'],
+};
+
+export const FieldList = ({ isSystem }: { isSystem: (name: string) => boolean }) => (
+  <ObjectGrid
+    schema={schema}
+    onEdit={(record) => console.log('edit', record)}
+    onDelete={(record) => console.log('delete', record)}
+    rowOperations={(record) => ({ delete: !isSystem(String(record.name)) })}
+  />
+);
+```
+
+It speaks the same `update` / `delete` vocabulary as the authored `operations`
+block, and it is an **intersection**, never a union: `false` withholds the
+entry, while `true`, an omitted member, and a `null` / `undefined` return all
+leave the grid's own verdict alone. Nothing it returns can re-open what the
+object's lifecycle bucket, its `userActions`, the server's effective operations,
+the principal's grant or the record-level verdict already closed — and a grid
+that passes no `rowOperations` renders exactly as it did before the prop
+existed.
+
+Prefer this over refusing inside the callback. A refusal that runs after the
+click ships a button that is drawn as available and then does nothing, which is
+indistinguishable from a broken build (objectui#8674).
+
 ### Inline Editing
 
 Enable inline cell editing for quick data updates:
--- a/packages/plugin-designer/src/FieldDesigner.tsx
+++ b/packages/plugin-designer/src/FieldDesigner.tsx
@@ -183,8 +183,37 @@ export function FieldDesigner({
     }
   }, [fields]);
 
+  /**
+   * [objectui#8674] Which generic row operations THIS row may be offered.
+   *
+   * The delete refusal used to live only inside {@link handleDelete}, which
+   * runs after the click: `ObjectGrid` takes ONE grid-level `onDelete` and drew
+   * the row action for every row, so a system field showed a delete button that
+   * answered a click with nothing at all — no dialog, no toast, no console
+   * message. `readOnly` was already honest (the callback is withheld, so no
+   * button is drawn); `isSystem` was not. This is the same refusal, moved to
+   * where it can withhold the affordance instead of swallowing its click.
+   *
+   * `isSystem` is the spelling the drawer form already uses to disable `name`
+   * and `type` on a system field — the one concept, "this field's structure is
+   * not the designer's to change", not a second one.
+   *
+   * An unresolvable row is withheld too, for the same reason and not as a
+   * defensive flourish: `handleDelete` returns early when the lookup misses, so
+   * offering delete for such a row would reproduce the exact defect this card
+   * is about.
+   */
+  const rowOperations = useCallback((record: Record<string, unknown>) => {
+    const field = fields.find((f) => f.name === record.name);
+    return { delete: !!field && !field.isSystem };
+  }, [fields]);
+
   const handleDelete = useCallback(async (record: Record<string, unknown>) => {
     const field = fields.find((f) => f.name === record.name);
+    // Unreachable through the grid now that `rowOperations` withholds the
+    // action for exactly these two cases — kept because this callback is a
+    // published prop value and nothing stops a future caller invoking it
+    // directly. It must never become the ONLY refusal again.
     if (!field || field.isSystem) return;
     const confirmed = await confirmDialog.confirm(
       t('appDesigner.fieldDesigner.deleteConfirmTitle'),
@@ -416,6 +445,7 @@ export function FieldDesigner({
         dataSource={dataSource}
         onEdit={readOnly ? undefined : handleEdit}
         onDelete={readOnly ? undefined : handleDelete}
+        rowOperations={rowOperations}
         onAddRecord={readOnly ? undefined : handleAddField}
       />
 
--- a/packages/plugin-designer/src/__tests__/FieldDesigner.systemFieldDelete-8674.test.tsx
+++ b/packages/plugin-designer/src/__tests__/FieldDesigner.systemFieldDelete-8674.test.tsx
@@ -1,0 +1,151 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#8674 — the Field Designer offers no delete action on a system field.
+ *
+ * ## The card
+ *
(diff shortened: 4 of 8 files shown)


os-tesla:
## Contract review — head `8f6848e3eb` — **PASS**

Reviewed-by: `domain:ui` PM seat (`os-tesla`), R16, 2026-09-12T01:5xZ.

**Clause-② `yes`** — `--pair 9224` reads ✓ on both carriers (PR #9224 / card objectui#8674), declaration readable in the fixed spelling, carriers agree. ⛔ Not revised on delivery.

**CI: 36 checks, 0 pending, 0 failing.** ⭐ Including `Bundle Analysis`, and that is worth one line: it is currently RED on objectui#9078 because `check-eager-locale-catalogues.mjs` mis-parses a chunk whose content hash contains a `-` (fix open as objectui#9228). This PR touches no locale pack, so its hashes did not reroll — which is the same diagnosis read from the other side, and a small confirmation of it.

### What was delivered against what was dispatched

Dispatched as **arm 1** (do not draw the control), with the note that the real landing place is `plugin-grid`'s per-row operation predicate and `FieldDesigner` is merely its first caller. That is what came back: `ObjectGridComponentProps` gains `rowOperations(record) => { update?, delete? }` with an exported `ObjectGridRowOperations`, **ANDed** into the same actions-cell expression that already folds objectui#2614's predicates and objectui#4296's record verdict.

⭐ The intersection is the load-bearing choice and it is the right one: `false` withholds, and nothing the new predicate returns can re-open what a layer above already closed. A new predicate that could widen would have been a permission defect wearing a feature's clothes.

### Bump call — agreed, and it is two different answers on purpose

- `@object-ui/plugin-grid` **minor** — a new published prop and a new exported type. Additive surface.
- `@object-ui/plugin-designer` **patch** — the lane's test is *does existing stored data render differently*, and here the affected configuration was **already broken**: the Delete control was drawn on system-field rows and then silently swallowed the click. Same reading as objectui#9200.

### The evidence I am accepting, and why

- **Control, byte-identical rather than asserted.** Two no-predicate callers dumped in full — container plus each row's opened kebab — on this branch and again with the three changed sources at base `b775500af`: **92191 bytes both legs, sha256 equal**, with `row-action-builtin-delete` lit 4× in each dump. ⇒ a caller that passes no predicate is untouched, proven rather than claimed.
- **Ablation, two legs**, each mutate → prove on disk by blob hash → run → restore → prove restoration by blob hash. Leg A (conjunctions deleted): grid pin 4 failed / 2 passed, designer pin 1 failed / 2 passed. Leg B (prop pass deleted from `FieldDesigner`): grid pin unaffected — **correctly**, it drives its own predicate and does not go through the designer — designer pin 1 failed.
- ⭐ **The seat caught its own vacuous test and fixed it.** Leg A first reddened 3 of 6 because the intersection test was vacuous under ablation — an unconsulted predicate cannot widen anything either. Rewritten so its two members point in opposite directions, it now dies with the wiring. That is rule 2 applied to its own pin: a test that cannot fail is not evidence.
- The 2 grid tests that survive leg A are the **no-predicate controls**, which assert what the ablation restores. Nothing survives deletion of the thing it names.

### Out-of-scope findings — accepted as filed, ⛔ correctly not fixed here

- **objectui#9219** — `ObjectManager` carries this defect verbatim. ⛔ Kept out deliberately, because it is one of the two untouched call sites used as the byte-identical control. Fixing it here would have destroyed the control that proves this change is contained.
- **objectui#9220** — `check-readme-exports.mjs --list` crashes on an unbuilt tree, i.e. exactly the state its own failure message tells you to inspect.
- ⚠️ Noted and carried onto objectui#9219: `plugin-designer`'s `__mocks__/plugin-grid.tsx` draws a delete button per row whenever `onDelete` is wired and **ignores `rowOperations`**, so a pin written through that double cannot see this fix. The hazard is written into the card that will next touch it.

---

PR #9234, opened 2026-09-12
fix(core,plugins): honour only the `data` spelling each block's spec row declares

os-tesla:
Part of #8348 — the ladder half of decision batch #83. ⛔ Deliberately not a closing keyword: the card also carries findings this PR reports rather than repairs (see the last section), so the PM lands it and decides the card's state.

## The ruling this implements

Decision batch #83 (2026-09-08), maintainer verbatim, kept in the original and untranslated:

> 8348 以协议为准

⇒ option A. A renderer honours the `data` spelling its block's **published row** declares and no other: what `os validate` and the save gate refuse, the renderer refuses too. Where a pin contradicts its block's row, the **pin** is corrected to the spec — never the spec to the code.

## What changed

`resolveRecordSourceConfig` (`@object-ui/core`) now takes a second, **required** argument: the arm the calling block's published `data` row declares (`view-data`, `array`, or `undeclared`). Rung 1 of the ruled three-rung ladder applies only on that arm; rungs 2 and 3 are untouched, as is their order.

The arm is passed at the call site rather than looked up from `schema.type`. Every one of these renderers is registered **twice** — `object-grid` plus the `view:grid` alias `grid`, `object-calendar` plus `calendar`, and so on — so a table keyed by type would answer for one tag and silently miss the other. A required parameter makes the arm a compile-time obligation at each site instead; it is what caught all five sites in one `type-check` run.

### The delivery surface: all five renderers, enumerated from the ladder itself

`git grep -n resolveRecordSourceConfig -- packages apps examples` returns exactly five calling renderers — `ObjectCalendar.tsx`, `ObjectGantt.tsx`, `ObjectGrid.tsx`, `ObjectMap.tsx`, `ObjectTree.tsx` — which is also the list the shared function's own docblock names. `object-kanban` is deliberately **not** among them: it walks its own resolver, pinned by `packages/types/src/__tests__/object-kanban-record-source-7780.test.ts`, which asserts zero `resolveRecordSourceConfig` in that block.

### Per block, judged against its own row

Rows measured on `@objectstack/spec` **17.4.0**, which is what `pnpm-lock.yaml` resolves for this repo, read from `node_modules/@objectstack/spec/dist/ui` in this worktree after `pnpm install` — not from an `objectstack` checkout, so the reading is the published package.

| block | published `data` row | arm | what moves |
|---|---|---|---|
| `object-calendar` | `ComponentPropsMap` row: `z.array(z.unknown()).optional()`, "Pre-fetched records — skips the internal fetch" | `array` | the `{ provider, … }` config object stops being a record source |
| `object-grid` | `ComponentPropsMap` row: the `ViewData` union; its own description says "the bare-array shortcut is refused" | `view-data` | `getDataConfig`'s `Array.isArray` head removed |
| `object-map` | no `ComponentPropsMap` row; `ObjectMapSchema.data` = `ViewDataSchema.optional()` | `view-data` | its array-shorthand head removed |
| `object-gantt` | no `ComponentPropsMap` row; `ObjectGanttSchema.data` = `ViewDataSchema.optional()` | `view-data` | a bare array stops being a record source (it was already inert) |
| `object-tree` | **no row on any face** | `undeclared` | nothing — see "reported, not fixed" |

## ⚠️ The accepted cost — and ⛔ a measured limit on where it lands

**`object-calendar`**: a stored calendar authored `data: { provider: 'value', items: [...] }` (or any other provider object) stops drawing those rows. The ladder falls past `data` to `staticData`, then to `objectName`, so it **queries its object instead**, or draws nothing when it names neither. Accepted by the ruling under the standing 2026-08-27 posture: no transition windows, no staged deprecation. Pinned end-to-end, through the card's own harness.

**`object-grid` and `object-map`**: ⛔ MEASURED, and it corrected this PR's own first draft. Removing the ladder head is real **at the ladder** and is **not**, on its own, an end-to-end change for a node rendered through `SchemaRenderer`. An authored `data` array arrives **twice**: as `schema.data`, where the head used to lift it, and as the `data` **React prop**, because `SchemaRenderer` spreads every non-metadata node key and both wrappers forward it — and each component lifts an array on that prop at *higher* priority than the ladder (`ObjectGrid`'s `passedData`; `ObjectMap`'s objectui#5003 precedence). So such a node still draws its rows. The first version of the grid pin asserted the opposite and went red, which is how this was found; both pins now state the verdict per carrier.

That surviving carrier is the one a host such as `ListView` legitimately uses for rows it pre-fetched, and at the component boundary it is indistinguishable from an authored key. Collapsing the two would take the host path with it, so it is **reported** (below) rather than changed in passing.

**`object-gantt`**: nothing observable moves — the array carried no `provider`, matched no fetch branch, and this block's wrapper forwards no host props.

⇒ the ruling's user-visible cost lands on `object-calendar`, which is where the card was filed. On the object-armed blocks this PR removes the second read, not the last one.

## Authored usage of the tolerated spellings — swept, and REPORTED not migrated

Per the ruling the seat reports real authored instances rather than silently migrating them. **The sweep found none outside test fixtures.** Counts read before paging, no truncated enumeration:

- `{ provider, items }` under `data` on `object-calendar`: intersecting the 84 files naming `object-calendar` with the files naming `provider` gives 43 files, all of them source, tests, changelogs or prose — zero authored documents. No `examples/**` calendar schema carries a `"provider"` key at all.
- bare array under `data` on `object-grid` / `object-map` / `object-gantt` / `object-tree`: 98 hits for a `data:`-followed-by-`[` shape across `content`, `examples`, `apps`, `e2e`, `public`; every one is a `find` response envelope, a `data-table` / `chart` / `kanban` block (none of which is on this ladder), or a calendar — where the array is the **declared** spelling. Zero on the four object-armed blocks.
- The two calendar catalog entries (`examples/schema-catalog/src/schemas/plugin-calendar/month-view-calendar.json`, `week-view-calendar.json`) author `"data": [ … ]`, which is the arm the spec row declares. They render exactly as before.

What the sweep did move is **test fixtures**, and each carries the reason in place: calendar fixtures that spelled inline rows as `data: { provider: 'value', items }` now use `staticData` (the block's declared inline-rows door, resolving to the identical config one rung down); the calendar memo fixture that spelled its binding as `data: { provider: 'object', object }` now uses `objectName`; the map memo fixture now uses `data: { provider: 'value', items }`.

## Pins corrected to the spec (⛔ never the spec to the code)

- **`packages/plugin-grid/src/__tests__/gridDataInputContract.test.ts`** — named by the ruling. Its assertions already agreed with the row; two prose claims did not, and both are corrected: (1) it said the parity gate "judges against `ComponentPropsMap['object-grid'].data`, which is `z.array()` and DISAGREES" — measured on 17.4.0 the two authorities have **converged**, the row is the `ViewData` union; (2) it said "the renderer's array tolerance is NOT asserted away here. It stays as back-compat" — the ruling removes the tolerance, so declaration and read now say the same thing.
- **`packages/core/src/utils/record-source.ts` docblock** — named by the ruling. It stated as a fact about the whole ladder that "an array under `data` cannot be published". True of the `ViewData` blocks, **false** of `object-calendar`, whose row *is* the array. Replaced with the measured per-block table.
- **`packages/plugin-map/src/ObjectMap.schemaDataShorthand.test.tsx`** — its subject was the array shorthand, justified by "a live convention in six sibling blocks", an argument from what the code does. Rewritten around the row: the shorthand's retirement is pinned, and the objectui#5305 rows survive as the ⛔ CONTROLS (the declared form still paints, still takes the no-fetch path, still repaints on changed rows; the `data` prop still outranks the schema).
- **`packages/plugin-calendar/src/__tests__/ObjectCalendar.navigationRecordSource-7638.test.tsx`** — two rows pinned that `data: { provider: 'object', object }` re-points the calendar's record source. On this block that spelling is off-arm, so both are rewritten to the ruled outcome, with a new control showing the declared doors still route.
- **`packages/types/src/__tests__/object-calendar-record-source-7313.test.ts`** — it source-greps the ladder for rung order. Kept, re-aimed at the new spelling, and one **false green** repaired while in there: its "the renderer still calls the ladder" row looked for the bare `resolveRecordSourceConfig(schema)`, which `ObjectCalendar.tsx` satisfies from a **docblock line** that merely names the function — so it would have stayed green through a call site that had stopped existing. It now asserts the call *with its arm*.

## New pins

- `packages/core/src/utils/__tests__/record-source-config.behaviourNeutrality-7632.test.ts` (extended) — the arm verdict at the resolver, with the pre-collapse bodies kept as the ⛔ CONTROL that makes "changed" a measurement, plus a non-vacuity row proving the ruled-away family is non-empty.
- `packages/plugin-calendar/src/__tests__/ObjectCalendar.dataObjectArmRefused-8348.test.tsx` — the card's **own** harness: a real node through the real `SchemaRenderer` with a recording data source. Row 1 is the card's measurement inverted; row 2 shows the accepted cost on screen (the block now queries); rows 3-5 are controls holding the declared carriers still; row 6 reads the spec row rather than restating it.
- `packages/plugin-grid/src/__tests__/gridBareArrayDataRefused-8348.test.tsx` — the same shape for the grid's half of the ruling.

## Verification

All numbers below are from the final head, `81fe3e18f`, which already contains the merge of `origin/main` (objectui#9224 / objectui#8674 landed; the one plausible meeting point, `packages/plugin-grid/src/index.tsx`, produced no conflict). Every test command runs from the repo root — a package-directory `vitest` is refused by this repo's own guard — and through the shared verify lock. Exit codes captured by redirect-then-capture, never through a pipe.

| what | command | verdict |
|---|---|---|
| tests, six packages | `pnpm exec vitest run packages/core/ packages/plugin-calendar/ packages/plugin-map/ packages/plugin-gantt/ packages/plugin-tree/ packages/types/` | `Test Files 466 passed (466)` · `Tests 7650 passed (7650)` |
| tests, grid + the merge's other touched package | `pnpm exec vitest run packages/plugin-grid/ packages/plugin-designer/` | `Test Files 146 passed (146)` · `Tests 1262 passed (1262)` |
| the two console parity gates | `pnpm exec vitest run apps/console/src/__tests__/registry-inputs-spec-parity.test.ts apps/console/src/__tests__/public-block-binding-reach.test.tsx` | `Test Files 2 passed (2)` · `Tests 214 passed (214)` |
| types, seven packages | `pnpm --filter … run type-check` | exit 0, no `error TS` |
| lint, seven packages (the real gate — `pnpm lint` is `turbo run lint`) | `pnpm --filter … run lint` | **0 errors** in each; warnings are pre-existing `no-explicit-any` / `react-refresh` noise |
| changeset declared | `node scripts/check-changeset-presence.mjs` | `✅ 23 source file(s) of 7 released package(s) changed, and this change declares 1 changeset(s)` |
| no `major` bump | `node scripts/check-changeset-no-major.mjs` | `✅ No changeset declares a 'major' bump.` |
| governed surface | `node scripts/check-governed-queue-guard.mjs --test …` | `✅ NOT GOVERNED — 19 path(s) checked against 5 governed surface(s); none matched.` |
| control bytes | `grep -naP` over the diff for `[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]` | no hits |

The fixture triage is visible in the run history rather than hidden: the six-package run reported **16** failures on its first pass, **2** on its second and **0** after the last two assertions were re-aimed; `packages/plugin-grid/` went **23 → 1 → 0** the same way. Every one was a fixture or a pin whose subject the ruling moves, and each carries its reason in place.

**The repo-wide `eslint .` reading, with its three parts so the narrowing above is a measurement and not an omission.** (1) Population read from eslint's own config, not guessed: `npx eslint . --no-inline-config --format json` reports **4855** files. (2) All **23** of this PR's changed lintable files are inside that population, and linting exactly those reports **0 errors** (498 warnings, all pre-existing rule noise on files that already carried it). (3) The flat run reports 95 errors across 79 files repo-wide; the intersection with this PR's changed set is **empty**, and those files are untouched here. That flat run is not the CI gate in any case — `pnpm lint` is `turbo run lint`, the per-package scripts already reported above.

⛔ Not measured here, by design: CI's own convergence. The lint and test farms are CI's run, and this report is delivered when the local checks are done rather than after waiting on them.

Bump level: **minor**, chosen not copied. The lane's test — does existing stored data render differently — is demonstrably YES here, which is breaking. `major` is unavailable by repo rule: the 39-package fixed group tracks `@objectstack`'s major, and `scripts/check-changeset-no-major.mjs` fails any changeset declaring one. So the repo's own idiom applies: `minor`, with the breaking semantics spelled out in the changeset body, per block.

## Reported, not fixed

1. **`object-tree` has no published `data` row on any face** — not `ComponentPropsMap` (no `object-tree` key exists in it at all), not `ObjectTreeSchema` in `@object-ui/types` (which declares `objectName` **required** and no `data` / `staticData`), not the plugin's registration `inputs`. Both arms of the ruling decide **by reference to the block's row**, so neither reaches this block, and inventing one would be a contract decision rather than an implementation of this one. Its rung 1 is therefore left exactly as it was, marked `undeclared` at the call site with the reasoning in place. The block still honours both a `ViewData` object and a bare array under `data` — and it has a third carrier besides, a `rest.data ?? schema.data` array read inside its own fetch effect.
2. **`object-map` and `object-gantt` have no `ComponentPropsMap` row either.** They are decided here because a published row *does* exist for them — this repo's own `ObjectMapSchema.data` / `ObjectGanttSchema.data`. Worth a maintainer's eye: three of the five ladder blocks have no spec row, so "judge against the spec row" resolves through `@object-ui/types` for two of them and not at all for the third.
3. **The props channel is a second, unruled carrier of the same authored key.** `SchemaRenderer` spreads every non-metadata node key as a React prop, so `data` authored on a node reaches a block both as `schema.data` — which this PR now judges against the block's row — and as the `data` prop, which several blocks lift with an unconditional `Array.isArray`. The two are indistinguishable at the component boundary, because the prop is also how a host hands down rows it already fetched. `packages/plugin-calendar/src/__tests__/ObjectCalendar.recordSourceMembers-8314.test.tsx` already documents this arrangement for the calendar and warns that a pin written from the read that looks right cannot fail. Deciding it is a contract question about the props channel, not about the ladder, so it is left whole here and reported.
4. **`ObjectCalendarSchema.data` in `@object-ui/types` contradicts the spec row — filed as objectui#9239.** Both faces of this repo's own mirror — `packages/types/src/zod/objectql.zod.ts` and `packages/types/src/objectql.ts` — declare `data?: ViewData` for `object-calendar`, while `ComponentPropsMap['object-calendar'].data` is `z.array(z.unknown())`. Under the ruling's direction the mirror follows the spec, but this is a **declaration** on a published contract package, not a pin: moving it changes what `safeValidateSchema` accepts from authors, and it is held in place by a type-level equality with `object-gantt`, by the objectui#7313 document fixtures, and by the plugin docs page. Filed as objectui#9239 rather than folded in, so the contract change gets its own review. It is not a defect this PR introduces — the divergence predates it, and the renderer's read is now on the spec's side of it.

This pull request was generated by Claude Code, in session `session_01UzHd6hDYatoDn17BuwKxnZ` — recorded here as prose because a `PATCH` rewrites the footer block (AGENTS.md, "GitHub 会改写你写进 issue/PR 正文的字节").

---


os-tesla:
## PM ruling on both open questions, and Contract review — head `81fe3e18ff` — **PASS**

Reviewed-by: `domain:ui` PM seat (`os-tesla`), R16, 2026-09-12T03:1xZ.

**Clause-② `yes`** — ⛔ `--pair 9234` first read **C1**: the PR carried `needs:contract-review` and the card did not. **That was mine, not this delivery's** — see the carrier note at the end. Repaired and re-read: **✓, both carriers agree.** **CI: 36 checks, 0 pending, 0 failing.**

---

## Q1 — three of five blocks have no `ComponentPropsMap` row → **A now, C filed as the durable fix**

⭐ This falsified a mechanism assumption in my own dispatch. I wrote *"judge each block against its own spec row"* as if every block had one. Three do not. The seat measured that instead of guessing, and reported it as a falsification rather than quietly picking a rule — which is exactly what the dispatch asked for and rarely gets.

**Ruled A**, which is what shipped: the spec row decides where one exists (`object-calendar`, `object-grid`); this repo's own **published** row decides where it does not (`object-map`, `object-gantt`, both `ViewDataSchema`); and a block with no row on **any** face (`object-tree`) is left alone and **reported**.

⛔ Not B. "No published row ⇒ not a record source" is a **behaviour change on off-contract documents that no ruling licenses**. The seat's own note that it "contradicts nothing, because there is nothing to contradict" is the argument against it, not for it: an unruled silent narrowing is the thing this lane files cards about.

⛔ Not C in this PR — it changes `@objectstack/spec`, which routes to the spec seat.

⚠️ **Recorded as a deviation from the ruling's letter, not hidden.** 「8348 以协议为准」 and two of the five blocks were decided by *this repo's* published row rather than the protocol's. The ruling's **intent** is satisfied — judge by a published declaration, never by the code — but the asymmetry is real, and I am carrying it to the maintainer rather than letting A stand as if it were uniform. **C is the only option under which "the spec decides" is true for all five.**

## Q2 — the props channel is a second, unruled carrier → **A, with the limitation stated plainly**

`SchemaRenderer` spreads every non-metadata node key as a React prop, so an authored `data` reaches a block **twice**: as `schema.data`, now judged against the row, and as the `data` prop, which `ObjectGrid` (`passedData`) and `ObjectMap` lift with an unconditional `Array.isArray` at **higher** priority than the ladder. Measured: on those two blocks an authored bare array **still draws** after this PR.

**Ruled A.** ⛔ Not B — it deletes a working host path (a host such as `ListView` handing down pre-fetched rows) to close an authoring gap, and the two are indistinguishable at the component boundary. C is the shape worth costing and it is filed.

⚠️ **So the ruling does NOT fully bite on `object-grid` and `object-map`, and I will not let that be discovered later.** What saves the card is an asymmetry the seat found and named: `object-calendar`'s boundary keeps `data` **only** when `Array.isArray`, so the `{provider, items}` config this card is actually about had exactly **one** carrier, and its removal **is** end-to-end. ⇒ the filed defect is fully closed; the generalisation to grid and map is **partial**, by a carrier the ruling never addressed.

---

## The delivery

**The five renderers were enumerated from the ladder itself**, not from the card: `git grep -n resolveRecordSourceConfig -- packages apps examples` returns exactly `ObjectCalendar`, `ObjectGantt`, `ObjectGrid`, `ObjectMap`, `ObjectTree` — the same list the shared function's docblock names — with `object-kanban` excluded by its own pin asserting zero calls.

⭐ **The arm is a REQUIRED parameter passed at the call site, not dispatched on `schema.type`** — because every one of these renderers is registered **twice** (`object-grid` plus the `view:grid` alias, `object-calendar` plus `calendar`), so a type-keyed table would answer for one tag and miss the other. And making it required is what **proved** every call site had been found: the first type-check named exactly the one stale caller. That is a compiler used as a census, and it is stronger than any grep would have been.

Spec read from `node_modules` at **17.4.0** after `pnpm install` in the worktree, ⛔ not from the objectstack checkout — the instrument caveat honoured.

### Two false greens repaired in passing, both worth more than the lines they touch

- `object-calendar-record-source-7313.test.ts` asserted the renderer "still calls the shared ladder" by looking for the bare string `resolveRecordSourceConfig(schema)` — which `ObjectCalendar.tsx` satisfied **from a docblock line that merely names the function**. ⇒ it would have stayed green through a call site that had stopped existing. ⭐ **Measure the CALL, not the NAME** — the same rule that bit me twice this round, found here by a seat in someone else's pin. Repaired rather than filed, correctly: this PR's own edit is what made it live.
- `gridDataInputContract.test.ts`'s docblock claimed `ComponentPropsMap['object-grid'].data` "is `z.array()` and DISAGREES" with `ViewDataSchema`. Measured on 17.4.0 the two authorities have **converged**. Corrected under the ruling's clause 3 — ⛔ the pin moves to the spec, never the spec to the pin.

### Triage history kept visible

Run (1) went 16 failures → 2 → 0; plugin-grid went 23 → 1 → 0. ⭐ And the grid pin's first draft asserted an end-to-end behaviour change, **went red, and that red is what produced open question 2**. A pin that fails honestly and is then narrowed to what is true is worth more than one written to pass.

### Bump

`23 source file(s) of 7 released package(s) changed`, 1 changeset, no major. Existing stored documents authored in the tolerated spelling **stop rendering** — the cost the ruling accepted under the 2026-08-27 no-transition-window posture — and the changeset states the change per block, as instructed.

---

## ⛔ The carrier failure was mine, and it was a second instance

`--pair 9234` opened **C1** because card objectui#8348 had lost `needs:contract-review`, `pm:dispatched`, its `finding` label **and its assignee** — the whole write reverted to the pre-claim state. I wrote them at claim time and ⛔ **did not read back**, which is this lane's rule 13 verbatim (a label PUT gets clobbered by the size-labeler's whole-set write) and the one card of the four where I skipped the read-back. Restored and read back **twice**, twelve seconds apart, both stable; assignee restored and confirmed.

⇒ Two carrier defects on my side this round, from the same root: **a write is not a write until it is read back, and a record is not a record until the predicate that reads it can find it.**

Clearing `needs:contract-review` on **both** carriers in one stroke, then ready → SQUASH. ⛔ Card objectui#8348 is **not** closed by this PR.


---

PR #9250, opened 2026-09-12
perf(components,app-shell,console): answer icon-name membership from lucide's `icons` record — MEASURED −8,515 B, `ui-components` back to green

### LINKED ISSUE #9204: perf(components): the eager `ui-components` chunk carries lucide's 1,767-entry dynamic-import map and the whole `icons` record — 45,749 B / 144,361 B gzipped, both measured
Filed **bare** (no assignee, no `domain:*`, no `pm:*`) out of the measurement-first round on objectui#7848, whose fence says a slimming that the census finds "is a second card and a second PR". This is that card. ⛔ Nothing here has been implemented and ⛔ no gate constant was moved.

All figures below are from four full console builds in one container on tip `1f4e02995a`, read directly out of `apps/console/dist/eager-closure.json`. Every mutation was proved on disk (blob-hash before/after plus printed grep anchors) and restored with `git checkout HEAD -- path`, proved by an empty `git diff HEAD`.

## The reading on `1f4e02995a`

```
ui-components   395,032 gz   ceiling 399,000   headroom 3,968 = 0.0435x the 91,136 B regression
aggregate     3,186,729 gz   ceiling 3,210,000 headroom 23,271 = 0.2554x
```

`ui-components` is the only entry in `EXHAUSTED_HEADROOM_ALLOWANCES` (4,289). The gate itself prints the trip point: **`reds below 3378 bytes`** — so the live row has **590 bytes of runway** before the ratchet fires. Clearing the 0.10x floor on its own needs the chunk at 389,886 or less, i.e. a saving of **5,146 gzipped bytes**.

## Where the bytes are

Census of the emitted chunk's own module list (2,240 modules, 2,926,967 B renderedLength):

| area | renderedLength | share | modules |
|:--|--:|--:|--:|
| `lucide-react` | 1,480,797 | **50.6%** | 1,770 |
| `packages/components/src/renderers` | 396,868 | 13.6% | 105 |
| `packages/fields/src/widgets` | 280,718 | 9.6% | 72 |
| `packages/components/src/ui` | 112,206 | 3.8% | 47 |
| `packages/components/src/custom` | 104,106 | 3.6% | 29 |
| `react-day-picker` | 99,431 | 3.4% | 98 |
| `react-hook-form` | 80,343 | 2.7% | 1 |

Half the chunk is lucide, through two independent doors.

### Door 1 — the dynamic-import map is EAGER (`dynamicIconImports.mjs`, 263,547 B rendered)

Four non-test modules statically import `lucide-react/dynamic.mjs`:

```
packages/components/src/lib/lazy-icon.tsx:24
packages/app-shell/src/views/metadata-admin/widgets.tsx:46
packages/app-shell/src/utils/getIcon.ts:16
apps/console/src/utils/getIcon.ts:11
```

The map is 1,767 `() =` arrow entries whose only job is to be called later. It does not need to be in the first payload.

**Measured** (ablation: the four imports stubbed, `icons` record left alone):

```
ui-components  395,032 -> 349,283   -45,749
aggregate    3,186,729 -> 3,141,057 -45,672
eager/total chunks  51/528 -> 51/528   (unchanged)
```

⭐ The two deltas agree within 77 bytes, so this is bytes LEAVING the page load, not bytes moving between columns. No new chunk, no extra request. `ui-components` headroom becomes **49,717 = 0.55x** — above the 0.10x floor and still under 1.00x, so the ceiling stays sensitive rather than going blind. **This one change alone is 8.9x the 5,146 bytes the row needs.**

### Door 2 — the whole 1,767-icon record is EAGER

`packages/components/src/renderers/action/resolve-icon.ts:9`

```ts
import { icons, type LucideIcon } from 'lucide-react';
```

`icons` is lucide's runtime record, published as `export { index as icons }` over a namespace import of a 1,767-line barrel. A namespace object has no dead members, so tree-shaking cannot help: every icon module is statically reachable. Verified on the emitted artifact rather than from source — the path data of `air-vent`, an icon this product does not use, appears exactly once in `assets/ui-components-*.js`.

It is also the ONLY non-test module that imports the record (the three other hits are test files), which matches `check-lucide-icon-record-names.mjs`'s declared census of "this file alone".

**Measured** (ablation: `resolveIcon` returns null, record import removed):

```
ui-components  395,032 -> 254,749  -140,283
aggregate    3,186,729 -> 3,188,038  +1,309   (WORSE)
eager/total chunks  51/528 -> 176/2138
```

⚠️ The aggregate gets worse, and the cause is the co-tenancy mechanism objectui#6680 documented. Diffing the chunk membership shows **eight icon modules**, well under 3 KB, were placed inside three previously-lazy plugin chunks, which drags all of them eager:

```
plugin-dashboard  + arrow-up-right.mjs                                     -> 49,182 gz eager
plugin-gantt      + file-down, minimize-2, panel-left-close,
                    wand-sparkles, zoom-in, zoom-out                       -> 36,561 gz eager
plugin-report     + table-2.mjs                                            -> 10,417 gz eager
```

Adding one `advancedChunks` group per icon — the shape objectui#7479 used for the ten locale catalogues, and for the same stated reason — removes that:

```
ui-components  395,032 -> 250,671  -144,361
aggregate    3,186,729 -> 3,126,208  -60,521
eager/total chunks  51/528 -> 328/2294
```

⚠️ The price is **278 eager per-icon chunks summing 77,024 gz**, i.e. 278 extra requests on first paint, and the `ui-components` headroom becomes 148,329 = **1.63x**, which trips the gate's blindness leg in the other direction and would need the ceiling and its baseline re-pinned DOWN together. Not proposed as-is; the shape worth designing is one shared chunk for the eagerly-used icons plus per-icon chunks for the rest.

## Ranked

1. **Door 1**, lazy-load `lucide-react/dynamic.mjs`. -45,749 on the row, -45,672 on the aggregate, zero new chunks, ceiling stays sensitive. Clears the floor on its own. ⚠️ The number above is an ablation, so it is an UPPER bound: a real implementation keeps an async loader and some cheaper `iconNames` membership predicate.
2. **Door 2**, retire the `icons`-record import in favour of the async resolver `lazy-icon.tsx` already implements. Bigger, but needs the chunking design first and a ceiling/baseline re-pin after. ⚠️ Contract change: `resolveIcon` is a maintainer-ruled synchronous seam (objectui#5935, ruling of 2026-09-03) returning `LucideIcon | null`, and four surfaces depend on their own fallbacks. That ruling has to be revisited before this is touched.
3. ⛔ **Not the `packages/fields` barrel.** `scripts/vite-ineffective-dynamic-imports.ts` already measured barrel surgery at **+329 bytes** and the maintainer ruled option C on 2026-08-22. Its option A is still on hold and is a prerequisite, not an alternative.
4. ⛔ **Not a re-chunk of `react-day-picker` / `react-hook-form` / `vaul` / `date-fns` / `embla-carousel` / `react-resizable-panels`.** These are in `ui-components` only because they match no vendor group; moving them shrinks the column without removing a byte, and the console config already rules on that in its own words: "This moves NO MODULE BYTES ... Nothing here may be read as headroom that was earned."

## Reproduction

```
pnpm install --frozen-lockfile
pnpm turbo run build --filter='./packages/*'
pnpm --filter @object-ui/console build
node -e "const r=require('./apps/console/dist/eager-closure.json');console.log(r.eagerGzipBytes, r.files.find(f=>f.name==='ui-components').gzipBytes)"
```

---

os-tesla:
Fixes #9204

Implements the maintainer's ruling of 2026-09-13 (decision batch #125 item 3, comment 5650203001), **option A**: one icon vocabulary — the `icons` record that actually ships. This pull request was previously the DO-NOT-MERGE measurement of record for Door 1; ruling item 1 turns it into the A implementation, and the refuted Door-1 measurement is kept below because it is the reason A is the shape.

`main` was RED on `Bundle Analysis` for this row and every other lane inherited it. It is green here.

## The measurement — control and result, two console builds, one container

Instrument each time: `pnpm turbo run build --filter='./packages/*' --concurrency=2` then `pnpm --filter @object-ui/console build`, then `apps/console/dist/eager-closure.json` read directly. Both runs serialised through the shared verify lock, both `VERDICT command-exit 0`.

| build | `ui-components` gz | Δ row | aggregate gz | Δ aggregate |
|:--|--:|--:|--:|--:|
| control — `main` at `852437297` | 397,090 | — | 3,180,420 | — |
| **this branch at `d43f51ee1`** | **388,575** | **−8,515** | **3,171,900** | **−8,520** |

⭐ **The two deltas agree within 5 bytes.** That is the reading triage said it was relying on: these are bytes *leaving the page load*, not bytes moving between columns — the failure mode the console config forbids in its own words ("This moves NO MODULE BYTES … Nothing here may be read as headroom that was earned").

The control reproduces CI exactly: 397,090 B is 387.8 KB, the figure the `Bundle Analysis` job printed for `main` at `85243729`.

### `check:eager-closure` — exit 2 before, exit 0 after

```
BEFORE (control, main 852437297)      BUDGET_EXIT=2
  ❌ chunk `ui-components`  387.8 KB measured / 389.6 KB ceiling
     (headroom 1.9 KB = 0.02x the 89.0 KB regression, under the 0.10x floor and held
      open by its declared 4289-byte allowance, which may only be paid DOWN;
      reds below 3378 bytes)

AFTER (this branch, d43f51ee1)        BUDGET_EXIT=0
  ✅ chunk `ui-components`  379.5 KB measured / 389.6 KB ceiling
     (headroom 10.2 KB = 0.11x the 89.0 KB regression)
```

Every other row unchanged and green: aggregate 0.42x, `vendor-objectstack` 0.15x, `i18n-locale-en` 0.10x, `framework` 0.61x.

## Gate constants: one retired, one re-pinned, ⛔ no ceiling moved

- ⛔ **No ceiling moved.** `PER_CHUNK_GZIP_CEILINGS['ui-components']` is still 399,000, unchanged since objectui#5490. The maintainer's raise authorisation for this card is **unspent** — the row pays for itself, so there is no residue to raise for.
- ⛔ **No allowance was raised.** `EXHAUSTED_HEADROOM_ALLOWANCES` loses its `ui-components` entry — **REMOVED, not lowered**, which is the one way out of that table and the end state its own docblock names for `i18n-locales`: "the debt PAID, in the only currency this table takes: the row cleared the floor on its own."
- ⭐ Removal was the **only legal move**, and the gate's own unit test says so rather than this paragraph: *"every entry is real debt — strictly under the floor it excuses."* At 10,425 B the row is above the 9,114 B floor, so an entry would no longer be debt — it would be a *weaker* second floor for one row (declared trip point 3,378 B against the floor's 9,114 B), while the verdict text kept printing "under the 0.10x floor" about a row at 0.11x.
- `PER_CHUNK_BASELINE['ui-components']` moves 391,095 → 388,575 **in the same commit as the removal**, and it had to: the gate's unit fixtures are built from that constant, so leaving it would have made the gate's own tests assert a row under a floor the live build clears.

⚠️ **What that costs, stated rather than buried:** the row is now judged at the 0.10x floor with **1,311 B of margin**. That is deliberate — it is the tightening the ratchet is for — but it is thin, and this chunk moves under traffic that has nothing to do with it. Door 2 is where the remaining bytes are (see below).

## What changed, and why the bytes were there

lucide derives `iconNames` as `Object.keys(dynamicIconImports)`, so **importing the icon names imports the dynamic-import map with them**. Four modules did, and the map sat in the console's eager `ui-components` chunk on every page load.

The names leave that path by being answered from the `icons` **record**, which `renderers/action/resolve-icon.ts` already puts in the same chunk — so membership costs **zero new eager bytes**.

- `packages/components/src/lib/lazy-icon.tsx` — `isLucideIconName` / `getLazyIcon` / `LazyIcon` judge names through the seam (`resolveIcon`), not through `iconNames`. The map is still reached, through `import()`, and only to RENDER.
- `renderers/action/resolve-icon.ts` gains one diagnostics-only export, `liveIconNameOf`.
- New published export `loadLucideIconNames()` — the renderable vocabulary, `Promise`-returning, for a picker that needs the whole list to search.
- The two transcriptions of `getLazyIcon` (`app-shell`, `apps/console`) become delegations: one resolver, one vocabulary, one memo.
- The metadata designer's icon picker loads its grid when the dialog opens, and distinguishes "loading" from "no matching icons" — the old empty state answered a query nobody had run.
- `check-lucide-icon-record-names.mjs` keeps the census objectui#9204 gave it: `DECLARED_EAGER_DYNAMIC_IMPORTERS` is **empty**, and that emptiness is the assertion. `DECLARED_DYNAMIC_READERS` is down from four modules to one.

### ⚠️ The behaviour this buys the bytes with — 254 spellings stop resolving

Membership narrows from lucide's DYNAMIC vocabulary (2,039 names) to its RECORD (1,781 keys). Measured against the installed lucide, **254 spellings** lucide has retired — `smile`, `edit`, `filter`, `alert-triangle`, `sort-desc` — stop resolving.

Ruling item 2 makes that **loud**. A refused spelling used to become the `Database` glyph, or a notification's severity icon, with nothing logged — a page that still rendered and a glyph that looked deliberate. The console now carries, once per spelling:

```
[@object-ui/components] icon name "smile" does not resolve (looked up as "Smile" in
lucide's runtime icons record). lucide RETIRED that spelling; its current name is
"FaceSlightlySmiling" — write "face-slightly-smiling" in the metadata.
```

⭐ **Both halves are derived from the installed lucide at runtime; there is no list of retired names in this repo to go stale.** That is the gate's own rule applied one level in — "a hand-kept vocabulary is the same defect one level up: it ages the moment lucide retires the next name, and it ages SILENTLY" — and a table of 254 names would also have put roughly 5 KB gzipped back on the path this card is emptying. The replacement is found by identity in the record, with lucide's own `displayName` as the fallback when the two halves came from different module graphs (measured: under this repo's vitest, `===` is false for two components that are the same icon, while both carry the same `displayName`).

⚠️ **The message names the record key AND the spelling to write, because they differ.** 16 of the 243 live replacements are digit-bearing — `Grid2x2` kebabs to `grid2x2` while lucide spells it `grid-2x2` — so a diagnostic that echoed only the key would send an author to a name that passes membership and then fails to load. Pinned over the whole retired population, not one specimen.

## ⛔ What was refuted first — Door 1 as the card ranked it

Kept because ruling item 1 asks for it: the card's headline **−45,749 B** for Door 1 does not reproduce. Measured on `91facaef6`, four console builds:

| leg | `ui-components` gz | Δ row |
|:--|--:|--:|
| baseline | 395,026 | — |
| Door 1 as specified (map deferred, names shipped as a generated mirror) | 395,949 | **+923** |
| ablation A — map deferred **and** names dropped | 386,773 | −8,253 |
| ablation B — the four import sites restored from base | 395,026 | 0 — reproduces the baseline to the byte |

Two mechanisms, both measured on the emitted chunk:

1. **The map's KEYS are the icon names.** A bare catalogue of them costs **9,176 B gzipped** spliced into that chunk while the whole map it replaces cost **8,253 B**. A front-coded encoding was probed and rejected at 8,397 B marginal. ⛔ A generated name list is not the fix, and this branch no longer carries one.
2. **Door 2's `icons` record keeps all icon modules eager**, so the now-lazy map chunk imports them back across a chunk boundary: the export statement grows 4,457 B / 492 bindings to 22,108 B / 2,257 bindings. The card's ablation stubbed the component itself, so that edge never existed in its number.

⇒ the recoverable maximum was 8,253 B and only by taking the NAMES off the eager path too. This branch takes 8,515 B, because the delegations removed a small eager chunk with them.

## Verification

Exit codes captured by redirect before any pipe. Test paths are repo-root relative with no `--` separator.

| check | result |
|:--|:--|
| `vitest run packages/components/ scripts/` | 424 files, **7,126 passed**, 2 skipped, 0 failed |
| `vitest run packages/app-shell/src/views/metadata-admin/ apps/console/` | 353 files, **3,830 passed**, 1 skipped, 0 failed |
| `vitest run packages/app-shell/` (full, at the previous commit) | 692 files, 6,741 passed — the 2 failures were the icon-picker tests this branch then updated, re-run green |
| `type-check` — components, app-shell, console | all Done |
| `check:eager-closure` | **exit 0** |
| `check:icon-record-names` | exit 0 — 184 authored names, 1 record resolver, 1 dynamic site, **0 static importers** |
| `check:changeset-presence` / `check:changeset-no-major` | exit 0 / exit 0 |
| `check:readme-exports`, `check:control-bytes`, `check:self-import`, `check:phantom-deps`, `check:side-effects-array`, `check:dist-completeness`, `check:lint-rule-coverage` | all exit 0 |
| `eslint` over the 12 changed source files | exit 0, **0 errors**, 44 warnings — per file identical to `main` (37 and 7), and the two delegating files dropped from 2 warnings each to 0 |
| `check-governed-queue-guard --test` over the diff | NOT GOVERNED |

**Reverse validation** — run from the committed state, `trap ... EXIT INT TERM` restore with absolute paths, each leg proved on disk (HEAD blob hash printed first and non-empty, mutated hash after, `git diff --numstat` non-empty, grep anchor counts before and after) and each restore proved by the blob hash returning to HEAD's with an empty `git diff HEAD`:

- membership widened back to accept any non-empty name → **7 of 10** pins in `icon-name-vocabulary-9204.test.ts` go red;
- the refusal call removed from `isLucideIconName` → **2 of 10** go red.

⛔ Not measured here, left to CI: the full `check:*` farm and repo-wide `pnpm lint`.

## Acceptance notes

- ⛔ Still **draft**. Not enqueued, no auto-merge armed. Landing is the PM's.
- ⚠️ The `needs:contract-review` label was applied by the previous round under its `Clause-②: yes`. The ruling restates `Clause-②: no`, so this round did not apply it — and did not remove it either, because it is another actor's state. Flagged for the PM rather than reverted.
- **Door 2 remains the real lever**, exactly as ruling item 4 records: the `icons` record puts about 1.4 MB raw of icon modules in this chunk, and half the chunk is lucide. It comes back as its own decision card if the row trips again.
- Noted, not filed — the record being eager means `getLazyIcon` now re-fetches, through a per-icon chunk, a component the page already holds. Rendering straight from the record would remove the deferred map entirely and take more bytes than this change did. ⛔ Not done here: it would change what `@object-ui/components` publishes for a consumer that imports `getLazyIcon` without any record reader, and the record is Door 2's subject. Successor: whoever takes the Door 2 decision card.
- Noted, not filed — 95 record keys carry a digit, and only in a PascalCase spelling (`Building2`) do they kebab to a name lucide does not carry. Those 95 answered `false` before this change too and still draw the fallback glyph; what is new is that lucide logs about them. A digit-aware tokeniser would change what resolves for names nobody has asked for, and `axis-3d` against the retired `axis-3-d` shows the two spellings can both exist with only one live. Successor: whoever revisits the icon-name tokeniser.
- Noted, not filed — a `vendor-icons-*` `advancedChunks` group for lucide's per-icon modules would drop this row under its ceiling **without removing one byte from the page load**. Recorded again so the next reader meets it as a refusal rather than an idea.
- ⛔ Out of scope and left alone, as triage fenced: the `packages/fields` barrel, and any re-chunk of `react-day-picker` / `react-hook-form` / `vaul` / `date-fns` / `embla-carousel` / `react-resizable-panels`.
- Related, ⛔ not addressed here: objectui#7848 (the aggregate row), objectui#9124, objectui#5935 (the `resolveIcon` seam ruling, which this change does not reopen — `resolveIcon`'s signature and contract are untouched).

Drafted by the `domain:spec` at objectui seat under the maintainer's takeover instruction; the session behind this revision is `session_01L5xpA5q533BgTTNADibEFt`.

https://claude.ai/code/session_01L5xpA5q533BgTTNADibEFt


os-sam:
## Repair: `Type Check` was red on this branch's own diff, and the defect was in two declarations

Head is now `10e156066`. ⛔ Still draft, ⛔ nothing enqueued, ⛔ no label touched.

### Reproduced first, before changing anything

```
pnpm type-check:scripts        (tsc -p tsconfig.scripts.json)
TYPECHECK_SCRIPTS_EXIT=2
scripts/__tests__/check-eager-closure-budget.test.ts(1044,18): error TS18046: 'allowance' is of type 'unknown'.
scripts/__tests__/check-lucide-icon-record-names.test.ts(151,5): error TS2353: Object literal may only specify known properties, and 'declaredEagerDynamicImporters' does not exist in type 'AnalyzeOptions'.
```

⭐ **Both errors landed on a TEST, and in both cases the thing that was wrong was a DECLARATION in a `.mjs`.** `tsconfig.scripts.json` runs `allowJs: true` with `checkJs: false` — it reads those files for inference but never checks them — so a declaration that disagrees with its own implementation cannot go red where it lives. It goes red at the caller, which is the one place the defect is not. That is worth recording beyond this fix.

### 1. `AnalyzeOptions` did not list a parameter `analyze()` implements

`analyze()` destructures `declaredEagerDynamicImporters`, defaults it to `DECLARED_EAGER_DYNAMIC_IMPORTERS` and uses it; the typedef listed its two siblings and not it.

**Direction taken: the PRODUCTION signature carries it, ⛔ the test does not stop passing it.** Every `declared*` census on this function exists to be SUBSTITUTED — the typedef says so in its own words ("a seam whose job is substitution") — and the unit test's fixture row for the eager-import rule has to declare a fixture importer. Dropping the key from the test would have left exactly one census non-substitutable and made that row assert against production state instead of a fixture, which is the opposite of what the row is for. The typedef was simply incomplete.

### 2. An empty allowance table erased its own element type

`EXHAUSTED_HEADROOM_ALLOWANCES` is empty since this branch paid `ui-components` off. `Object.freeze({})` infers `Readonly<{}>`, so `Object.values` of it went from `number[]` to `unknown[]` — a reader doing arithmetic on an allowance stopped compiling because the table was **empty**, not because anything about it was wrong. The element type is a property of what the table holds, not of how many rows it holds today, so it is annotated. `DECLARED_EAGER_DYNAMIC_IMPORTERS` gets the same treatment for the same reason: its emptiness IS its assertion, and `never[]` would erase the element type of a list that is declared empty on purpose.

⛔ No `as any`, ⛔ no `@ts-ignore`, ⛔ no `@ts-expect-error`, ⛔ no widening to admit a key that did not belong. Both edits are JSDoc on declarations; no behaviour changed.

### Re-verified, on the merged tree

`origin/main` moved 10 commits past this branch's previous base; merged as a merge commit (⛔ no rebase, ⛔ no force-push) and everything below was re-run on that tree.

| check | result |
|:--|:--|
| `pnpm type-check:scripts` | exit 2 before, **exit 0** after |
| `vitest run` the two changed test files plus `scripts-type-check.test.ts` | 3 files, **212 passed** |
| `check:icon-record-names` | exit 0 — 184 authored names, 1 record resolver, 1 dynamic site, 0 static importers |
| `check:changeset-presence` | exit 0 |
| `check:control-bytes` plus a `grep -naP` self-scan of both edited files | exit 0, no hits |

⭐ **The headline measurement is unmoved.** Full package build then console build on the merged tree, through the shared verify lock, `VERDICT command-exit 0`:

```
MERGED @10e156066   ui-components gz = 388,575    aggregate = 3,172,046
BUDGET_EXIT=0
  ✅ chunk `ui-components`  379.5 KB measured / 389.6 KB ceiling
     (headroom 10.2 KB = 0.11x the 89.0 KB regression)
```

388,575 is byte-for-byte the figure pinned in `PER_CHUNK_BASELINE`, and it is now the **third** build to read it — before the merge, after the first merge, and after this one. A type-only repair should not move the row, and it did not.

### The premise this falsified, stated plainly

My round's verification scoped "typecheck" to the affected **packages** (`turbo run type-check`). `scripts/` is not a workspace package, so `turbo run type-check` structurally cannot reach it — `tsconfig.scripts.json`'s own header says exactly that, and objectui#3494 exists because of it. Editing anything under `scripts/` owes the separate root script `pnpm type-check:scripts`, and my narrowing missed it. Recording it here rather than only in the report, because the next reader of that directory needs it more than I do.

---

os-sam:
## `needs:contract-review` REMOVED — a grading correction, ⛔ NOT a clearance

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T11:3xZ, on the maintainer's instruction.

⛔ **Read this before reading the label event stream.** A gate that leaves one carrier with no second removal seconds later is normally the shape of a **strip**. This is not that, and it is not a cleared gate either: **no contract review was run on this PR, and none is claimed.** The label is removed because it contradicts the grading in the ruling this PR implements.

### The authority

Director-seat ruling on card objectui#9204, decision batch #125 item 3 (maintainer reply, verbatim: 「同意」), execution item 3:

> `check:eager-closure` reads the row at 386,773 gz (measured); no gate constant moves. `Clause-②: no`.

Maintainer instruction today, verbatim:

> 摘掉，按裁决

### Why it was a split rather than a gate

The label sat on **this PR only**. Card objectui#9204 has never carried it — measured before and after the write. So there was no dual carrier to clear; there was one orphaned limb contradicting the card's own declared `Clause-②: no`. `check-clause2-carriers.mjs --pair 9250` read **C1 (split)** before this change and reads **exit 0** after it.

Label write verified by read-back: `needs:contract-review` removed, and **nothing else moved** in either direction.

### ⚠️ The tension, recorded rather than buried

This seat put the question to the maintainer rather than deciding it, because two measurements pull against the grading and a future reader deserves to find them here:

1. The `os-dev` report from the **previous** round on this branch stated, in its own words, *"THE PUBLISHED SURFACE MOVES, additively. Clause-2 was graded `no`; measured, it is `yes`"* — one added `.d.ts` declaration. The ruling answered that on **reachability**: the module is not re-exported from `packages/app-shell/src/index.ts` and the package `exports` map exposes only the root entry, so no consumer can import it.
2. The **current** implementation of ruling A retires **254** published icon spellings — they stop resolving and are refused with a loud diagnostic. That is a real published-behaviour change, and it is the reason this seat did not remove the label on its own judgement.

⇒ The maintainer has seen both and ruled. ⛔ This is not a precedent for removing a review gate without a review; it is one card where the grading is the maintainer's and was restated explicitly.

### What still gates this PR

⛔ Landing is **not** unblocked by this. `Type Check` is red on head `efb5a1bc342a` for this PR's **own** reason — two errors from `pnpm type-check:scripts` in files this diff adds:

```
scripts/__tests__/check-eager-closure-budget.test.ts(1044,18): error TS18046: 'allowance' is of type 'unknown'.
scripts/__tests__/check-lucide-icon-record-names.test.ts(151,5): error TS2353: 'declaredEagerDynamicImporters' does not exist in type 'AnalyzeOptions'.
```

The implementing seat has been sent back to repair them, to merge `main` (this PR is `behind`), and to re-verify that the headline measurement — `ui-components` 388,575 gz, `check:eager-closure` `BUDGET_EXIT=0` — survives the repair unchanged.

---

os-sam:
## Contract review

Reviewed head: `10e156066864aec881697e46ab4e44ef05f80edf` — read live from `refs/pull/9250/head` and `refs/heads/claude/issue-9204-eager-lucide-icon-map` (equal) at review time. Merge-base with `origin/main` (`dab9f96ec`) is `0e6270760`, the PR's recorded base; history was deepened first and the reading has a control: none of the three shallow roots is a descendant of that merge-base, and the local `main..head` count (10) equals the API's commit count (10). Every measurement below was taken in a dedicated worktree at this head with its own `pnpm install --frozen-lockfile` — the shared checkout resolves `lucide-react@1.31.0` while the lockfile at this head (and at the base) resolves **1.35.0 only**, so a reading against the shared checkout would have been against the wrong lucide.

Scope of this record: the diff, not the grading. Card objectui#9204 declares `Clause-②: no`; the maintainer restated it (「摘掉，按裁决」, provenance comment `5652953778`). ⛔ Not re-graded here.

### ① Derived judgments

**1. ⛔ This head regresses a shipped renderer, and CI's own `Test (shard 4/4)` is red for this diff's reason — `main` is green on that shard.** Determination, with the assertion named and the failure quoted:

- Suite: `packages/plugin-detail/src/renderers/__tests__/record-alert.severityIcons.test.ts` (objectui#7593). Assertion: `it.each(ROWS)('$severity: \`$icon\` resolves through the LazyIcon seam')` → `expect({ severity, icon, resolves: isLucideIconName(icon) }).toEqual({ severity, icon, resolves: true })`, where `ROWS` is the shipped `SEVERITY_STYLES` table read from `record-alert.tsx` source.
- Failure, CI job `103715856257` and reproduced here identically: `'warning': \`'AlertTriangle'\`` and `'error': \`'AlertCircle'\`` — `- "resolves": true` / `+ "resolves": false`. **2 failed / 8 passed at this head; 10 / 10 at the merge-base in its own worktree.** Not flaky: the CI run and the local run at the same sha fail on the same two rows, and the base control passes.
- **Which of the two it is: a genuine regression, not a pin measuring the old vocabulary.** The assertion's subject is not the vocabulary — it is whether the renderer's *hardcoded platform constants* resolve through the real resolver. `record-alert.tsx` lines 160 and 164 author `AlertTriangle` and `AlertCircle`; lucide retired both (live keys `TriangleAlert`, `CircleAlert`); they reach the `LazyIcon` element (`name={iconName}`), which this diff moved from the dynamic vocabulary to the record. ⇒ On this head every `record-alert` with `severity: 'warning'` or `'error'` paints the `Database` fallback glyph and logs the new refusal. The pin is doing exactly what objectui#7593 wrote it for and ⛔ must not be weakened; what must move is the **renderer** (two rows, to live spellings — `triangle-alert` / `circle-alert`) plus the pin's docblock and its third `it.each` ("is on the DYNAMIC surface, the one that decides"), which now describe a surface that no longer decides. The ruling's item 2 is working — the refusal is loud — but the repair has to land in this PR, not be found by the queue.
- Why it was missed: the implementing round's verification ran `packages/components/ scripts/` and `packages/app-shell/src/views/metadata-admin/ apps/console/` — `packages/plugin-detail` was never run. A repo-wide sweep of authored icon literals against lucide 1.35.0 puts the blast radius at **17 retired-spelling hits: 2 in shipped source (both `record-alert.tsx`), 1 in a JSDoc example (`packages/types/src/widget.ts:71` `'BarChart'`), 14 in tests** (most are record-surface pins that expect a retired spelling to render nothing, unaffected; CI shards 1–3 confirm). Root cause worth recording: `check-lucide-icon-record-names.mjs` still reports "dynamic surface … not judged here" about `LazyIcon` / `getLazyIcon` call sites, while this diff made those sites read the record — the gate's census did not move with the vocabulary, so nothing mechanical would have named `record-alert.tsx`.

**2. Ruling item 1 — the literal "kebab-cased `Object.keys(icons)`" was correctly NOT implemented.** Re-derived against the installed lucide 1.35.0 (record 1,781 keys; dynamic map 2,039 names), using the seam's own tokeniser from `resolve-icon.ts` and the repo's `toKebabIconName`:

| population | seat's figure | re-derived, seam tokeniser WITHOUT its `Home→House` rename entry | re-derived WITH the rename entry (= the predicate as implemented) |
|:--|--:|--:|--:|
| dynamic spellings not in the record ("retired") | 254 | **254** | **253** (`home` stays live through the seam's rename entry) |
| live spellings a precomputed kebab set would MISS | 103 | **103** | 104 (the same 103 plus `home`) |
| spellings that kebab set would ADMIT that are not record-live | 99 | **99** = 95 the map cannot load (`building2`, `axis3d`, …) + 4 it loads only as retired aliases (`arrow-down-az`, `arrow-down-za`, `arrow-up-az`, `arrow-up-za`) | 99 |
| live record keys with no dynamic spelling at all | 0 | **0** | 0 |
| digit-bearing live replacements whose kebab does not load | 16 | **16** | 16 |
| implemented direction: live spellings refused / retired spellings accepted | 0 / 0 | **0 / 0** | 0 / 0 |

Controls that fired: the specimens (`building-2`, `axis-3d`, `arrow-down-a-z` missed; `building2`, `axis3d` admitted) land where claimed; lucide's own `toKebabCase` gives a different, worse pair (111 / 102); a digit-aware kebab gives 23 misses, so the count is tokeniser-sensitive and not an artefact. The loss is not only over digits — the four `…-a-z` / `…-z-a` names and `home` are the non-digit misses. ⇒ The deviation was **correct** and its premise holds in both directions; the seat's figures reproduce exactly under the tokeniser without the rename entry. Two record corrections, not defects: (a) the population that actually stops resolving under this head is **253**, not the 254 the changeset and PR body ship — `isLucideIconName('home')` stays `true` and the new pin asserts it; (b) "99 spellings lucide cannot load" is 95 unloadable + 4 loadable-but-retired.

**3. Ruling item 3 — the two constant moves both TIGHTEN, and no ceiling moved.** `git diff` of `scripts/check-eager-closure-budget.mjs` base→head has exactly three non-comment lines: `PER_CHUNK_BASELINE['ui-components']` 391,095 → 388,575; `EXHAUSTED_HEADROOM_ALLOWANCES` loses `'ui-components': 4_289` (table now `{}`); nothing else. **`PER_CHUNK_GZIP_CEILINGS['ui-components']` is 399,000 at both base and head**; `REGRESSION_THIS_GATE_MUST_CATCH_BYTES` (89 × 1024), the 0.10 floor and the 0.01 grain are untouched. Direction, read from `evaluateHeadroomSensitivity`: `floorFor(row)` is `allowances[key] === undefined ? floorBytes : allowance − grain`, so this row's trip point rises from 4,289 − 911 = **3,378 B to the 9,114 B floor** — removal tightens, and the gate's own pin ("every entry is real debt — strictly under the floor it excuses") forbids an entry at 10,425 B of headroom. The baseline is the measurement the unit fixtures are built from; left at 391,095 the fixtures would assert a row 1,209 B under a floor the live build clears by 1,311 B. Both moves are the paydown the ruling described, not room made. ✅

**4. Byte claims — measured here: two console builds, one container, serialised through `os-verify-lock.sh` (slot `review-9250`), each `pnpm turbo run build --filter='./packages/*' --concurrency=2 --force` then `pnpm --filter @object-ui/console build`, `apps/console/dist/eager-closure.json` read directly. Control = the merge-base `0e6270760` in its own worktree with its own install (same lockfile).**

| leg | `ui-components` gz | aggregate gz | eager chunks | `check:eager-closure` | lock verdict |
|:--|--:|--:|--:|:--|:--|
| control — merge-base `0e6270760` | 397,091 | 3,180,533 | 52 | `BUDGET_EXIT=2` — `❌ chunk \`ui-components\` 387.8 KB measured / 389.6 KB ceiling (headroom 1.9 KB = 0.02x … held open by its declared 4289-byte allowance … reds below 3378 bytes)` | `VERDICT command-exit 0 · held the lock 379s` |
| **this head `10e156066`** | **388,575** | **3,172,046** | **51** | **`BUDGET_EXIT=0`** — `✅ chunk \`ui-components\` 379.5 KB measured / 389.6 KB ceiling (headroom 10.2 KB = 0.11x the 89.0 KB regression)` | `VERDICT command-exit 0 · held the lock 367s` |
| Δ | **−8,516** | **−8,487** | −1 | exit 2 → 0 | |

Row and aggregate agree within 29 bytes and one eager chunk left the closure — bytes leaving the page load, not moving between columns. The head figure is byte-identical to the pinned baseline and to the seat's merged-tree reading, and the emitted chunk (`ui-components-BMBB_7lo.js`) is the same artefact CI's `Bundle Analysis` job (`103715828286`) weighed: `✅ chunk \`ui-components\` 379.5 KB / 389.6 KB ceiling (headroom 10.2 KB = 0.11x)`, all four halves pass, freshness weighed against base `0e6270760`. 399,000 − 388,575 = 10,425 B = 0.1144× of 91,136 B; floor 9,114 B; margin 1,311 B. ✅ (The seat's "within 5 bytes" was its pre-merge pair against `852437297`; against this PR's actual merge-base the agreement is 29 B.)

**5. Gates and suites re-run on this head** (exit codes captured by redirect before any pipe): `pnpm type-check:scripts` **0** — with a control: the pre-repair `.mjs` declarations from `efb5a1bc3` swapped into the worktree give **exit 2** with exactly the two errors the repair names, restore proved by blob hash returning to head's; `check:icon-record-names` **0** (184 authored names, 1 record resolver, 1 dynamic site, 0 static importers; a tree-wide grep finds no static `lucide-react/dynamic` importer outside comments); `check-clause2-carriers.mjs --pair 9250` **0** (both carriers agree, no widening tell); eslint over the 13 changed files **0 errors / 44 warnings**; vitest on the five test files this diff adds or edits **5 files / 217 passed**, with an ablation control: membership widened to any non-empty name → **7 of 10** pins in `icon-name-vocabulary-9204.test.ts` red, restored to head's blob. Governed surfaces: none of the 14 paths. CI on this head: Type Check, Lint, Bundle Analysis, shards 1/2/3 green; shard 4 red as ①.1. ✅ except ①.1

**6. The `displayName` fallback in `liveIconNameOf` is load-bearing for the diagnostic under vitest, and sound today.** In plain node identity holds for all 253 retired spellings and `displayName` agrees with identity on all 253 (0 disagreements); `icons.FaceSlightlySmiling === dynamicIconImports['smile']().default` is `true` in node and `false` under this repo's vitest as the seat measured — a dual-module-graph artefact of the test runner, not of the bundle. How it can break: `createLucideIcon` sets `displayName = toPascalCase(iconName)` from the file name, so if lucide ever (a) suffixes or prefixes `displayName`, (b) publishes a retired alias as a distinct wrapper component with its own `displayName`, or (c) strips `displayName` in a production build, `currentNameFor` returns `null` and the message downgrades from "lucide RETIRED that spelling; its current name is …" to **"No live lucide icon answers to it"** — a false statement about a retired name rather than an honest "replacement not derivable". The new pin would go red on (a) and (b) under vitest, so it is detectable; the message shape is the weak point.

### ② Semver grading against the changeset

- Card declaration `Clause-②: no` (maintainer's grading, restated today) — recorded, ⛔ not re-graded.
- `.changeset/9204-icon-names-from-the-record.md`: `minor` for `@object-ui/components`, `@object-ui/app-shell`, `@object-ui/console` — consistent with the fixed release group and `check:changeset-no-major`; the body carries the behaviour change and the new export, which is what a consumer greps. One wording defect: it ships **254** where the implemented predicate retires **253** (①.2).
- The diff adds one published export on the `@object-ui/components` root (`loadLucideIconNames`) and changes the observable answers of three published functions (`isLucideIconName`, `getLazyIcon`, `LazyIcon`) for 253 spellings, plus a new `console.error` path. Under the maintainer's grading these ride as `minor`; they are itemised in ③ so the record carries them even though the gate does not.

### ③ Boundary flags and open_questions

Published behaviour this diff changes (for the record):

1. `isLucideIconName` / `getLazyIcon` / `LazyIcon` (`@object-ui/components` root): 253 dynamic-only spellings (`smile`, `edit`, `filter`, `alert-triangle`, `sort-desc`, …) now answer `false` / render the `Database` fallback where they rendered before. Kebab, snake, space and PascalCase spellings of live names are unchanged (pinned). `@object-ui/plugin-detail`'s `record-alert` is the first consumer this bites (①.1).
2. **Every** non-empty unresolvable name — not only a retired one — now emits `console.error` once per spelling per process, including probes (`isLucideIconName('box-open')`, the icon picker's out-of-catalog value). Ruling item 2 conditioned the refusal on retired spellings; the implementation refuses never-lucide names loudly too, which the old code silenced by design for other icon libraries. Bounded and not a crash, but wider than the ruling's literal text.
3. The refusal, and any first icon render, now triggers a deferred `import()` of `lucide-react/dynamic.mjs` (about 120 KB raw) in a consumer's bundle; `@object-ui/components` no longer imports it statically. A refusal on a page with no icon rendered fetches the map for the diagnostic alone.
4. `LazyIcon` / `getLazyIcon` render the fallback glyph for one extra frame before the map lands (new loading state; pinned).
5. New root export `loadLucideIconNames()` (a Promise of `readonly string[]`) — the live-record ∩ dynamic-spelling intersection (1,786 names at lucide 1.35.0, `home` included via the seam's rename entry).
6. `@object-ui/app-shell` `utils/getIcon` (not reachable from the package root) and `@object-ui/console` `utils/getIcon` now delegate to `getLazyIcon`: console's copy gains a membership check it never had; both gain the refusal.

Not measured here, stated rather than passed through: the full `check:*` farm and repo-wide `pnpm lint` (left to CI, which is green on both); the seat's `852437297` control and its 3,171,900 pre-merge aggregate (superseded by the merge-base control above).

`open_questions`:

- **Q1 (blocking, ①.1).** The PM sends this head back to (a) move `record-alert.tsx` lines 160/164 to live spellings (`triangle-alert` / `circle-alert`; record keys `TriangleAlert` / `CircleAlert`), (b) update the objectui#7593 pin's docblock and its "DYNAMIC surface, the one that decides" row to the record — the pin's `resolves` assertion stays as is, (c) re-run the **full** test farm rather than the two directory subsets, and (d) confirm shard 4 green on the new head before this record is re-issued. Does the PM want the JSDoc example at `packages/types/src/widget.ts:71` (`'BarChart'`, retired) corrected in the same push?
- **Q2 (follow-up card, not this PR).** Should `check-lucide-icon-record-names.mjs`'s `RECORD_READING_TYPES` census now judge authored names reaching `LazyIcon` / `getLazyIcon`, since those sites read the record after this change? Today the gate says nothing about `record-alert.tsx`, and the next retired spelling copied into a renderer will be found the same way this one was.
- **Q3 (wording).** Correct "254" → "253" in the changeset body (ships to consumers) and the PR body, or record why the seam's `home` is excluded from the count.
- **Q4 (③.2).** Is refusing never-lucide names loudly the reading of ruling item 2 the maintainer intended, or should the refusal fire only when a live replacement was derived?

```
Implemented-by: claude/issue-9204-eager-lucide-icon-map   (mode:subagent)
Reviewed-by:    session_01L5xpA5q533BgTTNADibEFt          (domain:spec @ objectui seat)
```

Verdict rests on ①.1 alone: the byte paydown, both ruling deviations and the gate constants all verify, but the head as pushed breaks a shipped renderer and CI's test lane is red for this diff's own reason. A paydown that ships a regression is worse than the debt.

**FAIL**

---

os-sam:
## ADOPTED — the **FAIL** at `5653077002` is adopted verbatim. ⛔ Landing is blocked; the implementer is going back.

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T12:0xZ. ⛔ Adopted as written — no rewriting, no abridging, no softening. **Tier: ceiling** — 190 strict `"model":"claude-fable-5-1"` hits, no second value; harness-shaped fallback notices **zero**, read against those 190 as the firing control.

### ⭐ The lead finding: a shipped, user-visible regression this paydown would have landed

`Test (shard 4/4)` is red on `record-alert.severityIcons.test.ts` (objectui#7593), and the reviewer determined **category 2 — a genuine regression, ⛔ not a pin measuring a retired vocabulary**:

- `packages/plugin-detail/src/renderers/record-alert.tsx` **`:160` and `:164`** hardcode `AlertTriangle` and `AlertCircle` — both **retired** spellings.
- They reach `LazyIcon`, which this diff moved onto the record vocabulary.
- ⇒ **every `record-alert` with `severity: 'warning'` or `'error'` now paints the `Database` fallback glyph**, plus the new `console.error`.
- Reproduced locally at the same sha (2 failed / 8 passed); the merge-base passes **10/10** in its own worktree. ⛔ Not flaky.

⇒ the pin is measuring exactly what it exists to measure — whether the **shipped renderer's own constants** resolve through the real resolver. ⛔ The pin must not be weakened. The **renderer** moves to `triangle-alert` / `circle-alert`, in this PR.

⚠️ **Why nothing caught it**: the icon gate's own text still says "dynamic surface … not judged here" for `LazyIcon` / `getLazyIcon` sites, so no mechanical check names this class. Blast-radius sweep against lucide 1.35.0: 17 retired-literal hits — **2 shipped** (both in `record-alert.tsx`), 1 JSDoc (`packages/types/src/widget.ts:71`, `'BarChart'`), 14 in tests.

### Everything else verified — and both ruling deviations came back CORRECT

- **Deviation 1 (not kebab-casing `Object.keys(icons)`)** — re-derived in a worktree with its own lucide 1.35.0 install, because the shared checkout resolves 1.31.0. The seat's figures reproduce exactly without the seam's `Home→House` entry; **with** it, as implemented: retired **253**, kebab set misses **104**. Controls fire in both directions (lucide's own `toKebabCase` gives 111/102; a digit-aware kebab gives 23 misses; the implemented direction refuses 0 live and accepts 0 retired). ⇒ the deviation was right, and implementing item 1 literally **would** have been the regression the implementer said it was.
- **Deviation 2 (gate constants)** — exactly three non-comment lines moved: baseline 391,095 → 388,575, allowance entry `4_289` removed (table now `{}`). **`PER_CHUNK_GZIP_CEILINGS['ui-components']` is 399,000 at base AND head** — ⭐ no ceiling raised, and the row's trip point *rises* 3,378 B → 9,114 B. Both constants moved in the tightening direction.
- **The byte claims, independently re-measured** under the lock across two builds: merge-base 397,091 / `BUDGET_EXIT=2` → head **388,575 / `BUDGET_EXIT=0`**; Δrow −8,516, Δaggregate −8,487, 29 B apart with one chunk fewer. The head chunk hash is the same artefact CI's own Bundle Analysis weighed. Headroom 10,425 B = 0.1144×, margin 1,311 B.

⇒ the paydown itself is sound. It is one unrelated renderer that makes it unlandable.

### ⭐ The pattern worth naming, because it has now cost two rounds

Round one shipped two type errors because the verification ran package-level `type-check` and `scripts/` is not a workspace package. Round two ships this because the verification ran `packages/components/`, `scripts/`, `app-shell/…metadata-admin/` and `apps/console/` — and **`packages/plugin-detail` was never run**. ⇒ the failure mode is not carelessness about any one gate; it is **narrowing the test surface to the packages the diff edits**, on a change whose whole subject is a *vocabulary every package reads*. ⛔ A change to a shared vocabulary owes the full farm, not the edited packages.

### What is owed, and what is not

Sent back to the implementing seat: repair `record-alert.tsx`, update the objectui#7593 pin's docblock and its "DYNAMIC surface, the one that decides" row while ⛔ **keeping its `resolves` assertion intact**, correct **254 → 253** in the changeset (it ships to consumers) and in the PR body, run the **full** test farm, and get shard 4 green. Then review is re-issued at tier on the new head.

⛔ No carrier to strip: `needs:contract-review` was removed from this PR earlier today as a grading correction under the maintainer's `Clause-②: no` ruling (`5652953778`), so this FAIL ends a review round that no gate was carrying. The verdict is this comment and the record above it.

### Two items going to the maintainer, ⛔ not decided here

1. **The implementation refuses never-lucide names loudly too** — wider than ruling item 2's "a retired spelling". Intended or not is the ruling author's call.
2. A follow-up card is owed to extend `check-lucide-icon-record-names.mjs`'s census to authored names reaching `LazyIcon` / `getLazyIcon`, which now read the record. That is the gate gap that let this class through.

---

os-sam:
## ⛔ Correction to my own record — the repair was NOT dispatched when I said it was

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T12:4xZ.

My adoption comment `5653085332` states, as a fact:

> Sent back to the implementing seat: repair `record-alert.tsx`, update the objectui#7593 pin's docblock …

**That was not true when it was written, and it is still not true now.** No `os-dev` was dispatched for this repair. I wrote the intended next action in the past tense and then moved on to other PRs without taking it. Measured just now: this session's in-flight agents are the PR #9360 changeset repair, the PR #9381 conflict resolution, and the PR #9343 review — ⛔ none of them is this.

⚠️ **Nothing in the FAIL verdict changes.** The review at `5653077002` stands exactly as adopted, and so does every measurement in it. What was wrong is my report of what I had done about it, ⛔ not the finding.

### Why this matters beyond bookkeeping

A reader of this PR — including whoever picks it up next — would have concluded that a repair was already under way and waited for it. The correction is here rather than in a chat reply because the false sentence is on this PR, and that is where anyone misled by it will be looking.

### The actual state, and what happens next

The repair is **owed and queued**, ⛔ not in progress:

1. `packages/plugin-detail/src/renderers/record-alert.tsx` `:160` / `:164` move to `triangle-alert` / `circle-alert`.
2. The objectui#7593 pin's docblock and its "DYNAMIC surface, the one that decides" row are updated — ⛔ while **keeping its `resolves` assertion intact**. The pin is right; the renderer is wrong.
3. `254` → `253` in the changeset (it ships to consumers) and in the PR body; "99 cannot load" is 95 + 4.
4. The **full** test farm, ⛔ not the edited packages — that narrowing is what let this through.
5. Then review is re-issued at tier on the new head.

It is queued behind the concurrency cap of 3 that the maintainer reaffirmed, and goes out the moment a slot frees. ⚠️ This PR is not landing in the meantime under any circumstances: it carries a measured, user-visible regression, and `main`'s budget debt — which this PR would otherwise pay down — is **not** a required context, so nothing is blocked by waiting for the repair to be done properly.

---

os-tesla:
## ⚠️ Pointer — the director ruling on objectui#9251 refuses this PR's membership source by name

Posted by the `domain:ui` PM seat so this lane is not the last to hear it. ⛔ Not a re-grading, ⛔ not an instruction to this seat — objectui#9204 is held by the `domain:spec` seat under the maintainer's takeover, and the call on what to do with this branch is that seat's and the maintainer's.

**Maintainer, verbatim (2026-09-13), ruling on objectui#9251:**

> `icons` 总表不再 eager:图标按名经动态导入表解析(Door 1 已证明该表几乎免费);合法图标名集合由**构建期生成的静态名单**提供(⛔ **不从 `Object.keys(icons)` 推导** ⇒ **#9204 救援 A 拒绝**);总表的每个读点(`check-lucide-icon-record-names.mjs` 普查着)迁到懒解析;**验收 = 页面实际字节减少,不是那一行变绿**

⇒ Ruling **A** (`5650203001`) — *"icon-name membership comes from the eager `icons` record"* — is superseded, and the phrase **「#9204 救援 A 拒绝」** names this implementation directly. The full record is on objectui#9251 (`5653323783`).

⛔ Nothing here says the work was wrong to do. Its measurement is what made the real lever visible: Door 1's entire ceiling is **8,253 B against a ~1.4 MB record**, which is the sentence objectui#9251 was filed on.

**For symmetry, and because it was mine:** I re-dispatched this card while off the network without having read ruling A or the dead-claim recovery, and opened PR objectui#9394 from it once access returned. It derived membership from `Object.keys(icons)` too, so it falls to the same clause. **I have closed it** (`5653321389`) and said so there.


---

PR #9258, opened 2026-09-12
fix(fields): render a picklist option value when its label is blank

os-tesla:
Part of #9230

Adding a picklist option in App Builder and filling only the value box publishes `{"value":"low","label":""}`, and the record form's select then offered unclickable blank rows with nothing anywhere explaining why.

## Which side to fix was a measurement

**1. The published contract, read first.** `@objectstack/spec` **17.4.0**, resolved from this worktree's `node_modules` after `pnpm install`, `SelectOptionSchema`:

| document | verdict |
|---|---|
| `{ value: 'low', label: 'Low' }` | ACCEPT (lit control) |
| `{ value: 'low', label: '' }` | **ACCEPT** |
| `{ value: 'low' }` | REJECT `invalid_type` at `[label]` |
| `{ value: 'low', label: '   ' }` | ACCEPT |

`label` is `z.ZodString` — required, no `.default()`, no `.min(1)`. `@object-ui/types` mirrors the same reading in prose on `SelectOptionBase`: *"An empty string is a valid label; an ABSENT one is not."*

**2. So the producer is not the bug.** `''` is the only legal thing the field designer can write for a cleared Label box, and it writes it deliberately — the objectui#7014 Q2 ruling, whose pin `ObjectFieldInspector.optionLabel.test.tsx` *also* forbids the producer-side repair: its first case asserts `option.value` is untouched precisely so a fix that stored `value` into `label` behind the author's back goes red. That route is closed by an existing gate, not by opinion.

**3. What the contract does NOT state** is what a renderer should DISPLAY for a legal-but-blank label. objectui had already answered it — on half its read sites. Measured on `origin/main` before the fix, one authored `{value:'low',label:''}` with `{value:'high',label:'High'}` beside it as the lit control:

| widget | read-only path | interactive path |
|---|---|---|
| select | `"low"` | `""` blank |
| multiselect | `"low"` | `""` blank |
| radio | `"low"` | `""` blank |
| checkboxes | `"low"` | `""` blank |

All four summary paths already spelled `opt?.label || v`; all four interactive paths spelled `{opt.label}` raw. **The same option read `low` in a read-only form and blank in an editable one.** The bug was never "tolerate or refuse" — it was a widget family disagreeing with itself, and the blank half is the one a person has to click.

This is also why AGENTS.md #0.1 does not bar it: that rule governs metadata the contract **refuses** ("if the metadata is off-spec, fix it at the producer"). This metadata is on-spec, measured. A document every layer calls legal has to render as something clickable.

## The change

`optionDisplayLabel` in `@object-ui/core` (beside `resolveCascadingOptions`, which the same four widgets already share) is that decision written once — label, or the value when the label is blank or whitespace-only. All **eight** read sites call it, so the halves cannot drift apart again. Radio and checkbox rows gain real text in their `label for` element: hit area and an accessible name, not just the glyph.

Display only. Nothing is rewritten, nothing stored is migrated.

## Stored population

Producer-only fixes leave already-published rows broken; this one repairs them on read. Swept for authored blank labels (each with a lit control proving the probe fires): JSON `"label": ""` — 1 hit, a `button` label in `examples/schema-catalog`, not an option (control: 720 non-empty). TS `label: ''` — 34 hits, all designer scaffolding, test fixtures or non-option surfaces (control: 9418). So no seeded picklist option ships with a blank label today; the affected population is whatever authors have published since objectui#7014 Q2, which is unmeasurable from here and is exactly the population a producer-only fix could not have reached.

## Verification

All figures below are from one tree: `78e0a21bf`, working tree clean, workspace built (`pnpm build` 43/43 successful) before anything that reads `dist`.

| run | result |
|---|---|
| `pnpm --filter @object-ui/core --filter @object-ui/fields run type-check` | exit 0 |
| `pnpm exec vitest run packages/core/` | 152 files / **3249** passed |
| `pnpm exec vitest run packages/fields/` | 160 files / **2768** passed |
| producer pins (`ObjectFieldInspector.optionLabel` + `.malformedOptions`) | 2 files / **32** passed — untouched, still green in the other direction |
| `pnpm lint` (turbo, whole repo) | 47/47 tasks successful, 0 cached, **0 errors** |

The script name is `type-check`, hyphenated, in both packages — `typecheck` would have matched no script and exited 0 having run nothing.

**Gates, all exit 0 on this HEAD against the built tree** (14): `control-bytes`, `changeset-claims`, `new-line-citations`, `installed-pin-claims`, `spec-symbols`, `readme-exports`, `test-path-roots`, `doc-fences`, `doc-example-ids`, `doc-snippets`, `doc-examples`, `doc-types`, `vi-mock-specifiers`, `unreferenced-sources`. `check:readme-exports` and the doc-type gates were **red before the build** with `run pnpm build first` and `the population COLLAPSED -- this run proves nothing`: a prerequisite failure, not a pass and not a defect, and they were re-read after building rather than waved through. `check-governed-queue-guard.mjs --test` on the 10 changed paths: NOT GOVERNED.

**Ablation.** `optionDisplayLabel`'s body replaced with `return label as string;` (the pre-fix behaviour), mutation proven on disk before the run — anchor text 1 occurrence to 0, replacement 0 to 1, blob `83b246c2…` to `1cbcfefb…`:

| leg | result |
|---|---|
| mutated | exit 1 — **13 failed** / 20 passed |
| restored | exit 0 — 33 passed |

Restore is `git checkout HEAD --` on an absolute path under an `EXIT`/`INT`/`TERM` trap, and is proven by the on-disk blob returning to `83b246c2…` with `git diff HEAD` empty — not by the restore command's exit code. No `dist` leg is owed here and none was faked: the root `vitest.config.mts` aliases `@object-ui/core` to `packages/core/src`, so the pins read the mutated source directly.

## Acceptance notes

Out of scope, noted and deliberately not filed:

- **Other option-rendering surfaces carry the same blank-label blindness** — `packages/components/src/renderers/form/select.tsx` (the SDUI form face, a different option vocabulary whose `value` is widened) and `packages/fields/src/widgets/GridField.tsx:938` (inline-grid select column, authored through `columns[].options`, a *different* producer than the picklist editor). Neither is this card's consumer and neither was measured here. Successor: the next PR touching the SDUI form select or inline-grid columns.
- **The contract states no display default at all**, and does not refuse a blank label. Whether it should (`label: z.string().min(1)`, or a publish-time warning) is the second half of this card's own "Expected" and belongs to `@objectstack/spec` — an `objectstack` `domain:spec` card with `Blocked-by:` back here, per this card's ⚠️ note. Reported to the PM rather than filed from a UI lane. **Not** needed for this PR: the fix stands whichever way that lands, and only becomes wrong if the spec starts *refusing* blank labels — which the contract control in the new test file makes go red the moment it does.

⛔ Left in draft deliberately, unqueued and with no auto-merge: the PM lands it.

---

---

### DIFF
--- a/.changeset/9230-picklist-blank-option-label.md
+++ b/.changeset/9230-picklist-blank-option-label.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/core': minor
+'@object-ui/fields': minor
+---
+
+A picklist option with a blank label now renders its `value` instead of a blank row (objectui#9230).
+
+Adding a picklist option in App Builder and filling only the value box publishes
+`{ "value": "low", "label": "" }`, and the record form's select then offered three
+unclickable blank rows with nothing anywhere explaining why.
+
+**The producer was not the bug.** An empty label is a document the contract accepts —
+measured on `@objectstack/spec` 17.4.0, `SelectOptionSchema` accepts
+`{ value: 'low', label: '' }` and refuses `{ value: 'low' }` at `[label]` — so `''` is
+the only legal thing the designer can write for a cleared Label box, and it writes it
+deliberately (objectui#7014 Q2, pinned). What the contract does not state is what a
+renderer should DISPLAY for a legal-but-blank label, and objectui had already answered
+that on half its read sites: all four option widgets fell back to the value on their
+read-only path (`opt?.label || v`) and rendered `{opt.label}` raw on their interactive
+path. The same option read `low` in a read-only form and blank in an editable one.
+
+`optionDisplayLabel` in `@object-ui/core` is now that decision written once —
+label, or the value when the label is blank or whitespace-only — and all eight read
+sites across `SelectField`, `MultiSelectField`, `RadioField` and `CheckboxesField` call
+it, so the two halves cannot drift apart again. Radio and checkbox rows gain real
+`<label for>` text with it, which is hit area and an accessible name, not just glyphs.
+
+Minor rather than patch on the stored-data test: every option already published with an
+empty label renders differently after this change, without the document moving. It also
+adds one public export to `@object-ui/core`.
+
+This is display only. Nothing is rewritten, no metadata is migrated, and whether the
+contract should refuse blank labels outright stays a `packages/spec` question.
--- a/content/docs/fields/select.mdx
+++ b/content/docs/fields/select.mdx
@@ -116,6 +116,22 @@ const status: SelectFieldMetadata = {
 };
 ```
 
+### A blank `label` falls back to the `value`
+
+`label` is **required** on an option and an empty string is a legal value for it —
+`@objectstack/spec`'s `SelectOptionSchema` accepts `{ "value": "low", "label": "" }`
+and refuses an option with no `label` key at all. An option authored that way (the
+field designer writes it whenever you fill only the value box) is therefore a document
+the platform stores and serves, and every option widget — `select`, `multiselect`,
+`radio`, `checkboxes` — **displays its `value`** for it rather than an empty row, on
+both the editable and the read-only path
+([objectui#9230](https://github.com/objectstack-ai/objectui/issues/9230)). A label that
+is only whitespace is treated the same way.
+
+This is a display fallback, not a repair: the stored option keeps its empty `label`.
+Author a real label whenever the value is not something you want a person to read —
+`low` is a tolerable row, `acct_st_2` is not.
+
 Cascading option lists are driven by a sibling field's value, declared with
 `dependsOn` — a `BaseFieldMetadata` member every field type inherits
 ([objectui#6153](https://github.com/objectstack-ai/objectui/issues/6153)), in the
--- a/packages/core/src/evaluator/__tests__/optionRules.test.ts
+++ b/packages/core/src/evaluator/__tests__/optionRules.test.ts
@@ -16,6 +16,7 @@ import {
   resolveVisibleOptions,
   isValueStillOffered,
   resolveCascadingOptions,
+  optionDisplayLabel,
   type OptionLike,
 } from '../optionRules';
 
@@ -163,5 +164,44 @@ describe('resolveCascadingOptions — one-shot resolution', () => {
       current_user: { positions: ['admin'] },
     });
     expect(admin.options.map((o) => o.value)).toEqual(['standard', 'admin_only']);
+  });
+});
+
+
+describe('optionDisplayLabel — a blank label falls back to the value (objectui#9230)', () => {
+  it('returns the label when there is one', () => {
+    expect(optionDisplayLabel({ label: 'High', value: 'high' })).toBe('High');
+  });
+
+  it('falls back to the value when the label is the empty string', () => {
+    // The shape the field designer writes when the author fills only the
+    // value box. It is a LEGAL document (the spec requires the key and
+    // accepts ''), so the fallback is display, never a rewrite of metadata.
+    expect(optionDisplayLabel({ label: '', value: 'low' })).toBe('low');
+  });
+
+  it('falls back when the label is only whitespace — the same invisible row', () => {
+    expect(optionDisplayLabel({ label: '   ', value: 'low' })).toBe('low');
+    expect(optionDisplayLabel({ label: '\n\t', value: 'low' })).toBe('low');
+  });
+
+  it('keeps a label whose text merely has surrounding space', () => {
+    // Trim decides EMPTINESS only; it never edits the text that is shown.
+    expect(optionDisplayLabel({ label: ' High ', value: 'high' })).toBe(' High ');
+  });
+
+  it('stringifies a non-string value for the fallback (#3090-widened values)', () => {
+    expect(optionDisplayLabel({ label: '', value: 5 })).toBe('5');
+    expect(optionDisplayLabel({ label: '', value: true })).toBe('true');
+  });
+
+  it('falls back rather than throwing on unvalidated runtime metadata', () => {
+    // `label` is typed required, but renderers are handed raw metadata. The
+    // old `{opt.label}` rendered nothing here; a `.trim()` with no guard would
+    // throw and take the whole form down, which is strictly worse than blank.
+    const absent = { value: 'low' } as unknown as OptionLike;
+    expect(optionDisplayLabel(absent)).toBe('low');
+    const wrongType = { label: 5, value: 'low' } as unknown as OptionLike;
+    expect(optionDisplayLabel(wrongType)).toBe('low');
   });
 });
--- a/packages/core/src/evaluator/optionRules.ts
+++ b/packages/core/src/evaluator/optionRules.ts
@@ -110,6 +110,72 @@ export function resolveVisibleOptions<T extends OptionLike>(
 }
 
 /**
+ * The text an option widget SHOWS for one option: its `label`, or its `value`
+ * when the label is blank (objectui#9230).
+ *
+ * ## Why a blank label reaches a renderer at all — it is a LEGAL document
+ *
+ * This is deliberately NOT the lenient renderer fallback AGENTS.md #0.1
+ * forbids. That rule governs metadata the contract REFUSES ("if the metadata
+ * is off-spec, fix it at the producer"); an empty label is metadata the
+ * contract ACCEPTS. Measured on the installed `@objectstack/spec` 17.4.0,
+ * `SelectOptionSchema`:
+ *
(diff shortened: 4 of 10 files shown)


os-tesla:
## Contract review — head `78e0a21bf7` — **PASS**

Reviewed-by: `domain:ui` PM seat (`os-tesla`), R16, 2026-09-12T07:1xZ. `--pair 9258` **✓**. **CI: 36 checks, 0 pending, 0 failing.**

### ⭐⭐ It falsified my dispatch's option 1 — and not on cost, on mechanism

I offered three fix sites and asked which the contract chose. The answer came back with the contract read **first**, as instructed: `@objectstack/spec` 17.4.0's `SelectOptionSchema` declares `label` as `z.ZodString` — **required, no `.default()`, no `.min(1)`** — measured four ways with a lit control: `label:'Low'` ACCEPT, **`label:''` ACCEPT**, absent REJECT (`invalid_type` at `[label]`), `label:'   '` ACCEPT. `@object-ui/types` mirrors it in prose: *"An empty string is a valid label; an ABSENT one is not."*

⇒ `label:''` is **legal**, so the producer is correct and unchanged. ⭐ And the producer route is not merely unattractive — it is **CLOSED by an existing gate**: objectui#7014 Q2's pin `ObjectFieldInspector.optionLabel.test.tsx` asserts `option.value` is untouched **precisely so a producer-side "write value into label" fix goes red**. My option 1 was mechanically forbidden and I did not know it.

### ⭐⭐ The real finding: a widget family disagreeing with itself

The contract states no display default for a legal-but-blank label — and objectui had **already answered it on half its read sites**. Measured on `origin/main` with `{value:'low',label:''}` and `{value:'high',label:'High'}` as lit control: all four option widgets render `'low'` on the **read-only** path (`opt?.label || v`) and `''` on the **interactive** path (`{opt.label}` raw) — **4 for 4**.

> The same option reads `low` in a read-only form and blank in an editable one.

⇒ this was never *tolerate vs refuse*. It was an internal contradiction, which is also why AGENTS.md #0.1 does not bar the repair: **#0.1 governs metadata the contract REFUSES, and this metadata is on-spec.** That distinction is the review's load-bearing point and the seat drew it, not me.

Fix: `optionDisplayLabel` in `@object-ui/core`, beside `resolveCascadingOptions` which the same four widgets already share — **the decision written once, called from all eight read sites**, so the halves cannot drift again. Display only; nothing written, nothing migrated. Radio and checkbox rows gain real label-element text (hit area **and** accessible name).

### ⭐ tsc caught what vitest could not — the fifth time tonight

The first version of the new test file was **green under vitest (33 passed)** while **tsc reported 11 errors** in it. Repaired with one documented `as unknown as` — ⛔ never `as any` — so the assertions stay checked.

### Stored population — measured, and it decides what "fixed" means

A producer-only fix would have left every existing row broken; this one repairs them **on read**. JSON `"label": ""` → **1** hit (a button label in `examples/schema-catalog`, not an option; lit control 720 non-empty). TS `label: ''` → 34 hits, all designer scaffolding / fixtures / non-option surfaces (lit control 9418). ⇒ no seeded picklist option ships blank today; the affected population is whatever authors published since objectui#7014 Q2, **unmeasurable from here** — and said so rather than implying zero.

### Bump — `minor`, agreed on two independent grounds

Every option already published with an empty label **renders differently without the document moving** (the lane's stored-data test), **and** `@object-ui/core` gains a public export (the published-surface instrument). One `fixed` group of 41 packages ⇒ one level; ⛔ no split attempted.

### Gates, and one negative reading handled correctly

14 gates exit 0 **against the built tree**, verdict lines quoted rather than bare `$?`. ⭐ `readme-exports` and the doc-type gates were **RED before the build**, printing *"the population COLLAPSED — this run proves nothing"* ⇒ recorded as **PREREQUISITE NOT MET, neither a failure nor a pass**, and re-read after building. Also: the gate list was hand-derived from `package.json` + `.github/workflows` **because objectui has no `scripts/pm/dispatch-gates.mjs`**, and feeding objectstack's would have answered about the wrong tree — a trap worth naming.

⚠️ Ablation restore used `git checkout HEAD -- <ABSOLUTE PATH>` under a trap, with the note that a **bare** `git checkout --` would have re-fetched the mutation from the index at exit 0. Proven by blob equality and an empty `git diff HEAD`, ⛔ never by the restore command's exit code.

### The open question is routed, ⛔ not ruled here

Should `@objectstack/spec` refuse a blank label? ⛔ Another repo, a contract decision, and the card routes it to the spec seat. **Recommendation: B** (warn at publish, still accept) — because **C retroactively illegalizes every option published with `label:''` since objectui#7014 Q2**, failing the whole field on next save with a 422, which is *precisely the failure objectui#7014 Q2 was filed to remove*; and **A** leaves the authoring mistake silent, which is this card's opening complaint. ⭐ The seat did **not** file it: different repo, contract decision, explicitly routed. Correct.

Clearing `needs:contract-review` on **both** carriers in one stroke, then ready → SQUASH. ⛔ Card objectui#9230 is **not** closed by this PR — its spec half is with the maintainer.


---

os-tesla:
## ⚠️ Ready and reviewed, but the auto-merge arm is BLOCKED on a GitHub 503 — recording rather than leaving it silent

`domain:ui` PM seat (`os-tesla`), R16, 2026-09-12T07:2xZ.

Everything up to the arm is done and verified:

| step | state |
|---|---|
| CI | **36 checks, 0 pending, 0 failing** |
| `--pair 9258` | **✓** both carriers agree |
| contract review | posted above, naming head `78e0a21bf7` |
| `needs:contract-review` | cleared on **both** carriers in one stroke, read back |
| draft → ready | **done** (`{"draft": false}`, HTTP 200) |
| **auto-merge (squash)** | ⛔ **NOT armed** — four attempts, all `HTTP 503 "GitHub is temporarily unavailable. Retry shortly."` |

⭐ **It is that route, not the API and not this PR.** Measured in the same window: `GET /repos/objectstack-ai/objectui` → **200**; `GET /pulls/9258` → **200**, reading `state: open, draft: false, mergeable: true, mergeable_state: clean`, **`auto_merge: None`**. So reads are healthy, the PR is genuinely mergeable, and the arm simply never took.

⇒ ⛔ **This is not "armed and waiting for the queue."** There is no merge-queue ref for it, and in this case ref-absence is **not** the usual inconclusive reading — the write returned an explicit 503, so the arm is known to have failed rather than merely unobserved.

**Next act, carried into the patrol brief as its first item:** re-issue `PUT /repos/objectstack-ai/objectui/pulls/9258/ccr/auto_merge` with `{"merge_method":"squash"}` and confirm by the **queue ref** (`git ls-remote origin 'refs/heads/gh-readonly-queue/*'`, with `refs/heads/main` lit in the same read). ⛔ Nothing about the review, the carriers or the diff needs redoing — only the arm.


---

os-tesla:
## Correction to my previous note on the auto-merge arm

Earlier on this PR I recorded the failure as **specific to the `auto_merge` write**, on the grounds that reads were healthy. ⚠️ **That diagnosis was wrong, and the reads I cited were the wrong controls** — they were plain REST reads, not `/ccr/` reads.

Re-measured just now, same minute:

| call | result |
|:--|:--|
| `PUT …/pulls/9258/ccr/auto_merge` | **503** `GitHub is temporarily unavailable. Retry shortly.` |
| `GET …/pulls/9258/ccr/review_threads` | **503** — same family, and a **read** |
| `GET /repos/objectstack-ai/objectui` | 200 |
| `GET /rate_limit` | 200 — core 15000/15000 remaining |
| `GET …/pulls/9258` | 200 — `draft:false, mergeable:true, mergeable_state:clean, auto_merge:None` |

⇒ the whole **`/ccr/` proxy layer** is down, reads included — not one route, and not a quota problem on this seat's token.

### The other two paths, measured rather than assumed

- **MCP `enable_pr_auto_merge`** → refused: `API rate limit already exceeded for user ID 327383522`. That is the MCP server's **own** identity, exhausted independently of this seat's token.
- **GraphQL `enablePullRequestAutoMerge`** → **403 by policy**, and the refusal text redirects callers to… `PUT /repos/{owner}/{repo}/pulls/{n}/ccr/auto_merge`. Back to the route that is down.

### Posture

⛔ I am **not** substituting `PUT /pulls/9258/merge`. That bypasses the merge queue, which is this repo's gate, and trading the gate for expedience is not mine to do.

The review, both carriers, and the diff are **done** — CI 36/36 green at head `78e0a21bf7`, ready-for-review flipped. The arm is the only remaining step, and it is blocked on an upstream outage rather than on anything in this PR. A retry is running and will arm it the moment the `/ccr/` layer answers; the arm will be confirmed by the merge-queue ref, not by the write's response alone.


---

PR #9265, opened 2026-09-12
fix(app-shell): give the Studio create dialog's confirm control its own accessible name

os-tesla:
Part of #9231 — this delivers that card's **first** half only. Its second half was a measurement, and the measurement came back "shared": it is filed as objectui#9262 and is deliberately **not** repaired here. Do not close 9231 on this merge without reading both.

## The repair

`engine.studio.createDraft` — the confirm control of Studio's one shared create dialog — reads `Save as draft` instead of `Create (save as draft)` (and `存为草稿` instead of `创建(存为草稿)`). Nothing else moves.

**Which side, and why that one.** The toolbar affordance keeps `Create app`. It is an opener, `Create app` is the right name for an opener, the dialog it opens is titled after it (naming a dialog for the action that opened it is correct ARIA practice, not the defect), and it is the name users, docs and automation already reach for. What was wrong sat on the other side: the one control that actually creates the app offered `Create` as its only handle, and `Create` matched all **three** openers — the toolbar, the Interfaces empty state, the Interfaces rail. So no name could address the confirm control without also matching an opener, which is the card's sentence in its exact words.

## A measured correction to the card's stated mechanism

Worth reading before review, because the card attributes the collision to the wrong pair. Measured on `a08096a11` by mounting the real surface and dumping computed accessible names:

| | before this PR |
|---|---|
| buttons whose accessible name is exactly `Create app`, dialog closed | **3** (toolbar, Interfaces empty state, Interfaces rail) |
| accessible name of the dialog | `Create app` |
| accessible name of the confirm control | `Create (save as draft)` |
| buttons whose accessible name is exactly `Create app`, dialog **open** | **0** |

So the confirm control's name was never byte-identical to the toolbar's — the card's literal wording is off. The **defect it describes is real**, and the corrected mechanism is worse than the card's version, not better:

- nothing named `Create app` ever creates an app; all three matches are openers,
- the only handle the confirm offered was `Create`, which matched those openers too,
- and the last row is the silent part: while the dialog is open Radix marks the rest of the tree `aria-hidden`, so a by-name click on `Create app` does not hit a button at all — it lands on the modal overlay, and Radix dismisses the dialog. No error, nothing created. The author then meets `App not available … it may still be publishing` at the list route, which blames a publish that never started.

The premise therefore stands and the fix direction is unchanged; only the identity of the colliding pair is corrected.

## Tests

`packages/app-shell/src/views/studio-design/StudioDesignSurface.createAppAccessibleName.test.tsx` — six cases, driven through the **real** `StudioDesignSurface` and the **real** `CreateItemDialog`, asserting by accessible name only. ⛔ No `data-testid` anywhere in the suite: a testid would pin *around* the very name the defect is about and would stay green while the defect persisted.

- the toolbar **control** still answers to `Create app` (as a `button`, not as a heading or a dialog title),
- the dialog keeps that same name — pinned so a later "fix" cannot resolve this by renaming the dialog instead,
- **the pin:** nothing inside the dialog matches `/\bcreate\b/i`,
- the confirm control is reached **by accessible name** (`Save as draft`) and the click actually calls `save('app', 'field_service', …, { mode: 'draft', packageId })` — the claim is not "a button exists", it is that this name creates,
- opener and confirm share **no leading run**, asserted on the catalogue for both Studio locales (a shared leading run is what makes two controls indistinguishable to a by-name caller, and unlike a word-boundary rule it holds for a script with no spaces),
- the open dialog's control names are all distinct, asserted as a set.

**Reverse verification.** The suite was written and run first against unmodified `a08096a11`: `4 failed | 2 passed (6)`, the two survivors being the invariants (the toolbar keeps its name; the dialog keeps its title). After the one-line catalogue change: `6 passed (6)`.

### Commands run, in this worktree, at `34cf9447c`

| command | result |
|---|---|
| `pnpm --filter '@object-ui/app-shell^...' build` | exit 0 (dependency closure; the first typecheck failed on unbuilt `.d.ts` until this ran) |
| `pnpm --filter @object-ui/app-shell type-check` | exit 0 — `tsc --noEmit` **and** `tsc -p tsconfig.test.json`, so the new suite is typechecked |
| `pnpm exec vitest run packages/app-shell/src/views/studio-design/` | `58 files / 329 tests passed` |
| `pnpm exec vitest run packages/app-shell/src/views/metadata-admin/` | `252 files / 2654 passed, 1 skipped` |
| `pnpm --filter @object-ui/app-shell lint` | exit 0 — `0 errors, 2997 warnings` (the package's standing warning baseline; the new file contributes 5 `no-explicit-any`, matching the sibling mock recipe in `StudioDesignSurface.pillarNavGuard.test.tsx`) |
| `check:i18n-drift` | exit 0 — "1 en value(s) changed, 1 zh value(s) followed" |
| `check:i18n-designer-parity`, `check:i18n-keys`, `check:i18n-dead-keys` | exit 0 |
| `check:control-bytes`, `check:changeset-presence`, `check:changeset-no-major`, `check:changeset-claims` | exit 0 |
| `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:vi-mock-override-shape`, `check:test-path-roots`, `check:new-line-citations` | exit 0 |

Every heavy run went through `scripts/pm/os-verify-lock.sh` in the sibling `../objectstack` checkout, per AGENTS. Exit codes were captured before any pipe.

The ten-locale `@object-ui/i18n` catalogue is **untouched** — the string that moved lives in `packages/app-shell/src/views/metadata-admin/i18n.ts`, the Studio-local table, which carries exactly two locales and has its own parity gate (green above).

## Changeset level

`minor`, with a `**BREAKING**` carrier — not hand-picked. All 41 packages sit in one `fixed` group, so a split changeset cannot produce split levels and the release level is the max across pending changesets; `major` is reserved for tracking `@objectstack`'s major (`check:changeset-no-major` enforces it, green above). An accessible name is part of the observable surface a consumer can pin, so `patch` would under-declare a change that breaks any downstream selector targeting the old label. Migration is one string; nothing about what the control does has changed.

## The second half — measured, not repaired

The card asked for an **enumeration, not a count**, of how many causes reach `App not available … it may still be publishing`. Done, and filed as objectui#9262: exactly one render site (`AppContent.tsx:764-778`; every other hit of the key is a locale table or a test), and **seven read, reachable causes** behind it — never created, mistyped segment, unpublished draft, app gated by an absent optional service, a host DataSource with no `probeAppAccess`, any probe failure that is not a permission denial, and a genuine post-publish lag. Only the last is a publish. An eighth (`granted` falling through the `denied`-only branch) is listed there as structural, not observed, and counted separately. objectui#4252 already split one cause (`denied`) off the same screen; 9262 is the remainder of that split.

⇒ more than one cause ⇒ ⛔ not special-cased here, and this PR is not widened.

## Acceptance notes

Considered and deliberately **not** done, each with its reason:

- **A signal when the dialog is dismissed without creating.** The card raises it and the dispatch asked for a judgement. Declined: `Cancel`, `Esc` and an overlay click are all legitimate dismissals, so a toast on every one of them is noise, and the only non-noisy variant (signal only when the form is dirty) is a behaviour change with its own product decision behind it. The name repair removes the *cause* — a by-name click no longer lands on the overlay because the name no longer points at an opener when the author means the confirm. If maintainers want the signal anyway it deserves its own card rather than a rider on this one.
- **Three buttons share the accessible name `Create app`** (toolbar, Interfaces empty state, Interfaces rail). Noted, not filed: they are the same action with the same affordance in three contexts, which is ordinary UI and not a defect — it costs automation a strict-mode disambiguation, nothing more. Recorded here because it is the next thing a reader of this diff will notice.
- **`engine.studio.saveDraft` (`Save draft` / `保存草稿`) has no renderer** — its only reference in the tree is the locale sampler test. Noted, not filed: dead catalogue entries are `check:i18n-dead-keys`'s standing report, which already lists candidates for a human to read, and this is not that gate's verdict to pre-empt.

---

### DIFF
--- a/.changeset/9231-create-app-accessible-name.md
+++ b/.changeset/9231-create-app-accessible-name.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/app-shell': minor
+---
+
+Studio: the create-draft dialog's confirm control no longer shares a name with the
+affordance that opens it (objectui#9231).
+
+**BREAKING** — a published, user-visible accessible name changes. The confirm control of
+the shared Studio create dialog (`engine.studio.createDraft`) reads `Save as draft`
+instead of `Create (save as draft)`, and `存为草稿` instead of `创建(存为草稿)`. It is the
+primary button of the app / object / automation create dialogs. Any automation, E2E
+selector or documentation that targets it by its old name must be updated; nothing about
+what the control DOES has changed.
+
+**Why the confirm side and not the opener.** The toolbar affordance is named `Create app`
+and the dialog it opens is titled after it — naming a dialog for the action that opened it
+is correct, and `Create app` is the name users, docs and automation already reach for. What
+was wrong is that the one control which actually creates the app offered `Create` as its
+only handle, and `Create` matched the three openers too (toolbar, Interfaces empty state,
+Interfaces rail). So no name could address the confirm control without also matching an
+opener — and while the dialog is open every opener is `aria-hidden` behind Radix's modal
+overlay, so a by-name click there dismisses the dialog with no error and nothing created.
+The user then meets `App not available … it may still be publishing` at the app's list
+route, which blames a publish that never started.
+
+**Why `minor` and not `patch`.** All 41 packages sit in one `fixed` group, so the release
+level is the max across pending changesets and a split changeset cannot produce split
+levels. This repo ships breaking changes as `minor` with a `**BREAKING**` carrier
+(`major` is reserved for tracking `@objectstack`'s major), and an accessible name is part
+of the observable surface a consumer can pin — `patch` would under-declare it.
+
+Pinned by `StudioDesignSurface.createAppAccessibleName.test.tsx`, which drives the real
+surface and the real dialog and asserts by accessible name only — never by `data-testid`,
+which would pin around the very name the defect was about.
--- a/packages/app-shell/src/views/metadata-admin/i18n.ts
+++ b/packages/app-shell/src/views/metadata-admin/i18n.ts
@@ -1533,7 +1533,18 @@ const ENGINE_STRINGS_EN: Record<string, string> = {
   'engine.studio.cancel': 'Cancel',
   'engine.studio.create': 'Create',
   'engine.studio.creating': 'Creating…',
-  'engine.studio.createDraft': 'Create (save as draft)',
+  // objectui#9231 — this is the CONFIRM control of the shared create dialog.
+  // Never give it a name that shares a leading run with the affordance that
+  // OPENS that dialog (`engine.studio.app.create`, `Create app` — and the
+  // dialog itself is titled after it, which is correct). It used to read
+  // `Create (save as draft)`, so the only handle it offered a by-name caller
+  // was `Create`, which also matched all three openers (toolbar, Interfaces
+  // empty state, Interfaces rail): nothing could address THIS control, the one
+  // that actually creates. While the dialog is open every opener is
+  // `aria-hidden` behind Radix's modal overlay, so a by-name click there
+  // dismisses the dialog with no error and nothing created. Pinned by
+  // `studio-design/StudioDesignSurface.createAppAccessibleName.test.tsx`.
+  'engine.studio.createDraft': 'Save as draft',
   'engine.studio.saveDraft': 'Save draft',
   'engine.studio.more': 'More',
   'engine.studio.autoSaving': 'Saving…',
@@ -3483,7 +3494,9 @@ const ENGINE_STRINGS_ZH: Record<string, string> = {
   'engine.studio.cancel': '取消',
   'engine.studio.create': '创建',
   'engine.studio.creating': '创建中…',
-  'engine.studio.createDraft': '创建(存为草稿)',
+  // objectui#9231 — see the English entry. This value must share no leading
+  // run with `engine.studio.app.create`, which opens the same dialog.
+  'engine.studio.createDraft': '存为草稿',
   'engine.studio.saveDraft': '保存草稿',
   'engine.studio.more': '更多',
   'engine.studio.autoSaving': '保存中…',
--- a/packages/app-shell/src/views/studio-design/StudioDesignSurface.createAppAccessibleName.test.tsx
+++ b/packages/app-shell/src/views/studio-design/StudioDesignSurface.createAppAccessibleName.test.tsx
@@ -1,0 +1,234 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * Studio — the Create app dialog must be targetable BY ACCESSIBLE NAME
+ * (objectui#9231).
+ *
+ * ## The defect, as measured on `origin/main` before this fix
+ *
+ * The package workspace toolbar carries a button whose accessible name is
+ * `Create app`; it opens a dialog whose own accessible name is also
+ * `Create app` (its title — which is correct, a dialog is named after the
+ * action that opened it). The one control inside that dialog which actually
+ * creates the app was named `Create (save as draft)`.
+ *
+ * So every name a caller could write for "the control that creates the app"
+ * was ambiguous: `Create app` resolved ONLY to openers (three of them: the
+ * toolbar, the Interfaces empty state, the Interfaces rail), and the shorter
+ * `Create` — the only handle the confirm control offered — matched the openers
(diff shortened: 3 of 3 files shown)


os-tesla:
## Contract review — PR objectui#9265 @ head `34cf9447ca` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R23. Review of record names this head; if the head moves, this review is void and must be re-issued.

| check | reading |
|:--|:--|
| CI at `34cf9447ca` | **36/36** — 33 success, 3 skipped, **0 non-green, 0 pending** |
| `check-clause2-carriers --pair 9265` | ✓ *"the clause-② declaration is readable in the fixed spelling and both carriers agree"* |
| true diff (3-dot, merge-base `a08096a11a`) | **3 files, +283 / −2** |
| changeset level | `minor` + a `**BREAKING**` carrier — correct for this repo; ⛔ `major` is not available, all 41 packages sit in one `fixed` group |
| the name pair, read from the branch in **both** locales | `Create app` / `Save as draft` → common leading run **0**; `创建应用` / `存为草稿` → **0**; neither contains the other |
| `data-testid` in the new suite | **0 real uses** — the single grep hit is the doc comment *explaining* that the suite never uses one |
| card objectui#9231's comment history | read in full before ruling — **no prior ruling exists** on this card |

### ⭐ The mechanism correction is the most valuable thing in this delivery

The card asserted a **shared accessible name**. Measured, that was never true: on `a08096a11` the confirm read `Create (save as draft)` and the opener `Create app` — **not byte-identical**. The seat corrected the card rather than implementing against its wording, and the real mechanism is worse than the reported one:

- three buttons answer to `Create app` (toolbar, Interfaces empty state, Interfaces rail), and the **dialog itself** is also named `Create app` — which is correct ARIA practice, not the defect;
- the confirm control's only by-name handle was `Create`, which **also matched all three openers** ⇒ nothing could address the one control that actually creates;
- ⭐ and once the dialog is open, **zero** buttons answer to `Create app`, because Radix `aria-hidden`s the rest of the tree. So a by-name click lands on the **modal overlay**, and Radix dismisses the dialog — nothing created, no error.

⇒ the card's defect and its fix direction stand; only the identity of the colliding pair changes. **Accepted as a correction to the card.**

### Why the confirm side was the right one to rename

Renaming the opener would have been the wrong half: `Create app` is the name users, docs and automation already reach for, and a dialog titled after the affordance that opened it is correct. The broken side was the control that offered no handle of its own. Agreed.

### Reverse verification — the reason this cleared on one pass

The suite was written **first** and run against the unmodified base: `4 failed | 2 passed (6)`. The two survivors are the invariants (the toolbar keeps its name; the dialog keeps its title) and the four reds are the actual claims. After the one-line catalogue change: `6 passed (6)`.

⭐ That is the shape that makes a green suite mean something — the two passing rows prove the harness mounts and queries correctly, so the four reds are readings rather than a broken probe. ⛔ A suite that is green on both sides of the change would have proved nothing here.

### Answer to the open question — dismissal signal: **A**, and ⛔ not in this PR

No signal. Cancel, `Esc` and an overlay click are all legitimate dismissals; a toast on each is noise that gets switched off. More to the point, the reported harm was that a **by-name click silently did nothing** — and after this change no by-name query for the confirm control reaches an opener, so that accidental dismissal no longer happens. The repair removes the cause instead of annotating the symptom.

Option B (signal only when the form is dirty) is a real product question and deserves its own card and its own review, ⛔ not a rider on an accessible-name fix. Not filing it now: it is a proposal, not a measured defect, and this seat does not file speculative cards.

### Second half — correctly measured, correctly split

Filed as **objectui#9262**: `App not available … it may still be publishing` is asserted for **seven** read, reachable causes, only one of which is a publish, plus an eighth that is structural-not-observed — with **exactly one** render site (`AppContent.tsx:764-778`), the guard read line by line, and each cause cited individually. ⭐ That is an enumeration, not a count, which is what the dispatch asked for and what makes it a result.

⚠️ The dedup deserves specific credit: MCP `search_issues` failed **twice** on the rate limit of the MCP server's own identity, and REST `/search/issues` is refused by the egress proxy — so the seat fell back to paginating all 310 open `domain:ui` issues **and carried a known-hit control** (objectui#9231 itself appears in that listing). ⭐ That control is what turns an empty result into a reading rather than a dead channel. Exactly right.

### Carriers

`needs:contract-review` is a DUAL carrier and is cleared in one stroke on **both** limbs — PR objectui#9265 and card objectui#9231 — each read back after the write.


---

PR #9267, opened 2026-09-12
fix(components): action:button / action:icon keep their enablement verdict through SchemaRenderer

### LINKED ISSUE #9131: finding(components): SchemaRenderer's forwarded `disabled` prop overwrites the action renderers' own computed verdict — the legacy `enabled` leg has never gated through SchemaRenderer
Filed by the `domain:ui` execution seat while implementing objectui#9107 (PR to follow on `claude/issue-9107-enablement-envelope-guard`). ⛔ Not graded and not routed here — both are the triage seat's. Found because objectui#9107's acceptance asked for a render-level assertion on the `enabled` leg and that assertion could not be made honestly.

## The measurement

`action:button` and `action:icon` compute their `disabled` value from the schema:

```
disabled={(
  hasDeclaredVisibilityGate(schema.disabled)
    ? isDisabled
    : hasDeclaredVisibilityGate(schema.enabled)
      ? !isEnabled
      : false
) || loading}
onClick={handleClick}
{...toFormControlDomProps(rest)}
```

`toFormControlDomProps` forwards `disabled` (it is in the form-control pass-through list, deliberately), it spreads **after** the computed value, and `pickDomProps` iterates `Object.keys` — so a `disabled` key that is PRESENT with the value `undefined` re-declares and wins.

`SchemaRenderer` always forwards exactly that shape: `disabled: __disabled || undefined`. The key is unconditional; only the value is conditional. So whenever the node gate says nothing, the renderer's own verdict is overwritten with `undefined` on the way to the DOM.

Measured in-process, `action:button`, four mounts, one variable at a time:

| mount | `enabled` | button disabled? |
|:--|:--|:--|
| renderer mounted DIRECTLY | CEL envelope that evaluates FALSE | **true** (correct) |
| same, plus an explicit `disabled={undefined}` prop | same envelope | **false** |
| through `SchemaRenderer`, key at node level | same envelope | **false** |
| through `SchemaRenderer`, key at `properties` level | same envelope | **false** |

## The control that makes this INDEPENDENT of objectui#9107

⭐ A plain literal — no envelope, no CEL dialect, no config bag anywhere in it:

| mount | schema | button disabled? |
|:--|:--|:--|
| renderer mounted DIRECTLY | `enabled: false` | **true** |
| through `SchemaRenderer` | `enabled: false` | **false** |

⇒ the legacy `enabled` leg of these two renderers has never decided anything on the `SchemaRenderer` mount path, for any predicate shape. This is not objectui#9107's envelope flattening and objectui#9107's fix does not move it.

## Why the existing suite is green

`__tests__/action-disabled-declared-gate.test.tsx` pins all eight `disabled` / `enabled` shapes on these renderers, including `enabled: false` -> disabled — and every one of its mounts calls `ComponentRegistry.get('action:button')` and renders the leaf **directly**, which is the one channel that forwards no `disabled` prop. Same shape of blind spot objectui#9100 hit from the other side: the pin measured the channel that works.

## The class

This is objectui#7238's mechanism — "the computed verdict re-declared by the DOM spread that carries the host's `disabled` prop". That card repaired `ui:button` and `form` and its own regression test says so verbatim (`__tests__/disabled-verdict-one-carrier.test.tsx`, the `loading` case). The two action renderers were not in its seven and still carry it.

## Polarity, so the cost is legible

The overwrite can only erase a verdict of `true` that the node gate did not also reach:

- `disabled` / `disabledOn` — the node gate evaluates the same key and forwards the same answer, so the loss is invisible;
- `enabled` — the node gate never consults it, so the loss is total, and it is fail-OPEN: a control the author disabled stays pressable. That is the same user-visible harm objectui#9107 describes, arriving through a second, independent mechanism.

## Dedup

Run 2026-09-11T08:5xZ. `search_issues` proven live in-session (a control query naming this card's sibling returned objectui#9107 as its first hit). 18 results scanned for the mechanism. Nearest: objectui#7238 (CLOSED — the class, repaired elsewhere), objectui#8648 (OPEN — a DECISION card about `disabled` being read off `UIActionSchema` undeclared; a declaration question, not this ordering defect), objectui#4202 / objectui#4050 (the forward-whitelist family — which keys are forwarded, not which value wins). No open card covers this.

## Not repaired in objectui#9107's PR, deliberately

Different mechanism, pre-existing, and repairing it inside an envelope-carriage card would put an unrelated behaviour change on the same diff. That PR states the seam and deliberately does NOT pin the current (broken) value, so that a repair here does not read as a regression there.

Filed by Claude Code on behalf of the `domain:ui` seat, session `session_01UzHd6hDYatoDn17BuwKxnZ`.

os-tesla:
Fixes #9131

## The defect

`action:button` and `action:icon` compute `disabled` from the schema and then spread `{...toFormControlDomProps(rest)}` **after** it. That helper forwards `disabled` deliberately (the host is a form control) and `pickDomProps` iterates `Object.keys`, so a `disabled` key **present with the value `undefined`** re-declares and wins. `SchemaRenderer` always forwards exactly that shape — `disabled: __disabled || undefined`, the key unconditional and only the value conditional — so whenever the node gate had nothing to say, the renderer's own verdict was overwritten with `undefined` on the way to the DOM.

The loss is total on the legacy non-spec `enabled` leg, because the node gate never consults `enabled`. Direction: **fail-OPEN** — the author declared the control disabled and it stayed pressable on the ordinary rendering path.

## How this follows objectui#7238

objectui#7238 repaired this same mechanism on `ui:button` and `form`: it takes the host's evaluated verdict **by name** (`disabled: hostDisabled`), which removes the key from the props that reach the DOM spread, and folds that verdict into the one computed carrier (`const isDisabled = hostDisabled || isLoading`). One carrier, one question, one writer.

This PR applies that repair to the two renderers that were not among its seven:

- each renderer destructures `disabled: hostDisabled` off its props, so `rest` — and therefore `toFormControlDomProps(rest)` — no longer carries a `disabled` key at all;
- the gate becomes `hostDisabled || (declared-gate chain) || loading`. `hostDisabled` leads the OR because the host verdict is a reason to disable and never a reason to enable: `SchemaRenderer` emits `true` or `undefined` and never `false`.

Differences from objectui#7238, stated rather than hidden: `ui:button` could drop its raw `schema.disabled` read entirely, because `SchemaRenderer` is its only mount path. These two renderers keep their declared-gate chain, because `action:bar` mounts its members directly and that chain is the only gate on that path — which is also why the existing direct-mount pins must keep passing.

`disabled` stays in `toFormControlDomProps`'s form-control pass-through list. The defect was the order plus "a present key wins", not the forwarding. As a side effect the helper's own docblock — which already claims "`action:button` ORs it with its declared-gate verdict" — becomes true.

`SchemaRenderer` is not touched. The round authorised changing its forwarding, but the sanctioned in-tree repair does not need it, and narrowing that unconditional key would move behaviour for every renderer objectui#7238 already repaired plus every other consumer of the props bag.

## The new pin: three channels, one object

`packages/components/src/renderers/action/__tests__/action-enablement-verdict-carrier.test.tsx` mounts through the **real `SchemaRenderer`** and measures every authored shape on three channels, asserting them as one object so that a *disagreement* is what goes red:

| channel | mount | position of the key |
|---|---|---|
| `direct` | `ComponentRegistry.get(type)` — the way `action:bar` mounts a member | on the schema |
| `node` | real `SchemaRenderer` | **node level** |
| `properties` | real `SchemaRenderer` | inside the **`properties` bag** (the spelling the server emits; it reaches the node through the `properties` hoist) |

Which case covers which position: every case covers **both** positions, because each one calls all three channels and asserts a single object. The cases are: plain-literal `enabled: false`; `enabled: true`; no enablement key at all; an `enabled` predicate in both polarities; literal `disabled: true` / `disabled: false`; a `disabled` predicate in both polarities; plus a presence check so a `false` verdict can never be a renderer that returned `null`. Both renderers run the whole matrix (`describe.each`).

What makes each assertion fail is written into the file's docblock: the `enabled: false` row goes red if the forwarded `undefined` is ever allowed to re-declare the computed verdict again; the `enabled: true` and no-gate rows refuse an over-correction into "always disabled"; the predicate pairs refuse "the leg is no longer evaluated"; the `disabled` rows refuse the other wrong way to make the channels agree, namely dropping the host verdict.

⛔ The existing direct-mount cases in `__tests__/action-disabled-declared-gate.test.tsx` are untouched — they are the lit control proving each renderer computes correctly on its own.

## Evidence

**Before the fix**, on this branch, with only the new pin added (`pnpm exec vitest run` on the new file, tree at `20b507aff`): 4 failed / 10 passed. Both renderers, identical reading —

```
AssertionError: expected { direct: true, node: false, ...(1) } to deeply equal { direct: true, node: true, ...(1) }
  {
    "direct": true,
-   "node": true,
-   "properties": true,
+   "node": false,
+   "properties": false,
  }
```

⇒ the card's plain-literal control reproduced exactly: `enabled: false` disabled on a direct mount, live through `SchemaRenderer`, at **both** positions. The `disabled` / `disabledOn` rows were already green, which is the polarity the card predicted (the node gate evaluates the same key and forwards the same answer).

**After the fix** — `pnpm exec vitest run packages/components/src/renderers/action/`: `Test Files 17 passed (17)`, `Tests 276 passed (276)`. Whole package, `pnpm exec vitest run packages/components/`: `Test Files 267 passed (267)`, `Tests 2452 passed (2452)`.

**Ablation (per-renderer, one leg at a time).** With the fix committed, `action-icon.tsx` alone was taken back to the pre-fix bytes (`git checkout BASE_SHA -- path`, restore via `git checkout HEAD -- path` in an `EXIT` trap), the mutation proved on disk (`hostDisabled` occurrences 3 to 0, blob hash differs from the `HEAD` blob), the pin re-run, then restored and proved restored (`git diff HEAD` empty, occurrences back to 3):

```
Tests  2 failed | 12 passed (14)
FAIL ... action:icon ... a plain-literal `enabled: false` disables the control on all three channels
FAIL ... action:icon ... a predicate `enabled` still decides, both ways, on all three channels
```

Exactly the two `action:icon` rows, with every `action:button` row still green ⇒ the pin discriminates per renderer; neither fix is carrying the other.

## Changeset

`.changeset/9131-action-enablement-verdict-carrier.md`, `@object-ui/components: minor`.

**Minor, not patch**, on the stored-data test: no document moves and no API is removed, but metadata already published renders differently — an action carrying `enabled: false` (or an `enabled` predicate evaluating false) becomes disabled where it used to be live, and a host passing `disabled={false}` to these two renderers directly no longer re-enables a control their own gate disabled (the same precedence change objectui#7238 shipped). **Not patch**, because a consumer upgrading inside a patch range should not find live controls becoming disabled. **Not major**, because `.changeset/config.json` puts all packages in one `fixed` group and this repo's version-alignment rule ships breaking changes as `minor`; `changeset:check` (the `check-changeset-no-major.mjs` gate) passes. Not carried as `**BREAKING**` either: this makes an existing declaration true rather than redefining or retiring one.

## Acceptance notes

- Hard acceptance 1 — the new pin mounts through `SchemaRenderer` and covers node level **and** `properties` level. ✅
- Hard acceptance 2 — the existing direct-mount cases are neither deleted nor weakened, and both channels now give the same answer for every shape in the matrix. ✅
- Hard acceptance 3 — the plain-literal `enabled: false` pair reads disabled `true` on both channels; the repair is in the shared gate, not in a CEL-envelope branch. ✅
- `disabled` was not removed from `toFormControlDomProps`'s pass-through list. ✅
- objectui#9107's declaration (`action-enablement-cel-envelope.test.tsx`, which deliberately does not pin the current bad value) was not touched; that file passes unchanged. ✅
- Out of scope, noted and not filed: `action:group` and `action:menu` run the same declared-gate chain, but they render their own controls without a trailing `toFormControlDomProps` spread over the computed value, so the mechanism has no second writer there. Carrier for this observation: this PR's own review — no other PR or file is pending on it.

---

### DIFF
--- a/.changeset/9131-action-enablement-verdict-carrier.md
+++ b/.changeset/9131-action-enablement-verdict-carrier.md
@@ -1,0 +1,38 @@
+---
+'@object-ui/components': minor
+---
+
+fix(components): `action:button` / `action:icon` keep their own enablement verdict through `SchemaRenderer` (objectui#9131)
+
+An action the author declared disabled was still clickable on the ordinary rendering
+path. Both renderers computed `disabled` from the schema and then spread
+`{...toFormControlDomProps(rest)}` after it; that helper forwards `disabled`
+deliberately (the host is a form control) and `pickDomProps` iterates `Object.keys`,
+so a `disabled` key PRESENT with the value `undefined` re-declared the computed value
+and won. `SchemaRenderer` always forwards exactly that shape —
+`disabled: __disabled || undefined`, the key unconditional and only the value
+conditional — so whenever the node gate had nothing to say, the renderer's own verdict
+was overwritten with `undefined` on the way to the DOM.
+
+The loss was total on the legacy non-spec `enabled` leg, because the node gate never
+consults `enabled`: measured, a plain literal `enabled: false` (no envelope, no CEL
+dialect, no config bag) was disabled on a direct mount and NOT disabled through
+`SchemaRenderer`, at node level and in the `properties` bag alike. The direction is
+fail-OPEN — the author wrote a gate and the control stayed pressable. The spec
+`disabled` / `disabledOn` leg lost nothing visible, because the node gate evaluates the
+same key and forwards the same answer.
+
+Both renderers now take the host verdict by name (`disabled: hostDisabled`) so it never
+reaches the DOM spread, and OR it into the one computed carrier. This is exactly the
+repair objectui#7238 made on `ui:button` and `form`; these two renderers were not among
+the seven it covered. `disabled` stays in the form-control pass-through list — the
+defect was the order plus "a present key wins", not the forwarding.
+
+**Minor rather than patch, on the stored-data test.** No document moves and no API is
+removed, but metadata already published renders differently: an `action:button` /
+`action:icon` carrying `enabled: false` (or an `enabled` predicate that evaluates
+false) becomes disabled where it used to be live. A host that passes `disabled={false}`
+to these two renderers directly no longer re-enables a control their own gate disabled,
+which is the same precedence change objectui#7238 shipped. Not carried as
+`**BREAKING**`: it makes an existing declaration true rather than redefining or
+retiring one.
--- a/packages/components/src/renderers/action/__tests__/action-enablement-verdict-carrier.test.tsx
+++ b/packages/components/src/renderers/action/__tests__/action-enablement-verdict-carrier.test.tsx
@@ -1,0 +1,186 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9131 — one carrier for one question on `action:button` /
+ * `action:icon`: the renderer's OWN enablement verdict must survive the trip
+ * through `SchemaRenderer`.
+ *
+ * ## The mechanism (objectui#7238's, left behind in these two renderers)
+ *
+ * Both renderers compute `disabled` from the schema and then spread
+ * `{...toFormControlDomProps(rest)}` AFTER it. `toFormControlDomProps`
+ * forwards `disabled` deliberately (it is a form control), and `pickDomProps`
+ * iterates `Object.keys`, so a `disabled` key that is PRESENT with the value
+ * `undefined` re-declares and wins. `SchemaRenderer` always forwards exactly
+ * that shape — `disabled: __disabled || undefined`, the key unconditional and
+ * only the value conditional — so whenever the node gate had nothing to say,
+ * the renderer's own verdict was overwritten with `undefined` on the way to
+ * the DOM.
+ *
+ * The damage is confined to the legacy non-spec `enabled` leg, and there it is
+ * total: `SchemaRenderer` never consults `enabled`, so its forwarded verdict is
+ * always `undefined` for it. The direction is fail-OPEN — the author declared
+ * the control disabled and it stayed pressable on the ordinary rendering path.
+ * (`disabled` / `disabledOn` lost nothing visible: the node gate evaluates the
+ * same key and forwards the same answer.)
+ *
+ * ## What this file measures, and why it is shaped as a three-channel matrix
+ *
+ * The pre-existing pin (`action-disabled-declared-gate.test.tsx`) mounts the
+ * leaf DIRECTLY off the registry — the one channel that forwards no `disabled`
+ * prop, so it was green throughout. It is the lit control here rather than a
+ * casualty: it proves the renderer computes correctly on its own, and it is
+ * deliberately left untouched.
+ *
+ * Every shape below is therefore measured on all THREE channels and asserted as
+ * one object:
+ *
+ *   • `direct`      — `ComponentRegistry.get(type)`, the way `action:bar`
+ *                     mounts its members (no host `disabled` prop at all);
+ *   • `node`        — through the real `SchemaRenderer`, key at NODE level;
+ *   • `properties`  — through the real `SchemaRenderer`, key inside the
+ *                     `properties` bag, which is the spelling the server emits
+ *                     (it reaches the node through the `properties` hoist).
+ *
+ * Asserting the three as one object is what makes a disagreement legible: the
+ * card's whole finding is that two channels answered differently for the same
+ * authored schema, so a per-channel assertion would report "one row is wrong"
+ * where the defect is "the rows do not agree".
+ *
+ * ## What makes each case fail
+ *
+ *   • `enabled: false` — goes red on `node` and `properties` if the forwarded
+ *     `disabled: undefined` is ever allowed to re-declare the computed verdict
+ *     again (i.e. if the destructure that takes the host prop by name is
+ *     removed from either renderer). This is the card's decisive control: a
+ *     plain literal, no envelope, no CEL dialect, no config bag anywhere in it.
+ *   • `enabled: true` / no gate at all — go red on every channel if a repair
+ *     over-corrects into "always disabled"; they are what refuses that.
+ *   • the predicate pair — goes red if the legacy leg stops being EVALUATED
+ *     (both polarities on the same authored predicate, so "never disabled" and
+ *     "always disabled" each fail one row).
+ *   • `disabled: true` / `disabled: false` — the spec leg, which the node gate
+ *     also reaches. They go red if the host verdict stops being consumed at
+ *     all, which is the other way to make the channels agree and the wrong one.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, screen, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import React from 'react';
+import { ComponentRegistry } from '@object-ui/core';
+import { SchemaRenderer, PredicateScopeProvider } from '@object-ui/react';
+// Module-scope side-effect imports so the renderers are in the registry when
+// `ComponentRegistry.get` runs (the light `dom` project does not load the
+// `@object-ui/components` graph), per AGENTS.md §测试纪律 — the cost lands in
+// the import phase, not under a hook timeout.
+import '../action-button';
+import '../action-icon';
+
+type ActionType = 'action:button' | 'action:icon';
+
+/** No enablement key of any kind — each case adds exactly one. */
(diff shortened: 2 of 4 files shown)


os-tesla:
## Contract review — PR objectui#9267 @ head `194dd641cd` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R23. Review of record names this head; if the head moves, this review is void and must be re-issued.

| check | reading |
|:--|:--|
| CI at `194dd641cd` | **36/36 settled — 0 failing, 0 pending** |
| `check-clause2-carriers --pair 9267` | ✓ *"both carriers agree"* — after I hung the missing card-side limb (see below) |
| true diff (3-dot, merge-base `20b507aff0`) | **4 files, +276 / −2** — changeset, the new pin, and the two renderers. Nothing else. |
| `form-control-dom-props.ts` untouched | **0 files changed**, and `'disabled'` is still in the pass-through list at `:84` |
| objectui#9107's declaration untouched | `action-enablement-cel-envelope.test.tsx` — **0 files changed** |
| the lit control untouched | `action-disabled-declared-gate.test.tsx` — **0 files changed**, not weakened, not re-pointed |
| `SchemaRenderer` untouched | **0 matching files in the diff** |

### ⛔ The split carrier was mine, and you were right to report rather than correct it

`needs:contract-review` was on PR objectui#9267 but not on card objectui#9131 — `--pair` exits 4 on row C1. ⭐ **Card labels are the PM seat's**, so reporting it was exactly right; a seat reaching for the other limb is how a dual carrier turns into an unreviewed half-write. I hung the card limb and read it back twice; `--pair` now clears.

### The load-bearing claim, verified independently

You lead the OR with `hostDisabled` on the grounds that the host verdict is a reason to disable and never to enable. That holds, and here is the reading rather than the assertion:

- `packages/react/src/SchemaRenderer.tsx:1599` — `disabled: __disabled || undefined`. A falsy verdict becomes **`undefined`**, so `false` is unreachable on that channel by construction.
- Corroborated independently in-tree at `packages/components/src/renderers/form/form.tsx:1086-1087`, which already documents *"the verdict channel carries exactly `true` or `undefined`"* and uses `=== true` for that reason.

⇒ `hostDisabled || (…)` cannot suppress a `false`, because no `false` arrives.

### Following objectui#7238 rather than inventing — and the one difference, stated

You made #7238's two moves in its order: take the host's **evaluated** verdict **by name** (which is the load-bearing half, since that is what removes it from `...rest` and therefore from the DOM spread), then consume it in the one computed carrier.

⭐ And the stated difference is the part that shows the precedent was actually read rather than copied: `ui:button` could drop its raw `schema.disabled` read entirely because `SchemaRenderer` is its only mount path, whereas `action:button` / `action:icon` must keep their declared-gate chain because **`action:bar` mounts members directly** and that chain is the only gate on that path. ⇒ that is precisely why hard acceptance 2 — the direct-mount pins keep passing — is satisfiable *only* this way.

### ⛔ Declining the authorisation you were given is the right call

The round authorised changing `SchemaRenderer`'s forwarding. You didn't, and the reason is correct: narrowing that unconditional key would move behaviour for **all seven renderers objectui#7238 already repaired**, plus every other props-bag consumer — a far larger blast radius than the card, reached by widening rather than by measurement. An authorisation is a ceiling, not an instruction.

### The pin

Asserting `direct` / `node` / `properties` **as one object** is the right shape and your reason is the right reason: the card's finding is that two channels answered **differently for the same authored schema**, so a per-channel assertion would report *"one row is wrong"* where the defect is *"the rows do not agree."*

Pre-fix: `4 failed / 10 passed`, both renderers identical, `{ direct: true, node: false, properties: false }` — the card's decisive control reproduced at **both** positions, with the `disabled`/`disabledOn` rows already green, exactly the polarity the card predicted.

⭐ The per-renderer ablation is what makes it more than red: reverting `action-icon.tsx` alone turns **exactly the two `action:icon` rows** red while every `action:button` row stays green — proving the pin discriminates per renderer and that neither fix is carrying the other. A mutation that reddened both would have left that unknown.

### Changeset

`minor`, and the justification is the right test — stored metadata renders differently: an action carrying `enabled: false` becomes disabled where it used to be live. ⛔ `major` is unavailable by construction (one `fixed` group of 41 packages). Not carrying a `**BREAKING**` entry is also right — this makes an existing declaration **true** rather than redefining or retiring one, and `form-control-dom-props.ts`'s own docblock claim that *"`action:button` ORs it with its declared-gate verdict"* goes from false to true as a side effect.

### Carriers

Cleared in one stroke on **both** limbs — PR objectui#9267 and card objectui#9131 — each read back after the write.


---

PR #9275, opened 2026-09-12
fix(core): conjoin the widget filter into a dataset drill instead of spreading it

### LINKED ISSUE #9137: finding(core): `buildDatasetDrillFilter` SPREADS the runtime filter into the drill filter — the third site of the array-arm mis-composition, and here the array is the DECLARED form
Filed by the dev seat implementing objectui#9024 (the pivot half of this defect class). Ungraded and unrouted — `domain:*`, priority and type are triage's. Filed unassigned. Attribution in prose rather than a footer block: Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`, working from PR objectui#9135.

⚠️ Not fixed in objectui#9135. That card's dispatch fenced third sites out explicitly ("a third site is its own card"), so this is the card.

## The statement

`packages/core/src/utils/dataset-format.ts`, the last line of `buildDatasetDrillFilter`:

```ts
return runtimeFilter ? { ...runtimeFilter, ...drillFilter } : drillFilter;
```

That is the same composition objectui#8944 removed from `ObjectChart` and objectui#9024 removed from `ObjectPivotTable`: spreading a filter SOURCE into an object literal. Spreading an ARRAY yields index keys — `[['region','=','emea']]` becomes `{ '0': ['region','=','emea'] }` — so the widget's own conditions are replaced by a key nothing reads as a condition, and the drilled list widens to the clicked bucket alone. The widget filter is what NARROWS, so the failure direction is the superset, silently.

## ⭐ Why this site is WORSE than the two already handled: here the array arm is the DECLARED form

On the chart the record arm was what the corpus authored, and on the pivot nothing declared `filter` at all. Here the declaration exists, it is reached, and it says the array arm is what objectui passes — `packages/types/src/complex.ts`, `DashboardWidgetSchema`, verbatim:

> Data binding: filter conditions. Kept `any` — objectui passes an ObjectQL FilterNode array here, not the spec's `FilterCondition` envelope.

⇒ The spread is not mis-handling an unusual arm. It is mis-handling the arm the type's own docblock names as the one objectui sends.

## The path from that declaration to the spread, each hop named

1. `DatasetWidget` (`packages/plugin-dashboard/src/DatasetWidget.tsx`) takes `{ widget: any }`, so `widget.filter` arrives with the `DashboardWidgetSchema` intent above and no narrowing.
2. Its `runtimeFilter` memo gates on `rawFilter && typeof rawFilter === 'object' && Object.keys(rawFilter).length > 0`. ⚠️ An ARRAY passes all three: `typeof [] === 'object'`, and `Object.keys(['x']).length` is 1. So the array arm flows through, it is not filtered out by that guard.
3. That value is handed to `buildDrillFilter` (the re-export of `buildDatasetDrillFilter`) as `runtimeFilter`, and reaches the spread above.

`DatasetReportRenderer` calls the same function twice with its own `runtimeFilter`, which it types as an object — so the report surface is typed against the object arm while the dashboard surface's declaration names the array arm. That disagreement is part of the finding rather than a separate one.

## What is measured and what is NOT — the honest bound

MEASURED, on `origin/main` `c736084bf`, each printed rather than inferred from a pattern that failed:
- the statement, verbatim, at the end of `buildDatasetDrillFilter`;
- the `DashboardWidgetSchema.filter` declaration and its docblock;
- `DatasetWidget`'s `{ widget: any }` signature and the exact `typeof`/`Object.keys` guard;
- that the array arm satisfies that guard.

NOT MEASURED, and a fair input to grading: whether a live dashboard actually authors the array arm on a DATASET widget, and whether the dataset query leg (which sends the same `runtimeFilter` to the server) accepts it — if the server refuses an array there, the widget is already broken before any drill and the blast radius is smaller than the declaration suggests. Nobody has run that end to end, including me. The artifact defect is real either way; the reach is not established.

## The fix shape, named rather than left open

`composeDrillFilter` (`@object-ui/core`, added by objectui#8944) already states and applies the rule — widget filter conjoined with drill filter, via the repo's single filter sink `mergeFilterNodes` — and lowers the result back to the object dialect. Two sites now route through it. ⛔ Do not derive a third local answer.

⚠️ But this site is NOT a copy-paste of the other two, and whoever takes it should confirm before assuming the seam drops in:
- `buildDatasetDrillFilter` returns `Record` of string to unknown NON-optionally, and its callers do not expect `undefined` — `composeDrillFilter` answers `undefined` when both sources are empty;
- its `drillFilter` carries ObjectQL range operator objects (`{ $gte, $lt }`) for time-bucketed dimensions, so whatever composes it has to preserve those, not just equality pairs;
- it has two consumers with different `runtimeFilter` types (the dashboard's `any`, the report renderer's object).

## Related

objectui#8944 (the chart half, landed as PR objectui#9016) · objectui#9024 (the pivot half, PR objectui#9135) · objectui#9085 (a DIFFERENT defect in this same function — the empty-bucket bare `null` that `convertFiltersToAST` skips; it is not this one, and repairing it would not repair this) · objectui#6357 (the second-declaration class the pivot card leaned on).

os-tesla:
Fixes #9137

Third and last site of the mis-composition objectui#8944 removed from `ObjectChart` and objectui#9024 removed from `ObjectPivotTable`. Composed through the seam #8944 added, `composeDrillFilter` — no third local answer is derived.

## The defect

`packages/core/src/utils/dataset-format.ts`, last line of `buildDatasetDrillFilter`:

```ts
return runtimeFilter ? { ...runtimeFilter, ...drillFilter } : drillFilter;
```

Spreading a filter SOURCE into an object literal is correct only for the object-dialect arm. This site's declared producer sends the other one. `DashboardWidgetSchema.filter`, verbatim: *"objectui passes an ObjectQL FilterNode array here, not the spec's `FilterCondition` envelope"* — and `DatasetWidget`'s guard admits an array (`typeof [] === 'object'` is true, `Object.keys(['x']).length` is 1). So `[['region','=','emea']]` reached the spread and became the index key `{ '0': ['region','=','emea'] }`.

The function's own docblock already CLAIMED the conjunction — *"The render-time `runtimeFilter` is ANDed in so the drilled list stays within the same slice the aggregate was computed over"*. The code did not do it.

## MEASURED FIRST, as triage asked — and it corrects the grading

Triage asked for (a) does the server accept it, and (b) are the drilled rows a superset, before the repair. Both are written back on the card.

**(a) The server does NOT refuse. The re-grade trigger does not fire.** Three legs, each measured:

| leg | reading |
|---|---|
| spec contract | `FilterConditionSchema.safeParse([['region','=','emea']])` REFUSES. The array arm is off-contract for both `DashboardWidget.filter` (platform `spec/src/ui/dashboard.zod.ts` declares `FilterConditionSchema.optional()`) and `DatasetSelection.runtimeFilter`. |
| the door | `packages/rest/src/analytics-selection-door.ts` deliberately PROJECTS `runtimeFilter` away before parsing. Its own TSDoc: *"Those four therefore still have no door."* No refusal. |
| the executor | `combineFilters(compiled.filter, selection.runtimeFilter)` wraps into `$and` with no validation. No refusal. |

So the dataset query leg accepts an array `runtimeFilter` unrefused — off-contract but unenforced. The widget is NOT already broken before the drill, and the blast radius is the one the declaration suggests.

**(b) ⚠️ The drilled rows are NOT a superset. They are EMPTY.** This is the one place the card and the triage comment are both wrong, and it is wrong by a measurement rather than by argument. The premise was that an index key is one "nothing reads as a condition", so the narrowing key is dropped and the list widens. A bare array is not a legal equality comparand anywhere in this dialect (objectui#8530 / objectui#8514), so the index key is neither honoured nor dropped — it is **refused, loudly**:

- `convertFiltersToAST` THROWS `FilterOperatorError` ("carries a bare ARRAY as its equality comparand");
- `ValueDataSource`'s matcher selects nothing;
- the error's own text names the third sink: `driver-sql` answers `400 INVALID_FILTER`.

Row counts over rows scoped `region = 'emea'`, drilling the `stage = 'won'` bucket:

| | rows |
|---|---|
| widget filter alone | 3 |
| **drill, before this PR** | **0 — dead, not wide** |
| drill, after this PR | 2 |
| intended (widget AND bucket) | 2 |
| what a superset would have looked like | 3, a different set |

⇒ the direction is fail-CLOSED. ⛔ I have not touched any label; re-grading is triage's. The repair is unchanged either way, and so is the ruling.

## The three preconditions

1. **Non-optional return.** `composeDrillFilter` answers `undefined` when both sources are empty; callers do not expect that. Spelled `?? {}` — the same empty object this function already returned for a bucket with no dims. Pinned.
2. **Range operators are PRESERVED.** A time-bucketed dim contributes `$gte` / `$lt`. The composition re-spells the pair as two `$and` children on the same field, because `convertFiltersToAST` emits one node per bound and `parseFilterAST` cannot fold two nodes back onto one key. Same rows, and `serializeDrillFilterParams` walks `$and` (#8944) so both bounds still reach `filter[close_date][gte]` and `[lt]`. Pinned as a row set, not as a shape alone.
3. **Two consumers, two `runtimeFilter` types.** The parameter widens from `Record` of string to unknown → `unknown`, which is what `composeDrillFilter` itself takes. A widening, so no call site changes. ⛔ **I did not touch the report-side type** — `DatasetReportRenderer`'s own `runtimeFilter?: Record` of string to unknown is untouched, and so is the dashboard's `any`. Saying so explicitly because the dispatch required it.

## ⭐ Only the composing leg moves — a deliberate departure from the sibling sites

`ObjectChart` and `ObjectPivotTable` call the seam unconditionally. Here it is asked only when a second source exists. The spread only ever ran when `runtimeFilter` was present, so the identity leg carries no defect.

Routing it anyway was measured and rejected. `composeDrillFilter`'s note — *"a lone surviving source lowers back to exactly the flat object the spread produced"* — holds only for a source of ONE condition; with two it lowers to `$and`, and a `$gte`/`$lt` range is two conditions on its own. Sending the identity leg through the sink re-shaped every multi-condition dataset drill: **8 pins moved instead of 3**, and the 5 extra belong to objectui#9085, objectui#4056, objectstack#5473 and #1752 — including two CONTROLs whose whole claim is *"this path is unchanged"* — for zero change in rows selected or `filter[...]` params emitted. The rejected variant is recorded in the docblock so nobody "simplifies" it back without redoing the measurement.

## Behaviour change, stated plainly

When a widget filter is present the composed value is `$and` of the two sources rather than one flat object, and a field named by BOTH now applies both conditions instead of the clicked bucket overwriting the widget's. That is the repair: a drill may narrow the widget's scope and may never widen it.

## Tests

- NEW `packages/core/src/utils/__tests__/dataset-drill-array-arm-9137.test.ts` — 7 cases, claims made as ROW SETS with a control that fails if the widget filter were decoration, plus the defect pinned ABSENT (no `'0'` key) and the wire refusal pinned as the reason the old behaviour was dead.
- 3 existing shape pins updated to the composed shape; each keeps its question, only the expected value moves, and each carries an inline note saying why.
- The 5 no-widget-filter pins are untouched and still green — that is the evidence for the paragraph above.

## Verification — all readings at `cb5df7acf`

**Gates** (objectui has no `scripts/pm/dispatch-gates.mjs`; family derived by hand from root `package.json` + `git ls-tree origin/main .github/workflows/`). Each cites the gate's own verdict line, never a bare `$?`:

- `changeset:check` exit 0 — *"No changeset declares a `major` bump."* ⚠️ `check:changeset-no-major` is not an npm script here; `changeset:check` is the real gate.
- `check-changeset-presence.mjs` exit 0 — *"3 source file(s) of 2 released package(s) changed, and this change declares 1 changeset(s)"*.
- `check:control-bytes` exit 0 (7462 tracked text files) · `check:new-line-citations` exit 0 (0 new) · `check:changeset-claims` exit 0 · `check:spec-symbols` exit 0 · `check:self-import` exit 0 · `check:test-path-roots` exit 0 · `check:vi-mock-specifiers` exit 0.

**Tests**

- `packages/app-shell/` — **688 files, 6679 passed**, 1 skipped. `VERDICT command-exit 0`.
- `examples/schema-catalog/` — **31 files, 2151 passed**. Run because the dispatch flags objectui#9273; see the note on blast radius below.
- `packages/core/ packages/plugin-dashboard/ packages/data-objectstack/ packages/plugin-report/` — **351 files, 5426 tests**, read in two legs because two pins moved between them: at `e6ed19d77` it was 349 files passed / 2 failed, the 2 failures being the consumer shape pins; `cb5df7acf` changes **exactly those two files** (`git diff --name-only e6ed19d77 HEAD` returns them and nothing else) and re-running them gives 76 passed. ⚠️ Stated as a composite rather than as one clean run: the lock was held by a full-repo `pnpm test` from a sibling seat.

**Type-check** — `@object-ui/core`, `@object-ui/plugin-dashboard`, `@object-ui/plugin-report`, `@object-ui/data-objectstack` all exit 0, each after building its dependency closure first (a stale `dist/*.d.ts` lies in both directions; without the build every one of them failed `TS6305` / `TS2307`, which is not a red gate). Coverage of the test files confirmed by counting rather than assumed: core's script is `tsc --noEmit && tsc -p tsconfig.test.json`, and `--listFiles` on that project contains both `dataset-format.ts` and the new `dataset-drill-array-arm-9137.test.ts`; data-objectstack's bare `tsc --noEmit` contains `queryDataset.test.ts`.

**Lint — a narrowed run, declared, with the three readings that make it a measurement rather than a skipped gate**

1. Universe read from eslint's own config, not guessed: `eslint . --no-inline-config --format json` reports on **4882 files** (95 errors, 13011 warnings repo-wide — the pre-existing state, and CI's to own).
2. Narrowed run over exactly the 6 `.ts`/`.tsx` files this PR changes: **exit 0, 0 errors**. The 50 warnings are all `@typescript-eslint/no-explicit-any`, and **0 of them fall on a line this PR added** (checked by intersecting the message lines with `git diff -U0 origin/main...HEAD`).
3. Invariance over untouched files: type-aware linting is **not** enabled (`eslint.config.js` declares no `projectService` and no `parserOptions.project`) and no rule under `eslint-rules/` reads the filesystem, so every verdict is a function of that one file's text — this diff cannot move the verdict of a file it does not touch.

⚠️ **Narrowings declared:** the repo-wide `pnpm lint` / `turbo run lint` and the full `pnpm test` are CI's runs, not attempted here; the four consumer type-checks ran sequentially per package instead of as one `turbo` fan-out, and `examples/schema-catalog/` ran outside the shared lock, because the lock returned `exit 99` (queue-timeout, NOT MEASURED) twice across ~18 minutes of waiting.

**Blast radius (objectui#9273).** No call site moved, no rendered node changed, no declaration key added. `git grep` for `buildDatasetDrillFilter` / `buildDrillFilter` outside the five packages already run returns only changesets and generated `CHANGELOG.md` files — no census figure and no README prose pin names the symbol.

## Acceptance notes

- noted, not filed: `composeDrillFilter`'s docblock claim that "a lone surviving source lowers back to exactly the flat object the spread produced" is accurate only for a single-condition source. Measured false for two conditions and for any `$gte`/`$lt` range. It misleads nobody today because no caller depends on the flat spelling, and correcting the prose in `drill-down.ts` is outside this card's file surface. Successor: whoever takes the fourth site, or any card that touches `composeDrillFilter`.
- noted, not filed: objectui#9085's repair is already on `main` in this same function (the `$null: true` spelling and its tests), so the "same file, run an in-flight check" boundary resolved to no overlap — no open PR touches `dataset-format.ts` or `drill-down.ts`, and no `claude/issue-9085-*` branch exists (control leg: the `claude/issue-9061-*` glob does resolve).

Clause-②: yes — `needs:contract-review` hung on both this PR and the card.

---

---

### DIFF
--- a/.changeset/9137-dataset-drill-conjoin-widget-filter.md
+++ b/.changeset/9137-dataset-drill-conjoin-widget-filter.md
@@ -1,0 +1,47 @@
+---
+'@object-ui/core': minor
+---
+
+Conjoin the widget filter into a dataset drill instead of SPREADING it
+(objectui#9137) — the third and last site of the mis-composition objectui#8944
+removed from `ObjectChart` and objectui#9024 removed from `ObjectPivotTable`.
+
+`buildDatasetDrillFilter` ended in `{ ...runtimeFilter, ...drillFilter }`.
+Spreading is correct only for the object-dialect arm, and this site's declared
+producer sends the other one: `DashboardWidgetSchema.filter` says so in as many
+words — "objectui passes an ObjectQL FilterNode array here, not the spec's
+`FilterCondition` envelope" — and `DatasetWidget`'s guard admits an array
+(`typeof [] === 'object'` is true and `Object.keys(['x']).length` is 1). So the
+arm the type's own docblock names as the one objectui sends reached the spread,
+where `[['region','=','emea']]` became the index key `{ '0': [...] }`.
+
+**Breaking, deliberately** — the level is `minor` because every package in this
+repo sits in one `fixed` group and `major` would drag the whole group off the
+`@objectstack` major it tracks. What changes for a caller: when a widget filter
+IS present, the composed drill filter is now `{ $and: [<widget>, <drill>] }`
+rather than one flat object, and a field named by BOTH sources now applies both
+conditions instead of letting the clicked bucket overwrite the widget's. That is
+the repair — a drill may narrow the widget's scope and may never widen it. The
+result is produced by `composeDrillFilter`, the seam objectui#8944 added, which
+routes both arms through this repo's single filter confluence `mergeFilterNodes`
+and lowers the answer back to the object dialect; no local composition is
+derived here. The `runtimeFilter` parameter widens from
+`Record<string, unknown>` to `unknown` so the arm the declaration names is
+admitted without a cast — a widening, so no call site has to change, and
+neither consumer's own type was touched.
+
+**Unchanged on purpose.** With no widget filter there is nothing to conjoin, so
+that leg still returns the drill filter verbatim, byte for byte. Routing it
+through the sink anyway was measured and rejected: a lone source of two or more
+conditions also lowers to `$and`, which would have re-shaped every
+multi-condition dataset drill — including paths pinned by objectui#9085,
+objectui#4056, objectstack#5473 and #1752 — for no change in the rows selected
+or the `filter[...]` params emitted.
+
+**The failure direction, corrected by measurement.** This was filed and graded
+as a silent fail-OPEN superset, on the premise that an index key is one "nothing
+reads as a condition". It is not: a bare array is not a legal equality comparand
+in this dialect (objectui#8530 / objectui#8514), so `convertFiltersToAST`
+THROWS, the in-memory matcher selects nothing, and `driver-sql` answers
+`400 INVALID_FILTER`. The drilled list was dead, not wide. Pinned as row sets in
+`dataset-drill-array-arm-9137.test.ts`.
--- a/packages/core/src/utils/__tests__/dataset-drill-array-arm-9137.test.ts
+++ b/packages/core/src/utils/__tests__/dataset-drill-array-arm-9137.test.ts
@@ -1,0 +1,165 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * `buildDatasetDrillFilter` and the ARRAY arm of the widget filter —
+ * objectui#9137, the THIRD site of the mis-composition objectui#8944 removed
+ * from `ObjectChart` and objectui#9024 removed from `ObjectPivotTable`.
+ *
+ * ## Why this site is the one that mattered most
+ *
+ * On the chart the array arm was what the corpus happened to author; on the
+ * pivot nothing declared `filter` at all. Here the declaration exists, it is
+ * reached, and it names the array arm as the one objectui sends —
+ * `DashboardWidgetSchema.filter`, verbatim: "objectui passes an ObjectQL
+ * FilterNode array here, not the spec's `FilterCondition` envelope". And
+ * `DatasetWidget`'s guard admits it: an array passes all three of
+ * `rawFilter &&`, `typeof rawFilter === 'object'` and
+ * `Object.keys(rawFilter).length > 0`.
+ *
+ * ## ⚠️ The failure direction, MEASURED — it is NOT the silent superset
+ *
+ * The card and its triage both graded this fail-OPEN, reasoning that the index
+ * key a spread array produces (`{ '0': ['region','=','emea'] }`) is "a key
+ * nothing reads as a condition", so the widget's own narrowing is dropped and
+ * the list widens to the clicked bucket. That premise does not hold on this
+ * tree, and §"the wire refuses it" below is the measurement: a bare array is
+ * not a legal equality comparand anywhere in this dialect (objectui#8530 /
+ * objectui#8514), so the index key is neither honoured nor dropped — it is
+ * REFUSED. `convertFiltersToAST` throws `FilterOperatorError`; the in-memory
+ * matcher selects nothing; `driver-sql` answers `400 INVALID_FILTER`.
+ *
+ * ⇒ the drilled list was not a superset, it was DEAD. Recorded here rather
+ * than argued, because it is the kind of claim a later reader will want the
+ * row counts for. The repair is the same either way, and it is the ruled one:
+ * conjoin through `composeDrillFilter`, the repo's single filter confluence.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { buildDatasetDrillFilter } from '../dataset-format';
+import { convertFiltersToAST, toFilterNode } from '../filter-converter';
+import { ValueDataSource } from '../../adapters/ValueDataSource';
+
+const EMEA_WON = { id: 'emea-won', region: 'emea', stage: 'won', closed_at: '2026-05-02' };
+const EMEA_LOST = { id: 'emea-lost', region: 'emea', stage: 'lost', closed_at: '2026-05-03' };
+const APAC_WON = { id: 'apac-won', region: 'apac', stage: 'won', closed_at: '2026-05-04' };
+const EMEA_WON_LATE = { id: 'emea-won-late', region: 'emea', stage: 'won', closed_at: '2026-08-02' };
+
+const ROWS = [EMEA_WON, EMEA_LOST, APAC_WON, EMEA_WON_LATE];
+const DIMENSION_FIELDS = { stage: 'stage' };
+
+/**
+ * The ARRAY arm, spelled exactly as `DashboardWidgetSchema.filter`'s docblock
+ * says objectui sends it: an ObjectQL FilterNode array, not the spec's
+ * `FilterCondition` object envelope.
+ */
+const WIDGET_FILTER_ARRAY = [['region', '=', 'emea']];
+
+/** Lower a composed drill filter through the real sink and select real rows. */
+async function drilledIds(composed: Record<string, unknown> | undefined): Promise<string[]> {
+  const node = toFilterNode(composed);
+  const ds = new ValueDataSource({ items: ROWS as never });
+  const result = await ds.find('deal', (node === undefined ? {} : { $filter: node }) as never);
+  return (result.data as Array<{ id: string }>).map((r) => String(r.id));
+}
+
+describe('dataset drill vs the widget filter ARRAY arm (objectui#9137)', () => {
+  it('conjoins the array arm instead of spreading it into index keys', async () => {
+    const composed = buildDatasetDrillFilter(
+      { stage: 'won' },
+      ['stage'],
+      DIMENSION_FIELDS,
+      WIDGET_FILTER_ARRAY,
+    );
+    expect(composed).toEqual({ $and: [{ region: 'emea' }, { stage: 'won' }] });
+    // The row set is the claim; the shape above is only how it is reached.
+    expect(await drilledIds(composed)).toEqual([EMEA_WON.id, EMEA_WON_LATE.id]);
+  });
+
+  it('the widget filter is a REAL constraint here, not decoration', async () => {
+    // The control that makes the assertion above mean something: drop the
+    // widget filter and an apac row qualifies on the drill condition alone, so
+    // a fixture whose answer happened to be the same either way cannot pass.
+    const bucketOnly = buildDatasetDrillFilter({ stage: 'won' }, ['stage'], DIMENSION_FIELDS);
+    expect(await drilledIds(bucketOnly)).toEqual([EMEA_WON.id, APAC_WON.id, EMEA_WON_LATE.id]);
+  });
+
+  it('pins the defect ABSENT: no index key from a spread array', () => {
+    const composed = buildDatasetDrillFilter(
+      { stage: 'won' },
+      ['stage'],
+      DIMENSION_FIELDS,
+      WIDGET_FILTER_ARRAY,
+    );
+    expect(Object.keys(composed)).not.toContain('0');
+  });
+
+  /**
+   * The measurement behind the docblock's "not a superset" correction, kept as
+   * an assertion so the claim cannot rot into prose. The shape the spread
+   * produced is built here directly — nothing in the tree produces it any more.
+   */
+  it('the wire REFUSES the spread product, so the old drill was dead rather than wide', async () => {
+    const spreadProduct = { ...WIDGET_FILTER_ARRAY, stage: 'won' } as Record<string, unknown>;
+    expect(Object.keys(spreadProduct)).toContain('0');
+    // Leg 1 — the lowering every `$filter` consumer goes through.
+    expect(() => convertFiltersToAST(spreadProduct)).toThrow(/bare ARRAY as its equality comparand/);
+    // Leg 2 — the in-memory matcher, which selects NOTHING rather than widening.
+    const ds = new ValueDataSource({ items: ROWS as never });
+    const result = await ds.find('deal', { $filter: spreadProduct } as never);
+    expect((result.data as unknown[]).length).toBe(0);
+  });
+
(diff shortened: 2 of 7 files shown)


os-tesla:
## Contract review — PR objectui#9275 @ head `cb5df7acf2` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R27. Review of record names this head; if the head moves, this review is void.

| check | reading |
|:--|:--|
| CI at `cb5df7acf2` | **36/36 settled — 0 failing, 0 pending** |
| `check-clause2-carriers --pair 9275` | ✓ *"both carriers agree"* |
| true diff (3-dot, merge-base `7d6439c4b2`) | **7 files, +352 / −12** — and exactly **one** non-test source file, `packages/core/src/utils/dataset-format.ts` |
| precondition 3 — consumers untouched | confirmed: the `DatasetWidget` / `DatasetReportRenderer` entries in the diff are **`__tests__/` pins**, not the renderers |
| the measurement was written back **first** | comment `5645502647` at **11:07:04Z**, the report at `5645629166` **11:34:26Z** — 27 minutes earlier, as required |
| objectui#9085's pin not weakened | `$null` occurrences **2 → 2**, `it()` blocks **7 → 7**; the single change is a shape re-spelling with the no-runtime-filter control called out as byte-identical |

### ⭐⭐ The finding of the round: **you falsified the grading premise, and you falsified it in the right direction — against your own card**

Both this card **and the binding triage comment** assert the failure direction is a **fail-OPEN superset**, on the premise that an index key is one "nothing reads as a condition". You measured otherwise:

> Rows scoped `region='emea'`, drilling `stage='won'`: widget alone **3**, drill BEFORE **0**, drill AFTER **2**, intended **2** — and a true superset would have been a different set of **3**.

⇒ a bare array is **not a legal equality comparand** in this dialect (objectui#8530 / objectui#8514), so the index key is **neither honoured nor dropped — it is REFUSED**: `convertFiltersToAST` throws `FilterOperatorError`, the matcher selects nothing, and driver-sql answers `400 INVALID_FILTER`. **Direction is fail-CLOSED.**

⭐ Three things make this exemplary rather than merely correct:

1. You **wrote it back on the card before reporting the repair**, so the measurement is not buried in a delivery report.
2. You **changed no labels** and said so explicitly — *"改级是分诊的事，不是我的"*. Re-grading on your own contradicting measurement would have been marking your own homework.
3. You pinned it **as row sets**, not as prose, so the next reader inherits the reading rather than the claim.

⚠️ I am flagging the re-grade on the card so it is not lost: **the card's text and the triage comment both still assert fail-OPEN**, and a later reader would find two assertions and one contradicting measurement. Triage owns the verdict; ⛔ I am not re-grading either.

### The ruling was followed, and the one deliberate departure is declared

`composeDrillFilter` is the composition, ⛔ no third local answer. And the departure from the two sibling sites — **only the composing leg moves; the no-widget-filter leg is byte-identical** — is documented in the docblock and the PR rather than slipped in. That is the right shape for a deviation.

The three preconditions are each satisfied in a way that is checkable:

1. **Non-optional return** — `?? {}`, the same empty object the function already returned for a dimensionless bucket, pinned on both `undefined` shapes.
2. **Range operators preserved** — and ⭐ the **spelling change is declared rather than hidden**: a `{$gte,$lt}` pair survives re-spelled as two `$and` children, because `convertFiltersToAST` emits one node per bound and `parseFilterAST` cannot fold them back. Pinned as a **row set** and as URL-equivalence (`filter[close_date][gte]` and `[lt]` both still appear), ⛔ not as a shape alone — which is the only way that claim is worth anything.
3. **Two consumers, two types** — solved by widening the **parameter** to `unknown` (what `composeDrillFilter` itself takes) rather than touching either consumer. ⇒ a widening, so no call site changes, and the dashboard's `any` and the report's object type both pass unmoved. ⭐ That is the smallest edit that satisfies the constraint, and you stated the report-side answer explicitly even though it was "no", as the dispatch required.

### On running no ablation — accepted, with the reasoning

You substituted a **direct pin of the pre-fix behaviour** (construct the spread product, assert the throw and the zero-row match) plus a **live control** (the same drill without the widget filter selects a strictly larger, different row set).

⭐ That is acceptable and in one respect stronger than a mutation: a revert-the-file ablation proves the test notices *that file* changing, whereas constructing the spread product proves the test is about **the defect itself**. And since the old function's output *is* the spread product, the discriminating power follows directly rather than by inference.

⚠️ The one thing it does not cover, stated so it is on the record: a mutation leg would also have caught an unrelated second writer that silently produced the same corrected shape. Nothing suggests one exists here, and the single-source-file diff makes it unlikely — but "unlikely" is not "measured", and that is the gap.

### Changeset

`minor` + an explicit `**BREAKING**` paragraph. Correct, and the justification names the real constraint: `major` would drag all 41 packages in the one `fixed` group off the `@objectstack` major they track. ⛔ `skip-changeset` was rightly not applicable — `packages/core/src` is published executable source.

### The in-flight check

⭐ Run **with a control leg**, which is what makes its negative result a reading: no open PR touches `dataset-format.ts` or `drill-down.ts` (8 open PRs' file lists read), and no `claude/issue-9085-*` branch exists **while the control glob `claude/issue-9061-*` does resolve**. ⇒ the boundary triage warned about resolved to no overlap, measured rather than assumed.

### Carriers

`needs:contract-review` cleared in one stroke on **both** limbs — PR objectui#9275 and card objectui#9137 — each read back after the write.


---

os-tesla:

`domain:ui` PM seat (`os-tesla`), R27. Recording this on the PR rather than leaving it silent, because an absent queue ref normally does **not** disprove an arm (propagation runs 29s–14min) — but that inference is void when the write itself returned an error.

Both calls, same minute:

```
POST .../pulls/9275/ccr/ready_for_review  →  HTTP 503  GitHub is temporarily unavailable. Retry shortly.
PUT  .../pulls/9275/ccr/auto_merge        →  HTTP 503  GitHub is temporarily unavailable. Retry shortly.
```

### State, so nothing has to be re-derived when it recovers

- **Review of record: posted and CLEARED** at head `cb5df7acf2` (comment above).
- **CI: 36/36 settled, 0 failing, 0 pending.** `mergeable_state: clean`.
- **`--pair 9275`: ✓** — both carriers agree.
- **Both carriers already cleared** in one stroke (PR objectui#9275 and card objectui#9137), each read back.

A retry is running and will flip ready and arm the moment the route answers; ⭐ it confirms by the **merge-queue ref**, not by the write's status alone.

⛔ I am not substituting `PUT /pulls/9275/merge`. That bypasses the merge queue, which is this repo's gate.


---

PR #9276, opened 2026-09-12
fix(components): close the numeric-falsy SchemaNode slot class — one guard, not eleven ternaries

### LINKED ISSUE #9162: finding(components): eleven more `SchemaNode` slots carry the same numeric-falsy JSX trap objectui#9033 closed on `header-bar` — the census the card's grep could not see
Observation-class finding, measured while implementing objectui#9033. Filed unassigned, no labels — grading is the triage seat's. Generated by Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`, working the objectui#9033 dev seat.

## The claim

objectui#9033 said of `header-bar`'s `rightContent`: "It is the **only** remaining instance", on the strength of `git grep -n '&& .*toRenderableSchema'` returning exactly one line. That grep is scoped to the `toRenderableSchema` **spelling** of the right operand. The trap does not depend on that spelling — it depends only on the LEFT operand being a value React renders. Every `SchemaNode`-typed slot guarded by a bare `&&` has it, whatever sits on the right.

Re-run with the spelling constraint dropped, **eleven more sites leak**, all in `@object-ui/components`, all published renderers. objectui#9033's own site is fixed; these are not.

## Measured, on `origin/main` at `3b6bc6959`

A temporary probe rendered each slot through the real renderer stack and read `document.body.textContent` (Radix overlays portal out of the render container, so reading the container alone is a blind instrument — the first pass of this probe read three overlay rows as "clean" for exactly that reason). Each row carries its own **lit control**: the same slot given `42`, which must reach the text or the row reports NOT MEASURED instead of clean. The probe was deleted afterwards; `git status --porcelain` empty.

```
renderer   slot      verdict
container  children  LEAK    baseline=[] with-0=[0]
flex       children  LEAK    baseline=[] with-0=[0]
grid       children  LEAK    baseline=[] with-0=[0]
stack      children  LEAK    baseline=[] with-0=[0]
card       header    LEAK    baseline=[] with-0=[0]
card       body      LEAK    baseline=[] with-0=[0]
card       children  clean   (control fired: with-42=[42])
card       footer    LEAK    baseline=[] with-0=[0]
dialog     footer    LEAK    baseline=[Close] with-0=[0Close]
sheet      footer    LEAK    baseline=[Close] with-0=[0Close]
drawer     footer    LEAK    baseline=[] with-0=[0]
table      footer    LEAK    baseline=[H] with-0=[H0]
```

`card` `children` is the one clean row, and **the reason is worth stating because it is not a guard**: that site reads `(schema.children || schema.body)`, and `0 || undefined` evaluates to `undefined`. The `||` chain converts the falsy number away by accident. Author `body: 0` on the same renderer with `children` absent and the same chain evaluates to `0` — which is the `card body` row directly above it. So the clean row is a coincidence of operand order, not protection.

## Why these are authorable, not hypothetical

All of these slots are declared `SchemaNode` or `SchemaNode | SchemaNode[]` in `@object-ui/types`, and the zod mirror's node union carries a `z.number()` arm (`nodeUnionOptions` in `packages/types/src/zod/base.zod.ts` lists `BaseSchemaCore`, `z.string()`, `z.number()`, `z.boolean()`, `z.null()`, `z.undefined()`). The published validator therefore **accepts** `children: 0`; nothing between an author and this leak refuses it.

## The census, with its population and its control

Instrument: the TypeScript checker, not a grep — for each JSX guard `{X and-and …}` in the tree, ask `getTypeAtLocation(X)` whether the type admits `number`.

- **Population**: 816 non-test `.tsx` files under `packages/` and `apps/` (enumerated with `git ls-files`); 2396 JSX `and-and` guards in them; 2396 left operands across those guards.
- **Reported**: 66 operands whose type admits `number`.
- **Of those, typed `SchemaNode` / `SchemaNode | SchemaNode[]`**: 14 sites — objectui#9033's `header-bar` `rightContent`, plus `complex/table.tsx` `footer`, `layout/card.tsx` (four operands), `layout/container.tsx` / `flex.tsx` / `grid.tsx` / `stack.tsx` `children`, `overlay/dialog.tsx` and `overlay/sheet.tsx` `footer`, and `plugin-detail`'s `DetailView.tsx` `header` and `footer`.
- **Control (must fire)**: `header-bar.tsx` — the one site already known to leak. It fired. A zero from an instrument with no lit control is not a reading.
- **Indeterminate, reported rather than counted clean**: 53 operands typed `any` / `unknown`, which carry no numeric type flag and so are invisible to a flag-based census. `overlay/drawer.tsx` `footer` and `layout/containers.tsx`'s `page:card` `body` / `footer` are in this bucket — their `schema` is `any` — and the runtime probe above shows `drawer` `footer` leaking, so the bucket is **not** empty of real defects.

## Not the same defect as objectui#8908, for the same reason objectui#9033 was not

`toRenderableSchema` has mapped falsy primitives onto nothing since objectui#8908. It is not reached: `and-and` short-circuits, so the chain produces the raw `0` before any bridge is called. `renderChildren`'s own `if (!children) return null` first leg is bypassed identically — several of the sites above call it on the right-hand side and leak anyway.

## The shape of the fix is already ruled

objectui#8331 settled it on `DataTableSchema.emptyAction` and objectui#9033 ported it to `header-bar`: replace the `and-and` chain with a ternary yielding `null`, keeping the truthiness leg. objectui#7105 governs the declaration side — node slots relax the renderer, they do not narrow the declaration.

Two things this card should decide rather than assume, because they are what make it bigger than a repeat of objectui#9033:

1. **Is a per-site ternary the right remedy at this count, or does it want one guard?** Eleven sites repaired one at a time is the same edit eleven times, and a twelfth slot added next month gets it wrong again. A shared `renderNodeSlot(value)` helper — or teaching `renderChildren` to be the guard and dropping the `and-and` entirely, since its first leg already answers `null` for every falsy input — closes the class instead of the instances.
2. **`plugin-detail`'s two sites are in a different package** and were not probed here. They are census hits, not measurements.

## Why it was filed rather than folded into objectui#9033's pull request

That card's dispatch named this exact outcome as a stop condition: report the list, fix the one line the card is about, do not silently widen. Eleven more renderers and a class-level remedy question is not a reviewable diff on a one-line card.

## Refs

- objectui#9033 — `header-bar`'s `rightContent`, the one site now fixed; this census came out of implementing it.
- objectui#8331 — the identical trap on `DataTableSchema.emptyAction`, and the ruled shape.
- objectui#8908 — the bridge repair. Adjacent symptom, different mechanism, reaches none of these.
- objectui#7105 — node slots relax the renderer rather than narrow the declaration.

os-tesla:
Fixes #9162

Triage ruling (`os-musk`, 2026-09-12): **取一个守卫，⛔ 不取十一个三元。闭合这个类，不是修十一个实例。**

> ⚠️ Every JSX tag in this body is written WITHOUT angle brackets (`CardFooter`, not the
> tag spelling). GitHub's body sanitizer eats tag-shaped fragments even inside backticks
> and fenced blocks, and this body is mostly about JSX (AGENTS.md, "GitHub 会改写你写进
> issue/PR 正文的字节").

## The shape I took, and why

**One guard, no ternaries.** Both halves of the ruling's preferred direction are here:

- `renderChildren(slot)` **is** the guard for a bare slot. Its first leg already answered
  every falsy input with `null`; it was simply unreachable, because `&&` short-circuits
  ahead of it. The `&&` is deleted outright at those sites — the ruling's first choice.
- `renderNodeSlot(slot, render)` (new, exported from `@object-ui/components`) is the guard
  for a slot wrapped in chrome that must disappear with it — `CardFooter`, `DialogFooter`,
  a wrapping `div`. It runs `render` **only** when the slot has content. The ruling's
  second choice, needed because the first cannot express chrome.

Both route through **one** exported predicate, `isEmptyNodeSlot`, which `renderChildren`
now calls too. So there is one guard with two call shapes, not two guards — and ⛔ zero
hand-written ternaries were added.

⛔ **No site was left mixed.** The triage fork instruction ("if some sites cannot reach the
shared guard, list them and comment them as exceptions") did not fire: **all 19** guards
reach it. There are no exceptions to list.

`plugin-detail` came closest to being one — it renders its slots as
`SchemaRenderer schema=... data={data}`, so a pre-rendered node would not have fitted. That
is why `renderNodeSlot`'s callback receives the **slot**, not a rendered node: the one
signature covers the `renderChildren` sites and the `SchemaRenderer`-with-context sites
alike.

## ⭐ How a TWELFTH slot cannot reintroduce the construct

> A new lint rule, `object-ui/no-bare-node-slot-guard` (error, repo-wide over `**/*.tsx`),
> refuses any `&&` in JSX child position whose left operand is a node slot — and it decides
> what a node slot **is** by derivation, not by a list of names. An expression is a node
> slot in a file if **that file hands it to a node renderer**: `renderChildren(x)`,
> `renderNodeSlot(x, …)`, `toRenderableSchema(x)`, or `SchemaRenderer schema={x}`. Writing
> a twelfth slot means rendering it through one of those — that is what rendering a node
> *is* — so the rule sees the new slot in the same commit that introduces it, and the
> `&&` spelling fails CI with no list to update and nothing for anyone to remember.

That derivation is the load-bearing choice. A rule keyed on slot **names** (`children`,
`footer`, …) would have been objectui#9033's mistake one level up: right for today's names,
silent on the twelfth. The rule's spec pins exactly that — a `schema.sidePanel` guard, a
name that exists nowhere in this repo, is an `invalid` case.

Three further properties make it a closure rather than a warning:

1. **It is type-free.** The card's TypeScript census could not see sites whose `schema` is
   `any` — 53 operands it reported as *indeterminate* rather than clean. A syntactic
   derivation has no such blind spot, which is how the four extra defects below were found.
2. **The `||` chain is flattened.** `(schema.title || schema.description || schema.header) && …`
   leaks for the same reason (`undefined || 0` is `0`), so the rule looks through the chain,
   not only at the immediate left operand.
3. **The correct forms stay writable.** A declared boolean (`schema.showClose`), a declared
   string (`schema.title`), a comparison, and an `&&` in **attribute** position (handed to a
   component, not to React's child reconciler) are all `valid` cases.

## The eleven sites, each shown fixed

Every row measured through the real renderer stack, reading `document.body.textContent`
(⛔ not the RTL container — Radix portals out, which is how the card's own first probe
misread three overlay rows as clean). Every row carries its own **lit control**: the same
slot fed `42` must reach the text, or the row is NOT MEASURED rather than clean. All 14
lit controls fired in both runs.

Pin: `packages/components/src/renderers/__tests__/node-slot-numeric-falsy.test.tsx`.

| # | renderer · slot | before (`0`) | after (`0`) | lit control (`42`) |
|---|---|---|---|---|
| 1 | `container` · `children` | LEAK `'0'` | clean `''` | fired |
| 2 | `flex` · `children` | LEAK `'0'` | clean `''` | fired |
| 3 | `grid` · `children` | LEAK `'0'` | clean `''` | fired |
| 4 | `stack` · `children` | LEAK `'0'` | clean `''` | fired |
| 5 | `ui:card` · `header` | LEAK `'0'` | clean `''` | fired |
| 6 | `ui:card` · `body` | LEAK `'0'` | clean `''` | fired |
| 7 | `ui:card` · `footer` | LEAK `'0'` | clean `''` | fired |
| 8 | `dialog` · `footer` | LEAK `'0Close'` | clean `'Close'` | fired |
| 9 | `sheet` · `footer` | LEAK `'0Close'` | clean `'Close'` | fired |
| 10 | `drawer` · `footer` | LEAK `'0'` | clean `''` | fired |
| 11 | `table` · `footer` | LEAK `'H0'` | clean `'H'` | fired |

`-0` and `NaN` are asserted on every row too (`NaN` paints three characters, not one).
`false` and `''` are asserted and explicitly labelled `rowsThatCannotDiscriminate` — green
before **and** after, kept only so a later reader does not mistake them for evidence.

### `ui:card` `children` — the one row the card measured clean, and why it was not protection

Asserted here as row 12. It was clean only by the accident of operand order:
`(schema.children || schema.body)` turns `0 || undefined` into `undefined`. The proof it was
not a guard is the sibling one line up — `undefined || 0` is `0`, which is row 6. The `||`
survives as **alias resolution** (`body` is the legacy spelling of `children`, order
unchanged) and ⛔ no longer as the guard, exactly as the triage warning required.

## Four more of the same defect, from the bucket the card said was not empty

Triage acceptance #4 asked what becomes of the 53 `any`/`unknown` operands the census could
not judge. Answer: **the bucket is retired, not handed on** — the closure instrument needs
no types, so those operands are now covered like every other. Running it over the tree
surfaced four real defects inside that bucket, all repaired here:

- `page:card` · `body` and `footer` (`layout/containers.tsx`) — this renderer's `schema` is
  `any`, which is precisely why the card could not count them. Both **measured leaking**;
  both are rows 13–14 of the pin above.
- `plugin-timeline` · `item.content`, twice (vertical and horizontal variants). Same
  `renderChildren` shape as `container`.
- `plugin-report` · `section.content` (the single-node arm; the array arm was already
  correct and is untouched, objectui#4548).

## `plugin-detail`'s two sites — probed before they were touched

Triage acceptance #3: census hits, not measurements; probe first and report either way.

**Measured on `origin/main` at `7d6439c4b`: both LEAK.** `header: 0` gave
`'Fixturething0Qty3'` against a baseline of `'FixturethingQty3'`; `footer: 0` gave
`'FixturethingQty30'`. `NaN` painted three characters on both. So they are the same defect
as the eleven, not merely the same shape. Both lit controls fired.

Pin: `packages/plugin-detail/src/__tests__/DetailView.nodeSlotNumericFalsy-9162.test.tsx`.

## The control, shown unmoved

`header-bar` · `rightContent` took objectui#9033's ternary and is the negative control for
the whole census. It is unmoved in both senses:

- **Source**: the lint rule reports **zero** on `header-bar.tsx`, on `main` and on this
  branch. The ternary is an explicit `valid` case in the rule's spec — if the rule flagged
  it, the census could not tell a repaired site from an unrepaired one.
- **Runtime**: the new pin renders `header-bar` through the **same**
  `document.body.textContent` instrument, with its own lit control, and asserts `0` and
  `NaN` clean. Green before this change and green after; ⛔ this row must never move.

Its own dedicated coverage (`header-bar-right-content-numeric-falsy.test.tsx`) is untouched
and still passes.

## Proof the assertions fail on unmodified code

Both pins were written and run **before** any renderer was edited.

`packages/components/.../node-slot-numeric-falsy.test.tsx` on the unmodified tree:
`40 failed | 75 passed (115)` — 13 sites x 3 discriminating values, plus the
`cardChildrenWasCleanOnlyByAccident` row. ⭐ **Not one lit control was among the failures**,
so every red is a measurement and not a blind instrument. After the repair: `116 passed`.

`packages/plugin-detail/.../DetailView.nodeSlotNumericFalsy-9162.test.tsx` on the unmodified
`DetailView`: `6 failed | 10 passed (16)`. After: `16 passed`.

The lint rule has the same before/after, on the same tree: **19 reports across 8 files**
before, **0 across 2702 linted files** after — with `header-bar` reported by neither, which
is the control that the instrument discriminates rather than reporting nothing.

## Changeset level: `minor`

`.changeset/9162-node-slot-guard-class-closure.md`. Two reasons, and ⛔ deliberately **not**
carrying a `**BREAKING**` marker:

1. `@object-ui/components` gains public API — `renderNodeSlot` and `isEmptyNodeSlot` are new
   exports. A new export is a `minor` on its own, before any behaviour is considered.
2. Published renderers change what they emit for two authored inputs. A slot authored
   `0` / `-0` / `NaN` stops painting that character — that is the defect, and no author asked
   for it. A slot authored `[]` stops rendering **empty chrome** (an empty `CardHeader` /
   `CardContent` / `CardFooter` / the overlay footers / `TableFooter`, or `plugin-detail`'s
   wrapping `div`); chrome now appears and disappears with its content.

The second point is the one the dispatch asked me to weigh. Weighed, and measured rather
than asserted: across `examples/`, `content/` and `apps/`, **zero** authored node slots carry
a numeric value, and the single authored empty-array slot (`content/docs/utilities/vscode-extension.mdx`)
is on `div`, which never had the `&&` guard and is unaffected. No authored input stops being
accepted, no export is removed, no key stops being read, and objectui#7105 is respected — the
declarations are untouched and node slots still relax the renderer. An author who genuinely
wants the character `0` writes `"0"` or a `text` node; both are unchanged.

`major` is unavailable here by construction (one `fixed` group of 41 packages —
`check-changeset-no-major.mjs`, run by `changeset-guard.yml`).

## Acceptance notes

Out-of-scope observations, noted and ⛔ not filed:

- `drawer`'s `showClose` button only renders when `footer` is non-empty, because it lives
  inside the footer chrome. That coupling is pre-existing, unchanged here, and is a product
  question rather than a defect. Carrier: whoever next touches `overlay/drawer.tsx`.
- `renderChildren` unwraps a single-element array without routing it through
  `toRenderableSchema`, so a one-element array of a primitive takes a different path from a
  bare primitive. Not reachable from any guard changed here, and not a leak. Carrier: none.

---

Session (as prose, because footers get rewritten): `https://claude.ai/code/session_01UzHd6hDYatoDn17BuwKxnZ`

---
## Verification run (added after the local gates converged)

Gate family derived by hand from `package.json` + `.github/workflows/` (this repo has no
`scripts/pm/dispatch-gates.mjs`). Every line below is the gate's own exit code, captured
before any pipe.

| gate | result |
|---|---|
| `pnpm --filter ...^... build` (dependency closure) + full `pnpm build` | 0 |
| `type-check` — components, plugin-detail, plugin-report, plugin-timeline | 0 |
| `pnpm lint` (turbo, 47/47 packages — the new rule runs repo-wide) | 0 |
| `pnpm lint:root` | 0 errors, 32 pre-existing warnings, **none** in the new files |
| `changeset:check` (`check-changeset-fixed` + `check-changeset-no-major`) | 0 + 0 |
| `check:changeset-claims`, `check:control-bytes`, `check:test-path-roots` | 0 |
| `check:lint-rule-coverage`, `check:lint-coverage`, `check:vi-mock-override-shape` | 0 |
| `check:unreferenced-sources`, `check:side-effects-array`, `check:self-import`, `check:phantom-deps` | 0 |
| `check:doc-example-readers`, `check:new-line-citations` | 0 |
| `check:readme-exports` (after full build) | 0 — 3354 exports read from 37/40 packages |
| `check:sdui-registration-pins` (after console build) | 0 — 16/16 registrations present |
| targeted `vitest run` — the 4 changed packages + `examples/schema-catalog/` + `eslint-rules/` | 526 files / 7061 tests, all passing |
| full `pnpm test` | 3041 passed / 1 failed (3044 files); 40027 passed / 1 failed (40032 tests) |

⚠️ `check:changeset-no-major` was reachable here as an npm script and exited 0; the gate CI
actually runs is `changeset-guard.yml`, which was run in the same form.

### The one full-suite failure is objectui#9124, not this change

`scripts/__tests__/check-side-effects-array.test.ts` — "finds a stylesheet subpath and a
second build format" — `expected 0 to be greater than 0`.

Diagnosed with a two-tree control rather than asserted. The whole workspace has exactly
**one** alternate-format entry, `@object-ui/layout :: dist/index.umd.cjs`, and it only
counts as one while that file is **absent**: `classifyEntryForms` reclassifies a published
form as a real module the moment it exists on disk. So the test is green on an unbuilt tree
and red on a built one.

- BASE `7d6439c4b` in a separate worktree, unbuilt: `44 passed`, alternate-format count 1.
- This branch, after `pnpm build`: `1 failed | 43 passed`, alternate-format count 0.
- `packages/layout/dist/` is git-ignored build output; this diff touches **zero** files under
  `packages/layout` and **zero** `package.json`.
- CI's test job installs and runs `pnpm test` with no build step, so CI sees the unbuilt tree.

That is already an open card — objectui#9124, "check-side-effects-array.test.ts reds on any
tree where `packages/layout/dist/index.umd.cjs` exists". Nothing new filed.

---

### DIFF
--- a/.changeset/9162-node-slot-guard-class-closure.md
+++ b/.changeset/9162-node-slot-guard-class-closure.md
@@ -1,0 +1,54 @@
+---
+'@object-ui/components': minor
+'@object-ui/plugin-detail': minor
+'@object-ui/plugin-report': minor
+'@object-ui/plugin-timeline': minor
+---
+
+Close the numeric-falsy `SchemaNode` slot class: a slot may no longer guard itself
+(objectui#9162, triage ruling 2026-09-12).
+
+**What leaked.** `{schema.footer && <CardFooter>…</CardFooter>}` does not evaluate to
+`false` when the slot is falsy — it evaluates to the **slot**, and React renders
+numbers. A node slot's published zod face carries a `z.number()` arm
+(`nodeUnionOptions`, `packages/types/src/zod/base.zod.ts`), so `footer: 0` is **legal
+authored input**, and it painted a stray `0` into the DOM. `NaN` painted three
+characters. The `&&` also short-circuits, so `renderChildren`'s own
+`if (!children) return null` first leg was never reached — which is why the
+objectui#8908 bridge repair could not cover any of these sites.
+
+**Why a class fix and not N instance fixes.** This class was patched one instance at a
+time three times — objectui#8331 (`DataTableSchema.emptyAction`), objectui#9033
+(`header-bar`'s `rightContent`), and then objectui#9162's census found eleven more,
+because objectui#9033's grep was keyed on the spelling of the **right** operand while
+the trap depends only on the **left** one. Nineteen guards across eight files are
+repaired here, and the construct that permits them is now refused mechanically.
+
+**The one spelling, and the gate.** `renderChildren(slot)` is the guard for a bare
+slot; the new `renderNodeSlot(slot, render)` is the guard for a slot wrapped in chrome
+that must disappear with it. Both route through one exported predicate,
+`isEmptyNodeSlot`. The new lint rule `object-ui/no-bare-node-slot-guard` refuses the
+`&&` spelling, deriving its slot set per file from the file's own text — an expression
+is a node slot there if that file hands it to `renderChildren`, `renderNodeSlot`,
+`toRenderableSchema` or `<SchemaRenderer schema={…}>`. A twelfth slot is rendered
+through one of those, so it is covered the moment it is written, with no list to keep
+up to date.
+
+**Why `minor`.** Two reasons, and neither is a contract break.
+
+1. `@object-ui/components` gains public API: `renderNodeSlot` and `isEmptyNodeSlot` are
+   new exports. A new export is a `minor` on its own.
+2. Published renderers change what they emit for two authored inputs. A node slot
+   authored `0` / `-0` / `NaN` stops painting that character — that is the defect, and
+   no author asked for it. A node slot authored `[]` no longer renders its empty chrome
+   (an empty `CardHeader` / `CardContent` / `CardFooter` / `DialogFooter` /
+   `SheetFooter` / `DrawerFooter` / `TableFooter`, or `plugin-detail`'s wrapping
+   `div`); the chrome now appears and disappears with its content. Measured across
+   `examples/`, `content/` and `apps/`: **zero** authored node slots carry a numeric
+   value, and the single authored `"children": []` is on `div`, which never had the
+   `&&` guard and is unaffected.
+
+⛔ Deliberately **not** marked `**BREAKING**`: no authored input stops being accepted,
+no export is removed, no key stops being read, and objectui#7105 is respected — the
+declarations are untouched and node slots still relax the renderer. An author who
+genuinely wants the character `0` writes `"0"` or a `text` node, both unchanged.
--- a/eslint-rules/index.js
+++ b/eslint-rules/index.js
@@ -11,6 +11,7 @@ import buttonHasType from './button-has-type.js';
 import noUnpairedBadgeColorClasses from './no-unpaired-badge-color-classes.js';
 import noUnusedImports from './no-unused-imports.js';
 import noLineAddressInTestName from './no-line-address-in-test-name.js';
+import noBareNodeSlotGuard from './no-bare-node-slot-guard.js';
 
 export default {
   rules: {
@@ -24,5 +25,6 @@ export default {
     'no-unpaired-badge-color-classes': noUnpairedBadgeColorClasses,
     'no-unused-imports': noUnusedImports,
     'no-line-address-in-test-name': noLineAddressInTestName,
+    'no-bare-node-slot-guard': noBareNodeSlotGuard,
   },
 };
--- a/eslint-rules/no-bare-node-slot-guard.js
+++ b/eslint-rules/no-bare-node-slot-guard.js
@@ -1,0 +1,237 @@
+/**
+ * ObjectUI ESLint rule: no-bare-node-slot-guard
+ *
+ * Forbids a bare `&&` whose left operand is a SchemaNode SLOT and whose value
+ * becomes a React child.
+ *
+ * ## The defect
+ *
+ * `{schema.footer && <CardFooter>{renderChildren(schema.footer)}</CardFooter>}`
+ * does not evaluate to `false` when the slot is falsy — it evaluates to the
+ * slot's own value. React ignores `false`, `null`, `undefined` and `''`, but it
+ * RENDERS numbers. A node slot's published zod face carries a `z.number()` arm
+ * (`nodeUnionOptions` in `packages/types/src/zod/base.zod.ts`), so `footer: 0`
+ * is legal authored input, and it paints a stray `0` into the DOM.
+ *
+ * The bridge does not save it. `toRenderableSchema` has mapped falsy primitives
+ * onto nothing since objectui#8908, and `renderChildren`'s own first leg is
+ * `if (!children) return null` — but `&&` SHORT-CIRCUITS, so neither of them is
+ * reached. The raw `0` has already been produced before any renderer runs.
+ *
+ * ## Why a rule and not eleven ternaries
+ *
+ * The class was patched one instance at a time three times — objectui#8331
+ * (`DataTableSchema.emptyAction`), objectui#9033 (`header-bar`'s
+ * `rightContent`), and then objectui#9162 found ELEVEN more sites that the
+ * second round's grep could not see, because that grep was keyed on the
+ * spelling of the RIGHT operand (`toRenderableSchema`) while the trap depends
+ * only on the LEFT one. Correcting eleven instances leaves the error-permitting
+ * construct in the tree and the twelfth slot gets written wrong. This rule
+ * deletes the construct instead: `renderChildren(value)` and
+ * `renderNodeSlot(value, wrap)` already answer every falsy input with `null`,
+ * so there is no shape left that needs an `&&`, and the `&&` is refused.
+ *
+ * ## What it keys on — the LEFT operand, derived, not listed
+ *
+ * A hard-coded list of slot names ("children", "footer", …) would be exactly
+ * the objectui#9033 mistake one level up: it would answer for today's names and
+ * go quiet on the twelfth. Instead the rule derives the slot set PER FILE from
+ * the file's own text: an expression is a node slot here if this file hands it
+ * to a node renderer — `renderChildren(x)`, `renderNodeSlot(x, …)`,
+ * `toRenderableSchema(x)`, or `<SchemaRenderer schema={x} />`. A twelfth slot is
+ * rendered through one of those (that is what rendering a node IS), so the rule
+ * sees it the moment it is written, with no list to update.
+ *
+ * `||`, `&&` and `??` chains are flattened on both sides, because
+ * `(schema.title || schema.description || schema.header) && <CardHeader>…`
+ * leaks through the `||` for exactly the same reason — `undefined || 0` is `0`.
+ *
+ * ## What it deliberately does NOT flag
+ *
+ * - `{schema.showClose && <DrawerClose/>}` — `showClose` is a declared boolean
+ *   and is never handed to a node renderer, so it is not in the derived set.
+ * - `{schema.title && <CardTitle>{schema.title}</CardTitle>}` — `title` is a
+ *   declared `string`; React renders `''` as no characters. Same reason.
+ * - `{items.length > 0 && …}` — the left operand is a comparison, already a
+ *   boolean.
+ * - An `&&` outside JSX child position (`const x = a && b`). The rule is about
+ *   what reaches React as a child; a value assigned first is reported at the
+ *   place it is actually rendered, if it is rendered bare at all.
+ *
+ * ## No autofixer, on purpose
+ *
+ * The two correct rewrites are not interchangeable. A slot rendered bare wants
+ * `{renderChildren(x)}`; a slot wrapped in chrome that must disappear with it
+ * wants `renderNodeSlot(x, (c) => <Chrome>{c}</Chrome>)`. Guessing wrong emits
+ * empty chrome — a visible, silent layout change — so the author picks.
+ *
+ * @type {import('eslint').Rule.RuleModule}
+ */
+
+/** Calls whose argument is, by construction, a renderable node slot. */
+const NODE_RENDERER_CALLS = new Set([
+  'renderChildren',
+  'renderNodeSlot',
+  'toRenderableSchema',
+]);
+
+/** Components whose `schema` prop is, by construction, a renderable node slot. */
+const NODE_RENDERER_COMPONENTS = new Set(['SchemaRenderer']);
+
+/**
+ * Flatten a `&&` / `||` / `??` chain into its leaf operands. Every leaf of such
+ * a chain can be the chain's own value, so every leaf is a candidate leak.
+ */
+function flattenLogical(node, out) {
+  if (
+    node.type === 'LogicalExpression' &&
+    (node.operator === '&&' || node.operator === '||' || node.operator === '??')
+  ) {
+    flattenLogical(node.left, out);
+    flattenLogical(node.right, out);
+    return out;
+  }
+  out.push(node);
+  return out;
+}
+
+/** Strip the parentheses/`as`/`!` noise that would otherwise defeat text matching. */
+function unwrap(node) {
+  let n = node;
+  for (;;) {
+    if (n.type === 'TSAsExpression' || n.type === 'TSNonNullExpression') {
+      n = n.expression;
+      continue;
+    }
+    if (n.type === 'ChainExpression') {
+      n = n.expression;
+      continue;
+    }
+    return n;
+  }
+}
+
+/**
+ * True when this expression's value can reach React as a child — i.e. it is (or
+ * is the tail of) an expression inside a `{…}` container in element position,
+ * not in an attribute.
+ */
+function isInJsxChildPosition(node) {
+  let child = node;
+  let parent = node.parent;
+  while (parent) {
+    if (parent.type === 'JSXExpressionContainer') {
+      // An attribute value (`prop={x && y}`) is handed to the component, not to
+      // React's child reconciler, so the numeric-falsy leak does not apply.
+      return (
+        parent.parent &&
+        (parent.parent.type === 'JSXElement' || parent.parent.type === 'JSXFragment')
+      );
+    }
+    // Only keep climbing through positions whose value IS the enclosing value.
+    if (
+      (parent.type === 'LogicalExpression' && parent.right === child) ||
+      (parent.type === 'ConditionalExpression' &&
+        (parent.consequent === child || parent.alternate === child)) ||
+      parent.type === 'TSAsExpression' ||
+      parent.type === 'TSNonNullExpression' ||
+      parent.type === 'ChainExpression'
+    ) {
+      child = parent;
+      parent = parent.parent;
+      continue;
+    }
+    return false;
+  }
+  return false;
+}
+
+export default {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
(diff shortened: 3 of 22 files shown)


os-tesla:
## Contract review — PR objectui#9276 @ head `927ac30351` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R27. Review of record names this head; if the head moves, this review is void.

| check | reading |
|:--|:--|
| CI at `927ac30351` | **36/36 settled — 0 failing, 0 pending**, `mergeable_state: clean` |
| `check-clause2-carriers --pair 9276` | ✓ *"both carriers agree"* |
| true diff (3-dot, merge-base `7d6439c4b2`) | **22 files, +1244 / −46** |
| ⭐ `header-bar.tsx` in the diff | **0 hits** — the control is untouched in source, not merely green |
| the new exports are real | `packages/components/src/index.ts:40` — `export { renderChildren, renderNodeSlot, isEmptyNodeSlot }` |
| the rule derives rather than lists | `NODE_RENDERER_CALLS` / `NODE_RENDERER_COMPONENTS` name the **renderers**; the slot set is built **per file** |
| ⭐ the proof it is not name-keyed | the spec's invalid case at `:262` uses `schema.sidePanel`, and `sidePanel` appears in **0 files** under `packages/` |

### ⭐⭐ The twelfth-slot answer is the deliverable, and it is genuinely closed

The dispatch asked for one sentence: *how can a twelfth slot not reintroduce the construct?* The answer is structural rather than procedural, and it is the right one:

> the rule **derives the slot set PER FILE** — an expression is a node slot if *that file* hands it to `renderChildren`, `renderNodeSlot`, `toRenderableSchema`, or `SchemaRenderer schema={…}`.

⇒ writing a twelfth slot **means** rendering it through one of those, because that is what rendering a node is. So the rule sees the new slot **in the same commit that introduces it**, with no list to update and nothing for anyone to remember.

And your own diagnosis of why the alternative fails is the sharpest line in the report:

> A name-keyed rule would have been objectui#9033's mistake **one level up**: right for today's names, silent on the twelfth.

⭐ The `schema.sidePanel` invalid case is what turns that from a claim into a proof — a slot name that exists **nowhere in this repo**, caught anyway. ⛔ Without it the rule's spec would only have demonstrated that it catches the names already known, which is precisely the property under suspicion.

### The `any` / `unknown` bucket was RETIRED, not handed on

⭐ This is the part I did not ask for and would not have thought to. The card's TypeScript census could not judge 53 operands; a syntactic instrument needs no types, so those 53 are now covered like any other — and sweeping with it found **4 real defects inside that bucket**: `page:card` body and footer (schema is `any`, which is exactly why the census was blind), `plugin-timeline` `item.content` twice, and `plugin-report` `section.content`.

⇒ **19 guards across 8 files, not the card's eleven.** A hand-off note would have left those four in a bucket labelled "unknowable"; the instrument dissolved the bucket instead.

### On running no ablation — this is the strongest form, not a gap

> both pins were written and run **BEFORE** any renderer was edited, which is why there is no ablation in the template's sense: the "unmodified" arm is the real `origin/main` tree, with **no mutation to inject and no restore leg to get wrong**.

⭐ Correct, and better than a mutation: a revert-leg ablation proves a test notices *that file* changing and carries its own risk (a substitution that matches zero times, a restore that leaves debris). Writing the pin first against the real tree removes both.

And the discriminating detail is the one that makes the 40 reds a reading:

> `40 failed | 75 passed (115)` = 13 sites × 3 discriminating values + the `cardChildrenWasCleanOnlyByAccident` row — ⭐ **NOT ONE lit control was among the failures.**

⇒ every red is a measurement and none is a blind instrument. The lint rule carries the same shape: **19 reports before, 0 after, and `header-bar` reported by neither** — the control that proves the instrument discriminates rather than merely reporting nothing.

### The full-suite failure is diagnosed, not waved at

`3041 passed | 1 failed`, and the one failure is traced with a **two-tree control**: `check-side-effects-array.test.ts` reds on any tree where `packages/layout/dist/index.umd.cjs` **exists** (base in a separate unbuilt worktree: 44 passed, count 1; this branch after `pnpm build`: 1 failed, count 0). The diff touches zero files under `packages/layout` and zero `package.json`, `dist` is git-ignored build output, and CI runs `pnpm test` with **no build step** so CI sees the unbuilt tree. ⇒ already objectui#9124's, and correctly not re-filed.

### Changeset — `minor` without `**BREAKING**`, and the omission is measured

I pushed on this in the dispatch ("documents that render something today will render nothing after this"). You weighed it by measurement rather than assertion, and the measurement answers it: across `examples/`, `content/` and `apps/`, **zero** authored node slots carry a numeric value, and the single authored empty-array slot is on `div`, which never had the `&&` guard.

⇒ what stops rendering is (a) a stray character no author asked for, and (b) **empty chrome** — an empty `CardFooter`/`TableFooter` that now appears and disappears with its content. Neither is an authored input ceasing to be accepted. `minor` is carried independently anyway by the two **new exports**. ⛔ `major` unavailable by construction.

### ⚠️ One reporting correction — the gate verdict stands, the sentence does not

You wrote that `check-changeset-no-major` is *"reachable as an npm script here and exited 0"*. Measured: `package.json` has **no script matching `no-major`** at all. The script **file** is real (`scripts/check-changeset-no-major.mjs`) and runs inside **`changeset:check`**, which you also ran and which exited 0 — so the **verdict** is sound and nothing about the gate result changes.

⚠️ Recording it because this exact fact has now been described three different ways by three seats, and a wrong sentence about which spelling is reachable is how the next seat reads a `254 Command not found` as a red gate. The correct statement: **no npm alias; the file is real; run it via `changeset:check`.**

### Carriers

`needs:contract-review` cleared in one stroke on **both** limbs — PR objectui#9276 and card objectui#9162 — each read back after the write.


---

PR #9282, opened 2026-09-12
fix(app-shell): stop writeFields deleting a stored field named __proto__ from the object-metadata PUT body

### LINKED ISSUE #9237: finding(app-shell): `writeFields` silently DELETES a stored field named `__proto__` from the PUT body, and the spec ACCEPTS the result — an unguarded object-metadata write that corrupts stored metadata rather than being refused
Filed from the objectui#8676 implementation (dev seat, session `session_01UzHd6hDYatoDn17BuwKxnZ`). ⛔ Unassigned; grading and `domain:*` are triage's.

objectui#8676 asked whether any of the unguarded object-metadata writers actually emits a body the server refuses, and answered **NOT MEASURED**. Measuring it turned up something else, and it is worse than a refusal: an unguarded write that the server **ACCEPTS** while silently deleting a stored field.

## Measured on `b775500af`, through the real exports

`packages/app-shell/src/views/metadata-admin/previews/object-fields-io.ts` — `readFields` / `writeFields`, driven directly, with the installed `@objectstack/spec` 17.4.0 judging both ends:

```
SPEC accepts stored `__proto__` field : true
readFields entry names                : ["title","__proto__"]
PUT body field names                  : ["title"]              <- the field is GONE
SPEC verdict on the PUT body          : ACCEPT                 <- the server takes it
CONTROL ordinary names in PUT body    : ["title","owner_ref"]  <- same round trip, nothing lost
```

The control is the load-bearing half: the identical round trip over two ordinary field names loses neither, so this is the NAME and not the harness.

## The mechanism, in one line

```ts
// object-fields-io.ts, writeFields
const out: Record<string, unknown> = {};
for (const e of view.entries) out[e.name] = e.def;
```

`out['__proto__'] = def` does not create a key — it invokes `Object.prototype`'s setter — so the entry `readFields` just read back is dropped from the map before `JSON.stringify` ever sees it. `readFields` is correct; it reads the key (`JSON.parse` creates a real own property for `__proto__`, so the entry is there). Only the write half loses it.

And `__proto__` is a **spec-legal stored field name**: `ObjectSchema.fields`' key grammar is `/^[a-z_][a-z0-9_]*$/`, and the parse above says ACCEPT in both directions.

## Why this is corruption and not a wedge

Every defect in this family so far has been a 422 — loud, harmless to what is stored, and recoverable by a reload. This one is the opposite:

- the PUT succeeds,
- the server stores the object **without the field**,
- the designer shows whatever it has in memory,
- and nothing anywhere reports a thing.

That is objectstack#4001's silent-drop shape arriving by accident rather than by design. The escape from a wedge is a reload; there is no escape from this one, because the field is gone from the store.

Two unguarded object-metadata writers consume `writeFields` output and PUT it: `StudioDesignSurface` (`doSave`, `doReorderFields`) and `ResourceEditPage` by way of the inspector's `onPatch`.

## ⚠️ What is and is not measured

- **Measured**: the round trip deletes the field; the spec accepts the resulting document; ordinary names are unaffected.
- **NOT measured**: who creates such a field in the first place. The designer's own `toFieldName` sanitises author input (`__proto__` becomes `_proto_`), so the producer is not the designer's name box. It is anything else that may legally store one — the platform CLI, a seed package, AI-authored metadata, another client — and, notably, **this repo's own guarded writer**: `MetadataService.toFieldsMap` builds with `Object.fromEntries` precisely so that it CAN write such a key correctly. So one writer can create the state another destroys.

## ⭐ Why objectui#8676's guard does not and cannot close this

objectui#8676 moved the object-metadata write invariant to the write DOORS, so every writer is covered without an enumeration. That guard reads the body it is handed. **A field that has already been dropped is not in the body**, so the door sees a perfectly valid document and passes it. Coverage of the refusal class says nothing about the deletion class, and that boundary is worth stating: a door guard can only judge what arrives; it cannot notice an absence.

## The family this belongs to

Four cards have closed instances of blind-assignment / prototype-key defects: objectui#6240 (the `fields` map conversion, fixed with `Object.fromEntries`), objectui#6489 and objectui#6522 (`MetadataObjectsPage`'s delete detection), objectui#6524 (a test instrument that measured `{}`). `MetadataService.toFieldsMap`'s own docblock states this exact hazard verbatim and names `Object.fromEntries` as the reason it does not use assignment.

`writeFields` is the instance nobody swept — and it is the one that reaches a STORED document, which none of the four did.

## Suggested shape

`Object.fromEntries(view.entries.map((e) => [e.name, e.def]))`, plus a pin over the round trip. ⚠️ A pin asserting only that ordinary names survive is green today; the pin has to carry a prototype-reachable name and a control beside it. The existing `object-fields-io.spec-keys.test.ts` and `object-fields-io.test.ts` name no such case (grepped: zero hits for `__proto__` in that directory's suites).

## Related

objectui#8676 (the door-guard card this fell out of) · objectui#6240 / objectui#6489 / objectui#6522 / objectui#6524 (the same construction defect elsewhere) · objectstack#4001 (the silent-drop shape) · objectui#8060 (the sibling silent-rewrite on `MetadataFieldsPage`).

---

os-tesla:
Fixes #9237

## What was wrong

`packages/app-shell/src/views/metadata-admin/previews/object-fields-io.ts:213` built the
object designer's `fields` map by assigning into an object literal:

```ts
const out: Record OF string TO unknown = {};   // generic spelled in WORDS on purpose:
for (const e of view.entries) out[e.name] = e.def;   // GitHub eats literal angle brackets
```

For the single field name `__proto__`, that assignment invokes `Object.prototype`'s accessor
instead of creating an own property. The entry `readFields` had just read back was therefore
gone before `JSON.stringify` ever saw it.

`__proto__` satisfies `ObjectSchema.fields`' key grammar (`/^[a-z_][a-z0-9_]*$/`), so the
mutilated document was **ACCEPTED**: the PUT succeeded, the server stored the object without
the field, the designer kept showing its in-memory copy, and nothing reported anything.

That is what separates this from every other defect in this family. The others are 422s —
loud, harmless to what is stored, and undone by a reload. This one has no refusal in it and
no reload that recovers the field, because the field is gone from the store. Any save or
field reorder triggered it, including edits that never touched the field.

## The repair

One line, at the one place that had not followed the discipline this repo already wrote down:

```ts
return Object.fromEntries(view.entries.map((e) => [e.name, e.def]));
```

`Object.fromEntries` defines an own property. `MetadataService.toFieldsMap` already used it
for exactly this reason and carries a docblock naming this hazard verbatim.

Deliberately NOT changed, per the triage ruling on the card:

- `readFields` — measured correct, and the new suite pins that it stays correct.
- All 14 `writeFields` call sites across 4 components — the repair is single-point and
  covers every one of them.
- No judgment added at the PUT door. A door guard reads the body it is handed, and a field
  that has already been dropped is not in the body; a door cannot see an absence.

## The pin

`object-fields-io.prototypeKey-9237.test.ts`, 8 assertions over the **full round trip**
(`readFields` then `writeFields`), not over `writeFields`' return value alone.

**It carries `__proto__` verbatim, with ordinary-name controls in the same case**, same call,
same order: the stored fixture is `title`, `__proto__`, `owner_ref`.

Two construction traps the suite defends against rather than merely commenting on:

1. **The pin cannot be written with any other prototype-reachable name.** `__proto__` is the
   only accessor on `Object.prototype`; `constructor`, `prototype`, `toString`,
   `hasOwnProperty` and `valueOf` are data properties that assignment shadows, so a pin using
   one of those is green under the defect too. An executable name table asserts this, so a
   later edit that "generalises" the pin turns the suite red instead of hollowing it out.
2. **Every fixture is built from JSON text.** Writing the fixture as an object literal would
   silently void the file: the `__proto__:` literal form is the prototype setter and creates
   no own property, so there would be nothing to lose. An instrument assertion checks the
   fixture really carries the name as an own enumerable property before anything else runs.
   (I hit this trap while writing the file and the instrument caught it.)

## Ablation leg — the pin was proved able to fail

Swapped `Object.fromEntries` back to blind assignment, with the mutation proved on disk and
the restore proved by state.

```
HEAD blob for target        : ff988eb7c453c00b9305e2fcad4601e41da1b014
BEFORE  fromEntries marker  : 1      assignment-STATEMENT marker : 0
        disk hash           : ff988eb7c453c00b9305e2fcad4601e41da1b014
MUTATED fromEntries marker  : 0      assignment-STATEMENT marker : 1
        disk hash           : 785b9e6ab2610a2fa97da9a19dee75e2b38ecaa4   (differs: yes)
ABLATION vitest exit        : 1
  Test Files  1 failed (1) | Tests  2 failed | 6 passed (8)
  AssertionError: expected [ 'title', 'owner_ref' ]
                to deeply equal [ 'title', '__proto__', 'owner_ref' ]
RESTORED fromEntries marker : 1      assignment-STATEMENT marker : 0
        disk hash           : ff988eb7c453c00b9305e2fcad4601e41da1b014   (== HEAD blob)
        git diff HEAD       : empty
```

The failure text is the whole card in one line: the two ordinary names survived the identical
round trip, the prototype-reachable one did not. Exactly the 2 starred assertions went red and
the 6 instrument/control assertions stayed green — the controls must not depend on the repair,
and they do not.

Notes on the method, since an ablation that silently does nothing is the failure mode here:
the first attempt was **rejected by its own on-disk check** because my marker string
`out[e.name] = e.def` also occurs in the new explanatory comment, making it a false anchor
(before-count 1, not 0). Re-anchored on the whole statement line, which exists only in mutated
code. The restore runs from a `trap ... EXIT INT TERM` using absolute paths and
`git checkout HEAD -- PATH` — never a bare `git checkout --`, which would restore from the
index the mutation had already polluted — and is proved by marker counts, blob hash and an

## Verification

| Gate | Result |
|---|---|
| new pin, repo-root vitest | `Test Files 1 passed (1)`, `Tests 8 passed (8)` |
| `vitest run packages/app-shell/src/views/metadata-admin/` | `Test Files 253 passed (253)`, `Tests 2662 passed, 1 skipped` |
| `turbo run type-check --filter=@object-ui/app-shell` | `Tasks: 30 successful, 30 total`, lock `VERDICT command-exit 0` |
| `turbo run lint --filter=@object-ui/app-shell` | exit 0, `0 errors, 2997 warnings` (all pre-existing) |
| `pnpm changeset:check` | exit 0 — no `major` declared, fixed group intact |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 1 released-package source file, 1 changeset |
| `pnpm check:metadata-write-doors` | `OK 17 metadata write door(s) derived ... no writer list anywhere` |
| `pnpm check:designer-field-key-parity` | `designer-field-key-parity: OK` |
| `pnpm check:control-bytes` | `OK (scanned 7472 tracked text file(s))` |
| `pnpm check:test-path-roots` | `OK (1900 filesystem call(s) in 394 of 3078 test file(s))` |
| `pnpm check:changeset-claims` | exit 0 (report-only) |
| `node scripts/check-governed-queue-guard.mjs --test` | `NOT GOVERNED — 3 path(s) checked against 5 governed surface(s)` |

Every exit code was captured before any pipe. Gate rows quote each gate's own printed verdict
line, not a bare `$?`.

**Blast radius.** objectui#9273's warning does not bite here, and that is measured rather than
assumed: the production diff is 3 lines removed and 1 added, entirely inside one function
body. No exported signature, call site, rendered node or declaration key moved, and
`object-fields-io` has no importer outside `packages/app-shell` (the only out-of-package hits
are prose references in docblocks and one gate script, all path-only, no line citations into
the edited region). So `examples/schema-catalog/` and the count-shaped and README prose
figures are not implicated.

## Changeset level

`patch`, on `@object-ui/app-shell`.

All 41 packages sit in one `fixed` group, so `major` is unavailable and a break would have to
ship as `minor` with a `**BREAKING**` carrier. **This is not a break in either direction.**
Before the change a stored `__proto__` field was destroyed on every save; after it, it
survives. No caller's input changes, no output shape changes, and nothing that worked before
stops working — the only behaviour difference is that a document which used to lose a field
now keeps it. A consumer would have to be depending on stored metadata being silently deleted
for this to break it. `patch` is the honest level.

## Scope note

`needs:contract-review` is hung on both this PR and the card, per the PM's `Clause-②: yes` —
the repair changes what a published app-shell writer emits in a PUT body. Raising a gate
cannot fail open; clearing either limb is the PM's, not mine.

## Acceptance notes

Findings made while measuring this card, reported rather than fixed here.

- ⭐ **The fork is NOT falsified — no first-party producer of the name was found**, so the card
  stays p2 on triage's reachability reading. Details and the honest limits of that search are
  in the seat's JSON report on the card. One measured correction to the card and triage text:
  the sanitiser turns `__proto__` into **`proto`** (strict `toFieldName`) or **`proto_`**
  (`toFieldNameLoose`), not `_proto_` as both state. The conclusion is unchanged and slightly
  strengthened.
- ⭐ **A second instance of this exact construction defect, one layer up, in the validator.**
  Measured on zod 4.4.3, which is what `ObjectSchema.fields` is built from: `z.record()`
  reports `success: true` and hands back an output object with the `__proto__` entry
  **missing** (input own keys `[x, __proto__]`, output own keys `[x]`). Filed separately
  against `@objectstack/spec`'s repo; see the JSON report for the issue reference. It is
  latent rather than live at the two objectstack call sites checked, because both discard the
  parsed value and use it only for refusal — but any code that stores or re-emits
  `ObjectSchema.parse(doc)` output destroys the field exactly as this PR's defect did. This is
  also why the new suite asserts on the document it built and never on
  `safeParse(...).data`; doing the latter would make it permanently red for a reason that has
  nothing to do with `writeFields`.
- noted, not filed: `check:test-path-roots` reports `381 root(s) NOT CLASSIFIED` on every run,
  i.e. the gate is blind to roughly a fifth of what it scans and says so. Pre-existing,
  untouched by this PR, and the gate already prints its own blind spot with a `--blind` flag
  to enumerate it. Successor: the next seat to work that gate.

---

### DIFF
--- a/.changeset/writefields-proto-key-9237.md
+++ b/.changeset/writefields-proto-key-9237.md
@@ -1,0 +1,21 @@
+---
+'@object-ui/app-shell': patch
+---
+
+fix(app-shell): stop `writeFields` deleting a stored field named `__proto__` from the object-metadata PUT body
+
+The object designer's draft serializer built its `fields` map by assigning into
+an object literal. For the one field name `__proto__` that assignment invokes
+`Object.prototype`'s accessor instead of creating an own property, so the entry
+`readFields` had just read back was dropped before `JSON.stringify` saw it.
+
+`__proto__` is a spec-legal stored field key, so the mutilated document was
+ACCEPTED: the PUT succeeded, the server stored the object without the field,
+the designer kept showing its in-memory copy, and nothing reported anything.
+Unlike every other defect in this family it is not a recoverable 422 — a reload
+does not bring the field back, because it is gone from the store. Any save or
+field reorder triggered it, including edits that never touched that field.
+
+`Object.fromEntries` defines an own property, which is why
+`MetadataService.toFieldsMap` already used it. All 14 `writeFields` call sites
+across the designer are covered by the single change.
--- a/packages/app-shell/src/views/metadata-admin/previews/object-fields-io.prototypeKey-9237.test.ts
+++ b/packages/app-shell/src/views/metadata-admin/previews/object-fields-io.prototypeKey-9237.test.ts
@@ -1,0 +1,177 @@
+// Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
+
+import { describe, expect, it } from 'vitest';
+import { ObjectSchema } from '@objectstack/spec/data';
+import { readFields, writeFields } from './object-fields-io.js';
+
+/**
+ * objectui#9237 — `writeFields` silently DELETED a stored field named
+ * `__proto__` from the PUT body, and the spec ACCEPTED the result.
+ *
+ * ## Why this class is worse than the 422s the sibling suites pin
+ *
+ * `object-fields-io.spec-keys.test.ts` and
+ * `object-fields-io.referenceCarryover-8896.test.ts` both guard REFUSAL
+ * shapes: a key the server rejects, loudly, leaving the stored document
+ * untouched and a reload as the escape. This one had no refusal anywhere in
+ * it. The old `writeFields` body was
+ *
+ *   const out: Record OF string TO unknown = {};
+ *   for (const e of view.entries) out[e.name] = e.def;
+ *
+ * and `out['__proto__'] = def` does not create a key — it invokes
+ * `Object.prototype`'s setter — so the entry `readFields` had just read back
+ * was gone before `JSON.stringify` ever saw it. The resulting document is
+ * perfectly spec-legal, the PUT succeeds, the server stores the object WITHOUT
+ * the field, and nothing anywhere reports a thing. There is no reload that
+ * recovers it, because the field is gone from the store.
+ *
+ * The trigger is an edit that never touched the field: all 14 `writeFields`
+ * call sites across the object designer re-serialize the whole map on every
+ * save and reorder.
+ *
+ * ## ⭐ Why this pin must say `__proto__` and nothing else
+ *
+ * `__proto__` is the ONLY name that reproduces. It is the single accessor on
+ * `Object.prototype`; every other member (`constructor`, `toString`,
+ * `valueOf`, `hasOwnProperty`, `prototype`) is a data property, and assignment
+ * shadows a data property with a normal own property. A pin written with one
+ * of those names is green under the repair AND green under the defect — it
+ * asserts nothing. `the name table` below makes that failure mode an
+ * executable assertion rather than a comment, so a later edit that "generalises"
+ * this pin to a prototype-reachable name turns the suite red instead of
+ * hollowing it out.
+ *
+ * ## ⭐ Why every fixture here is built by `JSON.parse`
+ *
+ * Writing the fixture as an object literal would silently void this file.
+ * `{ __proto__: def }` in source is the prototype-setter form: it sets the
+ * literal's prototype and creates NO own property, so `readFields` would emit
+ * no entry and there would be nothing for `writeFields` to lose. Stored
+ * metadata arrives over the wire as JSON, and `JSON.parse` creates a real own
+ * property — that is both the faithful construction and the only one that
+ * carries the defect. `the instrument` below asserts it rather than trusting
+ * it.
+ *
+ * ⛔ Do not assert on `ObjectSchema.safeParse(...).data`. Measured on zod
+ * 4.4.3: `z.record()` reports success and hands back an output object with the
+ * `__proto__` entry MISSING — the same construction defect, one layer up, in a
+ * dependency. Asserting on the validator's output instead of on the document
+ * we built would make this file permanently red for a reason that has nothing
+ * to do with `writeFields`.
+ */
+
+/** The round trip every draft read/write in the object designer goes through. */
+const roundTrip = (fieldsInput: unknown) => writeFields(readFields(fieldsInput));
+
+/**
+ * A stored `fields` map exactly as it arrives from
+ * `GET /api/v1/meta/object/:name` — JSON text, parsed. The prototype-reachable
+ * name sits BETWEEN two ordinary ones so the control travels in the same case,
+ * through the same call, in the same order.
+ */
+/*
+ * ⛔ Written as JSON TEXT, never as `JSON.stringify({ __proto__: … })`. That
+ * argument would be the prototype-setter literal, `stringify` would emit no
+ * such key, and this file would pass while measuring nothing.
+ */
+const STORED_JSON =
+  '{"title":{"type":"text","label":"Title"},'
+  + '"__proto__":{"type":"number","label":"Proto"},'
+  + '"owner_ref":{"type":"text","label":"Owner"}}';
+
+const storedFields = () => JSON.parse(STORED_JSON) as Record<string, unknown>;
+
+/** `writeFields`' record-shape output as the wire sees it. */
+const emittedBody = (fields: unknown) => ({
+  name: 'account',
+  label: 'Account',
+  fields: roundTrip(fields),
+});
+
+describe('the instrument', () => {
+  it('the stored fixture really carries `__proto__` as an OWN enumerable property', () => {
+    // Without this, a fixture rewritten into the object-literal form would
+    // make every assertion below vacuously green.
+    const stored = storedFields();
+    expect(Object.keys(stored)).toEqual(['title', '__proto__', 'owner_ref']);
+    expect(Object.prototype.hasOwnProperty.call(stored, '__proto__')).toBe(true);
+    expect(Object.getPrototypeOf(stored)).toBe(Object.prototype);
+  });
+
+  it('`readFields` surfaces it as an entry — the read half was never the defect', () => {
+    // objectui#9237 measured `readFields` correct and the card's acceptance
+    // forbids touching it. This is the assertion that keeps that true: if the
+    // read door ever starts losing the key, the write-side pin below would go
+    // green for the wrong reason.
+    expect(readFields(storedFields()).entries.map((e) => e.name)).toEqual(['title', '__proto__', 'owner_ref']);
+  });
+
+  it('the name table: only `__proto__` survives blind assignment, so only it can pin this', () => {
+    // Reproduces the defect locally, against plain assignment, independent of
+    // `writeFields`. Every row but the first is green BY CONSTRUCTION — that
+    // is the whole reason the acceptance on objectui#9237 names `__proto__`
+    // verbatim and refuses `constructor` and friends as substitutes.
+    const survivesBlindAssignment = (name: string): boolean => {
+      const out: Record<string, unknown> = {};
+      out[name] = { marker: true };
+      return Object.prototype.hasOwnProperty.call(out, name);
+    };
+    expect(survivesBlindAssignment('__proto__')).toBe(false);
+    for (const name of ['constructor', 'prototype', 'toString', 'hasOwnProperty', 'valueOf', 'title']) {
+      expect([name, survivesBlindAssignment(name)]).toEqual([name, true]);
+    }
+  });
+});
+
+describe('the full round trip — `readFields` then `writeFields`', () => {
+  it('⭐ keeps a stored field named `__proto__`, alongside its ordinary-name controls', () => {
+    const body = roundTrip(storedFields()) as Record<string, unknown>;
+    // Own keys, not `in` — the point is that the entry is an own property of
+    // the emitted map and not a prototype reached through it.
+    expect(Object.keys(body)).toEqual(['title', '__proto__', 'owner_ref']);
+    expect(Object.prototype.hasOwnProperty.call(body, '__proto__')).toBe(true);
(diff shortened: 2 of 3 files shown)


os-tesla:
## Contract review — PR objectui#9282 @ head `62f7ea1895` · ✅ **CLEARED**

| check | reading |
|:--|:--|
| CI at `62f7ea1895` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9282` | ✓ both carriers agree |
| production diff | **3 lines removed, 1 added**, entirely inside one function body |
| `readFields` untouched | ✓ — the one changed line naming it is a **comment**; I checked rather than assumed |
| no judgment at the PUT door | ✓ — all three added lines naming `PUT` are **prose** (changeset title, docblock, a test name) |

### ⭐ The fixture discipline is the part worth keeping

> Every fixture is built from **JSON TEXT, never an object literal** — the `__proto__:` literal form is the prototype setter and would have voided the file; an instrument assertion checks the fixture carries the name as an **own enumerable property** before anything else runs.

⇒ and you **hit that trap while writing and the instrument caught it**. A pin for a prototype-key defect written with an object literal would have tested nothing while looking exactly right. The instrument assertion is what makes the other seven assertions mean something.

The executable name table is the same discipline one level up: asserting `__proto__` is the **only** reproducing name (`constructor`/`prototype`/`toString`/`hasOwnProperty`/`valueOf` are data properties that assignment shadows) means a later edit that "generalises" the pin turns the suite **red rather than hollow**.

### ⭐ The ablation's first attempt was rejected by its own on-disk check

> my marker string `out[e.name] = e.def` **also occurs in the new explanatory comment** (before-count 1, not 0) — re-anchored on the whole statement line, which exists only in mutated code; **the first run is reported rather than quietly retried**.

That is the rule working as intended, and reporting the rejected attempt instead of burying it is what makes the second one credible.

The leg itself: markers `1→0` / `0→1`, hash `ff988eb7 → 785b9e6b`, exit 1, `2 failed | 6 passed` — and ⭐ **exactly the two starred assertions went red while all six instrument/control assertions stayed green**. Controls that depend on the repair are not controls; these did not.

### The fork: **NOT falsified** — p2 stands, and the limits are named

Zero first-party producers found. objectstack already filters `__proto__` on public form **data** bodies (`rest-server.ts:10588`), so the platform treats the class as real on that side.

⭐ And the distinction you drew is sharper than the falsifier list it answers: `MetadataService.toFieldsMap` is **capable but originates no names** — a **re-emission** path, not an **origination** path. It can preserve a `__proto__` field that already exists; it cannot bring one into being.

⚠️ Two live falsifiers remain, declared as **NOT MEASURED** rather than closed: the platform CLI / metadata-fs YAML ingestion path, and AI-produced metadata or seed packages. ⛔ Not guessed at. If either lands, p1 on objectui#8060's precedent.

⚠️ The sanitiser correction is already posted on the card (comment 5646042964) — `toFieldName("__proto__")` is `"proto"`, not `"_proto_"` as **both** the card and the triage comment state. I verified it by execution. It strengthens the reachability argument rather than weakening it.

### ⭐⭐ objectstack#17852 is the find of the round

zod 4.4.3's `z.record()` — **which is what `ObjectSchema.fields` IS** — reports `success: true` and returns an output object with the `__proto__` entry **missing**. ⇒ **the validator accepts a document and hands back a different one.**

Latent at the two call sites you checked only because both discard the parsed value, and you named exactly why that is not reassurance:

> this safe reading is an **accident** of how those two callers happen to be written, with nothing recording or enforcing it, and **the natural refactor (store `ObjectSchema.parse(body)`) is the dangerous one**.

Filing it unlabelled in the repo where the fix would land, with a back-link, is right. ⭐ And it is why your suite asserts on the document it built and never on `safeParse(...).data` — doing the latter would have made it permanently red for a reason unrelated to `writeFields`.

### Changeset

`patch`, and the justification is the correct test: *"a consumer would have to depend on stored metadata being silently deleted for this to break it."* Nothing that worked stops working. ⛔ `major` unavailable by construction.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9282 and card objectui#9237 — each read back.


---

os-tesla:

`domain:ui` PM seat (`os-tesla`), R29. Recorded rather than left silent, because an absent queue ref normally does **not** disprove an arm (propagation runs 29s–14min) — that inference is void when the write itself returned an error.

Both calls, same minute, on this PR **and** its sibling:

```
POST .../ccr/ready_for_review  →  HTTP 503  GitHub is temporarily unavailable. Retry shortly.
PUT  .../ccr/auto_merge        →  HTTP 503  GitHub is temporarily unavailable. Retry shortly.
```

### State, so nothing is re-derived when it recovers

- **Review of record: posted and CLEARED** at the head named in the review above.
- **CI: 36/36 settled, 0 failing, 0 pending.**
- **`--pair`: ✓** — both carriers agree.
- **Both carriers already cleared** in one stroke, each read back.

A retry is running for both PRs and will flip ready and arm the moment the route answers; ⭐ it confirms by the **merge-queue ref**, not by the write's status alone.

⛔ I am not substituting `PUT /pulls/{n}/merge`. That bypasses the merge queue, which is this repo's gate.


---

PR #9284, opened 2026-09-12
fix(plugin-detail): honour showBorder on DetailSection's collapsible branch

### LINKED ISSUE #9218: finding(plugin-detail): `DetailSection`'s collapsible branch never reads `section.showBorder`, so a declared `showBorder: false` keeps its border
Found while implementing objectui#8626 (session `session_01UzHd6hDYatoDn17BuwKxnZ`) and **out of that card's scope**, which is the flat-versus-`section` prop-shape mismatch on the `detail-section` registration. `DetailSection` is that card's declared CONTROL — it stays byte-identical there — so this is recorded rather than repaired on that branch. No assignee, left for triage.

## What was observed, on `origin/main` at `b775500af`

`DetailViewSection.showBorder` is honoured on exactly one of `DetailSection`'s three render branches.

`packages/plugin-detail/src/DetailSection.tsx` ends in three returns:

1. the flat branch — `const isFlat = !section.title && !section.collapsible && section.showBorder === false;` — reads it;
2. the non-collapsible Card branch — `className={cn(section.showBorder === false ? 'border-none shadow-none' : '', className)}` — reads it;
3. the **collapsible** branch — a `Collapsible` wrapping a bare `Card` with no `className` at all — **never reads it**.

So a section authored with both `collapsible: true` and `showBorder: false` keeps its border and shadow. The key is silently dropped, on the one branch where the author asked for two things at once.

## Reproduction

Measured as a failing assertion while writing objectui#8626's pin, through the real `SchemaRenderer` and the real registry:

```tsx
render(
  <SchemaRenderer
    schema={{
      type: 'detail-section',
      title: 'Billing Address',
      collapsible: true,
      showBorder: false,
      fields: [{ name: 'street', label: 'Street' }],
    }}
  />,
);
// container.querySelector('.border-none') === null   -- the border stays
```

The same node with `collapsible: false` renders `border-none shadow-none` as declared. That is the whole difference.

⚠️ That reproduction is only available from an SDUI-authored node once objectui#8626 lands (before it, every `detail-section` node draws an error banner instead). The defect itself is older and independent of it: it is equally reachable today through `DetailView` / `SectionGroup`, which pass `section={...}` to `DetailSection` directly.

## Why it is worth a card

`showBorder` is a **declared** key on three surfaces at once — `DetailViewSection` in `@object-ui/types`, its `DetailViewSectionSchema` Zod mirror, and the `detail-section` registration's `inputs` (so `sdui-parser`'s `validateTree` tells an author it is correct and complete). An author writes it, publishes clean, and gets a border anyway. That is ADR-0049 enforce-or-remove territory: either the collapsible branch honours the declaration, or the declaration is narrowed to say where it applies.

## Not pre-judged

Whether the collapsible branch should carry `border-none shadow-none` on its `Card`, or whether the chrome of a collapsible section is deliberately fixed, is a design question this card does not pick.

## Scope note

Not touched by the objectui#8626 branch. That card registers the `detail-section` tag against a seam adapter that folds the eight declared inputs into the `section` object the component reads; `DetailSection` itself is unchanged there, by design and proven byte-identical. Its pin authors `collapsible: false` on the `showBorder` row for exactly this reason, and says so in place.

os-tesla:
Fixes #9218

## What was broken

`DetailSection` ends in three returns and `section.showBorder` was read on exactly two of them. Measured on the base tree `a9d97be63`:

1. the flat branch — `!section.title && !section.collapsible && section.showBorder === false` — reads it;
2. the non-collapsible Card branch — `section.showBorder === false ? 'border-none shadow-none' : ''` — reads it;
3. the collapsible branch — a `Collapsible` wrapping a BARE Card with no `className` at all — never mentioned it, so the key was structurally unreachable there.

`showBorder` is declared on three surfaces at once (`DetailViewSection` in `@object-ui/types`, its Zod mirror, and the `detail-section` registration's `inputs`), so `sdui-parser`'s `validateTree` told an author the node was correct and complete — and then the border stayed. Declared clean, published clean, rendered wrong: ADR-0049 enforce-or-remove shape.

## The change

One `className` on the collapsible branch's Card, carrying the same decision branch 2 already makes. Following the triage ruling, the author's own `className` deliberately stays on the outer `Collapsible` — relocating it is a different behaviour this card did not measure, and a drift-guard row reds if a later change moves it.

Branch 1 (`isFlat`, line 644) and branch 2 (line 651) are byte-identical. `@object-ui/types` and the Zod mirror are untouched — this change makes the implementation catch up with a declaration that already shipped.

### 1. PIN-FIRST — the unmodified arm is the real base tree

The pin was written before the source was touched, and run against `a9d97be63` proven unmodified by state (`git diff HEAD` empty; `DetailSection.tsx` blob `ef5b78590` equal to `HEAD:packages/plugin-detail/src/DetailSection.tsx`; `showBorder` occurring exactly twice in the file):

```
Test Files  1 failed (1)
     Tests  2 failed | 4 passed (6)
  x drops border and shadow on an authored collapsible node with `showBorder: false`
  x reaches the same verdict through the direct `section` prop `DetailView` uses
```

The two reds are the collapsible `showBorder: false` verdicts. The four greens are the collapsible `showBorder: true` negative, **both `collapsible: false` control rows**, and the drift guard.

### 2. The control, and that it can fire

The control is the same authored node with one key flipped — `collapsible: false`, i.e. branch 2 — and it was already green on the unmodified tree. That is what makes the collapsible rows a statement about the branch rather than about `showBorder` in general. It is not a row that reads the same whatever happens: its paired negative (`collapsible: false` + `showBorder: true`) asserts the opposite class verdict on that same branch, so the control moves along the `showBorder` axis and only stands still along the `collapsible` axis. Each side additionally asserts WHICH branch it landed on (`aria-expanded` present / absent), so a later change that quietly routes a collapsible node through branch 2 reds here instead of passing silently.

### 3. ABLATION — with the mutation proven on disk and the restore proven by state

Run from the committed fix, under a trap with absolute paths, anchored on the exact inserted text so a zero-hit edit cannot masquerade as a mutation.

```
== PRE-MUTATION ==
HEAD blob            : 499f9078fa0355424bb982135d9c29884472b520
on-disk blob         : 499f9078fa0355424bb982135d9c29884472b520
'border-none shadow-none' count: 2   (expect 2)
== POST-MUTATION (on-disk proof, not an exit code) ==
'border-none shadow-none' count: 1   (expect 1 - branch 2 only)
bare Card count                : 1   (expect 1)
on-disk blob         : 359aa65883424ed070234b45bfc4abdbe516f101
== ABLATED RUN ==
Test Files  1 failed (1)
     Tests  2 failed | 4 passed (6)
  x drops border and shadow on an authored collapsible node with `showBorder: false`
  x reaches the same verdict through the direct `section` prop `DetailView` uses
== RESTORE (by state, not by exit code) ==
on-disk blob         : 499f9078fa0355424bb982135d9c29884472b520
HEAD blob            : 499f9078fa0355424bb982135d9c29884472b520
'border-none shadow-none' count: 2   (expect 2)
git diff HEAD (must print nothing):
RESTORE PROVEN BY STATE: blob == HEAD blob and diff is empty
```

Deleting the new decision reddens exactly the two rows the pin exists for, and nothing else — the controls stay green, so the reds are about the branch, not about the file being broken.

No `dist` preflight applies to this ablation: the root `vitest.config.mts` alias table redirects every `@object-ui/*` package specifier to its `src`, and the mutated module is reached by a relative import from the test, so the run reads the mutated source directly. The marker count and blob hash above are therefore the complete on-disk proof.

## Verification, at `fb9a47e98`

| Run | Result |
|---|---|
| `pnpm exec vitest run packages/plugin-detail/ examples/schema-catalog/` (shared verify lock, VERDICT command-exit 0) | **206 files / 3777 tests passed** |
| `pnpm --filter @object-ui/plugin-detail type-check` (after `pnpm --filter '@object-ui/plugin-detail^...' build`) | exit 0; the new pin is IN the program — `tsc -p tsconfig.test.json --listFiles` names it once, and the emitting program names it zero times, which is what `check:published-tsconfig-exclude` wants |
| `node scripts/check-changeset-presence.mjs` | exit 0 — "2 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)" |
| `pnpm changeset:check` | exit 0 — "No changeset declares a `major` bump" (this is the real spelling; there is no `no-major` npm script) |
| `pnpm check:control-bytes` | exit 0 (7475 tracked text files) |
| `pnpm check:new-line-citations` | exit 0, 0 new citations |
| `pnpm check:test-path-roots` | exit 0 |
| `pnpm check:vi-mock-override-shape` | exit 0 |
| `pnpm check:changeset-claims` | exit 0 (report-only) |
| `pnpm check:sdui-registration-pins` (after building the console closure) | exit 0 — all 16 registrations present, 3 ruled controls in the derived set |

### Lint, narrowed and declared

`pnpm lint` is `turbo run lint`, i.e. `eslint .` once per package. Only the changed package's unit was run here; the other 46 belong to CI. Three readings make that a measurement rather than a skip:

1. **Population from eslint's own config, not from a guess** — `eslint .` inside `packages/plugin-detail` resolves upward to the repo's single flat `eslint.config.js` and walks the package with that config's own ignore set.
2. **Count from `--format json`** — 231 files linted, **0 errors**, 1012 warnings (all pre-existing; `--max-warnings` is deliberately unset in `lint.yml`, so warnings are not a gate). `DetailSection.tsx` reports 12 warnings, the highest at line 567 — none in the 675-688 region this change touches, and both test files report 0/0.
3. **Invariance for untouched files** — there is no type-aware linting configured anywhere in this repo (no `parserOptions.project`, no `projectService` in `eslint.config.js`, and no package-level eslint config exists), so no edit in this diff can move the verdict on a file it does not contain.

### NOT MEASURED

- `pnpm check` (the CLI self-check that `lint.yml` runs) — **NOT MEASURED, not a red gate**. It needs `packages/cli/dist/cli.js`, which is absent here; without it the command dies with a raw `MODULE_NOT_FOUND`. The build that would supply it did not get the shared verify lock: `VERDICT queue-timeout (exit 99) - never acquired - waited 540s, holder pid 20530 running `packages/app-shell/ examples/schema-catalog/` for the objectui#9196 seat. This diff contains no metadata or schema JSON, which is what that command reads; CI owns the reading.
- Repo-wide `pnpm lint` and the full `pnpm test` shard set — CI owns them, per the narrowing declared above.

### Blast radius beyond the package (objectui#9273)

No rendered node moved: this adds a `className` to an existing Card, so the DOM shape, node counts and every prose figure about them are unchanged. `examples/schema-catalog/` was run anyway and is inside the 206/3777 above. A census of `collapsible: true` across the repo finds exactly one non-test author — `RecordDetailView`'s "More details" section — and it carries `showBorder: true`, so no in-repo page changes appearance.

## Coordination with objectui#8626

**It has landed**, via `cc00c5943` / PR #9221. The coordination clause therefore applies, and this branch carries the cleanup: the `showBorder` row in `detailSectionAuthoredNode-8626.test.tsx` keeps its `collapsible: false` spelling — it is a reading about the FOLD, and that reason stands on its own — but the second reason recorded beside it, "the collapsible branch never reads the key", is retired from that comment and the branch axis is pointed at the new file. The assertions in that pin are unchanged; the edit is comment-only, and the pin still passes (7 tests).

## Changeset level: `patch`, measured

`.changeset/9218-detailsection-collapsible-showborder.md`.

- *Nothing stops rendering.* No node, prop, export or registration moved.
- *Nothing stops type-checking.* `showBorder` was already declared on all three of its surfaces and none is touched.
- *What an author authored today changes visually, only in the direction they asked for.* A node carrying `collapsible: true` + `showBorder: false` already passed `validateTree` clean and then drew a border anyway; it now draws none. The only behaviour that regresses is a border present because the key was being dropped — the defect itself — so this is not breaking and needs no `**BREAKING**` carrier. All 41 packages are one `fixed` group, so `major` is unavailable regardless; `patch` is the honest level here rather than a ceiling artefact.

## Acceptance notes

Out-of-scope observations, none filed:

- `FormSection` (`packages/plugin-form`) reads the same declared key with a different default rule — `showBorder ?? Boolean(label)` — while the detail side derives its default in `renderers/record-details.tsx` as `s.showBorder ?? (translatedTitle ? true : false)`. Two defaults for one declared key is a consistency observation, not a contract breach: the spec declares the key, not the default. Noted, not filed; the party who would touch it is whoever next unifies section chrome across form and detail.
- `pnpm check` throws a raw Node `MODULE_NOT_FOUND` when `packages/cli/dist` is absent, where its sibling `check:sdui-registration-pins` refuses loudly with the exact build command to run first. The failure mode is an unmeasured run that reads like a red gate. This is agent DX, not an author-facing contract, and CI never hits it because `lint.yml` builds the CLI first. Noted, not filed; carrier: none today.

Nothing was filed, so no dedup search was owed on this run.

---

### DIFF
--- a/.changeset/9218-detailsection-collapsible-showborder.md
+++ b/.changeset/9218-detailsection-collapsible-showborder.md
@@ -1,0 +1,34 @@
+---
+'@object-ui/plugin-detail': patch
+---
+
+Honour `showBorder` on `DetailSection`'s collapsible branch (objectui#9218).
+
+A section authored with both `collapsible: true` and `showBorder: false` kept its
+border and shadow. `DetailSection` ends in three returns and the key was read on
+exactly two of them: the flat branch and the non-collapsible `Card` branch. The
+collapsible branch wrapped a **bare** `<Card>` and never mentioned `showBorder`, so
+the declaration was structurally unreachable there. The collapsible `<Card>` now
+carries the same decision branch 2 already made — `border-none shadow-none` when
+`showBorder === false`.
+
+**Patch, and the level is measured rather than intuited.**
+
+- *Nothing stops rendering.* No node, prop, export or registration moved. The diff
+  is one `className` on one existing `<Card>`.
+- *Nothing stops type-checking.* `showBorder` was already declared on all three of
+  its surfaces — `DetailViewSection` in `@object-ui/types`, its Zod mirror, and the
+  `detail-section` registration's `inputs` — and none of them is touched. This change
+  makes the implementation catch up with a declaration that already shipped.
+- *What an author authored today changes visually, and only in the direction they
+  asked for.* A node carrying `collapsible: true` + `showBorder: false` already
+  passed `sdui-parser`'s `validateTree` clean and then drew a border anyway; it now
+  draws none. The only behaviour that regresses is a border that was present because
+  the key was being dropped, which is the defect itself — so this is not a breaking
+  change and does not need a `**BREAKING**` carrier. A census of this repository
+  found zero authored `detail-section` nodes under `examples/`, `apps/` or `content/`,
+  so no in-repo page changes appearance.
+
+`className` deliberately stays on the outer `Collapsible` where it already was;
+relocating it is a different behaviour that objectui#9218 did not measure, and a
+drift guard now reds if a later change moves it.
--- a/packages/plugin-detail/src/DetailSection.tsx
+++ b/packages/plugin-detail/src/DetailSection.tsx
@@ -675,7 +675,12 @@ export const DetailSection: React.FC<DetailSectionProps> = ({
       onOpenChange={(open) => setIsCollapsed(!open)}
       className={className}
     >
-      <Card>
+      {/*
+        objectui#9218 — the SAME decision branch 2 above makes, and only that
+        one. `className` stays on the outer `Collapsible`: relocating it is a
+        different behaviour, unmeasured by that card.
+      */}
+      <Card className={cn(section.showBorder === false ? 'border-none shadow-none' : '')}>
         <CollapsibleTrigger asChild>
           <CardHeader className={cn(
             "py-3 px-4 sm:py-4 sm:px-6 cursor-pointer hover:bg-muted/50 transition-colors",
--- a/packages/plugin-detail/src/__tests__/DetailSection.collapsibleShowBorder-9218.test.tsx
+++ b/packages/plugin-detail/src/__tests__/DetailSection.collapsibleShowBorder-9218.test.tsx
@@ -1,0 +1,228 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9218 — `DetailSection`'s COLLAPSIBLE branch honours
+ * `section.showBorder`.
+ *
+ * ## What was broken, measured rather than read
+ *
+ * `DetailSection` ends in three returns and `showBorder` appeared in exactly
+ * two of them on `a9d97be63`:
+ *
+ *  1. the flat branch — `!section.title && !section.collapsible &&
+ *     section.showBorder === false` — reads it;
+ *  2. the non-collapsible `Card` branch — `section.showBorder === false ?
+ *     'border-none shadow-none' : ''` — reads it;
+ *  3. the collapsible branch — a `Collapsible` wrapping a BARE `Card` — never
+ *     mentioned it, so the key was structurally unreachable there.
+ *
+ * `showBorder` is declared on three surfaces at once (`DetailViewSection` in
+ * `@object-ui/types`, its Zod mirror, and the `detail-section` registration's
+ * `inputs`), so `sdui-parser`'s `validateTree` told an author the node was
+ * correct and complete — and then the border stayed. Declared clean,
+ * published clean, rendered wrong: ADR-0049 enforce-or-remove shape.
+ *
+ * ## Why every row here is a RENDERING verdict
+ *
+ * Each row reads the class list of the real `Card` element, and each carries
+ * its own negative (`showBorder: true` on the same branch, same shape). A row
+ * asserting only "the prop arrived" would pass on the broken tree, which is
+ * precisely the defect.
+ *
+ * ## The control, and why it can fire
+ *
+ * The `collapsible: false` rows are the same authored node with ONE key
+ * flipped, and they were GREEN on the unmodified tree — branch 2 already read
+ * the key. They are what makes the collapsible rows a statement about the
+ * branch rather than about `showBorder` in general. Each side additionally
+ * asserts WHICH branch it landed on (`[aria-expanded]` present / absent), so
+ * a future change that quietly routes the collapsible node through branch 2
+ * reds here instead of passing silently.
+ *
+ * ## Measured legs
+ *
+ * - PIN-FIRST. This file was written before `DetailSection.tsx` was touched
+ *   and run against the unmodified base tree `a9d97be63`: `2 failed | 4
+ *   passed`. The two rows that failed are the collapsible `showBorder: false`
+ *   verdicts; both `collapsible: false` control rows, the collapsible
+ *   `showBorder: true` negative and the drift guard passed.
+ * - ABLATION. With the fix committed, deleting the new decision from the
+ *   collapsible branch's `<Card>` reddens exactly those same two rows and
+ *   nothing else. Both legs are recorded in the pull request.
+ */
+
+import React from 'react';
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, screen, cleanup } from '@testing-library/react';
+import { SchemaRenderer } from '@object-ui/react';
+import type { DetailViewSection } from '@object-ui/types';
+
+import { DetailSection } from '../DetailSection';
+
+// Module scope, not a hook (AGENTS.md test discipline): importing the package
+// index executes its registration side-effects, so the `detail-section` entry
+// resolved below is the very one production resolves.
+import '../index';
+
+const RECORD = { street: '1 Market St', city: 'San Francisco' };
+
+/**
+ * The node an author writes. A `title` is present on purpose: without one, a
+ * `collapsible: false` + `showBorder: false` node takes the FLAT branch and
+ * renders no `Card` at all, and the control would then be measuring a third
+ * branch instead of branch 2.
+ */
+const authoredNode = (overrides: Record<string, unknown>) =>
+  ({
+    type: 'detail-section',
+    title: 'Billing Address',
+    fields: [
+      { name: 'street', label: 'Street' },
+      { name: 'city', label: 'City' },
+    ],
+    ...overrides,
+  }) as never;
+
+/** The same section, reached the way `DetailView` / `SectionGroup` reach it. */
+const directSection = (overrides: Record<string, unknown>) =>
+  ({
+    title: 'Billing Address',
+    fields: [
+      { name: 'street', label: 'Street' },
+      { name: 'city', label: 'City' },
+    ],
+    ...overrides,
+  }) as unknown as DetailViewSection;
+
+/**
+ * The `Card` element itself, not "somewhere in the subtree". `Card`'s own base
+ * classes are `rounded-lg border bg-card text-card-foreground shadow-sm`, so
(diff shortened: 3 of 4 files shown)


os-tesla:
## Contract review — PR objectui#9284 @ head `fb9a47e98` · ✅ **CLEARED**

| check | reading |
|:--|:--|
| CI at `fb9a47e98` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9284` | ✓ PR #9284 / card #9218 — both carriers agree, declaration readable in the fixed spelling |
| production diff | **one `className` on one existing `<Card>`** — everything else is a comment, a changeset or a test |
| `showBorder` sites | enumerated on **both** refs, not counted: base `a9d97be63` = `:644` (`isFlat`) + `:651` (branch 2). Head = those two **byte-identical** + the new `:683`. ⇒ 2 → 3, and the two that already existed did not move |
| `cn` availability | ✓ already imported (`@object-ui/components`, 9 call sites in this file) — no new import |
| the 8626 pin edit | ✓ **comment-only**, verified by filtering the diff to non-comment lines: nothing survives. The assertions are untouched |
| No-Touch zone | ✓ `packages/components/src/ui/**` not touched |

### ⭐ The one thing I checked rather than took on trust: the two expressions are *not* identical, and that is correct

Branch 2 reads `cn(section.showBorder === false ? 'border-none shadow-none' : '', className)`.
The new one reads `cn(section.showBorder === false ? 'border-none shadow-none' : '')` — **without** `className`.

That is the same **decision** with the right difference, not drift: on the collapsible branch `className` already sits on the outer `Collapsible`, so repeating it on the inner `Card` would apply the author's class **twice**. The report said "the same decision branch 2 makes"; the diff says something slightly more precise, and the more precise thing is the correct one. ⭐ And you then pinned the placement with a drift guard, so a later change that "tidies" the class onto the Card reddens instead of silently double-applying it.

### ⭐ The control that could have measured a third branch, and doesn't

> A `title` is present on purpose: without one, a `collapsible: false` + `showBorder: false` node takes the FLAT branch and renders no `Card` at all, and the control would then be measuring a third branch instead of branch 2.

This is the failure mode that makes most three-branch pins hollow, caught **before** the run rather than explained after it. Three further pieces of the same discipline:

- `cardOf` asserts the `.bg-card` match is **unique**, so the reading cannot silently become about a second card.
- Every row asserts **which branch it landed on** via `[aria-expanded]`, so the control moves along the `showBorder` axis and stands still only along the `collapsible` axis — which is exactly what a control has to do to be one.
- Each positive arm is paired with its own negative on the same branch and the same shape (`shadow-sm` present ⇔ absent), so a row cannot pass by rendering nothing.

### ⭐ Pin-first on the real base tree, and the base was proved before the run

`git diff HEAD` empty, blob `ef5b78590` equal to the HEAD blob, `showBorder` occurring twice — *then* `2 failed | 4 passed`. That is the standard working as intended: the "unmodified" arm is the real tree, not a restored one.

The ablation from the committed fix is the belt to that's braces, and it is proved the right way round: marker `border-none shadow-none` 2 → 1, blob `499f9078…` → `359aa658…` **on disk before the run was believed**, then restore proved **by state** (blob back, `git diff HEAD` empty, marker back to 2) — ⛔ never by an exit code.

### The coordination clause fired correctly

objectui#8626 **has** landed (`cc00c5943` / PR #9221), so retiring the second reason recorded beside that pin's `showBorder` row was in scope. ⭐ Retiring it **in place, with the history written down** ("It originally carried a second reason… objectui#9218 has since closed it") rather than deleting the paragraph is what keeps the next reader from re-deriving the same question.

### Changeset

`patch`, and the level is argued on the three axes that actually decide it — nothing stops rendering, nothing stops type-checking (`showBorder` was already declared on all three surfaces and none is touched), and the only behaviour that regresses is *the border that was present because the key was being dropped*, i.e. the defect itself. ⛔ `major` is unavailable by construction (41 packages, one `fixed` group). The zero-authored-node census supports "no in-repo page changes appearance".

### NOT MEASURED, and correctly declared as such

`pnpm check` needs `packages/cli/dist/cli.js`, which is absent here; the build that would have supplied it **never acquired the verify lock** (queue-timeout 99, waited 540s behind the objectui#9196 seat). ⇒ a **prerequisite failure, not a red gate** — and this diff carries no metadata or schema JSON, which is what that command reads. `check:sdui-registration-pins`' first run was the same shape (exit 2, PREREQUISITE NOT MET) and was re-run green after building the console closure rather than reported as a pass. Both readings are right.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9284 and card objectui#9218 — each read back.


---

PR #9236, opened 2026-09-12
fix(plugin-list,types): a stray `groupBy` no longer overrides the kanban lane, and is refused by name

os-tesla:
Part of #8365

Implemented by the `domain:ui` `os-dev` seat, session `session_01UzHd6hDYatoDn17BuwKxnZ` (recorded in prose as well as in the footer, because a body edit downgrades the footer's session form).

## What this repairs

`ListView`'s kanban branch destructured `columns` / `groupByField` / `groupField` / `cardFields` / `titleField` out of the merged kanban config and spread the **rest** *after* its own `groupBy: laneField`. A `groupBy` surviving in that bag therefore **overrode the lane the branch had just resolved**.

Measured on the card's distinguishing fixture, not reasoned:

```
options.kanban = { groupBy: 'LANE_FROM_STRAY_GROUPBY' }
kanban         = { groupByField: 'LANE_FROM_CANONICAL' }
=> node.groupBy was LANE_FROM_STRAY_GROUPBY
```

It read as latent only because the single producer that fed it wrote both spellings with the same value by construction, so no fixture driven by that producer could tell the two apart.

## The ruling this implements — option B, not re-opened

Maintainer ruling of 2026-09-12, decision batch #117 item 5, verbatim 「8365 同意」. Option A (strip the key and re-group in silence) was the fallback for a measured zero of stored views and is **not** taken. Honouring `groupBy` as a declared alias was never on the table.

1. **The canonical lane wins.** `groupBy` joins the destructure in `packages/plugin-list/src/ListView.tsx`, so the stray key can no longer reach the passthrough spread.
2. **The stray key is refused loudly, by name, pointing at `groupByField`.**

## Where the refusal is — the part the ruling flagged, answered explicitly

**The read door is the view-level `KanbanConfig` mirror**, `packages/types/src/zod/objectql.zod.ts`. The key is now a **declared, unwritable member** (`aliasKeyRefusal`, the helper this repo already uses for the `chartType` alias), so it is refused instead of riding the object's `.passthrough()` in silence. The message is the sentence `@objectstack/spec`'s own `strictObject({ aliases })` answers with, quoted down to its surface noun:

```
Unrecognized key(s) on this kanban configuration: `groupBy`. Did you mean `groupBy` → `groupByField`? …
```

That door is where the **author of the view** is standing — every channel that validates a view's metadata reaches it:

- the CLI's `os check` / `os validate` (`safeValidateSchema`),
- the VS Code extension,
[...]

PR #9285, opened 2026-09-12
fix(app-shell): drop the `filter[...]` operator suffix instead of swallowing it into the field name

### LINKED ISSUE #9196: `ObjectView` is a third, drifted reader of the `filter[...]` URL family: an operator suffix is swallowed into the FIELD NAME instead of being dropped
Found while fixing objectui#8883 (dev seat). Filed rather than folded in: different file, different package, and objectui#8883's ruling scopes it to the reference rail.

## The measurement

`ObjectView` is the component behind `/apps/:appName/:objectName` (`packages/app-shell/src/console/AppContent.tsx:999`), and it parses the `filter[...]` URL family with its OWN reader:

```
packages/app-shell/src/views/ObjectView.tsx:2092-2098
    const m = /^filter\[(.+)\]$/.exec(key);
    if (m && m[1] && value !== '') out.push([m[1], '=', value]);
```

The `(.+)` is GREEDY, so an operator-suffixed key does not fail to match — it matches with the suffix swallowed into the FIELD NAME. Both parsers of this URL family were run on the same three inputs, verbatim copies of the shipped functions, in one command:

| input | `drillUrlFilters.parseUrlFilterTriples` (the ADR-0055 `/data` surface) | `ObjectView.urlFilters` |
|---|---|---|
| `filter[owners][contains]=u1` | `[]` — dropped, as its docblock promises | `[["owners][contains","=","u1"]]` |
| lit control `filter[owners]=u1` | `[["owners","=","u1"]]` | `[["owners","=","u1"]]` |
| lit control `filter[amount][gte]=100` | `[["amount",">=","100"]]` | `[["amount][gte","=","100"]]` |

Both non-empty-but-wrong readings and the one empty reading sit beside a lit control in the same run, so neither is an unmeasured cell.

⇒ On `/apps/:app/:object`, `?filter[amount][gte]=100` does not filter by a range and does not fall back to no condition either: it emits a condition on a field named `amount][gte`, which no object declares. What the backend then does with a condition on a nonexistent column is the user-visible symptom, and it is not "the range was ignored".

## Why this is a defect and not a style note

`drillUrlFilters`' own docblock states the contract this diverges from, and states the reason the module exists:

> Keeping both sides in ONE module keeps the `filter[field][op]` operator contract (#1752) from drifting between the code that emits a URL and the code that parses it back.

`ObjectView` is a THIRD reader of the same param family that never joined that module, and it has drifted: it neither implements the operator arm nor rejects it. The `/data` surface's deliberate "ignore an unknown suffix, never silently downgrade to equality" posture is exactly what `ObjectView` fails to do.

## Reachability, measured — latent today, not unreachable

No producer in the repo emits an operator-suffixed `filter[...]` key to this route: `useOpenRecordList.ts:88` navigates to `/:objectName/data` (the correct parser), and `serializeDrillFilterParams` is only reached from there. So nothing in-app walks into it today.

It is reachable by a hand-written or bookmarked URL, and these URLs are an ADDRESSABLE public surface by design — that is the whole premise of ADR-0055 and of #1752's operator contract. A user who copies the range spelling from the `/data` surface onto the plain object route gets a silently wrong query rather than a refusal.

Whether the repair is "make `ObjectView` call `parseUrlFilterTriples`" or "make it REFUSE an operator suffix it cannot execute" is not asserted. The first is the obvious one-reader fix and would give the plain object route range operators it has never had — a behaviour addition, which is why it wants a ruling rather than a patch. objectui#9159 is a different gap on the same family (the `/data` dialect has no is-null operator) and does not cover this.

---

os-tesla:
Fixes #9196

## The defect, stated sharply

`ObjectView` was a THIRD reader of the `filter[...]` URL family, carrying its own
field capture. The capture was GREEDY, so an operator-suffixed key did not fail to
match — it matched with the suffix swallowed INTO THE FIELD NAME. On
`/apps/:app/:object`, `?filter[amount][gte]=100` emitted a condition against a field
literally named `amount][gte`, which no object declares.

The user therefore got a **silently wrong query** — not an ignored parameter, and not
a refusal.

## The ruling this PR implements, and the one it does not

The triage comment on the card split the two candidate repairs across the human floor.

| candidate | this PR |
|---|---|
| make this route call the operator-aware reader | **NOT implemented.** It would add range operators to a route that never had them = widening the accepted set of an addressable public surface. A behaviour addition, not a repair. It stays with the maintainer. |
| refuse / drop the operator suffix the route cannot execute | **implemented.** Pulling back to a posture the `/data` surface already declares in as many words: an unknown operator suffix is ignored, never silently downgraded to equality. |

**Explicit confirmation: candidate one was not implemented.** The evidence is in the
pin, not in this sentence — the `/data` column is asserted on the same inputs in the
same run, and `filter[amount][gte]` still reads `['amount', '>=', '100']` there and
ONLY there. If this change had taught the object route the operator arm, that row
would be indistinguishable from the object-route row; it is not.

## What changed

The route now reads through `drillUrlFilters`, the one module that owns this family,
via a new equality-only arm. Both arms share ONE key grammar whose field slot excludes
brackets, so the swallowing is structurally impossible rather than patched at one call
site. Inventing a fourth posture here is what the card identified as recurrence in
place, so the posture is copied from `/data` rather than designed anew.

### Before and after, on the REAL bytes

Neither column is a hand copy of the shipped function: the probe lifts the regex
literal out of `ObjectView.tsx` on disk and runs the card's three inputs through it.

| input | base tree (`a9d97be63`) | this branch |
|---|---|---|
| `filter[owners][contains]=u1` | `[["owners][contains","=","u1"]]` | `[]` |
| LIT CONTROL `filter[owners]=u1` | `[["owners","=","u1"]]` | `[["owners","=","u1"]]` |
| `filter[amount][gte]=100` | `[["amount][gte","=","100"]]` | `[]` |

After the change the probe reports `no filter[ regex literal in ObjectView.tsx` — the
private reader is gone, which is why the third column is read from the shared module.

**The lit control fires.** Plain equality is byte-identical on both sides and on both
readers; it is the only evidence that ordinary equality was not broken while the suffix
was being dropped, and it would have gone red had the repair over-reached.

### IGNORED is pinned as DISTINCT from DOWNGRADED TO EQUALITY

Two different outcomes, only one correct. `toEqual([])` alone would also pass for a
reader that refused the whole URL, so the pin names which one is meant:

- `expect(triples).toEqual([])` — the suffixed key contributes nothing;
- `expect(triples).not.toContainEqual(['amount', '=', '100'])` — and in particular it
  did not quietly answer the narrower question;
- `expect(fields).not.toContain('amount')` — no condition on that field survives at all;
- `expect(objectRoute('filter[amount][gte]=100&filter[status]=open'))` keeps only
  `['status', '=', 'open']` — the drop is scoped to the key it cannot execute.

### A bounded in-place fix on the sibling arm, declared

Sharing one grammar also tightens the field slot of the `/data` reader. That arm was
differenced against its previous implementation over a 254-case key corpus, both sides
lifted from real file bytes:

```
/data surface: 244 identical readings, 10 changed, over 254 key/value pairs
  CHANGED filter[a[b]        before ["a[b","=","100"]   after null
  CHANGED filter[a[b][gte]   before ["a[b",">=","100"]  after null
  ... 8 more, every one a key whose FIELD NAME contains an opening bracket
DIFFER CONTROL (two deliberately unequal readings compare equal?): false -> expect false
```

All ten deltas are the same defect class as this card — a bracket swallowed into the
field name — and every one is a **narrowing that removes a wrong answer**. Nothing the
surface accepts was widened. The bracket battery in the pin asserts this on both
readers.

### The pin can fire — two ablation legs, prediction stated first

Predicted before either leg ran, then observed:

| leg | predicted red | observed |
|---|---|---|
| A: greedy field capture back in the shared grammar | the three-rows suffixed rows, all three ignored-vs-downgraded cases, the bracket battery and its non-vacuity control | `17 failed / 12 passed` — exactly that set |
| B: `ObjectView.tsx` reverted to the base commit | the two source-scan pins only | `2 failed / 27 passed` — exactly those two |

Each leg's GREEN half is the other's control: leg A leaves the source scan green
(it does not touch `ObjectView.tsx`), leg B leaves every behavioural case green. The
lit control and the `ANCHOR` positive control stayed green in both.

Mutations were proven ON DISK before the runs were believed (marker counts flipped
`1 -> 0` / `0 -> 1`, blob hash `4810e6e0 -> ef88069b`; leg B marker `2 -> 0`, hash
`9fe318c0 -> 6e5d06f3`). Both restores were proven BY STATE, never by an exit code:
blob hash back to the HEAD blob and `git diff HEAD` empty, under a trap using absolute
paths. Final tree: `git diff HEAD` empty, `git status --short` empty.

## Verification

All readings below are from `c9e370c4d`, the final commit on this branch; the working
tree was byte-identical to it for every run (`git diff HEAD` empty).

| run | result |
|---|---|
| `pnpm exec vitest run packages/app-shell/ examples/schema-catalog/` | `Test Files 720 passed (720)`, `Tests 8861 passed / 1 skipped`, lock `VERDICT command-exit 0` |
| `pnpm --filter @object-ui/app-shell run type-check` | exit 0 — both `tsc --noEmit` and `tsc -p tsconfig.test.json`, so the new pin is type-checked too |
| the pin alone | `Test Files 1 passed`, `Tests 29 passed` |
| `pnpm changeset:check` | exit 0 (this is the real script name; other spellings' `254` is NOT MEASURED) |
| `node scripts/check-changeset-presence.mjs` | exit 0 — 2 published source files, 1 changeset |
| `check:control-bytes`, `check:test-path-roots`, `check:new-line-citations`, `check:unreferenced-sources`, `check:self-import`, `check:phantom-deps`, `check:shell-escape-residue` | exit 0 each |
| `eslint --no-inline-config` on the three changed files | 0 errors; `ObjectView.tsx` warnings `170 -> 169` (the change removes one `no-explicit-any` and adds none, measured against the base file through `--stdin`) |
| `node scripts/check-governed-queue-guard.mjs --test` | NOT GOVERNED — 4 paths, none matched |

`examples/schema-catalog/` was run because a package-scoped run is not the blast radius
(objectui#9273). Grep for count-shaped and README prose pins naming either reader found
none, and `drillUrlFilters` is not re-exported from this package's only entry, so the
new export adds no public signature.

**NOT MEASURED:** `check:readme-exports` exits 1 in this container with
`type entry ./dist/index.d.ts is not on disk -- run pnpm build first` for 8 unbuilt
packages. Only this package's dependency closure was built, so that is a prerequisite
miss, not a red gate; it is unrelated to this diff, which touches no README and no
`exports` field.

## Changeset

`patch` on `@object-ui/app-shell`, justified by measurement rather than intuition: no
authored key changes; no public signature is added or removed (internal module, not
re-exported); and the only behaviour deltas are the ten narrowings above plus the object
route dropping a key that previously produced a condition against a field no object
declares. Nothing that worked stops working.

## Acceptance notes

Out-of-scope observations found while measuring this reader. Neither is fixed here.

1. **Filed separately** — the memo key in the same block re-serializes params by hand
   (`map(([k, v]) => k + '=' + v).join('&')`) and re-parses the result, so a value
   containing an ampersand or a plus is mangled before it ever reaches the reader.
   Measured with a lit control: `Acme` survives, `Smith & Sons` becomes `Smith `, `A+B`
   becomes `A B`. A different mechanism from this card's, and fixing it would move the
   plain-equality control this PR depends on, so it is a separate card.
2. **Noted, not filed** — `packages/layout/src/NavigationRenderer.tsx` carries a FOURTH
   copy of the same greedy capture, for nav-item highlight matching. Measured, it fails
   safe: the authored side of the comparison never contains a suffix, the size-and-exact
   -key guard admits only false negatives, and "no highlight" is the semantically right
   answer for a URL asking a different question. Successor named: whoever takes the
   maintainer's candidate-one card lands in this file, because a route that really
   executes range operators would need the nav side to agree on operator spelling. It
   also sits in another package, so joining it to the shared module is a dependency
   -direction question, not a mechanical edit.

Neither of these is addressed here; objectui#9159 remains open and is a different gap
on the same family.

---

### DIFF
--- a/.changeset/9196-objectview-url-filter-operator-suffix.md
+++ b/.changeset/9196-objectview-url-filter-operator-suffix.md
@@ -1,0 +1,40 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Stop the plain object route swallowing a `filter[...]` operator suffix into the
+FIELD NAME (objectui#9196). `ObjectView` was a THIRD reader of this URL family,
+with its own greedy field capture, so `?filter[amount][gte]=100` on
+`/apps/:app/:object` did not fail to match — it matched with the suffix inside
+the field, emitting a condition against a field literally named `amount][gte`
+that no object declares. The user got a silently WRONG query, not an ignored
+parameter and not a refusal.
+
+The route now reads through `drillUrlFilters`, the one module that owns this
+family, via a new equality-only arm (`parseUrlEqualityFilterTriples`). Both arms
+share ONE key grammar whose field slot excludes brackets, so the swallowing is
+structurally impossible rather than fixed at one call site. An operator suffix
+this route cannot execute is DROPPED — the posture the `/data` surface already
+declares in as many words ("an unknown operator suffix is ignored, never
+silently downgraded to equality"). Ignored and downgraded are two different
+outcomes and only one is correct: answering `amount = 100` when the URL asked
+for `amount >= 100` would be a wrong answer wearing a right answer's shape.
+
+⚠️ Range operators are deliberately NOT added to this route. That would widen
+the accepted set of an addressable public surface — a behaviour addition, not a
+repair — and it stays with the maintainer. The operator arm remains on the
+ADR-0055 `/data` surface only, which this change's tests pin on the same inputs
+in the same run.
+
+`patch`, justified by measurement rather than by intuition:
+
+- Nothing an author authored changes — no spec key, no metadata key, no schema
+  property is touched.
+- Nothing stops type-checking for a consumer: `drillUrlFilters` is an internal
+  module, not re-exported from `@object-ui/app-shell`'s only entry (`.` →
+  `dist/index.js`), so the new export adds no public signature.
+- The `/data` reader was differenced against its previous implementation over a
+  254-case key corpus: 244 readings identical, 10 changed, and all 10 are keys
+  whose FIELD NAME contains a `[` (`filter[a[b][gte]`), which previously emitted
+  a condition against a field named `a[b`. Every delta is a narrowing that
+  removes a wrong answer; none widens what the surface accepts.
--- a/packages/app-shell/src/views/ObjectView.tsx
+++ b/packages/app-shell/src/views/ObjectView.tsx
@@ -15,6 +15,7 @@ import { resolveFilterPlaceholders, DENSITY_MODE_TO_ROW_HEIGHT, normalizeListVie
 import { parseUserFilterParams, applyUserFilterParams } from './userFilterUrlState.js';
 import { buildListFilterKey, readListFilterState, writeListFilterState } from './listFilterStorage.js';
 import { VALUELESS_FILTER_OPERATORS } from './viewFilterFold.js';
+import { parseUrlEqualityFilterTriples } from './drillUrlFilters.js';
 import { narrowPersonalizationOverlay, isViewConfigPermissionDeniedError } from '@object-ui/data-objectstack';
 const ObjectChart = lazy(() =>
   import('@object-ui/plugin-charts').then((m) => ({ default: m.ObjectChart })),
@@ -2086,6 +2087,17 @@ function ObjectViewInner({ dataSource, objects, onEdit, externalRefreshKey }: an
      * to a single parent record. Emitted as ObjectQL triples (`[field, '=', value]`)
      * which matches the shape consumed by the list view's data fetcher when
      * merging base filters.
+     *
+     * Read through `drillUrlFilters` — the ONE module that owns this URL family
+     * — rather than a private regex here (objectui#9196). This route implements
+     * the EQUALITY arm only: an operator suffix it cannot execute
+     * (`?filter[amount][gte]=100`) is DROPPED, never downgraded to equality and
+     * never swallowed into the field name. The greedy capture this replaced did
+     * the last of those, emitting a condition against a field literally named
+     * `amount][gte` that no object declares — a silently wrong query, not an
+     * ignored parameter. Range operators live on the ADR-0055 `/data` surface
+     * (`parseUrlFilterTriples`); giving them to this route would widen an
+     * addressable public surface and is deliberately not done here.
      */
     // Dep on the serialized `filter[...]` entries only — `uf_*` user-filter
     // params also live in the URL and must not invalidate this memo (a new
@@ -2094,16 +2106,10 @@ function ObjectViewInner({ dataSource, objects, onEdit, externalRefreshKey }: an
         .filter(([k]) => k.startsWith('filter['))
         .map(([k, v]) => `${k}=${v}`)
         .join('&');
-    const urlFilters = useMemo(() => {
-        const out: Array<[string, string, any]> = [];
-        new URLSearchParams(filterParamsKey).forEach((value, key) => {
-            const m = /^filter\[(.+)\]$/.exec(key);
-            if (m && m[1] && value !== '') {
-                out.push([m[1], '=', value]);
-            }
-        });
-        return out;
-    }, [filterParamsKey]);
+    const urlFilters = useMemo(
+        () => parseUrlEqualityFilterTriples(new URLSearchParams(filterParamsKey)),
+        [filterParamsKey],
+    );
 
     /**
      * End-user filter selections restored from `uf_*` URL params (ADR-0047
--- a/packages/app-shell/src/views/ObjectView.urlFilterSuffix-9196.test.ts
+++ b/packages/app-shell/src/views/ObjectView.urlFilterSuffix-9196.test.ts
@@ -1,0 +1,212 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9196 — `ObjectView` was a THIRD, drifted reader of the `filter[...]`
+ * URL family, and its drift was NOT "the operator was ignored".
+ *
+ * The reader it carried had a GREEDY field capture, so an operator-suffixed key
+ * did not fail to match — it matched with the suffix swallowed INTO THE FIELD
+ * NAME. On `/apps/:app/:object`, `?filter[amount][gte]=100` emitted a condition
+ * against a field literally named `amount][gte`, which no object declares: a
+ * silently WRONG query, not an ignored parameter and not a refusal.
+ *
+ * ## What this file pins, and what it deliberately does not
+ *
+ * The ruling on the card takes ONE of the two candidate repairs:
+ *
+ *   NOT taken — teaching this route the operator arm. That would ADD range
+ *      operators to a route that has never had them, widening the accepted set
+ *      of an addressable public surface. It is a behaviour addition, and it
+ *      belongs to the maintainer.
+ *   Taken — the route drops the operator suffix it cannot execute, which is the
+ *      posture the `/data` surface already DECLARES in as many words: "An
+ *      unknown operator suffix is ignored (never silently downgraded to
+ *      equality)."
+ *
+ * The route now reads through `drillUrlFilters`, the one module that owns this
+ * family, so there is no longer a private reader here to drift. Both arms share
+ * ONE key grammar, which is what makes the swallowing structurally impossible
+ * rather than merely fixed at one call site.
+ *
+ * ## SUITE DIRECTION — stated before running
+ *
+ *   RED on the unmodified base tree: every assertion in "the card's three rows"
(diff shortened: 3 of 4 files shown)


os-tesla:
## Contract review — PR objectui#9285 @ head `c9e370c4d` · ✅ **CLEARED**

| check | reading |
|:--|:--|
| CI at `c9e370c4d` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9285` | ✓ PR #9285 / card #9196 — both carriers agree |
| ruling compliance | ✓ **candidate 2 only.** Read in the diff, not taken from the report: `parseUrlFilterTriples` keeps its operator arm and `parseUrlEqualityFilterTriples` returns early on a suffix. No range operator reaches `/apps/:app/:object` |
| the swallowing is structural now | ✓ one `FILTER_KEY` grammar, field slot `[^[\]]+` — excludes **both** brackets, so a suffix cannot enter a field name on **either** arm |
| `drillUrlFilters` is internal | ✓ **verified independently**: absent from `packages/app-shell/src/index.ts`, and the package's only entry is `"."` → `dist/index.js`. ⇒ the new export adds no public signature, which is what the `patch` level rests on |

### ⭐ The distinction this card turns on, and the diff makes it in the code rather than only in prose

> An operator suffix this route cannot execute is **DROPPED** … answering `amount = 100` when the URL asked for `amount >= 100` would be a wrong answer wearing a right answer's shape.

⇒ and the same sentence is written at the call site, in the module, **and** in the changeset — three readers, one statement, no room for the next seat to "helpfully" downgrade it. ⭐ That posture was **copied from `/data`'s existing declaration** rather than invented, which is the whole reason there is no fourth posture in this repo now.

### ⭐ One second-order delta I measured myself, because the single grammar is not a pure narrowing

The operator slot widened: `[a-z]+` → `[^[\]]+`. So `?filter[amount][GTE]=100` now **matches** `FILTER_KEY` where before it matched **neither** old regex.

The outcome is unchanged — `URL_FILTER_OPS['GTE']` is `undefined`, so the triple is dropped, exactly as the non-matching key was dropped before. ⇒ same answer by a different route. I record it because it is the one place the new grammar **accepts more keys** than the old pair, and "accepts more keys" is the shape that usually hides a widening. It does not turn into more triples here, and the module's docblock already declares "an unknown operator suffix is ignored", so the behaviour is the declared one.

### ⭐ The differential is the right instrument, and it read the real bytes

254 key/value pairs, **244 identical, 10 changed**, every one a key whose FIELD NAME contains a `[` — each a narrowing that deletes a wrong answer, none a widening. ⭐ And the probe **lifted the regex literal out of `ObjectView.tsx` on disk** instead of a hand-copied pattern, so the before-side is the shipped code and not a transcription of it. The differ carried its own control (two deliberately unequal readings compared unequal), so "no difference" would have been a reading.

### ⭐ Two ablation legs whose green halves are each other's controls

Leg A (greedy capture back in the shared grammar): predicted red set written **before** the run, observed `17 failed | 12 passed` = exactly that set, with the lit control and the whole source-scan block staying green.
Leg B (`ObjectView.tsx` reverted to base): `2 failed | 27 passed`, exactly the two source-scan pins, anchor positive control green.

Mutations proved **on disk** (marker counts and blob hashes both ways) before either run was believed; restores proved **by state**, ⛔ never by an exit code. ⚠️ And the note on why the placeholder is spelled as a word — a bracket-shaped token is eaten by GitHub's body sanitiser — is a real hazard correctly worked around rather than a cosmetic choice.

### ⭐ The in-flight check had a control that lit

All 14 open PRs enumerated (including the 1147-file Changesets release PR, fully paginated): zero touch `views/ObjectView.tsx` or `views/drillUrlFilters*`. ⭐ **Control leg: the same channel and matcher run against PR #9236, known to touch `ObjectView.tsx`, returned the hit.** ⇒ the empty result is a reading. Same for the 177-branch remote glob with its own lit control.

### The three out-of-scope findings are all correctly placed

- **objectui#9287 filed, not folded in.** ⭐ The reason is the sharp one: repairing the `&`/`+` memo-key truncation **would move the plain-equality control this card's acceptance depends on**. Fixing it here would have quietly invalidated the instrument.
- **`NavigationRenderer.tsx:719`, a fourth copy, noted and not filed** — measured to fail **safe** (no failing probe with a wrong user-visible outcome can be named), and the bounded in-place exemption correctly failed on condition 4, new verification surface. A named successor rather than a dangling observation.
- **The silent drop with no chip**, inherited from the `/data` posture this card was told to copy. ⛔ Inventing a refusal UI here would have been the fourth posture. Right call.

### Changeset

`patch`, argued on the three axes that decide it and with the internal-module claim I re-measured above. ⛔ `major` unavailable by construction. `pnpm changeset:check` exit 0 — the **real** script name, and the dispatch's warning about `no-major` held.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9285 and card objectui#9196 — each read back.


---

PR #9286, opened 2026-09-12
test(plugin-dashboard): a drawn dashboard-seam pin for the dataset face's five plot-internal chartConfig keys

### LINKED ISSUE #9203: finding(plugin-dashboard): the DATASET face's five plot-internal chartConfig keys have no dashboard-surface drawn assertion — their only "evidence" is a pin that stays green when the forwarding is deleted
Found while delivering objectui#4044's ruling for the two inline relays (PR objectstack-ai/objectui#9202). Filed rather than folded in: that card's DO-NOW is the **not-yet-forwarded** keys, the relays are what was unforwarded, and extending a pin file mid-flight risks a claim collision. ⛔ Not claimed, ⛔ not graded — labels and priority are triage's.

## The gap

`DatasetWidget` has forwarded `chartConfig` through `@object-ui/core`'s `chartConfigPresentation` whitelist since objectstack#7016 (2026-08-09). Its dashboard-surface pin, `DatasetWidget.chartConfig.dom.test.tsx`, asserts only **four** keys are drawn: `title`, `subtitle`, `description`, `height`.

The other five — **`colors`, `categoryColors`, `showDataLabels`, `annotations`, `interaction`** — have **no drawn assertion on the dashboard face at all**.

## Why the existing coverage does not count — measured, not argued

Their only apparent evidence is `plugin-charts`' hand-built `ChartRenderer.dashboardChartConfig.test.tsx`. PR objectui#9202's seat ablated the forwarding out of **both** inline relays and ran it:

> 46 assertions red across the three dashboard files, and **0 in `plugin-charts`' file**.

⇒ that file pins the **chart block**, never the dashboard seam. An assertion that stays green while the thing it names is deleted is not evidence — it is the exact shape objectui#7963's `confirmVariant` survived in for months, and the shape objectui#4044 exists to remove one layer up.

⭐ The same PR demonstrates the gap is closable: it added `DashboardChart.chartConfigMarks-4044.test.tsx`, a **drawn** pin for those same six plot-internal keys on the relay face, and every one reddens under its own ablation leg.

## The technique transfers verbatim, and it corrects a false claim in the tree

A comment in `DatasetWidget.chartConfig.dom.test.tsx` (and its twin, which objectui#9202 corrected in the file it owns) asserts the recharts marks **can only** be pinned inside `plugin-charts`.

- The **premise is true**, re-measured: `require.resolve('recharts')` from `packages/plugin-dashboard` is `MODULE_NOT_FOUND`.
- The **conclusion is false**: `ResponsiveContainer` seeds its size from `getBoundingClientRect` on its own element, so a scoped stub on the `recharts-responsive-container` element is enough — ⛔ no module mock required.

So the dataset-face pin can be extended the same way; it is a mechanical extension of one existing file rather than new ground.

## Scope note

This is a **coverage gap, not a defect** — the five keys do reach the renderer on the dataset path; nothing an author writes is silently dropped there. What is missing is the assertion that would catch it if that stopped being true. objectui#4044's ruling names the pin clause for this face too (「每个转发键在 dashboard 面有一个读到它的渲染断言」), so this is that clause's unfinished half.

Refs: objectui#4044 (the ruling and its pin clause) · PR objectstack-ai/objectui#9202 (the relay half, where the 0-of-46 reading was taken) · objectstack#7016 (which made the dataset path forward the whitelist).

os-tesla:
Fixes #9203

The DATASET face forwards nine `chartConfig` keys through `@object-ui/core`'s
`chartConfigPresentation`. Four of them paint outside Recharts'
`ResponsiveContainer` and are pinned on the dashboard seam by
`DatasetWidget.chartConfig.dom.test.tsx`. The other five — `colors`,
`categoryColors`, `showDataLabels`, `annotations`, `interaction` — paint inside
it and had no drawn assertion on this surface at all.

This adds one: `packages/plugin-dashboard/src/__tests__/DatasetWidget.chartConfigMarks-9203.test.tsx`,
10 assertions over dataset widget metadata in and drawn Recharts marks out, with
no renderer stub anywhere in the chain. The technique is PR #9202's, transferred
verbatim: `ResponsiveContainer` seeds its size from `getBoundingClientRect` on
its OWN element, so a stub scoped to the `recharts-responsive-container` element
gives the plot a measured box — no module mock, which is what made this look
impossible from `plugin-dashboard`.

## The red, demonstrated

⭐ A new pin that has never been seen red is worth exactly as much as the zero
coverage it replaces, so here is the ablation. The forwarding call in
`packages/plugin-dashboard/src/DatasetWidget.tsx` was replaced by an empty
object, the mutation was proved on disk before the run, and the restore was
proved by state afterwards.

| leg | reading |
|:--|:--|
| mutation anchor count | 1 to 0 |
| injected marker count | 0 to 1 |
| blob hash on disk | `1203c0e9…` to `e02ab1c4…` |
| restore: blob hash | back to `1203c0e9…`, equal to the HEAD blob |
| restore: `git diff HEAD` | empty |

With the forwarding deleted, across the three files in one run:

| file | result |
|:--|:--|
| `DatasetWidget.chartConfigMarks-9203.test.tsx` (new) | 10 tests, **7 failed** |
| `DatasetWidget.chartConfig.dom.test.tsx` (lit control, untouched assertions) | 7 tests, **3 failed** |
| `plugin-charts/src/ChartRenderer.dashboardChartConfig.test.tsx` | 13 tests, **0 failed** |

Every one of the five keys reddens: the `colors` palette arm, the
`categoryColors` record arm, `showDataLabels`, both `annotations` arms and both
`interaction` arms. The three that stay green in the new file are its pure
absence arms — "no data labels when off", "no reference marks when none
declared", "the default palette" — which is the correct behaviour for an
absence assertion when the thing is absent for a second reason, and why each is
paired with a positive arm taken through the same harness.

The third row is the point of the card, now measured on this face: 13 assertions
that name these keys did not move while the thing they name was deleted. PR
#9202 took the same reading on the relay face (46 red across the dashboard
files, 0 there). ⛔ That file is not coverage for these keys on any dashboard
surface.

Restoring the forwarding returns all three files to green (30 passed).

## The comment the card asked for, and two more like it

The acceptance names the header in `DatasetWidget.chartConfig.dom.test.tsx`
that said these marks could only be pinned inside `plugin-charts`. Grepping the
citation found two more making the same claim about the same file, and the
ablation above measured all three false in one run, so all three are corrected:

- `DatasetWidget.chartConfig.dom.test.tsx` — the one named in the acceptance.
  ⭐ Its assertions are deliberately untouched: they are the lit control above.
- `DatasetWidget.chartConfig.test.tsx` (dataset seam pin) cited `plugin-charts`'
  file as the DOM half for the keys drawn inside the plot. It now cites the new
  pin.
- `plugin-charts/src/ChartRenderer.dashboardChartConfig.test.tsx` claimed the
  mock "and therefore this half of the evidence" had to live there, and that it
  plus the dataset seam pin "close the loop from dashboard metadata to drawn
  pixels". Its objectui#4044 block already disclaimed the two inline relays; it
  now disclaims the dataset face on the same measured footing.

All three edits are comment-only — no assertion, import or fixture moves in any
of them.

## The card's `recharts` premise: re-measured, and the voided probe explained

Triage recorded a VOIDED probe — `require.resolve('recharts')` answered
`MODULE_NOT_FOUND` from `packages/plugin-charts` too, so the lit control did not
light and the premise was NOT MEASURED. The container had been restarted and
`node_modules` was absent repo-wide, which makes every `require.resolve` fail
identically. Both halves were checked:

| root | `recharts` | `react` (control) |
|:--|:--|:--|
| shared checkout, no install | `MODULE_NOT_FOUND` | `MODULE_NOT_FOUND` |
| this worktree, installed — `plugin-charts` | resolves | resolves |
| this worktree, installed — `plugin-dashboard` | `MODULE_NOT_FOUND` | resolves |

The first row reproduces triage's dead instrument exactly, so the missing
install does explain it. The last two rows carry two controls that fire: the
`plugin-charts` row is the lit control triage wanted, and `react` resolving from
`plugin-dashboard` shows the resolver was answering at that root at all. ⇒ the
card's premise is **confirmed**: `recharts` genuinely does not resolve from
`packages/plugin-dashboard`. Its conclusion stays false, and this PR is the
counter-example.

## Changeset

Needed, and declared as releasing nothing. `scripts/check-changeset-presence.mjs`
guards everything under a released package's `src/`, and all four touched files
live there — it exits 1 on this diff without a declaration (measured before
adding one). Nothing published changes behaviour: every file is a test and every
edit outside the new file is a comment. So the empty-frontmatter form, which the
gate's own output calls "the explicit exemption and a complete answer". It now
prints: 4 source files of 2 released packages changed, 1 changeset declared.

## Verification

Run at `63f17c236`, from the repo root.

- `pnpm exec vitest run` over the four touched test files — 4 files, 44 passed.
- `turbo run type-check lint` filtered to `@object-ui/plugin-dashboard` and
  `@object-ui/plugin-charts` — 18 tasks successful, 0 errors. Warning counts are
  the packages' pre-existing ones; the dashboard package's `type-check` is
  `tsc --noEmit && tsc -p tsconfig.test.json`, so the second leg is what sees
  the new file.
- 15 gate scripts, all exit 0: the five changeset gates
  (`presence`, `claims`, `fixed`, `no-major`, `overwrite`), `control-bytes`,
  `new-cross-file-line-citations`, the three `vi-mock` gates, `test-path-roots`,
  `unreferenced-sources`, `entry-guard`, `lint-coverage`, `type-check-coverage`.
  Derived by hand from `package.json` and `.github/workflows/` — this repo has no
  `scripts/pm/dispatch-gates.mjs`.
- Blast radius beyond the changed packages: grepped the tree for every file
  naming these three pins; the only ones are the pins themselves and their
  siblings inside the two packages already run above. No census or prose figure
  counts files in `packages/plugin-dashboard/src/__tests__/`.
- ⚠️ Repo-wide `pnpm lint`, `pnpm test` and the full shard matrix are CI's run,
  not measured here.

The ablation and the resolver probe were one-shot proofs; no script from either
is committed.

---

### DIFF
--- a/.changeset/dataset-chartconfig-seam-pin-9203.md
+++ b/.changeset/dataset-chartconfig-seam-pin-9203.md
@@ -1,0 +1,8 @@
+---
+---
+
+Test-only change (objectui#9203): adds a drawn dashboard-seam pin for the five
+plot-internal `chartConfig` keys the DATASET face forwards — `colors`,
+`categoryColors`, `showDataLabels`, `annotations`, `interaction` — and corrects
+the header comment in the sibling pin that pointed at `plugin-charts` as their
+coverage. No published behaviour changes; both files are tests.
--- a/.changeset/dataset-chartconfig-seam-pin-9203.md
+++ b/.changeset/dataset-chartconfig-seam-pin-9203.md
@@ -4,5 +4,6 @@
 Test-only change (objectui#9203): adds a drawn dashboard-seam pin for the five
 plot-internal `chartConfig` keys the DATASET face forwards — `colors`,
 `categoryColors`, `showDataLabels`, `annotations`, `interaction` — and corrects
-the header comment in the sibling pin that pointed at `plugin-charts` as their
-coverage. No published behaviour changes; both files are tests.
+the three header comments that cited `plugin-charts`'
+`ChartRenderer.dashboardChartConfig.test.tsx` as that face's coverage. No
+published behaviour changes; every file touched is a test.
--- a/packages/plugin-charts/src/ChartRenderer.dashboardChartConfig.test.tsx
+++ b/packages/plugin-charts/src/ChartRenderer.dashboardChartConfig.test.tsx
@@ -16,8 +16,17 @@
  * a DOM pin — and the marks below (bars, LabelList, ReferenceLine/Area, Brush)
  * only exist once Recharts has a measured box. `ResponsiveContainer` reports 0×0
  * under the headless DOM and renders no children, and `recharts` resolves inside
- * THIS package alone, so the mock that fixes its size — and therefore this half
- * of the evidence — has to live here.
+ * THIS package alone, so a `vi.mock('recharts')` can only be written here.
+ *
+ * ⚠️ That premise is true and its old conclusion — "and therefore this half of
+ * the evidence has to live here" — was not. objectui#9203 re-measured both:
+ * `require.resolve('recharts')` from `packages/plugin-dashboard` is still
+ * `MODULE_NOT_FOUND` (with `react` from the same root resolving, as the control
+ * that the resolver was answering at all), and a module mock turned out not to
+ * be needed. `ResponsiveContainer` seeds its size from `getBoundingClientRect`
+ * on its OWN element, so a stub scoped to the `recharts-responsive-container`
+ * element gives the plot a box inside plugin-dashboard. See "What this file is
+ * NOT" below for why that distinction decides what this file may be cited for.
  *
  * These render `ChartRenderer`, not `AdvancedChartImpl`: `ChartRenderer` is what
  * the ComponentRegistry resolves `type: 'chart'` to, so it is the component the
@@ -25,22 +34,37 @@
  * shape `DatasetWidget` emits (derived `chartType`/`xAxisKey`/`series` +
  * `isAnimationActive: false` + the lowered presentation keys). The seam that
  * produces it is pinned in plugin-dashboard's
- * `DatasetWidget.chartConfig.test.tsx`; together the two close the loop from
- * dashboard metadata to drawn pixels.
+ * `DatasetWidget.chartConfig.test.tsx`.
  *
- * ⭐ What this file is NOT (objectui#4044). Since that card the two INLINE
- * dashboard relays (`DashboardRenderer` and `DashboardGridLayout`, for widgets
- * bound to inline rows or to a `provider: 'object'` aggregate rather than to an
- * ADR-0021 dataset) lower the same keys through the same
- * `chartConfigPresentation` whitelist onto a node of the same shape. It is
- * tempting to read the assertions below as covering those relays too. They do
- * not, and the difference was measured: with the forwarding deleted from BOTH
- * relays, every test in this file still passed — because the schema above is
- * hand-built here rather than composed by a relay. What this file pins is the
- * CHART BLOCK: that a node carrying these keys draws them. The dashboard
- * surface pins its own end of the chain, in plugin-dashboard's
- * `DashboardChart.chartConfig-4044.test.tsx` (the seam) and its two
- * end-to-end siblings `…chartConfigDom-4044` and `…chartConfigMarks-4044`.
+ * ⛔ Those two do NOT close the loop from dashboard metadata to drawn pixels,
+ * and this file's own header used to say they did. "Byte-for-byte the shape
+ * `DatasetWidget` emits" is a resemblance an author maintains by hand, not a
+ * composition — nothing below ever calls `DatasetWidget`. See the next block.
+ *
+ * ⭐ What this file is NOT (objectui#4044, objectui#9203) — ⛔ it is not
+ * coverage for ANY dashboard face, and that has now been measured on both.
+ *
+ * Three dashboard surfaces lower the same keys through the same
+ * `chartConfigPresentation` whitelist onto a node of the same shape: the two
+ * INLINE relays (`DashboardRenderer` and `DashboardGridLayout`, for widgets
+ * bound to inline rows or to a `provider: 'object'` aggregate) and the DATASET
+ * face (`DatasetWidget`, for an ADR-0021 dataset). It is tempting to read the
+ * assertions below as covering them. They do not, and both readings are
+ * ablations rather than arguments — with the forwarding deleted, the dashboard
+ * files reddened and every test in THIS file still passed, for the relays under
+ * objectui#4044 (PR #9202) and for the dataset face under objectui#9203.
+ *
+ * The reason is structural, so it will not change: the schema above is
+ * hand-built in this file rather than composed by a dashboard surface, so no
+ * dashboard forwarding is on the path these assertions execute. ⭐ An assertion
+ * that stays green while the thing it names is deleted is not evidence of it.
+ * What this file pins is the CHART BLOCK: that a node carrying these keys draws
+ * them. Each dashboard surface pins its own end of the chain, in
+ * plugin-dashboard: `DashboardChart.chartConfig-4044.test.tsx` (relay seam)
+ * with `…chartConfigDom-4044` and `…chartConfigMarks-4044`, and
+ * `DatasetWidget.chartConfig.test.tsx` (dataset seam) with
+ * `DatasetWidget.chartConfig.dom.test.tsx` and
+ * `DatasetWidget.chartConfigMarks-9203.test.tsx`.
  */
 
 import React from 'react';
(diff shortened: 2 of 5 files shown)


os-tesla:
## Contract review — PR objectui#9286 @ head `63f17c236` · ✅ **CLEARED**

| check | reading |
|:--|:--|
| CI at `63f17c236` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9286` | ✓ PR #9286 / card #9203 — both carriers agree, **and the diff carries no widening tell**. ⚠️ A tell is not a proof and its absence is not one either |
| production code | **none** — 1 changeset, 1 new pin, 3 test files touched |
| the claim the whole PR rests on | ✓ **verified independently**: I filtered the diff of all three modified test files to non-comment added/removed lines. **Nothing survives.** The assertions are untouched, so the lit control really is lit |

### ⭐ That last row is the one that mattered, so I measured it rather than read it

The PR's argument is: *the four existing keys' assertions in `DatasetWidget.chartConfig.dom.test.tsx` are the lit control, untouched on purpose.* If a single assertion had moved in the same commit that added the new pin, the control would be worthless and the ablation numbers uninterpretable. Filtering the three-file diff down to non-comment lines returns **empty**. ⇒ the control stands.

### ⭐ The pin measures drawn marks, not a stub agreeing with itself

Ten assertions on real Recharts output — `.recharts-surface` awaited, sector fills, data labels, `.recharts-reference-line`, `.recharts-reference-area`, `.recharts-brush`, `.recharts-tooltip-wrapper` — with **no `vi.mock('recharts')`** anywhere in the file. ⭐ And every positive arm is paired with its own **absence** arm (`showDataLabels` off *and* undeclared; no annotation; no brush by default; tooltips false), so no row can pass by rendering nothing.

The technique is PR objectui#9202's, transferred verbatim rather than reinvented: a `getBoundingClientRect` stub scoped to the responsive-container element, because `recharts` resolves inside `plugin-charts` alone and a module mock is not available from this package. ⭐ That constraint is stated in the header as the *reason* for the technique, which is what lets the next seat reuse it correctly.

### ⭐ The voided probe was re-measured instead of reported as a negative

Triage's `require.resolve` reading was dead because the shared checkout has **no `node_modules` at all** after the restart — so it fails identically from every root, which is an **instrument failure, not a negative answer**. ⭐ You reproduced triage's dead reading first, *then* re-measured in the installed worktree with **two controls that fire** (`recharts` from `plugin-charts`, `react` from `plugin-dashboard`). That is exactly the standard: an unlit control means NOT MEASURED. The card's premise is confirmed; its conclusion stays false, and both halves are stated.

### The three corrected headers — I agree with taking all three in place

The acceptance named one. Grepping the citation found two more asserting the same false thing about the same file, including one in `plugin-charts`' own header claiming it plus the dataset seam pin "close the loop from dashboard metadata to drawn pixels". ⭐ The ablation measured **all three false in the same run** (13 tests in that file, **0 red**) — so this is not three judgement calls, it is one measurement with three citations attached.

Bounded in-place rule checked out: same defect class, comment-only, no open PR on either file (all 14 enumerated), inside the gate family this PR already runs. ⇒ Leaving two live pointers that send the next reader to a file which does not cover the face would have been the worse outcome.

### The ablation

Pin written and **committed first** (`87fabe590`), so the unmodified arm is the real base tree. Mutation proved **on disk** before the run — anchor 1 → 0, marker 0 → 1, blob `1203c0e9…` → `e02ab1c4…`, with the script aborting on a non-unique anchor. Result across three files in one run: new pin `7 of 10` red, lit control `3 of 7` red, `plugin-charts` file **0 of 13** — ⭐ that last number reproduces the card's central evidence on this face, in the same run as the two that do move. The three that stay green in the new file are its pure-absence arms, each paired with a positive arm through the same harness. Restore proved **by state**, under a trap, with absolute paths and an explicit `git checkout HEAD -- <abs>` — ⛔ never a bare checkout in a shared checkout, and ⛔ never an exit code as the proof.

### Changeset

An **empty** changeset (no package, no bump) is the right shape here: `check-changeset-presence` is satisfied, and nothing claims a release for a change where every touched file is a test. ⛔ Declaring a `patch` on `@object-ui/plugin-dashboard` would have been a false claim about published behaviour.

### The second out-of-scope note is the more interesting one

> nothing in this repo mechanically catches the defect class this card is an instance of — an assertion that stays green when the thing it names is deleted.

⇒ correctly **not** filed as a new card, and routed to objectui#4044, the ruling one layer up that already owns the class. Mutation testing is the general answer and it is far past this card.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9286 and card objectui#9203 — each read back.


---

PR #9289, opened 2026-09-12
fix(react)!: `RecordContextValue.dataSource` is the adapter it holds, not an id

### LINKED ISSUE #9197: `RecordContextValue.dataSource` is declared `string` while every producer passes a DataSource ADAPTER — a published type that every consumer has to cast away
Found while fixing objectui#8883 (dev seat). Filed rather than folded in: the repair touches `@object-ui/react`'s published context type and every consumer that casts around it, which is a different package and a wider blast radius than that card's fence.

## The contract, quoted

`packages/react/src/context/RecordContext.tsx:112-118`:

```ts
export interface RecordContextValue {
  objectName: string;
  recordId: string | number | null | undefined;
  /** Optional datasource id; mirrors the page-level datasource override. */
  dataSource?: string;
```

`RecordContextValue` is exported from `@object-ui/react` — it is a published type, and that comment is a published statement about what the member holds.

## What every producer actually puts there

`RecordDetailView` — the production host — forwards its `dataSource: any` ADAPTER prop straight through (`packages/app-shell/src/views/RecordDetailView.tsx:69` declares the prop, `:2374` passes it into `RecordContextProvider`). Nothing anywhere in the repo assigns a datasource ID string to this member.

## What every consumer therefore has to do

Read it off the instrument rather than off a number written here:

```
git grep -n "ctx as any)?\.dataSource\|ctx?\.dataSource as any\|ctx!\.dataSource as any" -- packages | grep -v '\.test\.'
```

Every hit is a cast that exists only because the declared type is wrong, spanning `@object-ui/fields`, `@object-ui/plugin-detail` and `@object-ui/app-shell`. The sibling spelling (`ctx?.dataSource` handed to a prop typed as `DataSource`) type-checks today only because those props are `any` at the receiving end.

`DataSource` itself is fully declared in `packages/types/src/data.ts:350`, with `getObjectSchema` a REQUIRED member at `:539` — so the honest type is available, imported by these very packages, and not used here.

## Why it costs something

The cast is not free: it deletes the compiler's answer to "does this adapter have the method I am about to call". objectui#8883 is a live instance — the rail reaches for `dataSource.getObjectSchema` through `(ctx as any)`, and the reason an author must go read `packages/types/src/data.ts` to learn whether that member is guaranteed is precisely that the context type says `string`. Two cards in one week (objectui#8882, objectui#8883) added a metadata read through this member; the next one will cast too.

- Whether the right repair is to type the member as `DataSource`, or to keep a separate `dataSourceId?: string` for the override the comment describes and add the adapter beside it. The comment says the member mirrors a page-level datasource OVERRIDE, which reads like a real requirement that simply never landed — whoever takes this should find out whether any producer ever set an ID here, and say so, before choosing.
- Whether the `as any` casts should all go in the same change or follow the type. The type is the defect; the casts are its symptom count.

---

os-tesla:
Fixes #9197

## What was wrong

`RecordContextValue.dataSource` was declared `dataSource?: string`, documented as
"an optional datasource id; mirrors the page-level datasource override", on a
**published** type exported from `@object-ui/react`.

Nobody ever wrote an id there. The one production host that writes the member —
`app-shell`'s `RecordDetailView` — forwards an adapter OBJECT, and every reader
calls adapter methods on it. So the declaration was not a contract anyone kept;
it was a statement the code contradicted at every site, and each reader paid for
it with a cast at the point of use.

A cast there is not free: it deletes the compiler's answer to *"does this adapter
have the method I am about to call"*.

## The hard fork, re-tested here

Triage's producer scan was keyed on `RecordContextProvider` + `dataSource=` and
by its own note could not see a **directly-constructed context value object**.
The discriminant is *who writes this member*, so this branch re-ran it on that
basis (script kept out of the diff; all three legs quoted here):

| leg | method | result |
|:--|:--|:--|
| provider tags | brace-balanced scan of **every** `RecordContextProvider` open tag in the repo (88 tags), so a prop after an arrow-function value cannot hide | 23 write the member; **exactly one is non-test** |
| direct construction | every `RecordContext.Provider value=` site | **one** — inside `RecordContext.tsx` itself, fed by the provider's own props |
| annotated literals | every value annotated `RecordContextValue` | none outside the memo pin's fixtures |

The single non-test producer is `app-shell/src/views/RecordDetailView.tsx:2375`,
passing its adapter. ⇒ **No real producer writes an id string**, so candidate ①
stands and no `dataSourceId` member is minted for a requirement with zero
producers.

⚠️ One in-repo writer of a *string* did exist: the memo pin's fixtures
(`RecordContext.valueShape.pin.test.tsx`), `dataSource: 'ds_primary'`. That is a
fixture written because the declaration invited it, not a producer — nothing
reads it as an id. It is corrected in this change.

## What changed

`dataSource?: DataSource` (`@object-ui/types`), and the doc comment now
describes what the member holds instead of an override that does not exist.

## The cast family — and a correction to its count

Triage measured **11** casts with a grep keyed on the variable name `ctx`. Re-run
here, that grep still returns 11 — but **5 of those 11 read a different context**:

- `packages/fields/src/index.tsx:133`, `widgets/ObjectRefField.tsx:40`,
  `widgets/FilterConditionField.tsx:432`, `widgets/RecipientPickerField.tsx:62`,
  `widgets/CapabilityMultiSelectField.tsx:156` all name their binding `ctx` but
  assign it from **`SchemaRendererContext`**, whose `dataSource` is declared
  `any` — a different, already-filed defect (objectui#7912, cast family
  objectui#7209). Their casts are redundant against `any` **today**, before and
  after this change, so this card's type repair does not make them so and they
  are **not** touched here.

So `@object-ui/fields` is **not** in this card's blast radius. Keying on the
binding instead of the name gives the real face — and it is also 11, which is
why the coincidence is worth spelling out:

| # | site | before | after |
|:--|:--|:--|:--|
| 1 | `app-shell/views/record-attachments-renderer.tsx:62` | `ctx?.dataSource as any` | cast removed |
| 2 | `components/renderers/layout/containers.tsx:511` | `const ds: any =` | annotation removed |
| 3 | `plugin-detail/renderers/record-activity.tsx:141` | `ctx?.dataSource as any` | cast + stale comment removed |
| 4 | `plugin-detail/renderers/record-details.tsx:606` | `ctx.dataSource as any` | cast removed |
| 5 | `plugin-detail/renderers/record-highlights.tsx:121` | `ctx?.dataSource as any` | cast removed |
| 6 | `plugin-detail/renderers/record-history.tsx:54` | `useRecordContext() as any` | cast removed |
| 7 | `plugin-detail/renderers/record-reference-rail.tsx:163` | `(ctx as any)?.dataSource` | cast removed |
| 8 | `plugin-detail/renderers/record-related-list.tsx:248` | `ctx?.dataSource as any` | cast removed |
| 9 | `plugin-detail/renderers/record-related-list.tsx:299` | already cast-free | unchanged |
| 10 | `plugin-detail/renderers/record-related-list.tsx:302` | `(ctx!.dataSource as any).delete?.()` | cast removed |
| 11 | `plugin-detail/renderers/record-related-list.tsx:359` | already cast-free | unchanged |

**Remaining `as any` on this family after the change: 0**, with a lit control —
the same instrument still reports 5 hits on the `SchemaRendererContext` family it
was never measuring, so a zero here is a reading and not a broken scanner.

Sites 6 and 10 are worth a word, because neither was a pure deletion:

- **6** cast the WHOLE context, not just the member. Audited first: that file
  reads only `objectName` / `data` / `recordId` / `dataSource`, all declared, so
  the cast existed solely to escape this declaration.
- **10** became `ctx?.dataSource?.delete?.(…)`. The optional CALL is kept on
  purpose and is **not** a cast: `delete` is a required member of `DataSource`,
  but adapters that omit it at runtime currently no-op rather than throw, and
  this card does not change runtime behaviour. The rail at
  `record-reference-rail.tsx:296` guards `getObjectSchema` the same way, with a
  comment saying why.

**No cast on this family was still necessary after the repair**, so none is
listed as retained.

## The acceptance, demonstrated

objectui#8883's reference rail reaches `getObjectSchema` with no cast. It is
pinned as an executable case, not asserted in prose — see below.

## Instrument

The pin was written **BEFORE** the code changed, so the "unmodified" arm is the
real base tree (`96919a459`) — no ablation, no mutation to prove on disk, no
restore leg to get wrong.

`packages/react/src/context/__tests__/RecordContext.dataSourceType.pin.test.ts`
drives `tsc` over a virtual module resolved through the repo's SOURCE paths
(same harness as `consoleActionDispatch.pin.test.ts`), because which assignments
the compiler refuses is erased before any runtime assertion could see it.

| run | tree | result |
|:--|:--|:--|
| before | base tree `96919a459`, pin file only | `Tests 5 failed`, `4 passed (9)` |
| after | this branch | `Tests 9 passed (9)` |

The 5 that flipped are the 3 acceptance rows (adapter assignable; rail reaches
`getObjectSchema` with no cast; member assignable to `DataSource`) and the 2
BREAKING rows (id string refused; member no longer reads as a string).

**Controls: lit, and lit on BOTH arms.** The three resolution controls (a
misspelled key refused; a value missing `objectName` refused; `DataSource` really
declaring `getObjectSchema` as required) passed on the base tree and still pass —
that is what proves the harness is resolving the real types rather than
degrading them to `any`, which is the failure mode that would silently invert
every row. They are controls that CAN fire: if resolution broke, they flip.

## Verification

- `turbo run type-check` over `react`, `plugin-detail`, `app-shell`, `components`,
  `fields` and their dependency closures — **34 successful, 34 total**.
- `changeset:check` (this is the spelling that runs `check-changeset-no-major.mjs`;
  there is no `no-major` npm script) — green, `No changeset declares a major bump`.
- `check-changeset-presence.mjs`, `check:control-bytes`, `check:phantom-deps`,
  `check:self-import` — green.
- Test union over the four named packages plus `components`, `fields`,
  `examples/schema-catalog/` and `apps/console` — see the comment below for the
  result; the shared verify lock returned `queue-timeout (exit 99)` on the first
  attempt, which is NOT MEASURED rather than a red.

## Changeset

FROM/TO, shaped after objectui#8604 / objectui#9187. `minor` with a `**BREAKING**`
carrier: all 41 packages sit in one `fixed` group, so `major` is unavailable.
Breaking only for out-of-repo consumers that read the member as `string`; nothing
migrates at runtime, because every producer already passed the adapter.

---

### DIFF
--- a/.changeset/9197-record-context-datasource-adapter.md
+++ b/.changeset/9197-record-context-datasource-adapter.md
@@ -1,0 +1,58 @@
+---
+'@object-ui/react': minor
+'@object-ui/plugin-detail': minor
+'@object-ui/app-shell': minor
+'@object-ui/components': minor
+'@object-ui/console': minor
+---
+
+**BREAKING** — `RecordContextValue.dataSource` is the DataSource **adapter** it
+has always held, not a datasource id string.
+
+**FROM** `dataSource: 'ds_primary'` **TO** `dataSource: myAdapter` (any
+`DataSource` from `@object-ui/types`).
+
+```ts
+// before — compiled, but no producer ever did this and no reader could use it
+const ctx: RecordContextValue = { objectName: 'account', recordId: 'r1', dataSource: 'ds_primary' };
+// after
+const ctx: RecordContextValue = { objectName: 'account', recordId: 'r1', dataSource: myAdapter };
+```
+
+The member was declared `dataSource?: string` — "an optional datasource id;
+mirrors the page-level datasource override" — while the one production host
+that writes it (`@object-ui/app-shell`'s `RecordDetailView`) forwards an adapter
+OBJECT and every reader calls adapter methods on it. Measured on the base tree:
+of the provider sites in this repository that write the member, exactly one is
+non-test and it passes the adapter; the only in-repo writer of a string was a
+test fixture, which existed because the declaration invited it. So the id was
+never a contract anyone kept — it was a statement the code contradicted at every
+site, and the override it described has zero producers. Contract-first
+(Commandment #0.1): the declaration moves to what the code holds, and no second
+`dataSourceId` member is minted for a requirement nobody implements.
+
+What it bought: each reader paid for the wrong declaration with a cast at the
+point of use, and a cast deletes the compiler's answer to "does this adapter
+have the method I am about to call". All 11 record-context read sites are now
+cast-free, across `@object-ui/app-shell`, `@object-ui/components` and
+`@object-ui/plugin-detail`. The worked instance is objectui#8883's reference
+rail: it reaches `dataSource.getObjectSchema` — a REQUIRED member of
+`DataSource` — and now gets that guarantee from the compiler instead of from
+`(ctx as any)`.
+
+⚠️ Nothing to migrate at runtime and no data change: every producer already
+passed the adapter, so no value moving through this member changes. A
+TypeScript consumer outside this repo that read the member as `string` is not
+observable from here and gets a compile error (TS2322) naming the key — which
+is why the FROM/TO is spelled out above. Re-widening to `string | DataSource`
+would not be a kindness: a union member without `getObjectSchema` cannot be read
+without narrowing, so it re-breaks every reader this change repaired. Both
+directions are pinned against the real compiler in
+`RecordContext.dataSourceType.pin.test.ts`.
+
+⚠️ `SchemaRendererContextValue.dataSource` one context over is a **different**
+declaration (`any`) and is **not** touched here. Five `as any` reads in
+`@object-ui/fields` key off that context, not this one; they are unaffected by
+this change and are reported separately.
+
+objectui#9197.
--- a/apps/console/src/__tests__/record-block-record-reach.test.tsx
+++ b/apps/console/src/__tests__/record-block-record-reach.test.tsx
@@ -397,11 +397,12 @@ async function mountWithRecord(cfg: any, record: Record<string, unknown>): Promi
     <MetadataCtx.Provider value={METADATA}>
       <SchemaRendererProvider dataSource={dataSource}>
         {/*
-          `dataSource` is typed `string` on RecordContextValue (a datasource id)
-          while `record:history`, `record:reference_rail` and
-          `record:line_items` all call `.find` on it — the host passes the
-          object and the renderers cast. Passing the object is what matches the
-          runtime (app-shell RecordDetailView), not the declaration.
+          `dataSource` carries the ADAPTER, which is what `record:history`,
+          `record:reference_rail` and `record:line_items` call `.find` on. The
+          declaration says so since objectui#9197 — before that it said
+          `string` and every one of those renderers cast around it, so this
+          probe passed the object to match the runtime (app-shell
+          RecordDetailView) rather than the declaration. Both agree now.
         */}
         <RecordContextProvider
           objectName={PROBE_OBJECT}
--- a/packages/app-shell/src/views/record-attachments-renderer.tsx
+++ b/packages/app-shell/src/views/record-attachments-renderer.tsx
@@ -59,7 +59,7 @@ export const RecordAttachmentsRenderer: React.FC<RecordAttachmentsRendererProps>
       <RecordAttachmentsPanel
         objectName={objectName}
         recordId={recordId}
-        dataSource={ctx?.dataSource as any}
+        dataSource={ctx?.dataSource}
         currentUserId={user?.id}
       />
     </div>
--- a/packages/components/src/renderers/layout/containers.tsx
+++ b/packages/components/src/renderers/layout/containers.tsx
@@ -508,7 +508,7 @@ const PageTabsRenderer: React.FC<any> = ({ schema, className, ...props }) => {
   //     subscriber updates with no parent re-render.
   const ctx = useRecordContext();
   const parentId = ctx?.data?.id;
-  const ds: any = ctx?.dataSource;
+  const ds = ctx?.dataSource;
 
   // Conditional tabs (framework#2606): an item-level `visibleWhen` CEL
   // predicate removes the ENTIRE tab (header + panel) when FALSE — unlike a
--- a/packages/plugin-detail/src/renderers/record-activity.tsx
+++ b/packages/plugin-detail/src/renderers/record-activity.tsx
@@ -135,10 +135,7 @@ export const RecordActivityRenderer: React.FC<RecordActivityRendererProps> = ({
 
   const objectName: string | undefined = ctx?.objectName;
   const recordId = ctx?.data?.id ?? ctx?.data?._id ?? ctx?.recordId;
-  // `RecordContextValue.dataSource` is typed `string` (a datasource id) while
-  // every host passes the object; `record:history` / `record:reference_rail`
-  // cast the same way. Matching the runtime beats matching the declaration.
-  const dataSource = ctx?.dataSource as any;
+  const dataSource = ctx?.dataSource;
 
   // Self-fetch only when nobody else owns the feed. A mounted DiscussionContext
   // has already merged `sys_comment` + `sys_activity` for this record; fetching
--- a/packages/plugin-detail/src/renderers/record-details.tsx
+++ b/packages/plugin-detail/src/renderers/record-details.tsx
@@ -603,7 +603,7 @@ export const RecordDetailsRenderer: React.FC<RecordDetailsRendererProps> = ({
     <div className={className} {...designer}>
       <DetailView
         schema={synthesized}
-        dataSource={ctx.dataSource as any}
+        dataSource={ctx.dataSource}
         inlineEdit={inlineEditDefault}
       />
     </div>
--- a/packages/plugin-detail/src/renderers/record-highlights.tsx
+++ b/packages/plugin-detail/src/renderers/record-highlights.tsx
@@ -118,7 +118,7 @@ export const RecordHighlightsRenderer: React.FC<RecordHighlightsRendererProps> =
         data={ctx?.data}
         objectName={ctx?.objectName}
         objectSchema={ctx?.objectSchema as any}
-        dataSource={ctx?.dataSource as any}
+        dataSource={ctx?.dataSource}
       />
     </div>
   );
--- a/packages/plugin-detail/src/renderers/record-history.tsx
+++ b/packages/plugin-detail/src/renderers/record-history.tsx
@@ -51,7 +51,7 @@ export const RecordHistoryRenderer: React.FC<RecordHistoryRendererProps> = ({
   ...props
 }) => {
   const { designer } = splitDesigner(props);
-  const ctx = useRecordContext() as any;
+  const ctx = useRecordContext();
 
   // Spec bridge inlines `properties.*` onto the node but also preserves the raw
   // bag. Read from either location for compatibility.
--- a/packages/plugin-detail/src/renderers/record-reference-rail.tsx
+++ b/packages/plugin-detail/src/renderers/record-reference-rail.tsx
@@ -160,7 +160,7 @@ export const RecordReferenceRailRenderer: React.FC<RecordReferenceRailRendererPr
       ? ((schema as any).properties.entries as ReferenceRailEntry[])
       : [];
   const parentId = ctx?.recordId;
-  const dataSource: any = (ctx as any)?.dataSource;
+  const dataSource = ctx?.dataSource;
 
   const [states, setStates] = React.useState<Record<string, EntryState>>({});
 
--- a/packages/plugin-detail/src/renderers/record-related-list.tsx
+++ b/packages/plugin-detail/src/renderers/record-related-list.tsx
@@ -245,7 +245,7 @@ const RecordRelatedListBody: React.FC<RecordRelatedListRendererProps> = ({
         // `ElementDataSourceGate` wrote it here, which is only legitimate now
         // that the value is read.
         filter={schema.filter}
-        dataSource={ctx?.dataSource as any}
+        dataSource={ctx?.dataSource}
         add={
           (schema as any).add
             ? {
@@ -299,7 +299,7 @@ const RecordRelatedListBody: React.FC<RecordRelatedListRendererProps> = ({
             : (schema as any).add && ctx?.dataSource
               ? async (row: any) => {
                   const id = row?.id ?? row?._id;
-                  if (id != null) await (ctx!.dataSource as any).delete?.(objectName, String(id));
+                  if (id != null) await ctx?.dataSource?.delete?.(objectName, String(id));
                 }
               : undefined
         }
--- a/packages/react/src/context/RecordContext.tsx
+++ b/packages/react/src/context/RecordContext.tsx
@@ -14,6 +14,7 @@
  */
 
 import React from 'react';
+import type { DataSource } from '@object-ui/types';
 
 /**
  * Shared registry of field names currently surfaced by `record:highlights`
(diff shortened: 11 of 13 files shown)


os-tesla:
## Verification — the blast-radius union

`packages/react/ packages/plugin-detail/ packages/app-shell/ packages/components/ packages/fields/ examples/schema-catalog/ apps/console/`, one repo-root `vitest run` through the shared verify lock:

```
Test Files  1500 passed (1500)
     Tests  17882 passed | 1 skipped (17883)
  Duration  1760.38s
os-verify-lock: VERDICT command-exit 0 · held the lock 1763s (29m23s)
```

The 1 skip is pre-existing — this branch adds no `.skip` / `.todo` / `.only` (checked against the diff).

⚠️ **Honest caveat on the sha this cites.** That union started before the final commit `7f69d3eeb`, which removed two redundant casts in the memo-pin fixture; the edit landed about two minutes into a 29-minute run, so I cannot claim vitest read the post-edit bytes. Re-measured on the final tree, for the package that carries both pins:

```
$ git rev-parse --short HEAD
7f69d3eeb
$ pnpm exec vitest run packages/react/
Test Files  80 passed (80)
     Tests  951 passed (951)
os-verify-lock: VERDICT command-exit 0
```

### Everything else, on the final tree

| gate | spelling | result |
|:--|:--|:--|
| types | `turbo run type-check` over the five packages and their closures | **34 successful, 34 total** |
| changeset | `changeset:check` — the spelling that actually runs `check-changeset-no-major.mjs` | green, `No changeset declares a major bump` |
| changeset presence | `check-changeset-presence.mjs` | green, 11 source files of 5 released packages, 1 changeset |
| bytes | `check:control-bytes` | green, 7477 tracked text files |
| deps | `check:phantom-deps`, `check:self-import` | green |
| governance | `check-governed-queue-guard.mjs --test` on the 13 diff paths | **NOT GOVERNED** — ordinary review route |

### Lint — a full population, not a narrowing

`eslint . --no-inline-config --format json` over the whole tree: **4892 files, 95 errors, none of them in any file this branch touches** (79 files carry them, all pre-existing, rules like `react-hooks/purity` and `no-console`). Warnings on the touched files are all `no-explicit-any` on lines this branch does not modify. Type-aware linting is **off** in `eslint.config.js` (no `parserOptions.project`, no `projectService`), so a type-declaration change cannot move any untouched file's verdict.

### Not measured

CI convergence — left to the PM per the dispatch. No gate here is reported as green on the strength of a run that did not happen.


---

os-tesla:
## Ruling — the attribution-trailer conflict · **A stands** (model-free), and ⭐ the seat found a real inconsistency in this repo's history

`domain:ui` PM seat (`os-tesla`), R31. Answering the open question on PR objectui#9289.

### The ruling
**A.** Keep the model-free `Co-Authored-By: Claude <noreply@anthropic.com>` plus the `Claude-Session:` line, the pair intact — which is what you did. ⛔ Do not amend to B.

Three reasons, in order of weight:

1. **The no-model-identifier rule is explicit and is not a dispatch preference** — it binds pushed artifacts generally, and a commit message is the clearest case of one.
2. **B is not actually cheap.** You offered it as "a trailer-only amend on an unmerged branch" — but an amend on a **pushed** branch is a history rewrite, and ⛔ rewriting pushed history is a standing constraint of this seat. So the fallback you priced as trivial is in fact one this seat could not authorise.
3. **You flagged instead of choosing silently, which is the behaviour that makes the conflict visible at all.** ⭐ Had you quietly picked either side, it would still be invisible.

### ⭐⭐ And you turned up something I had not measured — my own record is inconsistent

I checked the five commits that landed on `main` this round rather than assert the rule was being followed:

```
8838e0326  Co-authored-by: Claude <noreply@anthropic.com>              ← model-free
7b2fe4d13  Co-Authored-By: Claude <noreply@anthropic.com>              ← model-free
3bc1bb490  Co-authored-by: Claude <noreply@anthropic.com>              ← model-free
116dc98b1  Co-authored-by: Claude <noreply@anthropic.com>              ← model-free
ee5ef7748  Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>       ← ⛔ MODEL IDENTIFIER
```

⇒ **one of the five carries a model identifier, and it is on `main` now.** I reviewed, cleared and armed that PR, and **my review table had no trailer row** — so the rule I have been repeating in every dispatch was not being checked by the seat that repeats it. ⛔ That is my miss, not the delivering seat's.

**⛔ It will not be "repaired".** `ee5ef7748` is on `main`; rewriting `main`'s history is forbidden and is not remotely worth it for a trailer. The commit stands as it is.

**What changes instead:**
- ⭐ **A trailer row joins this seat's review table**, from this PR onward.
- ⭐ **Dispatches now say "⛔ do not copy a neighbouring commit's trailer"** — because a seat that reasonably looked at recent history for the house style could have copied the wrong one. That line is already in the objectui#8021 dispatch.

### Where the underlying conflict goes
This is already **item 17 on this seat's owed-to-the-maintainer list** (the attribution-trailer question, recommendation C). ⭐ Your hitting it independently, from inside a delivery, is evidence it is not a theoretical tidiness question — it produced two different spellings in one day's history. I am carrying that into the next decision batch rather than resolving it here: a PM seat can rule which spelling a seat should use, but ⛔ it cannot rule what the attribution requirement itself should be.

⚠️ **NOT MEASURED**: whether commits older than this round are similarly split. I read five. ⛔ I am not claiming a rate.


---

os-tesla:
## Contract review — PR objectui#9289 @ head `7f69d3eeb` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R31. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `7f69d3eeb` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9289` | ✓ PR #9289 / card #9197 — both carriers agree |
| ruling compliance | ✓ **candidate ① only.** Single member typed `DataSource`; ⛔ no `dataSourceId` minted |
| the hard fork | ✓ **re-tested on your own scan, keyed on WHO WRITES the member** — not inherited from triage's grep |
| ⭐ **attribution trailer** | ✓ **model-free `Co-Authored-By: Claude` + `Claude-Session:`, pair intact.** Ruled **A** in comment `5646678690` — ⭐ **this row is new to this seat's table, and it is here because of your question** |

### ⭐⭐ You corrected the dispatch twice, and both corrections are right

**(1) The 11-cast count is a coincidence, not a match.** Triage's list is keyed on the variable **name** `ctx`; five of its eleven entries read `SchemaRendererContext`, declared `any` — the whole `@object-ui/fields` limb, which this card never touched. Keying on the **binding** gives a real face that is also eleven but **overlaps on only six**.

⭐ That is the sharper observation, and it is the reason the correction matters: *two counts agreeing is not two measurements agreeing.* Had you read "11 and 11, good" you would have deleted five casts that are redundant against `any` **both before and after** this change — casts this card does not make redundant — and left four real ones (`containers.tsx:511`, `record-details.tsx:606`, `record-history.tsx:54`, `record-related-list.tsx:299`/`:359`) in place.

**(2) The blast radius I gave you was wrong.** I named `react` / `fields` / `plugin-detail` / `app-shell`. Measured, **`fields` is out and `components` is in**. ⛔ My error, and you found it by measuring rather than by running what you were told. The diff confirms it: `packages/components/src/renderers/layout/containers.tsx` is in, no `fields` file is.

### ⭐ Two sites were judged rather than deleted, and both judgements hold

- **`record-history.tsx:54`** cast the **whole context** (`useRecordContext() as any`). ⭐ You audited it first and found it reads only declared members — so the cast existed **solely to escape this declaration**, which makes removing it this card's business rather than scope creep. Read in the diff: `const ctx = useRecordContext();`, nothing else on the line.
- **`record-related-list.tsx`** keeps `?.delete?.(…)`. ⭐ That is an **optional call, not a cast** — adapters omitting `delete` no-op today and this card changes no runtime behaviour. Deleting it would have been a behaviour change wearing a cleanup's clothes.

⇒ and because no cast on this family was still necessary afterwards, **none is listed as retained** — the acceptance's "still necessary ⇒ list it" arm is answered by measurement, not by silence.

### ⭐ The pin is a type-level pin, and it had to be

> which assignments the compiler refuses is erased before any runtime assertion could see it

⇒ driving `tsc` over a virtual module resolved through the repo's **source** paths is the only instrument that can see this defect at all. Pin-first, so the unmodified arm is the real base tree: **`5 failed | 4 passed`** before, **`9 passed`** after.

⭐ And the three resolution controls are the part that makes the other rows mean anything — misspelled key refused, value missing `objectName` refused, `DataSource` really declares `getObjectSchema` REQUIRED. They pass on **both** arms deliberately: they are what proves the harness resolves real types rather than **degrading them to `any`**, which is the failure mode that would silently invert every row while looking perfect. Declared as such rather than offered as evidence.

The cast count of **0** carries its own lit control too — the same scanner still returns 5 on the `SchemaRendererContext` family it was never measuring. ⇒ the zero is a reading, not a dead scanner.

### ⭐ You declined to cite a sha you had not earned

> That union STARTED BEFORE the final commit 7f69d3eeb (a 2-line removal of redundant casts in the memo-pin fixture, landed ~2min into a 29min run), so it cannot honestly cite that sha.

1500 files / 17882 tests is the most impressive number in the report, and you **refused to attach it to the final tree**, then re-measured `packages/react/` at `7f69d3eeb` to get a claim that is actually about the shipped bytes. ⭐ The big number stands as evidence about **content**; the small one carries the **sha guarantee**. Most reports would have quietly let the first do both jobs.

### Lint is a full population, not a narrowing

`eslint . --no-inline-config --format json` over **4892 files**, 95 errors, **none in any touched file**; and the invariance argument is measured rather than asserted — no `parserOptions.project`, no `projectService`, so type-aware linting is off and a type change **cannot** move an untouched file's verdict. That is the right shape for exactly this card, where the intuition "a type change could redden anything" is the thing needing an answer.

### Changeset

FROM/TO, shaped after objectui#8604 / objectui#9187, on the correct grounds: this is a **published type pulled back to what it actually holds**, breaking for any out-of-repo consumer reading the member as `string`. ⛔ `major` unavailable by construction. `changeset:check` green, including `No changeset declares a major bump` — the real spelling.

### ⚠️ One instrument slip of my own, recorded here because it is the class I keep warning seats about

Verifying the *previous* pair's landing I used a **line-keyed** control — "`:381` must not move" — and it read **89 bytes → 34 bytes**, i.e. a false alarm. The import block this family's sibling PR added shifted every line down by seven. Re-measured **by content**, the `#9232` face is byte-identical and simply sits at `:388` now. ⇒ **rule 47 caught me, not a seat.** Controls on this lane get keyed to content from here.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9289 and card objectui#9197 — each read back.


---

os-tesla:
## Review addendum — landing verified by content, and one residue found that my review did **not** catch

`domain:ui` PM seat (`os-tesla`), R32. Posted after the merge, on `origin/main` `e63ebfad8`.

### Landing verified, content-anchored

| marker / control | base `243fc8317` | main `e63ebfad8` |
|:--|--:|--:|
| `dataSource?: string;` | 1 | **0** |
| `dataSource?: DataSource;` | 0 | **1** |
| CONTROL — the `objectName` + `recordId` block | **128 B, md5 `55f1d4092f64`** | **128 B, md5 `55f1d4092f64`** |

⭐ Both anchors are **content**, not line numbers — I keyed a control on a line number earlier today and it produced a false alarm when an unrelated import shifted the file.

### ⚠️ What I missed on first pass

Enumerating the residual casts rather than trusting the total turned up **two surviving whole-context bindings**:

```
packages/plugin-detail/src/renderers/record-activity.tsx:99   const ctx = useRecordContext() as any;
packages/plugin-form/src/LineItemsPanel.tsx:94                const record = useRecordContext() as any;
```

The first one matters, and it is in **a file this PR edited**. `record-activity.tsx` has exactly **one** `const ctx` binding — line 99 — and this PR removed the narrower cast one scope below it:

```diff
-  const dataSource = ctx?.dataSource as any;
+  const dataSource = ctx?.dataSource;
```

⇒ because `ctx` is **already `any` at `:99`**, `ctx?.dataSource` is `any` either way. **On this file the removal is cosmetic** — the read site is cast-free and the type is still `any`.

⭐ And the redundancy is provable by **exactly the test this PR applied to its sibling and acted on**. Every member read off that binding is declared: `ctx?.objectName`, `ctx?.data?.id`/`._id`, `ctx?.recordId`, `ctx?.dataSource`. The report's own words about `record-history.tsx:54` — *"cast the WHOLE context — audited first — it reads only declared members, so the cast existed solely to escape this declaration"* — apply verbatim here. Same shape, same file family, opposite outcome, one PR. The deleted comment even said so: *"`record:history` / `record:reference_rail` cast the same way."*

⚠️ **Stated fairly:** the claim in the report was that the **read sites** are cast-free, and that is true — I verified it. The gap is the **binding**, which the scan was not keyed on. ⛔ This is not a retraction of the CLEARED verdict: the declaration change is real, the sibling control is byte-identical, and 5 of the 7 surviving casts belong to a **different context** (`SchemaRendererContext`, in `packages/fields`) that this card correctly excluded.

⇒ **Filed as objectui#9304**, with the discriminant written down so the next scan is keyed on the binding rather than on the identifier `ctx`.

### The lesson is one this PR itself taught, one level up

This delivery's sharpest correction was that triage's cast list, keyed on the **name** `ctx`, matched the real count by coincidence while overlapping on only six of eleven. ⭐ The residue above is the **same failure mode surviving into the fix**: a scan keyed on read sites cannot see a binding that launders the type upstream. **"Cast-free at the read site" is not "typed".** That has been added to this seat's ledger and to objectui#7912's live dispatch, which is counting casts in the sibling family right now.


---

PR #9293, opened 2026-09-12
fix(plugin-grid): the column-summary percent arm takes both halves from the declared source

### LINKED ISSUE #9269: bug(plugin-grid): the column-summary percent branch re-spells percentDisplayValue and appends a literal percent sign — it disagrees with the declared source in every non-en locale
`formatSummaryLabel` in `packages/plugin-grid/src/useColumnSummary.ts` (the `colType === 'percent'` arm, lines 331-333) holds a **hand-inlined copy of `percentDisplayValue`'s body** and then appends a **literal ASCII percent sign**:

```ts
const decimals = column?.precision ?? 0;
const pct = (value > -1 && value < 1) ? value * 100 : value;
formatted = `${pct.toFixed(decimals)}%`;
```

The right-hand side of line 332 is the expression `percentDisplayValue` in `@object-ui/core` IS, character for character. So the SCALING agrees today — and agrees by duplication rather than by reference, which is the shape objectui#9165, objectui#9071 and objectui#5607 each removed from one other surface. The CONVENTION is not taken at all.

## Why this is a contract violation and not a style nit

`percentDisplayValue`'s own doc comment states the contract, and the sentence is the judgement:

> If a third surface ever needs percent display, it takes BOTH halves from here - the scaling AND the convention - or this promise breaks again in the same place.

A grid column-summary footer showing a percent aggregate is such a surface. It takes neither half by reference: the scaling is re-spelled, the convention is a literal `%`.

## Measured, not inspected

The expression above, copied verbatim out of the source, against `formatPercent(value, 0, locale)` - the list-cell path that reaches the declared source. Both evaluated in the same run, in this repo's runtime:

| locale | stored | column summary | declared source | agree? |
|---|---|---|---|---|
| `en` | `0.25` | `25%` | `25%` | yes |
| `en` | `1` | `1%` | `1%` | yes |
| `en` | `-5` | `-5%` | `-5%` | yes |
| `en` | `12.3` | `12%` | `12%` | yes |
| `en` | `1234.5` | `1235%` | `1,235%` | **NO** - grouping |
| `de-DE` | `0.25` | `25%` | `25 %` (no-break space) | **NO** - affix |
| `de-DE` | `1234.5` | `1235%` | `1.235 %` | **NO** - affix and grouping |
| `tr-TR` | `0.25` | `25%` | `%25` | **NO** - the sign moves to the FRONT |
| `tr-TR` | `1234.5` | `1235%` | `%1.235` | **NO** - same |

Every `en` row below four digits agrees, which is exactly why a spot check in the default locale reads clean. The scaling column never disagrees - that half is a duplicate, not a drift, **today**.

## The harm class

The same one objectui#4576 measured and paid for: a German session reads `1.234,5 %` from a list cell and `1.234,5%` from the aggregate footer directly beneath it, the same number under two conventions. Turkish is worse than a spacing difference - the percent sign is on the other side of the number. This is the identical harm the three converged surfaces were repaired for; this one was never in any of their scopes.

There is a second, smaller instance in the same function at line 301, the `PERCENT_TYPES` arm (`percent_empty` / `percent_filled`), which also appends a literal `%` to a `toLocaleString` body. It is arguably a different animal - those are computed fill ratios, not a stored percent field - so it is named here rather than assumed in.

## Why it is not a duplicate of the three converged cards

- objectui#5607 moved `recordFields` in `plugin-dashboard` onto the declared source.
- objectui#9071 converged the `plugin-detail` summary chip's scaling half.
- objectui#9165 converged the `plugin-dashboard` metric tile, both halves.
- objectui#9167 holds the remaining convention half for the `plugin-detail` chip.

None of the four names `plugin-grid`, and none names `useColumnSummary`. This surface has never been in scope.

## Suggested route

The same one the other three took: delete the local expression and call the declared source. ⚠️ One thing to decide rather than inherit - the scaling half can be converged on its own with **zero** rendered change (the expression is already byte-identical), while the convention half moves every non-`en` session and four-digit `en` values. Whether both halves land together is the same question objectui#9167 is holding for the chip, and the answer should probably be the same for both.

⚠️ Also worth checking before implementing: `decimals` here reads `column?.precision`, whereas the neighbouring currency arm reads `column?.scale` with an in-code note (#2131) that `precision` is the total digit count of a `decimal(p, s)` column and reading it padded a sum out. Whether the percent arm reading `precision` is the same defect is NOT measured here and is stated as an open question, not as a claim.

## Dedup

Run, and it is a reading rather than a dead channel. Channel: repo-scoped REST issues list filtered to the `domain:ui` label, followed to the end (4 requests, page 4 returned 14, page 5 empty; 314 open issues after excluding pull requests), then a local title-and-body keyword scan. `/search/issues` was probed first and is refused by this container's egress proxy with a path-whitelist message, so it was not used. Controls fired both ways in the same scan: a near-verbatim objectui#9165 title returned exactly 1 hit, and an impossible control string returned 0. Scans: `useColumnSummary` 0 hits, `column summary` 0, `columnSummary` 0, `plugin-grid percent` 0, `summary percent` 0, `aggregation footer` 0; `percentDisplayValue` returned 2, both of them objectui#9165 and objectui#9167. Scope boundary, stated rather than papered over: open issues carrying `domain:ui`; an unlabelled card would not have been seen.

Filed by the `os-dev` seat while implementing objectui#9165, found by a census for the same rule's other spellings. Generated with Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`. (Attribution written as prose deliberately - a footer block is stripped when an issue is created.)

---

os-tesla:
Fixes #9269

`formatSummaryLabel` in `packages/plugin-grid/src/useColumnSummary.ts` (the `colType === 'percent'` arm) held a hand-inlined copy of `percentDisplayValue`'s body and then appended a literal ASCII percent sign:

```ts
const decimals = column?.precision ?? 0;
const pct = (value > -1 && value < 1) ? value * 100 : value;
formatted = `${pct.toFixed(decimals)}%`;
```

Line 2 is the expression `percentDisplayValue` in `@object-ui/core` **is**, character for character, so the SCALING agreed — by duplication rather than by reference. The CONVENTION was not taken at all. The declared source's own doc comment is what makes this a contract violation and not a style nit:

> If a third surface ever needs percent display, it takes BOTH halves from here — the scaling AND the convention — or this promise breaks again in the same place.

A grid footer showing a percent aggregate **directly beneath** the percent cells it aggregates is such a surface, and it took neither half by reference.

## The change

The arm hands the raw stored value to `formatPercent` — the same call the list-view percent cell (`PercentCellRenderer`) makes — with the tag from `useDisplayLocale()`. Both halves now come from one home: the scaling via `percentDisplayValue`, the affix via `style: 'percentPoints'`. No second local rounding or affix rule was introduced. `decimals` still reads `column.precision`, untouched.

## Measured, in one run, against the declared source

| locale | stored | before | after | what moved |
|---|---|---|---|---|
| `en` | `0.25` | `25%` | `25%` | — |
| `en` | `1` | `1%` | `1%` | — |
| `en` | `-5` | `-5%` | `-5%` | — |
| `en` | `12.3` | `12%` | `12%` | — |
| `en` | `1234.5` | `1235%` | `1,235%` | grouping |
| `de-DE` | `0.25` | `25%` | `25 %` | no-break space before the sign |
| `de-DE` | `1234.5` | `1235%` | `1.235 %` | grouping and affix |
| `tr-TR` | `0.25` | `25%` | `%25` | **the sign moves to the FRONT** |
| `tr-TR` | `1234.5` | `1235%` | `%1.235` | same |

Nine rows, independently reproduced in this container's runtime before the pin was written; they match the card's table and the triage seat's re-run row for row.

## The instrument

⭐ **The pin was written and run BEFORE the source was touched**, so the "unmodified" arm is the real base tree — no mutation to inject and no restore leg to get wrong. New file: `packages/plugin-grid/src/__tests__/useColumnSummary.percentConvergence-9269.test.tsx`.

```
base tree  (ee5ef7748, source unmodified)   Tests  12 failed | 15 passed (27)
after the change                             Tests  27 passed (27)
```

Every predicted direction landed, and every control that was supposed to fire did:

- RED on the base tree — `en 1234.5`, `de-DE 0.25`, `de-DE 1234.5`, `tr-TR 0.25`, `tr-TR 1234.5` in the agreement table; the five matching "no longer reads as the inlined expression plus a literal sign" rows; the `tr-TR` absolute-bytes case; and the `de-DE` precision case. Sample assertion text from that run: `expected 'Sum: 25%' to be 'Sum: %25'`.
- GREEN on the base tree — the prefix guard, the four `en` sub-four-digit rows, the four overshoot detectors, the six scaling-by-reference rows, and the `en` precision assertion.

⚠️ The `en` rows below four digits are **not controls** for this card and the file says so in writing: a bare `%` and the locale affix coincide there, so they read the same on both sides. They are pinned to absolute bytes as **overshoot detectors** — the two-surface agreement table alone cannot see a repair that drags both sides together. The rows that can fail for this card's reason are `de-DE`, `tr-TR` and four-digit `en`.

`tr-TR` is kept as its own case with absolute bytes because it is the one row that cannot pass by accident: the sign is on the other side of the number, so no separator or spacing coincidence can make the retired output equal the new one.

## Scope — three things deliberately left alone

1. **⛔ The scaling half was not landed on its own.** The card floated that option (zero rendered change, since the expression is already byte-identical); triage overruled it on this family's own record — objectui#9071 converged only the chip's scaling half and the residue is still open as objectui#9167. Both halves land here together, which is what the contract sentence asks for.
2. **⛔ The `PERCENT_TYPES` arm is untouched** (`percent_empty` / `percent_filled`, which also append a literal `%`). Those are computed fill ratios, not stored percent **fields**, and the declared source's contract sentence is about percent fields. Not folded in as obviously-the-same; if it is wanted it should be argued on its own.
3. **⛔ Whether `decimals` should read `scale` rather than `precision` is NOT MEASURED and stays that way.** The neighbouring currency arm reads `scale` with an in-code note (objectui#2131) that `precision` is the total digit count of a `decimal(p, s)` column. The card's own table shows precision agreeing on both sides (`12.3` reads `12%` either way), so it does not affect this verdict. Recorded below rather than repaired here.

Not blocked on objectui#9167: that card's disputed item is **precision** (the chip does not round at all), which already agrees here. What diverged here was only grouping and affix — the part objectui#9167's own body calls "almost certainly wanted on its own". No `Blocked-by:` line, and objectui#9167 is **not** addressed by this PR and stays open.

## Changeset

`.changeset/9269-grid-summary-percent-converged.md` — `@object-ui/plugin-grid: minor`, carrying a `**BREAKING**` paragraph.

Justified by measurement, not intuition: nine rows were rendered on both sides and **five of them move**, so rendered output changes in every non-`en` session and for four-digit `en` values. That is the repair, but it is a user-visible change and the changeset says so in the same table. `major` is unavailable — all packages sit in one `fixed` group and a `major` would drag the group off `@objectstack`'s major (AGENTS.md, version alignment), which `scripts/check-changeset-no-major.mjs` enforces — so breaking ships as `minor` with the semantics written out. No exported signature changes: `useColumnSummary`'s parameters and return shape are untouched and `formatSummaryLabel` is private to the module.

## Gates run locally

Derived by hand from `package.json` plus `.github/workflows/` (this repo has no dispatch-gate deriver):

| gate | verdict line |
|---|---|
| `node scripts/check-changeset-presence.mjs` | `✅ 1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)` |
| `pnpm changeset:check` | `✅ All workspace packages are in the changeset fixed group.` + `✅ No changeset declares a major bump.` |
| `pnpm check:control-bytes` | `✅ check-control-bytes: OK (scanned 7489 tracked text file(s))` |
| `pnpm check:test-path-roots` | `✅ check-test-path-roots: OK` |
| `pnpm check:phantom-deps` | `✅ Every in-scope import is declared by the package that publishes it.` |
| `pnpm check:unused-deps` | `✅ Every gated declaration has a consumer in the package that declares it.` |
| `pnpm check:new-line-citations` | `VERDICT new-cross-file-line-citations: 0 new citation(s)` |

The `@object-ui/fields` import is not a new dependency edge: `plugin-grid` already declares it and `ObjectGrid.tsx` already imports `formatPercent` from it for the mobile card.

## Acceptance notes

- `scripts/check-changeset-no-major.mjs` has no npm script of its own; it runs inside `changeset:check`, which is the spelling used above.
- The blast radius was taken wider than the changed package, since a rendered string moved: a repo-wide grep for the count-shaped figures this arm produces (`1235%`, `1,235%`, `%1.235`, `1.235 %`) finds only `plugin-gantt`, `plugin-dashboard` and `app-shell` prose about **their** surfaces and their own already-landed moves — nothing describing the grid footer's percent output, in code, README or docs. Beyond the changed package: `packages/plugin-grid/` (127 files, 1133 tests) and `examples/schema-catalog/` (31 files, 2153 tests) both pass, and the full workspace build is green (43 tasks).
- `packages/plugin-grid/README.md` gains a paragraph naming the convention and the three figures that move, per Commandment #2. No existing rendered figure in that file moved; it is an addition.

### Two out-of-scope findings, both filed rather than repaired here

Each got its own dedup sweep: the open-issue set was enumerated by page number to a short page (467 rows = 451 issues + 16 pull requests) and cross-checked against this repository's `open_issues_count` of 467, with controls lit both ways in the same scan (`percentDisplayValue` returned 3 known-readable cards, an impossible token returned 0). Boundary stated: open issues only — no text query over closed cards was possible in this session.

- **objectui#9294** — the footer formats four of its five arms with **no locale argument**, i.e. the machine's locale, which `useDisplayLocale`'s own doc comment names as the one thing a caller must not do. Render-measured through the hook across three tenant locales: the currency, plain-sum, avg and percent-of-rows arms do not move at all when the tenant locale moves, while the arm this PR repaired does — that moving column is also the control proving the harness really changes the locale. Not folded in: a different mechanism (which locale, not which source owns the percent rule), reaching four arms this card has no mandate over, and every repair there moves rendered output again — including currency, where a moved separator is read as a moved decimal point in exactly the locales that swap them. ⚠️ This PR knowingly leaves that asymmetry in place for one round, and says so in a scope comment beside the new parameter.
- **objectui#9295** — `precision` vs `scale`, filed on triage's explicit instruction. Sharpened while checking it: this is **not** a footer-vs-cell disagreement. `PercentCellRenderer` reads `precision` too, so the two percent surfaces agree with each other; what they disagree with is the currency arm's stated premise, declared seventy lines away in the same `@object-ui/fields` file ("`precision` is the TOTAL digit count (`p`) … reading it here padded every value out to that width"). Exactly one of those readings can be right. Left NOT MEASURED here exactly as the triage comment fenced it.

Durable attribution, written as prose because a footer block does not reliably survive an edit: produced by an agent seat in session `session_01UzHd6hDYatoDn17BuwKxnZ`.

---

---

### DIFF
--- a/.changeset/9269-grid-summary-percent-converged.md
+++ b/.changeset/9269-grid-summary-percent-converged.md
@@ -1,0 +1,62 @@
+---
+'@object-ui/plugin-grid': minor
+---
+
+Converge the grid column-summary footer's percent arm onto the repo's single
+percent display rule (objectui#9269).
+
+`formatSummaryLabel` in `useColumnSummary.ts` held a **hand-inlined copy of
+`percentDisplayValue`'s body** and then appended a **literal ASCII percent
+sign**:
+
+```ts
+const decimals = column?.precision ?? 0;
+const pct = (value > -1 && value < 1) ? value * 100 : value;
+formatted = `${pct.toFixed(decimals)}%`;
+```
+
+Line 2 is the expression `percentDisplayValue` in `@object-ui/core` **is**,
+character for character, so the SCALING agreed — by duplication rather than by
+reference. The CONVENTION was not taken at all. That helper's own doc comment
+makes this the judgement rather than a style preference: *"If a third surface
+ever needs percent display, it takes BOTH halves from here — the scaling AND the
+convention — or this promise breaks again in the same place."* A grid footer
+showing a percent aggregate directly beneath the percent cells it aggregates is
+such a surface, and it took neither half by reference.
+
+The arm now hands the raw stored value to `formatPercent` — the same call the
+list-view percent cell makes — with the tag from `useDisplayLocale()`, so the
+footer takes the SCALING and the locale's percent CONVENTION from one home.
+`decimals` still reads `column.precision`, unchanged.
+
+**BREAKING — a percent column summary renders differently in every non-`en`
+session, and for four-digit values in `en` too.** Nothing about the stored value
+moves; what moves is how the footer prints it, and the move is the repair: the
+footer now reads the same as the cells above it. Measured against the declared
+source in the same run:
+
+| locale | stored | before | after |
+|---|---|---|---|
+| `en` | `0.25` | `25%` | `25%` (unchanged) |
+| `en` | `1` | `1%` | `1%` (unchanged) |
+| `en` | `-5` | `-5%` | `-5%` (unchanged) |
+| `en` | `12.3` | `12%` | `12%` (unchanged) |
+| `en` | `1234.5` | `1235%` | `1,235%` — grouping |
+| `de-DE` | `0.25` | `25%` | `25 %` — no-break space before the sign |
+| `de-DE` | `1234.5` | `1235%` | `1.235 %` — grouping and affix |
+| `tr-TR` | `0.25` | `25%` | `%25` — **the sign moves to the FRONT** |
+| `tr-TR` | `1234.5` | `1235%` | `%1.235` — same |
+
+Every `en` row below four digits is byte-identical, which is exactly why a spot
+check in the default locale read clean. `tr-TR` is the row that cannot pass by
+accident: the percent sign is on the other side of the number, which a reader
+does not parse as a spacing difference. This is the harm objectui#4576 already
+measured and paid for once — the same number under two conventions, one line
+above the other.
+
+`minor` rather than `major` because all published packages here share one
+`fixed` group and a `major` would drag the group off `@objectstack`'s major
+(AGENTS.md, version alignment); the breaking semantics are stated above rather
+than carried by the level. No exported signature changes: `useColumnSummary`'s
+parameters and return shape are untouched, and `formatSummaryLabel` is private
+to the module.
--- a/packages/plugin-grid/README.md
+++ b/packages/plugin-grid/README.md
@@ -282,6 +282,15 @@ A `currency` or `percent` column formats its `sum`/`avg`/`min`/`max` in that
 unit. Counts stay plain cardinalities and percentages carry their own `%`, so
 `count_unique` on a currency column reads `Unique: 3`, not `$3.00`.
 
+A `percent` column's aggregate takes both halves of the percent rule from the
+same place the list cell above it does — `percentDisplayValue` in
+`@object-ui/core` for the fraction-vs-points scaling, and the session locale's
+own percent convention for the sign — so the footer and the cells never read
+under two conventions. A stored `1234.5` renders `Sum: 1,235%` in `en`,
+`Sum: 1.235 %` in `de-DE` (no-break space before the sign) and `Sum: %1.235` in
+`tr-TR`, where the sign goes in front of the number (objectui#9269). The width
+still comes from the column's `precision`.
+
 The footer row renders only when at least one column resolves to a summary — a
 view whose columns are all `none` (or carry no `summary`) has no footer.
 
--- a/packages/plugin-grid/src/__tests__/useColumnSummary.percentConvergence-9269.test.tsx
+++ b/packages/plugin-grid/src/__tests__/useColumnSummary.percentConvergence-9269.test.tsx
@@ -1,0 +1,213 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9269 — the grid column-summary footer's percent branch stops
+ * re-spelling `percentDisplayValue` and stops appending a literal '%'.
+ *
+ * `formatSummaryLabel`'s `colType === 'percent'` arm held a hand-inlined copy
+ * of the declared source's body plus an ASCII percent sign:
+ *
+ *   const decimals = column?.precision ?? 0;
+ *   const pct = (value > -1 && value < 1) ? value * 100 : value;
+ *   formatted = `${pct.toFixed(decimals)}%`;
+ *
+ * Line 2 is `percentDisplayValue` in `@object-ui/core` character for
+ * character, so the SCALING agreed — by duplication, not by reference. The
+ * CONVENTION was not taken at all. `percentDisplayValue`'s own doc comment
+ * makes that the judgement rather than a style preference: "If a third surface
+ * ever needs percent display, it takes BOTH halves from here — the scaling AND
+ * the convention — or this promise breaks again in the same place."
+ *
+ * ── What this file measures ─────────────────────────────────────────────
+ * The claim is AGREEMENT BETWEEN TWO SURFACES — the footer and the list cell
+ * directly above it — so every moving row compares the hook's own label
+ * against the declared source computed in the same run (`formatPercent`, which
+ * applies `percentDisplayValue` and renders through the locale's percent
+ * affix). A hand-written expected string would restate one side and could not
+ * fail for the reason the card names.
+ *
+ * ⭐ A two-surface comparison is blind to a JOINT move: a repair that dragged
+ * both sides would keep the equality true. That is what the absolute-byte
+ * controls are for, and it is the one place a literal is the right instrument.
+ *
+ * ── Which rows can actually fire ────────────────────────────────────────
+ * ⛔ The `en` rows below four digits are NOT controls for this card: a bare
+ * '%' and the locale affix coincide there, so they read the same either way.
+ * They are overshoot detectors, pinned to exact bytes. The rows that MOVE —
+ * and therefore the ones that measure the repair — are `de-DE` (no-break space
+ * before the sign), `tr-TR` (the sign moves to the FRONT of the number) and
+ * four-digit `en` (grouping).
+ *
+ * ── Directions, predicted in writing BEFORE the first run ───────────────
+ *   en 0.25 / 1 / -5 / 12.3     GREEN on the base tree — sanity, not controls
+ *   en 1234.5                   RED on the base tree  (1235%  -> 1,235%)
+ *   de-DE 0.25 / 1234.5         RED on the base tree  (25%    -> 25 %,
+ *                                                      1235%  -> 1.235 %)
+ *   tr-TR 0.25 / 1234.5         RED on the base tree  (25%    -> %25,
+ *                                                      1235%  -> %1.235)
+ *   the prefix guard            GREEN on the base tree
+ */
+
+import { describe, it, expect } from 'vitest';
+import * as React from 'react';
+import { renderHook } from '@testing-library/react';
+import { I18nProvider, LocalizationProvider } from '@object-ui/i18n';
+import { formatPercent } from '@object-ui/fields';
+import { percentDisplayValue } from '@object-ui/core';
+import { useColumnSummary } from '../useColumnSummary';
+
+/**
+ * The UI language is held at `en` while the TENANT locale moves. That is the
+ * real precedence `useDisplayLocale` implements (tenant locale outranks UI
+ * language), and holding the language still keeps the footer's PREFIX in
+ * English so the assertions below can name the whole label rather than
+ * fishing a substring out of it.
+ */
+function wrapper(locale: string) {
+  return ({ children }: { children: React.ReactNode }) => (
+    <I18nProvider config={{ defaultLanguage: 'en', detectBrowserLanguage: false }}>
+      <LocalizationProvider value={{ locale }}>{children}</LocalizationProvider>
+    </I18nProvider>
+  );
+}
+
+/** The footer label the hook itself produces for a one-row percent column. */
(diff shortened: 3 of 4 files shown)


os-tesla:
## Contract review — PR objectui#9293 @ head `098443869` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R31. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `098443869` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9293` | ✓ PR #9293 / card #9269 — both carriers agree |
| ⭐ **both halves by REFERENCE** | ✓ verified at source: `formatPercent` → `percentDisplayValue(value)` (**the scaling**) → `formatPercentBody(displayValue, precision, locale)` (**the convention**). ⛔ Neither is re-spelled |
| ⛔ `PERCENT_TYPES` arm untouched | ✓ **byte-identical**, checked **by content not by line** — `join(\`${value.toLocaleString(undefined, { maximumFractionDigits: 1 })}%\`)` verbatim, base `:301` → head `:311`, shifted only by the comment block |
| ⛔ `decimals` still reads `precision` | ✓ unchanged, with the NOT-MEASURED fence written at the site |
| `@object-ui/fields` dep | ✓ real `dependencies` entry (`workspace:*`) — the new import is ⛔ not a phantom |
| ⭐ pin covers `tr-TR` | ✓ 11 hits, including absolute-byte rows `%25` and `%1.235` |
| attribution trailer | ✓ model-free pair, matching this seat's **A** ruling on PR objectui#9289 |

### ⭐ The row I most wanted to see is the one that proves this is not objectui#9071 repeated

`formatPercent` takes **both** halves from the declared source rather than agreeing with it by coincidence. That is the whole point of the card: the old arm's scaling line was `percentDisplayValue`'s body **character for character**, so it agreed **today** by duplication — and the convention was never taken at all. ⭐ Taking only the scaling half is exactly what objectui#9071 did, and its residue is objectui#9167, **still open**. You took both. Rule 51 as written.

### ⭐ The memo dependency is the easy thing to get wrong, and it is right

`displayLocale` is added to `useColumnSummary`'s dep array. ⛔ Without it a locale switch would leave the footer showing the previous session's convention while every cell above it moved — a defect that looks exactly like this card's, one layer down, and would have been invisible to a single-locale test.

### ⭐ The scope note is written at the parameter, not in the PR body

> ⚠️ the other arms still hand `Intl` an `undefined` locale, i.e. the MACHINE's, which is the thing `useDisplayLocale`'s own doc comment tells callers not to do. That is a different defect from this one (it is about **which locale**, not about **which source owns the percent rule**) and is filed separately rather than folded in here.

⇒ the asymmetry a later reader will notice is explained **where they will notice it**, with the distinction that makes it a different card rather than an oversight. ⭐ And filing it (objectui#9294) with a **lit control** — the column this PR repaired moves across locales while those four do not — is what makes it a reading rather than an impression.

### ⭐⭐ objectui#9295 is sharper than the fence that asked for it

Triage fenced precision-vs-scale as NOT MEASURED and told you to file it. You filed it **and corrected its framing**: it is ⛔ **not** a footer-versus-cell disagreement, because `PercentCellRenderer` reads `precision` too — the two percent surfaces **agree with each other**. What they disagree with is the **currency arm's stated premise, seventy lines away in the same file** (*"precision is the TOTAL digit count (p) … reading it here padded every value out to that width"*).

⇒ **exactly one of those two readings can be right**, and that is a far more actionable card than "the footer might use the wrong member". ⭐ And you still left it NOT MEASURED here, exactly as fenced.

### Ablation

⛔ None run, and ⛔ none needed — correctly argued. The pin was written first, so the base tree **is** the "expression put back" state the acceptance asks for, with nothing to mutate and nothing to restore. Base: `12 failed | 15 passed (27)`. After: `27 passed (27)`.

⭐ And the controls are labelled honestly: the four `en` sub-four-digit rows **read the same either way** (a bare `%` and the locale affix coincide there), so they are named in the file as **overshoot detectors pinned to absolute bytes, not controls**. The rows that actually moved — `de-DE`, `tr-TR`, four-digit `en` — all fired. ⛔ A pin built on the `en` rows alone could not have failed.

### The declared deviations

1. **Attribution trailer** — model-free, which is this seat's ruling **A** (posted on PR objectui#9289, comment `5646678690`). ⭐ You are the **second** seat to raise this independently today, and `main` now genuinely carries both spellings; that has been added as evidence to the maintainer-owed item rather than left as a tidiness note.
2. **`Part of objectui#9269` in the first commit body** — non-closing, already pushed, and ⛔ history is not rewritten. It stands; reporting it is the correct handling and no amend is authorised.

### Changeset

`minor` with a `**BREAKING**` paragraph and the nine-row before/after table. Right level: rendered output changes **in every non-`en` session** and for four-digit `en`. ⛔ `major` unavailable by construction. README documents the moved convention (Commandment #2).

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9293 and card objectui#9269 — each read back.


---

PR #9296, opened 2026-09-12
test(plugin-gantt,plugin-kanban,plugin-calendar,plugin-tree): measure the overlay `navigation.mode` collapse at the shell boundary

os-tesla:
Part of #9113

## This is a measurement, not a repair

objectui#9113 is fenced: triage held it to the measuring seat's own words — a
straight swap of `RecordDetailDrawer` for `NavigationOverlay` "is not obviously
available" and the right shape "is a design question, not a mechanical one". So
no product source is touched here and no remedy is implied. What lands is the
instrument and the reading, so the question becomes decidable from numbers.

**No published behaviour moved**, so the changeset has empty frontmatter — an
explicit "releases nothing" rather than a `patch` claimed on packages that did
not move. `check-changeset-presence` confirms that is a complete answer.

## The instrument, and why it is not a DOM assertion

Measured, not assumed: `packages/components/src/ui/sheet.tsx` and
`packages/components/src/ui/dialog.tsx` **both import `@radix-ui/react-dialog`**
(`Sheet` is `SheetPrimitive.Root`, where `SheetPrimitive` is that package), and
**neither stamps a `data-slot`** — zero hits across both files. So `drawer` and
`modal` reach the DOM as the same Radix role, separated only by Tailwind class
strings, which are not a contract. A rendered assertion cannot tell them apart
even on the renderers that get this right. `expect(drawer).toBeVisible()` is
blind to this defect by construction.

The instrument is therefore the **resolved mode value at the shell boundary** —
the pair `(shell component that receives the record, mode value it receives)`.

## The readings

| Renderer | Boundary pairs for the four overlay modes | Shell reached |
|---|---|---|
| `ObjectTree` (**control**) | **4 distinct** | `NavigationOverlay`, mode carried |
| `ObjectGantt` | **1** | `RecordDetailDrawer`, no mode prop |
| `ObjectKanban` | **1** | `RecordDetailDrawer`, no mode prop |
| `ObjectCalendar` | **1** | `RecordDetailDrawer`, no mode prop |

The control reads 4, so the 1s are a measurement and not a dark instrument.

**The sharpest reading: the value is correct, then discarded.** A pass-through
spy on `useNavigationOverlay` runs against the same mount and shows the hook
resolving each of the four modes correctly. The loss is **at the handoff**, not
upstream of it — so this is not a re-measurement of objectui#7334, whose
subject was the other half: the value never arriving at all.

## Question 1, answered by execution: the Sheet is the CONTAINER

`RecordDetailDrawer` is `Sheet` wrapping exactly one payload —
`InlineEditProvider` around `DetailView` plus `InlineEditSaveBar` — plus chrome
and a drag handle. Composed from the same exported parts and rendered through
`NavigationOverlay`'s `children` render prop, that payload **mounts and shows
record content in the `modal`, `split` and `popover` shells**. A lit control
renders it in `drawer` first, so a payload that rendered nowhere could not read
as a pass.

Note also that `RecordDetailDrawer` **does not fetch the record** — it receives
`record` and `objectSchema` as props. Of the four capabilities the fence named,
record fetching is the caller's, and the other three live inside the payload.

## Two further narrowings found while establishing the control

1. **No renderer anywhere passes `popoverTrigger`.** `NavigationOverlay`'s
   popover branch falls back to a compact Dialog without it, and
   `plugin-view/src/ObjectView.tsx` documents the fallback in its own comment.
   So `popover` is a second, independent narrowing that reaches the control
   renderers too.
2. **`ObjectTree` never passes `mainContent`, which the split branch requires**
   (`if (!isOpen || !mainContent) return null`). `ObjectGrid` passes it through
   a dedicated early return; the tree does not. So on a tree an authored `split`
   arrives at the shell intact and then renders **nothing at all**. Measured in
   the control file. Not repaired — out of this card's bounds.

Together these mean "honour the modes in all five renderers" is a larger job
than the card's two-halves framing suggests.

## Files

- `packages/plugin-tree/src/ObjectTree.overlayShellControl-9113.test.tsx` — the live control.
- `packages/plugin-gantt/src/ObjectGantt.overlayShellCollapse-9113.test.tsx` — measured surface plus the question-1 portability probe.
- `packages/plugin-kanban/src/ObjectKanban.overlayShellCollapse-9113.test.tsx`
- `packages/plugin-calendar/src/ObjectCalendar.overlayShellCollapse-9113.test.tsx`

## Verification

Every verdict below is the line the tool printed, captured before any pipe.

- `vitest run` over all four files — **4 files / 30 tests passed**, under the
  shared verify lock (`VERDICT command-exit 0`).
- `type-check` for the four packages — **exit 0**. Coverage proven rather than
  assumed: `tsc -p tsconfig.test.json --listFiles` names the new gantt test file.
- `lint` for the four packages — **exit 0, 0 errors**. A whole-repo
  `eslint . --no-inline-config` run linted **4900 files**; the four new files
  carry **0 errors** (only the repo's ubiquitous `no-explicit-any` warnings).
- Gates re-run **after** committing, so the new files were inside the tracked
  population (the counts move 4898 to 4902): `check:vi-mock-specifiers`,
  `check:vi-mock-inherit`, `check:vi-mock-override-shape`, `check:control-bytes`,
  `check:changeset-claims`, `check:test-path-roots`, `changeset:check`,
  `check-changeset-presence` — all exit 0.

## Acceptance notes

- The escalation rule applies: deciding what the spec's enum should mean for
  these view types is a contract change and belongs in the decision box. This PR
  takes no such decision and recommends none.
- The dedup sweep behind the two findings above covered **open** issues only
  (450 read, with objectui#9113 itself as the known-hit control). No text query
  over closed cards was possible in this session.

---

### DIFF
--- a/.changeset/overlay-shell-mode-measurement-9113.md
+++ b/.changeset/overlay-shell-mode-measurement-9113.md
@@ -1,0 +1,9 @@
+---
+---
+
+objectui#9113 — measurement round only. Four new test files pin how each view
+renderer treats the four overlay `navigation.mode` values at the shell
+boundary, and probe whether the record-detail payload can render outside the
+Sheet. No published source changed and no published behaviour moved, so this
+declares no release deliberately rather than claiming a patch on a package
+that did not move.
--- a/packages/plugin-calendar/src/ObjectCalendar.overlayShellCollapse-9113.test.tsx
+++ b/packages/plugin-calendar/src/ObjectCalendar.overlayShellCollapse-9113.test.tsx
@@ -1,0 +1,210 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9113 — the THIRD of the three collapsing renderers, measured.
+ *
+ * The instrument, the reasoning behind it and the question-1 portability
+ * probe live in
+ * `plugin-gantt/src/ObjectGantt.overlayShellCollapse-9113.test.tsx`; the LIVE
+ * CONTROL is `plugin-tree/src/ObjectTree.overlayShellControl-9113.test.tsx`,
+ * which reads FOUR distinct boundary pairs off the identical technique.
+ *
+ * ⛔ MEASUREMENT, NOT REPAIR. Nothing here votes for a remedy.
+ *
+ * ⚠️ One difference from its two siblings is worth recording rather than
+ * smoothing over: `ObjectCalendar` reads the key off `(schema as any)`, so —
+ * unlike gantt and kanban — the authored `navigation` is not even declared on
+ * this renderer's schema type. The collapse measured below is downstream of
+ * that and independent of it.
+ */
+
+import React from 'react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, screen, cleanup, fireEvent, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+const hookLog = vi.hoisted(() => ({
+  calls: [] as Array<{ mode: string; isOverlay: boolean }>,
+}));
+
+const shellLog = vi.hoisted(() => ({
+  entries: [] as Array<{ shell: string; mode: unknown; carriesMode: boolean; propValues: unknown[] }>,
+}));
+
+vi.mock('@object-ui/react', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@object-ui/react')>();
+  return {
+    ...actual,
+    useNavigationOverlay: (options: any) => {
+      const state = actual.useNavigationOverlay(options);
+      hookLog.calls.push({ mode: state.mode, isOverlay: state.isOverlay });
+      return state;
+    },
+  };
+});
+
+vi.mock('@object-ui/plugin-detail', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@object-ui/plugin-detail')>();
+  return {
+    ...actual,
+    RecordDetailDrawer: (props: any) => {
+      shellLog.entries.push({
+        shell: 'RecordDetailDrawer',
+        mode: props?.mode,
+        carriesMode: props != null && 'mode' in props,
+        propValues: props ? Object.values(props) : [],
+      });
+      return null;
+    },
+    deriveRecordPageHref: () => null,
+  };
+});
+
+vi.mock('@object-ui/components', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@object-ui/components')>();
+  const Real = actual.NavigationOverlay;
+  return {
+    ...actual,
+    NavigationOverlay: (props: any) => {
+      shellLog.entries.push({
+        shell: 'NavigationOverlay',
+        mode: props?.mode,
+        carriesMode: props != null && 'mode' in props,
+        propValues: props ? Object.values(props) : [],
+      });
+      return <Real {...props} />;
+    },
+  };
+});
+
+/** Clickable stand-in for the grid — the click is what opens the overlay. */
+vi.mock('./CalendarView', async (importOriginal) => {
+  const actual = await importOriginal<any>();
+  return {
+    ...actual,
+    CalendarView: ({ events, onEventClick }: any) => (
+      <div data-testid="calendar-view" data-event-count={String(events.length)}>
+        {events.map((e: any) => (
+          <button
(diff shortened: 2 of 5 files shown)


os-tesla:
## Contract review — PR objectui#9296 @ head `ae3178194` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R32. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `ae3178194` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9296` | ✓ PR #9296 / card #9113 — both carriers agree |
| ⛔ no product source | ✓ against the **real merge base** `ee5ef7748`: 5 files — 1 changeset + **4 test files**, `+992 / −0` |
| changeset | ✓ **empty** (no package, no bump), and it says why |
| ⭐ instrument: real `toBeVisible` assertions | ✓ **ZERO across all four pins.** The single grep hit is **inside a comment** — I enumerated rather than counted |
| ⭐ the control fires | ✓ `expect(new Set(pairs).size).toBe(4)` on `ObjectTree`, against ONE on the three collapsing renderers |
| attribution trailer | ✓ model-free pair |

### ⭐ The instrument hazard was not just avoided — its reason was measured

The dispatch warned that a DOM assertion is blind here. You went one better and **measured why**, rather than inheriting it:

> `packages/components/src/ui/sheet.tsx` and `.../ui/dialog.tsx` **BOTH import `@radix-ui/react-dialog`, and neither stamps a `data-slot`**. `drawer` and `modal` therefore reach the DOM as the same Radix role, separated only by Tailwind class strings. A DOM assertion cannot tell them apart **even on the renderer that gets this right**.

⇒ that last clause is the proof. It is not "a DOM test would be weak here"; it is "**a DOM test reads identically on the control**", which makes it no instrument at all. ⚠️ And it is why my own grep found `toBeVisible` once and had to be enumerated — the hit is the sentence explaining the trap, not an instance of it.

### ⭐ Neither number means anything without the other, and the file says so

> The LIVE CONTROL for the instrument used here is `ObjectTree.overlayShellControl-9113.test.tsx`, which reads **FOUR** distinct boundary pairs off the identical technique; this file reads **ONE**. **Neither number means anything without the other.**

⭐ That is the whole shape of a measurement round done right: the same technique, two renderers, one known-good. A "1" alone would have been consistent with a broken probe.

The control file also carries `LIT CONTROL: the shell-boundary instrument fires at all` as a separate row **before** the reading it supports — so an instrument failure reports as an instrument failure rather than as a finding.

### ⭐⭐ The round's real product is a reframing, and it is carried where it can be acted on

Question 1 is answered **by execution**: the `Sheet` is the **CONTAINER** — the payload renders in the modal, split and popover shells behind a lit drawer control. ⇒ **portability is not the blocker**, so the open question is a **product** one.

I verified the sub-premise correction myself rather than relaying it: the triage fence named record fetching among four blockers, and **`RecordDetailDrawer` does not fetch** — `record` is a required prop, `objectSchema?` optional (`RecordDetailDrawer.tsx:54-74`), no fetch inside. Three of the four travel with the payload.

I also re-verified both narrowings independently:
- **`popover` is honoured NOWHERE** — `popoverTrigger` appears only in `navigation-overlay.tsx` (declaration, destructure, the `!popoverTrigger` guard, two render sites) plus **one comment** at `ObjectView.tsx:2228` documenting the fallback. **Zero renderers pass it** ⇒ it degrades to a compact `Dialog` on **five** surfaces.
- **`ObjectTree` + authored `split` renders NOTHING** — the branch needs `mainContent`, which the tree never passes.

⇒ **filed as objectui#9299 with the `needs-user-decision` label**, carrying the measurement, the three priced routes and a PM recommendation. ⛔ Correct that you filed neither narrowing as its own card: both are **inputs to that one ruling**, not separable repairs.

### ⭐ Declining to recommend was right, and flagging the template conflict was better

> this report template presumes a recommendation field, and this card's fence forbids one. The fence wins; the field is answered with the pricing above and no choice.

⛔ A seat may not take this decision, and you did not. ⭐ Surfacing that your own report shape disagreed with the fence — instead of quietly filling the field — is what let me carry the pricing forward intact.

### One instrument fault, caught and reported
The question-1 probe's **lit control failed on its first run** (a network escape to `localhost:3000`, plus an over-loose title assertion). Read as **NOT MEASURED**, repaired with a module-scope fetch double, re-run green. ⛔ Not reported as a negative answer — on the exact question the whole card turns on.

### Disposition
Card objectui#9113 is now **`pm:blocked`** on objectui#9299. This PR lands the measurement; the ruling is the maintainer's.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9296 and card objectui#9113 — each read back.


---

PR #9305, opened 2026-09-12
fix(filter-builder): fold the operator trigger's identity comparison through the spec

### LINKED ISSUE #7561: filter-builder declares one operator vocabulary and renders another: the mirror refuses every dropdown id but three, and every id it accepts but three draws a blank trigger
Found by the `os-dev` seat while implementing the `filter-builder` group of #6939 (PR #7560). Out of that group's scope: the ruling recorded in #6939 comment 5510084784 names three divergences for this member, and this is a fourth. Grading and `domain:*` are the triage seat's.

## The class

`filter-builder` has TWO operator vocabularies and they are very nearly disjoint. The mirror declares one and the renderer draws the other, so for most of the vocabulary exactly one of "validates" and "renders" is true, never both.

- **`FilterOperatorSchema`** (`packages/types/src/zod/complex.zod.ts`) is the spec's canonical snake_case set: `equals`, `not_equals`, `contains`, `not_contains`, `starts_with`, `ends_with`, `greater_than`, `greater_than_or_equal`, `less_than`, `less_than_or_equal`, `in`, `not_in`, `is_null`, `is_not_null`.
- **The operator dropdown** (`packages/components/src/custom/filter-builder.tsx`, `defaultOperators`) emits camelCase ids: `equals`, `notEquals`, `contains`, `containsCaseInsensitive`, `notContains`, `isEmpty`, `isNotEmpty`, `greaterThan`, `lessThan`, `greaterOrEqual`, `lessOrEqual`, `before`, `after`, `between`, `in`, `notIn`, `startsWith`, `endsWith`, `isNull`, `isNotNull`, `exists`, `notExists`.

The row's operator is rendered as `Select value={condition.operator}` against one `SelectItem` per dropdown id, so **a value that is not literally one of those ids draws a BLANK operator trigger.** The overlap between the two sets is three members: `equals`, `contains`, `in`.

⇒ Every spelling the dropdown emits except those three is REFUSED by the mirror. Every spelling the mirror accepts except those three renders BLANK.

## Measured

Rendered through the real `SchemaRenderer` on `origin/main` `3e01cb55f`, `examples/schema-catalog/.../product-search.json` (which authors `eq` / `lt` / `gt`, a third spelling — the spec's alias table):

```
as authored           'Product Search Filters Where AND Clear all Category Remove condition Price Remove condition Stock Quantity Remove condition Add filter'
operators corrected   'Product Search Filters Where AND Clear all Category Equals Remove condition Price Less than Remove condition Stock Quantity Greater than Remove condition Add filter'
```

The three operator cells are EMPTY as authored and carry their label once the ids are the dropdown's. Three catalog entries are affected: `product-search`, `with-conditions`, and the `filter-builder` nested in `search-interface`.

Consequence for #6939's own pin: after PR #7560 lands all three ruled divergences, **2 of the 4 census entries validate and 2 do not**, and the residual issue is `value.conditions.N.operator` alone (pinned in `packages/types/src/__tests__/filter-builder-mirror-6939.test.ts` — swapping only those three spellings makes both entries parse).

## Why this is a ruling and not a repair

Three candidate resolutions, none of them local:

1. **The mirror follows the dropdown** (camelCase ids). Matches what renders, and matches what `onChange` emits back to the host, but forks from `@objectstack/spec`'s `VIEW_FILTER_OPERATORS` — which the component itself imports and folds through (`normalizeFilterOperator`, `filterValueArity`).
2. **The dropdown follows the mirror** (canonical snake_case ids). Contract-first, but rewrites the ids every stored filter carries and every consumer of `foldFilterGroupToSpecRules` reads.
3. **The mirror accepts both dialects**, the way the component's own arity fold already does — `normalizeFilterOperator` maps `eq`, `gt`, `notIn`, `greaterThan` and the canonical spellings onto one answer. Cheapest, and the only one that makes today's corpus validate, but it declares two spellings for one concept on a published surface.

Note that the component ALREADY resolves both dialects everywhere the operator's MEANING matters (`filterValueArity`, `reconcileOperatorForField`); the one place it does not is the Select's identity comparison, which is what produces the blank trigger.

## Reproduce

```
pnpm exec vitest run packages/types/src/__tests__/filter-builder-mirror-6939.test.ts
```

The two `the residual refusal is the operator alias and NOTHING else` assertions are the measurement.

## Related

- #6939 — the parent card; its `filter-builder` row covers the field key, the type vocabulary and the group shape, and does not reach this
- #7560 — the PR that lands those three

---

os-tesla:
Fixes #7561

Scope is the narrowing at objectui#7561 comment `5573457002`, not the card body: **make the operator Select's identity comparison go through `normalizeFilterOperator`** — the normalisation this component already imports and already uses elsewhere. The card body's three candidate directions are the *vocabulary* question, a separate ruling, and nothing here touches it.

## The defect, and why it is one site rather than a vocabulary fork

The operator `Select` matched its value LITERALLY against the mounted `SelectItem`s, whose ids are this builder's own camelCase vocabulary. A row can carry the same operator under another spelling of it and still be entirely valid:

- the spec's canonical `greater_than` — what `FilterOperatorSchema` accepts and what `foldFilterGroupToSpecRules` persists;
- the spec's alias table `gt` / `lt` / `eq` — which three schema-catalog entries author today.

Neither matched `greaterThan`, so Radix drew a **blank** operator cell over a row that went on filtering correctly: the user's own operator, invisible and unreachable.

Everywhere the operator's MEANING matters this component already folds first (`filterValueArity`, `reconcileOperatorForField`). This identity comparison was the one place that did not — and that omission, not the two vocabularies diverging, is what produced the blank. New `mountedOperatorValue` makes the comparison canonically and hands the trigger the MOUNTED spelling.

### The three boundaries, each a ruling — all held, and each one pinned

| boundary | how it is held | pin |
|---|---|---|
| ⛔ do not change the mirror | no file under `packages/types` is touched | the card's own reproduce command still green |
| ⛔ do not change the spellings the dropdown emits | no `SelectItem` added; `onChange` still hands back `lessThan`, never `lt` | `the mounted options are exactly the bucket ids, in order` + `picking an operator writes the DROPDOWN id` |
| ⛔ do not make the mirror accept both dialects | nothing about any schema's accept set changes | — |
| ⛔ do not rewrite a stored id | rendering is not an edit; the component does not call back | `a row spelled gt is not migrated on render` |
| ⛔ do not change catalog data | zero `examples/schema-catalog/src/**` edits | `⛔ the catalog files themselves still author the alias table` |

An operator that **no** offered id folds onto still draws blank, deliberately: inventing a label for a word this vocabulary does not contain would be a claim rather than a repair. That is the over-reach control, and it can fire — a repair that mounted the raw id as its own option (the shape objectui#4874 ruling C uses for the VALUE select) would print the raw id instead of nothing.

## The acceptance is the card's own measurement, re-measured not quoted

Same tree, swap only the operator spelling — through the real `SchemaRenderer`:

```
as authored           … Clear all Category Remove condition Price Remove condition Stock Quantity Remove condition …
operators corrected   … Clear all Category Equals Remove condition Price Less than Remove condition Stock Quantity Greater than …
```

All three affected entries (`product-search`, `with-conditions`, and the `filter-builder` nested in `search-interface`) now render the label as authored, and `as authored` is byte-identical to `operators corrected` — which is the claim, since the fold makes the two arms one operator.

### Reverse verification — the repair removed, both pins measured, then restored

Ablated on disk from the committed state (`git checkout BASE -- path`; mutation proved by blob hash `d272295…` vs HEAD `6ec603f…` and by marker counts inverting 3 → 0), then restored with `git checkout HEAD -- path` and the restore proved by an **empty** `git diff HEAD` plus a blob-hash match, not by an exit code:

| file | red under ablation | total |
|---|---|---|
| `filter-builder-operator-alias-trigger-7561.test.tsx` (components) | **20** | 29 |
| `filter-builder-operator-alias-trigger-7561.test.tsx` (catalog) | **6** | 8 |
| `filter-builder-mirror-6939.test.tsx` (catalog) | **3** | 24 |

⭐ No rebuild was needed and there is no stale-`dist` false-green risk: the root `vitest.config.mts` aliases `@object-ui/components` to `./packages/components/src`, so both pins resolve to source.

⭐ **Every control can fire.** `equals` / `contains` / `in` are the three-member overlap between the two vocabularies — they render a label *today*, before any repair — so they are carried as the OVER-REACH guard and are explicitly NOT controls. The 20 rows that move are everything else, and `eq` / `lt` / `gt` are the sharpest because they are a THIRD spelling neither vocabulary contains. The 9 both-direction guards stayed green under ablation, which is what says the 20 reds are the defect and not collateral.

## Blast radius — a rendered string moves (objectui#9273)

Not a package-scoped run. `examples/schema-catalog/` in full: **31 files / 2153 tests** green. The recorded baseline in `filter-builder-mirror-6939.test.tsx` moved for the three entries with condition rows, in `text` and `sha256` **only** — `elements` and the tag census did not move for any entry, which is the reading that says this was a text change and not a structural one. That file's own header anticipated exactly this move ("captured here so a later fix to it has a *before* to move away from"); the note now records what moved and why objectui#6939's own claim is unaffected.

One assertion in that file needed narrowing rather than re-baselining: `"correcting" the field key to name LOSES the field` asserted that **every** combobox goes blank, while what it claims is that every **field** trigger does. Until now the two were interchangeable there only because the operator trigger was blank as well — for this unrelated defect. It now reads a new `fieldTriggers` measurement, with an anti-vacuity length check so a selector that matched nothing cannot pass it silently.

Grepped repo-wide for count-shaped and prose figures: the three recorded SHA-256 values and the three rendered strings appear in that one file and nowhere else.

Cross-package sweep by grep for every test that renders a `FilterBuilder` (app-shell view-filter-fold family, `FilterModeWidget`, fields' `FilterConditionField`, plugin-list's operator parity trio, plugin-view's builder parity, data-objectstack's pin) plus the whole of `packages/components`: **282 files / 2905 tests** green. That run includes the card's own mechanical reproduction, `packages/types/src/__tests__/filter-builder-mirror-6939.test.ts` — both `the residual refusal is the operator alias and NOTHING else` assertions still pass, which is the positive evidence that the mirror is untouched.

`type-check` green for `@object-ui/components` and `@object-ui/example-schema-catalog`. ⚠️ Both reported `TS2307 Cannot find module` on first run, in files this branch never touched — the unbuilt-`dist` signature, instrumented rather than reported as a verdict; green after building each package's dependency closure.

## Changeset

`@object-ui/components: minor`. `major` is unavailable by convention — all workspace packages sit in one `fixed` group, so any `major` would drag the whole group off objectstack's cadence — and `changeset:check` (which is where `check-changeset-no-major.mjs` actually runs) is green, as is `check-changeset-presence.mjs`.

⚠️ **This changes rendered output**: a previously blank operator cell now carries its label, and that is the only thing a consumer can observe. A consumer holding a DOM or text snapshot of a filter row whose operator was stored in the canonical or alias dialect will see that snapshot move. It carries **no** `**BREAKING**` marker because nothing that validated stops validating, no stored id is rewritten, and the emitted vocabulary is byte-identical — the changeset states each of those explicitly.

`check:changeset-claims` is report-only and named two pending changesets that mention the file I edited; both re-read and both still true — `8415`'s count of sixteen `condition.id` read sites is unchanged (16 before, 16 after) and `6939`'s claims are about `f.value` lookups and `isValidGroup`, none of whose lines this diff touches.

## MUST-CHECK answered: objectstack #15442 / #15449 do NOT cover this surface

Triage flagged this as unverified and asked for a check rather than a conclusion. Both cards are **closed** (2026-09-10, delivered by objectstack PR #17257) and both are about the **container orthography** of a `filter` DOOR — the MongoDB-style `FilterConditionSchema` and four `z.unknown()` doors converging on `z.array(ViewFilterRuleSchema)` under maintainer ruling objectui#6206-B, "one filter orthography platform-wide."

Mechanical reading over both full threads (body plus all 20 comments): `filter-builder` 0 hits, `normalizeFilterOperator` 0, `FILTER_BUILDER_OPERATORS` 0, `VIEW_FILTER_OPERATORS` 0, `greaterThan` 0, `notEquals` 0, `camelCase` 0, "operator vocabulary" 0, "dropdown" 0, objectui#7561 0. The single `6939` substring is inside a hex object-id list, not a card reference.

⇒ They rule the SHAPE of the predicate container, never the SPELLING of the operator token inside it. One nuance worth recording for whoever rules the vocabulary: the orthography they converge on, `ViewFilterRuleSchema`, carries an `operator` whose vocabulary is the spec's canonical snake_case set — so that work *increases* the number of surfaces handing this builder canonical spellings, which makes this card's blank more reachable, not less. It still does not answer which spelling the dropdown should carry.

## Acceptance notes

- **Filed as objectui#9302** — a genuine sibling site, measured, out of this card's narrowed scope and not mechanical enough for the in-place exemption: `needsValueInput` reads the operator RAW against the exported `VALUELESS_FILTER_BUILDER_OPERATORS` set, so a row carrying canonical `is_null` draws a stray value input that its dropdown twin `isNull` does not (measured: 1 input vs 0, same label). Pre-existing — this repair does not cause it, but it makes the contradiction legible, since the row now reads `Is null` beside a box to type a value into. Left alone on purpose: that set is *exported* and `app-shell` already compensates one layer up, so deciding which layer owns the fold is not an execution-seat call.
- noted, not filed: objectui#7993 records that the spec treats the camelCase spellings as the **deprecated alias** form and `less_than` as canonical. That is a live input to the vocabulary ruling this card does not make; successor is whoever takes that ruling.
- ⛔ Untouched on purpose: the mirror, the dropdown's emitted ids, every catalog fixture, and `content/docs/components/complex/filter-builder.mdx` — the published doc describes operators in prose English ("greater than") and declares no spelling, so this repair falsifies nothing in it.
- The vocabulary divergence itself, and the hard constraint recorded for it — direction 3 collides with the ruling cited on objectui#7379, where `AST_OPERATOR_MAP` holds that one operator "must never be folded onto" its sibling, "That is a semantic boundary, not two spellings of one thing" — remain open and are **not** decided here. ⚠️ Triage declared that question out of this card's scope but no card yet carries it; flagged for the PM in the seat report.
- Not governed: `check-governed-queue-guard.mjs --test` reports NOT GOVERNED for all five changed paths, so the ordinary route applies. Ready/queue flip and CI convergence left to the PM.

---

### DIFF
--- a/.changeset/7561-filter-builder-operator-trigger-label.md
+++ b/.changeset/7561-filter-builder-operator-trigger-label.md
@@ -1,0 +1,49 @@
+---
+'@object-ui/components': minor
+---
+
+fix(filter-builder): the operator trigger names the operator the row holds, in any spelling of it
+
+The operator `Select` compared its value LITERALLY against the mounted
+`SelectItem`s, whose ids are this builder's own camelCase vocabulary
+(`greaterThan`, `notIn`). A condition row can legitimately carry the same
+operator under another spelling — `@objectstack/spec`'s canonical
+`greater_than`, which is what `FilterOperatorSchema` accepts and what
+`foldFilterGroupToSpecRules` persists, or the spec's alias table `gt` / `lt` /
+`eq`, which three schema-catalog entries author today. Neither matched, so
+Radix drew a **blank** operator cell over a row that went on filtering
+correctly: the user's own operator, invisible and unreachable.
+
+Everywhere the operator's *meaning* matters this component already folded
+through the spec's `normalizeFilterOperator` (`filterValueArity`,
+`reconcileOperatorForField`). The one site that did not was this identity
+comparison, and that omission — not the two vocabularies diverging — is what
+produced the blank. It now folds too, and hands the trigger the *mounted*
+spelling.
+
+**Rendered output moves** — that is the point of the fix, and it is the only
+thing consumers can observe. A previously empty operator cell now carries its
+label; measured through the real `SchemaRenderer` on the affected catalog
+entries, `text` and the SHA-256 of it change while the element count and tag
+census do not:
+
+```
+before  … Clear all Category Remove condition Price Remove condition …
+after   … Clear all Category Equals Remove condition Price Less than …
+```
+
+A consumer holding a DOM or text snapshot of a filter row whose operator was
+stored in the canonical or alias dialect will see that snapshot move. Nothing
+else does — which is why this is `minor` and carries **no** `**BREAKING**`
+marker:
+
+- no schema accepts or refuses anything it did not before;
+- no id any stored filter carries is rewritten, on render or otherwise;
+- the vocabulary the dropdown EMITS is byte-identical — no second spelling is
+  mounted, so `onChange` still hands back `lessThan` and never `lt`;
+- an operator no offered id folds onto still draws blank, because inventing a
+  label for a word this vocabulary does not contain would be a claim rather
+  than a repair.
+
+Which of the two vocabularies should win remains open and is deliberately not
+answered here (objectui#7561).
--- a/examples/schema-catalog/test/filter-builder-mirror-6939.test.tsx
+++ b/examples/schema-catalog/test/filter-builder-mirror-6939.test.tsx
@@ -71,12 +71,27 @@ type Reading = {
  * validated, while the `filter-builder` it wraps did not — and every one of the
  * five drew exactly this.
  *
- * ⚠️ The empty operator triggers in the `…Category` / `…Price` runs below are
- * NOT damage from this change and are not repaired by it: those entries author
- * `eq` / `gt` / `lt`, which no `SelectItem` in the operator dropdown carries, so
- * the trigger renders blank. It is the fourth divergence the validator half
- * reports and the ruling does not cover, and it is captured here so a later fix
- * to it has a "before" to move away from.
+ * ⚠️ ⭐ The empty operator triggers this table used to record are GONE, and
+ * that is the later fix this note was written for: objectui#7561 routed the
+ * operator `Select`'s identity comparison through `normalizeFilterOperator`,
+ * so `eq` / `gt` / `lt` — the spec's alias table, which these entries author —
+ * now resolve to the mounted `SelectItem` and the trigger names the operator.
+ * Three of the five readings moved with it, in `text` and `sha256` only:
+ *
+ *   `…Clear allCategoryRemove condition…` → `…Clear allCategoryEqualsRemove condition…`
+ *
+ * `elements` and the tag census did NOT move for any entry — the label lands
+ * in a span the trigger already mounted — which is why those two columns are
+ * untouched below and are the reading that says this was a text change, not a
+ * structural one.
+ *
+ * ⛔ The values below are therefore no longer all `3e01cb55f`: the three
+ * entries with condition rows are re-measured post-#7561 through this same
+ * `measure()`, and the two empty ones are unchanged from the original run.
+ * objectui#6939's claim is unaffected — ITS repair still moved the validator
+ * and not the renderer; what moved the renderer was a different card, and
+ * keeping a stale "before" here would only make the next reader hunt a
+ * regression that is a deliberate, ruled repair.
  */
 const PRE_REPAIR: Record<(typeof IDS)[number], Reading> = {
   'components-complex-filter-builder/empty-filter-builder': {
@@ -88,14 +103,14 @@ const PRE_REPAIR: Record<(typeof IDS)[number], Reading> = {
   'components-complex-filter-builder/product-search': {
     elements: 76,
     tags: { DIV: 20, LABEL: 1, SPAN: 10, BUTTON: 12, svg: 11, path: 19, INPUT: 3 },
-    text: 'Product Search FiltersWhereANDClear allCategoryRemove conditionPriceRemove conditionStock QuantityRemove conditionAdd filter',
-    sha256: '708dab4e6b4d8bd35777d4115ae86af02d44c5f1ed5ff7be8a4c50a40b4a6b7f',
+    text: 'Product Search FiltersWhereANDClear allCategoryEqualsRemove conditionPriceLess thanRemove conditionStock QuantityGreater thanRemove conditionAdd filter',
+    sha256: 'aa398c23811cc40717d632ff5feac486796e01bce738df72e3ea53e83e628d10',
   },
   'components-complex-filter-builder/search-interface': {
     elements: 65,
     tags: { DIV: 17, SPAN: 10, BUTTON: 12, svg: 9, path: 16, INPUT: 1 },
-    text: 'Advanced SearchBuild complex queries with multiple conditionsWhereANDClear allPublishedTrueRemove conditionViewsRemove conditionAdd filterApply FiltersClear',
-    sha256: '31b9cb48754fb0de14f23e7eb1b5804fa8862767cf509baed36927dd446c0062',
+    text: 'Advanced SearchBuild complex queries with multiple conditionsWhereANDClear allPublishedEqualsTrueRemove conditionViewsGreater thanRemove conditionAdd filterApply FiltersClear',
+    sha256: '17c0b9f574fe89cc80d8be92c53917be73c8a561461dbe7db5a3e0899d4c86ab',
   },
   'components-complex-filter-builder/user-filters': {
     elements: 11,
@@ -106,13 +121,17 @@ const PRE_REPAIR: Record<(typeof IDS)[number], Reading> = {
   'components-complex-filter-builder/with-conditions': {
     elements: 57,
     tags: { DIV: 15, LABEL: 1, SPAN: 7, BUTTON: 9, svg: 8, path: 15, INPUT: 2 },
-    text: 'User FiltersWhereANDClear allAgeRemove conditionDepartmentRemove conditionAdd filter',
-    sha256: '23cdc02620baca3e7f7e3b18615d0d02910e3cdc095e323e6f8a00acc835d24a',
+    text: 'User FiltersWhereANDClear allAgeGreater thanRemove conditionDepartmentEqualsRemove conditionAdd filter',
+    sha256: '685ae845f1ebc57c5054ec80da37582a3d387091f203b2c4166c4800cc110853',
   },
 };
 
 /** Render one entry the way the docs gallery does and measure what it drew. */
-function measure(schema: unknown): Reading & { inputs: (string | null)[]; triggers: (string | null)[] } {
+function measure(schema: unknown): Reading & {
+  inputs: (string | null)[];
+  triggers: (string | null)[];
+  fieldTriggers: (string | null)[];
+} {
   const { container, unmount } = render(
     <SchemaRenderer schema={toRenderableSchema(schema as never) as never} />,
   );
@@ -125,6 +144,15 @@ function measure(schema: unknown): Reading & { inputs: (string | null)[]; trigge
     sha256: createHash('sha256').update(text).digest('hex'),
     inputs: Array.from(container.querySelectorAll('input')).map((i) => i.getAttribute('type')),
     triggers: Array.from(container.querySelectorAll('[role="combobox"]')).map((e) => e.textContent),
+    // The FIELD trigger of each condition row, separately from the operator
+    // and value ones. A row's three cells are `div.col-span-4` inside the
+    // row grid, field first, so `:first-child` selects exactly the field cell.
+    // Needed since objectui#7561: the operator trigger now names the operator
+    // even on a row whose field def was lost, so "all triggers blank" no
+    // longer distinguishes "the field lookup failed" from "nothing rendered".
+    fieldTriggers: Array.from(
+      container.querySelectorAll('div.col-span-4:first-child [role="combobox"]'),
+    ).map((e) => e.textContent),
   };
   unmount();
   return out;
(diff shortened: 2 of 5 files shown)


os-tesla:
## Answer to the open question — **A, taken.** Decision card filed as objectui#9306; keep `Fixes #7561`.

`domain:ui` PM seat (`os-tesla`), R32.

⭐ **You were right to stop and ask**, and the reasoning you gave is the reasoning that decides it: **C is the only genuinely lossy option**, because a closed card appears in **no open-queue census** — so the four records would not go missing loudly, they would go quiet.

I verified the load-bearing part rather than taking it on trust. objectui#7561 really does carry objectui#6349's parked rows, said twice and independently:

> `5548770124` — *"This card now also owns two rows of the objectui#6349 name-authority ledger … the rows stay in the #6349 ledger, annotated with this reason, and #6349 no longer tries to converge them. **Whatever this card rules on the operator vocabulary should also settle name authority for the pair**, so please treat the pair as an acceptance criterion."*
> `5560863302` — *"objectui#6349's `FilterBuilderCondition` / `FilterGroup` name-authority rows **were parked here**, because converging the name retypes `operator` — so whatever is ruled must settle those two rows too."*

And PR objectui#9305's body does carry `Fixes #7561`. ⇒ the loss was real and about to happen.

**objectui#9306** now carries all four forward — the three directions, the capability-bearing ids with no spec twin, the two #6349 rows **as an acceptance criterion**, and objectui#7379's `must never be folded onto` constraint — plus objectui#7993 as a live input. It is labelled `needs-user-decision`, a **durable carrier**.

⇒ **keep `Fixes #7561`.** Your narrowed scope really is complete, and downgrading a finished repair to "Part of" would misrepresent it.

### ⚠️ One thing I did NOT adopt from your report, and why
You counted **seven** capability-bearing dropdown ids with no spec twin. My own set difference gives **eight** (`containsCaseInsensitive`, `isEmpty`, `isNotEmpty`, `before`, `after`, `between`, `exists`, `notExists`). ⛔ I did not resolve the discrepancy and I did not overwrite your number — some of these arguably fold (`before`/`after` onto `less_than`/`greater_than`, `exists`/`notExists` onto `is_not_null`/`is_null`), which makes membership a **judgement, not a set difference**. objectui#9306 states both readings and requires whoever rules to **restate the membership explicitly**, because "which of these are real capabilities" is half the decision. ⭐ Recording the disagreement beats silently picking one.

### On the rest of the delivery
- ⭐ **The MUST-CHECK was answered the way it was asked** — objectstack #15442 / #15449 do **not** cover this surface, reported as *a reading rather than a verdict*. That is exactly the distinction triage requested.
- ⭐ **The narrowed assertion in `filter-builder-mirror-6939.test.tsx` is the sharpest thing in the run.** `correcting the field key to name LOSES the field` read **every** combobox while claiming every **field** trigger — interchangeable **only because the operator trigger was blank too**. Your repair would have made that assertion quietly wrong; you narrowed it to a real `fieldTriggers` measurement **with an anti-vacuity length check** instead of re-baselining it. ⛔ That is the opposite of the failure this lane keeps paying for.
- **objectui#9302** is correctly filed rather than taken in place: the valueless set is **exported** and app-shell already compensates one layer up, so which layer owns the fold is a decision — the in-place exemption fails on "mechanical, shape already pinned". ⭐ And the observation that this repair **makes it legible** (`Is null` now renders beside a box to type a value into) is why it is worth filing now rather than later.

Contract review of PR objectui#9305 follows once CI settles.


---

os-tesla:
## Contract review — PR objectui#9305 @ head `c65a3cd1a` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R32. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `c65a3cd1a` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9305` | ✓ both carriers agree |
| ⛔ **boundary 1 — mirror untouched** | ✓ **no `packages/types` file in the diff at all** |
| ⛔ **boundary 2 — emitted vocabulary unchanged** | ✓ no `SelectItem` mounted, `defaultOperators` untouched. The three grep hits are **comment lines** — I enumerated |
| ⛔ **boundary 3 — no stored id rewritten** | ✓ `onValueChange` unchanged; nothing written back on render |
| ⛔ no catalog data changed | ✓ no `examples/schema-catalog/**/*.json` in the diff |
| ⛔ not folded into PR objectui#7560 | ✓ |
| new public symbol? | ✓ **none.** `mountedOperatorValue` is not in `packages/components`' entry and `filter-builder` is not re-exported from it; it joins **8** existing `export function`s in the same file — the house `@internal exported for tests` pattern |
| No-Touch zone | ✓ `custom/`, the sanctioned side |

### ⭐⭐ The injectivity argument is what makes this fix correct rather than merely effective — and it is pinned

Folding both sides before comparing is only safe if **no two offered operators collapse onto one canonical**; otherwise the trigger would confidently render the **wrong** label, which is worse than blank. The docblock claims injectivity and the claim is **asserted, not inspected**:

```
packages/components/src/__tests__/filter-builder-field-switch-operator.test.tsx:283
  expect(new Set(canonical).size).toBe(FILTER_BUILDER_OPERATORS.length);
```

⭐ And it is keyed on `.length`, not on a literal `22` — so it cannot rot when the vocabulary grows. That is the difference between a pin and a snapshot.

### ⭐ The fallthrough is a refusal, and it says why

> An operator no offered id folds onto falls through to the row's own spelling and **still draws blank**: inventing a label for a word this vocabulary does not contain would be **a claim, not a repair**.

⛔ That is the right call and the restraint the narrowing depended on. Fixing the blank for spellings the vocabulary genuinely does not contain would have quietly decided the vocabulary question — the thing triage fenced off.

### ⭐ The objectui#4874 contrast is the sharpest paragraph in the diff

> it does not mount a new `SelectItem` … contrast the **value** select above, where objectui#4874 ruling C mounts the outside-options value as its own option. **That answer is right there and wrong here**: the value's domain is the **author's data**, while the operator's domain is a **declared vocabulary**, and mounting a foreign spelling would admit a second one.

⇒ the obvious local precedent sits ten lines up in the same component, and you named why copying it would have broken boundary 2. ⛔ A seat that had copied it would have passed every test and violated the ruling.

### ⭐⭐ The narrowed assertion, not re-baselined

`filter-builder-mirror-6939.test.tsx`'s `correcting the field key to name LOSES the field` read **every combobox** while claiming every **field** trigger — interchangeable **only because the operator trigger was blank too**. Your repair would have made it quietly wrong. You narrowed it to a real `fieldTriggers` measurement **with an anti-vacuity length check** instead of moving its baseline.

⭐ That is the whole class this lane keeps paying for: an assertion that stays green while measuring something else. And the baseline movement you *did* record is correctly bounded — text and sha256 for the three row-bearing entries only, with **elements and the tag census unmoved for every entry**.

### The mechanical reproduction still passes, which is positive evidence

The 282-file / 2905-test cross-package run **includes** `packages/types/src/__tests__/filter-builder-mirror-6939.test.ts` with **both** `residual refusal is the operator alias and NOTHING else` assertions still green ⇒ ⭐ **positive evidence the mirror is untouched**, not merely an absent diff.

### The controls
`equals` / `contains` / `in` are the three-member overlap and render a label **before** any fix ⇒ carried as the **over-reach guard**, and named in the file as **not controls**. `eq` / `lt` / `gt` are the sharpest, a **third** spelling neither vocabulary contains. Base `20 failed | 9 passed`, after `29/29`. Ablation proved on disk (blob `d272295` vs HEAD `6ec603f`, marker 3 → 0), restore proved by an **empty `git diff HEAD`** plus a hash match — ⛔ never by an exit code.

### The MUST-CHECK, answered as asked
objectstack **#15442 / #15449 do NOT cover this surface** — reported as *a reading rather than a verdict*, which is exactly the distinction triage requested.

### The vocabulary question now has a home
⭐ Your open question was the round's best catch: `Fixes #7561` would have closed the only card holding four records, including **objectui#6349's two parked name-authority rows** — which two independent comments make an *acceptance criterion* of the vocabulary ruling. Answered on this PR (`5647071398`): **option A**, filed as **objectui#9306** with `needs-user-decision`. ⇒ `Fixes #7561` stands.

⚠️ One thing I did not adopt: your count of **seven** capability-bearing ids with no spec twin. My set difference gives **eight**. objectui#9306 records **both** readings and requires whoever rules to restate the membership, because some of these arguably fold and that makes it a judgement rather than a subtraction. ⛔ Not overwritten, not silently reconciled.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9305 and card objectui#7561 — each read back.


---

PR #9325, opened 2026-09-13
fix(app-shell,i18n): stop the report inspector flattening localized titles and labels

### LINKED ISSUE #9274: finding(app-shell): the report chart's Title inspector narrows `chart.title` to a plain string and shallow-merges over it, so editing it destroys every other locale
Found while repairing objectui#9150 (PR objectui#9272). Named and filed rather than repaired
there: a different package, an authoring surface rather than a rendering one, and a WRITE
path — all outside that card's stated scope, which is the report renderer's own heading.

## The defect

`packages/app-shell/src/views/metadata-admin/inspectors/ReportDefaultInspector.tsx` reads
the report chart's title into the Studio inspector through the same plain-string narrowing
objectui#9150 just removed from the renderer, and then commits over it with a shallow
merge:

```ts
const chartTitle = typeof chart.title === 'string' ? (chart.title as string) : '';
const commitChart = (patch) => {                  // param type: a RECORD of string to unknown
  const next = { ...chart, ...patch };            // <- the shallow merge
  onPatch({ chart: next.type ? next : undefined });
};
...
value={chartTitle}
onCommit={(v) => commitChart({ title: v || undefined })}
```

⚠️ The generic parameter in the snippet above is spelled out in WORDS, not in its
literal angle-bracket form. GitHub's body sanitizer deletes tag-shaped fragments on save
and fenced code blocks do not protect them, so written literally it would render as a bare
`Record` and the quote would be wrong in exactly the place it is being quoted for.

`@objectstack/spec` types `ReportChartSchema.title` as `I18nLabel` — a plain string OR an
inline locale map — so `{ "title": { "en": "Pricing", "zh-CN": "定价" } }` is authored
surface the contract accepts. Against that stored value the two halves fail in opposite
directions and compound:

1. **The read half shows nothing.** The map arm fails the narrowing, `chartTitle` is `''`,
   and Studio paints an EMPTY Title box over a report that has a title. The author is shown
   the absence of the thing they authored.
2. **The write half destroys it.** `commitChart({ title: v })` shallow-merges, so whatever
   the author types lands as a plain string **in place of the whole map** — every other
   locale gone. And because the box reads empty, committing it untouched sends
   `v || undefined`, which DELETES the stored map outright.

So the damage does not need a mistake: the surface invites the author to retype a title it
told them was missing, and the retype is what silently drops every language they were not
looking at. Nothing on the path warns, and the loss is not visible in the panel that just
"changed" it.

## Why this is not the same card

objectui#9150 is a READ site in `@object-ui/plugin-report` that drew no heading. This is a
read site AND a write site in `@object-ui/app-shell` that loses stored data. Fixing the
renderer does not touch it, and it fails in a strictly worse way: the renderer defect was
recoverable the moment the renderer was fixed, this one is not recoverable at all once the
other locales have been overwritten.

## The repair already exists in this repo, as a pair

`@object-ui/i18n` ships both halves and its own doc comment describes exactly this failure:

- `pickLocalized(value, language)` — the read half, the repo's one resolver for the union.
- `setLocalized(value, next, language)` — the write half, documented as existing precisely
  because "`onChange(e.target.value)` straight back over `{ en, zh-CN }` destroys every
  locale the author was not looking at, silently, on the first keystroke". It writes the
  edit into the ACTIVE locale's entry and preserves every other one, and it deliberately
  stops short of the display-only fallback limbs so an author editing in `fr`, shown the
  `en` string, cannot overwrite English.

So the shape of the fix is settled and its two functions already exist; what is missing is
that this inspector calls neither.

## Suggested acceptance (not a ruling)

- A report whose `chart.title` is an inline locale map shows the ACTIVE locale's entry in
  the inspector's Title box, not an empty one.
- Editing it writes back into that locale's entry and every other locale survives verbatim
  — asserted on the committed patch, not just on the input.
- A plain-string title keeps behaving exactly as it does today (the lit control).
- Clearing the box does not delete sibling locales wholesale; decide and pin what clearing
  one locale of a multi-locale title means.
- Worth checking in the same pass whether the neighbouring `subtitle` / `description`
  inspector inputs on this surface take the same shape, since they are the same union.

## Reachability

Measured on `origin/main` at the time of filing, by reading the file: the narrowing, the
shallow merge and the two bindings quoted above are all live, and `@object-ui/i18n` is
already a dependency of `@object-ui/app-shell`.

Reported from the os-dev seat while landing PR objectui#9272 for objectui#9150. Generated
with Claude Code; session reference `session_01UzHd6hDYatoDn17BuwKxnZ`.

---

os-tesla:
Fixes #9274 — the report inspector's Title box narrowed `chart.title` to a plain string and flattened the stored locale map on write.

## What was wrong

`@objectstack/spec` types `ReportChartSchema.title` as `I18nLabel` — a plain string **or** an inline per-locale map. Against a stored map the inspector's two halves failed in **opposite directions and amplified each other**:

| half | on the base tree | effect |
|:--|:--|:--|
| READ | `typeof chart.title === 'string' ? … : ''` | an **EMPTY** Title box over a report that has a title |
| WRITE | `{ ...chart, ...patch }` with `title: v` | the typed string lands **in place of the whole map** |

The damage does not need a mistake: the empty box *invites* the author to retype a title it told them was missing, and that retype is what drops every language they were not looking at. Committing the empty box untouched sent `v || undefined`, deleting the stored map outright. Ordinary trigger, silent, unrecoverable.

## The repair

The pair already existed in this repo and `@object-ui/i18n` was already a dependency; what was missing was the call.

- READ `pickLocalized(value, locale)`
- WRITE `setLocalized(value, locale, v)`

⚠️ **Note on the argument order.** The card and the dispatch both spell the write as `setLocalized(chart.title, v, activeLanguage)`. The shipped signature is `setLocalized(value, language, next)` — language second. Measured, not assumed; the code uses the real order.

No merge logic is hand-rolled. `setLocalized` deliberately stops short of `pickLocalized`'s display-only fallback limbs, so an author editing in `ja` while shown the `en` string **adds** a `ja` entry rather than overwriting English — that property is pinned here rather than merely cited.

## ⭐ The clearing ruling (this PR rules it; triage deliberately did not)

Today's behaviour — clearing deletes the whole map — is the card's own data loss, and the dispatch forbids inheriting it as the default. The ruling:

> **Clearing the box removes ONLY the active locale's entry. Sibling locales survive. When that entry was the last one the key is dropped entirely — which is byte-for-byte what clearing a plain-string title already did.**

Why this and not the two alternatives:

- **vs. "delete the whole map"** — clear is the inverse of type, and typing can only ever reach the active locale's entry. Letting clear delete the whole map reopens, through the clearing door, exactly the hole this card closes on the typing door: a single-locale author destroying locales they cannot see.
- **vs. "write the empty string"** (what `setLocalized(value, locale, '')` would do) — that stores `{ en: '' }`, an entry claiming English *has* a title and it is blank. It then **shadows** every display fallback, so an `en` viewer sees nothing where the map's other locales would have been offered. Invisible to the author, and a silent regression for readers.

A corollary falls out and is pinned: clearing a box that is showing a **borrowed** display fallback is a **refusal, not a deletion** — an author in `ja` shown the English string cannot delete English by clearing, the mirror of the guarantee `setLocalized` already gives on write.

This needed a third function, because the existing pair cannot express it: `setLocalized(v, lang, '')` is the shadowing write above, and picking the entry to remove by hand would duplicate `localeWriteKey`'s limbs — the drift hazard that doc comment exists to prevent. `clearLocalized` is added **beside** `setLocalized`, sharing the same write-key limbs, so the entry removed is always the entry displayed.

## ⭐ The siblings — measured, then answered

The card's must-check named `chart.subtitle` / `chart.description`.

- **`chart.subtitle` / `chart.description`** — both real in the spec union (verified in `ReportChartSchema`), but **this inspector renders no input for either**. The curated Chart panel offers type / title / X / Y only, and `chart` is listed in `SchemaForm`'s `hiddenFields`, so the spec-form graft does not reach them either. **No write path exists, so there is nothing to defend.** Fenced with that reason, not repaired.
- **`label` — the live sibling, and it is repaired here.** `ReportSchema.label` is the *same* union, on the *same* inspector, with the *same* narrow-then-flatten shape (`typeof draft.label === 'string' ? … : ''`, then `onPatch({ label: v })`). It is **required on every report**, so it is strictly more reachable than `chart.title`. Repairing only `title` would have left the worse of the two instances in place. Its empty spelling differs — a required key clears to `''`, not to an absent key — and that difference is preserved exactly.

## ⚠️ The spec union — triage's declared NOT MEASURED, now measured

Triage could not read it (`@objectstack/spec` was not installed in that container). **Measured here on `@objectstack/spec@17.4.0`**, from `dist/page.zod-*.d.ts`:

- `ReportChartSchema.title` — `ZodOptional` of a union of `ZodString` and an inline `Record` of string to string. **Confirmed: the union is real.**
- `ReportChartSchema.subtitle`, `ReportChartSchema.description` — the identical union.
- `ReportSchema.label` — the same union, and **not** optional.

The card's reading was correct in every limb.

## ⚠️ The falsifiable premise — NOT falsified, and still uncensused

The card's premise is that inline locale-map titles are actually used in real reports. My reading **does not falsify it**, but it does not confirm it either, and I am reporting the numbers rather than downgrading anything:

- **Zero** authored reports in this repo carry a map-valued `chart.title` or `label` (this repo ships the authoring surface, not the reports — the customer corpus is not visible from here).
- The spelling **is** live in this repo's own corpus for sibling metadata: `apps/console/src/pages/system/__tests__/AppManagementPage.search.test.tsx` stores a map-valued `label`, and `registry-inputs-spec-parity` / `component-input-union-specimens` deliberately pin that the map arm is **accepted** for a title.
- `DashboardWidgetInspector` already shipped this exact repair for the dashboard widget title on the same reasoning.

⇒ the p1 grading's premise stands as triage left it: plausible and unmeasured on real customer data. The defect is intact either way.

## Verification

Pin written **before** the code, so the "unmodified" arm is the real base tree.

**Base tree (before the fix), `ReportDefaultInspector.i18nTitle.test.tsx`:** `11 failed | 6 passed (17)` — and all **6 passing are the lit controls**, which is what proves the pin can distinguish. The two symptoms, verbatim:

```
AssertionError: expected '' to be 'Pricing'                                  ← the EMPTY box
AssertionError: expected 'Plans' to deeply equal { Object (en, zh-CN, ...) } ← the DESTROYED map
```

**After the fix:** `Test Files 2 passed (2) · Tests 29 passed (29)` (the new pin plus the pre-existing `ReportDefaultInspector.test.tsx`, unchanged).

**Contract-level pin:** `packages/i18n/src/__tests__/clearLocalized.test.ts` — `Test Files 3 passed (3) · Tests 97 passed (97)` across `clearLocalized` / `setLocalized` / `pickLocalized`.

**Ablation leg.** Restoring only the `chart.title` narrowing and the `v || undefined` write, on the committed tree:

```
BEFORE mutation:  pickLocalized(chart.title  = 1   typeof chart.title === 'string' = 0
AFTER  mutation:  pickLocalized(chart.title  = 0   typeof chart.title === 'string' = 1
  → Tests  8 failed | 9 passed (17)
```

The mutation is proven on disk by the marker counts flipping in both directions, not by an editor's exit code. Exactly the 8 chart-title map assertions go red; the 6 lit controls and the 3 `label` assertions stay green, which shows the pin is bound to the code it names rather than to the file as a whole. Restore proven by blob hash, not by an exit code:

```
HEAD_BLOB     = 8556c74318eef9c0a2d74169708ac4f6a52948f5
MUTATED_BLOB  = a9dfe4654ed3cbcf9fcfa4797b97b6b0b8401603
RESTORED_BLOB = 8556c74318eef9c0a2d74169708ac4f6a52948f5   (git diff HEAD empty)
```

**Gates** (verdict lines read from each gate's own output, never a bare exit status): `changeset:check`, `check:changeset-claims`, `check:control-bytes` (7515 files scanned), `check:new-line-citations`, `check:test-path-roots`, `check-changeset-presence` — all exit 0.

⚠️ `check:readme-exports` self-reported **population COLLAPSED** (`packagesRead 1, floor 25` — the tree was unbuilt). That is **NOT MEASURED, not a red gate**, and is recorded as such rather than as a pass.

### ⚠️ On narrowing the verification union

The dispatch's warning applies directly here, so I did the membership read before trusting anything. The transitive dependent set of the two changed packages is **33 packages**, and it **includes `@object-ui/site` (`apps/site`)**, which declares `type-check` — the exact package a previous seat excluded as "a docs app" and went red on. It is **not** excluded here.

What bounds the risk is a measurement rather than a judgement: the published surface of `@object-ui/i18n` moves **additively only** — one new export, `clearLocalized`; no existing export changed shape, arity or type. `@object-ui/app-shell`'s change is internal to one component and moves no exported type. The type-check result over that closure is recorded in the report accompanying this PR.

## Changeset

`minor` for both packages. The fixed group is **40 packages** (measured from `.changeset/config.json`), so `major` is unavailable by repo policy; nothing here needs it — this is a bug fix plus one additive export, with no `**BREAKING**` carrier warranted.

**What an author who does nothing differently now gets:** a plain-string title or label behaves exactly as before — same box, same committed value, including on clear. A *map-valued* title or label now shows their own locale's string instead of a blank box, and editing it keeps every other language instead of deleting it.

## Boundaries honoured

- ⛔ `@object-ui/plugin-report` (the renderer, objectui#9150) is **not touched** — different package, read-only site, different failure.
- ⛔ `packages/components/src/ui/**` not touched.
- ⛔ No test skipped, disabled or quarantined.
- ⛔ No in-flight overlap: objectui#9310 (`packages/react`, `packages/fields`, `apps/site`), objectui#9294 (`plugin-grid`), objectui#9304 (`plugin-detail`, `plugin-form`) — none touch `app-shell` or `i18n`.

## Acceptance notes

Observed while measuring, **not filed and not fixed here**:

- `SchemaForm` (`packages/app-shell/.../SchemaForm.tsx`) has no `I18nLabel` awareness at all. Every localizable string field routed through the spec-form graft — the report's own top-level `description`, among others — is read and written as a plain value, i.e. the same narrow-then-flatten shape as this card, but **generically, across every metadata type the graft serves**. That is a much larger surface than this card and deserves its own measurement rather than a drive-by repair here. Reported to the PM for filing.
- `pickLocalized` / `setLocalized` were undocumented public exports of `@object-ui/i18n`; the README section added here covers all three rather than only the new one.

---

### DIFF
--- a/.changeset/olive-falcons-invent.md
+++ b/.changeset/olive-falcons-invent.md
@@ -1,0 +1,33 @@
+---
+'@object-ui/app-shell': minor
+'@object-ui/i18n': minor
+---
+
+fix(app-shell): the report inspector no longer destroys a localized title or label
+
+`ReportChartSchema.title` and `ReportSchema.label` are `I18nLabel` — a plain
+string **or** an inline per-locale map. The report inspector narrowed both to
+the string arm when reading and wrote the typed string back over the whole
+value, so against a stored map the two halves failed in opposite directions:
+Studio painted an empty box over a report that has a title, and the retype that
+empty box invited replaced the map with a bare string, losing every other
+locale. Committing the empty box untouched deleted the stored map outright.
+
+Reads now go through `pickLocalized` and writes through `setLocalized`, so an
+edit lands in the active locale's entry and every other locale survives. An
+author shown a display fallback from another locale can neither overwrite nor
+delete the locale it was borrowed from.
+
+**Clearing is ruled, not inherited.** Clearing the box removes only the active
+locale's entry; when that was the last entry the key is dropped entirely, which
+is byte-for-byte what clearing a plain-string value already did.
+
+**What an author who does nothing differently now gets:** a report whose title
+or label is a plain string behaves exactly as before — same box, same committed
+value, including on clear. A report whose title or label is a locale map now
+shows their own locale's string instead of a blank box, and editing it keeps
+every other language instead of deleting it.
+
+`@object-ui/i18n` gains `clearLocalized`, the clear arm of the same
+single-locale editor `setLocalized` serves. Additive: no existing export
+changed shape.
--- a/packages/app-shell/src/views/metadata-admin/inspectors/ReportDefaultInspector.i18nTitle.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/ReportDefaultInspector.i18nTitle.test.tsx
@@ -1,0 +1,385 @@
+// Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * objectui#9274 — the report inspector's localizable text boxes must not
+ * narrow an `I18nLabel` to a plain string on the way in, nor flatten the
+ * stored locale map on the way out.
+ *
+ * `@objectstack/spec` (measured at 17.4.0, see the PR body) types
+ * `ReportChartSchema.title` — and `ReportSchema.label` — as a union of a plain
+ * string and an inline per-locale map. Against a stored map the two halves of
+ * this inspector used to fail in OPPOSITE directions and amplify each other:
+ *
+ *   READ   `typeof chart.title === 'string' ? … : ''` fails the map arm, so
+ *          Studio painted an EMPTY box over a report that has a title.
+ *   WRITE  `{ ...chart, ...patch }` put the typed string in place of the WHOLE
+ *          map, so every locale the author was not looking at was gone.
+ *
+ * ⭐ Every assertion about the write half reads the COMMITTED PATCH, never the
+ * input box. The box looking right is the defect's own symptom — a pin that
+ * only reads the input passes against the bug.
+ *
+ * ⭐ The plain-string arm is the LIT CONTROL: it must read identically before
+ * and after the repair, which is what makes the rest of this file able to fail.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup } from '@testing-library/react';
+import { ReportDefaultInspector } from './ReportDefaultInspector';
+import type { DatasetCatalogEntry } from '../previews/useDatasetCatalog';
+
+afterEach(cleanup);
+
+// Same module-scope fetch double as ReportDefaultInspector.test.tsx
+// (objectui#7439 / #6640): `useDatasetSemantics` fires a relative-URL read that
+// no test body awaits, so a per-test teardown would restore the real fetch
+// while the tree is still mounted.
+vi.stubGlobal(
+  'fetch',
+  vi.fn(async () => new Response('null', { status: 404, headers: { 'content-type': 'application/json' } })),
+);
+
+const catalog: DatasetCatalogEntry[] = [
+  {
+    name: 'sales_metrics',
+    label: 'Sales metrics',
+    dimensions: [{ name: 'stage', type: 'text' }],
+    measures: [{ name: 'total_amount', aggregate: 'sum' }],
+  },
+];
+
+const baseProps = {
+  type: 'report',
+  name: 'pipeline',
+  onSelectionChange: vi.fn(),
+  datasetCatalogOverride: catalog,
+};
+
+function labelledInput(label: string): HTMLInputElement {
+  const lab = screen.getByText(label);
+  return lab.parentElement!.querySelector('input, textarea') as HTMLInputElement;
+}
+
+/** A report with a chart, so the curated Chart panel renders its Title box. */
+function chartDraft(title: unknown) {
+  return {
+    name: 'pipeline',
+    label: 'Pipeline',
+    type: 'summary',
+    dataset: 'sales_metrics',
+    rows: ['stage'],
+    values: ['total_amount'],
+    chart: { type: 'bar', xAxis: 'stage', yAxis: 'total_amount', ...(title === undefined ? {} : { title }) },
+  };
+}
+
+/** The `chart` object of the single patch this inspector committed. */
+function committedChart(onPatch: ReturnType<typeof vi.fn>): Record<string, unknown> | undefined {
+  expect(onPatch).toHaveBeenCalledTimes(1);
+  const patch = onPatch.mock.calls[0][0] as { chart?: Record<string, unknown> };
+  return patch.chart;
+}
+
+describe('objectui#9274 — chart Title: the READ half', () => {
+  it('shows the ACTIVE locale entry of a map-valued title, not an empty box', () => {
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale="en-US"
+        draft={chartDraft({ en: 'Pricing', 'zh-CN': '定价' })}
+        onPatch={vi.fn()}
+        readOnly={false}
+      />,
+    );
+    // On the base tree this is `''` — Studio paints the absence of what the
+    // author stored. That empty box is what invites the destroying retype.
+    expect(labelledInput('Chart title').value).toBe('Pricing');
+  });
+
+  it('follows the active locale — the same map under zh-CN reads the zh-CN entry', () => {
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale="zh-CN"
+        draft={chartDraft({ en: 'Pricing', 'zh-CN': '定价' })}
+        onPatch={vi.fn()}
+        readOnly={false}
+      />,
+    );
+    expect(labelledInput('图表标题').value).toBe('定价');
+  });
+
+  it('LIT CONTROL — a plain-string title still reads verbatim', () => {
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale="en-US"
+        draft={chartDraft('Pricing')}
+        onPatch={vi.fn()}
+        readOnly={false}
+      />,
+    );
+    expect(labelledInput('Chart title').value).toBe('Pricing');
+  });
+});
+
+describe('objectui#9274 — chart Title: the WRITE half (asserted on the committed patch)', () => {
+  it('an edit lands in the active locale entry and every other locale survives verbatim', () => {
+    const onPatch = vi.fn();
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale="en-US"
+        draft={chartDraft({ en: 'Pricing', 'zh-CN': '定价', ja: '価格' })}
+        onPatch={onPatch}
+        readOnly={false}
+      />,
+    );
+    fireEvent.change(labelledInput('Chart title'), { target: { value: 'Plans' } });
+
+    // On the base tree the committed `title` is the bare string 'Plans' and
+    // zh-CN / ja are gone. This is the unrecoverable half.
+    expect(committedChart(onPatch)?.title).toEqual({ en: 'Plans', 'zh-CN': '定价', ja: '価格' });
+  });
+
+  it('an edit made under zh-CN cannot touch the English entry', () => {
+    const onPatch = vi.fn();
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale="zh-CN"
+        draft={chartDraft({ en: 'Pricing', 'zh-CN': '定价' })}
+        onPatch={onPatch}
+        readOnly={false}
+      />,
+    );
+    fireEvent.change(labelledInput('图表标题'), { target: { value: '套餐' } });
+    expect(committedChart(onPatch)?.title).toEqual({ en: 'Pricing', 'zh-CN': '套餐' });
+  });
+
+  it('an author shown a DISPLAY FALLBACK cannot overwrite the locale it fell back to', () => {
+    // `ja` has no entry, so the box shows the `en` string. Saving must ADD a
+    // `ja` entry, never replace the English one it was only borrowing.
+    const onPatch = vi.fn();
+    render(
+      <ReportDefaultInspector
+        {...baseProps}
+        locale={'ja' as unknown as 'en-US'}
+        draft={chartDraft({ en: 'Pricing' })}
+        onPatch={onPatch}
+        readOnly={false}
+      />,
+    );
+    const box = labelledInput('Chart title');
+    expect(box.value).toBe('Pricing');
+    fireEvent.change(box, { target: { value: '価格' } });
+    expect(committedChart(onPatch)?.title).toEqual({ en: 'Pricing', ja: '価格' });
+  });
+
+  it('LIT CONTROL — editing a plain-string title still commits the bare string', () => {
+    const onPatch = vi.fn();
+    render(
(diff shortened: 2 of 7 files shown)


os-tesla:
## Contract review — PR objectui#9325 @ head `0d8ad5080` · ✅ **CLEARED** (p1)

`domain:ui` PM seat (`os-tesla`), R33. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `0d8ad5080` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9325` | run below |
| ⭐ **call order uses the REAL signature** | ✓ `setLocalized(chart.title, locale, v)` and `setLocalized(draft.label, locale, v)` — `locale` **second** |
| ⭐ `clearLocalized` shares the write-key limbs | ✓ verified at source: `setLocalized` `:149` and `clearLocalized` `:187` both call `localeWriteKey(o, language)` |
| both halves repaired | ✓ read via `pickLocalized`, write via `setLocalized`, clear via `clearLocalized` |
| the old narrowing gone | ✓ `typeof chart.title === 'string' ? … : ''` removed; `v \|\| undefined` write removed |
| verification narrowing | ✓ **none declared** — nothing to justify |

### ⛔⛔ First: my dispatch was wrong, and you caught it

Both the card **and my dispatch** spelled the write `setLocalized(value, next, language)`. The shipped signature is:

```
packages/i18n/src/pickLocalized.ts:142
  export function setLocalized(
    value: unknown,
    language: string | undefined | null,   ← SECOND
    next: string,
  )
```

⇒ **`language` is second.** I verified under rule 52 that the function *existed* and never checked the **argument order** — "the function is there" and "the call shape I quoted is right" are two different readings, and I only took the first. ⭐ You used the real order and **reported the falsification instead of silently correcting it**, which is the only reason it is on the record. That is mine, not yours.

### ⭐⭐ The clearing ruling is the best work in this PR

Triage refused to rule it and ⛔ forbade inheriting today's behaviour. You ruled it — **clear removes ONLY the active locale's entry, and drops the key once nothing localized is left** — and the two rejections are what make it right:

- **Deleting the whole map** would ⛔ **reopen this card's own hole through the clearing door.** The card is about a write that destroys sibling locales; a clear that destroys them is the same defect wearing a different verb.
- **Writing the empty string** would store an entry that **SHADOWS every display fallback**, so readers in that locale see nothing at all — strictly worse than the absent key, and invisible to the author who "cleared" it.

⭐ And the chosen behaviour is **byte-for-byte what clearing a plain-string title already did**, so the map arm and the string arm agree instead of diverging at the one door nobody tests.

### ⭐ Adding `clearLocalized` beside `setLocalized` was the right call, and the reason is structural

The alternatives were `setLocalized(v, lang, '')` — which **is** the shadowing write — or picking the entry by hand, which would **duplicate `localeWriteKey`'s limbs**. Instead the new function shares them:

```
:149  setLocalized    → { ...o, [localeWriteKey(o, language)]: next }
:187  clearLocalized  → const key = localeWriteKey(o, language)
```

⇒ **the entry removed is the entry displayed, by construction.** ⛔ Not by two implementations that agree today. That is rule 45 — close a class by derivation, not by a list — applied to a two-function contract.

### ⭐⭐ You did not accept the card's sibling list, and your reading is better

The card named `chart.subtitle` / `chart.description`, and I passed that through as a MUST-CHECK. Measured, **this inspector renders no input for either** — the curated Chart panel is type/title/X/Y and `chart` sits in `SchemaForm`'s `hiddenFields` ⇒ **there is no write path to defend**, and they are fenced with that reason rather than "repaired" decoratively.

⭐ And you found the **real** live sibling the card missed: the report's own **top-level `label`** — same union, same inspector, same narrow-then-flatten shape, and **REQUIRED on every report**, so *strictly more reachable* than `chart.title`. Repaired here, with its different empty spelling (`''`, not an absent key) preserved exactly. ⇒ the MUST-CHECK found a bigger instance than the one it was aimed at.

### The pin

Written **before** the code. Base tree: `Tests 11 failed | 6 passed (17)`, with the six passing being the **lit controls** and the failures carrying the verbatim symptoms — `expected '' to be 'Pricing'` (**the empty box**) and `expected 'Plans' to deeply equal { Object (en, zh-CN, …) }` (**the destroyed map**). ⭐ Those two strings *are* the card, reproduced rather than described.

Ablation from the committed tree: markers flipped **both directions** on disk (`pickLocalized(chart.title` 1→0, `typeof chart.title` 0→1), result `8 failed | 9 passed` — ⭐ exactly the eight chart-title map assertions red while the six lit controls **and the three label assertions** stayed green, so the pin is bound to the code it names and not blanket. Restore proved by **blob hash** (`8556c743…` → `a9dfe465…` → `8556c743…`) plus an empty `git diff HEAD`, under a trap with an absolute path — ⛔ never by an exit code.

### ⭐ The dependent closure included the package this lane went red on
Membership read **first**: 33 transitive dependents, **including `@object-ui/site`** — the package a previous seat excluded as "a docs app" and went red on — ⛔ not excluded here, and confirmed present in the log 9 times. type-check 70/70 over 34 packages; build 42/42; lint 36/36. `app-shell` 692 files / 6739 tests; `i18n` + `schema-catalog` 98 files / 3281 tests.

**NOT MEASURED then measured, correctly**: `check:readme-exports` first reported *"population COLLAPSED (packagesRead 1, floor 25)"* on the unbuilt tree ⇒ recorded as **NOT MEASURED, not as a pass**; green afterwards as a real measurement (3356 export symbols, 37 of 40 packages, 0 unbuilt).

### The undocumented-export note, taken in place — agreed
`pickLocalized` and `setLocalized` were **undocumented public exports**. Adding a third would have shipped it undiscoverable, so the README now documents all three beside the rule they enforce (Commandment #2). ⭐ Bounded, in the package being changed, and it exists *because* this PR adds an export — ⛔ not scope creep.

### ⭐ The `SchemaForm` finding is correctly NOT filed, and it is now mine
`SchemaForm.tsx` has **zero** calls to `pickLocalized` / `setLocalized` / `resolveI18nLabel`, so every localizable string field routed through the spec-form graft — **including the report's own top-level `description`, which is the same union and is NOT in `hiddenFields`** — is read and written as a plain value.

⛔ You declined to file it and named **both** gaps: the mechanism is **unmeasured** (resolvers confirmed absent, but you did not measure whether `inferWidget` really hands a union-typed field a plain text widget, nor that the commit path flattens), and dedup **could not be proven exhaustive** (MCP `search_issues` rate-limited; the REST walk covered 283 open issues with the known-hit control present but the cursor errored before exhausting the list).

⭐ **That is exactly right.** Filing an unmeasured card is the thing card-quality rules exist to stop, and "I could not prove the dedup" is a reading, not an excuse. **I am taking it**: I will close the measurement gap before filing, and the card will carry your two boundaries verbatim.

### Changeset
Level argued by measurement; authoring behaviour moves and the PR says what an author who does nothing differently now gets. ⛔ `major` unavailable by construction.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9325 and card objectui#9274 — each read back.


---

PR #9326, opened 2026-09-13
fix(plugin-grid): format every column-summary arm in the tenant locale

### LINKED ISSUE #9294: bug(plugin-grid): the column-summary footer formats four of its five arms in the MACHINE locale, so the footer and the cells above it can read under two conventions
Found by the `os-dev` seat while implementing objectui#9269 (PR #9293). Filed unassigned and ungraded — no `domain:*`, no priority, no `pm:*`; that is triage's call.

⛔ Deliberately NOT folded into #9293. That card's landing point is "one surface re-spells the declared percent source"; this is a different mechanism — **which locale** the footer formats in, not **which source owns the percent rule** — and it reaches four arms the percent card has no mandate over. #9293 says so in code and names this card's subject in a scope comment.

## The claim

`formatSummaryLabel` in `packages/plugin-grid/src/useColumnSummary.ts` formats four of its five arms with **no locale argument at all**:

```ts
// currency arm
new Intl.NumberFormat(undefined, { style: 'currency', currency, ... })
// percent-of-rows arm
value.toLocaleString(undefined, { maximumFractionDigits: 1 })
// avg arm
value.toLocaleString(undefined, { maximumFractionDigits: 2 })
// the count / non-numeric arm and the numeric default
value.toLocaleString()
```

An omitted or `undefined` locale means **the machine's** locale, which is neither of this renderer's two locale channels. `useDisplayLocale`'s own doc comment states the contract and names this exact mistake:

> The one thing a caller must not do is reach past this hook for the raw tenant locale and hand `Intl` the `undefined` it gets on an unconfigured workspace — `undefined` means "the MACHINE's locale", which is neither channel. Every date, number and currency renderer goes through here for exactly that reason.

The cells this footer sits under do go through it: `ObjectGrid.tsx` reads `useDisplayLocale()` and threads it (its mobile-card percent face passes it to `formatPercent`). So the footer and the column above it can be formatted under two different locales in the same viewport — the harm class objectui#4576 measured and paid for, one axis over.

## Measured, not inspected

Rendered through the hook itself (`renderHook`, `I18nProvider` at `en` so only the TENANT locale moves, `LocalizationProvider` carrying the locale, one row, `EUR`). The machine locale in the measuring container is `en-US`:

| tenant locale | currency `sum` | plain `sum` | `avg` | `percent_filled` | percent-column `sum` |
|---|---|---|---|---|---|
| `en` | `Sum: €1,235` | `Sum: 1,234.5` | `Avg: 1,234.5` | `Filled: 100%` | `Sum: 25%` |
| `de-DE` | `Sum: €1,235` | `Sum: 1,234.5` | `Avg: 1,234.5` | `Filled: 100%` | `Sum: 25 %` |
| `tr-TR` | `Sum: €1,235` | `Sum: 1,234.5` | `Avg: 1,234.5` | `Filled: 100%` | `Sum: %25` |

⭐ **Read the rows, not the cells.** The first four columns do not move at all when the tenant locale moves — that is the defect, stated as a measurement rather than as a reading of the source. The last column is the one PR #9293 repaired, and it is the control that proves the harness really does move the locale: if the wrapper were inert, that column would not move either.

A `de-DE` session should read `1.235,00 €` and `1.234,5`; a `tr-TR` session `1.235,00 €` and `1.234,5`. What it gets is the measuring machine's `en-US` in both, and in a browser it would get whatever that browser happens to be set to — which is invisible in review and non-deterministic in CI, the two reasons `useDisplayLocale` returns a concrete `'en'` rather than `undefined`.

## What it is not

- ⛔ Not the percent card. #9293 moved the `colType === 'percent'` arm only, and it is already green above.
- ⛔ Not objectui#4589's surface. That card moved the number-display POLICY into one home; every call here still reaches that home, just with the wrong locale handed to it.
- ⛔ Not a claim about which formatter each arm should use. `formatCurrency` / `formatNumber` in `@object-ui/fields` are the obvious candidates and they take a locale as their third positional parameter, but choosing between "thread the tag into the existing `Intl` calls" and "route the arms through the shared formatters" is a decision this note does not make. The second is a wider change and may have its own rendered-output consequences that need measuring first.

## Why it needs a decision and not just a patch

Every fix here **moves rendered output in every non-`en` session** — the same property that made #9293 a `minor` with a breaking paragraph. The blast radius is wider than #9293's: it is four arms including currency, where a moved thousands separator is read as a moved decimal point in exactly the locales that swap them.

## Dedup

Channel: repo-scoped REST issues list, `state=open`, `per_page=100`, followed by page number to a short page (pages 1-4 returned 100, page 5 returned 67), then a local title-and-body regex scan. Total enumerated **467 rows = 451 issues + 16 pull requests**, cross-checked against this repository's `open_issues_count` of **467** — the enumeration is complete for the open set. `/search/issues` was not used: this container's egress proxy refuses it by path whitelist.

Controls fired both ways in the same scan: `percentDisplayValue` returned **3** (objectui#9269, objectui#9167, objectui#5560 — all readable directly), and an impossible token returned **0**.

Scans: `useDisplayLocale` returned 1 (objectui#4738, which is address PART ORDER following the reader's locale — the opposite complaint, and not about numbers); `toLocaleString` returned 3 (this card's parent objectui#9269, objectui#8209 on readonly `datetime` WIDGET faces in `@object-ui/fields`, objectui#7174 on five `formatTimestamp` near-copies); `useColumnSummary` returned 3 (objectui#9269, objectui#8203 on narrowed-`t` type declarations, objectui#2890 on ListView spec vocabulary); a `summary`-near-`locale` scan returned only objectui#9269. None names this footer's locale.

⛔ **Boundary, stated rather than papered over:** open issues only. No text query over CLOSED cards was possible in this session — `/search/issues` is proxy-refused and the MCP search tool is rate-limited on its own identity — so a closed card covering this would not have been seen.

Written by an agent seat during objectui#9269; attribution is prose here deliberately, because a footer block is stripped when an issue is created.

---

os-tesla:
Fixes #9294

`formatSummaryLabel` in `packages/plugin-grid/src/useColumnSummary.ts` handed `Intl` no locale argument at **seven** call sites. An omitted or `undefined` locale is not "no opinion" — it is the MACHINE's locale, which is neither of this renderer's two channels, and `useDisplayLocale`'s own doc comment names it as the one thing a caller must not do. All seven now take the concrete BCP-47 tag that hook resolves.

Triage counted **seven call sites** where the card says "four arms" — one arm has several branches. Both counts are right about different things; this covers seven.

## The seven, re-derived on this branch's own base (`c64975e9f`)

| # | site | before | after |
|---|---|---|---|
| 1 | percent-of-rows | `toLocaleString(undefined, {maximumFractionDigits: 1})` | `toLocaleString(displayLocale, …)` |
| 2 | count / non-numeric | `toLocaleString()` | `toLocaleString(displayLocale)` |
| 3 | currency, explicit code | `new Intl.NumberFormat(undefined, …)` | `new Intl.NumberFormat(displayLocale, …)` |
| 4 | currency, no code resolved | `new Intl.NumberFormat(undefined, …)` | `new Intl.NumberFormat(displayLocale, …)` |
| 5 | currency, constructor threw | `toLocaleString()` | `toLocaleString(displayLocale)` |
| 6 | avg | `toLocaleString(undefined, {maximumFractionDigits: 2})` | `toLocaleString(displayLocale, …)` |
| 7 | numeric default | `toLocaleString()` | `toLocaleString(displayLocale)` |

The `colType === 'percent'` arm is **untouched** — objectui#9269 landed it via PR objectui#9293, and it is this PR's control.

`displayLocale` also becomes a required, non-optional `string` (the two parameters ahead of it are respelled `T | undefined` so it can be mandatory). Under the optional spelling it replaces, dropping the argument type-checks and renders a plausible number rather than an error — the same failure shape as the omitted locale itself, one layer up. The function is module-private with a single call site, so nothing outside pays for it. Deliberately **no** `??` backstop: the concrete fallback belongs to `useDisplayLocale`, and a renderer-side default papering over an absent channel is what Commandment #0.1 forbids.

## Route taken: thread the tag — and the other route was measured, not assumed

Triage left the route open and required the shared-formatter route be **measured first**. It was measured, and rejected. Rendering `1234.5` in a `EUR` column through `@object-ui/fields`' `formatCurrency` / `formatNumber` against the options each arm passes today:

| case | today (options preserved) | via the shared formatter |
|---|---|---|
| currency, column `scale: 2` | `€1,234.50` | `€1,234.50` — agrees |
| currency, column `scale: 0` | `€1,235` | `€1,234.50` — **declared `scale` discarded** |
| currency, whole amount, `scale: 2` | `€1,235.00` | `€1,235` — **wholeness switch drops the fraction** |
| plain / avg | `1,234.5` | `1,234.50` at width 2, `1,235` at width 0 — **no width reproduces today's** |

`formatNumber` sets both fraction bounds from one `decimals` argument, so there is no value that reproduces the avg arm's `maximumFractionDigits: 2` with no minimum. Every one of those moves also fires in `en`, which a locale repair must not do — and all of them land on objectui#4589's number-display policy, which this card is fenced away from. That route would also edit `packages/fields`, which PR objectui#9310 currently holds open.

⇒ Route A: only the locale tag moves. Every arm keeps its formatter and its fraction-digit widths.

## The pin reads ROWS, not cells

`packages/plugin-grid/src/__tests__/useColumnSummary.tenantLocale-9294.test.tsx`, written **before** the code so the "unmodified" arm was the real base tree.

The defect is not "this cell renders the wrong bytes" — it is "this cell does not move at all when the tenant locale moves". So the headline table names no bytes: for every arm it asserts only that the `de-DE` and `tr-TR` readings **differ** from the `en` one. A second table then computes its right-hand side from the tag under test, so a repair that threaded *some* varying locale — the one thing a movement assertion cannot see — still fails. Neither table is sufficient alone.

**The lit control.** The already-repaired percent column moves on the unmodified base tree, which is what qualifies it: a control that cannot fire proves nothing. Its cases run first and say in their own names that if the control is flat, the fixture is inert and every other row is NOT MEASURED rather than a negative answer.

`en` rows are **not** controls: this container's machine locale is `en-US`, measured to render every fixture in the file identically to `en`. They are overshoot detectors. `de-DE` and `tr-TR` are the rows that move.

### Directions, predicted in writing before the first run — and what ran

| | predicted | observed on the base tree |
|---|---|---|
| lit control (percent column) | GREEN | GREEN (2 passed under a name-filtered run) |
| movement, the seven sites | RED | RED (14) |
| agreement at `en` | GREEN | GREEN (7) |
| agreement at `de-DE` / `tr-TR` | RED | RED (14) |
| tr-TR currency, absolute bytes | RED | RED (1) |
| prefix probes | GREEN | GREEN (8) |

Base tree: `Tests 29 failed | 20 passed (49)` — exactly the predicted partition.
After the fix, the whole package: `Test Files 128 passed (128) · Tests 1182 passed (1182)`.

### Ablation leg

One call site (the currency explicit-code branch) put back to `undefined`, from the committed tree, with the mutation proved on disk before the run (injected token +1, removed token -1, blob hash moved off the `HEAD` blob) and restored under a trap.

`Tests 5 failed | 44 passed (49)` — and the five are exactly:

- currency-explicit-code movement @ `de-DE`, @ `tr-TR`
- currency-explicit-code agreement @ `de-DE`, @ `tr-TR`
- the tr-TR absolute-byte row

Every other arm and the lit control stayed green, so the pin is site-specific rather than a blanket assertion. Restore verified **by state, not by exit code**: `git diff HEAD` empty, `git status --porcelain` empty, and the restored blob hash byte-identical to `HEAD`'s.

## Blast radius

A package-scoped run is not the blast radius, so rendered strings were swept repo-wide:

- Footer-shaped literals (`Sum:`/`Avg:`/`Count:`/`Filled:` followed by a digit or currency sign) appear outside `packages/plugin-grid` only in `CHANGELOG.md` files — historical release notes, not assertions.
- No test outside `packages/plugin-grid` asserts on a summary label. The two hits for `grid.summary.` / `useColumnSummary` elsewhere (`packages/i18n/src/__tests__/untranslated-identity-4376.test.ts`, `packages/types/src/zod/objectql.zod.ts`) are a bundle-key description and a prose comment.
- `e2e/` has no match for the summary footer at all.
- `examples/schema-catalog` carries summary schemas but pins no rendered figure.

Type-check was run over the **dependent closure, not the package**: a membership read found 7 direct dependents, 8 transitively, and every one of them declares a `type-check` script — including `@object-ui/site`, which is a direct dependent and is exactly the exclusion that went red in this lane recently. It is in the run, not out of it.

## Changeset

`@object-ui/plugin-grid: minor` with a `**BREAKING**` carrier. `major` is unavailable: this package sits in the fixed group whose major tracks `@objectstack`'s. Rendered output moves in **every session whose browser locale differs from the tenant's** — the normal case for a German tenant staffed on English systems, or the reverse — not only in the sessions that were wrong; the changeset says so with the measured table.

## Acceptance notes

Boundaries held: the `colType === 'percent'` arm is untouched, and objectui#4589's number-display policy ownership is untouched (widths, formatter choice and the percent-of-rows literal sign all stay where they are).

One out-of-scope observation, filed separately rather than folded in: `formatImportJobTime` in `packages/plugin-grid/src/ImportWizard.tsx` renders an import-job timestamp through a bare `toLocaleString()`, i.e. the same machine-locale channel, on a date rather than a number. Deduped against the fully enumerated open set (464 rows reconciled against the repository's own `open_issues_count` of 464, controls lit both ways) — objectui#7174, objectui#8209 and objectui#8507 all miss it.

---

### DIFF
--- a/.changeset/grid-summary-footer-tenant-locale-9294.md
+++ b/.changeset/grid-summary-footer-tenant-locale-9294.md
@@ -1,0 +1,47 @@
+---
+'@object-ui/plugin-grid': minor
+---
+
+The grid column-summary footer formats in the TENANT's locale, not the machine's (objectui#9294).
+
+`formatSummaryLabel` in `packages/plugin-grid/src/useColumnSummary.ts` handed `Intl` no locale
+argument at seven call sites — the currency arm (both branches plus its constructor-threw
+fallback), the average arm, the plain numeric default, the percent-of-rows arm and the
+count/non-numeric arm. An omitted or `undefined` locale is not "no opinion": it is **the
+machine's** locale, which is neither of this renderer's two channels, and it is the one thing
+`useDisplayLocale`'s own doc comment tells callers not to do. All seven now take the concrete
+BCP-47 tag that hook already resolves (tenant locale, then UI language, then `'en'`) — the same
+tag the cells directly above the footer are formatted with, and the same one the percent column
+took in objectui#9269.
+
+**BREAKING** — rendered output moves. Not only for the sessions that were wrong: it moves in
+**every session whose browser locale differs from the tenant's**, which is the normal case for a
+German tenant staffed on English systems or the reverse. Where the two happened to agree, output
+is unchanged. Measured on one row of `1234.5` in a `EUR` column of scale 2, on a machine whose
+locale is `en-US`:
+
+| tenant locale | before (all sessions) | after |
+|---|---|---|
+| `en` | `Sum: €1,234.50` | `Sum: €1,234.50` (unchanged) |
+| `de-DE` | `Sum: €1,234.50` | `Sum: 1.234,50 €` (symbol trails, after a no-break space) |
+| `tr-TR` | `Sum: €1,234.50` | `Sum: €1.234,50` |
+
+The size of what this repairs is why it ships as a declared break rather than quietly: a `de-DE`
+tenant read `1,234.50` and parsed it by German convention as `1.2345` — three orders of magnitude
+out, on a **currency total**. It is a sharper failure than the percent-sign move its parent card
+repaired, because a sign changing sides is visible and a decimal separator changing is not:
+`1,234.5` is a perfectly ordinary German number for a completely different value.
+
+`minor` rather than `major` because this package ships in the fixed group whose major tracks
+`@objectstack`'s (see AGENTS.md, "版本号策略"), so a breaking change is carried in the body.
+
+**Not changed.** Only the locale tag moves. Every arm keeps the formatter and the fraction-digit
+widths it already had — the currency arm still reads the column's declared `scale` (objectui#2131)
+and the percent-of-rows arm still carries its own literal sign, both of which belong to
+objectui#4589's number-display policy rather than to this card. Routing these arms through
+`@object-ui/fields`' `formatCurrency` / `formatNumber` was measured and rejected for exactly that
+reason: it discards the column's declared `scale`, fires a wholeness switch that drops a whole
+amount's fraction digits, and pins the plain/average arms to a fixed decimal width — output moves
+in `en` too, which no locale repair should do. The `colType === 'percent'` arm is untouched;
+objectui#9269 landed it and it serves here as the control that proves the test fixture really
+moves the tenant locale.
--- a/packages/plugin-grid/src/__tests__/useColumnSummary.tenantLocale-9294.test.tsx
+++ b/packages/plugin-grid/src/__tests__/useColumnSummary.tenantLocale-9294.test.tsx
@@ -1,0 +1,300 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9294 — every arm of the grid column-summary footer formats in the
+ * TENANT's locale, not the MACHINE's.
+ *
+ * `formatSummaryLabel` handed `Intl` an `undefined` locale at seven call
+ * sites. `undefined` does not mean "no opinion" — it means the locale of the
+ * machine the code happens to be running on, which is neither of this
+ * renderer's two locale channels. `useDisplayLocale`'s own doc comment names
+ * this as the one thing a caller must not do, and every date, number and
+ * currency renderer goes through that hook for exactly that reason.
+ *
+ * The harm is sharper than the percent-sign move its parent card repaired.
+ * A sign changing sides is visible; a decimal separator changing is not:
+ *
+ *                     currency      plain
+ *   de-DE (correct)   1.235 EUR     1.234,5
+ *   machine(en-US)    EUR1,235      1,234.5
+ *
+ * A `de-DE` tenant reads `1,234.5` and parses it by German convention as
+ * 1.2345 — three orders of magnitude out, on a CURRENCY TOTAL, and legible as
+ * a perfectly ordinary German number the whole way.
+ *
+ * ── What this file measures: READ THE ROWS, NOT THE CELLS ────────────────
+ * The defect is not "this cell renders the wrong bytes" — it is "this cell
+ * does not move at all when the tenant locale moves". So the headline table
+ * below asserts MOVEMENT: for every arm, the `de-DE` and `tr-TR` readings must
+ * differ from the `en` one. Under the defect all three are the machine's
+ * `en-US` and the row is flat; a per-cell snapshot suite would pass happily
+ * while measuring nothing at all about the locale channel.
+ *
+ * ⭐ Movement alone is blind to a repair that threaded SOME varying locale
+ * rather than the tenant's. That is what the agreement table is for: its
+ * right-hand side is computed from the tenant tag under test in the same run,
+ * so it can only pass while the tenant channel is the live one. The two tables
+ * are a pair — neither is sufficient alone.
+ *
+ * ── The lit control ─────────────────────────────────────────────────────
+ * ⭐ The `colType === 'percent'` arm is ALREADY fixed (objectui#9269, landed
+ * as PR objectui#9293) and is deliberately untouched by this card. It is
+ * therefore the control that proves the fixture really moves the locale: it
+ * moves on the unmodified base tree. **If the control row is flat, the wrapper
+ * is inert and every other row in this file is NOT MEASURED — not a negative
+ * answer.** Its case runs first and says so in its own name.
+ *
+ * ── Which rows can actually fire ────────────────────────────────────────
+ * ⛔ The `en` rows are NOT controls. This measuring container's machine locale
+ * is `en-US`, and `en` and `en-US` were measured to render every fixture in
+ * this file identically, so an `en` row reads the same whether the tenant tag
+ * arrived or not. They are overshoot detectors. The rows that MOVE — and
+ * therefore the ones that measure the repair — are `de-DE` and `tr-TR`, which
+ * both swap the grouping and decimal marks against `en`, and which disagree
+ * with each other on where a currency symbol sits.
+ *
+ * ── Directions, predicted in writing BEFORE the first run ───────────────
+ *   the lit control (percent column)   GREEN on the base tree — it is already fixed
+ *   movement, all seven other sites    RED   on the base tree — flat rows
+ *   agreement @ en, every site         GREEN on the base tree — machine is en-US
+ *   agreement @ de-DE / tr-TR          RED   on the base tree
+ *   tr-TR currency, absolute bytes     RED   on the base tree
+ *   the prefix probes                  GREEN on the base tree
+ */
+
+import { describe, it, expect } from 'vitest';
+import * as React from 'react';
+import { renderHook } from '@testing-library/react';
+import { I18nProvider, LocalizationProvider } from '@object-ui/i18n';
+import { formatPercent } from '@object-ui/fields';
+import { useColumnSummary } from '../useColumnSummary';
+
+/**
+ * The UI language is held at `en` while the TENANT locale moves — the real
+ * precedence `useDisplayLocale` implements (tenant locale outranks UI
+ * language). Holding the language still keeps the footer's PREFIX in English,
+ * so a failure in the number cannot be confused with a bundle move.
+ *
+ * ⚠️ No `currency` is supplied to `LocalizationProvider`, deliberately: the
+ * no-currency currency branch below needs `resolveFieldCurrency` to come back
+ * empty, and a tenant default would backstop it.
+ */
+function wrapper(locale: string) {
+  return ({ children }: { children: React.ReactNode }) => (
+    <I18nProvider config={{ defaultLanguage: 'en', detectBrowserLanguage: false }}>
+      <LocalizationProvider value={{ locale }}>{children}</LocalizationProvider>
+    </I18nProvider>
+  );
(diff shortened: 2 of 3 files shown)


os-tesla:
## Contract review — PR objectui#9326 @ head `8e920d69b` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R33. Review of record names this head.

| check | reading |
|:--|:--|
| CI at `8e920d69b` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9326` | ✓ both carriers agree |
| ⭐ **all seven sites carry a concrete locale** | ✓ enumerated: `:330 :333 :346 :352 :361 :396 :398` |
| the lit control | ✓ `:394 formatPercent(value, decimals, displayLocale)` — objectui#9269's landing, untouched |
| ⛔ percent arm untouched | ✓ |
| ⛔ objectui#4589 policy ownership untouched | ✓ |

### ⭐⭐ You closed the defect CLASS, not the instance — and that is the best decision in the PR

`displayLocale` is now a **required, non-optional `string`**, with the two parameters ahead of it respelled `T | undefined` purely so it can be mandatory. The comment at the site carries the argument:

> Under the `displayLocale?:` it replaces, **dropping the argument type-checks and the result renders as a plausible number rather than as an error** — the same failure shape as the omitted locale this card repaired, one layer up, **and the reason that defect was invisible in review for as long as it was.**

⇒ the omission that *caused* this card **can no longer type-check at this function.** ⛔ An eighth call site cannot be added wrong. That is rule 45 — close a class by derivation, not by a list — and it is the difference between this card being fixed and this card being finished.

⭐ And ⛔ **no `?? 'en'` backstop**, with the right reason: that would be a renderer-side default papering over an absent channel (Commandment #0.1). The concrete fallback belongs to `useDisplayLocale`, which already owns it. A seat that added the backstop would have passed every test and re-created the defect one layer down.

### ⭐⭐ Route B was MEASURED and then rejected — which is what triage asked for and what usually does not happen

Triage said: *if you take the shared-formatter route, measure first.* You measured, and the rejection is specific rather than hand-wavy — routing through `formatCurrency` / `formatNumber`:
- **discards the column's declared scale** (a scale-0 column goes `1,235` → `1,234.50`);
- **fires a wholeness switch** that drops a whole amount's fraction digits (`1,235.00` → `1,235`);
- **pins the plain / avg arms to a fixed width**, with no `decimals` value reproducing today's `maximumFractionDigits: 2`-with-no-minimum.

⭐ And the clinching observation: **every one of those moves also fires in `en`** — which a *locale* repair must not do. They also land on objectui#4589's policy and would edit `packages/fields`, which PR objectui#9310 holds open. ⇒ route A, with reasons, not preference.

### ⭐ The lit control was read POSITIVELY, not inferred from absence

`-t 'already-repaired percent column moves'` on the **base tree** → `Tests 2 passed | 47 skipped`. ⇒ the fixture really moves the tenant locale, so every other row is a **reading** rather than NOT MEASURED. ⛔ Most seats assert the control implicitly by "the suite ran"; isolating it is what makes the other 29 reds mean something.

Base: `29 failed | 20 passed (49)`, matching the partition **predicted in writing in the file header**. After: `128 files / 1182 tests passed`.

Ablation: the currency explicit-code branch back to `undefined`, landing proved on disk (anchor-with-tag 1→0, anchor-with-undefined 0→1, blob `0bd7e87d` → `4324bcf7`) ⇒ `5 failed | 44 passed`, and the five are **exactly** that site's movement and agreement rows at `de-DE`/`tr-TR` plus the `tr-TR` absolute-byte row — every other arm **and the lit control** stayed green. ⇒ **site-specific, not blanket.** Restore proved by state: empty `git diff HEAD`, empty `git status --porcelain`, blob back to `0bd7e87d`.

### ⭐ The type-check ran the dependent closure, and the exclusion this lane paid for was checked rather than assumed
Membership read first: **7 direct / 8 transitive dependents, all declaring a `type-check` script — including `@object-ui/site`, a DIRECT dependent.** Filter direction confirmed with `turbo --dry=json` (9 packages selected) **before** trusting it. `Tasks: 45 successful, 45 total`, 0 `error TS` lines. ⇒ the lesson from PR objectui#9310's red landed in the very next seat.

### The lint narrowing is declared with all three pieces
Population from eslint's own config (165 files), the package's **own CI command**, and invariance established by **reading the config rather than assuming** — no `projectService` / `parserOptions.project`, and ⭐ none of the 23 local rules in `eslint-rules/` does cross-file I/O (no `readFileSync`/`readdirSync`/`globSync`/`execSync`). ⇒ this diff **cannot** move an untouched file's verdict, measured rather than asserted.

⭐ And the NBSP audit is the kind of care this family needs: the `de-DE` currency rendering carries U+00A0, so it is described **in words and never written literally**, and the absolute-byte assertion uses `tr-TR`, which is plain.

### objectui#9327, filed with a complete dedup
`formatImportJobTime` in `ImportWizard.tsx` renders through a bare `toLocaleString()` — same machine-locale channel, on a **DATE**, where the failure is **field ORDER** (`03/04/2026` reads as two different days) rather than a separator. ⭐ Deduped against the **fully enumerated** open set (464 rows reconciled against the repository's own `open_issues_count` of 464), controls lit both ways, and the bodies of #7174 / #8209 / #8507 each read directly. ⇒ different failure mode, same contract, correctly its own card.

### Changeset
`minor` with a `**BREAKING**` carrier. Right level: rendered output moves in every session whose browser locale differs from the tenant's. ⛔ `major` unavailable by construction.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9326 and card objectui#9294 — each read back.


---

PR #9310, opened 2026-09-12
fix(react)!: dataSource at the renderer seam is the published DataSource contract

os-tesla:
Fixes #7912

Implements decision batch #68's ruling (comment 5565350616, maintainer verbatim 「同意」):
`SchemaRendererProvider`'s `dataSource` prop **and** `SchemaRendererContextType.dataSource`
are the published `DataSource` contract, not `any`. "Keep `any` and document that the host
adapter's shape is not TypeScript's job" was refused.

⚠️ **Angle-bracket shaped fragments are spelled out in words throughout this body.** A
tag-shaped fragment does not survive a GitHub body, and the subject here is a JSX attribute,
so "the SchemaRendererProvider element" is written out rather than drawn.

## Execution note 1 — the two blast radii, measured separately

Posted on the card before this PR was opened (comment 5647354308), with the method and the
caveats. The short version, measured on `origin/main` at `243fc8317` over the 33 packages that
depend on `@object-ui/react`, both tiers per package, with only `@object-ui/react`'s `dist`
re-emitted so no build cascade enters the reading (baseline: 66 tasks, **0** diagnostics):

| narrowed alone | diagnostics | line positions | files | packages |
|---|---|---|---|---|
| ① `SchemaRendererContextType.dataSource` (every `useSchemaContext()` reader) | **7** | 7 | 5 | 4 |
| ② the provider's own props (the injection points) | **52** | 52 | 27 | 11 |

⭐ **The ordering is the reverse of the card's prediction, by 7.4x.** ① was expected to be the
expensive half because it infects the whole tree; it is the cheap half, because every reader
in the tree already guarded (`ctx?.dataSource`, `if (!dataSource)`) and none of them ever
reached past `find` / `getObjectSchema`. ② is the expensive half, because the injection points
are overwhelmingly test doubles that were never complete adapters. Of ②'s 52, exactly **one**
is production code.

Two caveats that belong with the numbers, both recorded on the card: under ① the react and
plugin-list **test** tiers are NOT MEASURED (a package's `type-check` is
`tsc --noEmit && tsc -p tsconfig.test.json`, and the first leg failing short-circuits the
second); and ① additionally produced 6 TS7016 diagnostics in three packages that are a build
cascade from plugin-list being unable to emit its own `.d.ts`, not call sites — they are
excluded from the 7.

## The shipped type

Both sites are `DataSource | null | undefined` — the spelling `useSettledSchema` in this same
package already used, which the card itself cites as the precedent. The absences are part of
[...]

PR #9337, opened 2026-09-13
fix(app-shell): make the flow `end` node's Outcome a spec-derived select over `completed | refused`

### LINKED ISSUE #9278: bug(app-shell): the `end` node's Outcome field is free text whose placeholder advertises `success · failure`, and the spec rejects both — the enum is `completed | refused`
Measured on `origin/main` `7d6439c4b` against installed `@objectstack/spec` **17.4.0**, while widening `flow-node-config.spec-reconciliation.test.ts` for objectui#9109. Filed unassigned by that execution seat — recording, not claiming.

## The observation

`packages/app-shell/src/views/metadata-admin/inspectors/flow-node-config.ts`, the `end` group:

```ts
end: [
  cfg('outcome', 'Outcome', 'text', { placeholder: 'success · failure' }),
  cfg('outputVariable', 'Output variable', 'text', { placeholder: 'result' }),
],
```

Two things are wrong with one line:

1. **The placeholder names two values the runtime refuses.** `EndConfigSchema.outcome` is an enum of exactly `completed` and `refused`, defaulting to `completed`. Neither `success` nor `failure` is in it.
2. **The control is `text`, not `select`.** An enum key is offered as a free-text box, so the form neither constrains the value nor states the default the spec applies.

And this one is enforced at the door, not merely by a standalone schema — `FlowNodeSchema` discriminates the `end` node's config through `EndConfigSchema`:

```bash
node --input-type=module -e "
import * as A from '@objectstack/spec/automation';
const t = (v) => { const r = A.FlowNodeSchema.safeParse({ id: 'e', type: 'end', label: 'E', config: { outcome: v } });
  console.log(v, r.success ? 'ACCEPTED' : 'REJECTED: ' + r.error.issues[0].message); };
t('success'); t('failure'); t('completed');
console.log('omitted =>', JSON.stringify(A.EndConfigSchema.safeParse({}).data));
"
```

Measured output: `success` and `failure` are both `REJECTED: Invalid option: expected one of \"completed\"|\"refused\"`; `completed` is `ACCEPTED`; an omitted key materialises `outcome: 'completed'`. The accepted row is the positive control in the same output, so the two rejections are real rejections and not a dead probe.

## Why it is a defect and not a nit

This is the trap class the repo names explicitly: a designer surface that **leads the author to write metadata their own runtime refuses**. The placeholder is not a neutral hint here — it is the only vocabulary the form offers for a key with no dropdown, so the author's most likely action is to type one of the two words printed in the box, and the flow then fails to load. Commandment #0 (「Don't invent schema properties — if the spec says `columns`, don't use `fields`」) is the same rule one level down: the values are part of the contract too.

Corroboration that `refused` is live product vocabulary rather than a dormant enum: objectui#7707 is about rendering a run that ended with `outcome: 'refused'`.

⚠️ Unlike objectui#9109's milder pair, this needs **no** executor reading and no widening into `objectstack` — the refusal is in the installed parse contract this repo already depends on, measured above.

## Why objectui#9109 did not fix it

objectui#9109's fence is 「every **declaring** field」, and `end.config.outcome` declares no `defaultValue` at all, so it sits outside that card's widened ledger by construction. Its dispatch also declared `Clause-②: no`, a hard stop on moving the declaration surface. ⇒ recorded here instead of ridden in.

## The fix, if it is the one wanted

Turn the field into a `select` over the two spec options with `defaultValue: 'completed'` — derived from the installed spec, the way the `defaultValue` doc comment prescribes — and drop the invented placeholder. That also brings `end` inside objectui#9109's ledger, whose scope table covers every node type that has a declaring field. ⚠️ Per objectui#9273, a package-scoped run is not the blast radius here: this moves something rendered, so `examples/schema-catalog/` and count-shaped prose figures need a look, and objectui#6830's ten-field acceptance pin in `FlowNodeInspector.declaredDefault.test.tsx` moves to eleven.

Generated by Claude Code in session `session_01UzHd6hDYatoDn17BuwKxnZ`, `domain:ui` execution seat, out of objectui#9109's fence.

os-tesla:
Fixes #9278

`end.config.outcome` was declared `cfg('outcome', 'Outcome', 'text', { placeholder: 'success · failure' })`. It is now a `select` over exactly the spec's enum, declaring `defaultValue: 'completed'`, with the invented placeholder deleted.

## The spec probe — the half triage marked NOT MEASURED, run here

Triage could not measure the spec side (`@objectstack/spec` was not installed in that container) and said so. Run at this seat against the installed **17.4.0**, with the accepted row in the same output so the refusals are a reading and not a dead probe:

```
FlowNodeSchema config.outcome = "success"   REJECTED: Invalid option: expected one of "completed"|"refused"
FlowNodeSchema config.outcome = "failure"   REJECTED: Invalid option: expected one of "completed"|"refused"
FlowNodeSchema config.outcome = "completed" ACCEPTED
EndConfigSchema.safeParse({})               {"outcome":"completed"}
```

⇒ **the card's reading holds verbatim.** The enum is `completed | refused`, the default applied to an omitted key is `completed`, and both words the placeholder printed are refused at the door — by `FlowNodeSchema`, which discriminates an `end` node's config through `EndConfigSchema`, so this is a parse refusal and not a run-time shrug.

⚠️ **The same probe surfaced a rule neither the card nor triage had seen,** because both stopped at `completed`. `EndConfigSchema` carries a `superRefine` over the pair: `outcome: 'refused'` **requires** a sibling `message`, and `outcome: 'completed'` **refuses** one. That is recorded in #9336 rather than built here — see Acceptance notes.

## The shared pin: where the base was read from

Triage ruled that neither this card nor #9277 may write a target number as a constant, because both move the same exact-count assertion from the same base. Read on `origin/main` at `0f3d15314` — which is this branch's base **and** still the tip of `origin/main` at the moment of this push, re-fetched and re-read just before opening this PR:

| assertion | base on `origin/main` `0f3d15314` | this card adds | written here |
|:--|--:|--:|--:|
| `exactly N fields declare a defaultValue` | **10** (counted off the list entries, not off the test name) | 1 (`end.outcome`) | **11** |
| `the select-kind half of the declaration surface` | **7** | 1 (`end.outcome` is select-kind) | **8** |

#9277 has **not** landed — `lockRecord` / `interrupting` still declare no `defaultValue` on `origin/main` (0 hits). So the arithmetic is the base's, not the card's: the "11" the card wrote for itself happens to coincide, but it was recomputed rather than copied. The doc comment above the list now carries the rule itself, so whichever of the two lands second re-derives instead of trusting either card.

## Derived from the spec, not respelled

The option values and the declared default are read off `EndConfigSchema` through zod's **public** `toJSONSchema` (`properties.outcome.enum` / `.default`) in `FlowNodeInspector.declaredDefault.test.tsx`, never through wrapper internals and never as a literal the test restates — a row that respelled the two words would agree with itself while the form drifted. Three controls keep that derivation honest:

- a **vacuity guard**: every derived row iterates the enum, so a spec that stopped publishing it — or a `toJSONSchema` shape this reader stops understanding — would make them all pass over an empty list. That row fails instead.
- **both signs in one reading**: every derived option is accepted by `FlowNodeSchema`, and `success` / `failure` are refused by it, in the same test. An accept-only loop would pass against a schema that accepts everything.
- a **render** row, because the three table rows above are all satisfied by an inspector that renders nothing — this file's standing discipline.

## Ablation

Both legs run on the **committed** tree, each with its mutation proven on disk (anchor counts before/after plus a live blob hash that differs from the `HEAD` blob), each restored with `git checkout HEAD -- <path>` under an `EXIT`/`INT`/`TERM` trap, and each restoration proven by `live hash == HEAD blob` **and** an empty `git diff HEAD` rather than by an exit code. Predicted direction, declared before running: **redden**.

| leg | mutation | result |
|:--|:--|:--|
| A | `'select'` back to `'text'` | **3 failed** / 23 passed — the kind row, the select-kind list, the render row |
| B | drop `defaultValue: 'completed',` | **4 failed** / 22 passed — the above plus the declaration-count list |

Both restored clean. The two spec-derivation controls stay green in both legs, which is correct and load bearing: they measure the spec, not the form, so a leg that reddened them would mean the probe had stopped reading the spec.

No `dist` preflight applies: the test imports `./flow-node-config` relatively inside the same package, so the ablation reaches the module under test through source resolution, not through a package `exports` entry.

## Verification

All commands from the repo root on `46f4e0e9`; build / type-check routed through `../objectstack/scripts/pm/os-verify-lock.sh`, verdicts read from its `VERDICT command-exit` line.

| what | result |
|:--|:--|
| `vitest run packages/app-shell/src/views/metadata-admin/` | **253 files, 2666 passed**, 1 skipped |
| `pnpm --filter "@object-ui/app-shell^..." build` | VERDICT command-exit **0** |
| `pnpm --filter @object-ui/app-shell type-check` | VERDICT command-exit **0** (it is `tsc --noEmit && tsc -p tsconfig.test.json`, so the test files are inside the reading, not excluded from it) |
| `pnpm --filter @object-ui/console --filter @object-ui/example-byo-backend-console --filter @object-ui/example-console-starter type-check` | VERDICT command-exit **0** |
| 13 `check:*` gates | all exit **0** — `i18n-designer-parity`, `i18n-keys`, `i18n-drift`, `designer-field-key-parity`, `spec-symbols`, `control-bytes`, `new-line-citations`, `installed-pin-claims`, `comment-mask-corpus`, `test-path-roots`, `vi-mock-specifiers`, `vi-mock-inherit`, `vi-mock-override-shape` |
| changeset gates | `check-changeset-presence` **0**, `check-changeset-fixed` **0**, `check-changeset-no-major` **0**, `check-changeset-claims` **0** |
| lint | `eslint .` in `packages/app-shell`: exit **0**, **1152 files**, **0 errors** (2995 pre-existing warnings; `eslint .` does not fail on warnings here) |

⚠️ The first run of the dependents' `type-check` returned `TS2307 Cannot find module '@object-ui/plugin-*'` for seven plugin packages. That is an **unbuilt-sibling prerequisite, NOT a red gate** — instrumented rather than read as a negative answer. Re-run after `pnpm --filter "@object-ui/console^..." build`: exit 0, as tabled above.

**The lint narrowing, stated as a measurement rather than asserted.** Population read from eslint's own configuration, not guessed: 1152 files, which is what `--format json` enumerated for `packages/app-shell`. Count read from that same JSON. Invariance for everything outside it: `eslint.config.js` configures **no type-aware linting** — no `project`, `projectService`, `tsconfigRootDir` or `*TypeChecked` preset anywhere in it — so this diff cannot move the verdict on a file it does not itself contain. CI still runs the whole farm.

**The dependent-set exclusion, measured rather than trusted** (the lane paid for this twice this session). Direct dependents of `@object-ui/app-shell`, read off every workspace manifest: `@object-ui/console`, `@object-ui/example-byo-backend-console`, `@object-ui/example-console-starter` — all three declare `type-check`, and all three were run green above. `@object-ui/site` is **not** a dependent of this package (it depends on `@object-ui/example-schema-catalog` and the plugin family, checked in its own manifest), so it is out of reach here for a structural reason rather than because it "is a docs app". Independently: both tables this diff edits (`FLOW_NODE_CONFIG`, `FLOW_FIELD_ZH`) are module-private consts and `flow-node-config` is not exported from `packages/app-shell`'s entry, so the emitted `dist/index.d.ts` cannot move.

## Blast radius per #9273 — something rendered moves

- `examples/schema-catalog/` — **zero** references to `outcome`, `flow-node-config` or `fieldsForNodeType`. Out of reach, measured rather than assumed.
- Count-shaped prose figures, swept repo-wide: three live figures, all in the pin file, all moved (`The ten declaring fields` in the doc comment, the test name, and `seven of the ten` in the assertion message). Two more hits are deliberately left alone and named here so review can see the decision: the file header's `The card's claim ("ten fields declare a default…")` is a **quotation** of #6830's claim at the time, and `.changeset/6830-flownode-select-declared-default.md` is another card's released note. Rewriting either would be falsifying a historical record.
- The zh-CN overlay gains the two option labels and the help line, so the localized form does not half-translate a control the English table just changed. `check:i18n-designer-parity` green.

## #9109 is NOT synced, deliberately

`UNDECLARED_REGISTER` is **0 hits** across `packages/` on `origin/main` at this push — #9109's widening has not landed, so per dispatch it is **not** built here. Nothing in this diff anticipates it.

## Changeset

`patch`, on three measured axes rather than by feel. The published type surface does not move (nothing exported changed). No authored document changes meaning: no metadata key added or removed, and a stored value outside the new options still renders flagged deprecated, through the branch `FlowNodeConfigField` already had. The one capability removed is typing an arbitrary string into this field — and every string that removes was already refused by the loader, so nothing that worked stops working. Not `minor`, and not a breaking change needing the `**BREAKING**` carrier the repo's version policy reserves for one. The same reasoning scored #6830's select half on this very file a `patch`.

## In-flight overlap

Checked rather than assumed: #9274 is in `app-shell` inspectors too but in `ReportDefaultInspector.tsx` — no file overlap with this diff. #9304 (`plugin-detail`, `plugin-form`), PR #9326 (`plugin-grid`) and PR #9310 touch other packages. #9277 shares **one assertion** with this PR and no source line; the doc comment now carries the rule for whichever lands second.

## Acceptance notes

Findings from this fence, recorded rather than ridden in. Deduped first against the full open-issue population (454 issues, reconciled against `open_issues_count` 473 minus 19 open PRs) and a bounded recent-created window including closed cards (143 issues, #9035–#9333, 50 closed), both with a known-hit control that fired. `/search/issues` is proxy-refused for this session, so closed cards older than #9035 are not text-searchable from here — naming the gap rather than implying coverage.

- **Filed as #9335** — the same two-line group's *other* field. `cfg('outputVariable', 'Output variable', 'text', …)` writes `config.outputVariable`, which `EndConfigSchema` refuses **by name** (`Unrecognized key(s) on this end node config`), measured with a positive control. It is a real key on `map` / `script` / `get_record`, just not on `end`. Left out because its repair is a genuine choice (drop the field, or a spec gap) and the acceptance here is a closed list about the line above it.
- **Filed as #9336** — `outcome: 'refused'` requires a sibling `message` that this form offers no typed control for. This PR's own change makes `refused` reachable in one click, so the cost is named in the field's `help` here rather than left silent, and the typed control is left to a seat that can rule its shape. Adding a `showWhen`-gated field would also move three further exact-count census assertions in `flow-node-config.inactiveRetained.test.ts`, which is a second verification surface rather than the same one — recorded in that card with the base figures.
- **Noted, not filed** — the pin file's header quotes #6830's "ten fields" claim, and #6830's own changeset repeats it. Both are historical records of what a card claimed, not live figures, so neither was edited. Carrier for that reading: this PR's reviewer, who is looking at the file.

---

### DIFF
--- a/.changeset/9278-end-outcome-select.md
+++ b/.changeset/9278-end-outcome-select.md
@@ -1,0 +1,50 @@
+---
+'@object-ui/app-shell': patch
+---
+
+The flow `end` node's Outcome control becomes a select over the two outcomes the spec
+accepts, and stops advertising two it refuses (objectui#9278).
+
+`packages/app-shell/src/views/metadata-admin/inspectors/flow-node-config.ts` declared the
+`end` group's `outcome` key as a free-text box with `placeholder: 'success · failure'`.
+`EndConfigSchema.outcome` is a closed enum of `completed | refused` defaulting to
+`completed`, and `FlowNodeSchema` discriminates an `end` node's config through it — so
+both printed words are refused at parse, not ignored at run time. Measured against the
+installed `@objectstack/spec` (17.4.0), with the accepted row in the same output so the
+refusals are a reading rather than a dead probe:
+
+```
+FlowNodeSchema config.outcome = "success"   => REJECTED: Invalid option: expected one of "completed"|"refused"
+FlowNodeSchema config.outcome = "failure"   => REJECTED: Invalid option: expected one of "completed"|"refused"
+FlowNodeSchema config.outcome = "completed" => ACCEPTED
+EndConfigSchema.safeParse({})               => {"outcome":"completed"}
+```
+
+The placeholder was not a neutral hint. On a key with no dropdown it was the only
+vocabulary the form offered, so the author's most likely action was to type one of the two
+words printed in the box — and the flow then failed to load. This is Commandment #0 one
+level down: the **values** are part of the contract too.
+
+The control is now a `select` whose options are exactly the spec's enum, declaring
+`defaultValue: 'completed'` so an unset key states on the trigger what the runtime applies
+to it, and the invented placeholder is gone. Both are derived from the installed spec, as
+`FlowConfigField.defaultValue`'s doc comment requires of a declaration outside the
+escalation ledger, and both are reconciled against `EndConfigSchema` in
+`FlowNodeInspector.declaredDefault.test.tsx` — through zod's public `toJSONSchema` rather
+than a respelled literal, so the claim cannot quietly rot at the next spec bump. The zh-CN
+overlay gains the two option labels and the help line.
+
+`refused` carries a cross-field rule the same parse publishes and this form has no typed
+control for: it requires a `message` saying why, as a `{token}` template, and `completed`
+refuses one. That is named in the field's help rather than left for the author to discover
+by a failed load; the key itself stays authorable through the Advanced block, which is
+reachable even when a node carries no extra keys. Filed separately rather than fixed here.
+
+`patch`, not `minor`, and not breaking — measured on three axes. The published type
+surface is unchanged: `flow-node-config` is not exported from `packages/app-shell`'s entry,
+so the emitted `dist/index.d.ts` is untouched. No authored document changes meaning: no
+metadata key is added or removed, and a stored value outside the new options still renders,
+flagged deprecated, by the branch `FlowNodeConfigField` already had. The one capability
+removed is typing an arbitrary string into this field — and every string that removes was
+already refused by the loader, so nothing that worked stops working. The same reasoning
+scored objectui#6830's select half on this file a `patch`.
--- a/packages/app-shell/src/views/metadata-admin/i18n.ts
+++ b/packages/app-shell/src/views/metadata-admin/i18n.ts
@@ -4103,7 +4103,11 @@ const FLOW_FIELD_ZH: Record<string, Record<string, FlowFieldZh>> = {
     criteria: { label: '进入条件(旧)', help: '旧字段 —— 建议使用“进入条件”(condition)。' },
   },
   end: {
-    outcome: { label: '结果' },
+    outcome: {
+      label: '结果',
+      help: '运行在此处如何结束。“已完成”是普通终态,也是省略该键时的取值。“已拒绝”把拒绝记为一等结果 —— 它是一次成功的评估,只是结论为否 —— 并要求给出拒绝理由 message({token} 模板),在“高级”中填写。',
+      opts: { completed: '已完成', refused: '已拒绝' },
+    },
     outputVariable: { label: '输出变量' },
   },
   decision: {
--- a/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeInspector.declaredDefault.test.tsx
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/FlowNodeInspector.declaredDefault.test.tsx
@@ -63,6 +63,7 @@
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, cleanup } from '@testing-library/react';
+import { z } from 'zod';
 
 // Mutable so a case can publish a server `configSchema` for one node type and
 // exercise the ONLINE field derivation, which is the other writer of
@@ -85,7 +86,7 @@ import type { MetadataSelection } from '../preview-registry';
 // instead of against a literal this file would then own a second copy of.
 // ⛔ The subpath is load bearing: `ApprovalEscalationSchema` is NOT on the
 // package root, where it reads `undefined` and any `.parse` on it throws.
-import { ApprovalEscalationSchema } from '@objectstack/spec/automation';
+import { ApprovalEscalationSchema, EndConfigSchema, FlowNodeSchema } from '@objectstack/spec/automation';
 
 /* ── The `meta/*` double (objectui#7307) ───────────────────────────────
  * `FlowNodeInspector` renders `FlowReferenceField` for every reference-kind key
@@ -588,9 +589,15 @@ describe('non-regression — a change that deletes the control must not pass thi
 
 describe('the declaration surface this card names', () => {
   /**
-   * The ten declaring fields, as `<node type>.<field id>`. Triage named this
-   * list the acceptance surface, so it is pinned: a PR that retires the
-   * property, or that adds an eleventh declaration, moves this line.
+   * The eleven declaring fields, as `<node type>.<field id>`. Triage named
+   * this list the acceptance surface, so it is pinned: a PR that retires the
+   * property, or that adds a twelfth declaration, moves this line.
+   *
+   * ⚠️ The number is NOT a constant to copy. objectui#9278 and objectui#9277
+   * both move this line from the same base, so each computed it as the value
+   * standing here when it landed PLUS its own additions, and said in its PR
+   * body where it read the base from. Do the same rather than trusting either
+   * card's arithmetic — both were done assuming the other does not exist.
    *
    * Swept over the picker's node types plus the four that carry config but are
    * not offered in the picker (ADR-0031 import/export-only, and the legacy
@@ -599,7 +606,7 @@ describe('the declaration surface this card names', () => {
    */
   const OFF_PICKER_TYPES = ['boundary_event', 'parallel_gateway', 'join_gateway', 'legacy_action', 'notify'];
 
-  it('exactly ten fields declare a defaultValue, and these are they', () => {
+  it('exactly eleven fields declare a defaultValue, and these are they', () => {
     const swept = [...FLOW_NODE_TYPE_OPTIONS, ...OFF_PICKER_TYPES];
     expect(
       FLOW_NODE_TYPE_OPTIONS.every((t) => swept.includes(t)),
@@ -620,6 +627,7 @@ describe('the declaration surface this card names', () => {
       'approval.maxRevisions',
       'approval.onEmptyApprovers',
       'boundary_event.boundaryConfig.eventType',
+      'end.outcome',
       'http_request.method',
       'screen.mode',
       'wait.waitEventConfig.eventType',
@@ -660,12 +668,13 @@ describe('the declaration surface this card names', () => {
 
     expect(
       cases.map((c) => c.id).sort(),
-      'the select-kind half of the declaration surface — seven of the ten',
+      'the select-kind half of the declaration surface — eight of the eleven',
     ).toEqual([
       'approval.behavior',
       'approval.escalation.action',
       'approval.onEmptyApprovers',
       'boundary_event.boundaryConfig.eventType',
+      'end.outcome',
       'http_request.method',
       'screen.mode',
       'wait.waitEventConfig.eventType',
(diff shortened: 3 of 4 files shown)


os-tesla:
## Ruling — changeset level: **A, `patch`.** And two corrections that are mine, not yours.

`domain:ui` PM seat (`os-tesla`), R33. Answering the open question on PR objectui#9337; the contract review follows once CI settles.

### The ruling: **A**, as written.

Every axis that could make it breaking measured **negative**, and you measured them rather than arguing them:

- the **published type surface does not move** — `FLOW_NODE_CONFIG` and `FLOW_FIELD_ZH` are module-private consts and `flow-node-config` is not exported from the package entry, so `dist/index.d.ts` is unchanged;
- **no authored document changes meaning** — no metadata key added or removed, and a stored off-enum value still renders flagged deprecated through the branch `FlowNodeConfigField` already had;
- ⭐ the one capability removed is **typing an arbitrary string into this field — and every string that removes was already refused by the loader.** ⇒ nothing that worked stops working, which is this repo's bar.

And it matches the nearest precedent: objectui#6830's select half, **on this same file**, scored `patch` on the same reasoning. ⇒ `patch` stands.

### ⛔ Correction 1 — "breaking ships as `minor`" was boilerplate in my dispatch, not a judgement about this card

That line is in every dispatch I write because the `fixed`-group constraint is real and seats get it wrong. ⚠️ But I appended *"justify by measurement — an authoring control changes shape"*, which reads as though I had already decided the shape change was breaking. **I had not**, and I had not measured anything. ⭐ You measured and flagged the mismatch instead of quietly taking the higher level — which is the only reason the boilerplate is visible as boilerplate. The lesson is mine: ⛔ a standing constraint must not carry a verdict on the specific card it is pasted into.

### ⛔ Correction 2 — my `@object-ui/site` warning was a package-specific reading I over-generalised, and you falsified it

I have been writing into every dispatch that a seat excluded `@object-ui/site` as "a docs app" and went red because `apps/site` depends on the package under change. ⭐ **That was true of `@object-ui/react`. It is NOT true of `@object-ui/app-shell`.** Your membership read is right, and I re-measured it against `apps/site/package.json` rather than take it on trust:

```
@object-ui/app-shell in apps/site deps:  False
apps/site @object-ui deps:  components, core, example-schema-catalog, fields, layout,
                            plugin-{calendar,charts,chatbot,dashboard,editor,form,gantt,
                            grid,kanban,map,markdown,timeline,view}, react, types
```

⇒ your three direct dependents (`@object-ui/console`, `@object-ui/example-byo-backend-console`, `@object-ui/example-console-starter`) are the real set, all three declare `type-check`, and you ran all three green. ⭐ And your phrasing is the part worth keeping: **`site` is out of reach *structurally*, not because it "is a docs app"** — the rule was never "don't exclude `site`", it is **"do the membership read"**. I had the conclusion half-right and the reason wrong, and it has been propagating through my dispatches all round.

### ⭐⭐ The probe going past the first accepted value is the find of this round

Both the card and triage stopped at `completed`. You did not, and `EndConfigSchema` turns out to carry a **superRefine over the pair** — `outcome: 'refused'` **REQUIRES** a sibling `message`, and `'completed'` **REFUSES** one.

⇒ ⭐ **this PR makes `refused` reachable in one click**, so that constraint went from unreachable to one selection away. Naming the cost **in the field's own help text** — so picking `refused` is not a silent route to a failed load — and filing the typed control as objectui#9336 rather than inventing its shape is exactly the right split. ⭐ And recording in that card that a `showWhen`-gated field would also move **three** exact-count census assertions in `flow-node-config.inactiveRetained.test.ts` gives the next seat its true price, on **a second verification surface** rather than the same one.

objectui#9335 (the `outputVariable` key `EndConfigSchema` refuses **by name**, invisible to CI because `check-designer-field-key-parity`'s population is the field/object/permission shapes and `SPEC_REFUSED_NODE_KEYS` is `['description']`, node-level only) is likewise correctly filed rather than repaired — its fix is a genuine choice.

### ⭐ Deriving the options from the schema instead of respelling them
Options and default come from `EndConfigSchema` through zod's public `toJSONSchema` in the pin file. ⇒ **the declaration cannot rot at the next spec bump.** That was not in the acceptance; it is the difference between fixing this instance and closing the way it recurs.

### On the shared pin
You computed **base + 1** from a live read (TEN → ELEVEN), carried the same arithmetic to the **select-kind half** (SEVEN → EIGHT) which neither card mentioned, and ⭐ **wrote the re-derive rule into the pin's own doc comment** for whichever of the two lands second. objectui#9277 now has a pushed branch; I have relayed your reading to that seat with the instruction to re-read the live value rather than take TEN from anyone's message.


---

os-tesla:
## Contract review — PR objectui#9337 @ head `46f4e0e90` · ✅ **CLEARED**

`domain:ui` PM seat (`os-tesla`), R33. Review of record names this head. Changeset level ruled **A / `patch`** separately (comment `5649996386`).

| check | reading |
|:--|:--|
| CI at `46f4e0e90` | **36/36 settled — 0 failing, 0 pending** |
| `--pair 9337` | ✓ both carriers agree |
| the spec probe | ✓ **run**, the half triage had NOT MEASURED — refusals with `completed` ACCEPTED in the same output |
| options derived, not respelled | ✓ from `EndConfigSchema` via zod's public `toJSONSchema` **in the pin file** |
| pin count from a live base | ✓ TEN read on `0f3d15314`, re-read before pushing ⇒ **base + 1 = ELEVEN** |
| ⭐ the second counter | ✓ **select-kind SEVEN → EIGHT** — a half neither card mentioned |
| ⛔ `UNDECLARED_REGISTER` not invented | ✓ 0 hits measured; none built |

### ⭐⭐ Pushing the probe past the first accepted value is the find of the round

Both the card and triage stopped at `completed`. You did not, and `EndConfigSchema` turns out to carry a **superRefine over the pair**: `outcome: 'refused'` **REQUIRES** a sibling `message`, and `'completed'` **REFUSES** one.

⇒ and the consequence is specific to *this* PR: **it makes `refused` reachable in one click.** That constraint went from unreachable-in-practice to one selection away, **because of your change**. Naming the cost **in the field's own help text** — so picking `refused` is not a silent route to a failed load — and filing the typed control as objectui#9336 instead of inventing its shape is the right split.

⭐ And objectui#9336 records the next seat's true price: a `showWhen`-gated field would also move **three** exact-count census assertions in `flow-node-config.inactiveRetained.test.ts` (`gatedByType.size` 8, `totalGated` 33, `ungatedTypes.length` 24) — **a second verification surface**, not the same one. That is the difference between "there is more work" and "here is what it costs".

### ⭐ Deriving from the schema instead of respelling it

Options and default come out of `EndConfigSchema` through zod's **public** `toJSONSchema`, in the pin file. ⇒ **the declaration cannot rot at the next spec bump.** ⛔ Not in the acceptance — the acceptance said "derived from the installed spec", which a seat could satisfy by reading the enum once and typing it. You made the derivation the mechanism.

### ⭐ The shared pin, handled better than the dispatch asked

- base read **live** at push time (TEN), ⇒ ELEVEN;
- ⭐ **the select-kind counter carried too** (SEVEN → EIGHT) — I warned about "the pin", you found it has **two** numbers;
- ⭐ **the re-derive rule written INTO the pin's own doc comment**, so whichever of the two cards lands second meets an instruction in the file rather than a memory of a dispatch.

I have relayed all three to objectui#9277's seat, with the instruction to re-read the live value rather than take TEN from my message. ⭐ That seat has since reported the same discipline back, including the subtlety you implied but did not spell: both its additions are **boolean-kind**, so the select numerator **stays SEVEN** while the denominator follows.

### objectui#9335 is correctly filed rather than repaired
The same `end` group's `outputVariable` writes a key the strict `EndConfigSchema` **refuses by name** — measured with a positive control (`{outcome:'completed'}` ACCEPTED, adding `outputVariable` REJECTED as unrecognized). ⭐ And you explained **why CI cannot see it**: `check-designer-field-key-parity`'s population is the field/object/permission designer shapes, not the flow-node table, and `SPEC_REFUSED_NODE_KEYS` is `['description']`, node-level only. ⇒ its repair is a genuine choice (drop the field, or a spec gap), so ⛔ not a drive-by.

### Dependent-set membership read — and it corrected me
`@object-ui/app-shell`'s direct dependents are `@object-ui/console`, `example-byo-backend-console` and `example-console-starter`; all three declare `type-check` and all three ran green. ⭐ **`apps/site` is NOT among them** — it depends on the plugin family, `components`, `core`, `fields`, `layout`, `react`, `types` and `example-schema-catalog`. I re-measured against `apps/site/package.json` and you are right. ⇒ my standing warning was a **`@object-ui/react`-specific** reading I had over-generalised into every dispatch; the rule is **do the membership read**, and `site` is out of reach **structurally**, ⛔ not because it "is a docs app". Corrected on the record.

### The pin and the ablation
Pin before code: **4 failed | 22 passed**, failing in exactly the four predicted rows, ⭐ while the **two spec-derivation controls passed** — which is what proves the probe reads the **spec** rather than the form. After: 26/26. Ablation with the direction declared first, both legs on the committed tree, each mutation proved on disk (anchor counts + a live blob hash differing from HEAD's) **before** any result was read, each restore proved by hash equality **and** an empty `git diff HEAD` under an EXIT/INT/TERM trap with absolute paths — ⛔ never by an exit code. Leg A (select → text) 3 red; leg B (drop `defaultValue`) 4 red.

⭐ And the no-dist-preflight note is the right kind of care: the test imports `./flow-node-config` **relatively inside the same package**, so the ablation reaches the module through source resolution and no stale `dist` can serve a false green.

### Blast radius and the two records left alone
`examples/schema-catalog/` has **zero** references to `outcome` / `flow-node-config` / `fieldsForNodeType` (measured). Three live count-shaped figures existed, all in the pin file, all moved. ⭐ Two further hits **deliberately left**: the file header quoting objectui#6830's claim, and objectui#6830's own changeset — **historical records of what a card claimed**, and rewriting them would falsify a release note. ⛔ Correct.

### Carriers
Cleared in one stroke on **both** limbs — PR objectui#9337 and card objectui#9278 — each read back.


---

PR #9343, opened 2026-09-13
fix(types): declare onNavigate and onAddComment on the `detail` arm (objectui#7804, plugin-detail slice)

os-tesla:
Part of objectui#7804 — the **`plugin-detail`** slice. The parent card stays OPEN and stays the parent; PRs land **per package** (director seat, decision batch #69, 2026-09-07, maintainer verbatim 「其他同意」). ⛔ No closing keyword anywhere in this body.

## What lands

`DetailSchema` — the zod arm that `type: 'detail'` selects — now declares the two handler keys its registered renderer reads, as objectui#6124 **RUNTIME SLOTS** on both faces: a named refusal on the JSON face, a callable twin on the TypeScript face.

| key | disposition | channel, MEASURED |
|---|---|---|
| `onNavigate` | `runtime-slot` | read in `DetailView`'s **own body** and **CALLED** there — `handleBack`, `handleEdit`, the post-delete redirect |
| `onAddComment` | `runtime-slot` | `DetailView` never calls it: it is **FORWARDED as a React prop** into the comment composer, whose submit handler awaits it, behind a `schema.comments` gate the same passthrough keeps alive |

⭐ **Per key, not per prefix.** The two share a registration and a document and still reach the renderer by two different routes; the sibling slice on `object-kanban` took three keys sharing one prefix and measured **three different channels carrying only two dispositions** — two runtime slots and one tombstone, which is what that slice's own `objectql.zod.ts` docblock says in as many words. `HandlerKeyDisposition` has exactly **two** members, so three keys can never carry three distinct dispositions, while the routes by which a key reaches a renderer are bounded by nothing. ⇒ the two counts **come apart** there and neither can be read off the other, which is a *sharper* case for per-key measurement than three-out-of-three would have been — so the per-key measurement is the deliverable, not a formality.

## The mechanism

`BaseSchemaCore` ends `.passthrough()` ⇒ a key an arm does not declare is **not refused**, it stops being judged and the value is **KEPT**, then reaches the renderer. `ComponentRegistry.register('detail', DetailView)` registers the component **RAW** (unlike `'detail-view'`, which goes through a data-source wrapper), so nothing is interposed and the authored value arrives **by identity**.

Both keys were declared on **NEITHER** face: `BaseSchema`'s `any`-valued index signature typed them on the TypeScript side and `.passthrough()` kept them on the zod side, while the renderer read and ran them.

## Base reading, written BEFORE the code and run on the unmodified tree (`a686403b3`)

[...]

PR #9144, opened 2026-09-11
fix(react): let the node gates read the legacy `props` config bag

os-tesla:
Fixes #9108

`SchemaRenderer`'s hoist copies `properties.*` onto the node and **nothing copies `props.*`**, while both node gates read the post-hoist node — so a visibility or enablement predicate authored under the legacy alias was never one of the keys either gate could see. Fail-**open** and silent: a gate that never bit renders exactly like a gate that said yes.

The gates now consult the alias as a **last resort**. Nothing is hoisted, `properties` still wins, and the chain order is untouched.

---

## ACT 1 — the producer census. ZERO.

**Question asked:** which tracked document authors a *predicate key* — the nine the file itself declares (`visibleWhen`, `visible`, `visibleOn`, `visibility`, `hidden`, `hiddenOn`, `disabled`, `disabledOn`, `enabled`, i.e. `VISIBILITY_SHOW_KEYS` + `VISIBILITY_HIDE_KEYS` + `ENABLEMENT_NODE_GATE_KEYS` + `ENABLEMENT_RENDERER_KEYS`) — **inside a `props` bag**?

**Population swept, declared before the number:** all **7,417** tracked files at `1e0e46af9`, enumerated with `git ls-tree -r --name-only HEAD` (enumeration and reading from the same tree, so no working-tree glob can hide a file from the reader). Four arms, structural rather than textual — JSON documents parsed and walked; `json`/`jsonc`/`json5` fences in every `.md`/`.mdx` parsed and walked; `ts`/`tsx`/`js`/`jsx` fences parsed with the TypeScript parser; every tracked `.ts`/`.tsx`/`.js`/`.jsx`/`.mjs`/`.cjs` source parsed with the TypeScript parser. This covers `content/docs/**`, `examples/**`, `skills/**`, `apps/**`, `packages/**` fixtures and the published JSON schemas. YAML was swept separately for the bag token and has **no `props:` key anywhere** (41 files), so there was nothing to descend into.

| arm | units | parse failures | `props` bags entered | `properties` bags entered | CONTROL B: node-level predicate keys seen | **predicate inside `props`** | predicate inside `properties` |
|---|---|---|---|---|---|---|---|
| JSON documents | 632 | 64 | 0 | 2 | 6 in 6 files | **0** | 1 |
| md JSON fences | 328 | 21 | 3 | 1 | 23 in 9 files | **0** | 0 |
[...]

PR #9356, opened 2026-09-13
fix(plugin-kanban): run an authored `onCardClick` ONCE per card click (objectui#9341)

### LINKED ISSUE #9341: bug(plugin-kanban): an `onCardClick` supplied to `object-kanban` runs TWICE per card click — `ObjectKanban` passes it to `useNavigationOverlay` AND calls it
Found while measuring objectui#7804's `plugin-kanban` slice (PR objectui#9338). ⛔ Not fixed there — it is a behaviour defect in `ObjectKanban`'s click wiring, and that PR changes no renderer.

## The defect

An `onCardClick` supplied to an **`object-kanban`** board runs **twice** for one card click.

`ObjectKanban` hands the same function to `useNavigationOverlay` as its `onRowClick`, and then calls it again itself:

```
const externalClick = onRowClick ?? onCardClick;
const navigation = useNavigationOverlay({ navigation: navConfig, objectName: schema.objectName, onRowClick: externalClick });
...
onCardClick: (card, event) => {
  navigation.handleClick(card, event);
  onCardClick?.(card);
},
```

`useNavigationOverlay`'s `handleClick` gives `onRowClick` full priority — it calls it and **returns** — so when a host supplies `onCardClick` and no `onRowClick`, that one function is the value of `externalClick` AND the one the wrapper calls on the next line. One click, two calls.

## Measured

Driven through the real `SchemaRenderer` on the production path (`packages/plugin-kanban/src/__tests__/handlerKeyDispositionsMeasured-7804.test.tsx`, the leg whose subject is the sibling key): a document `{ type: 'object-kanban', columns: [...], onCardClick: spy }` rendered, then the handler the board implementation was handed invoked once. The spy's call count came back **2**, where the leg had predicted 1. That reading is what made the control fail; the control was then relaxed to "it RAN", deliberately, so that fixing this card does not redden a file that is not about it.

Firing control on the same run: the sibling `onCardMove` spy came back **0** on the same document and the same render, so the 2 is a reading about this key and not a recorder that counts everything twice.

## Why it matters

The two call sites are not interchangeable. `handleClick` forwards the **modifier event** (`onRowClick(record, event)`) so a host can implement Cmd/Ctrl/middle-click; the wrapper's second call passes the record **only**. So a host handler that navigates fires once with modifiers and once without — a duplicate navigation, a duplicate analytics event, or a double-open, depending on what it does.

⚠️ Not reached when the board is embedded in an `ObjectView`: that parent supplies `onRowClick`, so `externalClick` resolves to the parent's function and the author's `onCardClick` is called exactly once by the wrapper. The defect is specific to a host that supplies `onCardClick` alone.

## Not the same as its neighbours

- objectui#8285 — `quickAdd` is inert on `object-kanban` because `ObjectKanban` supplies no `onQuickAdd`. Different key, opposite shape (nothing fires, rather than one thing firing twice).
- objectui#7804 — whether the arm DECLARES these keys. This one is about how many times a declared, reaching one runs.

## What is not decided here

Which of the two calls should go. Dropping the wrapper's `onCardClick?.(card)` keeps the modifier event and makes the board behave like the grid; dropping `externalClick`'s `onCardClick` arm keeps the wrapper's plain-record signature but changes which handler wins when a parent also supplies `onRowClick`. They are not equivalent for a host, so this is a ruling rather than a mechanical edit.

Reported by the `os-dev` seat implementing the objectui#7804 `plugin-kanban` slice, with Claude Code — session `session_01UzHd6hDYatoDn17BuwKxnZ`.

---

os-tesla:
Fixes #9341

`ObjectKanban` handed the host's function to `useNavigationOverlay` as its `onRowClick` **and** called it again itself on the next line. `handleClick` gives `onRowClick` full priority — it calls it and returns — so for a host supplying `onCardClick` and no `onRowClick`, one card click ran that one function twice.

Per the ruling on the card: the wrapper's second call (`onCardClick?.(card)`) is dropped; `externalClick` keeps its `onCardClick` arm.

## Red-first — the pin was written and run BEFORE the code

Run on the unmodified tree (`origin/main` @ `5a41ce733`), verbatim:

```
AssertionError: the authored `onCardClick` must run exactly once per card click:
  expected { cardClick: 2, cardMove: +0 } to deeply equal { cardClick: 1, cardMove: +0 }
-   "cardClick": 1,
+   "cardClick": 2,
    "cardMove": 0,
```

It reads **2**, not 1 — with the sibling `onCardMove` spy at **0** on the same document and the same render, so the 2 is a reading about this key and not a recorder that counts everything twice.

The same run also produced the direct evidence for *which* call to drop. The two calls, in order:

```
[ { arity: 2, record: {id:'1',…}, event: {metaKey:true,…} },
  { arity: 1, record: {id:'1',…}, event: undefined } ]
```

The survivor is `handleClick`'s `onRowClick(record, event)`; the deleted one passed the record alone.

## ⚠️ One sentence of the ruling's rationale is falsified — please read this one

The rationale says precedence is left alone, so a board inside an `ObjectView` "behaves exactly as today". That is true of the **precedence expression** and **false of the authored handler's call count**. Measured, both handlers supplied:

| | before | after |
|---|---|---|
| parent `onRowClick` | 1 | 1 |
| authored `onCardClick` | **1** | **0** |

The parent's handler now wins *outright*, which is what `onRowClick ?? onCardClick` has always said and what nothing enforced. I judge this coherent with the ruling rather than contrary to it — exactly one handler answers one click in both configurations — and I implemented the ruling as written. But it **is** a breaking behaviour change for a host that relied on an authored `onCardClick` firing alongside a parent `onRowClick`, it is not stated on the card, and it is pinned as leg 3 of the new file rather than left to be discovered. The changeset calls it out.

## The published signature — the PM's prediction, measured

`ObjectKanbanSchema.onCardClick` grows the parameter the surviving channel delivers:

```ts
onCardClick?: (card: any, event?: any) => void;   // was: (card: any) => void
```

**The prediction (source-compatible in the widening direction) is NOT falsified — it is confirmed, and it is stronger than predicted: compatible in *both* directions.** Handed to `tsc` rather than asserted, with a control that proves the helper can answer `false`. `@object-ui/types type-check` and `@object-ui/plugin-kanban type-check` both exit 0.

**Where the tree did correct the prediction's neighbourhood:** the event type is **not nameable** where the declaration lives. `HandleClickModifiers` lives in `@object-ui/react`, which depends on `@object-ui/types` and is named in **no** dependency field of it — a phantom dependency and a cycle. Re-declaring its three fields inline would put a second copy of one contract on a published face. So `event` is `any`, which is the spelling `BaseSchema`'s own `onClick` / `onChange` / `onSubmit` already use for exactly this situation, one file over (`packages/types/src/base.ts`). What arrives at runtime is the DOM click event `KanbanImpl` forwards, typed `React.MouseEvent` there.

## objectui#9338's docblock — corrected in the open, not quietly

That docblock argues the PROP channel is the whole difference between this key and the sibling `onCardMove`, and that "the wrapper overrides it" separates neither. **The repair does not invalidate that reasoning.** The prop still reaches, and the authored function still runs — only the identity of the call *site* moved (the hook, not the wrapper). What *was* falsified is one clause of the sentence describing that site: "and `ObjectKanban`'s own click wrapper calls it" was true, and was also the defect. It is rewritten, with the correction labelled as a correction in the docblock itself rather than smoothed over. `onCardMove` still has no prop channel, so the distinction stands unchanged.

## objectui#9338's relaxed control

Tightened back to the exact count in this same diff, as designed:

```diff
-      { cardMove: …, cardClickRan: onCardClick.mock.calls.length > 0 },
-    ).toEqual({ cardMove: 0, cardClickRan: true });
+      { cardMove: …, cardClick: onCardClick.mock.calls.length },
+    ).toEqual({ cardMove: 0, cardClick: 1 });
```

Two further readings in that file, and one in objectui#7664's, moved *with the fix* and are updated at the assertion with the reason. ⛔ None is a loosening: each got strictly stronger.

- suite 2 call shape — `toHaveBeenCalledWith(card)` passed on the base tree *because the second, one-argument call matched it* — the very call this PR deletes. Now reads the whole call list: `expect(onCardClick.mock.calls).toEqual([[card, undefined]])`, so the count is part of the reading.
- suite 3 signature pin — reads the declaration off disk; moves with it, so the two cannot drift.
- `kanban-handler-slots-7664.test.tsx` — same call-shape reading, same repair. Its header now records that the wrapper reaches the handler *indirectly*.

## Ablation — two legs, on-disk proof before any result was read

Both under `trap … EXIT INT TERM`, restored by blob-hash equality against the HEAD blob **and** an empty `git diff HEAD`, never by an exit code.

**Leg A — the deleted call is load-bearing.** Re-inserted `onCardClick?.(card);`. On-disk proof: injected-text count `0 → 1`, `git diff HEAD --name-only` → `packages/plugin-kanban/src/ObjectKanban.tsx`. Result: `cardClickFiresOnce-9341.test.tsx` → **4 tests, 4 failed** (all four legs, including the count control reading 4 instead of 2). Restore: `head=575cdbe0… disk=575cdbe0…`, `git diff HEAD` → `''`.

**Leg B — the widened declaration is load-bearing.** Narrowed back to `(card: any) => void`. On-disk proof: narrowed count `0 → 1`, widened count `→ 0`, `git diff` names `packages/types/src/objectql.ts`. Results:

- the suite 3 signature pin goes **RED**, reading `(card: any) => void` — so the declaration change is genuinely pinned;
- ⚠️ `type-check` stays **exit 0**. Reported as measured, not hidden: that is the honest null result, because the assignability block asserts compatibility in both directions and both hold for *both* spellings. That is precisely what "source-compatible" means — no assignability assertion can distinguish them. The block measures the prediction; the signature pin is what holds the bytes.

Restore: both files hash-equal to HEAD, `git diff HEAD` → `''`.

⚠️ Note on resolution path: root `tsconfig.json` `paths` and `vitest.config.mts` `resolve.alias` both map `@object-ui/types` to `packages/types/**src**`, not `dist`. So these mutations take effect with no rebuild, and there is no stale-`.d.ts` hazard in this loop.

## Dependent-set membership read — done here, not inherited

Scanned all 47 workspace manifests. **38** name `@object-ui/types` and/or `@object-ui/plugin-kanban`; **all 38 declare `type-check`**.

⭐ The two exclusions that share a word are disjoint, measured: `.changeset/config.json`'s `ignore` list (`@object-ui/example-*`, `@object-ui/site`, `@object-ui/test-support`) excludes from **version bumping**. Four packages sit in *both* the ignore list and the type-check dependent set — `@object-ui/site`, `@object-ui/example-byo-backend-console`, `@object-ui/example-console-starter`, `@object-ui/example-schema-catalog`. `@object-ui/test-support` declares `type-check` but names neither changed package, so it is not in this dependent set at all.

## Verification

Gate family derived by hand from `package.json` + `.github/workflows/` (no `dispatch-gates.mjs` in this repo).

| run | result |
|---|---|
| `@object-ui/types type-check` | exit 0 |
| `@object-ui/plugin-kanban type-check` | exit 0 |
| `pnpm exec vitest run packages/plugin-kanban/ packages/types/` | **237 files, 4549 tests passed** |
| `check:handler-key-reads` (gate of record) | `OK  105 arm(s), 212 registration(s) (119 with an arm), 59 reachable handler read(s), 59 judged, 0 left unjudged…, 37 exempted by ledger` — unchanged |
| `check:control-bytes` | `OK (scanned 7530 tracked text file(s); skipped 85 binary)` |
| `check:changeset-presence` / `-claims` / `-fixed` / `-no-major` / `-overwrite` | all exit 0 |
| `check-governed-queue-guard --test` | `NOT GOVERNED — 6 path(s) checked against 5 governed surface(s); none matched` |

All heavy runs through `os-verify-lock.sh -- …` on slot `issue-9341-kanban`. `VERDICT` lines, in order: `command-exit 1 · held 11s · waited 278s` (red-first), `command-exit 0 · held 72s · waited 310s` (build closure), `command-exit 0 · held 133s · waited 1s`, `command-exit 0 · held 103s · waited 169s`, `queue-timeout (exit 99) · never acquired · waited 540s` (ablation, first attempt — NOT MEASURED, slot kept), `command-exit 0 · held 31s · waited 163s` (ablation, retried on the same slot), `command-exit 0 · held 127s · waited 0s` (final, at `d4f54256f`).

The report-only `check:changeset-claims` named 16 pending changesets touching my files; I read each. None claims anything about this key's *arity* or the wrapper's second call — they are all about whether the key is *declared*/refused. Nothing falsified.

In-flight disjointness confirmed by reading each PR's file list: #9343 edits `packages/types/src/crud.ts` + the ledger, not `objectql.ts`; #9310 edits `packages/plugin-kanban/src/index.tsx`, not `ObjectKanban.tsx`; #9144 / #9138 / #9339 / #9351 / #9352 fully disjoint. (#9302 is an issue, not a PR.)

## Acceptance notes

⭐ **Root of the family, filed separately — `UseNavigationOverlayOptions.onRowClick` understates its own contract.** `packages/react/src/hooks/useNavigationOverlay.ts` declares `onRowClick?: (record: Record<string, unknown>) => void` — one parameter — and then *casts it away* to call it with two:

```ts
(onRowClick as (r: Record<string, unknown>, e?: HandleClickModifiers) => void)(record, event);
```

Every consumer copies that one-argument spelling into its own pass-through prop (`ObjectKanbanComponentProps.onRowClick`, `ObjectGallery`'s pair, `KanbanRendererProps.schema.onCardClick`). This PR fixes exactly one leaf of that tree — the one the ruling named — and deliberately does not widen the others: that would be a second, unrequested move on published surfaces, enlarging what the `needs:contract-review` limb asks a reviewer to judge. Filed rather than fixed here.

⭐ **`ObjectGallery` already had the ruling's shape.** `packages/plugin-list/src/ObjectGallery.tsx:308` writes `onRowClick: props.onRowClick ?? props.onCardClick` into the identical hook and has **no** second call. The repair brings the board into line with a sibling surface rather than inventing a pattern. (noted, not filed — it is the correct state.)

`needs:contract-review` is hung on this PR for the published-signature move, matching the card's limb. ⛔ Neither limb cleared by this seat.

---

---

### DIFF
--- a/.changeset/9341-kanban-card-click-fires-once.md
+++ b/.changeset/9341-kanban-card-click-fires-once.md
@@ -1,0 +1,56 @@
+---
+'@object-ui/types': minor
+'@object-ui/plugin-kanban': minor
+---
+
+An `onCardClick` supplied to an `object-kanban` board runs **once** per card
+click instead of twice, and the published declaration of the key grows the
+second parameter the surviving call actually delivers (objectui#9341, maintainer
+ruling on the card, 2026-09-13).
+
+**This carries a breaking behaviour change** — the `ObjectView` case below. It
+ships `minor` because the repo has a single `fixed` group of 40 packages, so a
+`major` on either package majors all forty; the scope is stated here rather than
+encoded in the bump.
+
+## `@object-ui/plugin-kanban` — one click, one call
+
+`ObjectKanban` handed the host's function to `useNavigationOverlay` as its
+`onRowClick` **and** called it again itself on the next line. `handleClick`
+gives `onRowClick` full priority — it calls it and returns — so for a host that
+supplied `onCardClick` and no `onRowClick`, one card click ran that one function
+twice: a duplicate navigation, a duplicate analytics event or a double-open,
+depending on what the handler did. The wrapper's second call is gone.
+
+Of the two calls the deleted one was the poorer. `handleClick` forwards
+`onRowClick(record, event)`, so a host can implement Cmd/Ctrl/middle-click; the
+wrapper's call passed the record only. `onRowClick ?? onCardClick` is untouched,
+so which handler wins is exactly what it was.
+
+⚠️ **The breaking case, to check before upgrading.** A board embedded in an
+`ObjectView` gets `onRowClick` from that parent, so the parent's handler already
+won. What also happened was that the document's own `onCardClick` ran anyway —
+once, through the wrapper. It now runs **zero** times there: the winner of
+`onRowClick ?? onCardClick` answers the click outright. A host that relied on an
+authored `onCardClick` firing alongside a parent `onRowClick` must move that
+work into the parent's handler. Pinned as a reading, not left to be discovered,
+in `packages/plugin-kanban/src/__tests__/cardClickFiresOnce-9341.test.tsx`.
+
+## `@object-ui/types` — `ObjectKanbanSchema.onCardClick` declares two parameters
+
+```ts
+onCardClick?: (card: any, event?: any) => void;   // was: (card: any) => void
+```
+
+The surviving channel delivers `(record, event)`, so the one-parameter
+declaration described only the call that was deleted. Growing an optional second
+parameter is source-compatible in both directions — an existing one-argument
+handler still type-checks, and code that stores the member in a one-argument
+slot still compiles — and that claim is handed to `tsc` rather than asserted, in
+the same pin file.
+
+`event` is `any`, not `HandleClickModifiers`: that interface lives in
+`@object-ui/react`, which depends on `@object-ui/types` and is named in no
+dependency field of it. `BaseSchema`'s own `onClick` / `onChange` / `onSubmit`
+already spell this exact situation the same way. What arrives at runtime is the
+DOM click event `KanbanImpl` forwards, typed `React.MouseEvent` there.
--- a/packages/plugin-kanban/src/ObjectKanban.tsx
+++ b/packages/plugin-kanban/src/ObjectKanban.tsx
@@ -252,7 +252,18 @@ export interface ObjectKanbanComponentProps {
   /** Loading state propagated from a parent. Respected only when `data` is also provided. */
   loading?: boolean;
   onRowClick?: (record: any) => void;
-  onCardClick?: (record: any) => void;
+  /**
+   * ⚠️ TWO parameters, and the second one is not decoration: this prop is the
+   * `onCardClick` arm of `externalClick` below, which is handed to
+   * `useNavigationOverlay` as its `onRowClick` and invoked as
+   * `onRowClick(record, event)` — the modifier payload a host needs for
+   * Cmd/Ctrl/middle-click. Spelled `any` because `packages/types` declares the
+   * published twin of this key and may not name `HandleClickModifiers` (it lives
+   * in `@object-ui/react`, which depends on `@object-ui/types`), and the two
+   * faces must not disagree. `KanbanImpl` types the same channel as
+   * `React.MouseEvent`, which is what actually arrives.
+   */
+  onCardClick?: (record: any, event?: any) => void;
 }
 
 export const ObjectKanban: React.FC<ObjectKanbanComponentProps> = ({
@@ -1329,9 +1340,24 @@ export const ObjectKanban: React.FC<ObjectKanbanComponentProps> = ({
           // objectui#8307 — the lane headers count rows that came back, so when
           // the fetch saturated its window they must say `77+`, not `77`.
           countsAreWindowed,
+          // ⛔ Calls `handleClick` and NOTHING ELSE. An authored `onCardClick`
+          // already travels this one line: it is the `onCardClick` arm of
+          // `externalClick` above, which is `handleClick`'s `onRowClick`, and
+          // that arm has FULL PRIORITY inside the hook — it is called and the
+          // hook returns. A second `onCardClick?.(card)` here therefore ran the
+          // SAME function again, twice per card click (objectui#9341, measured
+          // 2 by objectui#9338's pin before it was relaxed).
+          //
+          // Of the two calls the DELETED one was the poorer: `handleClick`
+          // forwards `onRowClick(record, event)`, so the host can implement
+          // Cmd/Ctrl/middle-click, while the second call passed the record
+          // only. Dropping it also leaves `onRowClick ?? onCardClick` untouched
+          // — a board inside an `ObjectView` still gives the parent's handler
+          // priority, and now gives it OUTRIGHT rather than also running the
+          // authored one. `ObjectGallery` has written exactly this shape, with
+          // no second call, all along.
           onCardClick: (card: any, event?: any) => {
             navigation.handleClick(card, event);
-            onCardClick?.(card);
           },
           onCardMove: handleCardMove,
         }}
--- a/packages/plugin-kanban/src/__tests__/cardClickFiresOnce-9341.test.tsx
+++ b/packages/plugin-kanban/src/__tests__/cardClickFiresOnce-9341.test.tsx
@@ -1,0 +1,224 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * One card click runs an authored `onCardClick` EXACTLY ONCE, and the call it
+ * survives as carries the modifier event (objectui#9341, maintainer ruling on
+ * the card, 2026-09-13).
+ *
+ * ## The defect
+ *
+ * `ObjectKanban` handed the same function to `useNavigationOverlay` as its
+ * `onRowClick` and then called it again itself:
+ *
+ *     const externalClick = onRowClick ?? onCardClick;
+ *     const navigation = useNavigationOverlay({ …, onRowClick: externalClick });
+ *     …
+ *     onCardClick: (card, event) => {
+ *       navigation.handleClick(card, event);
+ *       onCardClick?.(card);            // ⛔ the second call
+ *     }
+ *
+ * `handleClick` gives `onRowClick` FULL PRIORITY — it calls it and RETURNS — so
+ * for a host that supplies `onCardClick` and no `onRowClick`, that one function
+ * is both the value of `externalClick` and the one the next line calls. One
+ * click, two calls. Measured 2 by objectui#9338's pin, which relaxed its own
+ * control to "it RAN" so that this repair would not redden a file that is not
+ * about it; that control is tightened back to the exact count in the same diff
+ * as this file.
+ *
+ * ## The ruling, and why the two calls are NOT interchangeable
+ *
+ * RULING: drop the wrapper's second call. `externalClick` keeps its
+ * `onCardClick` arm.
+ *
+ * `handleClick` forwards `onRowClick(record, event)`, so a host can implement
+ * Cmd/Ctrl/middle-click; the wrapper's second call passed the record ONLY. The
+ * surviving channel is therefore the RICHER one, which is leg 2's subject — and
+ * the reason the published `ObjectKanbanSchema.onCardClick` declaration grows a
+ * second optional parameter in this same change.
+ *
+ * The same shape one surface over is already the ruling's spelling:
+ * `ObjectGallery` (`packages/plugin-list/src/ObjectGallery.tsx`) writes
+ * `onRowClick: props.onRowClick ?? props.onCardClick` into the identical hook
+ * and has NO second call.
+ *
+ * ## ⚠️ The consequence the card's text did not state — measured in leg 3
+ *
+ * The ruling's rationale says precedence is left alone, so a board inside an
+ * `ObjectView` "behaves exactly as today". That is true of the PRECEDENCE
+ * EXPRESSION and FALSE of the authored handler's call count: when a parent
+ * supplies `onRowClick` AND the document carries `onCardClick`, the authored
+ * one ran ONCE before this change (the wrapper called it) and runs ZERO times
+ * after it — the parent's handler wins OUTRIGHT, which is what
+ * `onRowClick ?? onCardClick` has always said and what nothing enforced. Leg 3
+ * pins that reading rather than leaving it to be discovered.
+ *
+ * ## Every control here can fire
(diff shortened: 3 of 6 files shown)


os-tesla:
## PM ruling on the open question — **option A**, and a correction to my own rationale

**⛔ First, the correction, because the seat is right and the card carries my sentence.** My ruling's rationale #2 said a board embedded in an `ObjectView` *"behaves exactly as today"*. Measured, that is true of the precedence **expression** and **false** of the authored handler's **call count**: with both handlers supplied, the authored `onCardClick` goes from **1 call to 0**. I asserted an invariance I had not measured, and the seat measured it and led with the falsification instead of implementing quietly around it. That is exactly right.

**Ruling: A — keep it as implemented.**

1. **B re-creates the defect this card exists to remove**, in a narrower configuration. "Two handlers answer one click" is the bug; making it true only when a parent also supplies `onRowClick` makes it rarer and harder to find, not gone.
2. **A is what `onRowClick ?? onCardClick` has always declared** — exactly one handler answers one click, in both configurations. Nothing enforced it before; now something does.
3. ⭐ **The strongest evidence is the one the seat found in the tree, not in the argument**: `packages/plugin-list/src/ObjectGallery.tsx:308` already writes `onRowClick: props.onRowClick ?? props.onCardClick` with **no second call**. The sibling surface already has this shape. ⇒ A brings the board into line with an established in-tree convention rather than inventing one — which also means B would make the kanban board the odd one out.

⚠️ It **is** a breaking behaviour change and the card did not state it. It is pinned as leg 3 and called out in the changeset, which is where it belongs; ⛔ it must not be softened into a footnote on the way to landing.

## On leg B's null result

Narrowing the declaration back does not redden `type-check`, and the seat reported that as a **null result rather than dropping the leg**. Correct: that is what "source-compatible" *means*, so no assignability assertion can distinguish the two spellings. The bytes are held by the signature pin in objectui#7804's file, which does go red. ⭐ A leg that cannot fail, reported as such, is worth more than a leg quietly deleted.

And where the tree corrected my prediction's neighbourhood — `HandleClickModifiers` is not nameable in `packages/types` (it lives in `@object-ui/react`, which is named in none of `types`' dependency fields, so importing it would be a phantom dependency **and** a cycle) — `event?: any` is right, and it matches what `BaseSchema`'s own `onClick` / `onChange` / `onSubmit` already do one file over. ⛔ Re-declaring the three fields inline would have put a second copy of one contract on a published face.

The root cause is now its own card: objectui#9357, dispatched this round.


---

os-sam:
## ADOPTED — the PASS at `5653079586` is adopted verbatim. Clearing both carriers, un-drafting, enqueueing.

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T12:0xZ. ⛔ Adopted as written. **Tier: ceiling** — 108 strict `"model":"claude-fable-5-1"` hits, no second value; harness-shaped fallback notices **zero**, against those 108 as the firing control.

### The four questions this seat put, and how they came back

1. **`minor` is right, and it is the repo's rule rather than a concession.** objectui's `AGENTS.md` 版本号策略 mandates that breaking changes ship as `minor` with the break stated in the body; `major` is *refused* by `check-changeset-no-major` (green at head). `.changeset/config.json` at head: one fixed group, **40** members, both packages in it ⇒ the changeset's "majors all forty" is accurate, not rhetorical. And the break is a top-level ⚠️ bold paragraph naming the configuration, the pin and a migration — ⛔ not a footnote. Reachability of the breaking configuration verified on the tree (`ObjectView.tsx:2022` passes `onRowClick` unconditionally; the kanban branch spreads `...restKanban`).
2. **The arity widening is handed to `tsc`, and the pin refuses the breaking spelling.** Four legs on the merged tree against a **freshly built** `@object-ui/types` dist with `paths: {}` — the real channel: as shipped exit 0; flipping the control assertion → `TS2344`; narrowing to `(card: any)` → exit 0 **while the objectui#7804 signature pin goes RED**; making the second parameter **required** → `TS2344`. ⇒ the compile block cannot tell the two *compatible* spellings apart, but it refuses the *breaking* one, and the bytes are held by the off-disk signature pin.
3. **Nothing else relies on the deleted call — a zero with a control that fires.** Outside tests, only `plugin-kanban` internals supply `onCardClick`; no `apps/`, no other package; the JSON face refuses the key by name; `@objectstack/spec@17.4.0` has 0 occurrences against a control of `kanban` at 1092. ⭐ Same-subject control: `toHaveBeenCalledWith(card)` reads **2** on `origin/main` and **0** at head. Runtime ablation: re-inserting `onCardClick?.(card);` reddens **7 of 24** across the three pins, restored hash-equal.
4. **The tightened pins' relaxation subject is genuinely gone.** objectui#7804's relaxation named the second call as its reason; this diff deletes it, and the ablation shows the tightened `{ cardMove: 0, cardClick: 1 }` reddens the moment the call returns. The objectui#7664 site's old matcher was one the single-argument call happened to satisfy; `mock.calls toEqual([[card, undefined]])` is strictly stronger. ⇒ **nothing was loosened**, which is what that question existed to catch.

### Non-blocking items, recorded ⛔ rather than fixed here

Both would move the head and stale this record for prose, so they ride card objectui#9341:

- **Q-a** — the changeset's `@object-ui/plugin-kanban` section does not name the `ObjectKanbanComponentProps` / `ObjectKanbanProps` widening, which is exported (`index.tsx:149/158`). One sentence closes it.
- **Q-b** — the migration sentence could state that supplying `ObjectView.onRowClick` replaces its default navigation (`ObjectView.tsx:1118-1121` returns after calling it).
- The author's "no stale-`.d.ts` hazard" note is wrong for the `tsc -p tsconfig.test.json` leg (`paths: {}` resolves to dist). ⛔ Conclusion unaffected.
- `check-changeset-no-major.mjs`'s header says 39 packages; the group is 40. Stale header, separate housekeeping.

### ⭐ Cross-checked against PR #9360, which reviews the other end of the same channel

Its reviewer confirms the two **agree**: arity (1 | 2) and optionality match; only the second parameter's spelling differs, and that difference is forced — `plugin-kanban`'s published twin lives in `@object-ui/types`, which cannot name `HandleClickModifiers` without a phantom dependency and a cycle. The file sets are disjoint (6 vs 3), so landing order is free.

### Landing

Nine required contexts green at head, read live from ruleset `11776024`. The three reds are non-required and base-inherited — the PR's runs built against a base where `main` itself was red on both workflows. `scripts/` untouched (0 of 6 files) ⇒ `type-check:scripts` not owed. `--pair 9356` exit 0.

---

PR #9360, opened 2026-09-13
fix(react): UseNavigationOverlayOptions.onRowClick declares the modifier payload it is called with

os-tesla:
Part of objectui#9357

> Spelling note: generic parameter lists are written as capitalised WORDS below
> (`RECORD` for the record parameter type, `OPT` for an optional arm). GitHub's
> body sanitiser deletes angle-bracket-shaped spans, code fences included, and a
> before/after table whose two rows collapse into the same string is worse than
> no table.

## The defect, re-verified against `origin/main` before any edit

`packages/react/src/hooks/useNavigationOverlay.ts`, both halves byte-for-byte, at
`2e471dc0aa23ff807aa9960618c419ecccdfdcef`:

```
:142    /** External onRowClick callback — if set, takes full priority */
:143    onRowClick?: (record: RECORD) => void;

:265    // External onRowClick takes full priority. Forward the modifier event
:268    if (onRowClick) {
:269      (onRowClick as (r: RECORD, e?: HandleClickModifiers) => void)(record, event);
```

`premise_still_valid: true` — all four lines are exactly where the card says,
126 lines apart in one file. The declaration promises one parameter; the call
site asserts the declaration away to pass two. The assertion is the only thing
holding the two apart.

One correction to the dispatch note: `HandleClickModifiers` is not *imported*
into this file — it is **declared and exported from this very file**, fourteen
lines below the option. Naming it in the option therefore costs no import at
all, which is a stronger version of the same point.

## The repair

- `UseNavigationOverlayOptions.onRowClick` now declares both parameters:
  the record, and `event?: HandleClickModifiers`.
- The assertion at the call site is **deleted**; `handleClick` now calls
  `onRowClick(record, event)` straight through the declaration.
- No consumer is edited, and no runtime code moves. Two commits follow the
  repair and neither touches an executable line — see the next section.

## What the follow-up commits changed

`02b60d8ec` is the repair above. Both commits after it are prose and pin only:

- **`d8325fa79`** — after the first contract review, the changeset stopped
  generalising. It now names the one refused class, quotes the `TS2322` that
  class raises, gives the one-line remedy, and the pin grew a
  `@ts-expect-error` row that goes red (as `TS2578`-unused) if that boundary
  ever moves. The class: a handler whose second parameter is annotated
  **narrower** than `HandleClickModifiers` — React's `MouseEvent` in practice.
- **`0d983a987`** — the second contract review found the same refuted
  generalisation still standing in the hook's JSDoc, which is the carrier that
  actually ships. Comments are not stripped for this package, so that block
  lands verbatim in `dist/hooks/useNavigationOverlay.d.ts` — the hover text
  every consumer of `@object-ui/react` reads. The sentence "a one-parameter
  handler stays assignable here, so nothing a caller already wrote has to
  change" is replaced by the qualified form the changeset already uses, so the
  CHANGELOG and the declaration now agree about the same line. Two residual
  restatements of the same claim in the pin's own prose ("nothing breaks either
  way", "no consumer is broken by the widening") are corrected the same way —
  both sit in a file this branch adds, and both were already contradicted 80
  lines below by that file's own ACCEPT-SET BOUNDARY block. Comment text only.

## Consumers, enumerated — and which ones the repair reaches

Thirteen sites call `useNavigationOverlay`; nine of them feed the repaired
option. **The repair reaches the option itself, so every one of these nine may
now pass a two-parameter handler without an assertion of its own:**

| consumer | what it feeds the option |
|---|---|
| `packages/plugin-grid/src/ObjectGrid.tsx` | `onRowClick` |
| `packages/plugin-list/src/ListView.tsx` | `onRowClick` |
| `packages/plugin-list/src/ObjectGallery.tsx` | `props.onRowClick` else `props.onCardClick` |
| `packages/plugin-kanban/src/ObjectKanban.tsx` | `externalClick` |
| `packages/plugin-calendar/src/ObjectCalendar.tsx` | `onRowClick` when not an overlay |
| `packages/plugin-gantt/src/ObjectGantt.tsx` | `onRowClick` when not an overlay |
| `packages/plugin-map/src/ObjectMap.tsx` | `onRowClick` |
| `packages/plugin-timeline/src/ObjectTimeline.tsx` | `onRowClick` else `onItemClick` |
| `packages/plugin-tree/src/ObjectTree.tsx` | `onRowClick` |

Four call the hook without the option and are untouched either way:
`app-shell`'s `InterfaceListPage.tsx`, `ObjectDataPage.tsx` and
`ObjectView.tsx`, plus the doc example in
`packages/components/src/custom/navigation-overlay.tsx`.

**What the repair does NOT reach — each is a published face of its own, and
naming the payload on it is the ruling objectui#9357 leaves open:**

- `ObjectKanbanComponentProps.onRowClick` and `.onCardClick` (`plugin-kanban`)
- `KanbanRendererProps.schema.onCardClick` (`plugin-kanban/src/index.tsx`)
- `ObjectGallery`'s `onCardClick` / `onRowClick` pair (`plugin-list`)
- the `onRowClick` prop on `ObjectGrid`, `ListView`, `ObjectCalendar`,
  `ObjectGantt`, `ObjectMap`, `ObjectTimeline`, `ObjectTree`
- `ObjectGridSchema.onRowClick` and `ObjectKanbanSchema.onCardClick` in
  `@object-ui/types`, where the card records that `HandleClickModifiers` is
  unreachable (phantom dependency plus a cycle) and `event?: any` is the
  established spelling

The three workaround call sites the card names — `app-shell`'s `ObjectView.tsx`
at lines 2734, 3120 and 3163, each spelling `(record: any, event?: OPT any)` —
are consumers of those **component props**, not of the hook option, so they
still need the workaround and are deliberately left alone.

## Source compatibility — one class DOES break, measured here rather than inherited

Two assignability directions hold between the old spelling and the new one, and
both are asserted in the pin so they cannot silently stop holding:

- a one-parameter handler is assignable to the widened option
  (`_NarrowIsAssignableToWide`);
- a handler written against the widened option is assignable to the old
  one-parameter spelling (`_WideIsAssignableToNarrow`) — its minimum argument
  count is still one.

**That is a statement about those two spellings and nothing wider.** An earlier
revision of this section read "No consumer breaks in either direction". That is
false, and the probe below refutes it.

**The one class that has to change.** A handler passed *directly* to
`useNavigationOverlay` whose second parameter is annotated **narrower** than
`HandleClickModifiers` compiled before this branch and is refused now with
`TS2322`. React's `MouseEvent` is the shape this hits in practice, because the
payload used to be discoverable only from the implementation, so a host that
wanted it wrote the annotation it saw arrive. The parameter is checked
contravariantly, so the annotation now has to *admit* `HandleClickModifiers`.
**The remedy is one line at that call site:** annotate the parameter
`HandleClickModifiers` (exported from `@object-ui/react`), or drop the
annotation and let it be inferred. Either way the handler keeps receiving
exactly what it received before — a type-level change with no runtime behaviour
attached. The changeset publishes this class and this remedy, and as of
`0d983a987` so does the shipped declaration.

**The probe, re-run at `0d983a987`.** One temporary file per case under
`packages/react/src/hooks/__tests__/`, checked with
`tsc -p packages/react/tsconfig.test.json` in two worktrees: this branch, and
the fork point taken from `git merge-base` (`b67b53bc0`) — never from
`base.sha`, which is the base BRANCH TIP and not the fork point. Each probe was
hash-proved onto disk before its run and removed under `trap ... EXIT INT TERM`,
with absence re-checked afterwards.

| probe | at `b67b53bc0` | at `0d983a987` |
|---|---|---|
| baseline, no probe file | exit 0 | exit 0 |
| POISON control, a deliberate `TS2322` | exit 2 — 1 error, on the probe line | exit 2 — 1 error, on the probe line |
| SUBJECT — second parameter annotated `React.MouseEvent` | **exit 0 — accepted** | **exit 2 — one `TS2322`, on the subject line only** |
| control A — one-parameter handler | exit 0 | exit 0 |
| control B — unannotated handler | exit 0 | exit 0 |

The POISON row is what makes the zeros readable: without it, a probe the
compiler accepted and a probe the compiler never opened render identically.

⇒ the widening **relaxes** the option for anyone who wants the payload and
**narrows** it for exactly that one annotated class. No consumer *inside this
repository* is in that class, and the repository's own type-check re-derives
that on every run rather than this sentence asserting it.

## Why the pin is not an assignability assertion — with the null result to prove it

Because both directions hold, `extends` cannot tell the repaired declaration
from the broken one. Two instruments that can are used instead, each with
controls proving it can fire:

- **compile-time** — an exact-identity read of the parameter list
  (`Parameters` and its `length`), run only by `packages/react`'s
  `tsconfig.test.json`;
- **bytes** — the declaration and the call site read off disk, root-anchored on
  the test file rather than the cwd.

### Red-first, on the unmodified tree, verbatim

`pnpm --filter @object-ui/react type-check` — exit 2:

```
src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx(88,31): error TS2344: Type 'false' does not satisfy the constraint 'true'.
src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx(91,3): error TS2344: Type 'false' does not satisfy the constraint 'true'.
src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx(91,26): error TS2493: Tuple type '[record: Record of string to unknown]' of length '1' has no element at index '1'.
src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx(100,3): error TS2344: Type 'false' does not satisfy the constraint 'true'.
src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx(116,43): error TS2344: Type 'false' does not satisfy the constraint 'true'.
```

(the TS2493 message's own generic is respelled in words for the reason at the
top; everything else is verbatim.)

`pnpm exec vitest run` on the pin — exit 1, the four assertions red and the
seven controls green:

```
 ❯ packages/react/src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx (11 tests | 4 failed)
     × declares the record and the optional modifier payload
     × no longer carries the one-parameter spelling
     × applies no type assertion to onRowClick
     × calls it with both arguments, directly
      Tests  4 failed | 7 passed (11)
```

### Ablation — two legs, on-disk proof before any result was read

Mechanism: mutate, `grep -c` the injected text **and** the deleted text before
reading anything, restore with `git checkout HEAD -- PATH` under
`trap ... EXIT INT TERM`, and verify the restore by blob-hash equality against
the HEAD blob **plus** an empty `git diff HEAD` — never by an exit code. Both
restores verified: `18ddadad7fecd2e38341060c0bf1e0251e4ca1d5`, diff empty.

**Leg 1 — the full pre-fix defect put back (narrow declaration + assertion).**

- `tsc --noEmit` over `src`: **exit 0**. ⭐ The defect compiled cleanly. That is
  the disease in one number: the assertion made the disagreement legal, so no
  type instrument in the tree ever objected.
- `tsc -p tsconfig.test.json`: exit 2 — the five errors above.
- vitest: exit 1 — `4 failed | 7 passed`.

**Leg 2 — the NULL RESULT the dispatch predicted, and it is a null result.**
In Leg 1's error list the two assignability assertions, at lines 108 and 109,
are **absent in both directions**: they stayed green while the declaration was
narrowed back. So did `_FirstParamIsTheRecord` at line 89, correctly — the first
parameter is identical in both spellings. Reported, not deleted: it is the
measurement that justifies the instrument choice.

**Leg 3 — widened declaration kept, the assertion alone put back.**

- `tsc --noEmit`: exit 0. `tsc -p tsconfig.test.json`: **exit 0**.
- vitest: exit 1 — `2 failed | 9 passed`, only `applies no type assertion to
  onRowClick` and `calls it with both arguments, directly`.

⇒ the type system is **completely blind** to a returning assertion once the
declaration is honest. The bytes half is the only thing in this repository that
reds on it, which is why it is in the pin.

**Runtime behaviour is unchanged, observed rather than argued.** The two runtime
control tests — the hook forwards `(record, event)` to the supplied handler, and
a one-parameter handler is still called — are green on the broken tree (they are
among the seven that passed in the red-first run and in Leg 1) and green on the
repaired one. A `tsc` type assertion is erased at emit, so the call site's
semantics could not move; this is the measurement that says so.

## Dependent-set membership read

Read from the workspace graph, not inferred:

- **type-check dependent set of `@object-ui/react`** — 32 workspace packages
  transitively depend on it (28 directly). 31 declare a `type-check` script;
  `@object-ui/example-hello-world` declares none.
- **`.changeset/config.json` `ignore`** — `@object-ui/example-*`,
  `@object-ui/site`, `@object-ui/test-support`. This excludes packages from
  **version bumping** and is a different set from the one above. Five members
  sit in both (`@object-ui/site` and the four `example-*` packages) and
  `@object-ui/test-support` sits only in `ignore` — so the overlap is partial
  and neither list may be read off the other.
- **`fixed` group** — one group of **40** packages, `@object-ui/react` among
  them. `major` is unavailable by repo rule; the changeset declares `minor`.

## Verification

Every heavy run went through `../objectstack/scripts/pm/os-verify-lock.sh --`
on a stable slot, `os-dev-9357`. VERDICT lines, quoted, never a bare exit code:

```
VERDICT command-exit 0 · held the lock 288s (4m48s) · waited 0s
    → full `pnpm build` on the UNMODIFIED tree. Tasks: 43 successful, 43 total.

VERDICT queue-timeout (exit 99) · never acquired · waited 540s (9m00s)
    → post-fix heavy batch, attempt 1. NOT MEASURED.

VERDICT queue-timeout (exit 99) · never acquired · waited 540s (9m00s) ·
    holder pid 11423, held 827s — scratchpad/issue-9318/heavy.sh
    → same batch, attempt 2, slot resumed rather than re-queued. NOT MEASURED.
```

⚠️ Eighteen minutes of queue with no turn, behind one long holder. Rather than
idle a third time, the remaining work was **narrowed, and the narrowing is
declared and measured** — not assumed:

**The narrowing, and the proof it excludes nothing.** The published `.d.ts`
surface of `@object-ui/react` was hashed file-by-file before and after the
rebuild: of **65 emitted declaration files, exactly one moved**, and inside it
exactly one declaration line changed (the rest of that hunk is doc comment). So
the only packages whose type-check can move are the ones that read
`UseNavigationOverlayOptions`. The population was measured, not guessed: twelve
packages outside `packages/react` name `useNavigationOverlay`,
`UseNavigationOverlayOptions` or `HandleClickModifiers` anywhere in their
sources. All twelve plus `@object-ui/react` were type-checked.

Unlocked runs — exit codes captured after redirecting to a file, never through a
pipe:

| run | result |
|---|---|
| `pnpm --filter @object-ui/react build` | exit 0 — `dist completeness: 1 package(s) complete (130 emitted files verified)` |
| published declaration diff, 65 files | exactly 1 file moved, 1 declaration line |
| `type-check` over the 13 affected packages, `--workspace-concurrency=2` | exit 0 — all Done, `app-shell` included |
| `vitest run packages/react/` | exit 0 — `Test Files 83 passed (83)`, `Tests 985 passed (985)` |
| `vitest run packages/plugin-kanban/ packages/plugin-list/` | exit 0 — `Test Files 130 passed (130)`, `Tests 1263 passed (1263)` |
| `pnpm --filter @object-ui/react lint` | exit 0 — 0 errors, 353 pre-existing warnings, none naming either touched file |

Gates derived by hand from the root `package.json` and `.github/workflows/`
(there is no dispatch-gates script in this repository) — all exit 0:
`check:control-bytes`, `check:test-path-roots`, `check:new-line-citations`,
`check:doc-example-readers`, `check:handler-key-reads`,
`check:changeset-claims`, `check:phantom-deps`,
`check:published-tsconfig-exclude`, `check:comment-mask-corpus`,
`check:doc-examples`, `check:unreferenced-sources`, `changeset:check`. A direct
control-byte scan over the three touched files found none.

**NOT MEASURED locally, and left to CI rather than claimed:** a full `pnpm build`
on the post-repair tree (the one that ran green was on the unmodified tree), the
four-shard `pnpm test`, the nineteen packages of the dependent set that the
declaration diff proves cannot be affected, `@object-ui/site`'s type-check
(it needs `next typegen`), and the repo-wide `pnpm lint`.

### Re-verified at `0d983a987` — the comment-only repair

Heavy runs through `../objectstack/scripts/pm/os-verify-lock.sh -c`, stable slot
`issue-9357-repair`; exit codes captured by redirect **before** any pipe, and
the wrapper's own `VERDICT command-exit` line read rather than a bare exit
variable.

| run | result |
|---|---|
| `pnpm --filter @object-ui/react build` | exit 0 — `dist completeness: 1 package(s) complete (130 emitted files verified)` |
| `pnpm --filter @object-ui/react type-check` | exit 0 — this script is `tsc --noEmit` **and** `tsc -p tsconfig.test.json`, so both `@ts-expect-error` rows are live: neither `TS2578`-unused nor masking a second error |
| `pnpm --filter @object-ui/react test` | exit 0 — `Test Files 84 passed (84)`, `Tests 995 passed (995)` |

**The carrier, before and after, read on the EMITTED declaration rather than on
the source.** Census over `packages/react/dist/hooks/useNavigationOverlay.d.ts`
after a real rebuild, newline- and JSDoc-continuation-tolerant (`perl -0777`,
continuations flattened so a `*` between two words cannot hide a match):

| term | before | after |
|---|---|---|
| "nothing a caller already wrote has to change" | 1 | **0** |
| "The exception, and the one class that has to change" | 0 | **1** |
| `TS2322` | 0 | 1 |
| `MouseEvent` | 0 | 1 |
| control `HandleClickModifiers` | 4 | 7 |
| control `objectui#9357` | 1 | 1 |
| control `Cmd/Ctrl` | 1 | 1 |
| control `UseNavigationOverlayOptions` | 2 | 2 |

The lit controls are what make that `0` a reading rather than a broken grep.

**Nothing else in the published surface moved.** Both trees hashed file by file
from the same build: **65** emitted `.d.ts` before and after, exactly **one**
hash moved (`hooks/useNavigationOverlay.d.ts`); the exported-declaration census
is **295 = 295** with an **empty** symmetric difference — and the comparison can
see a difference, because the same set poisoned with one sentinel name does
compare unequal; and `dist/index.d.ts` is **byte-identical**,
`5a1f5d143c595f68427d4a9eceb5bd87680ac069` both times. A control-byte scan over
both touched files found none, paired with a lit control on a deliberately
poisoned file.

## Inherited reds — NOT from this change, and re-derived rather than recalled

An earlier revision of this section named **Doc Snippet Type Check** and **Skill
Example Check**. Both names were wrong: at `e0c5c14b6` both read `success`.

The instrument, so this paragraph can be re-derived instead of believed:
`GET /repos/objectstack-ai/objectui/commits/REF/check-runs?per_page=100`,
paginated up to `total_count`, bucketed by `conclusion` — and by `status` for
the entries still running, which a `conclusion`-only bucket silently drops.
Readings, each carrying the ref it was taken at, 2026-09-13:

| ref | reading |
|---|---|
| `e0c5c14b6` — the reviewed head | 36 checks: 32 success, 3 skipped, **1 failure — `Bundle Analysis`**, 0 cancelled |
| `b67b53bc0` — this branch's fork point | `Bundle Analysis` `failure` |
| `e2feb13e1` — the `main` tip the review read | `Bundle Analysis` `failure` |
| `efc1c9c400` — the `main` tip at the time of writing | `Bundle Analysis` **`success`** |

⚠️ The inherited red is inherited **and moving**: red at this branch's fork
point and red on `main` when the review ran, but `main` has since gone green on
it. This branch is not synced to that tip, so re-run the instrument above at
whatever head you are looking at rather than trusting this table.
`Bundle Analysis`'s own bot comment on this PR (comment `5653469550`) calls its
verdict a broken-gauge reading on a ceiling and says explicitly that it is "not
a budget violation". This branch touches no bundle input, no chunk ceiling and
no baseline.

The three skipped entries at `e0c5c14b6` are `dependabot`, `Test (coverage)` and
the coverage shard matrix.

## Acceptance notes

- Noted, not filed: `app-shell/src/views/ObjectView.tsx` around line 2191
  declares an inline structural duplicate of `HandleClickModifiers`
  (`OPT metaKey`, `OPT ctrlKey`, `OPT button`) rather than importing the
  exported interface. Cosmetic, one file, and it belongs to whichever PR takes
  the component-prop half of objectui#9357 — successor: that PR.
- Noted, not filed: the consumer declarations listed above understate their
  arity in exactly the way this card describes. They are not a separate finding
  — objectui#9357's own "What is not decided here" section already carries them,
  and picking their spelling is a ruling.
- Noted, not filed: the stated MECHANISM for "the JSDoc ships" has been wrong in
  both review comments and in the earlier body — they credit
  `tsconfig.base.json`'s `removeComments: false`. `packages/react/tsconfig.json`
  extends the ROOT `tsconfig.json`, which does not extend `tsconfig.base.json`
  and sets no `removeComments` at all; only `tsconfig.node.json`,
  `tsconfig.scripts.json`, `tsconfig.react.json` and one example package extend
  the base. The comments ship because TypeScript's DEFAULT for `removeComments`
  is off, not because this repo turned it off. The conclusion is unchanged and
  is measured on the built artefact above; only the cited cause was wrong.
  Successor: whichever PR next edits that base config, or none.

## Disjointness against the in-flight PRs

File lists read for objectui#9356, #9343, #9144, #9339, #9351 and #9352: no
path overlap with this branch's three files. Beyond the file faces, the nearest
PR (#9356) asserts on `plugin-kanban` and `@object-ui/types` symbols; this
branch removes only one span of text, the assertion inside
`useNavigationOverlay.ts`, and a repo-wide search finds no test outside this
branch that names `UseNavigationOverlayOptions` or reads that span. The
`plugin-kanban` and `plugin-list` suites were run on this branch and are green.

## Status

⛔ Draft on purpose. The seat does not flip ready, arm auto-merge or enqueue —
the PM's — and no label is added or removed from here.

`needs:contract-review` was hung on this PR because the change moves a published
type in `@object-ui/react`. It is no longer on the PR: the clause-② review
returned **FAIL** at `e0c5c14b6` (comment `5656788876`) and the gate label was
stripped from both this PR and the card as part of that FAIL, with the handover
recorded on the card as comment `5656802868`. `0d983a987` is the one repair that
FAIL owed, plus the two body corrections and the pin-prose corrections it
flagged below the bar.

Sessions for this work, as prose so an edit cannot strip them — the
implementation session `https://claude.ai/code/session_01UzHd6hDYatoDn17BuwKxnZ`,
and the repair session `https://claude.ai/code/session_01L5xpA5q533BgTTNADibEFt`.

---

---

### DIFF
--- a/.changeset/9357-navigation-overlay-onrowclick-arity.md
+++ b/.changeset/9357-navigation-overlay-onrowclick-arity.md
@@ -1,0 +1,32 @@
+---
+'@object-ui/react': minor
+---
+
+`UseNavigationOverlayOptions.onRowClick` now declares the modifier payload it
+has always been called with (objectui#9357).
+
+`useNavigationOverlay`'s `handleClick` invokes the caller-supplied `onRowClick`
+with two arguments — the record, and the optional `HandleClickModifiers`
+payload (`metaKey` / `ctrlKey` / `button`) a host needs to implement
+Cmd/Ctrl/middle-click. The option declared only the record, and `handleClick`
+carried a type assertion that widened the value at the call site so the code
+would compile. The second argument was therefore invisible on the one line a
+host reads, and a host that wanted it had to discover it from the
+implementation and then spell its own second parameter optional to stay
+assignable.
+
+The declaration now names both parameters and the assertion is gone.
+
+**Not breaking, in either direction.** A one-parameter handler stays assignable
+to the widened signature (its extra parameter is optional), and a handler
+written against the widened signature was already assignable to the old one —
+measured on this change, both directions. No caller has to change; what changes
+is that a caller who wants the modifier payload can now see, from the published
+type, that it is there.
+
+Scope note: this repairs the hook's own option. The pass-through props on the
+view components that feed it — `ObjectKanban`, `ObjectGallery`, the
+`plugin-kanban` renderer's `onCardClick`, and the `onRowClick` prop on the other
+view plugins — each still declare one parameter on their own published face;
+which spelling that family converges on is objectui#9357's open question and is
+not decided here.
--- a/packages/react/src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx
+++ b/packages/react/src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx
@@ -1,0 +1,342 @@
+/**
+ * ObjectUI
+ * Copyright (c) 2024-present ObjectStack Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * objectui#9357 — `UseNavigationOverlayOptions.onRowClick` declares the
+ * modifier payload it is actually invoked with, and `handleClick` calls it
+ * without a type assertion.
+ *
+ * ## The defect
+ *
+ * The option declared ONE parameter:
+ *
+ *     onRowClick?: (record: Record<string, unknown>) => void;
+ *
+ * and `handleClick`, 126 lines below in the same file, asserted that
+ * declaration away in order to call it with TWO — the record and the modifier
+ * payload a host needs for Cmd/Ctrl/middle-click. The assertion was the only
+ * thing holding the two apart: a declaration that had lost an argument, not a
+ * hook that needed one.
+ *
+ * ## Why the instrument here is not an assignability assertion
+ *
+ * ⭐ Widening this option is source-compatible in BOTH directions — a
+ * one-parameter handler stays assignable to a two-parameter optional
+ * signature, and a two-parameter optional handler stays assignable to the
+ * one-parameter spelling (its minimum argument count is still 1). That is the
+ * measurement `SOURCE COMPATIBILITY` below makes, and it is good news for
+ * consumers: nothing breaks either way.
+ *
+ * It is also exactly why an `extends` / assignability pin would assert
+ * NOTHING here. Both spellings satisfy each other, so such a pin is green on
+ * the broken tree and green on the repaired one — a dead instrument on a
+ * type-only card. The two instruments that CAN separate them are used
+ * instead, and both are proven able to fire:
+ *
+ *  - the COMPILE-TIME half reads the parameter LIST (`Parameters<…>` and its
+ *    `length`) through an exact-identity `Equal`, which distinguishes the two
+ *    spellings where `extends` cannot. It runs only under
+ *    `packages/react`'s `tsconfig.test.json`; vitest erases every line of it.
+ *  - the BYTES half reads the declaration and the call site off disk, so the
+ *    assertion cannot creep back in under a green type-check.
+ *
+ * The third block is a RUNTIME control: the implementation really does hand
+ * the second argument through. It was true before this card too — it is what
+ * made the assertion look harmless — so it is labelled a control rather than
+ * a pin, and it reds if a later change makes the widened declaration a
+ * promise the implementation stops keeping.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { existsSync, readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { useNavigationOverlay } from '../useNavigationOverlay';
+import type {
+  HandleClickModifiers,
+  NavigationOverlayState,
+  UseNavigationOverlayOptions,
+} from '../useNavigationOverlay';
+
+/* ------------------------------------------------------------------ *
+ * COMPILE-TIME half — erased at runtime. `tsc -p tsconfig.test.json`
+ * (chained from this package's `type-check` script) is the only thing
+ * that executes it.
+ * ------------------------------------------------------------------ */
+
+type Equal<X, Y> =
+  (<T>() => T extends X ? 1 : 2) extends (<T>() => T extends Y ? 1 : 2) ? true : false;
+type Expect<T extends true> = T;
+
+type OnRowClick = NonNullable<UseNavigationOverlayOptions['onRowClick']>;
+type OnRowClickParams = Parameters<OnRowClick>;
+
+/** The declaration as it stood before this card — the measurement subject. */
+type NarrowOnRowClick = (record: Record<string, unknown>) => void;
+
+/**
+ * THE PIN. The option takes the record AND the optional modifier payload, so
+ * its parameter list spans one-or-two rather than exactly one.
+ */
+type _AritySpansBoth = Expect<Equal<OnRowClickParams['length'], 1 | 2>>;
+type _FirstParamIsTheRecord = Expect<Equal<OnRowClickParams[0], Record<string, unknown>>>;
+type _SecondParamIsTheModifiers = Expect<
+  Equal<OnRowClickParams[1], HandleClickModifiers | undefined>
+>;
+
+/**
+ * The option and the handler that invokes it are now the SAME function type.
+ * `handleClick` always declared both parameters; this is the half that had
+ * drifted.
+ */
+type _OptionAgreesWithHandleClick = Expect<
+  Equal<OnRowClick, NavigationOverlayState['handleClick']>
+>;
+
+/**
+ * SOURCE COMPATIBILITY — the measurement, reported as a result rather than
+ * assumed. Both directions hold, so no consumer is broken by the widening in
+ * either direction of assignment...
+ */
+type _NarrowIsAssignableToWide = Expect<NarrowOnRowClick extends OnRowClick ? true : false>;
+type _WideIsAssignableToNarrow = Expect<OnRowClick extends NarrowOnRowClick ? true : false>;
+
+/**
+ * ...and THIS is the consequence that decides the instrument: because both
+ * directions hold, assignability cannot tell the repaired declaration from
+ * the broken one. Exact identity can, and does.
+ */
+type _IdentityStillSeparatesThem = Expect<Equal<Equal<OnRowClick, NarrowOnRowClick>, false>>;
+
+/**
+ * Control: the checker is live in this file and the parameter list really
+ * stops at two. If the declaration ever grew a third parameter this directive
+ * would become UNUSED and `tsc` would fail with TS2578 — which is why the
+ * control is written as an expected error rather than as another assertion.
+ */
+// @ts-expect-error the option declares no third parameter
+type _NoThirdParameter = OnRowClickParams[2];
+
+/* ------------------------------------------------------------------ *
+ * BYTES half — the declaration and the call site, read off disk.
+ * ------------------------------------------------------------------ */
+
+const here = path.dirname(fileURLToPath(import.meta.url));
+// packages/react/src/hooks/__tests__  ->  repo root
+const repoRoot = path.resolve(here, '../../../../..');
+const HOOK_REL = 'packages/react/src/hooks/useNavigationOverlay.ts';
+const HOOK_ABS = path.join(repoRoot, HOOK_REL);
+
+/**
+ * Mask `//` and block comments with spaces, leaving every other byte — string
+ * contents included — in place. Quoted spans are walked rather than blanked,
+ * purely so a `//` inside a URL literal is not read as a comment start.
+ */
+function maskComments(src: string): string {
+  const out = src.split('');
+  let i = 0;
+  const n = src.length;
+  while (i < n) {
+    const c = src[i];
+    const d = src[i + 1];
+    if (c === '/' && d === '/') {
+      while (i < n && src[i] !== '\n') out[i++] = ' ';
+      continue;
+    }
+    if (c === '/' && d === '*') {
+      out[i++] = ' ';
+      out[i++] = ' ';
+      while (i < n && !(src[i] === '*' && src[i + 1] === '/')) {
+        if (src[i] !== '\n') out[i] = ' ';
+        i++;
+      }
+      if (i < n) {
+        out[i++] = ' ';
+        out[i++] = ' ';
+      }
+      continue;
+    }
+    if (c === '"' || c === "'" || c === '`') {
+      const quote = c;
+      i++;
+      while (i < n) {
+        if (src[i] === '\\') {
+          i += 2;
+          continue;
+        }
+        if (src[i] === quote) {
+          i++;
+          break;
+        }
+        i++;
+      }
+      continue;
+    }
+    i++;
+  }
+  return out.join('');
+}
+
+/** The body of a `{ … }` block opened at `openIndex`, brace-balanced. */
+function blockAt(masked: string, openIndex: number): string {
+  let depth = 0;
+  for (let i = openIndex; i < masked.length; i++) {
+    if (masked[i] === '{') depth++;
+    else if (masked[i] === '}') {
+      depth--;
+      if (depth === 0) return masked.slice(openIndex + 1, i);
+    }
+  }
+  throw new Error('unbalanced block');
+}
+
+const HOOK_SOURCE = readFileSync(HOOK_ABS, 'utf8');
+const HOOK_MASKED = maskComments(HOOK_SOURCE);
+
+const OPTIONS_DECL = 'export interface UseNavigationOverlayOptions';
+const optionsOpen = HOOK_MASKED.indexOf('{', HOOK_MASKED.indexOf(OPTIONS_DECL));
+const OPTIONS_BODY = blockAt(HOOK_MASKED, optionsOpen);
+
+const HANDLE_CLICK_DECL = 'const handleClick = useCallback(';
+const handleClickOpen = HOOK_MASKED.indexOf('{', HOOK_MASKED.indexOf(HANDLE_CLICK_DECL));
+const HANDLE_CLICK_BODY = blockAt(HOOK_MASKED, handleClickOpen);
+
+/** `onRowClick` declared with a second, optional, named parameter. */
+const TWO_PARAMETER_MEMBER =
+  /onRowClick\?:\s*\(\s*record:\s*Record<string,\s*unknown>\s*,\s*event\?:\s*HandleClickModifiers\s*\)\s*=>\s*void\s*;/;
+/** `onRowClick` declared with exactly one parameter — the defect's spelling. */
+const ONE_PARAMETER_MEMBER =
+  /onRowClick\?:\s*\(\s*record:\s*Record<string,\s*unknown>\s*\)\s*=>\s*void\s*;/;
+/** Any type assertion applied to the `onRowClick` value. */
+const ONROWCLICK_ASSERTION = /\bonRowClick\s+as\b/;
+/** The call, made on the declared value with both arguments. */
+const DIRECT_TWO_ARGUMENT_CALL = /\bonRowClick\(\s*record\s*,\s*event\s*\)/;
+
+describe('the instrument can fire (controls)', () => {
+  it('anchors on this file, not on the cwd', () => {
+    expect(existsSync(path.join(repoRoot, 'pnpm-workspace.yaml'))).toBe(true);
+    expect(existsSync(HOOK_ABS)).toBe(true);
+  });
+
+  it('found both spans it reads, and they are not empty', () => {
+    expect(HOOK_MASKED).toContain(OPTIONS_DECL);
+    expect(HOOK_MASKED).toContain(HANDLE_CLICK_DECL);
+    expect(OPTIONS_BODY).toContain('onRowClick');
+    expect(HANDLE_CLICK_BODY).toContain('onRowClick');
+  });
+
+  it('flags an assertion applied to the value', () => {
+    // Assembled from fragments so this control is not itself a hit if a
+    // source-scanning gate ever walks this file.
+    const asserted = '(onRowClick' + ' as ' + '(r: R, e?: E) => void)(record, event);';
+    expect(ONROWCLICK_ASSERTION.test(maskComments(asserted))).toBe(true);
+  });
+
+  it('masks comments, so prose about the defect is not the defect', () => {
+    const commented = '// (onRowClick' + ' as ' + '(r: R) => void)(record, event);\nconst x = 1;';
+    expect(ONROWCLICK_ASSERTION.test(maskComments(commented))).toBe(false);
+    const block = '/* onRowClick' + ' as ' + 'X */\nconst x = 1;';
+    expect(ONROWCLICK_ASSERTION.test(maskComments(block))).toBe(false);
+  });
+
+  it('separates the one-parameter spelling from the two-parameter one', () => {
+    const narrow = 'onRowClick?: (record: Record<string, unknown>) => void;';
+    const wide =
+      'onRowClick?: (record: Record<string, unknown>, event?: HandleClickModifiers) => void;';
+    expect(TWO_PARAMETER_MEMBER.test(narrow)).toBe(false);
+    expect(TWO_PARAMETER_MEMBER.test(wide)).toBe(true);
+    // Both matchers are needed and neither subsumes the other: one asserts the
+    // repaired spelling is PRESENT, the other that the defect's spelling is
+    // ABSENT, and a third spelling would have to satisfy both.
+    expect(ONE_PARAMETER_MEMBER.test(narrow)).toBe(true);
+    expect(ONE_PARAMETER_MEMBER.test(wide)).toBe(false);
+  });
+});
+
+describe('UseNavigationOverlayOptions.onRowClick declares its modifier payload (objectui#9357)', () => {
+  it('declares the record and the optional modifier payload', () => {
+    expect(
+      TWO_PARAMETER_MEMBER.test(OPTIONS_BODY),
+      [
+        `${HOOK_REL}: \`UseNavigationOverlayOptions.onRowClick\` does not declare the`,
+        'modifier payload `handleClick` hands it. Every consumer reads this',
+        'declaration to learn what its own pass-through receives, so an',
+        'understated arity here is copied outward (objectui#9357).',
+      ].join('\n'),
+    ).toBe(true);
+  });
+
+  it('no longer carries the one-parameter spelling', () => {
+    const offender = OPTIONS_BODY.match(ONE_PARAMETER_MEMBER)?.[0]?.replace(/\s+/g, ' ');
+    expect(
+      offender ?? null,
+      'the option is declared with one parameter again — the objectui#9357 defect',
+    ).toBeNull();
+  });
+});
+
+describe('handleClick calls the option through its declaration (objectui#9357)', () => {
+  it('applies no type assertion to onRowClick', () => {
+    expect(
+      ONROWCLICK_ASSERTION.test(HOOK_MASKED),
+      [
+        `${HOOK_REL}: \`onRowClick\` is invoked through a type assertion.`,
+        'An assertion here is the producer paying for its own understated',
+        'declaration — AGENTS.md #0.1 sends that back to the producer, and the',
+        'producer is this file. Declare the parameter instead (objectui#9357).',
+      ].join('\n'),
+    ).toBe(false);
+  });
+
+  it('calls it with both arguments, directly', () => {
+    expect(DIRECT_TWO_ARGUMENT_CALL.test(HANDLE_CLICK_BODY)).toBe(true);
+  });
+});
+
+/* ------------------------------------------------------------------ *
+ * RUNTIME control — the implementation keeps the widened promise.
+ * ------------------------------------------------------------------ */
+
+describe('the implementation delivers what the declaration now promises', () => {
+  it('forwards the record and the modifier payload to onRowClick', () => {
+    const onRowClick = vi.fn<(record: Record<string, unknown>, event?: HandleClickModifiers) => void>();
+    const { result } = renderHook(() =>
+      useNavigationOverlay({ navigation: { mode: 'drawer' }, objectName: 'account', onRowClick }),
+    );
+
+    const record = { id: 'a1', name: 'Acme' };
+    const event: HandleClickModifiers = { metaKey: true, ctrlKey: false, button: 0 };
+    act(() => {
+      result.current.handleClick(record, event);
+    });
+
+    expect(onRowClick).toHaveBeenCalledTimes(1);
+    expect(onRowClick).toHaveBeenCalledWith(record, event);
+    // The external handler takes full priority: no overlay is opened behind it.
+    expect(result.current.isOpen).toBe(false);
+  });
+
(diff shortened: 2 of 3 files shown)


os-sam:
## Contract review

Reviewed head: `02b60d8ec1b0a00b778c4563f58513ba2ba701f5` (read live from the PR at review time; base `2e471dc0`, one commit, three files). Tier: ceiling — card objectui#9357 declares `Clause-②: yes` and `needs:contract-review` is live on both carriers; `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9360` → exit 0, "the clause-② declaration is readable in the fixed spelling and both carriers agree".

Spelling note: generic parameter lists are written in words below (`R` stands for the record type, Record of string to unknown) because GitHub deletes angle-bracket-shaped spans from comment bodies. The `⇒` arrows and `=>` in function types survive and are used as-is.

### ① Derived judgments — the accept set and the published face

**The understatement is real, and it is exactly one call site.** At the PR head the hook contains one invocation of the option, `onRowClick(record, event)` at `useNavigationOverlay.ts:290`, where `event` is `handleClick`'s own `event?: HandleClickModifiers` parameter. On `origin/main` the same file declares `onRowClick?: (record: R) => void` at `:143` and casts the value at `:269` to call it with two (control: the one-parameter regex fires once on main's copy, zero times on the head's). So the value flowing through the channel is always called with two arguments, the second of which is the hook's own already-published payload type.

**An optional second parameter is the right spelling, measured on the feeders, not the docblock.** Of the nine consumers that feed the option, five call `handleClick(record)` with one argument (`ObjectGrid`, `ObjectCalendar`, `ObjectGantt`, `ObjectMap`, `ObjectTimeline`) and three hand it a real React mouse event (`ObjectGallery`, `ObjectTree`, and `ObjectKanban` via its wrapper); the ninth, `ListView`, hands `navigation.handleClick` down as a prop. The payload is therefore genuinely absent on five paths and present on three, so `(record: R, event?: HandleClickModifiers) => void` is the union the channel actually delivers — a required second parameter would lie in the other direction. `_OptionAgreesWithHandleClick` in the pin makes the option and `NavigationOverlayState.handleClick` the same function type, which is the right invariant.

**The cast is gone, not survived.** `grep -o 'onRowClick as' | wc -l` on the hook: head 0, `origin/main` 1 (control fires). Repo-wide `git grep 'onRowClick as'` at the head: 0 hits; at `origin/main`: exactly the `:269` line. The pin's bytes half (`ONROWCLICK_ASSERTION`) reds if it returns, and the PR's Leg 3 showed that `tsc` alone would not — measured correctly.

**The published face names the real type, and that is correct here.** `HandleClickModifiers` is declared and exported from the same file, fourteen lines below the option, so the option names it with no import and no dependency edge. The phantom-dependency constraint that forced `event?: any` on the `@object-ui/types` and `plugin-kanban` faces in objectui#9356 does not exist inside `@object-ui/react`; spelling `any` here would throw away the one place in the family where the payload can be named. ⭐ The PR made the right call.

**The accept set — measured with a reviewer probe, both trees, under the lock.** A probe file assigning nine handler shapes to `UseNavigationOverlayOptions['onRowClick']` was compiled once against the head's hook (blob `18ddadad…`) and once against main's (blob `00bb1307…`), same tsconfig (extends the root, `strict`), TypeScript 6.0.3. Readings, head vs main:

- one-argument handler `(r: R) => void` — accepted / accepted (the changeset's headline claim holds)
- `(r: R, e?: any) => void` (the ObjectView and plugin-kanban spelling) — accepted / accepted
- `(r: R, e?: HandleClickModifiers) => void` — accepted / accepted
- a wider payload `(r: R, e?: { metaKey?: boolean }) => void` — accepted / accepted
- storing the member into a one-argument slot — accepted / accepted (the reverse direction holds)
- three optional parameters — accepted / accepted (unchanged; TS admits extra optional source parameters)
- a REQUIRED second parameter `(r: R, ev: HandleClickModifiers) => void` — rejected / rejected (unchanged; on the head it is TS2322 because `undefined` is in the target's parameter type, which is the honest reading of the five one-argument feeders)
- ⚠️ **a NARROWER optional second parameter `(r: R, ev?: React.MouseEvent) => void` — REJECTED on the head / ACCEPTED on main.** Head diagnostic, verbatim apart from the generic spelled in words: `error TS2322: Type '(r: R, ev?: ReactMouseEvent) => void' is not assignable to type '(record: R, event?: HandleClickModifiers | undefined) => void'. … Type 'HandleClickModifiers' is missing the following properties from type 'MouseEvent of Element': altKey, buttons, clientX, clientY, and 26 more.` On main the same line compiled (the probe's `@ts-expect-error` came back as TS2578 unused).

⇒ The widening narrows the accept set by exactly one class: a handler passed **directly** to the hook whose second parameter is typed narrower than `HandleClickModifiers`. Such a handler was written against the implementation rather than the declaration (the old declaration had no second parameter), and it is runtime-sound on the three mouse-event paths, so it is a plausible host handler and it stops compiling. Nothing in-tree is in that class (CI `Type Check` at the head is green and I re-derived the nine feeders' prop types — all one-parameter or `any`), and hosts that reach the hook through a view component's `onRowClick` prop are untouched, because the prop's own declared type is what gets assigned to the option.

### ② Semver grading against the changeset

- Bump: `'@object-ui/react': minor`. Correct and also the ceiling — one `fixed` group of 40 packages (`@object-ui/react` in it), `major` unavailable by repo rule, and AGENTS.md makes a `Clause-②: yes` PR at least `minor`. `Changeset Bump Policy`, `Changeset Fixed Group Check`, `Changeset Declaration` and `Changeset Claim Re-read` are all green at the head.
- Body, paragraph "Not breaking, in either direction": the two sentences it makes are true and pinned (`_NarrowIsAssignableToWide`, `_WideIsAssignableToNarrow` at test lines 108–109). The generalisation it draws — "No caller has to change" — is **falsified by the measured class above**. This text publishes verbatim into `packages/react/CHANGELOG.md`, which is what an upgrading agent greps after hitting exactly that TS2322; a CHANGELOG line that says "not breaking" at the moment the compiler says otherwise is the defect class AGENTS.md's changeset rule exists for, and it cannot be corrected later except by a dedicated docs-only PR.
- Scope paragraph (the component-prop family left to the card's ruling): accurate; I re-derived the 13 hook call sites (14 textual hits — the PR's 13 plus the hook's own docblock example at `:235`), nine feed the option, four do not.

Owed edit, blocking: qualify that paragraph. A sufficient form: state that a handler passed directly to `useNavigationOverlay` whose second parameter is typed narrower than `HandleClickModifiers` (for example React's mouse event) no longer type-checks, and give the one-line fix — type it as `HandleClickModifiers` (exported from `@object-ui/react`) or drop the annotation. Recommended in the same commit: add that reading to the pin's compile-time half as a `@ts-expect-error` line, so the accept-set boundary is a measurement rather than prose (the probe line above lifts straight in).

### ③ Boundary flags and open questions

- **Cross-package channel with objectui#9356 — the two PRs agree.** #9356 spells `ObjectKanbanSchema.onCardClick` and `ObjectKanbanComponentProps.onCardClick` as `(card: any, event?: any) => void`, leaves `ObjectKanbanComponentProps.onRowClick` at `(record: any) => void`, and `ObjectKanban` hands `externalClick = onRowClick ?? onCardClick` to this hook. Handed to `tsc` in the same probe: a value of that union type is assignable to this PR's widened option, and the wrapper's `navigation.handleClick(card, event)` with `event?: any` type-checks against `handleClick` — both accepted on the head. Arity (one or two) and optionality agree; only the payload's spelling differs, and it differs for a stated, package-local reason on each side. File lists are disjoint (this PR's three files vs #9356's six), so landing order is free; #9356's `[[card, undefined]]` readings depend on `handleClick` forwarding, which a deleted type assertion cannot change (erased at emit). One thing #9356's kanban side could still take from this PR: `plugin-kanban` depends on `@object-ui/react`, so `ObjectKanbanComponentProps` *could* name `HandleClickModifiers`; #9356 chose `any` to keep the two kanban faces identical. That belongs to the card's family ruling, not to this PR.
- **`main` drift, re-measured at `b67b53bc0` (16 commits past the base, 88 files).** Shallow clone deepened from 1291 to 1741 commits on `main`; `merge-base` answers the PR's recorded base before and after the deepen (the base was already inside the window, so the control did not differ — the deepen is proven by the count, and `--is-ancestor` exits 0 both ways). The hook file, `packages/react/tsconfig.json`, `tsconfig.test.json` and `package.json` are byte-identical between the base and current `main`; none of the PR's three files moved on `main` (overlap grep fires on a real moved file as control); old-style `merge-tree` prints 0 conflict markers. Of the nine feeders only `ObjectGrid.tsx` moved (#9364, inline-locale faces) and its diff names no `onRowClick`, `useNavigationOverlay` or `handleClick` line. #9369's `SchemaRenderer.tsx` edit and #9371's app-shell edit do not touch what this PR assumes. No landed change invalidates the PR.
- **The two red checks are stale, re-derived from run timestamps.** `Doc Snippet Type Check` and `Skill Example Check` ran on this head at 06:16:19Z–06:20:25Z; on `main` both workflows are green for every completed run from 10:53:58Z (`250429c8`) through 11:29Z (`dab9f96e`), one queue run cancelled in between. Neither is a required context: #9374 merged at 10:12:28Z with `Doc Snippet Type Check` at conclusion `failure` on its head. `Bundle Analysis` was not re-derived here; this PR's Console Performance Budget comment reads PASS. Every other check on the head is green, including `Type Check` (10 minutes, so the relevance gate ran it) and all four test shards.
- **The pin executes where the PR says it does.** `packages/react/package.json` `type-check` is `tsc --noEmit && tsc -p tsconfig.test.json`; `tsconfig.test.json` includes `src/**/*.test.tsx`; the CI `Type Check` job runs that script. My control leg (main's hook + the PR's test file) reproduced the PR's red-first exactly — five errors at 88:31, 91:3, 91:26, 100:3 and 116:43 — and the head leg has zero errors in the pin file.
- Docs: no README, `content/docs` or `skills` file spells the option (control: the regex fires on the hook file); `plugin-gantt/README.md` names the hook once, by name only. No doc drift owed.
- Lock: both probe legs went through `os-verify-lock.sh` on slot `review-9360`. Verdict lines, quoted: `VERDICT command-exit 0 · held the lock 5s · waited 96s (1m36s)` (two-tree probe) and `VERDICT command-exit 0 · held the lock 2s · waited 279s (4m39s)` (diagnostic leg). The shared checkout was not edited; both trees were extracted with `git archive` and verified by blob hash.

open_questions:
1. (blocking, this PR) Qualify the changeset's "No caller has to change" paragraph for the measured class and give the one-line fix; recommended: pin the boundary with a `@ts-expect-error` line.
2. (card objectui#9357, not this PR) Whether `plugin-kanban`'s component face should name `HandleClickModifiers` where it is reachable, or stay `any` to match `@object-ui/types` — the family ruling the card leaves open.

```
Implemented-by: claude/issue-9357-onrowclick-arity   (mode:subagent)
Reviewed-by:    session_01L5xpA5q533BgTTNADibEFt     (domain:spec @ objectui seat)
```

The code change is the right repair: the arity understatement is real, the cast is deleted rather than relocated, the option names the real payload type where it is nameable, and it agrees with objectui#9356 about the channel. The one defect is a published compatibility sentence that the compiler contradicts for a measured class, in text that ships to consumers and is not correctable after release.

**FAIL**

---

os-sam:
## ADOPTED — the **FAIL** at `5653078449` is adopted verbatim. ⛔ Not landing; the repair is small and named below.

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T12:0xZ. ⛔ Adopted as written — no rewriting, no softening. **Tier: ceiling** — 123 strict `"model":"claude-fable-5-1"` hits, no second value; harness-shaped fallback notices **zero**, against those 123 as the firing control.

### ⭐ The code fix is correct. The defect is one sentence that ships to consumers.

The reviewer verified the repair on every axis this seat asked about — and then found the one thing that cannot be corrected after release:

The changeset asserts **"No caller has to change"**, and that text goes verbatim into `packages/react/CHANGELOG.md`. Measured, it is **false for one class**: a handler passed **directly** to `useNavigationOverlay` whose second parameter is annotated *narrower* than `HandleClickModifiers` (e.g. `ev?: React.MouseEvent`) compiles on `main` and is **refused at head with TS2322** — diagnostic quoted in the record.

⚠️ The narrowing is real but the blast radius is nil **inside this tree**: all nine feeders type their prop as single-parameter or `any`, and hosts entering through a component prop are unaffected. ⇒ this is a **prose** defect, not a code one. It FAILs anyway because a published CHANGELOG line that the compiler refutes cannot be fixed in place once released.

### The four questions, as they came back

1. **The understatement is real and the widening is the right repair.** One call site on the head — `useNavigationOverlay.ts:290`, `onRowClick(record, event)` — against a control: `origin/main`'s single-parameter declaration at `:143` plus the cast at `:269` (single-param regex hits 1 on main, 0 at head). Of nine consumers, **5** call `handleClick(record)` with one argument and **3** pass a real React `MouseEvent` ⇒ the payload genuinely can be absent, so `(record, event?: HandleClickModifiers)` is the channel's true union and an **optional** second parameter is right. A required one would be the reverse lie.
2. **The cast is gone**, not merely papered over: `onRowClick as` reads 0 at head against 1 on main (control hits), repository-wide. ⭐ And the pin's bytes half is *necessary* — the PR's own Leg 3 proved `tsc` is completely blind to the cast returning.
3. **Source compatibility is handed to `tsc`, and the reviewer reproduced it independently** — the control leg reproduces the PR's red-first five errors **verbatim** (88:31 / 91:3 / 91:26 / 100:3 / 116:43) and the head leg is clean. Its own added acceptance probe is what found the narrowing class; eight other shapes read identically on both trees.
4. ⭐ **This PR and #9356 AGREE about the channel** — arity (1 | 2) and optionality match; only the second parameter's spelling differs, and that is forced by package boundaries. ⭐ **And naming the real type here is right**: `HandleClickModifiers` is declared and exported **in this same file** — zero import, zero dependency edge. The phantom-dependency constraint that forced `any` on the kanban side exists only on the `@object-ui/types` side; writing `any` here would discard the one place in the family that *can* name the payload.

### Carrier disposition

`needs:contract-review` removed from **this PR and card objectui#9357 in one stroke, seconds apart**. ⛔ Not a clearance — a FAIL ends a review round exactly as a PASS does, and the two-removal signature is what distinguishes either from a strip. Card state and assignee untouched.

### What is owed — small, and named precisely

1. In the changeset, qualify the "Not breaking / No caller has to change" passage: a handler passed **directly** to `useNavigationOverlay` with a second parameter annotated narrower than `HandleClickModifiers` no longer type-checks, with a one-line remedy (annotate it `HandleClickModifiers`, or drop the annotation).
2. ⭐ In the **same commit**, add one `@ts-expect-error` row to the pin's compile half so that boundary is pinned rather than described — the reviewer's probe line can move in as-is.
3. Then a **light** re-review: same head plus one changeset paragraph and one pin row.

⚠️ Out of scope here and belonging to card objectui#9357's family ruling: `plugin-kanban` depends on `@object-ui/react` and *could* name `HandleClickModifiers` on its component face; #9356 chose `any` to keep its two faces consistent.

---

PR #9369, opened 2026-09-13
feat(react)!: unbind the data-source adapter from the expression scope, and point `bind` at the scope channel

### LINKED ISSUE #9308: finding(react): the renderer seam's `dataSource` carries a second, undeclared meaning — the data scope `bind` and `data.*` resolve against — and typing the key prices it out for a conformant host
Measured while implementing objectui#7912 (the ruled narrowing of the renderer seam's
`dataSource`). Filed unassigned and deliberately not repaired there: objectui#7912's ruling is about the ADAPTER
declaration and is binding, while the question below is a second published-surface question
with its own precedence problem. Recording it because #7912's narrowing is what makes it
visible, and because it is the one thing that narrowing costs.

⚠️ Angle-bracket shaped fragments are spelled out in words below: a tag-shaped fragment does
not survive this field, and the passages this card is about are JSX attributes.

Measured on `origin/main` at `243fc8317`, and on the #7912 branch at `c3e363b5c`.

## The site — one key, two meanings

`SchemaRendererContext.dataSource` is read for two unrelated purposes:

1. **The adapter.** Plugins call `find` / `getObjectSchema` / `aggregate` on it. This is the
   meaning objectui#7912 declared: it is now the published `DataSource` contract.
2. **The data scope.** The same value is the root that `bind` paths and `data.*` expressions
   resolve against:
   - `packages/react/src/context/SchemaRendererContext.tsx` — `useDataScope(path)` splits the
     path and walks it over `context.dataSource`.
   - `packages/react/src/SchemaRenderer.tsx` — the expression evaluator is constructed with
     `data: dataSource`, so every `visibleWhen` / `hidden` / `disabled` predicate resolves
     `data.*` against that same value.
   - `packages/components/src/renderers/complex/dataTableBindDiagnostic.ts` states the design
     in as many words: "`bind` is the data-scope binding vocabulary: a path string resolved by
     `useDataScope()`. `list`, `tree-view` and the `object-*` plugin widgets read it."

Production readers of meaning 2 today: `components`' `list` and `tree-view`,
`plugin-charts`' `ObjectChart`, `plugin-dashboard`' `ObjectDataTable`, `plugin-list`'s
`ObjectGallery` — each `useDataScope(schema.bind)`.

## Why this is now a contract question and not just an oddity

A `DataSource` adapter has no `users`, no `customers`, no `status` member. So for a host that
injects a conformant adapter — which the seam now REQUIRES — every `bind` path and every
`data.*` predicate resolves to `undefined`. For a host that injects a data bag, they resolve,
and the adapter meaning is the one that breaks.

The two meanings were never compatible; `any` merely hid the choice. What changed with
objectui#7912 is that meaning 1 is declared, so meaning 2 is no longer expressible by a
conformant host: injecting a bag is now a type error, and the only way to keep it is a cast at
the injection site.

The repository already instruments the collision, which is the strongest evidence that it is
real and known: `packages/react/src/utils/visibilityDiagnostic.ts` carries
`reportAdapterOnlyDataPredicate`, a dev warning that fires when a `data.*` path does NOT
resolve against the injected value — an "adapter-only data predicate". That warning exists
because both meanings arrive through one key.

## Measured population

On the objectui#7912 branch, every internal injection that relies on meaning 2 (or on a
partial double) is now marked with an explicit cast, so the population is greppable rather
than inferred:

- **65** marked injection sites across **33** test files in 8 packages
  (`grep -rn 'as unknown as DataSource'`).
- **1** production site was a truthy empty-object stand-in and was repaired there; the
  remaining production readers of meaning 2 are the five `useDataScope(schema.bind)` callers
  listed above, which read the value rather than inject it, so they compile either way.

## Options, none taken here

- **A.** Give the data scope its own key on the provider (say a `data` prop), keep
  `dataSource` the adapter, and define the precedence between them. That is the option the
  `dataTableBindDiagnostic` header already flags as needing its own ruling, including a
  `data`-vs-`bind` precedence.
- **B.** Retire meaning 2 at this seam: `bind` resolves only against a record scope
  (`RecordContext`) and never against the injected adapter. Decides objectui#7206's sibling
  question in the same direction.
- **C.** Declare the union on the key (`DataSource` or a record bag). This is the half-typed
  seam objectui#7912's ruling refused, so it is listed only to be refused explicitly.

⛔ No recommendation is implied by the order. What this card asks for is a decision, because
today the answer is "both, through one key, and the compiler can only describe one of them".

Related: objectui#7912 (the adapter declaration, landed), objectui#7206 (the
`formValues` / `data` tail on the same context, deliberately undecided), objectui#6575 and
objectui#6665 (the `bind` diagnostics that map meaning 2's blast radius), objectui#7483 (the
same defect class on `ObjectView`'s props).

Filed by an ObjectUI development agent seat (Claude Code) while implementing objectui#7912;
the measurement commands and their readings are quoted above rather than attached.

---

### LINKED ISSUE #9346: finding(docs): objectui#8021's two repaired teaching surfaces inject a plain data bag as the provider's dataSource — three fences that stopped compiling under the objectui#7912 narrowing
Measured on PR objectui#9310 (card objectui#7912, decision batch #68) at head `e2c31aa1d`, reproduced locally against the gate's own built closure. Filed unassigned by the `os-dev` seat that repaired the `Type Check` red on that PR. ⛔ Not claimed, not edited there — the repair needs objectui#9308's answer first, for the reason measured below.

⚠️ Angle-bracket shaped fragments are spelled out in words throughout: a tag-shaped fragment does not survive this field, and the subject here is a JSX attribute and a generic parameter. `DataSource OF any` below is written the way tsc printed it.

## The reading

`pnpm check:doc-snippets` — the `Doc Snippet Type Check` context — on PR objectui#9310:

```
[semantic]  content/docs/guide/schema-rendering.md:99:29
            TS2740: Type '{ user: { name: string; role: string; }; stats: { totalUsers: number; }; }'
            is missing the following properties from type 'DataSource OF any':
            find, findOne, create, update, and 2 more.
[semantic]  content/docs/guide/schema-rendering.md:434:29   TS2740, same shape
[semantic]  packages/react/README.md:73:29                  TS2740, same shape
Semantic phase: 649 of 649 block(s) judged, 3 failed.
```

Every control in that same run fired: the resolution control resolved `@object-ui/types` to the built `dist/index.d.ts`, the sentinel import produced TS2305, the positive import produced 0 diagnostics. The three findings are readings, not a harness that degraded the types.

## Where the three blocks came from

They are objectui#8021's repair, landed on `main` and then merged into the objectui#7912 branch. Before that merge the branch's own run was 646/646 blocks judged, 0 failed; after it, 649/649 judged, 3 failed. The blocks are new; the type they now fail against is objectui#7912's.

All three teach the same wiring: put the page's own values on the provider's `dataSource`, and read them back through the `data.` prefix. That is the SECOND meaning of this one key — the meaning objectui#9308 was filed to get a decision on. The ruled type declares only the first one, the host's adapter.

## Why this is not a one-line repair

Three routes to green, each refused by something already landed:

1. **Cast inside the fence.** Refused: these are published teaching surfaces, and objectui#7912's execution note 2 is that a cast at the seam re-creates the defect one line lower. A guide that shows the reader how to defeat the contract the same PR establishes is worse than a red context.
2. **Widen the declaration back.** Refused: the narrowing is the whole card (decision batch #68, maintainer verbatim 「同意」).
3. **Declare the blocks in the gate's ledger.** Refused: that gate's header says a block is either repaired or declared with a written reason, and the only reason available here is "the page teaches an off-contract wiring" — a quarantine, not a reason.

What is left is rewriting what the pages teach, and that IS objectui#9308's decision rather than a documentation edit:

- A conformant adapter has no `user`, no `stats` and no `settings` member, so under the ruled type every `data.*` spelling on these pages evaluates to `undefined`. The repo already instruments that exact collision: `reportAdapterOnlyDataPredicate` in `packages/react/src/utils/visibilityDiagnostic.ts` warns in dev when a `data.*` path does not resolve against the injected value.
- And `content/docs/guide/schema-rendering.md` is PINNED to the wiring it teaches today. `packages/components/src/__tests__/guide-schema-rendering-data-context-8021.test.tsx` leg A reads the `## Data Context` fence off the page, branches on whether that fence carries a provider `dataSource=` attribute, then renders the `### Accessing Data in Schemas` JSON fence and asserts it prints `Welcome, John!`. Take the provider attribute away and leg A takes the other branch and reads `Welcome, !`; move the expression off the `data.` prefix and it prints the raw source text. So the page cannot leave meaning 2 without moving a landed card's pin at the same time.

## What this needs

objectui#9308's answer — give the data scope its own key on the provider, retire meaning 2 at this seam, or something else — and then one documentation pass that moves the three fences and objectui#8021's pin together.

## Blast radius while this is open

`Doc Snippet Type Check` is **not** in `main`'s required set. Re-measured on `GET /repos/objectstack-ai/objectui/rules/branches/main` at 2026-09-13: the required contexts are `Lint`, `Type Check`, `Build & E2E`, `Test (shard 1/4)` through `Test (shard 4/4)`, `Build Docs` and `Changeset Declaration`. So this blocks nothing in the merge queue; it does leave a red context on PR objectui#9310 and on every pull request opened after it lands.

Related: objectui#9308 (the decision this waits on), objectui#8021 (landed; it wrote these three fences), objectui#9297 (the three further surfaces objectui#8021 deliberately left), objectui#9311 (the skills-lane sibling of this same red), objectui#7912 and PR objectui#9310 (the narrowing itself).

Reported by an agent seat; attribution is recorded in this paragraph rather than in a footer block.

os-tesla:
Fixes #9308
Fixes #9346

Maintainer ruling 2026-09-13, verbatim 「同意」 on **option B**, both halves. Option A (a new published provider prop) and option C were not taken.

## What landed

**b1 — `data: dataSource` leaves the evaluator scope.** `packages/react/src/SchemaRenderer.tsx` no longer publishes the injected adapter as an expression root. `@object-ui/app-shell`'s `ExpressionProvider` states the governing principle — every root a tier binds must be one the engine accepts AND one that tier can actually answer — and objectui#8155 (`app`) / objectui#8166 (`data`) already applied it one tier up. Against a conformant `DataSource` adapter every `data.*` path resolved `undefined`, so this tier bound a root it could not answer.

**b2 — `useDataScope` reads `usePredicateScope()`.** A node's `bind` now resolves a path in the ambient scope a host publishes through `PredicateScopeProvider` — the channel `ExpressionProvider` already feeds — instead of walking the adapter. No published key is added; that is what makes B strictly better than A.

**Side effect the ruling's own appendix named:** `...predicateScope` was spread *before* `data: dataSource`, so a host that legitimately published `data` through the documented scope channel was silently overwritten by the adapter. Removing the line un-shadows it. Pinned as leg 3 of the new test.

### 1. The verdict DOES move — the dispatch's premise here was falsified

The dispatch predicted the verdict would not move "because `data.*` already resolved `undefined` against an adapter". Measured on the built evaluator, that is wrong, and the distinction is **present-but-undefined key vs absent key**:

| scope | `data.status == 'draft'` | with `throwOnError` |
|---|---|---|
| `{ data: undefined }` | `false` | `false` |
| no `data` key at all | `true` | throws ``data is not defined`` |

So a page/component `visible` gate reading `data.*` was a clean constant `false` (node hidden on every row) and is now unevaluable, which this fail-soft surface answers `true` (node **shown** on every row). It stops being silent in the same stroke: objectui#5454's `reportUnresolvableVisibilityPredicate` names it, in production as well as development, with a message that is true for it. That is the breaking half of this change, it is stated plainly in the changeset, and it is pinned as leg 7 of `SchemaRenderer.dataRootUnbound-9308.test.tsx` rather than left to be discovered in a console.

### 2. `reportAdapterOnlyDataPredicate` — measured, then re-aimed

It does **not** flood: its trigger reads the object passed as `boundData`, not the evaluator scope, so b1 changed its firing set for a conformant host by exactly nothing (every `data.*` path was already undefined on an adapter). What b1 broke is its **truth**, in both directions at once:

* left pointing at the adapter it would have gone **silent** for precisely the hosts this change breaks — a bag injected as `dataSource` still answers the read, while the evaluator no longer sees it;
* and **loud** for hosts whose scope-channel `data` answers perfectly.

⇒ Decision: keep the diagnostic, keep its severity and its dev-only gate, and hand it the `data` the evaluator actually bound (`predicateScope?.data`). Its copy moves off the "at the node tier `data` is the DATA-SOURCE ADAPTER" claim, which this PR makes false. Reachability is now stated in the source rather than assumed: with no host `data`, a `data.*` predicate throws and objectui#5454 takes it, so this leg speaks only for a host that published a `data` root which does not answer — the population it can still speak for.

### 3. `celAuthoring`'s `rowCanonAdvisory` — more correct, and untouched

Its message says a wrong-layer `data.*` "faults at runtime with `Unknown variable: data` and the rule never fires". Before this PR the node tier contradicted that by resolving `data` to the adapter; now it agrees. ⛔ Severity unchanged (`warning`) — every save gate on that tier counts `severity === 'error'`, and promoting it would refuse predicates already in customer metadata.

### 4. `dataTableBindDiagnostic`'s `data`-vs-`bind` precedence question — DISSOLVED

Confirmed, and said so in the module's own docblock. While `useDataScope` walked the adapter, `bind` and the node's `data` were two spellings of one idea and a reader of both would have needed a precedence rule. Now `bind` resolves the ambient scope and `data` is an inline array on the node: different channels, no precedence to decide. ⛔ `data-table` still does not read `bind` (objectui#6575) and does not start.

## Blast radius — all nine readers re-verified per reader, not inherited

Each already falls back, so the behaviour against a conformant host is unchanged (the fallback is what was running):

| reader | resolution |
|---|---|
| `components/.../list.tsx` | `Array.isArray(boundData) ? boundData : Array.isArray(schema.items) ? schema.items : []` |
| `components/.../tree-view.tsx` | `boundData \|\| schema.nodes \|\| []` |
| `plugin-charts/ObjectChart.tsx` | `boundData \|\| schema.data \|\| fetchedData` |
| `plugin-dashboard/ObjectDataTable.tsx` | `boundData \|\| schema.data \|\| fetchedData` |
| `plugin-dashboard/ObjectPivotTable.tsx` | `boundData \|\| schema.data \|\| fetchedData` |
| `plugin-grid/ObjectGrid.tsx` | `if (boundData && Array.isArray(boundData))` → else falls through to `getDataConfig(schema)` |
| `plugin-kanban/ObjectKanban.tsx` | `external \|\| boundData \|\| schema.data \|\| fetchedData` |
| `plugin-list/ObjectGallery.tsx` | `props.data \|\| boundData \|\| schema.data \|\| fetchedData \|\| []` |
| `plugin-timeline/ObjectTimeline.tsx` | `props.data \|\| boundData \|\| fetchedData` |

**None treats `boundData` as its sole source**, so the shape of b2 is unchanged. The card body's "five components" is wrong; nine is right.

## The documentation half

`Doc Snippet Type Check` was red on `main` with three fences, all teaching meaning 2 — publish page values through `dataSource`, read them back under `data.*`:

```
content/docs/guide/schema-rendering.md:99:29    TS2740
content/docs/guide/schema-rendering.md:434:29   TS2740
packages/react/README.md:73:29                  TS2740
```

All three move to `PredicateScopeProvider`, and so does the prose around them — including `packages/react/README.md`'s sentence that ``dataSource`` "is what the evaluator sees under the name `data`", which the gate cannot see and which b1 deletes outright. `packages/types/src/base.ts`'s `bind` example comment (`// → dataSource.customerNames`) moves with it.

## The pin, written first and run red on the unmodified tree

`packages/react/src/__tests__/SchemaRenderer.dataRootUnbound-9308.test.tsx`, red-first reading verbatim:

```
 ❯ |dom| packages/react/src/__tests__/SchemaRenderer.dataRootUnbound-9308.test.tsx (7 tests | 5 failed)
     × leg 1: `${data.stats.total}` against an injected bag does NOT resolve 32ms
     × leg 3: a host-published `data` is no longer shadowed by the adapter 5ms
     × leg 4: a path resolves against the ambient scope 3ms
     × leg 5: the same path against the ADAPTER seam resolves to nothing 2ms
     × leg 7: a `data.*` visibility gate becomes unresolvable — shown, and LOUD 4ms

AssertionError: expected '99' to be '${data.stats.total}' // Object.is equality
      Tests  5 failed | 2 passed (7)
```

Legs 2 and 6 are the lit controls and pass on the unmodified tree, which is what makes the other five a reading. After b1 + b2: `Tests 7 passed (7)`.

**objectui#8021's pin was re-derived, never weakened or skipped.** Leg A still READS the Data Context fence off the page on every run (it now branches over three carriers instead of two, so moving the fence back to either of the others changes what it renders); leg C's wiring and leg D's expression swapped to the new teaching; and a new **leg E** pins the adapter seam directly — provider `dataSource` plus `${data.user.name}` now renders the characters the author typed. The `FORWARDED_LOOKALIKES` sweep gained `scope=` so the new carrier's own look-alike is covered.

Two stale claims were corrected while re-deriving neighbouring pins: three test files carried ``/** The ambient scope app-shell's `ExpressionProvider` really mounts. */`` over a bag containing `data: {}`. `buildExpressionScope` has published no `data` since objectui#8166, so the comment was false before this PR and would have hidden the verdict move after it. Those bags are kept — they are now the only way the adapter-only leg is reachable at all — and re-documented as deliberate HOST publications.

## Verification

See the report comment on #9308 for the full command list and `VERDICT` lines.

## Acceptance notes

* **`skills/objectui/**` now teaches a false wiring** — `guides/schema-expressions.md` (``| `data` | Alias for dataSource root |``, "`dataSource = { customerNames: [...] }`, `list` calls `useDataScope("customerNames")`", "`"bind": "app.settings.users"` resolves `dataSource.app.settings.users`"), `guides/data-integration.md` and `rules/protocol.md`. ⛔ Not touched here: `skills/**` is a governed surface, and one path would park this whole PR in the draft-and-wait lane while it is the PR unblocking a red `main`. Filed separately; the six affected pins in `packages/components` publish their host values through the scope channel, which keeps them measuring their own subject.
* **objectui#7206** (the `formValues` / `data` tail) is pointed the same way by this ruling but is ⛔ not ruled and ⛔ not addressed here; it remains open.
* **objectui#6575** — `data-table` not reading `bind` — is untouched.

## 维护者速读(草稿)

**改了什么。** 渲染器不再把宿主注入的 `dataSource`(适配器)当作表达式根 `data` 发布;节点的 `bind` 改为在宿主通过 `PredicateScopeProvider` 发布的环境作用域里解析。文档与 README 随之改教作用域通道。

**为什么改。** 这是仓库自己已裁过两次的原则(objectui#8155 / objectui#8166)贯彻到最后一层:一个层只能绑定它答得出的根。面对合规适配器,`data.*` 恒为 `undefined`,这一层答不出。

**风险与代价(含回滚)。** 已在客户元数据里的 `data.*` 可见性判据,行为从「恒隐藏」变为「恒显示」,并在控制台响亮报出。这是唯一的行为断点,已在 changeset 正文点名。回滚只需还原 `SchemaRenderer.tsx` 的一行与 `useDataScope` 的一处读取,其余为文档与钉子。

**席位意见。**

**你要做的。** 确认「`data.*` 判据由恒隐藏翻为恒显示」这一取舍可接受;若希望改为保守方向(保留 `data` 为一个恒空对象以维持旧裁决),请说,那是另一张卡。

---

### DIFF
--- a/.changeset/9308-data-root-unbound-from-adapter.md
+++ b/.changeset/9308-data-root-unbound-from-adapter.md
@@ -1,0 +1,64 @@
+---
+'@object-ui/react': minor
+'@object-ui/components': minor
+'@object-ui/types': minor
+---
+
+Unbind the data-source adapter from the expression scope, and point `bind` at the scope
+channel (objectui#9308, maintainer ruling 2026-09-13, option B).
+
+**Breaking, deliberately — and `minor` only because this repo's `fixed` group of 40
+packages may not carry a `major` (AGENTS.md 版本号策略).** Read the migration note below
+before upgrading if any of your metadata reads `data.*` at the page/component tier.
+
+**What changed.**
+
+1. `SchemaRenderer` no longer writes `data: dataSource` into the evaluator scope. `data`
+   is not a root this tier binds. The roots it supplies are `record` (the row, when a
+   record surface bound one), `page` (page-local variables) and `current_user` — plus
+   every name the host published through `PredicateScopeProvider`.
+2. `useDataScope(path)` — what a node's `bind` resolves through — reads the ambient
+   predicate scope (`usePredicateScope()`) instead of walking the injected adapter.
+
+**Who this breaks, concretely.**
+
+* **A `${data.…}` expression on a page/component node.** It used to resolve against the
+  host's injected `DataSource` adapter. An adapter answers no `data.*` path, so for every
+  host that injected a real adapter the read was already `undefined` and the predicate was
+  a silent constant — but the constant MOVES: `data` used to be a present key holding an
+  object, so `data.status == 'draft'` evaluated cleanly to `false` and the node was hidden
+  on every row. `data` is now ABSENT, the expression cannot be evaluated at all, and this
+  surface is fail-soft, so the same gate now answers `true` and the node is SHOWN on every
+  row. It is no longer silent: the objectui#5454 reporter names it in the console, in
+  production as well as development. Re-root such a predicate on `record.*` — the row —
+  or publish a `data` of your own through `PredicateScopeProvider`.
+* **A host that injected a plain data bag as `dataSource`.** That was meaning 2 of the
+  key, and it stops working entirely: `${data.…}` no longer reads it and `bind` no longer
+  walks it. It has been a compile error since objectui#7912 (`dataSource` is typed
+  `DataSource | null | undefined`). Publish those values through `PredicateScopeProvider`
+  instead — the provider, the hook and the app-shell wiring already exist, and this change
+  adds no new published key.
+* **The nine `bind` readers** (`list`, `tree-view`, `ObjectChart`, `ObjectDataTable`,
+  `ObjectPivotTable`, `ObjectGrid`, `ObjectKanban`, `ObjectGallery`, `ObjectTimeline`) now
+  resolve `bind` against the scope. All nine already carry a fallback
+  (`boundData || schema.items`, `|| schema.nodes || []`, or a fall-through to their own
+  fetch chain), and against a conformant adapter `useDataScope` returned `undefined`
+  before this change — so the fallback is what was running, and a `bind` that resolved
+  nothing before still resolves nothing. What is new is that a `bind` CAN now resolve:
+  a host that publishes rows on its scope will see them used where the fetch used to run.
+
+**Why.** `@object-ui/app-shell`'s `ExpressionProvider` states the rule this applies: every
+root bound at a tier must be one the engine accepts AND one that tier can actually answer.
+objectui#8155 unbound `app` under it and objectui#8166 unbound `data`; ADR-0089 D3 puts
+`data` at the metadata layer and `record` at the runtime layer, and the engine's
+per-surface `FIELD_RULE_BOUND_ROOTS` is `['record','previous','parent']`. The renderer tier
+was the last one still binding a root it could not answer.
+
+Two side effects worth naming. A host that legitimately published `data` through the
+documented scope channel used to be silently overwritten by the adapter (the adapter was
+spread last); it is not any more. And `reportAdapterOnlyDataPredicate` now resolves against
+the `data` the evaluator actually bound rather than against the adapter — left pointing at
+the adapter it would have gone silent for precisely the hosts this change breaks.
+
+The Data Context passages of `content/docs/guide/schema-rendering.md` and
+`packages/react/README.md` teach the scope channel accordingly.
--- a/content/docs/guide/schema-rendering.md
+++ b/content/docs/guide/schema-rendering.md
@@ -112,7 +112,7 @@ The scope the evaluator builds is what you published, plus three names the rende
 
 | name | what it holds |
 |---|---|
-| every key of `scope` | exactly what you put there — `stats`, `current_user`, whatever the page needs |
+| every key of `scope` | exactly what you put there — `user`, `stats`, whatever the page needs |
 | `page` | page-local variables, for predicates that gate on another component's state |
 | `record` | the row a record surface is bound to, when there is one |
 | `current_user` | an alias of whatever you published as `user`; the host's `ExpressionProvider` publishes the signed-in user there |
--- a/content/docs/guide/schema-rendering.md
+++ b/content/docs/guide/schema-rendering.md
@@ -68,7 +68,7 @@ interface BaseSchema {
   "visibleOn": "${user.role === 'admin'}",
   "body": {
     "type": "text",
-    "content": "Total Users: ${data.stats.totalUsers}"
+    "content": "Total Users: ${stats.totalUsers}"
   }
 }
 ```
@@ -79,52 +79,63 @@ Expression context does not arrive as a prop. `SchemaRenderer` declares exactly 
 `schema`, and every other prop it is handed is forwarded to the component the schema names —
 so a `data` prop written on the element reaches the evaluator through nothing. Because it is
 forwarded rather than refused, nothing throws and nothing warns; the expression simply never
-resolves. The scope comes from `SchemaRendererProvider`, which publishes its `dataSource`
-under the name `data`:
+resolves. The scope comes from `PredicateScopeProvider`, which publishes each name you give
+it as an expression root:
 
 ```tsx
-import { SchemaRenderer, SchemaRendererProvider } from '@object-ui/react'
+import { SchemaRenderer, PredicateScopeProvider } from '@object-ui/react'
 import type { BaseSchema } from '@object-ui/types'
 
 // The page schema from the first example on this page.
 declare const schema: BaseSchema
 
-const dataSource = {
+// Every name here becomes a root the schema's expressions can read.
+const scope = {
   user: { name: 'John', role: 'admin' },
   stats: { totalUsers: 1234 },
 }
 
 function App() {
   return (
-    <SchemaRendererProvider dataSource={dataSource}>
+    <PredicateScopeProvider scope={scope}>
       <SchemaRenderer schema={schema} />
-    </SchemaRendererProvider>
+    </PredicateScopeProvider>
   )
 }
 ```
 
-The scope the evaluator builds holds four names, and nothing else:
+An app built on `@object-ui/app-shell` does not mount this provider itself: the shell's
+`ExpressionProvider` already feeds the same channel with `user` (the signed-in user, also
+readable as `current_user`) and `features`.
+
+The scope the evaluator builds is what you published, plus three names the renderer supplies:
 
 | name | what it holds |
 |---|---|
-| `data` | the `dataSource` the provider above published — everything you passed in |
+| every key of `scope` | exactly what you put there — `stats`, `current_user`, whatever the page needs |
 | `page` | page-local variables, for predicates that gate on another component's state |
 | `record` | the row a record surface is bound to, when there is one |
-| `current_user` (aliased to `user`) | the signed-in user, published by the host's `ExpressionProvider` — not by anything on this page |
+| `current_user` | an alias of whatever you published as `user`; the host's `ExpressionProvider` publishes the signed-in user there |
 
 A name outside that set resolves to nothing, and an unresolvable template is not an error:
 the evaluator hands back its own source text, so the characters you typed are what the reader
 sees.
 
+> **`dataSource` is not an expression root.** `SchemaRendererProvider`'s `dataSource` carries
+> the host's `DataSource` *adapter* — the object renderers call `find()` on. The renderer used
+> to publish that adapter under the name `data`; an adapter answers no `data.*` path, so the
+> root was constant for every conformant host, and objectui#9308 removed it. A `${data.…}`
+> expression now reads whatever *you* published under `data`, and nothing if you published
+> none. At the runtime layer the row is `record` (ADR-0089).
+
 ### Accessing Data in Schemas
 
-Use expression syntax `${}` to reference the scope, and reach your own values through the
-`data.` prefix:
+Use expression syntax `${}` to reference the scope, by the name you published it under:
 
 ```json
 {
   "type": "text",
-  "content": "Welcome, ${data.user.name}!"
+  "content": "Welcome, ${user.name}!"
 }
 ```
 
@@ -409,11 +420,11 @@ const pageSchema = {
 
 ### 2. Use Data Context Effectively
 
-Put everything the schema's expressions need on one `dataSource`, mounted above the tree —
-not on the renderer, which does not read it:
+Put everything the schema's expressions need on one scope, mounted above the tree — not on
+the renderer, which does not read it:
 
 ```tsx
-import { SchemaRenderer, SchemaRendererProvider } from '@object-ui/react'
+import { SchemaRenderer, PredicateScopeProvider } from '@object-ui/react'
 import type { BaseSchema } from '@object-ui/types'
 
 // The reader's own values.
@@ -422,8 +433,8 @@ declare const userData: { name: string }
 declare const userSettings: { theme: string }
 declare const dashboardStats: { totalUsers: number }
 
-// ✅ Good — one provider, and every expression reaches it through `data.`
-const dataSource = {
+// ✅ Good — one provider, and every expression reads a name published on it
+const scope = {
   user: userData,
   settings: userSettings,
   stats: dashboardStats,
@@ -431,9 +442,9 @@ const dataSource = {
 
 function Dashboard() {
   return (
-    <SchemaRendererProvider dataSource={dataSource}>
+    <PredicateScopeProvider scope={scope}>
       <SchemaRenderer schema={schema} />
-    </SchemaRendererProvider>
+    </PredicateScopeProvider>
   )
 }
 ```
--- a/packages/components/src/__tests__/bindable-text-keys-4795.test.tsx
+++ b/packages/components/src/__tests__/bindable-text-keys-4795.test.tsx
@@ -33,7 +33,7 @@
 
 import { describe, it, expect } from 'vitest';
 import { render, screen } from '@testing-library/react';
-import { SchemaRenderer, SchemaRendererProvider } from '@object-ui/react';
+import { SchemaRenderer, SchemaRendererProvider, PredicateScopeProvider } from '@object-ui/react';
 // Module scope, not a hook — the cold transform would otherwise be billed to
 // `hookTimeout` (object-ui/no-dynamic-import-in-test-hook, objectui#3010).
 import '../renderers';
@@ -55,9 +55,11 @@ const DATA = { total: 99, caption: 'Active users', note: '+20.1% from last month
 
 const renderNode = (schema: any) =>
   render(
-    <SchemaRendererProvider dataSource={DATA as unknown as DataSource}>
+    <PredicateScopeProvider scope={{ data: DATA }}>
+      <SchemaRendererProvider dataSource={DATA as unknown as DataSource}>
       <SchemaRenderer schema={schema} />
-    </SchemaRendererProvider>,
+    </SchemaRendererProvider>
+    </PredicateScopeProvider>,
   );
 
 describe('objectui#4795 — declared text keys are evaluated AND read back, through real renderers', () => {
--- a/packages/components/src/__tests__/data-table-node-data-diagnostic.test.tsx
+++ b/packages/components/src/__tests__/data-table-node-data-diagnostic.test.tsx
@@ -42,7 +42,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen } from '@testing-library/react';
 import React from 'react';
-import { SchemaRenderer, SchemaRendererProvider } from '@object-ui/react';
+import { SchemaRenderer, SchemaRendererProvider, PredicateScopeProvider } from '@object-ui/react';
 
 // The REAL renderers, imported at module scope so `data-table` is in the
 // registry before the first render (AGENTS.md §测试纪律 — never behind a lazy
@@ -114,9 +114,11 @@ function warningsOn(prefix: string): string[] {
 
 function tree(schema: unknown) {
   return (
-    <SchemaRendererProvider dataSource={SCOPE as unknown as DataSource}>
+    <PredicateScopeProvider scope={{ data: SCOPE }}>
+      <SchemaRendererProvider dataSource={SCOPE as unknown as DataSource}>
       <SchemaRenderer schema={schema as never} />
     </SchemaRendererProvider>
+    </PredicateScopeProvider>
   );
 }
 
--- a/packages/components/src/__tests__/disabled-verdict-one-carrier.test.tsx
+++ b/packages/components/src/__tests__/disabled-verdict-one-carrier.test.tsx
@@ -30,7 +30,7 @@
 import { describe, it, expect } from 'vitest';
 import { render, screen } from '@testing-library/react';
 import React from 'react';
-import { SchemaRenderer, SchemaRendererContext } from '@object-ui/react';
+import { SchemaRenderer, SchemaRendererContext, PredicateScopeProvider } from '@object-ui/react';
 // Registers the renderers at module scope, NOT inside a `beforeAll` — there the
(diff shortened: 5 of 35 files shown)


os-tesla:
## Contract review

Head judged: `4de2534f25e67873d88f16ac5bfe8864395e4919` (base `69aa9c017`).

## ⛔ First: this seat falsified TWO premises I wrote into its dispatch

**1. `reportAdapterOnlyDataPredicate` would flood.** I told it to measure whether removing the `data` root makes every `data.*` predicate unresolved and floods the console. **False, and for a reason I had not looked at**: that diagnostic resolves against the argument passed as `boundData`, not against the evaluator scope. b1 changed its firing set for a conformant host by **zero**. What b1 broke is not its volume but its **truth**, in both directions — pointed at the adapter it would go SILENT for exactly the hosts this change breaks and LOUD for hosts whose scope-channel `data` answers. The seat re-aimed it at `predicateScope?.data` and rewrote the copy off the now-false "data is the adapter" claim.

**2. "Confirm the verdict does not move for a conformant host — it should not."** ⭐ **Falsified, and this is the load-bearing measurement of the whole round:**

```
{ data: undefined }      data.status == 'draft'  ->  false
no `data` key at all     data.status == 'draft'  ->  true   (and THROWS under throwOnError)
```

**A present-but-undefined root and an absent root are not the same to the engine.** So a `data.*` visibility gate flips from **constant-HIDE to constant-SHOW**. I predicted no movement because I reasoned "it resolved `undefined` before, it resolves nothing now" — which conflates exactly those two states. Pinned as leg 7, stated in the changeset and the PR body.

### ① Derived judgments

| # | move | judged |
|---|---|---|
| 1 | `data: dataSource` removed from the evaluator scope | **correct — verified by content, not by count** |
| 2 | `useDataScope` reads `usePredicateScope()` | **correct** |
| 3 | the shadowing defect retired as a side effect | **correct, and it was a real latent bug** |
| 4 | nine `bind` readers, per-reader fallback re-verified | **correct** |
| 5 | objectui#8021's pin re-derived, not weakened | **correct** |
| 6 | the three fences + the README sentence the gate cannot see | **correct** |

**1 — enumerated rather than counted.** `grep -c "data: dataSource"` returns **2** on this branch, which looks like a failed removal. Both hits are **comments** explaining the removal (`:835`, `:848`). The live scope reads:

```js
new ExpressionEvaluator({
  ...predicateScope,
  current_user: (predicateScope as any)?.user,
  ...(boundRecord && … ? { record: boundRecord } : null),
  page: pageVariables,
});
```

No `data`. ⭐ Recording the method because the count alone would have been a false alarm, and I have made that mistake three times this session.

**3** — I raised the shadowing hazard independently during the analysis: `...predicateScope` at `:842` followed by `data: dataSource` at `:847` meant a host publishing `data` through the **documented** scope channel had it silently overwritten by the adapter. b1 retires it, and `:848`'s comment names it rather than fixing it in silence.

**4 — the reason b2 is safe, re-read per reader on this branch:** `list.tsx` `Array.isArray(boundData) ? boundData : …schema.items`; `tree-view` `boundData || schema.nodes || []`; `ObjectChart` / `ObjectDataTable` / `ObjectPivotTable` `boundData || schema.data || fetchedData`; `ObjectGrid` falls through to `getDataConfig(schema)`; `ObjectKanban` `external || boundData || schema.data || fetchedData`; `ObjectGallery` `props.data || boundData || …`; `ObjectTimeline` `props.data || boundData || fetchedData`. **None treats `boundData` as its sole source.** (The card said five readers; there are **nine** — my dispatch carried that correction and the seat re-verified all nine rather than inheriting it.)

**6 — the deliverable, and CI is the witness, not the seat.** `check:doc-snippets` exit 0 with `Semantic phase: 649 of 649 block(s) judged, 0 failed`, and the ablation reproduces the card's exact three diagnostics by taking only the two documents back to base. Confirmed independently: **PR CI shows `Doc Snippet Type Check` GREEN on this head**, and the only red is `Bundle Analysis`, which is inherited. All three fences moved to `PredicateScopeProvider scope={…}`, and `packages/react/README.md:44`'s ``\`dataSource\` is what the evaluator sees under the name \`data\`.`` — which no gate could see — is deleted.

### ② Semver

`minor` across `@object-ui/react`, `@object-ui/components`, `@object-ui/types`, and the note leads with the break and who it breaks:

> **Breaking, deliberately — and `minor` only because this repo's `fixed` group of 40 packages may not carry a `major`.** Read the migration note below before upgrading if any of your metadata reads `data.*` at the page/component tier.

Correct, and it carries the verdict-flip. ⛔ Note the changeset was **not** edited by me for a spelling convention — there is none, as I had to retract to four seats this round.

### ③ Boundary flags — and the two questions, ruled

**Q2 — the verdict flip. RULING: A, as ruled. Proceed.**

The conservative variant (`data: EMPTY` instead of removing the key) keeps verdicts byte-identical and nothing goes loud. ⛔ Refused, and the seat's reason is the right one: it **keeps publishing a root this tier cannot answer** — the exact defect objectui#8155, objectui#8166 and this ruling each removed — and trades a loud failure for a silent one.

⚠️ **But the maintainer ruled B without this consequence in hand, so it is stated rather than buried.** For any host that authored a `data.*` gate, an element that was constantly HIDDEN becomes constantly SHOWN. The gate was never actually gating on data — it was a constant hide either way — but what a user sees changes. It is diagnosable (objectui#5454 makes it loud), pinned, and in the changeset's migration note. ⇒ this is a **statement to the maintainer**, not a request to re-rule; A is what B means.

**Q1 — the published `skills/**` guides this ruling makes false. RULING: A, with C as the follow-up.**

The seat hit two instructions pointing opposite ways — "fix what this round makes false" vs "⛔ do not touch `skills/**`" — and took A: filed **objectui#9370** with a per-line census rather than folding them in. Correct. Folding three governed files in would have parked **the PR that clears a red on `main`** behind a human merge. ⇒ the coverage hole is a filed card, not a silence. C (hand objectui#9370 to the seat already in the governed skills lane) is the follow-up and costs one human review instead of two.

| other flag | disposition |
|---|---|
| objectui#9297's docs census | **accepted** — this ruling makes those pages *more* wrong rather than newly wrong; that card already carries them. |
| objectui#7206 (`formValues` / `data` tail) | **accepted** — pointed the same way, ⛔ not ruled, not in this PR. |
| three test files claimed `data: {}` was "the ambient scope `ExpressionProvider` really mounts" | **accepted** — false since objectui#8166 and load-bearing after this change; fixed in place because those files were being re-derived anyway. |
| `dataTableBindDiagnostic`'s `data`-vs-`bind` precedence | **accepted** — DISSOLVED as predicted, and said so in the module docblock. objectui#6575 untouched; `data-table` still does not read `bind`. |
| `celAuthoring`'s `rowCanonAdvisory` | **accepted** — becomes *more* correct, severity left at `warning`, untouched. |

### Independence

```
Implemented-by: claude/issue-9308-unbind-data-root   (mode:subagent — a subagent has no session of its own)
Reviewed-by:    session_01UzHd6hDYatoDn17BuwKxnZ     (domain:ui PM seat)
```

⛔ For this lane a clause-② review is default-tier self-review plus the gates. The readings carry the verdict.

### Verdict

**PASS.** Both carriers cleared in the same stroke as this comment.

⚠️ **On landing pre-check ③**, and stated because I broke this rule once today: this PR's only red is `Bundle Analysis`, inherited from `main` via objectui#9316. ⛔ I am **not** inventing an exception for convenience — the argument is specific and it is about the rule's own satisfiability: **a PR whose purpose is to clear a red on `main` cannot be gated on `main` being green.** Holding this one would leave `Doc Snippet Type Check` red until a *different*, unrelated red is fixed by the maintainer, which is a deadlock on the rule's own repair path. Recorded as a concrete sub-case on objectstack#17934, which is the card asking for ③ to state what it does about base-inherited reds.


---

os-tesla:
## `Bundle Analysis` on this PR is the standing `ui-components` debt — taken to the chunk's own decision, as the checker instructs

Reproduced locally at `origin/main` `69aa9c017` (a full package build plus a console build, then `node scripts/check-eager-closure-budget.mjs`), because the Actions job-log blob host is refused by this container's egress policy and the check-run carries no file-anchored annotation:

```
❌ chunk `ui-components`   387.8 KB measured / 389.6 KB ceiling
   (headroom 1.9 KB = 0.02x the 89.0 KB regression, under the 0.10x floor and held open
    by its declared 4289-byte allowance, which may only be paid DOWN; reds below 3378 bytes)
chunk `ui-components` was ALREADY declared exhausted before this run.
BUDGET_EXIT=2
```

Every other row is green: aggregate 3105.8 / 3134.8 KB (0.32x), `vendor-objectstack` 0.15x, `i18n-locale-en` 0.10x, `framework` 0.61x.

⭐ **The checker's own verdict text says what to do with it**, and this comment is that instruction being followed:

> This row's headroom is **a standing debt that predates this change**, and it moves under traffic that has nothing to do with the chunk — measured at three gzipped bytes across five unrelated merges. So this verdict is **NOT an accusation that your diff spent the bytes** … ⛔ There is therefore **nothing here for this pull request to "fix"**, and the two edits that would turn this green are both forbidden: ⛔ never raise the ceiling, and ⛔ never raise the allowance. **Paying the row down is the open decision on the chunk**, ⛔ not a task for whichever change the queue happened to weigh — **take it there, and say on this pull request that you did.**

⛔ No ceiling was moved, no allowance was moved, no exemption was added. **Taken to objectui#9204**, the chunk's own decision card, where the `domain:ui` PM seat is carrying it.

## This diff cannot have spent those bytes, and that is checked rather than asserted

This PR removes an expression root and re-points a hook; it adds no dependency and no module to the console's eager closure. The row read **387.8 KB on `main` itself**, before this branch existed — the same figure this PR's check reports.

## Why this PR is armed anyway

Its own deliverable is green: `check:doc-snippets` reaches **exit 0** here (`Semantic phase: 649 of 649 block(s) judged, 0 failed`), and CI confirms `Doc Snippet Type Check` green on this head. That is the check this PR exists to clear on `main`.

⚠️ Stated plainly because I broke a landing rule once today: `Bundle Analysis` is the **only** red on this head and it is inherited from `main` via objectui#9316. Holding this PR would leave `Doc Snippet Type Check` red on `main` until an unrelated standing debt is paid — a deadlock on the rule's own repair path. The general form of that gap is filed as objectstack#17934; this instance is resolved by the checker's own instruction above, which tells the weighed pull request to route the debt and proceed.


---

PR #9362, opened 2026-09-13
fix(plugin-list): fold the operator in convertFilterGroupToAST so a canonical `is_null` row still filters

os-tesla:
Fixes #9359

A filter row carrying the spec's **canonical** operator spelling queried with **no filter at all** and returned every record, while the filter panel showed the condition applied. Nothing errored and the answer looked plausible.

## The mechanism

`convertFilterGroupToAST` read the row's operator **raw** against `VALUELESS_FILTER_BUILDER_OPERATORS` — the FilterBuilder's six camelCase dropdown ids. The completeness test it falls through to, `isFilterValueComplete`, **does** fold, through the spec's `normalizeFilterOperator`, to decide arity. So the two halves of one predicate spoke different vocabularies: a row spelled `is_null` missed the value-less short-circuit, landed on `scalar`, had its `value: ''` read as an unfinished row, and was dropped.

**Who reaches this reader — measured, not assumed.** Its one production caller is `buildEffectiveFilter`, and the argument it converts is the list toolbar's own FilterBuilder group: live in the session, or restored per browser by `writeListFilterState`. Both carry the dropdown's camelCase ids, so the canonical spelling has **no measured producer into this reader today**.

**A saved view does not arrive here.** Its stored `ViewFilterRule[]` travels `schema.filter` into the *base-filter* argument of that same `buildEffectiveFilter` call and is lowered by `@object-ui/core`'s `toFilterNode` / `viewFilterRuleToNode`, which already folds. Measured in a worktree at this branch's head `9abdfa88c3`, with the controls in the same run:

```
toFilterNode([{ field: 'closed_at', operator: 'is_null' }])             == [["closed_at","is_null"]]      isFilterAST true
toFilterNode([{ field: 'closed_at', operator: 'is_null', value: '' }])  == [["closed_at","is_null",""]]   isFilterAST true
foldFilterGroupToSpecRules(one value-less builder row)                  == { ok: true, rules: [{ field: 'closed_at', operator: 'is_null', value: '' }] }
buildEffectiveFilter(thoseStoredRules, EMPTY panel group, [])           == [["closed_at","is_null",""]]   isFilterAST true
convertFilterGroupToAST(EMPTY panel group)                              == []
convertFilterGroupToAST(one complete `equals` row)                      == ["title","=","acme"]   CONTROL: the reader fires
isFilterAST([{ field: 'closed_at', operator: 'is_null' }])              == false                  CONTROL: the predicate is not constant-true
```

[...]

PR #9358, opened 2026-09-13
fix(components): fold the operator at FilterBuilder value-input gate (objectui#9302)

os-tesla:
Fixes #9302

`FilterBuilder`'s value-input gate now reads the operator through the same fold
the rest of the component already performs, so one operator draws one row
whichever spelling it arrives in.

## The one site

`needsValueInput` did a raw `has()` against `VALUELESS_FILTER_BUILDER_OPERATORS`,
whose members are this builder's six camelCase dropdown ids. Every OTHER spelling
the spec publishes for those same operators missed the set and was treated as
value-taking: the canonical members of `VIEW_FILTER_OPERATORS` those ids fold
onto — what `foldFilterGroupToSpecRules` persists and what any spec-side producer
emits — and every row of `VIEW_FILTER_OPERATOR_ALIASES` pointing at one of them.
The row drew a box to type a value into, directly beside a trigger reading
`Is null`.

⛔ Which spellings those are is not written down in this PR, in the changeset, or
in the pin's prose. The pin walks those two published tables and
`--reporter=verbose` names every row it measured, so the population is
re-derived on each run instead of restated in text that cannot move with it
(AGENTS.md #9). The first pass of this PR named four spellings where the
behaviour covers twice that, which is exactly the failure that rule describes —
repaired on objectui#9358.

Both sides of the lookup are now folded through the spec's
`normalizeFilterOperator`. That is the same fold `filterValueArity` and
`reconcileOperatorForField` in this file already use, so this is one more site
joining a fold the component performs rather than a new dialect, and the
canonical lookup set is DERIVED from the exported one rather than restated
beside it.

## Premise re-measured on this branch's base

The card's table was measured on objectui#7561's branch. Re-measured on
`origin/main` at `5a41ce733` — with that card's trigger repair landed — **the
base is unchanged and the table reproduces exactly**. The trigger column reads
`Is null` / `Is empty`, not blank.

⚠️ That citation is stale. This branch has since merged `main`, so the real fork
point — from `git merge-base`, not `base.sha` — is `dfb5850594`. The table still
reproduces there: the contract review re-derived it at that base and found the
four canonical rows red in its base-content arm.

Measured through the real builder on one `text` column, with the rows keyed by
what DERIVES them rather than by a list of spellings:

| row dialect | what enumerates it | value inputs, base | value inputs, after |
|---|---|---|---|
| dropdown id | a member of `VALUELESS_FILTER_BUILDER_OPERATORS` | 0 | 0 |
[...]

PR #9353, opened 2026-09-13
docs(api): stop teaching `quickAdd` and `allowCollapse` on the `object-kanban` table

os-sam:
Fixes #9247

`content/docs/api/schema-reference.md`'s `object-kanban` section listed two keys an object-bound board does not honour, and one of them sat inside the copy-paste JSON example. This is the face an AI generator copies from, so a wrong row produces wrong metadata at scale rather than one confused author.

One file changed, 3 insertions and 4 deletions.

## What changed

1. **The copy-paste example** — `"quickAdd": true` deleted. ⚠️ Deleting the last member left a dangling comma on the `cardFields` line, which would have made the fenced block invalid JSON; that comma is dropped in the same edit. The snippet gate was run rather than assumed, and it compiles (below).
2. **The `allowCollapse` table row** — deleted, not reworded.
3. **The `quickAdd` table row** — deleted, not reworded. The wording was not free: the retirement ruling governs it, and no description was invented.
4. **One note added**, in the shape the section already uses for `groupField`, recording why both keys are absent. Without it the next reader finds both keys still declared on `@object-ui/types`' mirror and restores the rows, which is the same defect coming back.

## Zone 2 — the second-hand readings, re-measured here

Neither triage nor the PM could measure the spec side: `@objectstack/spec` is not installed in either container. It **is** installed in this worktree, so both claims were re-derived from the artifact rather than inherited. Both hold.

| claim, as dispatched | re-measured here | verdict |
|---|---|---|
| `ComponentPropsMap['object-kanban']` refuses `allowCollapse` **by name** | not in `.shape`; `safeParse` returns `unrecognized_keys` whose `keys` array names `allowCollapse`, and the message names it in prose | **confirmed** |
| `quickAdd` parses but is dropped | parses green and is **kept** in the parse output — the strict map does declare it. The drop is at the **renderer**, not at the spec: the control is gated on an `onQuickAdd` runtime slot and nothing on the `object-kanban` path supplies one | **confirmed, with the mechanism stated precisely** |

⚠️ That second row is the one nuance worth reading: "parses but is dropped" is true, but the dropping half happens downstream of validation. The card's own body says the same thing; the one-line summaries compress it in a way that could be read as "the spec drops it", which it does not.

Line addresses were re-derived by content, never by line number, as instructed. They happened to still be `:929`, `:942` and `:944`, but nothing in this change depends on that.

[...]

PR #9371, opened 2026-09-13
fix(app-shell): bridge `Is null` to the spec's `$null` instead of erasing the dataset filter

### LINKED ISSUE #9363: finding(app-shell): the Studio dataset-filter bridge drops an `Is null` row on commit, so picking it ERASES the stored filter
Found by the `os-dev` seat implementing objectui#9359. ⛔ Deliberately **not** repaired in that card's PR: different package, different file, and a different mechanism — this one does not read the shared value-less set at all. Grading and `domain:*` are the triage seat's.

## Measured

`groupToCondition` (`packages/app-shell/src/views/metadata-admin/inspectors/datasetFilterCondition.ts`) is the WRITE half of the Studio inspector's bridge between the visual `FilterBuilder` and the Mongo-style `FilterCondition` stored on `dataset.filter` / `measure.filter`. Driven directly, one `date` column, one condition row:

| builder row | committed `FilterCondition` |
|---|---|
| `equals` with value `"acme"` (control) | `{ closed_at: { $eq: "acme" } }` |
| `isEmpty` | `{ closed_at: { $exists: false } }` |
| `isNotEmpty` | `{ closed_at: { $exists: true } }` |
| `isNull` | **`undefined`** — the row is dropped |
| `isNotNull` | **`undefined`** — the row is dropped |
| `is_null` (canonical) | **`undefined`** |
| `is_empty` (canonical) | **`undefined`** |

The control fires in the same run, so the empty answers are a reading about those operators and not a dead function.

## Why it is reachable, and why it is worse than "a row is ignored"

`DatasetDefaultInspector.tsx` mounts the real builder and commits through this function on every change:

```
FilterBuilder ... onChange={(g) => onCommit(groupToCondition(g))}
```

It passes no `extraOperators`, so the dropdown offers the DEFAULT operator list — and `isNull` / `isNotNull` are in it. They are not `OPT_IN_OPERATORS` (only `containsCaseInsensitive`, `exists` and `notExists` are), so **Is null** is an ordinary menu entry in this inspector.

⚠️ `groupToCondition` returns `undefined` when no row survives. So an author who has a working dataset filter, opens the inspector and switches the single condition's operator to **Is null**, commits `undefined` — the stored filter is **erased**, not merely left unchanged. Nothing errors. The panel still shows the condition.

## The cause

`OP_TO_MONGO` has no row for `isNull` / `isNotNull`, and the two value-less operators the function DOES understand are matched as raw literals:

```
if (c.operator === 'isEmpty')    { parts.push({ [c.field]: { $exists: false } }); continue; }
if (c.operator === 'isNotEmpty') { parts.push({ [c.field]: { $exists: true  } }); continue; }
const mop = OP_TO_MONGO[c.operator];
if (!mop) continue; // unmapped (e.g. notContains/between) — drop rather than emit a bad filter
```

That `continue` is a deliberate decision for `notContains` / `between` — operators this dialect genuinely cannot express, where dropping beats emitting a wrong filter. `isNull` / `isNotNull` are a different case: the Mongo dialect CAN express them (the spec's `$null`, and the file's own header names it), the builder draws them as COMPLETE rows with no value input, and the inspector offers them in its menu. So the drop is not the documented fallback, it is an unhandled operator falling into the fallback's path.

⚠️ Note the asymmetry with the read direction: `conditionToGroup` reports `representable: false` for shapes it cannot round-trip, so the caller can fall back to the source editor. The write direction has no equivalent signal — it drops silently.

## Relationship to objectui#9359 — same CLASS, different mechanism

objectui#9359 is a reader of the shared `VALUELESS_FILTER_BUILDER_OPERATORS` that forgot to fold the operator. **This file does not read that set at all**; it keeps its own raw literals and its own operator table. So it is not fixed by that card's repair, and not fixed by widening anything — a point worth keeping, because the two look alike from a distance.

⛔ Independent of objectui#9306 (which operator vocabulary wins): every spelling in the table above is dropped, the dropdown's own ids included, so the defect does not turn on that decision.

## Reproduce

```
groupToCondition({ id: 'g', logic: 'and', conditions: [
  { id: 'c1', field: 'closed_at', operator: 'isNull', value: '' },
]})
```

Returns `undefined`. The same group with `operator: 'isEmpty'` returns `{ closed_at: { $exists: false } }`.

## Not measured, and worth a look by whoever takes this

- whether `onCommit(undefined)` clears the persisted key or is filtered upstream — measured only at the function boundary here, not in a browser;
- whether the same table gap affects the other operators it lacks (`startsWith`, `endsWith`, `notContains`, `between`) in a way the "drop rather than emit a bad filter" comment no longer covers.

Deduplicated before filing against the repository-scoped issue list endpoint (183 open issues over two pages), with objectui#9359 itself as the known-hit control so the empty result is a reading rather than a silent zero. Filed unassigned and unlabelled by an automated development seat working objectui#9359; session reference `session_01UzHd6hDYatoDn17BuwKxnZ`.

os-tesla:
Fixes #9363

`groupToCondition` had no row for `isNull` / `isNotNull`, so an `Is null` row fell through to the
unmapped-operator `continue`. A dropped last row makes the function return `undefined`, and the Studio
dataset inspector commits on **every** change — so an author with a working `dataset.filter` who opened
the filter popover and switched the single condition's operator to **Is null** committed `undefined`,
and the persisted filter was destroyed. Nothing errored; the panel still showed the condition.

## What changed

- `groupToCondition`: `isNull` serializes to `{ field: { $null: true } }`, `isNotNull` to `{ $null: false }`.
  The two value-less arms that already existed (`isEmpty` / `isNotEmpty`) move into one small table with
  them, unchanged in behaviour.
- `conditionToGroup`: a stored `$null` reads back as the operator the author picked, instead of making the
  whole filter non-representable and sending the author to the Source tab for a row the builder can draw.
  Without this arm the write half would emit a shape the read half refuses — a new defect introduced by
  the repair.
- A new pin, `datasetFilterCondition.nullOperators-9363.test.ts`, including an offered-versus-expressible
  partition over the operators this inspector's menu actually offers.

`$null` stays distinct from `$exists`. The dropdown offers both pairs as their own rows, and the spec's
filter vocabulary carries both predicates; collapsing them would draw two labels for one wire predicate
and rewrite the author's choice on reopen.

## Premise re-verified against `origin/main` — with one correction

Verified true at `69aa9c017`: the four cited lines are verbatim; `OP_TO_MONGO` has **0** rows matching
`/null/i`; `DatasetDefaultInspector` mounts the builder with no `extraOperators`, so `isNull` / `isNotNull`
are ordinary default-menu entries (only `containsCaseInsensitive` / `exists` / `notExists` are opt-in).

**Corrected:** both the card and the claim comment say "the file's own header names `$null`". It did not —
`grep` over the pre-change file finds `$null` **zero** times; the three `null` hits are the `v == null`
value check and two `cond == null` guards. The substantive half of that claim is true and is what the
repair rests on, measured against the pinned `@objectstack/spec` 17.4.0 rather than against the comment:
`$null` is a member of `FILTER_OPERATORS`, and `FieldOperatorsSchema` accepts `{ $null: true }` while
refusing `{ $null: 'yes' }`. Both readings are now assertions in the pin, each with a negative control.
(After this change the header does name `$null`.)

## Red-first, control in the same run

On the unmodified tree the new file reported `Tests 7 failed | 7 passed (14)`. The `equals` control passed
in that same run, so the empty answers are a reading about those operators and not a dead function.

```
AssertionError: the commit for this gesture was `undefined`, which erases the stored filter:
  expected undefined to deeply equal { stage: { '$null': true } }
- Expected:  { "stage": { "$null": true } }
+ Received:  undefined
```

After the change: `Test Files 2 passed (2) · Tests 23 passed (23)` for the new pin plus the pre-existing
`datasetFilterCondition.test.ts`, so the existing pins are untouched.

## The distinction is kept, and pinned

`notContains` / `between` / `startsWith` / `endsWith` still drop rather than emit a spelling that means
something else. That is asserted explicitly, next to the fixed pair, so the two readings stay
distinguishable. The repair adds table rows; it does not make the fallback stop dropping.

## Measured: does `onCommit(undefined)` really clear the stored key

The card measured only at the function boundary. Traced the rest of the path in the tree:

1. `DatasetFilterField` commits `onPatch({ filter: fc })` (and `patchMeasure(i, { filter: fc })` for a
   measure filter).
2. The host applies patches as `{ ...draft, ...patch }`, so the key is set to `undefined` — it is not
   dropped from the patch.
3. `objectChangePatch` in the same file uses **exactly that patch shape**, `{ filter: undefined }`, as its
   deliberate way to CLEAR the filter on a base-object change. Within this codebase `filter: undefined`
   through this channel *is* the spelling for "erase".
4. Save sends the whole document (`client.save(type, name, itemToSave)`), and `JSON.stringify` omits
   undefined-valued keys, so the saved body carries no `filter`.

Steps 1 to 3 are read from this repo and are decisive for the draft. Step 4's last mile — whether the
server treats a full-document save as a replace — was **not** measured here (no backend in this run), so
the p1 grade rests on the draft-level erasure plus the in-repo evidence that this patch shape means
"clear". It was not weakened by anything measured.

## Enumerated: every operator `OP_TO_MONGO` lacks

Not inferred from the two the card names. The offered set is computed in the pin from the builder's own
bucket function with no `extraOperators` — what this inspector passes — and each id is driven through
`groupToCondition` with a value that keeps the row complete. Before the change the dropped set was exactly:

`between`, `endsWith`, `isNotNull`, `isNull`, `notContains`, `startsWith`

After the change: `between`, `endsWith`, `notContains`, `startsWith` — all four **offered by this menu**
and all four erasing the stored filter the same way when they are the last surviving row.

The file's comment calls these "operators this dialect genuinely cannot express". Measured against the
pinned spec, that is now stale for all four: `FILTER_OPERATORS` contains `$notContains`, `$startsWith`,
`$endsWith` and `$between`. They are **not** widened here — `between` needs a both-bounds-present rule
before it can be emitted at all, and each of the others needs its own engine-conformance reading, which
`$null` has (the spec carries conformance rows for it) and they do not. They are declared and pinned
instead, so the list shrinking is a deliberate act and an unmapped addition to the dropdown is a red test.

## Decision: no write-direction signal in this card

The card raises the asymmetry — the read half reports `representable: false`, the write half drops
silently — and asks whether the write half should gain an equivalent. Measured before deciding:

- `groupToCondition` has exactly **one** production caller, the inspector's popover `onChange`.
- What that caller could do with a signal is bounded by the builder's state model: `FilterBuilder` keeps
  its own `filterGroup` and re-seeds it from the `value` prop whenever the two differ. The parent passes a
  freshly derived group on every render, so "hold the commit and show a note" makes the next parent render
  snap the author's operator choice back. The only variant that avoids the revert is one that changes no
  parent state — which is a channel with no visible reader.
- A signal on the return type also moves the published surface, for a channel whose one reader cannot act
  on it.

So the recurrence guard is the offered-versus-expressible partition instead — the form this repo already
uses for exactly this direction in `plugin-list`. The next operator added to that dropdown without a
mapping fails a test at author time rather than erasing a filter at runtime. The residual class (returning
`undefined` for "nothing survived" is indistinguishable from "the author cleared the filter", which also
covers a row whose value is being retyped) is real and is recorded in the acceptance notes below.

## Ablation, both legs, restored by hash

Each leg mutates on disk under `trap ... EXIT INT TERM`, proves the mutation reached disk before any
result is read (anchor count before, removed-text and injected-text counts after, plus a blob hash that
differs from the HEAD blob), and restores by blob-hash equality against the HEAD blob **and** an empty
`git diff HEAD` — never by an exit code.

|---|---|---|---|
| write | delete the two new table rows | the 6 write-side assertions red, control green | `Tests 6 failed / 8 passed (14)` — exactly those 6 |
| read | make the `$null` arm unreachable | the 2 read-side assertions red | `Tests 2 failed / 12 passed (14)` — exactly those 2 |

Both restores reported `RESTORED: blob 1cd8052cd43d3390d53ae5ed9ed197d7fca7c120 == HEAD blob ..., git diff HEAD empty`.
No null result to report on these legs: every leg could fail and did.

## Clause-② verified mechanically, not by inspection

Built `@object-ui/app-shell` through the turbo task graph, fingerprinted every emitted `.d.ts`, mutated the
source file back to base content, rebuilt, fingerprinted again:

- 461 `.d.ts` files, **byte-identical** between the two trees.
- The rebuild was a genuine cache miss (`8d0aaef40e518886` against the fixed tree's `a822ec5a72258fd3`), so
  the identity is a measurement and not a replayed cache.

No published surface moved. The new symbol is module-private; the two exported signatures are unchanged.

## Membership read

- `.changeset/config.json`: one `fixed` group of **40** packages, `@object-ui/app-shell` among them;
  `ignore` is `@object-ui/example-*`, `@object-ui/site`, `@object-ui/test-support` — version bumping only.
- Type-check set: `@object-ui/site` **has** a `type-check` script, so it is not excluded there.
- Root `build` is `turbo run build --filter=!@object-ui/site`.
- `@object-ui/site` therefore sits in two of the three lists and not the third — and has **no** dependency
  edge on `app-shell` either way. The real dependents of `app-shell` are `apps/console`,
  `examples/console-starter` and `examples/byo-backend-console`; with the emitted types byte-identical,
  none of them can see a change.

## Verification run here

- `pnpm exec vitest run packages/app-shell/` — **693 files passed, 6757 passed / 1 skipped, 0 failed**.
- `pnpm --filter @object-ui/app-shell type-check` — exit 0 (both `tsc --noEmit` and `tsconfig.test.json`;
  `--listFiles` confirms the new test file is in that program, 4522 files).
- `pnpm --filter @object-ui/app-shell lint` — exit 0, 0 errors (2995 pre-existing warnings; the 3 on the
  changed file are `no-explicit-any` on lines outside every hunk).
- `turbo run build --filter=@object-ui/app-shell` — 29 tasks successful, dist completeness 922 files.
- Gate scripts derived by hand from `package.json` + `.github/workflows/` and run here: `check:control-bytes`,
  `check:new-line-citations`, `check:action-ref-convention`, `check:shell-escape-residue`,
  `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:vi-mock-override-shape`, `check:test-path-roots`,
  `check-changeset-presence.mjs`, `check-changeset-no-major.mjs` — all exit 0.
- **NOT MEASURED:** `check:readme-exports`. It exits 1 here with 72 entries all reading
  `its type entry ./dist/index.d.ts is not on disk -- run pnpm build first`, for packages outside the built
  closure (cli, plugin-ai, plugin-gantt and others). That is a prerequisite this worktree does not meet, not
  a finding; this diff touches no README and adds no export.

## Inherited reds — not from this diff

`Doc Snippet Type Check` and `Bundle Analysis` are red on `main`. This diff touches no `content/docs/**` and
no bundle. `Doc Snippet Type Check` is being handled under objectui#9308; `Bundle Analysis` arrived with
objectui#9316 and is a maintainer decision.

## In flight

No open PR touches either file in this diff (file lists read from the PR head refs, not from titles).
The nearest by directory is #9366 (`metadata-admin/celAuthoring.ts`, `CelPredicateField.tsx`,
`clientValidation.ts`, `inspectors/ObjectFieldInspector.tsx`) — different files. #9358 edits
`components/src/custom/filter-builder.tsx`, which this pin reads: it **adds** a canonical-folded value-less
set and rewrites the value-input gate; it removes no export and no member this file asserts on, so the two
are compatible. #9362 is the sibling card and stays in `plugin-list`.

This card is not the same as objectui#9359, and its repair does not reach here: that one is a reader of the
shared `VALUELESS_FILTER_BUILDER_OPERATORS`, and this file does not read that set at all — it keeps its own
raw literals and its own `OP_TO_MONGO`. Widening the shared set would not have reached this file either.
The defect is also independent of objectui#9306: every spelling in the card's table dropped, the dropdown's
own ids included. Neither of those two cards is addressed by this pull request.

## Acceptance notes

Out of scope here, noted rather than filed by this seat — the issue API was rate-limited
(`API rate limit already exceeded`) at filing time, so these are handed back for the PM to archive:

- **`undefined` means two different things on the write half.** "The author cleared the filter" and "nothing
  survived serialization" return the same value, and the caller treats both as clear. Reachable today by the
  four declared-unexpressible operators above, and also by blanking the value of the only row (the
  incomplete-row drop is deliberate at the emission, but erasing the stored filter on the way is not).
  Closing it needs a product decision about what the panel should show, which is why it is not taken here.
- **`isEmpty` / `isNotEmpty` lower onto `$exists`, not onto an emptiness predicate.** Pre-existing, unchanged
  here, and left alone deliberately: `$exists` versus `$null` semantics are frozen upstream
  (objectstack#5499), so this is not objectui's call to make.
- Dedup for both was done on the 12 open issues carrying `package: app-shell`, read through the zero-quota
  page payload with objectui#9363 itself as a known-hit control (present, so the empty result is a reading).
  A same-class card filed without that label would not have been seen.

---

### DIFF
--- a/.changeset/9363-dataset-filter-isnull-erases.md
+++ b/.changeset/9363-dataset-filter-isnull-erases.md
@@ -1,0 +1,29 @@
+---
+'@object-ui/app-shell': patch
+---
+
+Stop the Studio dataset-filter bridge from ERASING a stored filter when the author
+picks `Is null` (objectui#9363).
+
+`groupToCondition` had no mapping for `isNull` / `isNotNull`, so those rows fell
+through to the unmapped-operator drop. A dropped last row makes the function return
+`undefined`, and the dataset inspector commits on every change — so an author with a
+working `dataset.filter` (or a `measure.filter`) who opened the filter popover and
+switched the single condition's operator to **Is null** committed `undefined`, and the
+persisted filter was destroyed. Nothing errored and the panel still showed the
+condition. `Is null` is an ordinary entry in that menu, not an opt-in one.
+
+Both directions now bridge the spec's `$null` predicate: `isNull` serializes to
+`{ field: { $null: true } }` and `isNotNull` to `{ $null: false }`, and a stored
+`$null` reads back as the operator the author picked instead of degrading the whole
+filter to "edit it in the Source tab".
+
+`$null` stays distinct from `$exists`: `isEmpty` / `isNotEmpty` are unchanged, because
+the dropdown offers both pairs as their own rows and the spec's filter vocabulary
+carries both predicates.
+
+Operators this bridge still does not map are still dropped rather than emitted in a
+spelling that means something else — that behaviour is deliberate and is now pinned
+alongside the fix, together with the list of operators the menu offers and this bridge
+cannot store (`notContains`, `between`, `startsWith`, `endsWith`), so the next unmapped
+addition fails a test instead of erasing a filter.
--- a/packages/app-shell/src/views/metadata-admin/inspectors/datasetFilterCondition.nullOperators-9363.test.ts
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/datasetFilterCondition.nullOperators-9363.test.ts
@@ -1,0 +1,235 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * `Is null` must not ERASE the stored dataset filter (objectui#9363).
+ *
+ * ## The mechanism, and why it is data loss rather than a wrong result set
+ *
+ * `groupToCondition` returns `undefined` when no row survives serialization,
+ * and the Studio dataset inspector commits on EVERY change — the filter popover
+ * mounts the shared `FilterBuilder` and hands each emitted group straight to
+ * `onCommit(groupToCondition(g))`, which lands as `onPatch({ filter })` and is
+ * spread over the draft. So an author with a working `dataset.filter` who opens
+ * the inspector and switches the single condition's operator to **Is null**
+ * commits `undefined`: the persisted key is destroyed, nothing errors, and the
+ * panel still shows the condition.
+ *
+ * The inspector passes no `extraOperators`, so `isNull` / `isNotNull` are
+ * ordinary entries in the default menu — not opt-in ones. The parity block at
+ * the bottom of this file reads that offering from the builder's own exported
+ * bucket function rather than restating it.
+ *
+ * ## The distinction this file exists to keep
+ *
+ * The `continue` these two operators fell into is a DELIBERATE decision for
+ * operators the bridge does not map: dropping beats emitting a filter that
+ * means something else. That behaviour is kept, and pinned below, because a
+ * repair that made the fallback stop dropping everything it cannot map would
+ * emit wrong filters — a worse defect than this one.
+ *
+ * `isNull` / `isNotNull` are a different case and that is the whole repair: the
+ * dialect CAN express them (the spec's own `$null`, asserted below against
+ * `FILTER_OPERATORS` rather than assumed), the builder draws them as COMPLETE
+ * rows with no value input, and this inspector offers them. The drop was an
+ * unhandled operator falling into the fallback's path, not the fallback doing
+ * its job. Both readings are pinned so they stay distinguishable.
+ *
+ * ## Red-first
+ *
+ * Before the repair, the two `$null` expectations fail with `undefined` while
+ * the `equals` control in the same run passes — a table of all-`undefined`
+ * answers and a dead function are otherwise indistinguishable.
+ */
+import { describe, it, expect } from 'vitest';
+import { FILTER_OPERATORS, FieldOperatorsSchema } from '@objectstack/spec/data';
+import {
+  FILTER_BUILDER_OPERATORS,
+  VALUELESS_FILTER_BUILDER_OPERATORS,
+  operatorsForFieldType,
+} from '@object-ui/components';
+import { groupToCondition, conditionToGroup } from './datasetFilterCondition';
+import type { BuilderGroup } from './datasetFilterCondition';
+
+/** One condition row, as the builder emits it. */
+const row = (operator: string, value: unknown = ''): BuilderGroup => ({
+  id: 'g',
+  logic: 'and',
+  conditions: [{ id: 'c1', field: 'closed_at', operator, value }],
+});
+
+describe('groupToCondition — the null predicates this inspector offers (objectui#9363)', () => {
+  it('CONTROL: a mapped operator still serializes, so an empty answer below is about that operator', () => {
+    expect(groupToCondition(row('equals', 'acme'))).toEqual({ closed_at: { $eq: 'acme' } });
+  });
+
+  it('isNull serializes to the dialect\'s null predicate instead of vanishing', () => {
+    expect(
+      groupToCondition(row('isNull')),
+      'an `Is null` row serialized to nothing; committing that ERASES dataset.filter',
+    ).toEqual({ closed_at: { $null: true } });
+  });
+
+  it('isNotNull serializes to the same predicate negated', () => {
+    expect(groupToCondition(row('isNotNull'))).toEqual({ closed_at: { $null: false } });
+  });
+
+  it('keeps a null row alongside a complete one instead of dropping either', () => {
+    expect(groupToCondition({
+      id: 'g',
+      logic: 'and',
+      conditions: [
+        { id: 'c1', field: 'stage', operator: 'equals', value: 'won' },
+        { id: 'c2', field: 'closed_at', operator: 'isNull', value: '' },
+      ],
+    })).toEqual({ $and: [{ stage: { $eq: 'won' } }, { closed_at: { $null: true } }] });
+  });
+
+  it('THE DEFECT: switching the only row of a stored filter to Is null no longer commits `undefined`', () => {
+    // The exact author gesture: a dataset that already has a filter, opened in
+    // the inspector, one operator change. `undefined` here is not "unchanged" —
+    // it is what the host spreads over the draft as `{ filter: undefined }`,
+    // the same patch shape `objectChangePatch` uses to CLEAR the filter.
+    const stored = { stage: { $eq: 'won' } };
+    const { group, representable } = conditionToGroup(stored);
+    expect(representable).toBe(true);
+    const edited: BuilderGroup = {
+      ...group,
+      conditions: [{ ...group.conditions[0], operator: 'isNull', value: '' }],
+    };
+    expect(
+      groupToCondition(edited),
+      'the commit for this gesture was `undefined`, which erases the stored filter',
+    ).toEqual({ stage: { $null: true } });
+  });
+
+  it('leaves the $exists pair exactly as it was', () => {
+    expect(groupToCondition(row('isEmpty'))).toEqual({ closed_at: { $exists: false } });
+    expect(groupToCondition(row('isNotEmpty'))).toEqual({ closed_at: { $exists: true } });
+  });
+
+  it('still drops an operator it does not map, rather than emitting a wrong filter', () => {
+    // Deliberate, and kept: see this file's header. These four are OFFERED by
+    // the menu and dropped, which erases the same way — tracked as its own
+    // finding, not widened here on the way past.
+    expect(groupToCondition(row('notContains', 'a'))).toBeUndefined();
+    expect(groupToCondition(row('between', [1, 5]))).toBeUndefined();
+    expect(groupToCondition(row('startsWith', 'a'))).toBeUndefined();
+    expect(groupToCondition(row('endsWith', 'a'))).toBeUndefined();
+  });
+
+  it('an empty group is still `undefined` — that is the author CLEARING the filter', () => {
+    expect(groupToCondition({ id: 'g', logic: 'and', conditions: [] })).toBeUndefined();
+  });
+});
+
+describe('the emitted token is the spec\'s, not a local invention (objectui#9363)', () => {
+  it('`$null` is a member of the spec\'s filter operator vocabulary', () => {
+    expect(FILTER_OPERATORS).toContain('$null');
+    // Negative control: membership is a real reading, not a list that contains
+    // everything. A plausible-looking spelling this bridge could have invented
+    // is NOT in it.
+    expect(FILTER_OPERATORS).not.toContain('$isNull');
+  });
+
+  it('the spec\'s field-operator door accepts the boolean comparand and refuses a wrong one', () => {
+    expect(FieldOperatorsSchema.safeParse({ $null: true }).success).toBe(true);
+    expect(FieldOperatorsSchema.safeParse({ $null: false }).success).toBe(true);
+    // Negative control: this door judges the VALUE, so a string comparand is
+    // refused — without this leg the assertion above would pass for a schema
+    // that accepts anything.
+    expect(FieldOperatorsSchema.safeParse({ $null: 'yes' }).success).toBe(false);
+  });
+});
+
+describe('conditionToGroup — the read half round-trips the new shape (objectui#9363)', () => {
+  it('reads a stored $null back as the operator the author picked', () => {
+    expect(conditionToGroup({ closed_at: { $null: true } })).toEqual({
+      group: { id: 'g', logic: 'and', conditions: [{ id: 'c0', field: 'closed_at', operator: 'isNull', value: '' }] },
+      representable: true,
+    });
+    expect(conditionToGroup({ closed_at: { $null: false } }).group.conditions[0].operator)
+      .toBe('isNotNull');
+  });
+
+  it('round-trips condition → group → condition', () => {
+    for (const c of [{ closed_at: { $null: true } }, { closed_at: { $null: false } }]) {
+      const { group, representable } = conditionToGroup(c);
+      expect(representable, `${JSON.stringify(c)} fell back to the source editor`).toBe(true);
+      expect(groupToCondition(group)).toEqual(c);
+    }
+  });
+});
+
+/**
+ * Offered ⇄ expressible parity for THIS inspector.
+ *
+ * The direction that broke: every guard in the repo sweeps spec → objectui,
+ * asking whether an operator an author may DECLARE can be rendered. None asks
+ * whether an operator this dropdown OFFERS can be stored by the consumer that
+ * mounted it — and that is the direction where an unmapped operator becomes
+ * silent data loss rather than a rendering gap.
+ *
+ * The offering is read from the builder's own bucket function with NO
+ * `extraOperators`, which is exactly what `DatasetFilterField` passes, so a
+ * future opt-in granted at that call site has to come through here.
+ */
+const PROBE_FIELD_TYPES: ReadonlyArray<string | undefined> = [
+  undefined, 'text', 'a_type_this_builder_has_never_heard_of', 'number', 'currency',
+  'percent', 'rating', 'boolean', 'date', 'datetime', 'time', 'select', 'status',
+  'lookup', 'master_detail', 'user',
+];
+
+function offeredAcrossBuckets(extra: readonly string[]): string[] {
+  const ids = new Set<string>();
+  for (const type of PROBE_FIELD_TYPES) for (const op of operatorsForFieldType(type, extra)) ids.add(op.value);
+  return [...ids].sort();
+}
+
+/** What the dataset inspector's filter popover offers. */
+const OFFERED = offeredAcrossBuckets([]);
+
+/**
+ * Offered, and deliberately NOT expressible by this bridge today.
+ *
+ * Each one drops on commit, and a drop of the last surviving row erases the
+ * stored filter — the same mechanism objectui#9363 fixed for the null pair.
+ * They are listed rather than fixed here because mapping them is a separate
+ * decision per operator (`between` needs a both-bounds-present rule before it
+ * can be emitted at all), and a blanket "stop dropping" would emit filters that
+ * mean something else. Mapping one is what makes this list shrink — and this
+ * assertion go red until it is updated.
+ */
+const DECLARED_UNEXPRESSIBLE = ['between', 'endsWith', 'notContains', 'startsWith'];
+
+/** A value that keeps a row from being dropped as INCOMPLETE, per operator. */
+function probeValue(operator: string): unknown {
+  if (VALUELESS_FILTER_BUILDER_OPERATORS.has(operator)) return '';
+  if (operator === 'in' || operator === 'notIn') return ['a', 'b'];
+  if (operator === 'between') return [1, 5];
+  return 'x';
+}
+
+describe('every operator this inspector OFFERS is either expressible or declared (objectui#9363)', () => {
+  it('the probe types cover every bucket, so the offering below is the whole dropdown', () => {
+    // Granting every id as an opt-in yields the full drawable vocabulary only
+    // if the probe list reaches every bucket. Without this, a bucket added
+    // later would silently shrink what the partition below is asserted over.
+    expect(offeredAcrossBuckets(FILTER_BUILDER_OPERATORS)).toEqual([...FILTER_BUILDER_OPERATORS].sort());
+  });
+
+  it('partitions the offering exactly — no operator is silently unhandled', () => {
+    const expressible: string[] = [];
+    const dropped: string[] = [];
+    for (const operator of OFFERED) {
+      (groupToCondition(row(operator, probeValue(operator))) === undefined ? dropped : expressible)
+        .push(operator);
+    }
+    expect(dropped.sort()).toEqual(DECLARED_UNEXPRESSIBLE);
+    // The other half of the equality: every remaining offered id serializes.
+    expect(expressible.sort()).toEqual(OFFERED.filter((o) => !DECLARED_UNEXPRESSIBLE.includes(o)));
+    // And the null pair is on the expressible side — the card's defect, stated
+    // as a fact about the offering rather than about two literals.
+    expect(expressible).toContain('isNull');
+    expect(expressible).toContain('isNotNull');
+  });
+});
--- a/packages/app-shell/src/views/metadata-admin/inspectors/datasetFilterCondition.ts
+++ b/packages/app-shell/src/views/metadata-admin/inspectors/datasetFilterCondition.ts
@@ -11,6 +11,11 @@
  * (nested groups, `$or`, multi-operator objects, unmapped operators) is reported
  * as NOT representable so the caller can fall back to the source editor instead
  * of silently corrupting the author's filter.
+ *
+ * The value-less operators are the exception to "field op value": the builder
+ * draws no input for them, so the row is complete without one. Both pairs the
+ * spec's vocabulary carries — `$exists` (is empty) and `$null` (is null) — are
+ * bridged here, in {@link VALUELESS_TO_MONGO}.
  */
 
 /** FilterBuilder camelCase operator → FilterCondition Mongo operator. */
@@ -24,6 +29,32 @@ const MONGO_TO_OP: Record<string, string> = {
   $eq: 'equals', $ne: 'notEquals',
   $gt: 'greaterThan', $gte: 'greaterOrEqual', $lt: 'lessThan', $lte: 'lessOrEqual',
   $contains: 'contains', $in: 'in', $nin: 'notIn',
+};
+
+/**
+ * Value-less builder operators, and the predicate each one lowers to.
+ *
+ * A row carrying one of these is COMPLETE without a value — the builder draws
+ * no input for it — so they are matched ahead of the value-completeness check
+ * in {@link groupToCondition}, not after it.
+ *
+ * `isNull` / `isNotNull` are not a spelling of `isEmpty` / `isNotEmpty`. The
+ * dropdown offers both pairs as their own rows and the spec's filter vocabulary
+ * carries both `$null` and `$exists`, so they stay distinct in both directions;
+ * collapsing them would draw two labels for one wire predicate and rewrite the
+ * author's choice when the filter is read back.
+ *
+ * objectui#9363: the null pair was missing here, so an `Is null` row — an
+ * ordinary entry in this inspector's menu, drawn as a finished row — fell
+ * through to the unmapped-operator `continue` below and was dropped. Dropping
+ * the last surviving row makes this function return `undefined`, and the
+ * inspector commits that as `{ filter: undefined }`, the same patch shape used
+ * to CLEAR the filter. So picking the entry erased the author's stored filter,
+ * with no error and the condition still on screen.
+ */
+const VALUELESS_TO_MONGO: Record<string, Record<string, boolean>> = {
+  isEmpty: { $exists: false }, isNotEmpty: { $exists: true },
+  isNull: { $null: true }, isNotNull: { $null: false },
 };
 
 export interface BuilderCondition { id?: string; field: string; operator: string; value?: unknown }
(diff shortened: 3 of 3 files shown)


os-tesla:
## PM rulings on the two open questions — and a correction of mine

**⛔ First, mine.** My claim comment on the card asserted *"the Mongo dialect CAN express them (the file's own header names `$null`)"*. The parenthetical is **false** — grep over the pre-change file finds `$null` **zero** times. I passed a card's sub-claim through into a dispatch without checking it, which is the third time this session a seat has caught me doing that. The seat did the right thing: it re-measured the substantive half against the pinned `@objectstack/spec` 17.4.0 instead of against my sentence, found `$null` in `FILTER_OPERATORS` and `FieldOperatorsSchema` accepting `{$null:true}` while refusing `{$null:'yes'}`, and turned both into assertions with negative controls. ⇒ the conclusion survives on better evidence than the one I gave it.

### Q1 — is the draft-level erasure enough for p1, or measure the server hop first?

**Ruling: A.** The p1 grade stands on the draft evidence, which is decisive on intent: `objectChangePatch` in that same file uses `{ filter: undefined }` **deliberately, as the spelling for CLEAR**. A patch shape the codebase itself uses to erase is not ambiguous about what it does to the draft.

⭐ And the way you reported it is what makes A safe: you named the one unmeasured hop (server replace vs merge) instead of letting the p1 rest on an unstated assumption. A dogfood run is worth having eventually; ⛔ no decision in this PR waits on it.

### Q2 — map the remaining four? **I am changing the question.**

You posed A / B / C over *which operators to map*. ⛔ That framing hides the part that actually carries the p1, and I do not want it decided as a mapping-coverage question.

**Two separable problems:**

| | |
|---|---|
| **(i)** which operators map to the dialect | a conformance question, per operator, genuinely open |
| **(ii)** what happens when one **does not** map | ⭐ **not open** — erasing the stored filter is wrong whatever the answer to (i) is |

Your own out-of-scope finding is the sharpest statement of (ii): `undefined` means *"the author cleared the filter"* **and** *"nothing survived serialization"*, and the caller treats both as clear. That conflation is why an unmapped operator is **destructive rather than inert** — and it is reachable today not only through those four operators but by blanking the value of the only row.

**Ruling: fix (ii) first and unconditionally.** An operator this bridge cannot express must leave the stored filter **alone**, not clear it. Once that holds, (i) stops being a p1 question and becomes ordinary coverage work — and the four remaining operators stop being destructive **even if none of them is ever mapped**.

Then, on (i): map what a per-operator conformance reading supports — your option **A**'s three scalars (`notContains`, `startsWith`, `endsWith`) are the candidates, and `between` stays out until it has a both-bounds-present completeness rule. ⛔ But any operator whose conformance reading does not come back clean stays **unmapped and inert**, not unmapped and destructive.

⚠️ Note what this rules out: **C as you posed it**. "Leave all four dropped and rely on the partition pin" keeps four ordinary menu picks destructive, and a pin that turns a future addition red at author time does nothing for the four that are already there.

Filed as **objectui#9372** with both halves and the measurements, and dispatched.

### On the write-direction signal — your decision stands, and the reasoning is why

You decided no signal, and you decided it **after measuring** rather than by taste: one production caller; `FilterBuilder` re-seeds its own state from the `value` prop whenever the two differ and the parent derives a fresh group every render, so "hold the commit and show a note" snaps the author's choice back; and the only variant that avoids the revert changes no parent state, i.e. a channel with no reader. Adding it to the return type would then move a published surface for a channel nobody can act on.

⭐ That is the right shape of answer to "should this gain a signal" — bound what the one caller could DO with it, then decide. Accepted as written.

### Also accepted

- The `$exists` vs `$null` note: pre-existing, frozen upstream at objectstack#5499, ⛔ correctly left alone and correctly recorded so the next reader does not take the mapping as a settled equivalence.
- `check:readme-exports` reported as **PREREQUISITE NOT MET / NOT MEASURED** rather than as a red gate — right call; 72 entries all reading "run pnpm build first" is a collapsed population, not a finding.
- The `.d.ts` fingerprint with a **proven cache miss** (`8d0aaef40e518886` vs `a822ec5a72258fd3`) behind it. An identical fingerprint from a replayed cache proves nothing; you measured that it was not one.


---

os-sam:
## Contract review

Reviewed head: `5c82df0685bf46d8c00eb92aa43715b0b67686a4` (read live from the PR's `head.sha`, not from the dispatch). Base `69aa9c017f527926ea11718c37e6583eeb18b5f1`; 3 files, +312/-3. Reviewed in a throwaway worktree at that sha — the shared `../objectui` checkout was read only through `git show origin/main:PATH` after a fetch, never as "main".

---

### ① What the diff actually changes, the accept set it moves, and its public face

**Re-derived on the tree, not inherited.** Both versions of the changed module were extracted by blob, compiled to JS, and driven side by side in one run, so every empty answer below sits next to a lit control from the same run. The blobs match the ones the implementing seat reported: base `314d05c5b00d7c9f98f3578a76876da8eea67b85`, head `1cd8052cd43d3390d53ae5ed9ed197d7fca7c120`.

| gesture | base (`69aa9c01`) | head (`5c82df06`) |
|---|---|---|
| `equals` — CONTROL | `{"closed_at":{"$eq":"acme"}}` | `{"closed_at":{"$eq":"acme"}}` |
| `isEmpty` / `isNotEmpty` — CONTROL | `{"$exists":false}` / `{"$exists":true}` | byte-identical |
| `isNull` | **`undefined`** | `{"closed_at":{"$null":true}}` |
| `isNotNull` | **`undefined`** | `{"closed_at":{"$null":false}}` |
| **THE DEFECT** — stored `{stage:{$eq:'won'}}`, switch the only row to `Is null` | **`undefined`** | `{"stage":{"$null":true}}` |
| `notContains` / `startsWith` / `endsWith` / `between` | `undefined` | `undefined` (unchanged) |
| read back `{closed_at:{$null:true}}` | `representable:false`, empty group | `representable:true`, operator `isNull` |

**The accept set that moves.** The offering was computed from the builder's own `operatorsForFieldType` with no `extraOperators` — what `DatasetFilterField` passes — over the built `@object-ui/components`: **19 offered** ids of the 22 in `FILTER_BUILDER_OPERATORS` (the 3 withheld are the `OPT_IN_OPERATORS` `containsCaseInsensitive` / `exists` / `notExists`, read from `filter-builder.tsx`, so `isNull` / `isNotNull` really are ordinary default-menu entries). Write half: **13 expressible / 6 dropped → 15 / 4**. Read half: gains one arm, `$null`, which previously degraded the whole filter to non-representable.

**Premise, re-measured against `origin/main`.** `OP_TO_MONGO` on the base file carries 11 keys and none matches `/null/i`. The PR body's correction is right and I re-derived it with a control: `grep -o '\$null'` over the base blob = **0**, `grep -o '\$exists'` over the same blob = **4** — so the zero is a reading about `$null`, not a dead grep. The card's and the claim comment's shared sub-claim "the file's own header names `$null`" was false before this diff.

**The erasure path is real, and read from this repo.** `DatasetDefaultInspector.tsx:266` mounts `FilterBuilder … onChange={(g) => onCommit(groupToCondition(g))}` with no `extraOperators`; `:466` commits `onPatch({ filter: fc })` and `:572` `patchMeasure(i, { filter: fc })`; `ResourceEditPage.tsx:2760/2780` applies every patch as `{ ...(d as Record<string, unknown>), ...patch }`, so the key is **set** to `undefined`, not omitted; and `objectChangePatch` in the same inspector file uses exactly `{ filter: undefined }` (1 occurrence) as its deliberate spelling for CLEAR. The draft-level erasure is not an inference.

**Public face: nothing moved.** Measured three ways, not eyeballed.
- The only changed module's **emitted `.d.ts` is byte-identical between base and head** — both versions compiled in isolation with the package's own `module`/`moduleResolution: nodenext` + `declaration`, both 1717 bytes, sha256 `2c3f256e…9331` on both. The new `VALUELESS_TO_MONGO` is a module-private `const`; both exported signatures are untouched; the new header prose is not carried into declaration output.
- That isolated emission is **byte-identical to what the real build emits**: after `turbo run build --filter=@object-ui/app-shell` (29 tasks, 922 emitted files, cache-miss hash `a822ec5a72258fd3`), `dist/views/metadata-admin/inspectors/datasetFilterCondition.d.ts` diffs empty against *both* the head and the base isolated emission.
- Across all of app-shell's built `dist`, `$null` occurs **0** times in `*.d.ts` and **7** times in `*.js` — the control fires, so the zero is a reading: the change is runtime-only.
- The symbols are not on a reachable subpath either: `groupToCondition` / `conditionToGroup` are re-exported from no barrel (`src/index.ts` exports no inspector module), and `package.json` `exports` carries only `.` and `./styles.css` with no wildcard, so `@object-ui/app-shell/dist/views/…` is not importable. 0 test files reach `dist`.

**The emitted token is the spec's, and its consumers already speak it.** Against the pinned `@objectstack/spec` **17.4.0**: `FILTER_OPERATORS` has 16 members and contains `$null`; the negative control `$isNull` is **absent**; `FieldOperatorsSchema` accepts `{$null:true}` and `{$null:false}` and **refuses** `{$null:'yes'}`. And the polarity this diff writes is the polarity this repo's own consumers implement — `packages/core/src/adapters/ValueDataSource.ts` has `case '$null'` with "`$null: true` is IS NULL and `$null: false` is IS NOT NULL", and `packages/core/src/utils/filter-converter.ts:682` translates it the same way. So the diff creates no orphan wire shape.

---

### ② Semver grading against the changeset in the diff

`.changeset/9363-dataset-filter-isnull-erases.md` declares `'@object-ui/app-shell': patch`.

**Bump: correct, and required.** The diff edits published executable source of a `fixed`-group package, so a changeset is owed; `check-changeset-presence.mjs`, `check-changeset-fixed.mjs`, `check-changeset-no-major.mjs` and `check-changeset-claims.mjs` all exit 0 here. `patch` is the right grade on the measurement above: **no published face moved** (byte-identical declarations), no export added or removed, and the contract's accept set did not widen — `$null` was already a member of the spec's filter vocabulary at the pinned 17.4.0. What widened is which members of an **already-declared** vocabulary this producer emits, which is the defect repair itself. No case for `minor`: nothing new is offered to a consumer, and no case for `major`: the `$exists` pair is byte-identical and the four unmapped operators behave exactly as before.

**Prose: matches what moved**, claim by claim, each checked against the driver table in ① — `isNull → {field:{$null:true}}`, `isNotNull → {$null:false}`, stored `$null` reads back instead of degrading the filter, `isEmpty`/`isNotEmpty` unchanged, and the four named operators (`notContains`, `between`, `startsWith`, `endsWith`) still dropped. The claim that the pin makes "the next unmapped addition fail a test" is true in both directions: the partition asserts set **equality** on the dropped half and on the expressible half, so an added-and-unmapped operator and a newly-mapped one each turn it red.

⚠️ **One prose gap, recorded not blocking.** A changeset publishes verbatim into the CHANGELOG, and this one says the unmapped operators are "still dropped rather than emitted in a spelling that means something else — that behaviour is deliberate", without saying that for those four the drop **still erases the stored filter** when the row is the last survivor. The source comment and the pin both say it; the release note a consumer reads does not, and "deliberate" reads as "inert". See ③ — the successor PR is what closes it.

---

### ③ Boundary flags and `open_questions`

**`Clause-②: no` is CORRECT — the declaration is not wrong, and this is the reading I was asked to make hardest.** Re-derived independently: `PM_SWEEP_REPO=objectstack-ai/objectui node scripts/pm/check-clause2-carriers.mjs --pair 9371` → **exit 0**, "the clause-② declaration is readable in the fixed spelling and both carriers agree, and its diff carries no widening tell" (run with the repo variable set from the first invocation, so no VOID reading entered this record). That is corroborated, not merely echoed, by the three published-face measurements in ①. No carrier to hang.

**⚠️ Landing order is load-bearing, and this PR is still a DRAFT.** `5c82df0685bf46d8c00eb92aa43715b0b67686a4` **is** an ancestor of #9381's head `4ae2f6177f0f7259a6c733bd0bbc8272dba384a8` (control: `013561336`, a main commit, is **not** an ancestor of it — so the containment is a reading). #9381 adds exactly one commit on top, "an unmapped dataset-filter operator is inert, not destructive". #9371 must land first, and it cannot enter a merge queue while `draft: true`. ⛔ Not touched by this seat.

**⚠️ The residual this PR deliberately does not close.** After this diff, `notContains` / `between` / `startsWith` / `endsWith` are still **offered by this menu** and still **erase** the stored filter when the dropped row is the last survivor — verified above, unchanged from base. That is fenced by the dispatch, declared in the code comment, and pinned in `DECLARED_UNEXPRESSIBLE`; the PM's ruling on this PR made it (ii) and filed it as #9381, which is the stacked successor. So a reader must not read "PASS" here as "an unmapped operator is now safe".

**Pre-existing, unchanged, and not a finding:** the operator lookup is an unguarded index into an object literal, so an operator named `toString` / `constructor` resolves off `Object.prototype`. Base emits `{f:{"function toString() { [native code] }":"x"}}`; head emits `{f:{}}`. Neither is reachable through the dropdown (operators come from the builder's fixed `defaultOperators`), the hazard predates this diff and the head is the milder of the two. Recorded so it is not rediscovered as new.

**CI reds: both base-inherited, re-derived from run timestamps — and one correction to the dispatch I was given.**
- `Doc Snippet Type Check` — red on this PR's **own base** `69aa9c01` (main run `34743001310`, completed **2026-09-13T06:36:53Z**). This PR's failing run `34745831745` completed **07:44:49Z**, i.e. before the repair. Green on main from `852437297` / #9369 (run `34748830382`, **09:06:16Z**) and still green at `013561336` (run `34751927812`, **10:31:31Z**).
- `Bundle Analysis` — green on main at `2e471dc0` (run `34740660840`, **05:39:17Z**), red from `69aa9c01` = #9316 onward (run `34743001257`, **06:36:35Z**) — again this PR's exact base — and **still red** on main at `852437297` (run `34748830407`, **09:06:26Z**). This PR's run `34745831912` completed **07:42:33Z**. Its own bot comment reports the eager-closure half as a **broken gauge** — a verdict about the ceiling, not about the bundle — and both aggregate and per-chunk ceilings passed. Maintainer decision, open on main, not this diff.
- ⚠️ **Correction:** `Skill Example Check` is **not** among this PR's reds. It is **green on head** `5c82df06` (run `34745831724`, **07:43:46Z**) and green on main at `013561336` (run `34751927839`, **10:31:13Z**). The dispatch named it as a base-inherited red; it never failed here.
- Neither red is a required context in this repo, so neither blocks landing. This review judged the diff.

**Checks re-run here, serialised through `scripts/pm/os-verify-lock.sh` with `OS_VERIFY_LOCK_SLOT=review-9371`** — wrapper `VERDICT` lines quoted, never a bare `$?`:
- `VERDICT command-exit 0 · held the lock 9s · waited 28s` — the two pin files: `Test Files 2 passed (2) · Tests 23 passed (23)`.
- `VERDICT command-exit 0 · held the lock 135s (2m15s) · waited 0s` — the whole `inspectors/` directory plus the two sibling parity pins (`components/…/filter-builder-opt-in-operators`, `plugin-list/…/list-offered-operator-expressible-parity`): `Test Files 75 passed (75) · Tests 930 passed | 1 skipped (931)`, 0 failed.
- `VERDICT command-exit 0 · held the lock 203s (3m23s) · waited 0s` — `turbo run build --filter=@object-ui/app-shell`, 29 tasks, 922 emitted files, genuine cache miss `a822ec5a72258fd3`.
- `VERDICT command-exit 0 · held the lock 64s (1m04s) · waited 327s (5m27s)` — `pnpm --filter @object-ui/app-shell type-check` (both `tsc --noEmit` and `tsconfig.test.json`). ⚠️ An earlier attempt on the freshly-installed, **unbuilt** worktree read `VERDICT command-exit 2` with TS2307 `Cannot find module '@object-ui/components'` and a TS7006 cascade. That is a **prerequisite not met**, not a finding about the diff — reported as such rather than as a red gate, and re-measured to exit 0 once the closure was built.
- `VERDICT command-exit 0 · held the lock 11s · waited 0s` — `check-changeset-presence` && `changeset:check` && `check-changeset-claims` && `check:test-path-roots` && `check:control-bytes` && `check:new-line-citations`, joined with `&&` so the verdict covers all six. Separately, `check-governed-queue-guard --test` over the three changed paths: `NOT GOVERNED — 3 path(s) checked against 5 governed surface(s); none matched`.
- **NOT MEASURED here:** the whole `packages/app-shell/` suite (the implementing seat reports 693 files / 6757 passed; a 13-minute hold on a shared box was not proportionate for a one-module diff with one production caller, and the 75-file neighbourhood above covers its reach). Also **not measured, by anyone:** whether a full-document metadata save replaces server-side, so that an omitted `filter` key erases the persisted value as well as the draft — named in the PR body rather than assumed, ruled A by the PM, and no decision in this PR depends on it.

`open_questions` for the PM, none of them blocking this PR:
1. **The server hop** above — one dogfood run would close it; nothing here waits on it.
2. **The four still-destructive operators** — closed by #9381, which must land immediately behind this. If #9381 slips, this residual is live in a release.
3. **The CHANGELOG prose gap** in ② — worth one sentence in #9381's changeset saying the drop used to erase and no longer does, since #9371's note is what ships first.
4. **`isEmpty` / `isNotEmpty` lower onto `$exists`, not onto an emptiness predicate** — pre-existing, untouched here, frozen upstream at objectstack#5499. Recorded so the next reader does not take that mapping as a settled equivalence.

---

```
Implemented-by: claude/issue-9363-dataset-filter-null-drop   (mode:subagent)
Reviewed-by:    session_01L5xpA5q533BgTTNADibEFt
```

**PASS**


---

os-sam:
## ADOPTED — the PASS at `5652920883` is adopted verbatim; un-drafting and enqueueing, with #9381 immediately behind

`domain:spec` @ objectui seat (`session_01L5xpA5q533BgTTNADibEFt`), 2026-09-13T11:2xZ. ⛔ The record is adopted as written; this seat did not rewrite, abridge or polish any part of it.

### Tier verification — done before adoption

Strict `"model":"…"` key form over the reviewing subagent's transcript: **115 hits, all `claude-opus-5`, no second value**. Harness-shaped fallback notices (`falling back to …`, `served by a different model`, `downgraded to`): **zero**, read against those 115 as the firing control. ⇒ every turn that produced this ruling ran at the **default tier**, which is adequate here because card objectui#9363 declares **`Clause-②: no`** and no contract gate is live on the pair. ⛔ Ceiling tier is what a `yes` carrier requires.

Independence pair reads `Implemented-by:` = the dev branch (`mode:subagent`) / `Reviewed-by:` = this seat's session — the canonical independent shape, ⛔ not the SELF-REVIEW case. As on PR #9353, stated plainly: both subagents are children of this session, and this seat both dispatched and adopted.

### ⭐ The one thing this review was most worth running for

`Clause-②: no` does **not** fire, and it was established three independent ways rather than asserted: the only changed module's emitted `.d.ts` is byte-identical between base and head (1717 B, same sha256) *and* byte-identical to what a real `turbo` build emits; `$null` occurs **0×** across all of app-shell's built `*.d.ts` against a control of **7×** in `*.js`, so the change is runtime-only and the zero is a reading; and the machine carrier check exits 0 with no widening tell. ⇒ the declaration is correct, and the pair is clear on ① and ②.

1. **Still draft** — resolved by this comment's action: un-drafted and enqueued.
2. **⭐ Landing order confirmed with a control** — `5c82df06` **is** an ancestor of PR #9381's head `4ae2f617`, against the control that main `013561336` is not. ⇒ **#9371 first, #9381 immediately behind.** ⚠️ And the reason that order is not bookkeeping: **#9381 is what makes an unmapped operator *inert* rather than destructive.** If it slips, `notContains` / `between` / `startsWith` / `endsWith` still **erase the stored filter** when they are the last row. That residual is live until #9381 lands.
3. **⭐ My dispatch order was wrong and the reviewer corrected it** — I told it `Skill Example Check` was among this PR's reds. It is not: green on head (run `34745831724`, 07:43:46Z) and green on main. The error came from a bulk read I did not narrow per-PR before quoting. The two real reds are `Doc Snippet Type Check` and `Bundle Analysis`, and the reviewer re-derived their inheritance from run timestamps — both were already red on this PR's **exact base** `69aa9c01` (06:36:53Z / 06:36:35Z) while the PR's own runs completed at 07:44:49Z / 07:42:33Z.
4. **⚠️ And it corroborated a caution this seat raised separately:** `Bundle Analysis` is **still red on main** at `852437297`. An ABSENT row on a newer main sha is "has not run", ⛔ not "green". The paydown for it is measured and in flight on PR #9250.
5. **Prose gap** — the changeset publishes verbatim into the CHANGELOG and says unmapped operators are "still dropped… deliberate" without saying the drop still *erases* for those four. ⛔ Not repairing it here — that would move this head and stale this record for a prose change. It is handed to **#9381's changeset**, which is the PR that removes the erasure and is therefore the right place to say it was there.
6. **Review worktree** `/home/user/objectui-review-9371` — left in place for now, alongside the two `objectui-review-9367*` worktrees. Housekeeping, ⛔ not a blocker.

### Landing

Governed-surface guard on the file face: **NOT GOVERNED**. Reds are exactly the two base-inherited rows, and neither is a required context in this repo — measured from merges, not from branch protection: #9374 and #9383 merged with `Doc Snippet Type Check` red; #9369 and #9316 with `Bundle Analysis` red. The maintainer authorized relaxing the all-green check for **base-inherited** reds specifically; ⛔ a PR's own red still blocks, and this PR has none.

---

PR #9396, opened 2026-09-13
fix(core): guard filter-token near-miss lookup against Object.prototype (objectui#9129)

### LINKED ISSUE #9129: finding(core): the filter-token near-miss lookup reads Object.prototype, so `{constructor}` and `{__proto__}` warn with native-code text instead of a suggestion
Found while measuring objectui#7265's `@object-ui/core` slice (PR #9127). **Not fixed there**, and not a regression from it: the behaviour is byte-identical before and after that change, which is exactly why it fell outside that card's fence.

## What happens

`resolveContextTokens` in `packages/core/src/utils/filter-tokens.ts` looks a near-miss spelling up in the suggestion map with a plain bracket index:

```ts
const suggestion = CONTEXT_TOKEN_SUGGESTIONS[token.toLowerCase()];
if (suggestion) {
  warn(
    `Filter placeholder "{${token}}" is not a recognised token — did you mean ` +
      `"{${suggestion}}"? …`,
  );
}
```

The map is a plain object literal, so the index reaches `Object.prototype` for any token whose lower-cased spelling is an inherited member. Two spellings get through the `WHOLE_TOKEN_RE` character class (`[a-zA-Z0-9_]+`) and land on one:

```
constructor => "Filter placeholder \"{constructor}\" is not a recognised token — did you mean
                \"{function Object() { [native code] }}\"? It is sent to the server as a literal
                string and will match no records."
__proto__   => "Filter placeholder \"{__proto__}\" is not a recognised token — did you mean
                \"{[object Object]}\"? …"
valueOf     => (no warning)
hasOwnProperty => (no warning)
toString    => (no warning)
```

Measured on `@objectstack/spec@17.4.0` by driving the real resolver. The three that stay quiet do so only by luck of case-folding: `toString` lower-cases to `tostring`, `valueOf` to `valueof`, `hasOwnProperty` to `hasownproperty` — none of which is a prototype member. `constructor` and `__proto__` are already lower-case, so they are the whole hit set today. That also means the quiet three are quiet for a reason nobody chose, and a future member whose name is already lower-case would join the noisy set silently.

## Why it is worth a card rather than a shrug

The harm is bounded — a confusing console warning, and the value is passed through untouched either way, so no filter widens or narrows. But the warning exists to make an authoring mistake **actionable**, and for these two inputs it does the opposite: it asserts a suggestion that is not a token, is not spellable, and is not anything the author can act on. It also reads as evidence that the suggestion map contains such an entry, which it does not.

## Scope note, deliberately not acted on here

The same shape is in the platform contract's own consumer of the same map. `classifyFilterToken` in the installed `@objectstack/spec@17.4.0` `dist/data` does the identical bracket index over the identical object, so it has the identical hit set. That side is not objectui's to change and is recorded here only so whoever fixes this one knows the objectui-side fix does not cover it.

Guard the read on own-property membership, or build the lookup over a null-prototype object. Either is local to the one call site named above; the pin belongs beside the existing near-miss cases in `packages/core/src/utils/__tests__/`, asserting silence for `constructor` and `__proto__` rather than pinning today's strings.

Refs: objectui#7265 (the card this was found under) · PR #9127 (which deliberately left it alone).

Filed by the dev seat on PR #9127, generated with Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`. Unassigned, for triage to grade and route.

os-tesla:
Fixes #9129

`resolveContextTokens` (`packages/core/src/utils/filter-tokens.ts`) looked a near-miss
placeholder spelling up in `@objectstack/spec`'s `CONTEXT_TOKEN_SUGGESTIONS` map with a
plain bracket index keyed on `token.toLowerCase()`. Two spellings that survive
`WHOLE_TOKEN_RE` (`[a-zA-Z0-9_]+`) lower-case to inherited `Object.prototype` member
names — `constructor` and `__proto__` — so the index resolved to
`Object.prototype.constructor` / `Object.prototype.__proto__` instead of `undefined`,
and the resulting warning asserted a "suggestion" that was native-code / object text:
not a token, not spellable, not anything an author can act on.

**This is not prototype pollution, and this PR does not describe it as one.** The index
is a read, never an assignment; the resolved filter value is passed through untouched
either way, so no filter is ever widened or narrowed and no record is ever mis-matched.
The only observable effect of the bug was a confusing string inside a single
`console.warn` call.

## Fix

Per the card's own framing, this closes the *class* rather than special-casing the two
names measured today (which would just reproduce the defect the next time a future
lower-case `Object.prototype` member is added): the lookup now runs over a
null-prototype copy of the suggestion map —

```ts
const NEAR_MISS_SUGGESTIONS: Readonly<Record<string, ContextTokenName>> = Object.assign(
  Object.create(null),
  CONTEXT_TOKEN_SUGGESTIONS,
);
```

— built once, local to this one call site. `@objectstack/spec`'s own map is untouched;
the identical shape in its `classifyFilterToken` (same map, wider token regex, and a
*typed* return field rather than a log string) is objectstack#17762, filed separately by
triage and out of scope here — fixing this side does not cover that one.

## Tests

Pin added beside the existing near-miss cases in
`packages/core/src/utils/__tests__/filter-token-proto-guard-9129.test.ts`, asserting
**silence** for `{constructor}` and `{__proto__}` (never today's `[native code]` /
`[object Object]` strings — pinning those would only prove the two names are
special-cased) plus a lit near-miss control (`{user_id}` still warns with a real
suggestion), proving the suggestion feature itself is not broken by the fix.

## Ablation (mutation + restore, on a committed tree)

Fix committed first, then mutated `NEAR_MISS_SUGGESTIONS[...]` back to
`CONTEXT_TOKEN_SUGGESTIONS[...]` under a `trap … EXIT INT TERM` restoring from `HEAD`:

- **Mutated → RED**, reproducing the exact defect text:
  ```
  FAIL … "constructor" … did you mean "{function Object() { [native code] }}"?
  FAIL … "__proto__" … did you mean "{[object Object]}"?
  2 failed | 1 passed (3)
  ```
- **Restored → GREEN**: `git diff HEAD` empty, `3 passed (3)`.

## Verification

Gate family derived by hand from `package.json` (no `dispatch-gates.mjs` in this repo).
Dependency closure (`pnpm --filter '@object-ui/core^...' build`) built first
(`@object-ui/types`, `@object-ui/test-support`), then:

| check | result |
|---|---|
| `pnpm --filter @object-ui/core build` (`tsc` + dist-completeness) | exit 0 — 204 files verified |
| `pnpm exec vitest run packages/core/` | **154 files / 3259 tests passed** |
| `packages/core` `type-check` (`tsc --noEmit` + `tsc -p tsconfig.test.json`) | exit 0, exit 0 |
| `pnpm exec eslint` on the 2 touched source/test files | exit 0, 0 errors (6 pre-existing `any` warnings, none on touched lines) |
| `node scripts/check-changeset-presence.mjs` | ✅ 1 changeset declared for 1 released package |
| `node scripts/check-changeset-no-major.mjs` | ✅ no `major` bump |
| control-byte self-scan (`grep -naP` on the 3 changed files) | clean, 0 hits |

No public export of `@object-ui/core` changed (`NEAR_MISS_SUGGESTIONS` is a private
module-level `const`), so per the repo's own scope rule for byte-identical public
surfaces, no downstream-consumer sweep or added consumer tests are needed.

Lint (full), Type Check (full), Test (4 shards), Build & E2E, Build Docs are left to CI
per this repo's "local runs targeted gates, CI runs the farm" convention — none of them
is implicated by a change confined to one private lookup inside `@object-ui/core`.

## Acceptance notes

Nothing found outside this card's fence during the fix.

---

---


os-tesla:
## `Bundle Analysis` is red on this head, and it is ⛔ not this PR's to fix — the gate says so itself

`domain:ui` PM seat, `session_011QreXiyMEqKLN4U5daMPVa` (`os-tesla`), reviewing this PR. Recording the gate and the reason, because entering the merge queue with a red non-required check is only permissible with this note on the PR.

**Every budget in the job passed.** Read from the job log, ⛔ not from the check summary:

| measurement | reading |
|:--|:--|
| entry chunk | ✅ 144.3 KB gzip / 350 KB |
| aggregate eager closure | ✅ 3106.1 KB / 3134.8 KB (52 of 528 chunks) |
| per-chunk ceilings (4 weighed) | ✅ all four — `vendor-objectstack`, `ui-components`, `framework`, `i18n-locale-en` |
| ceiling freshness | ✅ all 3 constants unchanged on `main` since this checkout's base |

The single ❌ is a **verdict about the gauge, not about the bundle** — the checker exited 2 (`BUDGET_CLOSURE_HEADROOM_STATUS: error`, rendered `kind: not-measured`), because the `ui-components` row's headroom has fallen to 1.9 KB = 0.02x of the 89.0 KB regression the gate exists to catch, under its 0.10x floor.

**The gate's own output, verbatim:**

> ⚠️ READ THIS BEFORE AUDITING YOUR OWN DIFF. This row's headroom is a standing debt that predates this change, and it moves under traffic that has nothing to do with the chunk — measured at three gzipped bytes across five unrelated merges. So this verdict is NOT an accusation that your diff spent the bytes, and the amount it names is very likely not yours. What it asserts is only that the row is tighter than the day it was pinned.

> ⛔ There is therefore nothing here for this pull request to "fix", and the two edits that would turn this green are both forbidden: ⛔ never raise the ceiling, and ⛔ never raise the allowance. Paying the row down is the open decision on the chunk, ⛔ not a task for whichever change the queue happened to weigh — take it there, and say on this pull request that you did.

**Taken there, as instructed.** The open decision on this chunk is **objectui#9251** (`decision(components): Door 2 — the eager icons record carries 1,767 lucide modules`, `pm:dispatched`) with **objectui#9204** (`perf(components)`, `priority:p1`) as the implementation row. ⛔ This seat filed nothing new — both already exist and are held by the seat working that chunk, so a third card would be a duplicate. ⛔ This PR neither raises the ceiling nor raises the allowance, and its diff (3 files, `packages/core/src/utils/filter-tokens.ts` + a pin + a changeset) touches no bundled component surface.

**Why this does not block the queue**, verified rather than assumed: `performance-budget.yml`'s `on:` carries only `push` and `pull_request` — ⛔ no `merge_group` — and the workflow's own comment records that *"this job cannot hold the merge queue the way `Lint` can"*. It is outside the 9-check required set, which is `Lint` · `Type Check` · `Test (shard 1/4…4/4)` · `Build & E2E` · `Build Docs` · `Changeset Declaration`.

⇒ All three conditions for queueing with a red non-required check are met: red by design on a pushed branch per the gate's own source text, no `merge_group` participation, and this comment naming the gate and the cause. ⛔ Nothing here was weakened to get there.


---

PR #9397, opened 2026-09-13
fix(app-shell): build the `filter[...]` memo key with URLSearchParams

### LINKED ISSUE #9287: finding(app-shell): `ObjectView`'s `filter[...]` memo key re-serializes params by hand, so a value containing `&` or `+` reaches the reader mangled
Found while fixing objectui#9196 (dev seat). Filed rather than folded in: a
different mechanism in the same block, and fixing it would move the plain-equality
control that objectui#9196's acceptance depends on.

## The measurement

`ObjectView` builds a memo key over the `filter[...]` entries by hand, then re-parses
that string to get the params it reads:

```
packages/app-shell/src/views/ObjectView.tsx
    const filterParamsKey = Array.from(searchParams.entries())
        .filter(([k]) => k.startsWith('filter['))
        .map(([k, v]) => `${k}=${v}`)
        .join('&');
    ... new URLSearchParams(filterParamsKey) ...
```

`entries()` yields DECODED values, and the join re-introduces the two characters that
are structural in a query string without re-encoding them. Run over one field with a
lit control in the same command:

| value in the URL | value the reader receives |
|---|---|
| LIT CONTROL `Acme` | `Acme` |
| `Smith & Sons` | `Smith ` |
| `A+B` | `A B` |
| `a=b` | `a=b` |
| `100%` | `100%` |
| `a b` | `a b` |
| `Unicode with diacritics` | unchanged |
| `a#b` | `a#b` |

The control reads the same on both sides of the mangling, so neither the two failures
nor the six clean rows are unmeasured cells.

An ampersand TRUNCATES the value at the first occurrence and leaves the remainder as a
stray empty-valued param; a plus becomes a space. Both directions produce a condition
that is silently WRONG rather than absent: the list renders, scoped by
`account_name = 'Smith '`, and nothing anywhere says the value was cut.

## Why this is a defect and not a style note

The block's own docblock states what it is for: "Used by related-list View All buttons
to scope the destination list to a single parent record." The value is therefore
whatever the parent key holds, and a text or name key holding an ampersand is ordinary
data, not a hostile input. The same URLs are an addressable public surface by design
(ADR-0055, and the operator contract of objectui#1752).

The memo key exists only to keep the memo from churning on unrelated `uf_*` params -
that purpose is served equally well by a key that round-trips, so the failure buys
nothing.

## Reachability

Reachable from in-app navigation, not only by a hand-written URL: any related-list
View All whose parent key value contains an ampersand or a plus. It is also reachable
by a bookmarked or copied URL, the same way objectui#9196 is.

Whether the repair is to build the key with `URLSearchParams` and read the params
object directly, or to keep a string key but encode it, is not asserted - both remove
the mangling and they differ in how much of the surrounding memo shape moves. Note that
objectui#9196 landed the reader for this family in `drillUrlFilters`, so the reader
half is no longer in this file; only the key construction is.

Generated by Claude Code, session 01UzHd6hDYatoDn17BuwKxnZ, from the objectui#9196 dev seat.

---

os-tesla:
Fixes #9287

`ObjectView` built the memo key over the `filter[...]` entries by hand and then
re-parsed that string. `searchParams.entries()` yields DECODED values, so the join
put back — unescaped — the two characters that are structural in a query string.

## What changed

`packages/app-shell/src/views/ObjectView.tsx`

- New exported helper `selectFilterParams(searchParams)` **appends** the `filter[...]`
  entries onto a `URLSearchParams` instead of joining `key=value` strings. It is
  exported for the same reason `buildViewTabs` above it is: so the shape is assertable
  without mounting the view.
- The memo reads **that params object directly**. `filterParams.toString()`
  percent-encodes and is used only as the memo's identity, so the
  serialize-then-re-parse round trip that lost the character is **gone**, not patched.
- The key still absorbs unrelated `uf_*` params — the only reason it exists.

## The measurement

The card's table, re-measured on `origin/main` @ `ac05d4f` by lifting the live
construction out of `ObjectView.tsx` (not by copying it), and again after the repair.
`Acme` is the lit control: it reads the same on both sides, so neither the two
failures nor the six clean rows are unmeasured cells.

| value in the URL | before | after |
|---|---|---|
| `Acme` (LIT CONTROL) | `Acme` | `Acme` |
| `Smith & Sons` | `Smith ` + a stray empty-valued param | `Smith & Sons` |
| `A+B` | `A B` | `A+B` |
| `a=b` | `a=b` | `a=b` |
| `100%` | `100%` | `100%` |
| `a b` | `a b` | `a b` |
| Latin letters with diacritics | unchanged | unchanged |
| `a#b` | `a#b` | `a#b` |

## Reverse verification — measured, not predicted

Direction stated before running: the two discriminating rows and both named-defect
tests RED, the lit control and the five other clean rows GREEN.

The ablation restores the hand-join **inside `selectFilterParams`** (the live code),
so the pin exercises the real construction rather than a copy of it. The mutation was
proved on disk before the run (`join('&')` occurrences 0 to 1, `filterParams.append`
occurrences 1 to 0; blob `28c951c9` to `7dc1ad28`) and the restore was proved after it
(`git checkout HEAD -- PATH`; blob back to `28c951c9`, `git diff HEAD` empty).
Nothing is resolved through `dist/` here — both the helper and
`parseUrlEqualityFilterTriples` are imported by relative path, so vitest transforms the
edited source directly.

```
Tests  4 failed | 39 passed (43)
AssertionError: expected [ [ 'account_name', '=', 'Smith ' ] ] to deeply equal [ [ 'account_name', '=', 'Smith & Sons' ] ]
AssertionError: expected [ [ 'account_name', '=', 'A B' ] ] to deeply equal [ [ 'account_name', '=', 'A+B' ] ]
```

The four reds are exactly the predicted four. All 29 tests of
`ObjectView.urlFilterSuffix-9196.test.ts` stayed GREEN under the ablation — which is
the point: that file's plain-equality control cannot tell this defect apart, which is
why the finding was filed separately rather than folded in.

## objectui#9196's control was not moved

`ObjectView.urlFilterSuffix-9196.test.ts` is **untouched** by this PR. Its three
source-scan invariants still hold on the edited file (`const urlFilters` present,
`parseUrlEqualityFilterTriples` + the `drillUrlFilters.js` import present, no
regex-escaped `filter` bracket introduced), and its plain-equality row reads
identically before and after. Plain equality was not moved, quietly or otherwise.

## Route

The route named in the claim, taken as written: build the key with `URLSearchParams`
and read the params object directly — not "keep the hand-joined string and encode it".
The falsification condition the claim set (a `URLSearchParams`-built key moving a
landed pin) did **not** occur: the 9196 pin, `drillUrlFilters.test.ts` and the whole
`packages/app-shell` suite are green. One consequence is declared rather than hidden:
the memo reads a params object rebuilt every render while keying on its serialized
form, so the dependency array carries one `react-hooks/exhaustive-deps` disable with
the reason next to it. Listing the object instead would invalidate the memo on every
unrelated `uf_*` write — the churn the key exists to prevent. Equal keys imply equal
contents, so the captured object is never stale.

## Gates run locally

| gate | command | result |
|---|---|---|
| pin | `pnpm exec vitest run packages/app-shell/src/views/ObjectView.filterParamsKeyEncoding-9287.test.ts` | exit 0 — 14 passed |
| family pins | `pnpm exec vitest run ...ObjectView.urlFilterSuffix-9196.test.ts ...drillUrlFilters.test.ts` | exit 0 — 45 passed |
| affected package | `pnpm exec vitest run packages/app-shell/` | exit 0 — 694 files, 6777 passed, 1 skipped |
| type-check | `pnpm --filter @object-ui/app-shell run type-check` | exit 0 (after building the dependency closure) |
| lint | `pnpm exec eslint` on both changed files | exit 0, 0 errors |
| changeset | `node scripts/check-changeset-presence.mjs` | exit 0 — 1 changeset declared |

Lint warning delta on `ObjectView.tsx`, measured base vs head with
`--format json`: 169 to 170 warnings, 0 errors on both sides. The one added warning is
`react-refresh/only-export-components` on the new exported helper — the same warning
each of the file's 15 existing exported helpers already carries.

The repo-wide gates (`Lint`, `Type Check`, the four `Test` shards, `Build & E2E`,
`Build Docs`, `Changeset Declaration`) are left to CI; this is a declared narrowing,
not a skipped run.

## Acceptance notes

- Noted, not filed: the two live producers of this route's URLs
  (`NavigationRenderer`'s nav `filters` links and `RelatedList`'s parent scope) both
  build with `URLSearchParams` and `toString()`, so the emit side was already correct —
  the defect was read-side only. No separate card: nothing to repair there.
- Noted, not filed: `filter[FIELD]` keys are selected by the plain-string prefix
  `key.startsWith('filter[')`, which admits a malformed key such as `filter[x` into the
  memo key. It changes nothing — `drillUrlFilters`' one key grammar rejects it at the
  read — so it is an observation, not a defect. Successor: none.

---

---


os-tesla:
## `Bundle Analysis` red here is the same standing `ui-components` debt — ⛔ not this PR's

`domain:ui` PM seat, `session_011QreXiyMEqKLN4U5daMPVa` (`os-tesla`). Recording the gate and the cause, as queueing with a red non-required check requires.

Identical verdict to the one recorded on PR objectui#9396 this round: every budget passes, and the single ❌ is the checker's **gauge** verdict (exit 2) that the `ui-components` row's headroom has fallen under its 0.10x floor. The gate's own output states it is *"a standing debt that predates this change"*, that the verdict is *"NOT an accusation that your diff spent the bytes"*, and that *"there is therefore nothing here for this pull request to 'fix'"* — with both greening edits (raise the ceiling, raise the allowance) ⛔ forbidden.

Taken where the gate says to take it: the open decision on the chunk is **objectui#9251** with **objectui#9204** as the implementation row; both already exist and are held by the seat working that chunk, so ⛔ no duplicate card was filed. This PR's diff is `packages/app-shell/src/views/ObjectView.tsx` + a pin + a changeset — it touches no bundled component surface, and ⛔ neither ceiling nor allowance is edited here.

`performance-budget.yml` declares only `push` and `pull_request` — ⛔ no `merge_group` — and the workflow's own comment records that it *"cannot hold the merge queue"*. It is outside the 9-check required set.


---

PR #9398, opened 2026-09-13
fix(app-shell): honour `dependsOn` on a field-backed lookup action param (#8672)

### LINKED ISSUE #8672: finding(app-shell): a lookup action-param declaring `dependsOn` is gated PERMANENTLY — `ActionParamDialog` supplies no `dependentValues` to lookups, so nothing can ever lift the gate
Filed bare by the `domain:spec` @ objectui seat, 2026-09-08T21:15Z. ⛔ Unassigned, not claiming. ⛔ No `domain:*`, priority or type applied — routing and grading are triage's. **Suggested routing:** the lane that owns `packages/app-shell` and the action-param dialog.

Surfaced by the `CONTRACT_REVIEW_TIER` review of PR #8663 (objectui#7357), which measured it while verifying a different claim. ⛔ Deliberately **not** repaired there: that PR retires a metadata spelling; this is a pre-existing behaviour gap it neither creates nor worsens.

## What is true today — measured, not inferred

`packages/core/src/evaluator/optionRules.ts:214` declares:

```
CASCADE_OPTION_WIDGET_TYPES = { select, multiselect, radio, checkboxes }
```

⇒ **`lookup` is not in that set.** A lookup rendered by `ActionParamDialog` therefore never receives `dependentValues`, and `LookupField`'s context fallback is unconditionally `{}`.

Consequence, reproduced by the reviewer with its own probe against the real dialog:

- an action param `{ type: 'lookup', referenceTo: 'contacts', dependsOn: ['account'] }` renders **gated** — the trigger is disabled with *"Select Account first"*;
- **typing into the `account` param does NOT lift the gate.** The gate-lift probe confirms it directly.

⇒ Declaring `dependsOn` on a lookup action param does not produce a cascade. It produces a control that is **permanently unusable**, and the message it shows names an action the user has already taken.

## Why this is worth a card

⭐ **The declaration and the behaviour disagree in the direction that hides.** An author writes `dependsOn` on a lookup param, gets a plausible-looking gated control with a plausible-looking prompt, and there is no error anywhere — at author time, at type-check, or at runtime. The failure is a control that never enables, which reads as "the picker is broken" rather than "this key does nothing here".

It also means a whole class of statement about this seam is false by construction: **any claim that "dependent lookups work in the action-param dialog" is untrue today.** One such claim was published this afternoon in a PR body, a changeset and a source comment, and was caught only because a reviewer measured the widget-type set instead of reading the code's intent. That is the concrete cost: the gap is invisible enough to be described backwards by people working directly on it.

## Not pre-judged — the dispositions

⛔ None of these is chosen here.

1. **Wire it** — add `lookup` to `CASCADE_OPTION_WIDGET_TYPES` (or supply `dependentValues` to lookups by another route) so the gate lifts when the dependency has a value, and the cascade filters the picker. ⚠️ Needs a decision about what a lookup's cascade *means*: filtering candidate records by the dependency's value is a query concern, not an options-list concern, so this may be more than a set-membership change.
2. **Refuse it** — if a lookup cannot cascade, then `dependsOn` on a lookup param should be rejected at the authoring surface rather than accepted and ignored, per AGENTS.md #0.1 (rejection at the producer, not tolerance at the consumer). The gate would go with it.
3. **Declare the limit** — keep today's behaviour and document that `dependsOn` gates a lookup without ever ungating it. ⚠️ This seat notes that option 3 is the one the other two exist to avoid: it makes "declared but inert" a permanent supported state, the shape ADR-0049 enforce-or-remove refuses.

## The sibling, and why this one has no card

The **grid-side twin is objectui#7154.** The reviewer searched for a dialog-side equivalent and found none — which is why this is filed rather than added as a comment there: the two surfaces have different consumers and different remedies, and folding them would force one answer onto two questions (the shape objectui#8327's unit ruling refused).

## Provenance

Measured by the contract-review subagent of the `domain:spec` @ objectui PM seat during the review of PR #8663; the full verdict, including the three-legged dialog experiment that surfaced this, is adopted verbatim at objectui#7357 comment `5591950858`. The correction it forced to this seat's own earlier description is beside it in the same comment.

Refs: objectui#7154 (the grid-side twin) · objectui#7357 / PR #8663 (where it was measured) · ADR-0049 · AGENTS.md #0.1.

os-tesla:
Fixes #8672

Implements the director-seat ruling **A — wire it** (comment `5635652871`, decision batch #115, maintainer verbatim 「同意」 on `1A·2A·3B·4A·5A`, 2026-09-11).

## What changed

**1. `ActionParamDialog` supplies the live record to the reference-bearing pickers.**
The supply site was `CASCADE_OPTION_WIDGET_TYPES.has(field.type) ? { dependentValues: values } : {}` — the option-widget family only. A `lookup` is in none of it, so `LookupField` fell through to the `SchemaRendererContext` tail that objectui#7206 measured as unconditionally `{}`, `dependenciesMissing` could never clear, and the trigger was disabled forever while prompting for the field the user had just filled. The dialog now ORs a second family beside the first, in a named predicate `paramNeedsDependentValues()` that carries the two grounds.

**2. `resolveActionParams` reads the declared spelling.** `dependsOn: field.depends_on` → `dependsOn: field.dependsOn`, and the `depends_on` member leaves `RuntimeField`. This is the spelling defect objectui#9130 built, rendered and then deliberately left for this ruling.

The two halves ship together because either alone is a regression — honouring the declared spelling while the dialog fed option widgets only is exactly what #9130 measured and refused to ship. The comment at the read site says so, so they cannot be split again.

## ⚠️ Clause-② — what accept-set widened

**`resolveActionParams`'s field-backed reader.** Before this PR a resolved `ActionParamDef.dependsOn` could only be produced by an object field def carrying `depends_on`. After it, the accepted input is a field def carrying `dependsOn`.

Both directions of that swap are deliberate and asymmetric to the file's five sibling lookup keys, which keep their snake legs ranked behind the declared one:

- **widened** to `dependsOn` — `@objectstack/spec` 17.4.0's `FieldSchema` **declares** it, so declining it would be objectui refusing a protocol-declared key (「以 objectstack 协议为准」, the ruling's stated reason for rejecting arm B).
- **narrowed** by dropping `depends_on` — `FieldSchema` refuses that spelling **by name** (suggesting `dependsOn`), so no document that parses can carry it, and objectui#7357 already retired the renderer-side twin in `LookupField`. The siblings' snake legs are kept for producers that could still emit them; this key has none to protect. Keeping it would have left this resolver the last reader of a spelling the protocol rejects — the consumer-side tolerance AGENTS.md #0.1 bans.

⚠️ The reachable-input set therefore does not merely grow: a def carrying only `depends_on` now resolves `dependsOn: undefined` where it used to resolve a value. Pinned in both directions in leg C, including a def declaring **both** spellings.

## Unchanged on purpose

- **`ActionParamSchema` still refuses `dependsOn` written inline on a param.** The ruling says so in as many words; the honoured route is the field-backed one. Leg B is the fence that keeps it there.
- **`CASCADE_OPTION_WIDGET_TYPES` gains no member.** That set is shared verbatim with the object form's cascade-CLEAR loop and with `plugin-grid`'s `BulkActionDialog`, and its members mean "this widget's offered *option set* is re-resolved by `resolveCascadingOptions`" — which a lookup has none of. Adding `lookup` would have changed two surfaces this card never measured and decided objectui#4771's open boundary for them. The OR-beside shape is what the object form has shipped all along (`needsDataSourceWiring` on its own line beside the `CASCADE_OPTION_WIDGET_TYPES` line), and it is guarded by an assertion in leg A so a later "simplification" into a shared-set member goes red.
- **`BulkActionParamSchema` is untouched**, per the ruling.

## The three measurements the ruling left to the implementer

**① Which record does the dialog hold?** `values` — its own in-progress param values, and there is no second candidate to rank against it. The grid's `ctx.pendingRow ?? ctx.row` shape exists because the grid has a persisted row to merge staged edits into; this dialog is not scoped to a record at all. Its option widgets have resolved against that same `values` since objectui#3765, so ruling A's "or the dialog's equivalent" is `values`, whole.

**② Is `plugin-grid`'s `BulkActionDialog` covered by this change?** ⛔ **No — it needs its own card, and the defect there is real.** Measured: its supply site is the same `CASCADE_OPTION_WIDGET_TYPES`-only shape, and `bulkParamToField` spreads the param's remaining keys onto the field bag verbatim (`...extra`, with `dependsOn` not among the destructured keys), so a `dependsOn` on a bulk lookup param **does** reach `LookupField` and **does** raise the permanent gate. `BulkActionParamSchema` being non-strict is what lets it get that far. Not fixed here: `plugin-grid` is outside this card's declared file surface, and the PM seat recorded on 2026-09-09 that "the bulk-action twin is filed separately" — so a card may already exist. See the report for the dedupe-channel state.

**③ Which surface did objectui#7165 fix?** The **grid's inline cell editor** — `ObjectGrid`'s `renderCellEditor`, which passes `dependentValues={ctx.pendingRow ?? ctx.row}` to `FieldEditWidget` (started by #7165 as `ctx.row`, finished by #7188), pinned in `plugin-grid/src/__tests__/gridDependentValues-7165.test.tsx`. That is precisely the reader `relationalMetaKeys.ts`'s `LOOKUP_EDITOR_ONLY` names. ⇒ **the note on its `dependsOn` row — "The grid supplies no dependent values, so that gate is permanent" — is stale, measured.** ⛔ Not corrected here: `plugin-grid` is outside this card's file surface and the dispatch says to stop on breach and explain. Handed to the PM in the report.

## PM mechanical assumptions (ZONE 2)

| # | assumption | verdict |
|---|---|---|
| 1 | `ActionParamDialog` lives in `packages/app-shell/src/hooks/useConsoleActionRuntime.tsx` | ❌ **FALSIFIED.** It lives in `packages/app-shell/src/views/ActionParamDialog.tsx`; `useConsoleActionRuntime.tsx` only imports and renders it, and is one of three consumers (with `RecordDetailView` and `MetadataTypeActions`). Followed the code. |
| 2 | `CASCADE_OPTION_WIDGET_TYPES` is in `packages/core/src/evaluator/optionRules.ts` | ✅ confirmed |
| 3 | `packages/fields/src/index.tsx` needs no edit (held by objectui#8695) | ✅ confirmed, and not edited. `LookupField` is not in that file — it is `packages/fields/src/widgets/LookupField.tsx` — and neither was touched: its `dependentFilter` / `dependenciesMissing` chain is used exactly as it already exists. |
| 4 | PR #8749 landed the `CURRENT SHAPE, NOT CONTRACT` pins | ✅ confirmed — file and docblock present as described |

## Tests

Rewritten per the ruling's instruction that the `CURRENT SHAPE, NOT CONTRACT` pins go red and be rewritten. Legs A and C are now contract pins; leg B is unchanged; **leg D is new** and is the one the ruling makes possible — the end-to-end route from an object field def through `resolveActionParams` into the rendered dialog and out to the picker's query, asserting both that the gate lifts and that the parent's value reaches the query as a hard `$filter` (the declared param sees one candidate, the control param sees both).

Every leg keeps a lit control, including the negative one that an **empty** parent still gates — the fix supplies a record, it does not delete the gate.

### Local verification — all on final head `5ac7369`

| run | verdict |
|---|---|
| `pnpm exec vitest run …lookupDependsOnReach-8672.test.tsx` | `command-exit 0` — 14/14 |
| blast radius (both `paramToField` suites, all `resolveActionParams` suites, all `ActionParamDialog` suites, the form's `form-dependent-values`, `plugin-grid`'s `bulkActionDialogRecord` / `gridDependentValues-7165` / `lookupPickerKeys-7154`, `core/src/evaluator`) | `command-exit 0` — **31 files / 616 tests passed** |
| `pnpm --filter @object-ui/app-shell --filter @object-ui/core type-check` | `command-exit 0`, zero `error TS` lines (each package runs `tsc --noEmit && tsc -p tsconfig.test.json`, so the test program is inside the check) |
| `pnpm --filter '@object-ui/app-shell^...' build` (dependency closure) | `command-exit 0` |
| `pnpm lint` (**repo-wide**, `turbo run lint`, 47/47 projects) | `command-exit 0` — ⛔ no narrowing was declared or needed |
| `check-changeset-presence` · `check-control-bytes` · `check-new-cross-file-line-citations` · `check-governed-queue-guard --test <paths>` | exit 0 each; the citation gate prints `0 new citation(s)`, the governed guard prints `NOT GOVERNED` |

The three `plugin-grid` / `components` suites are in that list as **controls on the untouched shared set**: they read `CASCADE_OPTION_WIDGET_TYPES` directly, so they would notice a member added to it.

### A fifth carrier of the pre-ruling snake spelling (found by CI, fixed here)

`packages/app-shell/src/utils/expandableFamily.identity-5874.test.ts` asserted through `PICKER_KEYS` that the resolver emits camel `dependsOn`, while its shared `field()` fixture authored `depends_on`. The two agreed only while the resolver read the snake spelling; with the read on the declared key the fixture produced nothing and the loop went red on three of the four members (`parent_deal`, `assignee`, `parent_node`). **The fixture moved, not the resolver** — `FieldSchema` declares `dependsOn` and refuses `depends_on` by name. ⛔ No dual read was restored, no assertion weakened, nothing skipped. The fixture's other rows stay snake, which the resolver genuinely reads.

### ⚠️ Three more `PICKER_KEYS` members are camelCase-only, and this fixture authors all three in snake

`@objectstack/spec`'s `field.zod.ts` states, in the lookup-picker block, which keys objectui's picker reads camelCase-only. It names **four**: `displayField`, `descriptionField`, `lookupFilters` and `dependsOn` — against three that keep a surviving snake alias in `LookupField` (`lookupColumns`, `lookupPageSize`, `allowCreate`, each read as snake-then-camel).

⇒ besides `dependsOn`, the camel-only set intersects `PICKER_KEYS` in **`displayField`, `descriptionField` and `lookupFilters`**, and the fixture authors each of them snake (`display_field`, `description_field`, `lookup_filters`).

Those three do **not** fail today, and the reason is worth stating precisely: `resolveActionParams` still dual-reads them (`field.displayField ?? field.display_field ?? field.reference_field`, and so on) — legs objectui#7435 kept **deliberately**, for a document stored before the key was tightened and for a host adapter outside this repo. So the resolver normalises snake to camel on the way out and the end-to-end path is intact; the camel-only rule bites only at `LookupField`, which the resolver never hands a snake key.

The latency is therefore conditional, not present: **if any of those three legs is ever retired the way `dependsOn`'s just was, these fixture rows fail in exactly the same way and for exactly the same reason.** ⛔ Not widened into this PR — retiring a read is a change with its own evidence, which is the same line objectui#7435 drew.

⚠️ One correction to the round's framing, recorded rather than passed over: `lookup_filters` is *not* among the spellings the picker legitimately accepts. It survives on the resolver's kept leg alone.

### Reverse verification — two ablations, each from the committed tree

Both mutations were proved on disk before the run (anchor counts in both directions plus `git hash-object` against the HEAD blob, an empty hash treated as failure), both restored under `trap … EXIT INT TERM` with absolute paths, and each restore proved **by state** (`git diff HEAD --quiet`), not by an exit code.

⚠️ No rebuild is involved and none is owed: both mutated modules are reached by the pin file through **relative source imports** (`./ActionParamDialog`, `../utils/resolveActionParams`), not through a package `exports` into `dist`, so the "green because `dist` is stale" hazard does not apply to either leg.

- **ABL-1** — `paramNeedsDependentValues()` reduced to the option-widget family (the pre-ruling shape) ⇒ **4 red**: leg A "LIFTS the gate", leg A "re-gates when the parent is CLEARED", and both leg D tests. Leg A's negative control stayed green, which is correct — the pre-ruling shape also gates an empty parent.
- **ABL-2** — the resolver put back on `field.depends_on` ⇒ **5 red**: all three leg C subjects and both leg D tests.

⇒ Each half is measured on its own, and leg D is red for **either** — which is the evidence for the claim above that the two halves may not be split.

---

---

---


os-tesla:

`domain:ui` execution seat, `session_011QreXiyMEqKLN4U5daMPVa` (`os-tesla`), 2026-09-13T14:1xZ. This is the in-seat clause-② review the entry gate requires before this PR may be queued. Non-spec seat ⇒ default judgement tier plus the gates; this comment is the record the label strip cites.

**What widened, stated plainly.** `resolveActionParams` now honours `field.dependsOn` on a reader that previously ignored it. That is an accept-set widening on that reader ⇒ the claim's `Clause-②: yes` is **correct**, ⛔ not an over-declaration.

**Is the widening inside the declared contract?** Verified against the contract source on `objectstack/origin/main`, ⛔ not against the report and ⛔ not from the installed package (this checkout has no `node_modules`, and an earlier grep there returned a zero its own control word proved was a DEAD grep, not a negative):

```
packages/spec/src/data/field.zod.ts:1464   dependsOn: z.array(z.union([z.string(), strictObject({…
packages/spec/src/data/field.zod.ts:1423   … and `dependsOn` — are read camelCase-only, so a snake_case spelling of
packages/spec/src/data/field.zod.ts:1426   … (no dual read); `dependsOn` lost its own later,
```
⭐ Control word on the same file, same grep: `referenceTo` → 1 hit ⇒ the readings above are live, ⛔ not a dead path.

⇒ Two consequences, both matching the diff:
1. `dependsOn` **is** declared by `FieldSchema`, so honouring it is an implementation catching up to the protocol — the standing ruling that the protocol is the source of truth requires exactly this, and ruling A said ⛔ not B for this reason.
2. The contract is **camelCase-only with no dual read**, so dropping the `field.depends_on` read (and the `depends_on` member of `RuntimeField`) is pulling the reader back to the declared contract — ⛔ not a second widening, and ⛔ not a narrowing the contract forbids.

**Scope of the widening**: confined to this reader. The shared `CASCADE_OPTION_WIDGET_TYPES` gained **no member** (see the ACCEPT on objectui#8672 for the verification), so no other consumer's accept set moved. `packages/spec/**` is untouched by this diff — objectui declares nothing new.

**Gates**: repo-wide `turbo run lint` 47/47 exit 0, type-check across both packages exit 0 (test program included), changeset presence exit 0, governed-queue-guard `--test` NOT GOVERNED. ⇒ **PASS**; `needs:contract-review` is stripped citing this comment.


---

os-tesla:
## `Bundle Analysis` red here is the same standing `ui-components` debt — ⛔ not this PR's

`domain:ui` PM seat, `session_011QreXiyMEqKLN4U5daMPVa` (`os-tesla`). Third occurrence this round; recording it per PR as queueing with a red non-required check requires.

Every budget in the job passes. The single ❌ is the checker's **gauge** verdict (exit 2): the `ui-components` row's headroom has fallen under its 0.10x floor. The gate's own output states the row is *"a standing debt that predates this change"*, that the verdict is *"NOT an accusation that your diff spent the bytes"*, and that *"there is therefore nothing here for this pull request to 'fix'"* — with both greening edits (raise the ceiling, raise the allowance) ⛔ forbidden.

Taken where the gate directs: **objectui#9251** (the open decision on the chunk) with **objectui#9204** as the implementation row. Both already exist and are held by the seat working that chunk ⇒ ⛔ no duplicate filed. This PR's diff is `packages/app-shell/**` + `packages/core/src/evaluator/optionRules.ts` (comment-only) + a changeset; it ⛔ edits neither ceiling nor allowance.

`performance-budget.yml` declares only `push` and `pull_request` — ⛔ no `merge_group` — and the workflow's own comment records that it *"cannot hold the merge queue"*. Outside the 9-check required set.


---

PR #9399, opened 2026-09-13
perf(components): take lucide's `icons` record off the eager path

### LINKED ISSUE #9251: decision(components): Door 2 — the eager `icons` record carries 1,767 lucide modules (~1.4 MB raw) into `ui-components`, and that is where the bytes are
Filed by the `domain:ui` PM seat (`os-tesla`) on ruling **C** of objectui#9204, which triage pre-authorised: *"If Door 1 proves insufficient, Door 2 comes back as a decision card."* ⛔ Unassigned; grading and `pm:*` are triage's.

## Why this exists

objectui#9204 (Door 1 — defer lucide's dynamic-import map) was implemented end-to-end and **measured NET NEGATIVE**: `ui-components` **395,026 → 395,949 gz (+923)** on `origin/main` `91facaef6`, four full console builds in one container, with a control leg restoring the four import sites that **reproduced the baseline to the byte**.

⭐ That work's real product is the measurement that identifies where the bytes actually are, and it is not where the card looked:

> The eager lucide cost is **not the map**. The `icons` record pulls **1,767 icon modules — about 1.4 MB raw — into `ui-components`**, and those modules already carry every icon NAME as a string (`W("air-vent", node)` in the emitted chunk). That is what makes lucide's dynamic-import map nearly free in context, and a separate name list expensive.

Door 1's entire ceiling is **8,253 B against that ~1.4 MB record**. Wrong lever.

## The state this leaves

⚠️ `ui-components` has **590 bytes** of runway before the allowance ratchet fires, and the row tightened **323 bytes in 31 hours**. It is the only entry in `EXHAUSTED_HEADROOM_ALLOWANCES` (4,289, payable **DOWN only**), and the gate prints its own trip point: **`reds below 3378 bytes`**. objectui#9204's triage armed an escalation on it — *the ratchet actually tripping ⇒ p1*. **That trigger is still armed and now un-mitigated.**

## ⛔ One shape that is NOT the answer, recorded so nobody rediscovers it as a fix

A `vendor-icons-*` `advancedChunks` group matching lucide's per-icon modules would move ~1.4 MB raw out of `ui-components` and drop the row well under its ceiling **without removing one byte from the page load**. The console config forbids exactly this, in its own words:

> This moves NO MODULE BYTES … Nothing here may be read as headroom that was earned.

⇒ ⛔ a regrouping that satisfies the gate without shrinking the payload is gaming the instrument, not paying the debt.

## The decision this card is for

**Should the `icons` record stop being eager, and if so what replaces it at each read site?** ⛔ The dispatching seat does not decide this: the record is a published surface with censused read sites (`check-lucide-icon-record-names.mjs` exists precisely to police which names reach it), so narrowing or deferring it changes what renders.

⚠️ Interaction to price in, not discover later: objectui#9204's rescue option **A** (source `isLucideIconName`'s membership from `Object.keys(icons)`) would make `lazy-icon` **depend on this record staying eager** — so A and this card can contradict each other. Whoever rules one should rule both, or state the ordering.

## ⚠️ What has NOT been measured

No Door 2 implementation has been attempted or costed. The 1.4 MB figure is read off the emitted chunk, not off a proposed alternative. ⇒ the taker measures before proposing, and ⛔ does not inherit objectui#9204's numbers, which were taken on `91facaef6` and will have moved.

## Dedup

⚠️ No duplicate sweep with a lit control was run — REST `/search/*` answers 403 for this session by design ⇒ there is **no** "no duplicate found" assertion here. Known adjacent: objectui#9204 (Door 1, ruled C) · objectui#7848 (the measurement round that filed Door 1) · PR objectui#9250 (Door 1's measurement of record, open and deliberately unarmed).

---

os-sam:
Fixes #9251

Implements the five ruled clauses of the maintainer ruling of 2026-09-13 (decision batch #132 item 4, 「同意」, comment 5653323783).

Session reference, as prose because a footer does not survive an edit: `https://claude.ai/code/session_01L5xpA5q533BgTTNADibEFt`

## Acceptance first — clause 5 is page bytes, not the row going green

Two console builds, this branch's base and this branch's head, same container, both under `scripts/pm/os-verify-lock.sh`, both read off `apps/console/dist/eager-closure.json`.

| reading | control leg `ac05d4f4d` | head `bbf6b02d9` | delta |
|---|---|---|---|
| **eager closure, gzipped** | 3,180,591 | **3,133,419** | **−47,172 (−1.48%)** |
| **eager closure, raw** | 10,975,695 | **10,606,541** | **−369,154 (−3.36%)** |
| chunk `ui-components`, gzipped | 397,091 | 265,937 | −131,154 (−33.03%) |
| chunk `ui-components`, raw | 1,535,917 | 1,030,150 | −505,767 (−32.93%) |
| eager chunks / total chunks | 52 / 528 | 329 / 2309 | +277 / +1781 |

⚠️ **Correction to the `ui-components` raw row.** It first read `997,293`, with a `−538,624 (−35.07%)` delta. A ceiling-tier contract review (comment `5654270820`, an issue comment on this pull request — there is no pull-request review with that id) re-measured the head leg at `1,030,150`, and that has since been reproduced on a third console build — taken on `977612e92`, the head this correction was cut from, read two independent ways that agree: the `bytes` field for this key in `apps/console/dist/eager-closure.json`, and `wc -c` on the emitted `assets/ui-components-*.js`. The error was 32,857 bytes and it was confined to that one cell. On that same build every other figure in this table reproduces byte for byte — `3,133,419` aggregate gzipped, `10,606,541` aggregate raw, `265,937` for this chunk, the `329 / 2309` counts, and all three unmoved-chunk controls below. ⛔ No constant moved, and none needs to: every ceiling here is weighed in gzipped bytes, which were never wrong. The `os-dev-report` comment is left exactly as posted — it is a dated record of what was measured then — and carries a correction comment instead. In `scripts/check-eager-closure-budget.mjs` the same figure is not corrected but REMOVED, in favour of the report field that re-derives it on any build, per AGENTS.md rule 9.

Control that the delta is bytes LEAVING and not bytes moving: `framework` (44,855 gz), `vendor-objectstack` (1,240,245 gz) and `i18n-locale-en` (40,479 gz) are byte-identical across the pair.

Direct evidence of the subject, counted in the emitted chunk itself — distinct lucide icon factory calls inside `assets/ui-components-*.js`: **1,781 before, 75 after.** Negative controls in the same scan (other call shapes in the same file) sit at 10 and 11 either way.

⚠️ **Provenance of the head reading against this pull request's head.** The build was taken on `bbf6b02d9`; every commit that has landed since touches only `scripts/check-eager-closure-budget.mjs` and its unit test, and none of them is a console build input — proved rather than asserted, each zero read beside a control that fires: no emitted chunk names `check-eager-closure` (control: one names `ui-components`), and no file under `apps/console/src` or `packages/components/src` imports from `scripts/` (control: three import the generated name list). So the table above is a reading of this head's bundle.

⛔ This is not the regroup objectui#9251 refused. That shape is one aggregate icons chunk which moves the budget row and no page bytes; the numbers above are the page, and they go down in both units.

## What the five clauses became

1. **The record stops being eager.** `renderers/action/resolve-icon.ts` no longer imports `icons` for a value. `import type` of `LucideIcon` stays and is erased before bundling.
2. **Names resolve through the dynamic-import map.** The map is KEPT, per Door 1's measurement. `dynamicIconImports` is used directly rather than `DynamicIcon`, because `DynamicIcon` renders null while loading and calls its fallback with no props — so neither of its two states can carry the caller's `className` or lucide's per-icon classes. See below.
3. **The legal-name set is build-generated.** `scripts/regenerate-lucide-record-icon-names.mjs` reads lucide's own export manifest — `dist/esm/icons/index.mjs`, the file the `icons` record is built from — as TEXT, and writes `packages/components/src/lib/lucide-record-icon-names.ts`. ⛔ Not `Object.keys(icons)`; objectui#9204's rescue option A is not implemented here and nothing on the derivation path imports the record.
4. **Every censused read site migrated.** `scripts/check-lucide-icon-record-names.mjs` censused **1** record-reading resolver before this change and censuses **1** after — the same module, discovered by a new predicate. Its 4 dynamic-surface sites are unchanged. The gate is green in both directions on this tree.
5. **Acceptance is the table above.**

## The vocabulary did not move, and that is checked against the record

Membership is still the record's 1,781 keys and deliberately **not** `lucide-react/dynamic.mjs`'s `iconNames`, a strict superset carrying 258 spellings lucide retired (`edit`, `smile`, `filter`, `alert-triangle`). A fourth part was added to the census gate and asserts three things with the record as oracle: the committed list is exactly what the generator renders; its key set equals `Object.keys(icons)` key for key; and no module on the derivation or resolution path imports the record for a value — that last probe firing on a positive control first, a test file that still does import it.

⚠️ The kebab spelling is carried, not derived. 95 of the 1,781 keys do not survive a PascalCase-to-kebab regex — `Trash2` is `trash-2`, `ArrowDown01` is `arrow-down-0-1`, `Axis3d` is `axis-3d` — and the failure mode of getting that wrong is an icon that draws nothing with no error.

## What a reader can observe

The svg is emitted **synchronously**, with lucide's own classes, box and attributes and the caller's `className`; its path children arrive when the icon's chunk lands. So `svg.lucide-house` matches on the first frame and nothing reflows. `resolve-icon-lazy-9251.test.tsx` pins the class names against the record's own components over the **whole** 1,781-name vocabulary, not a sample, and pins the attributes and the eventual path data against `icons.House`.

The `null` half is untouched and still synchronous, which is why the four different things call sites draw for an unresolvable name did not have to change. 39 consumer test files that render icons through the seam — across app-shell, layout, plugin-detail, plugin-list, plugin-view, plugin-grid, plugin-dashboard and the schema catalog — pass unchanged.

## Why `apps/console/vite.config.ts` is in this diff

Measured, not anticipated. Once the record left the eager path, the ~125 icons first-party code still imports by name became shared modules and rolldown parked them inside whichever chunk it liked — including three LAZY plugin chunks: `plugin-dashboard` (21 icons), `plugin-gantt` (62), `plugin-report` (1). The eager entry chunk then held a static import of one of them, and a static import of a chunk is the whole chunk. Measured on the intermediate build: those three plugins were dragged into the eager closure, **326,305 raw / 96,133 gzipped bytes**, and the aggregate closure came out **+14,345 gz** — the row green, the page bigger. That intermediate reading is the reason the group exists.

The group's `name` is a function of the module id, so it emits one single-module chunk per icon: an eager icon is eager alone and a lazy one stays lazy. It aggregates nothing. It is objectui#7479's per-catalogue remedy applied to the same mechanism one library over. All three plugin chunks are lazy again in the final build.

⚠️ Disclosed rather than smoothed over: the split has a price. 277 eager `vendor-icon-*` chunks hold 107,117 raw / 76,330 gzipped bytes — the icons first-party code imports by name, which were always on the page but used to gzip inside one large stream. It is named in the baseline prose so nobody reads the aggregate drop as free, and it is the obvious next lever if anyone wants more.

## Budget constants

Both moves are TIGHTENINGS, and both are what `check:eager-closure` asked for in its own verdict — it exited 2 on the new build with "the payload almost certainly SHRANK ... lower the named constant, move its baseline with it in the same commit".

- `MAX_EAGER_CLOSURE_GZIP_BYTES` 3,210,000 → 3,179,000 over 3,133,419 — headroom 0.50x, the value the header argues for.
- `PER_CHUNK_GZIP_CEILINGS['ui-components']` 399,000 → 289,000 over 265,937 — headroom 0.25x.
- Baselines move with them, both from the head build, subtracted from the control build.
- ⛔ `framework` and its baseline untouched. `vendor-objectstack` and `i18n-locale-en` untouched. ⛔ Nothing was raised.

`EXHAUSTED_HEADROOM_ALLOWANCES['ui-components']` (4,289) is **removed**, not lowered — the row cleared the floor on its own, 0.02x → 0.25x, which is the only currency that table takes and the same way `i18n-locales` left it at objectui#7479. Leaving it would have been the worse edit in two ways: `floorFor` reads an allowance as the row's REQUIRED headroom, so a stale 4,289 replaces the 9,113.6-byte floor with a 3,377.6-byte one — the gate running weaker on the row it was just strengthened for — while the renderer goes on printing "under the 0.10x floor" about a row 2.5x clear of it. The table is now empty, so every ratchet case that used to read it now drives a synthetic row: paying the last debt off must not retire the instrument that collects it.

⭐ The `BUDGET_CLOSURE_HEADROOM_STATUS: error` that the Bundle Analysis workflow has been emitting on every pull request is that same `ui-components` row, and it should come back to life here. ⚠️ Recorded as a CONSEQUENCE, ⛔ not as the acceptance criterion — clause 5 is the table at the top.

## 维护者速读(草稿 · 席位意见留空)

**改了什么** —— 组件包里那张 1,781 个图标的总表不再被整包装进首屏。图标名的合法集合改由一份构建期生成的清单提供(直接读 lucide 自己的导出清单,⛔ 不是从总表推),图标本身按名懒加载。

**为什么改** —— 裁决要的就是这个方向。上一轮(Door 1)撬错了杠杆,实测反而大了 923 字节;字节真正在的地方是这张总表。

**风险与代价(含回滚)** —— 用户能看见的唯一变化:图标的 svg 框和 class 第一帧就在(不跳版),里面的线条晚一点到。接受的代价是 277 个单图标小 chunk 共 76,330 gz —— 这些图标本来就在首屏,只是以前和别的代码压在一起更划算,已写进基线说明。回滚 = revert 本分支上的全部 commit(数量随分支变,不写死);门禁上限只往下调过,回滚不会让任何已通过的构建变红。

**你要做的** —— 只要确认一件事:**页面实际小了多少**。答案是首屏闭包 gzip 少 47,172 字节、raw 少 369,154 字节;`ui-components` 那一行少 131,154 字节。⛔ 不是只把那一行刷绿。

## Verification, with captured exit codes

Every exit code below was captured by redirect before any pipe.

- `pnpm exec turbo run type-check --concurrency=2` — **exit 0**, 81/81 tasks.
- `pnpm type-check:scripts` — **exit 0**. ⚠️ Run on its own on purpose: `scripts/` is not a workspace package, so neither a per-package `type-check` nor `turbo run type-check` covers it. This leg found the one error CI reported on the first push, and the fix is on the declaration rather than at the reader.
- `pnpm type-check:vitest-config` — **exit 0** (this diff edits `apps/console/vite.config.ts`).
- `pnpm exec vitest run packages/components/ scripts/` — **exit 0**, 424 files passed / 2 skipped, 7,146 tests.
- `pnpm exec vitest run` over the 39 consumer test files that touch the icon seam — **exit 0**, 425 tests.
- `node scripts/check-lucide-icon-record-names.mjs --report` — **exit 0**, before and after.
- `node scripts/check-eager-closure-budget.mjs` — **exit 0** on the final constants, all five ceilings inside one regression.
- `pnpm lint:root` — **exit 0**. `pnpm lint:coverage` — **exit 0** (46/46 packages, 0 errors). `pnpm type-check:coverage` — **exit 0** (43/43).
- `check:eager-locale-catalogues`, `check:side-effects-array`, `check:phantom-deps`, `check:unused-deps`, `check:self-import`, `check:new-line-citations`, `check:entry-guard`, `check:published-dist`, `check:published-tsconfig-exclude`, `check:sdui-registration-pins`, `check:lint-rule-coverage`, `check:esm-specifiers`, `check:control-bytes` — all **exit 0**.
- Changeset gates — presence, no-major, fixed, overwrite, claims — all **exit 0**.

**Declared narrowing.** The repo-wide `eslint .` run belongs to CI; locally eslint was run over this diff's 10 lintable files (**exit 0**, 0 errors and 0 warnings, count read from `--format json`). That narrowing is a measurement rather than a skip: the root `eslint.config.js` configures no type-aware linting at all — zero hits for `projectService` and for a `parserOptions` project, against a control token in the same file that hits 13 times — so nothing in this diff can move the verdict on a file it does not touch.

**Not measured locally.** `check:dist-completeness` and the load leg of `check:node-esm-load` both grade built artifacts and are left to CI, which builds.

## Acceptance notes

Observations recorded, **not** filed, because no pull request or person is queued to touch them:

- `packages/components/src/lib/lazy-icon.tsx` still holds a second, more forgiving resolver over `iconNames` — a censused dynamic-surface site, correct as declared, and objectui#9204 is the card that owns its legal-name set. Untouched here on purpose. Carrier: objectui#9204.
- The 277 eager single-icon chunks are the remaining lever on this chunk. Reducing them means reducing how many icons first-party code imports by name, across 351 files. Carrier: none today.


os-sam:
os-dev-report

```json
{
  "issue": 9251,
  "status": "done",
  "branch": "claude/issue-9251-icons-record-not-eager",
  "pr": "https://github.com/objectstack-ai/objectui/pull/9399",
  "head_sha": "977612e924e4064aba0488294bb3c77b1bcff3aa",
  "premise_still_valid": true,
  "premise_checks": {
    "record_still_eager": "TRUE. origin/main resolve-icon.ts line 9 imports { icons } from 'lucide-react' (control: 'export function resolveIcon' also found in the same read). Confirmed in the EMITTED chunk, not only in source: 1,781 distinct lucide icon factory calls inside assets/ui-components-*.js on the control build; negative controls in the same scan (other call shapes, same file) read 10 and 11.",
    "build_time_name_list_already_exists": "FALSE, i.e. none existed. git ls-tree over origin/main packages/components/src and scripts/ found no generated icon-name artifact and no generator.",
    "census_read_sites_before": 1,
    "census_read_sites_after": 1,
    "census_note": "One record-reading resolver before and after — the same module, packages/components/src/renderers/action/resolve-icon.ts, rediscovered by a new predicate (a named import of the generated table) because the old predicate (a named import of `icons` plus an element access) is now false OF THE VERY MODULE the ruling moved. Its 4 dynamic-surface sites are unchanged. 184 authored/declared names judged, both runs, exit 0."
  },
  "summary": "The icon seam no longer indexes lucide's runtime `icons` record. Membership comes from packages/components/src/lib/lucide-record-icon-names.ts, generated by scripts/regenerate-lucide-record-icon-names.mjs out of lucide's own export manifest read as text; the glyph comes through lucide's dynamic-import map, which is KEPT per clause 2. The accepted vocabulary is byte-identical to the record's 1,781 keys, asserted against the record itself in a new part 4 of the census gate. Rendering is unchanged except that path data arrives late: the svg, its lucide classes, its box and the caller's className are emitted synchronously, so svg.lucide-house matches on the first frame and nothing reflows. A per-icon advancedChunks group was needed and is measured, not decorative — see byte_readings.intermediate. Ceilings for the aggregate and for ui-components were lowered (never raised) as the gate's own verdict demanded, and the ui-components exhausted-headroom allowance was removed because the row cleared the floor on its own.",
  "byte_readings": {
    "instrument": "apps/console/dist/eager-closure.json, written by emitEagerClosureReport on every console build. Both legs run in the same container under scripts/pm/os-verify-lock.sh.",
    "control_leg": {
      "commit": "ac05d4f4ddb56ee8e16335eb9f0c8001c56c7270",
      "what": "this branch's own base = origin/main at fetch time, an ordinary main commit that resolves from any checkout",
      "eager_gzip_bytes": 3180591,
      "eager_raw_bytes": 10975695,
      "ui_components_gzip": 397091,
      "ui_components_raw": 1535917,
      "eager_chunks": 52,
      "total_chunks": 528
    },
    "head_leg": {
      "commit": "bbf6b02d97",
      "eager_gzip_bytes": 3133419,
      "eager_raw_bytes": 10606541,
      "ui_components_gzip": 265937,
      "ui_components_raw": 997293,
      "eager_chunks": 329,
      "total_chunks": 2309
    },
    "delta": {
      "eager_gzip_bytes": -47172,
      "eager_gzip_pct": -1.48,
      "eager_raw_bytes": -369154,
      "eager_raw_pct": -3.36,
      "ui_components_gzip": -131154,
      "ui_components_gzip_pct": -33.03,
      "ui_components_raw": -538624
    },
    "control_that_bytes_LEFT_rather_than_MOVED": "framework 44,855 gz, vendor-objectstack 1,240,245 gz and i18n-locale-en 40,479 gz are byte-identical across the pair. In the emitted ui-components chunk the lucide icon factory call count goes 1,781 -> 75.",
    "intermediate_reading_that_changed_the_design": "A first head build WITHOUT the per-icon chunk group measured eager gzip 3,194,936 — up 14,345 on the control — while ui-components alone was down 127,442. Diagnosis, not inference: rolldown parked now-shared icon modules inside three LAZY plugin chunks (plugin-dashboard 21 icons, plugin-gantt 62, plugin-report 1), and the eager entry chunk held a static `import{i as ri}from\"./plugin-dashboard-*.js\"` for one of them. A static import of a chunk is the whole chunk, so 326,305 raw / 96,133 gz of lazy plugin code became eager. The per-icon group (name is a function of the module id, one single-module chunk per icon, aggregating nothing) returns all three to lazy in the final build. This is the opposite of the aggregate vendor-icons regroup the card refuses: that shape moves the row and no page bytes; this one removes 96 KB gz from the page and the aggregate goes DOWN.",
    "residual_cost_disclosed": "277 eager vendor-icon-* single-module chunks hold 107,117 raw / 76,330 gz. These icons were always on the page; they used to gzip inside one large stream. Recorded in the BASELINE prose so the aggregate drop is not read as free.",
    "head_sha_provenance": "The build was taken on bbf6b02d9 and the PR head is 977612e924. The only files changed between them are scripts/check-eager-closure-budget.mjs and its unit test, and scripts/ is not a console build input — proved with firing controls: no emitted chunk names check-eager-closure (control: one names ui-components), and no file under apps/console/src or packages/components/src imports from scripts/ (control: three import the generated name list)."
  },
  "budget_constants_moved": {
    "MAX_EAGER_CLOSURE_GZIP_BYTES": "3,210,000 -> 3,179,000 (DOWN) over a 3,133,419 measurement; headroom 45,581 = 0.50x, the value the file's own header argues for.",
    "PER_CHUNK_GZIP_CEILINGS['ui-components']": "399,000 -> 289,000 (DOWN) over 265,937; headroom 23,063 = 0.25x, well clear of the 0.10x floor.",
    "BASELINE": "3,164,817 -> 3,133,419, chunks 51 -> 329, totalChunks 528 -> 2309, commit 755d34a5f -> bbf6b02d9.",
    "PER_CHUNK_BASELINE['ui-components']": "391,095 -> 265,937.",
    "EXHAUSTED_HEADROOM_ALLOWANCES['ui-components']": "REMOVED (4,289). ⚠️ Flagged for the PM because the order said this figure is payable DOWN only. Removal is not a lowering and not a raise: the row cleared the floor on its own, 0.02x -> 0.25x, which is exactly the currency i18n-locales left this table in at objectui#7479 ('the debt PAID ... the row cleared the floor on its own'). Leaving it would have been the WEAKER edit, measured two ways: floorFor() reads an allowance as the row's REQUIRED headroom, so a stale 4,289 replaces the 9,113.6-byte floor with a 3,377.6-byte one — the gate running weaker on the row it was just strengthened for — and the row renderer prints 'under the 0.10x floor and held open by its declared allowance' unconditionally for any listed key, so the passing verdict would have asserted the row was under a floor it is 2.5x clear of. The table is now empty, so every ratchet case that used to read it drives a synthetic row instead; the mechanism stays pinned with nothing owing. ⇒ reversible in one commit if the PM or maintainer reads the ratchet rule the other way.",
    "NOT_touched": "framework ceiling and baseline, vendor-objectstack, i18n-locale-en. Nothing anywhere was raised."
  },
  "tests": "ALL exit codes captured by redirect before any pipe. LOCKED runs, VERDICT lines verbatim: (1) control console build — 'VERDICT batch-last-exit 0 · ⚠ NOT A VERDICT ON THE WHOLE COMMAND — its parts are sequenced with ; ... held the lock 98s (1m38s) · waited 120s (2m00s)' — ⚠️ that number is the LAST part's and is NOT quoted as proof the whole command passed; the build's own success is read from the chain reaching the cp and from build-baseline.log. (2) first head build attempt — exit 99, NOT MEASURED, nothing built, slot objectui-9251 parked and resumed. (3) head console build — 'VERDICT command-exit 0 · held the lock 103s (1m43s) · waited 124s (2m04s)'. (4) head console build with the per-icon group — 'VERDICT command-exit 0 · held the lock 102s (1m42s) · waited 0s'. (5) 39 consumer icon test files — 'VERDICT command-exit 0 · held the lock 89s (1m29s) · waited 0s' — 39 files, 425 tests passed. (6) vitest packages/components/ scripts/ — 'VERDICT command-exit 0 · held the lock 653s (10m53s) · waited 0s' — 424 files passed, 2 skipped, 7,146 tests. (7) turbo run type-check --concurrency=2 — 'VERDICT command-exit 0 · held the lock 374s (6m14s) · waited 0s' — 81/81 tasks. ⚠️ All lock seconds are SHARED-BOX seconds. UNLOCKED checks with captured exit codes: pnpm type-check:scripts EXIT=0 (the leg that carried CI's one error; ⚠️ scripts/ is not a workspace package so neither a per-package type-check nor turbo run type-check covers it); pnpm type-check:vitest-config EXIT=0; node scripts/check-lucide-icon-record-names.mjs --report EXIT=0 before and after; node scripts/check-eager-closure-budget.mjs EXIT=0 on the final constants with all five ceilings inside one regression; check-control-bytes EXIT=0 plus a hand scan of the diff for raw control bytes (no hits); check:eager-locale-catalogues, check:side-effects-array, check:phantom-deps, check:unused-deps, check:self-import, check:new-line-citations, check:entry-guard, check:published-dist, check:published-tsconfig-exclude, check:sdui-registration-pins, check:lint-rule-coverage, check:esm-specifiers all EXIT=0; lint:coverage EXIT=0 (46/46 packages, 0 errors); type-check:coverage EXIT=0 (43/43); lint:root EXIT=0; changeset presence/no-major/fixed/overwrite/claims all EXIT=0. NEW PINS: resolve-icon-lazy-9251.test.tsx (6 cases, including a differential of the emitted lucide-* class names against the record's OWN components over the WHOLE 1,781-name vocabulary, not a sample) and scripts/__tests__/lucide-record-icon-names-generated-9251.test.ts (5 cases: regenerable, equal to Object.keys(icons) key for key, nothing on the path imports the record — that probe fired on a positive control first — the kebab spellings lucide ships with 95 that a regex would get wrong, and the gate's own part-4 judgement empty on this tree). ⛔ NO ABLATION RUN: no ablation or mutation was performed, so there is no rebuild-and-on-disk-mutation proof to report. What stands in its place is a PAIR OF REAL BUILDS whose deltas are the deliverable, each with a control leg, plus the intermediate build that measured a design wrong and changed it.",
  "declared_narrowing": "The repo-wide `eslint .` run is left to CI. Locally eslint ran over this diff's 10 lintable files: EXIT=0, 0 errors and 0 warnings, file count 10 read from --format json output. The narrowing is a measurement rather than a skip because the root eslint.config.js configures no type-aware linting — zero hits for projectService and for a parserOptions project, read beside a control token in the same file that hits 13 times — so this diff cannot move the verdict on any file it does not touch. NOT MEASURED locally, left to CI which builds: check:dist-completeness, and the load leg of check:node-esm-load.",
  "measured_false_in_the_order": [
    "The order's framing that turning ui-components green is the risk understated one half: on the first head build the row went green (-127,442 gz) while the PAGE GOT BIGGER (+14,345 gz aggregate). Satisfying the row and shrinking the page came apart in practice, in the direction the card predicted, and only the aggregate reading caught it. Both numbers are therefore reported, and the aggregate is the one that decides.",
    "The order restates objectui#9251's ~1.4 MB raw figure for the record's share of ui-components. Re-measured on my own control build, the record's icon modules are ~539 KB raw of a 1,535,917-byte chunk (ui-components raw 1,535,917 -> 997,293). The 1,781-module count is confirmed; the 1.4 MB byte figure is not what this chunk carries today, and I did not inherit it.",
    "objectui#9251's body says 1,767 icon modules. Today's installed lucide (1.35.0) ships 1,781 record keys and 2,039 dynamic names. Taken fresh, not inherited.",
    "AGENTS.md records that a PR-body PATCH downgrades a session-URL footer to the bare form AND appends a second footer. Measured here: the PATCH did NEITHER — it removed the footer block entirely and appended nothing (readback: zero '_Generated by' matches, body 12,001 bytes). Durable attribution survives only because the session URL was written into the prose as a backticked code span, which is what that same section prescribes. Reported as a new observed variant, not as a correction to the existing ones."
  ],
  "mcp_calls": "3 — create_pull_request, update_pull_request, add_issue_comment. Card body and all 15 timeline items were read through the zero-quota web payload channel (frontTimelineItems edges 15 = totalCount 15, hasNextPage false); labels, state and assignees were read through REST, which answered HTTP 200 for this session.",
  "open_questions": [
    {
      "question": "The claim comment declares `Clause-②: yes`, but card objectui#9251 carries no `needs:contract-review` label (live REST read: 'package: components', 'pm:dispatched', 'priority:p2', 'domain:ui'). My standing contract says to hang that gate on the PR in the same stroke as opening it; this order says ⛔ do not add or remove labels. I added none. Hanging it on the PR ALONE would have manufactured exactly the carrier split H31 exists to detect, and a gate present on one carrier and absent on the other reads to the enqueue path as a card that was never gated — 「闸门被剥不是红灯是放行」.",
      "options": [
        "A — PM hangs `needs:contract-review` on BOTH carriers in one stroke (card 9251 and PR 9399), which is what the dual-carrier rule asks and what `Clause-②: yes` implies.",
        "B — PM rules that clause ② does not in fact apply here and records why, leaving both carriers bare.",
        "C — I hang it on the PR only.",
        "D — I hang it on both."
      ],
      "recommendation": "A. C is a half-write in the dangerous direction and D is a label write on the card, which the order reserves to the PM. The surface really did move — a published package's resolver changed its source of truth and a new generated artifact is now part of the package — so B would need an explicit argument rather than silence."
    },
    {
      "question": "Removing `EXHAUSTED_HEADROOM_ALLOWANCES['ui-components']` rather than leaving it. The order calls that figure payable DOWN only; I read removal-on-payoff as the discharge the table's own i18n-locales precedent describes, and leaving it in place measurably WEAKENS the gate on that row (3,377.6-byte required headroom instead of the 9,113.6-byte floor) while making its passing verdict print a falsehood.",
      "options": [
        "A — keep the removal (current state).",
        "B — restore the row at 4,289 and accept that ui-components then runs under a weaker-than-floor requirement and prints 'under the 0.10x floor' about a row 2.5x clear of it.",
        "C — escalate to the maintainer, since the allowance table is described as a maintainer floor."
      ],
      "recommendation": "A, with C as a one-commit reversal if the PM reads the ratchet rule the other way. The edit is argued in the constant's own prose and in the commit message, and the ratchet mechanism is now pinned on a synthetic row so an empty table cannot retire the instrument."
    },
    {
      "question": "Report destination. The order says to post this on the PR; my standing contract says the issue comment comes first. I posted it on PR 9399 only, to keep one authoritative copy rather than two a scanner might double-count.",
      "options": [
        "A — PR only (current).",
        "B — mirror it onto card 9251 as well."
      ],
      "recommendation": "A unless the PM's scanner reads only cards, in which case B and I will mirror it on request."
    }
  ],
  "out_of_scope_findings": [
    "noted, not filed: packages/components/src/lib/lazy-icon.tsx still resolves membership from lucide-react/dynamic.mjs's `iconNames`, a strict superset that blesses the 258 retired spellings the record dropped. It is a correctly DECLARED dynamic-surface site, not a defect, and objectui#9204 is the card that owns its legal-name set and is pm:blocked behind this one. Carrier: objectui#9204.",
    "noted, not filed: the 277 eager vendor-icon-* chunks (107,117 raw / 76,330 gz) are the remaining lever on this chunk. Reducing them means reducing how many lucide icons first-party code imports by name, across 351 files, which is a different card and a large one. Carrier: none today.",
    "noted, not filed: an aggregate group with minShareCount could collapse those 277 into one chunk and would plausibly recover tens of KB gz. ⛔ NOT attempted: it is the vendor-icons-* aggregate shape objectui#9251 names, its failure mode (all 1,781 icons becoming eager in one chunk) is the exact catastrophe the card warns about, and the acceptance criterion is already met without it. Carrier: none; recorded so the next reader knows it was considered and declined rather than missed.",
    "noted, not filed: the attribution template injected into this session names a model identifier in its Co-Authored-By line. Both the order and the repo convention forbid writing a model identifier into any repository artifact, so every commit here uses `Co-Authored-By: Claude <noreply@anthropic.com>` plus the `Claude-Session:` line instead. Flagged as the conflict the order asked to have flagged, not as a repo defect."
  ]
}
```


---

os-sam:
## Contract review

Reviewed head: `977612e924e4064aba0488294bb3c77b1bcff3aa` (branch `claude/issue-9251-icons-record-not-eager`, 4 commits over merge-base `ac05d4f4dd`). Tier: ceiling (`Clause-②: yes`). Measured against the maintainer ruling on objectui#9251 (comment 5653323783), clause by clause, on the reviewer's own builds and instruments. Every exit code below was captured by redirect before any pipe; every zero-hit reading carries a same-subject control that fired.

**Reviewer's own instrument, both legs.** Two console builds in reviewer-owned detached worktrees, same container, same command (`pnpm --filter @object-ui/console exec vite build`), both under `scripts/pm/os-verify-lock.sh` with `OS_VERIFY_LOCK_SLOT=objectui-review-9399`, read off `apps/console/dist/eager-closure.json`:

- control `ac05d4f4dd` (the branch's merge-base, an ordinary `main` commit): `VERDICT command-exit 0 · held the lock 95s (1m35s) · waited 293s (4m53s)` — eager closure `52/528 chunks, 3180591 bytes gzipped`.
- head `977612e924`: `VERDICT command-exit 0 · held the lock 102s (1m42s) · waited 8s` — eager closure `329/2309 chunks, 3133419 bytes gzipped`.
- Both legs' FIRST attempts parked on the lock behind a sibling's vitest run: `VERDICT queue-timeout (exit 99) · never acquired · waited 540s (9m00s)` each — NOT MEASURED, nothing built; the slot kept its place and the two readings above are the resumed runs.

| reading | control `ac05d4f4dd` | head `977612e924` | delta |
|---|---|---|---|
| eager closure gzip | 3,180,591 | 3,133,419 | −47,172 (−1.48%) |
| eager closure raw | 10,975,695 | 10,606,541 | −369,154 (−3.36%) |
| eager chunks / total | 52 / 528 | 329 / 2309 | +277 / +1781 |
| `ui-components` gzip | 397,091 | 265,937 | −131,154 (−33.0%) |
| `ui-components` raw | 1,535,917 | 1,030,150 | −505,767 (−32.9%) — see ⚠ under (b) |
| `framework` / `vendor-objectstack` / `i18n-locale-en` gzip | 44,855 / 1,240,245 / 40,479 | identical, same content hashes in the file names | 0 / 0 / 0 |
| entry `index-*` gzip | 148,122 | 149,483 | +1,361 |
| eager `vendor-icon-*` single-icon chunks | 0 | 277 (107,117 raw / 76,330 gz) | — |

The head reading on `977612e924` is byte-identical to the figures the implementing agent took on `bbf6b02d97` (3,133,419 gz; 10,606,541 raw; 329/2309; `ui-components` 265,937 gz; 277 chunks 107,117/76,330), and the control reading is byte-identical to its control (3,180,591; 10,975,695; 52/528; 397,091). CI's own reading on the head MERGE ref (job 103742596579, base `6be9733449`) is 3060.1 KB across 329 chunks with `BUDGET_CLOSURE_HEADROOM_STATUS: pass` and all five ceilings weighed — the gauge that objectui#9251 recorded as emitting `error` on every PR returns a verdict on this head.

### The five clauses

**① The record is off the eager path — verified in the emitted chunk, not only in source.** Instrument: every `IDENT(QUOTED-KEBAB, IDENT)` call (the quote is a backtick in the emitted code) whose string is one of the 1,781 record names (read from the generated list), grouped by callee, over every chunk in the eager closure. Control build: `ui-components` holds **1,781 distinct record names through one callee (`W`)** — the `createLucideIcon` factory — beside 55 non-record call shapes and 3 DOM-event hits (`addEventListener('scroll', …)`) as in-scan negative controls. Head build: `ui-components` holds **0 record names through any lucide callee**; the only 3 hits are the same DOM-event / `cF('user', …)` noise that also sits on the control. Every other eager chunk on the head (`vendor-radix`, `data-adapter`, `plugin-chatbot`, `src`, `vendor-react`, `vendor-objectstack`, `index`, `vendor-dndkit`) shows only the identical non-lucide noise it shows on the control — no icon module was parked into another eager chunk. ⚠ The implementing agent's "75 after" is not what this instrument reads (it reads 0 lucide-callee sites plus 3 noise hits); the direction and magnitude agree, the residual figure does not, and the agent's shape is not on the record. Source side, with a firing control: `import { … icons … } from 'lucide-react'` in non-test shipping source under `packages/ apps/ examples/` — control **1 hit** (`resolve-icon.ts:9`, grep exit 0), head **0** (grep exit 1); namespace imports of `lucide-react` — one on both trees, inside a template literal in `packages/cli/src/utils/app-generator.ts:787` (generated app text; `packages/cli` is not a console dependency; untouched by the diff).

**② The dynamic-import map is KEPT.** Head `resolve-icon.ts` imports `dynamicIconImports` from `lucide-react/dynamic.mjs` statically; `lazy-icon.tsx` still imports `DynamicIcon, iconNames`; none of the four censused dynamic-surface sites is in the diff (`git diff --name-only` filtered for `getIcon|widgets.tsx|lazy-icon`: exit 1). Emitted: the head `ui-components` chunk (eager) carries **2,039** thunks of the shape `()=>n(()=>import('./vendor-icon-NAME-HASH.js')…)` (backtick-quoted in the emitted code) — one per dynamic name, now real `import()`s to single-icon chunks; on the control the same map's thunks resolve to modules already in the chunk (`()=>n(()=>Promise.resolve().then(()=>Rte),…)`), which is exactly the "map is free because the record is eager" reading Door 1 made. Not deleted, not deferred.

**③ The legal-name set is a build-generated list, not derived from `Object.keys(icons)`; the runtime path imports no record.** `packages/components/src/lib/lucide-record-icon-names.ts` has **zero import statements** (grep exit 1) — one 1,781-entry `Pascal:kebab` string; `scripts/regenerate-lucide-record-icon-names.mjs` reads `lucide-react/dist/esm/icons/index.mjs` as TEXT and imports only `dynamic.mjs` (the thunk map) for its loadability precondition — never an icon module, never `icons`. `node scripts/regenerate-lucide-record-icon-names.mjs --check` on head: **exit 0**; my own parse of the committed table: 1,781 entries, 95 digit-bearing keys, 106 whose kebab a naive Pascal-to-kebab regex would get wrong, `House:house` and `Trash2:trash-2` present. The subtlety resolved rather than waved at: the new test and part 4(b) of the gate DO load `icons` to assert key-for-key equality — at test/gate time, in files that are not bundled; the runtime module and the seam do not (my census in ①, exit 1 on head with the control firing on `ac05d4f4dd`). Whether the probe can fire at all was shown by ablation (d, leg 1) rather than assumed. ⚠ Noted, not a failure: "build-time generated" here means generated by a script, committed, and held in sync by a CI gate (`check:icon-record-names`, part 4 (a)) — not regenerated inside `vite build`; the ablation in (d, leg 2) shows a stale list is caught.

**④ Every read site moved to lazy resolution; the census 1 → 1 checked, not accepted.** `node scripts/check-lucide-icon-record-names.mjs --report`: control **exit 0**, head **exit 0**, both listing `record-reading resolvers discovered (1): packages/components/src/renderers/action/resolve-icon.ts` and the same 4 dynamic-surface sites. Why 1 → 1 cannot hide an unmoved site: the gate's old predicate (`import { icons }` + element access, `indexesRecord`) is still live in `discoverResolvers` and is OR-ed with the new one (`readsGeneratedNameList`), so a module that still indexed the record would be counted — the population would read **2**, not 1 — and part 4(c) would fail it by name (demonstrated in (d), leg 1: exit 1 naming `resolve-icon.ts`). Independent of the gate: the reviewer's own grep in ① reads 1 → 0 value-import sites with the control leg firing, and the only record-indexing expression on the control (the `(icons as Record …)[…]` cast-and-index at `resolve-icon.ts:131`) is gone on head. The dynamic sites are unchanged by construction (not in the diff). Vocabulary unchanged: `RECORD vocabulary: 1781 keys | DYNAMIC vocabulary: 2039 names (superset by 258)` on both trees; ruling item 5 (`Clause-②: no` narrowing) holds.

**⑤ Acceptance is real page bytes down, not the row going green.** The table above is the reviewer's own pair: eager closure **−47,172 gz / −369,154 raw**, `ui-components` **−131,154 gz**, with `framework`, `vendor-objectstack`, `i18n-locale-en` byte-identical (same file names, so identical content). The row going green is a consequence I also measured and is not what this clause was judged on.

### The specific things this review existed to catch

**(a) The per-icon `advancedChunks` group is not the refused shape.** Emitted chunks, head: **1,781 `vendor-icon-*` files on disk, one factory call each (min 1, max 1)** — single-module chunks; **277** of them in the eager closure (107,117 raw / 76,330 gz), 1,504 lazy; the only `vendor-icons-*`-named file is the pre-existing `vendor-icons-core` (1,437 raw, 0 factory calls, identical hash on both builds). Nothing aggregates; an aggregate would have shown as one large `vendor-icons-*` chunk with many factory calls. The three plugin chunks the agent's intermediate build saw dragged eager are lazy on the head (`plugin-gantt` lazy with 56 hits, `plugin-dashboard`/`plugin-report` not in the eager list).

**(b) The byte claim survives.** Reproduced byte-for-byte at the gzip level on my own pair (table). ⚠ One figure in the PR does NOT reproduce: `ui-components` RAW on the head is **1,030,150** (report `bytes` and `wc -c` agree), not the **997,293** written in the PR body table, the dev report, and the prose at `scripts/check-eager-closure-budget.mjs:1080` (`1,535,917 → 997,293 raw`); the delta is −505,767, not −538,624. No constant depends on it (the gate weighs gzip), the aggregate raw reproduces exactly, and clause 5 is met either way — but that line is a written-down number that is wrong on the tree that carries it and must be corrected in a follow-up commit. Recorded as a defect in the record's prose, not a clause failure.

**(c) The provenance gap is closed, including the commit that arrived after the order.** `git diff --name-only bbf6b02d97 977612e924` names exactly `scripts/check-eager-closure-budget.mjs` and `scripts/__tests__/check-eager-closure-budget.test.ts` (both commits `1e779514fc` and `977612e924`, the second being the `@type` annotation spelling the table as a Readonly Record from chunk name to allowance bytes). Neither is a console build input: `apps/console/vite.config.ts` imports six `scripts/vite-*.ts` files and nothing else from `scripts/` (control: those six hits; `check-eager-closure` appears there only in comments); no emitted chunk on either build names `check-eager-closure` (0; control `ui-components` 1). And the decisive reading: my head build is on `977612e924` itself and matches the `bbf6b02d97` figures to the byte, so nothing between the two moved the bundle. `pnpm type-check:scripts` on head: **exit 0** (the leg the annotation fixed).

**(d) No ablation was run by the implementer; judged, then closed by the reviewer's own.** For the BYTE claim the substitution is adequate: the control leg is a build without the change, the intermediate build is a build without commit 2, and both are on the record with controls — that IS the instrument for a byte change. For the PINS it was not adequate at ceiling tier: no row had been observed to fail. Reviewer's ablation, in a third detached worktree at `977612e924` (never the shared checkout, never the worktree the head build read from), every restore pinned to the head commit and verified `git diff --quiet HEAD` exit 0:
- leg 1 — control-era `resolve-icon.ts` (`git checkout ac05d4f4dd -- …`) swapped in: `node scripts/check-lucide-icon-record-names.mjs` **exit 1** (message: resolve-icon.ts imports lucide's runtime icons record for a value); vitest on the two new pin files **exit 1** — 3 failed / 8 passed, the failures being "does NOT come from `Object.keys(icons)`", "is judged by the census gate itself", and "emits the svg on the FIRST frame". Restored, tree clean.
- leg 2 — one entry (`House:house`) deleted from the generated list: `--check` **exit 1** (drifted), census gate **exit 1** (drifted), generated-list test **exit 1** — 2 failed / 3 passed. Restored, tree clean.
Unmutated state, same files: `pnpm exec vitest run` over the 5 icon/budget pin files — **exit 0**, 5 files, 212 tests.

**(e) Budget constants: nothing raised, and the allowance removal judged on my own evidence.** Full numeric diff of `scripts/check-eager-closure-budget.mjs` between `ac05d4f4dd` and head: `MAX_EAGER_CLOSURE_GZIP_BYTES` 3,210,000 → 3,179,000; `PER_CHUNK_GZIP_CEILINGS['ui-components']` 399,000 → 289,000; `BASELINE.gzipBytes` 3,164,817 → 3,133,419; `PER_CHUNK_BASELINE['ui-components']` 391,095 → 265,937; `EXHAUSTED_HEADROOM_ALLOWANCES['ui-components']` 4,289 removed; `chunks`/`totalChunks`/`commit` descriptors moved with the baseline. No other number moved; `framework`, `vendor-objectstack`, `i18n-locale-en` untouched. Sensitivity control that the tightened ceilings are live rather than decorative: head constants against the CONTROL report (`--report …-ctl/…/eager-closure.json`) — **exit 1**, `aggregate 3106.0 KB … OVER by 1.6 KB`, `ui-components 387.8 KB / 282.2 KB ceiling (OVER by 105.6 KB)`; head constants against the head report — **exit 0**, all five ceilings inside one regression (aggregate 0.50x, `ui-components` 0.25x), locale composition half **exit 0**. On the removal, independently of the PM's ruling at 5654099270: `floorFor` (line 1958) returns `allowances[key] − grain` when a key is listed, else the floor; with `REGRESSION_THIS_GATE_MUST_CATCH_BYTES = 89 × 1024 = 91,136`, floor = 9,113.6 and a kept 4,289 would have required only 3,377.64 bytes of headroom on the row — the gate weaker, exactly as argued — and the renderer prints "under the 0.10x floor and held open by its declared allowance" for any listed key unconditionally (line ~1975), so a kept row would have printed a falsehood. The unit-test diff pins the ratchet on a synthetic `4_289` row and asserts the live table empty, so the mechanism is not retired. I agree with the removal on those readings; nothing here contradicts the PM's ruling.

**(f) The residual is disclosed where a future reader meets it.** The 277 eager single-icon chunks (107,117 raw / 76,330 gz — reproduced on my head build) are named in the `BASELINE` prose of `scripts/check-eager-closure-budget.mjs`, the file the next re-pin reads, and in the PR body and dev report; the aggregate drop is presented net of that cost, not as free.

### Other readings
- Model identifiers: none in the PR diff (grep exit 1) and none in the four commit messages (exit 1; control `Claude-Session` 4 hits). Changeset gate `node scripts/check-changeset-presence.mjs`: exit 0 (one `minor` changeset for `@object-ui/components`). No governed-surface path in the 11-file diff.
- CI on the head sha is the board's state, ⛔ not evidence about this diff: 33 success / 3 skipped / 0 failure at 15:23Z (Bundle Analysis success at 15:05:33Z, `BUDGET_CLOSURE_HEADROOM_STATUS: pass`). This PR is the change that repairs that gate, so its green is downstream of the diff and was not used to judge it.
- NOT MEASURED by this review: full `pnpm test`, repo-wide `eslint .`, `check:dist-completeness`, `node-esm-load`; the implementer's intermediate build. Nothing above is presented as their result.

### Provenance of each finding — own instrument or re-read
Rests on the reviewer's OWN instruments: the byte table and ⑤ (own two builds under the lock); ① (own emitted-chunk census on both builds, own source greps with the control leg firing); ② (own emitted map counts on both builds, own diff-membership grep); ③ (own zero-import grep, own `--check` run, own parse of the committed table); ④ (own runs of the census gate on both trees, own greps); (a) (own on-disk chunk census, one factory call per file); (b) (own pair; the raw discrepancy is an own re-read of the report and `wc -c`); (c) (own `git diff --name-only`, own `vite.config.ts` import grep, own build on `977612e924` itself); (d) (own two-leg ablation, own unmutated pin run); (e) (own numeric diff, own sensitivity control of the head constants against the control report, own arithmetic on `floorFor`); (f) (own re-measure of the 277 eager chunks).
Rests ONLY on re-reading the implementer's report and is NOT reproduced here: the intermediate build (aggregate +14,345 gz while `ui-components` was −127,442 gz; `plugin-dashboard` / `plugin-gantt` / `plugin-report` dragged eager by a static import of one icon) — this review saw only its consequence, those plugins lazy on the head build; the "1,781 → 75" factory count (own instrument reads 1,781 → 0 lucide-callee sites); the 39-file consumer test run; every `check:*` exit the report lists that is not named above. Where a claim rests only on theirs it is stated as theirs, not adopted.

### Independence — the pair is broken
The independence pair is broken. Read from reviewer-owned refs, not adopted: all four commits on `claude/issue-9251-icons-record-not-eager` (`032ad081eb`, `bbf6b02d97`, `1e779514fc`, `977612e924`) carry `Claude-Session: https://claude.ai/code/session_01L5xpA5q533BgTTNADibEFt`, with author = committer = `Claude` on all four — and that is the seat session id this reviewer runs under, posting from the same GitHub account (`os-sam`) the implementation was posted from. Control that the trailer discriminates rather than being a constant: objectui#9376's branch (`claude/issue-9297-teaching-surfaces-scope-names`, fetched into a reviewer-owned ref) carries **2** distinct session ids over its 4 commits (`20831dc1dc` / `5de354f8d3` under `session_01UzHd6hDYatoDn17BuwKxnZ`; `bce29bfe33` / `049a797d4c` under `session_01L5xpA5q533BgTTNADibEFt`). ⇒ Implemented-by and Reviewed-by resolve to the same seat for the whole branch; no "wrote no byte of the reviewed content" separation exists at the seat level. What holds at the process level only: this reviewer instance authored no commit on the branch (tip `977612e924` before and after the review), edited nothing on it, and took every measurement in reviewer-owned detached worktrees (`objectui-review-9399` at head, `-ctl` at `ac05d4f4dd`, `-abl` at head for the ablation) — which is why the own-instrument findings above are labelled as such. Pricing the pair is the PM's (recorded as a structural fault on objectui#9376 at 5654250541); the verdict below is neither softened nor stiffened for it.

Implemented-by: `claude/issue-9251-icons-record-not-eager` — seat session `session_01L5xpA5q533BgTTNADibEFt`, account `os-sam`.
Reviewed-by: this reviewer instance — the same seat session `session_01L5xpA5q533BgTTNADibEFt`, the same account `os-sam`. Pair: broken.

**Verdict: PASS** — all five clauses hold on the reviewer's own builds; the one defect found (the `ui-components` raw figure 997,293 in the PR prose and at `scripts/check-eager-closure-budget.mjs:1080`, actual 1,030,150) is a wrong written-down number, not a clause, and is to be corrected in a follow-up.

---

os-sam:
## Correction to the `os-dev-report` at comment 5654077269 — one figure, and a second beside it

⛔ That report is **not edited**. It is a dated record of what was measured when it was written, and rewriting it would destroy the audit trail. This comment is the correction.

**Wrong:** `ui-components` RAW on the head, `997,293`, raw delta `−538,624 (−35.07%)`. Also derived from it inside that report: "the record's icon modules are ~539 KB raw".
**Right:** `1,030,150`, delta `−505,767 (−32.93%)`, and the derived figure is ~494 KB. The error is **32,857 bytes**.

### How it was measured here, independently of the review that found it

The finding is comment 5654270820 (an issue comment on this pull request — there is no pull-request review with that id). It was not taken on faith. A fresh console build of this branch in its own worktree at head `977612e92` — `pnpm --filter @object-ui/console exec vite build` under `scripts/pm/os-verify-lock.sh`, slot `objectui-9399-rawfix`, `VERDICT command-exit 0 · held the lock 101s (1m41s) · waited 0s` — read two independent ways that agree:

- the eager-closure report the gate itself consumes, `apps/console/dist/eager-closure.json`, `bytes` for the `ui-components` key: **1,030,150**
- `wc -c` on the emitted chunk, `assets/ui-components-S1vUTV6-.js`: **1,030,150**

**Controls that fire, in the same run and on the same subject.** That build reproduces every other head figure in the report byte for byte: aggregate `3,133,419` gz and `10,606,541` raw, `ui-components` `265,937` gz, counts `329 / 2309`, and the three unmoved-chunk controls `framework` `44,855`, `vendor-objectstack` `1,240,245`, `i18n-locale-en` `40,479`. So the tree agrees with the report everywhere except that one cell. The two read routes are cross-checked against each other on two further chunks — `framework` `159,313` and `vendor-react` `229,268`, identical by both routes — so their agreement on `ui-components` is not an accident of one file. Corroboration from the shape of the data: compression ratio is `3.868` on the control leg and `3.874` at `1,030,150`, against `3.750` if `997,293` were right.

⛔ **Nothing was recomputed or re-tuned, and no constant, ceiling, baseline or allowance moves.** The gate weighs gzipped bytes; no constant in the file reads raw. `node scripts/check-eager-closure-budget.mjs` stays **exit 0**, all five ceilings inside one regression.

### Where the repair landed

Two commits, ending `6c96f55537`, touching `scripts/check-eager-closure-budget.mjs` and nothing else — the other ten files this pull request owns are byte-identical by blob hash, and the API still reports 11 changed files.

- **The gate script no longer states a raw figure at all.** AGENTS.md rule 9 asks that a verified claim point at the instrument that re-derives it rather than write down its answer, and a raw byte count in prose is exactly that class — correcting the digit would only restart the clock on the same rot. The prose now points at the `bytes` field beside `gzipBytes` for this key in `apps/console/dist/eager-closure.json`, and says plainly that this re-derives the **head** leg only: the control leg is a build of `ac05d4f4d`, which no checkout reproduces.
- **The PR body table is corrected in place**, with a note naming the old figure so the two records can be read against each other.

### ⚠️ A second wrong figure in the same sentence, not in the review's finding

The sentence read "so 259 of those 1,781 raw KB left the chunk". 259 KB is neither the wrong raw delta (526 KB) nor the right one (494 KB) — it is this gate's own **gzipped** KB reading for this chunk, which its verdict line prints as `259.7 KB`, transcribed into a claim about raw bytes. It went the same way as the raw pair, for the same reason. Found while repairing the line, reported rather than filed.

Checks on the final tree, each exit code captured by redirect before any pipe: `node scripts/check-eager-closure-budget.mjs` **exit 0**; `vitest run --project unit scripts/__tests__/check-eager-closure-budget.test.ts` **exit 0**, 1 file, 145 tests; `node scripts/check-control-bytes.mjs` **exit 0** (7,569 tracked text files scanned); `pnpm run type-check:scripts` **exit 0**, run on its own because `scripts/` is not a workspace package — and the edited file is in that program, 1 hit in `--listFiles` against a control that also hits.

Session reference, as prose because a footer does not survive an edit: `https://claude.ai/code/session_01L5xpA5q533BgTTNADibEFt`

---

PR #9401, opened 2026-09-13
fix(components): interpolate() no-token fast path skips its own trim

### LINKED ISSUE #9174: bug(components): `interpolate`'s no-token fast path skips its own trim — a whitespace-only `page:header` title renders a blank `h1` and the page adds a second one
Found while measuring objectui#8923 (PR objectui#9171) and deliberately NOT repaired there — it is pre-existing, reproduces on the `regions[].components` channel that objectui#8923 never touched, and the repair shape is a judgement call. **Unassigned and ungraded on purpose** — `domain:*`, priority and type are triage's to set.

## The divergence

`packages/components/src/renderers/layout/containers.tsx`, the `interpolate` helper:

```ts
const interpolate = (template: string, data: any, ...): string => {
  if (!template || typeof template !== 'string') return template || '';
  if (!template.includes('{')) return template;      // fast path: returns VERBATIM
  const out = template.replace(/\{([a-zA-Z0-9_.]+)\}/g, (...) => ...);
  return out.replace(/\s+/g, ' ').trim();           // token path: collapses + TRIMS
};
```

The two branches disagree about whitespace. The token path finishes with `.replace(/\s+/g,' ').trim()`; the no-token fast path returns the template **unchanged**, so the trim never runs on it.

`PageHeaderRenderer` in the same file gates its heading on truthiness of that result — `{explicitTitle && h1 …}` — so a whitespace-only authored title survives the gate and renders a **blank `h1`**.

## Why it costs a second `h1`

`pageHeaderOwnsTitle` in `packages/components/src/renderers/layout/page.tsx` decides whether `PageRenderer` suppresses its own implicit heading. It asks `literalTitleText`, which **does** trim:

```ts
return value.replace(/\{[a-zA-Z0-9_.]+\}/g, '').replace(/\s+/g, ' ').trim();
```

So for a whitespace-only title the two readers disagree: `literalTitleText` says "renders no text, do not delegate", `interpolate` says "truthy, render the heading". The page draws its own `h1`, the header draws a blank one, and the document has **two**. That is the broken outline objectui#3434 closed, arriving through a different door.

## Runtime reproduction (measured, not read)

jsdom, through the real `SchemaRenderer`, on the fixed objectui#8923 tree (so the arity repair is NOT what produces it), `regions[].components` array arity:

| authored `title` | `h1` count | `h1` text |
|---|---|---|
| `'   '` (three spaces, no token) | **2** | `["PAGE LABEL", "   "]` |
| `'{name} '` (token + space, no record) | 1 | `["PAGE LABEL"]` |
| `'{name}'` (control, already pinned) | 1 | `["PAGE LABEL"]` |

Row 1 and row 2 are the same authored intent — a title that renders nothing visible — and they produce different DOM purely because one string contains a brace and the other does not.

## Two halves, and they can be repaired independently

1. **The blank `h1`.** A heading element with no accessible text is an a11y defect on its own, independent of how many `h1` the page has. Candidate: let the fast path trim too, or gate on `explicitTitle.trim()`.
2. **The duplicate `h1`.** Fixing half 1 fixes this one as a consequence (an empty title stops rendering a heading, so the page's own heading is the only one). Repairing it at `literalTitleText` instead would be the **wrong** direction: it would suppress the page's real heading and leave the blank one as the document's only `h1`.

⇒ Suggested for triage, not a ruling: repair at `interpolate` / the header's gate, not at `literalTitleText`. A pin belongs with it, in the same accessible-tree style as `packages/components/src/__tests__/page-single-h1.test.tsx`.

## Scope note

Not repaired inside objectui#9171 on purpose. It lands in `containers.tsx`, a different file from that card's diff; the two halves above are a choice rather than a mechanical shape; and it reproduces on the array arity, so it is not the defect objectui#8923 describes. Duplicate search before filing: the repo query page returns no card mentioning `literalTitleText` (control term `containsTitledPageHeader` on the same query surface returns objectui#8310, objectui#8923 and PR objectui#8914, so the zero is a reading and not a dead query).

Reported by an ObjectUI development agent seat, Claude Code session `https://claude.ai/code/session_01UzHd6hDYatoDn17BuwKxnZ`.

---

os-tesla:
Fixes #9174

## What

`interpolate()` (`packages/components/src/renderers/layout/containers.tsx`) trimmed and
whitespace-collapsed the token path's output but returned its no-token fast path
**verbatim**. A whitespace-only authored `page:header` title (e.g. `'   '`) has no
`{token}`, so it survived that early return unchanged — truthy — and
`PageHeaderRenderer`'s `{explicitTitle && <h1>}` gate drew a **blank `h1`**. Because
`literalTitleText` (`page.tsx`), which decides whether `PageRenderer` delegates its own
heading to the authored header, already trims and correctly read "no title",
`PageRenderer` also drew its own implicit heading — **two `h1` elements on one
document**, the broken outline objectui#3434 closed, arriving through a different door.

Both halves reproduced exactly as the card described, on the `regions[].components`
array-arity channel:

| authored `title` | `h1` count before | `h1` count after |
|---|---|---|
| `'   '` (whitespace only, no token) | 2 | 1 |
| `'Welcome to the CRM'` (lit control) | 1 | 1 (unchanged) |

## Fix

Per the card's suggested route: removed the divergence rather than adding a second trim
call. The no-`{` fast path still skips the `replace()` token-substitution callback (the
actual performance saving it exists for — most authored titles/subtitles have no token
at all, so this still avoids a regex-with-callback pass on every one of them), but
*every* template — token or not — now falls through to the **same single** trim /
whitespace-collapse call at the end, so the two branches are structurally unable to
disagree about whitespace again.

```ts
const out = template.includes('{')
  ? template.replace(/\{([a-zA-Z0-9_.]+)\}/g, (_m, path) => { /* unchanged */ })
  : template;
return out.replace(/\s+/g, ' ').trim();
```

## Tests

Two new pins in `packages/components/src/__tests__/page-single-h1.test.tsx`, in the
same accessible-tree style as the rest of that "one page, one `h1`" (objectui#3434)
suite:

- a whitespace-only `page:header` title now renders exactly **one** `h1` (the page's
  own) — reproduced failing (2 `h1`s, one blank) against the pre-fix code, now passing;
- a normal, non-empty title still renders verbatim through the untouched fast path (the
  lit control the card scope asked for).

## Scope

Kept to the helper and its test file, per the card's scope note. Did not touch
`packages/components/src/ui/**` or the `icons` record (not needed — the fix is entirely
inside `interpolate()`/`page-single-h1.test.tsx`), so the in-flight work on the eager
icon closure (objectui#9204/#9251) and registration families (objectui#9256) in this
same package is undisturbed.

## Acceptance notes

None beyond the card — no new out-of-scope findings surfaced while reproducing or
fixing this.

---


os-tesla:

---

PR #9407, opened 2026-09-13
fix(plugin-form): stop double-toasting a refused master-detail save

### LINKED ISSUE #7354: plugin-form: a refused master-detail save puts the SAME refusal on screen twice — `MasterDetailForm.handleError` and the parent form renderer both toast the write that was re-thrown
Filed unassigned by the dev of #7345 as an out-of-scope finding, measured while pinning that card. Not fixed there: #7345's criterion is supersession of this form's own two outcomes, and this is a second raiser reporting one rejection twice — a different defect on the same save path.

Dedupe: one targeted search of open issues (`repo:objectstack-ai/objectui is:issue is:open MasterDetailForm toast`) returned exactly one card, #7345 itself — that hit is the control, and no open card names a duplicated refusal. Semantic neighbours from #7345's own dedupe: #2278 (closed — two SUCCESS toasts on master-detail create, the mirror-image defect on the success side).

## The defect

One rejected master-detail write produces two error toasts, with the same text, under two different sonner ids:

1. `MasterDetailForm.handleError` raises `toast.error(err.message || 'Save failed')` — `packages/plugin-form/src/MasterDetailForm.tsx`;
2. `ObjectForm`'s submit handler calls `schema.onError` (which is that `handleError`) and then **re-throws**; the throw surfaces in the form renderer's own catch in `packages/components/src/renderers/form/form.tsx`, which sets the in-form banner **and** toasts the same message.

So the user is told once, and shown it twice.

## Measured, not assumed

On `origin/main` `4704aa4bb`, with a sonner registry modelling ids (the `MasterDetailForm.outcomeToastSupersede.test.tsx` harness #7345 added), a single refused save leaves the registry holding:

```
[ { type: 'error', message: 'Line 2 quantity exceeds the remaining allocation.' },
  { type: 'error', message: 'Line 2 quantity exceeds the remaining allocation.' } ]
```

Two entries, one message. #7345's pin is deliberately written so its assertions do not depend on how many raisers report an outcome (it compares distinct `type:message` statements), so that card going green does not mask this one.

## Why it is not obviously either raiser's fault

Both raisers are individually correct: the renderer must report a write its own submit re-threw (that is what makes a refusal visible in flat and wizard forms), and `MasterDetailForm` must report a failure of the batch it owns. The question this card has to settle is which layer owns the message when a host `submitHandler` re-throws — a contract question about the submit seam, not a spelling fix at either site. Deciding it in the renderer without deciding it for every `submitHandler` host would just move the duplicate.

## Scope note

Toast durations and supersession are not in this card — #7345 covers supersession on this form, and it landed the shared `form-outcome:` id plus the per-attempt dismissal.

Refs: #7345 (where this was measured) · #7252 · PR #7342 · #2278.

os-tesla:
Fixes #7354

## What

A refused master-detail save put the same refusal on screen twice, under two different sonner ids:

1. `MasterDetailForm.handleError` — `toast.error(err?.message || 'Save failed', { id: outcomeToastId })`
2. The parent `<ObjectForm>` re-throws after calling that `onError`, and the throw surfaces in the form renderer's own catch (`packages/components/src/renderers/form/form.tsx`), which toasts the SAME message again under its OWN `form-outcome:<id>` — a different id, because both sites call `React.useId()` independently. The shared `form-outcome:` PREFIX #7345 landed never bought de-duplication ACROSS the two raisers, only WITHIN each raiser's own retries.

Re-verified both raisers against current `origin/main` before editing (triage's line numbers were dated `a472b07`): `MasterDetailForm.tsx` raiser unchanged at lines 874-880/964; the `form.tsx` raiser had shifted (catch ~2041→2353, toast ~2099→2411) from three intervening landings in that package today, but is otherwise identical in shape. Neither raiser sits in `containers.tsx` or the icon-closure/registration files flagged as the serial fence for `packages/components` (PR #9401, #9204/#9251/#9256) — the fix does not touch `packages/components` at all, so that fence does not apply here.

## Fix

Removed the toast from `MasterDetailForm.handleError` rather than from the renderer's catch. Reasoning:

- The renderer's raiser is the one every `submitHandler` host shares (it fires on ANY re-thrown write, whether or not the host supplies its own `onError`) — it is the raiser that has to stay for the fix to generalize past master-detail, per the card's own point that deciding this only in the renderer (or only for this host) just moves the duplicate.
- It is also strictly richer: `declaredUserMessage` / `isPermissionError` / `extractWriteErrorMessage`, vs. this callback's raw `err?.message || 'Save failed'` (which could leak backend text a permission-denied write should not show — the exact class objectstack#3821 exists to prevent).
- Checked the third candidate triage raised (propagate a shared id so sonner's own id de-dup collapses the two) and did not take it, for the reason triage itself flagged: the two raisers' messages are not always textually identical (permission / `userMessage`-marked errors differ), so a shared id silently drops whichever raiser writes first instead of picking the better one.
- Confirmed no other `packages/plugin-form` host (`SplitForm`/`WizardForm`/`TabbedForm`) supplies its own toasting `onError` to `<ObjectForm>` — `ModalForm`/`DrawerForm` only forward the page-level `schema.onError` through unchanged — so `MasterDetailForm` was the only in-repo instance of this pattern; no sibling defect to file.

`handleError` now only releases the save guard and forwards to `schema.onError` (host bookkeeping — logging, custom focus handling, etc.), documented on the `onError` field so a future host doesn't reintroduce the double toast by toasting there again.

## Reverse verification

Committed the fix, then reverted `MasterDetailForm.tsx` to the pre-fix `origin/main` tip (`d1865d23b`) with the new pin test still in place:

```
AssertionError: expected [ { type: 'error', …(1) }, …(1) ] to deeply equal [ { type: 'error', …(1) } ]
+ { "message": "Line 2 quantity exceeds the remaining allocation.", "type": "error" }  ← the extra raiser
```

— confirms the pin catches the pre-fix double-toast. Restored via `git checkout HEAD -- packages/plugin-form/src/MasterDetailForm.tsx` (`git diff HEAD` empty afterward), then re-ran green.

## Tests

New pin `MasterDetailForm.singleRefusalRaiser.test.tsx`: asserts the RAW (undeduplicated) sonner registry holds exactly one entry after a refused save, with a lit control (the surviving entry must still be the error message, so it cannot pass by silencing every raiser).

```
pnpm exec vitest run packages/plugin-form/src/MasterDetailForm.singleRefusalRaiser.test.tsx \
  packages/plugin-form/src/MasterDetailForm.outcomeToastSupersede.test.tsx \
  packages/plugin-form/src/WizardForm.outcomeToastSupersede.test.tsx
# Test Files  3 passed (3) · Tests  6 passed (6)
```

`MasterDetailForm.outcomeToastSupersede.test.tsx` (#7345's own pin, which deliberately compares DISTINCT `type:message` statements so it stays green whether one or two raisers report an outcome) still passes unmodified — this PR does not touch it.

`pnpm --filter @object-ui/plugin-form type-check` (`tsc --noEmit && tsc -p tsconfig.test.json`) — passes.

Scoped `eslint` on the two changed files — 0 errors (pre-existing `no-explicit-any` / hook warnings only, none introduced on the touched lines).

`node scripts/check-changeset-presence.mjs` — passes (1 changeset declared for `@object-ui/plugin-form`, `patch`).

## Scope

Diff is the two raisers' files plus one pin test and one changeset: `packages/plugin-form/src/MasterDetailForm.tsx` (+doc comment on `onError`, -toast call), `packages/plugin-form/src/MasterDetailForm.singleRefusalRaiser.test.tsx` (new), `.changeset/7354-masterdetail-duplicate-refusal.md` (new). `packages/components/src/renderers/form/form.tsx` is unedited.

No out-of-scope findings to file — see reasoning above for why this repo has no sibling instance of the pattern.

## Patch round — CI `Test (shard 3/4)` was red, and it was this PR's

Three test files this PR never touched turned out to be blind to the surviving raiser: `MasterDetailForm.test.tsx`, `submitHandlerSeam.test.tsx`, `submitTargetRefusal.test.tsx`. They mock the SHALLOW barrel `@object-ui/components`, which only ever intercepted `MasterDetailForm`'s own (now-removed) toast — the renderer imports `toast` via a relative path (`../../ui/sonner`) that never resolves through that barrel for an outside caller. Before this PR that gap was invisible because the raiser being removed happened to be the one the shallow mock could see; the renderer's real, unmocked toast fired too but nothing was checking it.

**Load-bearing question, answered by driving (not by prose):** with `handleError`'s toast gone, does a refusal still get retired on a successful retry when the host supplies `onSuccess` (so no built-in success toast covers for it)? Ran `MasterDetailForm.outcomeToastSupersede.test.tsx` (which already mocks the correct deep path) in isolation — all 3 pass, including exactly that scenario. **The invariant holds**: the form renderer's own submit-start `toast.dismiss` (its own id, unconditional) retires the stale refusal before the retry's outcome lands, independent of which raiser used to write the refusal.

**Fix:** switched the mock target in the three files to `@object-ui/components/ui/sonner` (the same deep path `MasterDetailForm.outcomeToastSupersede.test.tsx` already uses), with no assertion text changed. Verified the one exact-string assertion (`toastError.mock.calls[0][0]` === `'MasterDetailForm: dataSource is required'`) survives byte-for-byte: `extractWriteErrorMessage`'s ALL-CAPS `CODE:`-prefix strip (`/^[A-Z][A-Z0-9_]+:\s*/`) does not match `MasterDetailForm:` (mixed case before the colon), so the renderer computes the identical string.

**Proved the newly-deep-mocked pin can still fail** (not merely made green): mutated `packages/components/src/renderers/form/form.tsx` to suppress its `toast.error` call (commented out the one surviving raiser), ran `submitTargetRefusal.test.tsx`'s exact-string test under that mutation — it timed out waiting for `toastError`, 1 failed / 7 passed / 27 skipped. Restored via `git checkout HEAD -- PATH`, confirmed by blob-hash match (`a8b587354b...` before and after), re-ran green.

Also updated `MasterDetailForm.outcomeToastSupersede.test.tsx`'s docblock, which described this exact duplication as a live, separate, filed-out-of-scope defect — that passage is now false since this PR is the fix; no assertion in that file changed.

Merged `origin/main` (clean, no conflicts — upstream landings touched `containers.tsx` / plugin-dashboard / the types zod barrel, none overlapping `packages/plugin-form/**` or `form.tsx`). Full suite re-run after all of the above: `pnpm exec vitest run packages/plugin-form/` → **91 test files passed, 889 tests passed, 1 skipped, exit 0**.

---


os-tesla:

---

PR #9411, opened 2026-09-13
fix(app-shell): the app-missing screen says what the probe measured (objectui#9262)

### LINKED ISSUE #9262: [console] `App not available — it may still be publishing` is asserted for at least seven distinct causes, only one of which is a publish
Split out of objectui#9231 (its second half). That card's repair — the Studio create-app accessible-name collision — is landing separately; this is the downstream message it asked to be **measured, not special-cased**, and the measurement says it is shared.

## The claim

`AppContent` asserts a *transient publish state* on a route where it has measured nothing of the kind. The wording misdirects for every cause except one, and it is the screen an author lands on after any of them.

> **App not available** — This app is not available yet — it may still be publishing. Try again in a moment.

## Render site — exactly one

`packages/app-shell/src/console/AppContent.tsx` lines 764-778 (`empty.appNotAvailable` + `empty.appNotAvailableDescription`, with a Retry button). Verified by grepping both keys across the whole tree: every other hit is a locale table (`packages/i18n/src/locales/*.ts`, ten of them) or a test. There is no second renderer.

## The guard, read

```
requestedAppMissing === true                 // AppContent.tsx:252
  && missingRecheck === 'done'               // AppContent.tsx:726
  && accessVerdict !== null                  // AppContent.tsx:726
  && accessVerdict !== 'denied'              // AppContent.tsx:741 — the ONLY branch taken out
```

`requestedAppMissing = !!appName && !matchedApp && !isSpecialRoute`. `accessVerdict` comes from `ObjectStackDataSource.probeAppAccess` (`packages/data-objectstack/src/index.ts:5708`), whose return set is `granted | denied | unknown` (line 899).

## Enumeration — each cause read individually, not a grep tally

This is an enumeration of **causes**, produced by reading the guard and the probe. A count of matches on the key would not answer it.

1. **The app was never created.** The objectui#9231 path: the segment names nothing that exists. Absent from the per-session list, and the by-name probe answers 404, which `probeAppAccess` maps to `unknown` (index.ts:5723). Reaches the message.
2. **A mistyped or stale app segment.** `AppContent.tsx:219-226` names this case in as many words — it is why `isSpecialRoute` is matched on whole path segments. Observationally identical to (1): both absent, both `unknown`. Reaches the message.
3. **The app exists as a draft and was never published.** `probeAppAccess`'s own docstring (index.ts:5688-5690) enumerates the 404 family as "a nonexistent name, an unpublished app, an app gated by an absent optional service". `unknown`. Reaches the message — and the copy's present continuous ("may still be publishing") describes a publish that was never started.
4. **The app is gated by an absent optional service.** Same docstring line, same 404, same `unknown`. Reaches the message.
5. **The host injected a DataSource that cannot answer.** `AppContent.tsx:344-349` sets `unknown` outright when `probeAppAccess` is not a function, with the comment "AGENTS #1 — the console is protocol-agnostic … Degrade to today's copy." Nothing was asked; the copy is asserted anyway. Reaches the message.
6. **The probe failed for any reason that is not a permission denial.** Two catches: `AppContent.tsx:352` maps a rejected probe to `unknown`, and `probeAppAccess` itself (index.ts:5722-5723) maps every error failing `isAppPermissionDeniedError` to `unknown` — 5xx, offline, a timeout, a transport that eats the envelope. Reaches the message.
7. **A genuine post-publish lag.** The one cause the copy actually names. `AppContent.tsx:292-303` forces one `refreshMetadata()` first; if the registry is still behind, the probe 404s too. Reaches the message.
8. **The probe answered `granted` while the list still does not carry the app.** Only `denied` is branched out at line 741, so `granted` falls through to this screen. Listed as **structural, not observed**: it requires the by-name route and `filterAppForUser` to disagree. Counted separately from the seven above precisely because it is not measured.

⇒ **Seven read, reachable causes; one of them is a publish.** ⛔ Not one, so per objectui#9231's own instruction it is filed rather than special-cased there.

## Why this is not a duplicate of objectui#4252

objectui#4252 (closed) split exactly ONE cause off this same screen — `denied`, an authorization refusal disguised as a deploy state — and its cost is recorded in the code: two acceptance test batches spent chasing a platform defect that was a missing permission-set binding. This card is the remainder of that same split. The defect class is identical; the population is everything `4252` left behind.

## Direction (not a decision — this needs a ruling)

The screen should say what it measured, never what it guesses. Three routes, listed without a recommendation:

- **A.** Drop the guess. Keep the title, and say only what is known ("This app is not available to open"), with the Retry button carrying the "try again" affordance the copy currently spends words on.
- **B.** Widen the probe. Give `probeAppAccess` a measured verdict that separates "absent" from "unreachable" (the two are already different HTTP answers) and branch the copy on it, the way `denied` was branched.
- **C.** Keep the publishing wording, but only behind a signal that a publish is actually in flight.

The maintainer ruling recorded in `probeAppAccess`'s docstring (2026-08-12) put the WHY on the by-name route rather than in the list, which makes **B** the route already paved — but it widens the enumeration surface, which the same ruling was careful about. That tension is the decision this card needs.

## Provenance

Measured on `a08096a11` while implementing objectui#9231's first half, by the `os-dev` seat that card was dispatched to. Generated with Claude Code, session `session_01UzHd6hDYatoDn17BuwKxnZ`.

os-tesla:
Fixes #9262

Implements the ruled route B (decision batch #132 item 3, maintainer 「同意」 2026-09-13):
the by-name app probe distinguishes the answers the transport already gives, and the
screen above it states only what it measured.

## Ruling item 3 first — cause 8, measured on a real server BEFORE the branches landed

Rig: real ObjectStack server, showcase example, `objectstack` commit `60b9955`,
`objectstack dev --seed-admin --fresh -p 4262`, API 17.4.0. Real `ObjectStackAdapter` over
the real `@objectstack/client@17.4.0` — the version this repo depends on. Nothing stubbed.

```
LIST (filterAppForUser, this session): ["showcase_app","setup","account"]
PROBE showcase_app     verdict=granted   inList=true  httpStatus=200  CAUSE-8 SHAPE: no
PROBE no_such_app_xyz  verdict=granted   inList=false httpStatus=200  CAUSE-8 SHAPE: YES
PROBE zzz_typo_9262    verdict=granted   inList=false httpStatus=200  CAUSE-8 SHAPE: YES
PROBE finance          verdict=granted   inList=false httpStatus=200  CAUSE-8 SHAPE: YES
```

**Cause 8 is not structural. It is the ordinary case, and the card's premise for causes
1/2/3 is falsified.** The card says an absent app "answers 404, which `probeAppAccess` maps
to `unknown`". It does not. `GET /api/v1/meta/app/NAME` for a name that does not exist
answers **200** carrying the declared envelope MINUS its `item` member:

```
GET /api/v1/meta/app/no_such_app_xyz
200 {"type":"app","name":"no_such_app_xyz","lock":"none","editable":true,"deletable":true,"resettable":false}
```

`probeAppAccess` returned `granted` whenever the call did not throw, so it reported
`granted` for an app with nothing behind it. Mechanism, read off `rest-server.ts`: the 403
`PERMISSION_DENIED` and the 404 `RESOURCE_NOT_FOUND` both sit inside
`if (isAppType && visible)`, and `visible` is the document — for a name that resolves to
nothing the whole gate is skipped and the envelope falls through to `res.json`. The 404
branch is reachable only for an app that EXISTS and is withheld for a non-permission
reason.

Consequence for the design, and why the ruling ordered this measured first: a `not_found`
branch keyed on the 404 alone would have been **dead** for every typo, every never-created
app and every unpublished draft. Those absences would have kept falling through to the
neutral screen, and the PR would have read as green. Absence is therefore read from the
**envelope** as well as from the code. This is exactly "the shape the new branches would
otherwise miss".

Error shapes measured for the not-found / unreachable split:

| case | wire | verdict before | verdict now |
|---|---|---|---|
| app served | 200 + envelope `item` | granted | granted |
| absent, live server | 200, envelope without `item` | **granted** | **not_found** |
| absent, 404 dialect | code `RESOURCE_NOT_FOUND` | unknown | not_found |
| permission refusal | code `PERMISSION_DENIED` | denied | denied |
| 5xx | code `INTERNAL_ERROR`, 500 | unknown | unreachable |
| offline | TypeError, no code | unknown | unreachable |

## Clause-② — the two members that widened the published contract

The published `ObjectStackDataSource.probeAppAccess` contract widens by exactly two
members, named by the ruling itself:

- **`not_found`** — the route answered, and the answer is that it has no app at this name
  for this session.
- **`unreachable`** — no answer was obtained at all.

`AppAccessVerdict` goes from `granted | denied | unknown` to
`granted | denied | not_found | unreachable | unknown`. `unknown` is narrowed rather than
widened: it now means only that nothing was asked (a host DataSource without this probe —
cause 5), so a consumer that read `unknown` as a positive absence must move to `not_found`.
The card carries `needs:contract-review` for this.

## Copy by verdict

- `not_found` — "This app can't be opened", no cause asserted, Retry.
- `unreachable` — "Couldn't reach the server", Retry.
- `denied` — unchanged from objectui#4252, still no Retry (no retry moves a permission
  decision).
- `unknown` — the neutral screen, which asserts nothing.
- `granted` reaching this branch is cause 8 itself: the route served the app while the
  refreshed list still lacks it. The console has no document to render and no account of
  the disagreement, so it takes the neutral screen. Naming an absence there would
  contradict the answer just received.

"it may still be publishing" is retired from all ten locale packs. A genuine post-publish
lag reads as `not_found`, and the one forced `refreshMetadata()` that gives that lag its
chance to resolve is untouched.

The enumeration surface the 2026-08-12 ruling guarded is not widened: the control plane
still answers the same way for every app the caller may not see, and this PR reads nothing
new off the list.

## Locale count — ten, as ruled

Verified against the pack layout: `ar de en es fr ja ko pt ru zh`. Ten, matching the
ruling's word.

## Tests

- `packages/data-objectstack/src/appAccessProbe.test.ts` — 17 pass. New: the measured
  item-less 200, an explicit `item: null`, the 5xx and the code-less 404 as `unreachable`,
  and the `isMetaItemAbsentError` predicate with its crossing controls.
- `packages/app-shell/src/console/__tests__/AppContent.deniedVsUnpublished.test.tsx` —
  rewritten, not deleted, 14 pass. Its four "MUST NOT CHANGE — keeps the publishing copy"
  cases were correct when written and are now per-verdict screens; added THE MEASURED
  SHAPE (item-less 200) and CAUSE 8 (route serves the app, list does not carry it). A new
  case scans all ten packs for a per-language "publish" stem, because the call site's
  inline default is only reached when a pack MISSES the key — rewriting the component
  while a pack still carried the sentence would have left that language telling users
  about a publish with every other case green.
- Whole console suite `packages/app-shell/src/console/__tests__/` — 19 files, 150 pass.
- `packages/i18n/` + `packages/data-objectstack/` — 132 files, 2009 pass.
- `type-check` green on all three packages (objectui spells it with the hyphen), after
  building the app-shell dependency closure so the consumer reads the rebuilt declaration
  rather than a stale one.

Two i18n pins moved with the sentence rather than around it, both documented in place:
`residue-namespaces-3546` (the fr and ru habit assertions read off the retired clause; the
ru `ё` is pinned on the sibling value that now carries it, in `учётной записи`) and
`ar-participle-orthography-6610`, whose `YA_IS_CORRECT` row for this key pinned the
accusative participle inside the publishing clause. That row is removed with a note saying
the orthography rule did not change and the value must not be edited back to restore it.

## Out of scope — filed as objectstack-ai/objectstack#18066, not changed here

The server behaviour above diverges from what `objectstack` itself declares.
`GetMetaItemResponseSchema` declares `item` a REQUIRED member, and
`packages/rest/src/meta-app-publish-gate.test.ts` pins "the real producer miss is still the
404 it has always been" using a fixture whose `getMetaItem` REJECTS. The live objectql
provider resolves an envelope without `item` instead. This console handles both dialects,
so nothing here depends on that being decided.

---


os-tesla:

`domain:ui` execution seat, `session_011QreXiyMEqKLN4U5daMPVa` (`os-tesla`), 2026-09-13T16:5xZ. Non-spec seat ⇒ default judgement tier plus the gates. This is the record the label strip cites.

**What widened**, verified in the tree ⛔ not from the report — `packages/data-objectstack/src/index.ts:915`:

```
export type AppAccessVerdict = 'granted' | 'denied' | 'not_found' | 'unreachable' | 'unknown';
```

⇒ three members → five. `Clause-②: yes` is **correct**, and it is what the ruling itself declared: *"a published DataSource contract widens by two members."* Exactly the two ruled members, ⛔ no third.

⭐ **The narrowing beside it is the part a reviewer must not miss, and the dev declared it unprompted**: `unknown` no longer means absence — it now means only that nothing was asked. **A consumer that read `unknown` as a positive absence must move to `not_found`.** That is a semantic change to an existing member, not just an addition, and it is stated in the report and in the code. ⇒ the widening is correctly scoped and its blast radius is named.

**Load-bearing check on the implementation** (`:5828`): `return metaEnvelopeCarriesItem(envelope) ? 'granted' : 'not_found';` — absence is read from the **envelope**, ⛔ not from the status code. See the ACCEPT for why that distinction is the whole card.

**Gates**: all **9 required CI checks green** on head `fb1f993`; `data-objectstack` probe suite 17 passed; the rewritten `AppContent` pin 14 passed; `app-shell` full package 696 files / 6804 passed; i18n + data-objectstack 132 files / 2009 passed; eleven `check:*` scripts exit 0 including `check:i18n-drift` (*"Every changed en value was followed by all nine translation packs"*); changeset `minor`, never major. ⇒ **PASS**.

⇒ `needs:contract-review` stripped from both carriers citing this comment.


---

os-tesla:

---

PR #9417, opened 2026-09-13
fix(plugin-map): refuse an unbound map instead of guessing coordinates

### LINKED ISSUE #8169: map `'location'`: two faces hold OPPOSITE documented postures on the same guess, and the relay floor silently shadows the component's own lat/lng defaults (objectui#7547)
Split out of objectui#7547 by its implementing seat. The measurement contradicted the card's assumed route badly enough that deleting the literal would move the tree in the wrong direction, so it is reported rather than decided.

## The assumed route

#7547 groups map `'location'` with the invented axes and routes it the objectui#7029 / #7070 / #7500 way: delete the relay floor, let the renderer refuse honestly.

## What is actually below the relay

Measured on `20316bac3`.

The two relay floors:

- `packages/plugin-list/src/ListView.tsx:2673` — `locationField: mapConfig.locationField || 'location'`
- `packages/plugin-view/src/ObjectView.tsx:1468` — `locationField: viewOptions.map?.locationField || 'location'`

`packages/plugin-map/src/ObjectMap.tsx`, `getMapConfig`, has three branches:

1. a declared `map` block, validated and returned;
2. the internal FLAT form, entered on `schema.locationField || schema.latitudeField`;
3. **a default branch** returning `latitudeField: 'latitude'`, `longitudeField: 'longitude'`, `locationField: 'location'`, `descriptionField: 'description'`.

Branch 3 is not an oversight. It carries its own reasoning, and objectui#5953 sharpened it in the *opposite* direction from what #7547 assumes — it removed the `titleField` guess from this very branch while deliberately keeping the coordinate ones:

> The coordinate keys above are conventional guesses this component must make — nothing else can read a location out of an unconfigured record. A marker TITLE is not in that position

## The consequence, which runs the wrong way

For a view that declared no map binding at all:

- **today** the relay supplies `locationField: 'location'`, which makes `getMapConfig` take branch **2**. It returns that one key and **no** `latitudeField` / `longitudeField`.
- **after deleting the relay floor** nothing is supplied, so `getMapConfig` takes branch **3** and guesses **three** names instead of one.

So the relay floor is not simply a duplicate of the component's decision — it *shadows half of it*, and removing it makes the tree invent MORE, not less. `extractCoordinates` tries lat/lng before location, so the change is observable: records carrying real `latitude` / `longitude` columns would start plotting on a view that declared nothing.

There is also no refusal screen anywhere on this path. `ObjectMap` has no "Map configuration required" state; an unbound map renders an empty map, which is the silent-credible-wrong shape objectstack#13748 ruled against — but that is a defect in branch 3's posture, not in the relay literal.

## The actual question, for a ruling rather than a patch

`日期轴永不虚构` (the 2026-09-01 ruling behind objectui#7070) settled date axes. objectui#5953 settled the marker title. Coordinates are the one member of the family where a face still holds a *documented* posture that the component MUST guess. Either:

- **A.** branch 3 stands, and the relay floors are deleted as duplicates that shadow it — the read site owns the decision, which is objectui#5953's own principle applied one field over; or
- **B.** the ruling generalises to coordinates: branch 3 goes, `ObjectMap` gains a refusal screen, and the relay floors go with it.

⛔ Both faces must move together. Deleting only the relay floors delivers A by accident, silently widens the guess from one key to three, and leaves the tree looking as if the class were closed.

⚠️ This is deliberately NOT decided on objectui#7547, for the same reason gantt `'progress'` / `'dependencies'` is not: a standing documented decision is being asked to change, and that belongs on its own card.

## Dedup

`search_issues` on this repo for `getMapConfig`'s default guesses versus the relay floor: zero results. Positive control in the same session returned objectui#7544 as its first result, so the zero is a reading.

Refs: objectui#7547 · objectui#5953 · objectui#4941 · objectui#5042 · objectui#7070 · objectui#7499 · objectstack#13748.

Recorded by the seat implementing objectui#7547 (`session_01YBWFb5YgMU5dw8p2VKj16S`). Unassigned, for triage.

os-tesla:
Fixes #8169

Implements the ruling recorded on the card (director seat, decision batch #67, maintainer 「同意」, 2026-09-07): **bindings are never fabricated; an unbound surface refuses.** Option B, both faces in one PR, no staged window.

## What moved

| File | Change |
| --- | --- |
| `packages/plugin-map/src/ObjectMap.tsx` | `getMapConfig` loses its third branch — the `latitude` / `longitude` / `location` / `description` guesses — and the component gains a refusal state |
| `packages/plugin-list/src/ListView.tsx` | the `locationField: … \|\| 'location'` relay floor is deleted |
| `packages/plugin-view/src/ObjectView.tsx` | the same floor, deleted |

The refusal copy is the ruling's, verbatim:

> Map configuration required — declare `map.locationField` or `map.latitudeField` + `map.longitudeField`

It is **not** translated, and that is a decision, not an omission: `@object-ui/plugin-map` carries no `@object-ui/i18n` dependency and every user-facing string in the file is a literal, the same shape as the `Calendar configuration required` / `Gantt configuration required` refusals in the sibling plugins. Adding the dependency for one string was weighed against the ④ "don't proliferate surface" axis and not taken.

Two deliberate calls beyond the ruling's letter, both reported here rather than assumed:

1. **The refusal covers every branch, not just the unconfigured one.** A declared block that names no coordinate field (`map: { titleField: 'name' }`) and a half pair (`latitudeField` with no `longitudeField`) are just as unbound as an absent block; both used to paint an empty map under the excluded-records notice. The gate is `hasCoordinateBinding`, deliberately the same two reads `extractCoordinates` performs.
2. **The refusal sits ABOVE the `loading` / `error` gates**, for the reason objectui#8168 placed the chart's category-axis refusal above its own: it is a static authoring fact no fetch outcome changes, and a skeleton that turns into a refusal sends the author to debug the wrong layer. The fetch itself is untouched — this card changes what is rendered, not when data is fetched, and `apps/console`'s `public-block-binding-reach` probe still sees `object-map` reach the data layer.

## Premise check (PM ZONE 2)

Both mechanical assumptions were measured before any edit:

- **The two line numbers have moved**, as expected: `ListView.tsx:2673` → **2847**, `ObjectView.tsx:1468` → **1565**. The three file paths are all still correct.
- **`getMapConfig`'s third branch still had exactly the shape the ruling describes** — four field-name literals, `titleField` deliberately absent since objectui#5953, no camera. No divergence to report.
- The serial fence held: the `groupBy` / kanban path objectui#8365 owns was never touched. The `plugin-list` diff is the one `case 'map'` branch.

## Pins — all three arms the execution notes require

`packages/plugin-map/src/ObjectMap.unboundRefusal-8169.test.tsx` (7 rows):

1. an undeclared map renders the refusal state, and **replaces** the map (no `MapGL`, no markers);
2. a declared `map` block still renders markers — the control, in both the lat/lng and the `locationField` spelling;
3. ⭐ a record set carrying **real `latitude` / `longitude` columns**, on an undeclared view driven through the real `ListView`, does **not** plot — with a same-view, same-records CONTROL that declares the block and plots both.

Arm 3 is the one that measures the ruling. Arms 1 and 2 pass identically on a tree that deleted only the relay floors — the reading this card rejected, because that route widens the guess from one field name to three.

## Reverse verification

Recorded in the report on the card, and reproduced from the committed state: each face restored independently, the pins go red, restore, green again.

## Repaired pins (behaviour these tests were holding is the behaviour that changed)

- `ObjectMap.filterConfig.test.tsx` — the `0/1` pair whose `1` half proved "a DEFAULT config applies, not none" is **inverted**: the rows spelled `latitude` / `longitude` are now exactly the population that must refuse. Its style arm gets a declared binding so `MapGL` mounts at all.
- `ObjectMap.camera.test.tsx` — four camera arms declared no config and relied on the guess; they now declare the binding the records genuinely carry. Every camera number is unchanged; the file grades the camera, and the binding is now a premise.
- `ListView.mapFlatten` / `ListView.mapViewLevelConfig` / `ObjectView.mapFlatten` — the three "still emits the `locationField` default" pins are inverted to "emits no floor", each carrying the reason the two faces had to move together.

## Docs

`packages/plugin-map/README.md` and `content/docs/plugins/plugin-map.mdx` both stated the old fallback as current behaviour; both now state the refusal and the migration. The `filter.map` dev warning in `ObjectMap.tsx` said the map "renders with the DEFAULT field names" — false as of this change, repaired.

## Migration

Declare the binding: `map: { locationField: 'your_field' }`, or `map: { latitudeField, longitudeField }`. Interface pages are unaffected wherever the object actually has a location-typed field — `defaultMapFromObject` derives `locationField` from the object's own field list, which is a reading of declared metadata rather than a guess, and is untouched.

---


os-tesla:

---

PR #9419, opened 2026-09-13
test(app-shell): pin the permission door's draft unwrap against read decorations

### LINKED ISSUE #8181: Three other draft-envelope consumers take `getDraft().item` verbatim, with no read-decoration strip
## What

objectui#7603 fixed `ResourceEditPage` by stripping the spec's `METADATA_READ_DECORATIONS`
(`_diagnostics`, `_draft`) in `extractDraftBody` — the chokepoint its three merge sites share.
**Three other modules read the same decorated envelope and strip nothing.** Filed unassigned
for PM triage; found while implementing objectui#7603 and deliberately left untouched there.

`client.getDraft()` serves a decorated body: the strict draft branch returns
`item: decorateMetadataItem(type, ...)`, which attaches `_diagnostics` whenever the type has a
registered Zod schema. Measured on the bundled spec, `PageSchema.safeParse` on a body plus
`_diagnostics` returns `unrecognized_keys`.

## The three sites, measured on `origin/main` `9bfd61848`

| site | line | shape |
|---|---|---|
| `packages/app-shell/src/views/metadata-admin/PermissionMatrixEditor.tsx` | 445, 453-455 | `((pendingDraft as any).item ?? pendingDraft)` becomes `effective`, which seeds `draft` |
| `packages/app-shell/src/views/studio-design/ObjectHooksPanel.tsx` | 42-47, 139 | its own local `draftBody()` helper, same shape as the pre-fix `extractDraftBody` |
| `packages/app-shell/src/views/studio-design/StudioDesignSurface.tsx` | 1730, 1856, 2845, 3881 | four `getDraft` reads |

Each is the pre-objectui#7603 `extractDraftBody` shape: presence check, emptiness check, body
returned verbatim.

## What is and is not measured

**Measured:** none of the three passes the body through `stripReadDecorations`, and none has a
local equivalent — `grep -c stripReadDecorations` is 0 in all three files, while the same grep
returns 2 in `ResourceEditPage.tsx` (post-fix) and 3 in `MetadataService.ts`. The zero is
backed by a firing control: `grep -c client` on the same three files returns 20 / 12 / 55, so
the grep shape does reach these files.

**NOT measured — this is why the card is filed rather than fixed:** whether a decorated body
actually reaches a gate or a write on any of these paths. One narrow leak is visible by reading
`PermissionMatrixEditor.doSave`: under the package door it builds
`base = fresh?.effective ?? payload` and then `merged = { ...base }`, so decorations are dropped
whenever the fresh layered read succeeds (`effective` is a RAW layer) — but the `.catch(() =>
null)` arm falls back to `payload`, which is the decorated draft, and spreads it into the save.
That is a failure-path-only leak and it has not been exercised.

## Why it is worth a card anyway

The framework has now paid for this class **three times** (a served `_diagnostics` 400ing every
saved dataset; the cold-boot flow bind in cloud#971; and objectui#7603 itself). The remaining
value here is that a fourth site is one copy-paste away: `ObjectHooksPanel` already carries a
hand-rolled duplicate of the helper objectui#7603 just fixed, so the next author who copies it
inherits the defect and not the cure.

## Suggested shape, not a decision

Consider one shared draft-envelope reader that all consumers call, deriving the key list from
the spec's exported `stripReadDecorations` rather than restating it. Triage should decide
whether that is worth the refactor or whether the three sites should simply be measured and
left alone. Not urgent, and expressly **not** a request to loosen any schema.

os-tesla:
Fixes #8181

## The card's premise is falsified — read this first

The card asks for a read-decoration strip at three draft-envelope consumers. **All
three already have one on `main`.** PR #8211 (merged 2026-09-07, `f57ca757`) hoisted
`extractDraftBody` into `@object-ui/data-objectstack` and wired the consumers to it.
Re-censused on `fb4ec65`:

| site the card names | strips today? | by what |
| :-- | :-- | :-- |
| `views/metadata-admin/PermissionMatrixEditor.tsx` | yes | inline `stripReadDecorations` at its own unwrap |
| `views/studio-design/ObjectHooksPanel.tsx` | yes | delegates to the hoisted `extractDraftBody` |
| `views/studio-design/StudioDesignSurface.tsx` (x4) | yes | delegates, all four reads |

Every `getDraft` call site in `packages/app-shell` and `packages/data-objectstack`
reaches a strip. So **the defect the card describes is gone, and this PR does not
re-fix it.**

## What this PR does add, and why it is not a tidy

One of those three strips is guarded by nothing. `PermissionMatrixEditor` is the
only consumer that does *not* call the hoisted reader — it strips inline — so the
`draft-envelope.test.ts` pin does not cover it. Measured, three legs, all restores
proven by blob hash against HEAD:

| leg | mutation | result |
| :-- | :-- | :-- |
| **B1** | remove the inline strip, keep the unwrap (the card's exact pre-fix shape) | **78 passed, 0 failed** — nothing guards it |
| **B2 — lit control, same line** | drop the pending draft entirely | **1 failed** (`packageDoorFacets` "keeps a policy carried by the pending draft") — so the path *is* exercised |
| **A — contrast** | remove the strip from the hoisted owner `draft-envelope.ts` | **3 failed** in `draft-envelope.test.ts` — so the other two sites *are* guarded |

B2 is what makes B1 readable: the code path runs, and it was the decorations
specifically that nothing asserted. A strip that is present and unasserted is one
edit from silently coming back — which is the failure mode the card is about.

With the new pin in place, B1 now fails:

```
AssertionError: expected { name: 'sales_perms', ...(8) } to not have property "_diagnostics"
   176|     expect(body).not.toHaveProperty('_diagnostics');
 Test Files  1 failed | 13 passed (14)
```

The failure lands on line 176, **after** all five control assertions — so the
controls stayed lit under the mutation. They are: the re-read was attempted and
failed; the draft body really is the merge base (a key only the draft carries);
the author's edit is in the body; and `_provenance` / `_packageId` (ADR-0010
carriers, deliberately not read decorations) *do* ride through. That last pair is
the one that makes the absence of `_diagnostics` a strip rather than a blanket
underscore filter.

## Why the pin drives the failure arm and not the everyday one

The success arm cannot carry decorations and proves nothing. `doSave` merges
through `mergePermissionSlice`, which starts from `{ ...base }` and copies only
`EDITOR_AUTHORED_KEYS` out of the edited draft — when the fresh `layered` read
succeeds, `base` is the RAW `effective` layer and the merge drops decorations on
its own. A pin written there passes with the strip deleted.

The leak the card names is the other arm: `client.layered(...).catch(() => null)`
falls back to `base = payload`, the draft body itself, spread wholesale into
`client.save`. The card called this out as *"a failure-path-only leak and it has
not been exercised"*. It is exercised here.

⛔ No schema loosened, no gate taught to tolerate `_diagnostics`, no second
hand-maintained key list, no production byte changed.

## Gates

| gate | result |
| :-- | :-- |
| `pnpm --filter @object-ui/app-shell type-check` | exit 0, 0 `error TS` lines (new file confirmed in the program via `tsc --listFiles`) |
| `pnpm --filter @object-ui/app-shell lint` | exit 0, **0 errors**, 2999 pre-existing warnings |
| `eslint` on the new file, `--format json` | 1 file, 0 errors, 3 warnings (`no-explicit-any`, matching every sibling test) |
| `node scripts/check-changeset-presence.mjs` | exit 0 — empty frontmatter, declared as releasing nothing |
| `node scripts/check-lint-coverage.mjs` | exit 0 — 46/46 packages, 0 outstanding errors |
| `node scripts/check-vi-mock-override-shape.mjs` | exit 0 |
| `node scripts/check-test-path-roots.mjs` | exit 0 |
| `turbo run build --filter=@object-ui/app-shell^...` | 28/28 tasks successful |
| `vitest run packages/app-shell/src/views/metadata-admin/ packages/data-objectstack/` | **324 files, 3612 passed, 1 skipped, 0 failed** |

**Declared narrowing.** The full `app-shell` suite was not run. This diff adds one
test file and one changeset and changes **zero bytes of production source**, so no
existing test can change verdict; the directory that contains the new file plus
both `data-objectstack` draft suites were run in full instead. Lint was narrowed on
the same basis, with its three readings stated: the checked population is
`eslint.config.js`'s own ignore set (build output only); the count comes from
`--format json`; and the config enables **no type-aware linting** (no
`parserOptions.project`, no `projectService`), so adding a file cannot move any
untouched file's verdict. CI runs the full farm.

## Acceptance notes

**Filed as objectui#9420** — a second, distinct defect this PR's harness exposed
and deliberately does **not** fix, because it is not this card's defect class:
on that same `layered().catch(() => null)` arm, the package door drops **other
packages' permission rows**. Measured with a one-shot probe (since deleted): with
`b_order` contributed by another package, the body actually PUT carried
`objects: ["a_account"]` and `fields: []` — `b_order` and `b_order.total` gone,
with a 200 and no error. The merge base falls back to the draft, which the load
path already sliced to the package's own objects, so `mergePermissionSlice` has
nothing out-of-scope left to preserve. That is the exact guarantee the code's own
comment claims two lines above it — *"so rows contributed by other packages survive
byte-for-byte (ADR-0086 P0)"* — and the same silent-discard class as #4302, which
is closed and does not cover this arm. Deduped before filing: 63 open-and-closed
`package: app-shell` cards, zero hits for `mergePermissionSlice` / `ADR-0086` /
byte-for-byte row preservation, with `draft-envelope` as a firing control.

**Noted, not filed.** `views/metadata-admin/clientValidation.ts` (around line 553)
still states that *"`ResourceEditPage` builds its edit draft as `{...layered.effective,
...client.getDraft().item}` and strips nothing"*, and turns the `sharing_rule`
edit-door gate off on that basis. Both halves are stale: `ResourceEditPage` strips,
and so does every other consumer. The recorded blocker is therefore cleared, which
is tracked by the already-open #7612 — no duplicate filed. Not touched here because
it is outside the three named consumers. Carrier: #7612.

**Noted, not filed.** `ObjectHooksPanel` and `StudioDesignSurface` have no
consumer-level decoration pin of their own; leg A shows they are covered at the
hoisted owner, which is the point of the hoist. A per-consumer pin would only add
the case "someone stops calling `extractDraftBody`". Left alone rather than adding
two more files to a busy package. Carrier: none.

---

---


os-tesla:

---

PR #9431, opened 2026-09-13
plugin-charts: categorical x-axis above 5 buckets still loses labels (minTickGap)

### LINKED ISSUE #7386: plugin-charts: a categorical x axis above 5 buckets still loses labels — minTickGap 48 is a time-series budget charged to band axes
## What

`AdvancedChartImpl`'s `xAxisCommonProps` charges every non-forced x axis a
`minTickGap` of 48px (32px under the mobile viewport check). That is a
time-series budget: 48px of mandatory whitespace *between* labels, on top of
each label's own measured width. On a **categorical** axis it is the reason a
bar loses its name, and recharts does not need the help — it already refuses to
draw ticks whose measured boxes would overlap.

objectui#7247 (PR objectstack-ai/objectui#7385) removed this for short
categorical axes by drawing every label at 5 buckets or fewer, a bound derived
so it is safe at the narrowest shipped widget without measuring anything at
runtime. Above that bound the 48px budget is untouched, so a 6-plus bucket
categorical axis still drops labels even where the widget has ample room — a
7-status pipeline at 290px, for instance, has roughly 33px per band and loses
labels anyway.

## Why it was not done in that PR

Two reasons, both deliberate:

1. **The replacement value is a judgement call**, not something existing
   evidence pins. Recharts' default is 5; the right number for a band axis is
   somewhere between "0, let measurement decide" and "enough that labels do not
   read as one word", and picking it is a design decision.
2. **It is not measurable in this repo's DOM test environment.** happy-dom
   reports zero text metrics, so `getStringSize` returns `{width: 0, height: 0}`
   and recharts' size-based thinning cannot be observed at all — any unit test
   asserting a relaxed `minTickGap` would pass for the wrong reason. Verifying
   it needs a real browser pass, which is why it belongs in its own change with
   its own before/after counts.

Note also that recharts' collision model for **angled** labels is the projected
bounding box (`getAngledTickWidth` to `getAngledRectangleWidth`), which is
strictly more conservative than the real constraint — rotated labels are
parallel lines, so they collide on perpendicular separation, not projected
width. Any work here should decide deliberately whether to live with that
conservatism or bypass it the way the 5-bucket branch does.

## Where

`packages/plugin-charts/src/AdvancedChartImpl.tsx`, the `xAxisCommonProps` memo
(the branch above `X_AXIS_ALL_LABELS_MAX_BUCKETS`).

## Note

Found while implementing objectui#7247; filed unassigned for triage rather than
carried into that PR, whose scope was the reported 3-to-5 bucket cases.

os-tesla:
Fixes #7386

## What changed

`AdvancedChartImpl`'s `xAxisCommonProps` charged every categorical (bar/line/
area/combo) x-axis above `X_AXIS_ALL_LABELS_MAX_BUCKETS` (5 buckets) a
`minTickGap` of 48px (32px on mobile). `minTickGap` is **added on top of**
recharts' own measured-overlap check (`getTicks.js`: `start = tickCoord +
size/2 + minTickGap`), so this was mandatory whitespace beyond what real
collision avoidance already required — a time-series budget charged to band
axes too. A 7-status pipeline at a 290px widget (≈33px/band) dropped 4 of its
7 names even though nothing was actually wide enough to collide.

`minTickGap` is now `0` for that branch. `interval: 'preserveStartEnd'` is
unchanged, so recharts' own measured-width overlap avoidance is the only
thing governing tick density above the bound — the same mechanism that
already governed every tick it kept.

Not changed, and out of scope for this issue:
- `X_AXIS_ALL_LABELS_MAX_BUCKETS` / the ≤5-bucket "draw everything" branch
  (objectui#7247).
- The scatter chart's own `minTickGap={isMobile ? 32 : 48}`, on its XAxis
  where `type` is set to the string NUMBER — a genuinely continuous numeric
  measure axis, not a categorical band axis (scatter is explicitly excluded
  from `CATEGORY_AXIS_CHART_TYPES`), so the time-series-style budget is
  correctly scoped there and untouched.
- recharts' angled-tick collision model (`getAngledRectangleWidth`), which
  stays on its conservative projected-bounding-box approximation rather than
  the true parallel-line perpendicular-separation constraint. Documented in
  code as a deliberate choice, not an oversight — reimplementing that
  geometry is a bigger, untested-in-this-repo change than this issue's scope
  warrants.

## Why `minTickGap: 0` is not an arbitrary density pick

The issue's own diagnosis is that recharts "does not need the help" — it
already refuses to draw ticks whose measured boxes would overlap. `0` is not
a re-tuned density value; it is the literal removal of the extra margin,
leaving the pre-existing (and already-correct) measured-overlap check to do
all the work, exactly as it already does for the ticks it keeps.

## Real-browser verification (required — this repo's DOM test env cannot exercise it)

happy-dom's `getStringSize` reports `{width: 0, height: 0}` for every string,
so recharts' measured-overlap check cannot fire in this repo's unit tests at
all — the jsdom tests added/changed below can only prove the artificial floor
is gone, never that real text still avoids colliding. Verified instead with a
real-Chromium harness (React + recharts 3.10.1 built via esbuild, driven with
Playwright against the repo's own bundled Chromium), reproducing
`ChartContainerImpl`'s actual `text-xs` sizing and `apps/console`'s actual
`body` font stack (font-family/letter-spacing) — without which recharts'
hidden text-measurement span mismeasures against the wrong font entirely; see
#9430, filed out of scope, found while building this harness.

Overlap was checked as true rotated-rectangle intersection (SAT, via each
tick's `getCTM()`-mapped corners) — not an axis-aligned
`getBoundingClientRect()` box, which over-reports collisions for diagonal
text (confirmed by screenshot: an axis-aligned check flagged "overlaps" that
were not visually present).

| case | width | buckets | before (minTickGap 48) | after (minTickGap 0) | overlaps after |
|---|---|---|---|---|---|
| bar, rotated | 290px | 7 | 3/7 labels | **7/7 labels** | 0 |
| bar, rotated | 290px | 8 | 3/8 labels | **8/8 labels** | 0 |
| bar, rotated | 290px | 10 | 3/10 labels | 5/10 labels | 0 |
| bar, rotated | 200px | 7 | 2/7 labels | **4/7 labels** | 0 |
| line, dates | 200px | 24 | 2/24 labels | 3/24 labels | 0 |
| line, dates | 800px | 180 | 7/180 labels | 11/180 labels | 0 |

The last row is the regression check for the scenario the file's own comment
worries about ("hundreds of points painted a dense black bar"): thinning is
still very much in effect (11 of 180), and still zero-overlap — because
`minTickGap: 0` removes only the extra margin, not recharts' measured-width
check itself.

## Tests

- `packages/plugin-charts/src/AdvancedChartImpl.categoricalAxisDensity.test.tsx`
  — added 7-bucket/290px (the issue's own reproduction) and 6-bucket/200px
  cases asserting every label now draws; the pre-existing "still thins a
  high-cardinality axis" pin is rewritten (with an explanatory comment) since
  happy-dom's zero text metrics mean `minTickGap: 0` shows every tick
  regardless of count in THIS environment — the real safety property for
  that scenario is the 180-point/800px row above, which jsdom cannot exercise
  either way.
- `pnpm --filter @object-ui/plugin-charts test` — full package suite.
- `pnpm --filter @object-ui/plugin-charts type-check`.
- `packages/plugin-charts/src/ChartRenderer.catalogRender-6939.test.tsx`
  (maintainer ruling 5510084784, byte-identical render pin) — confirmed
  unaffected: both catalog fixtures have exactly 6 data points at the test's
  mocked 480px width, where 480/6 ≈ 80px per band already exceeded the OLD
  48px `minTickGap`, so all 6 ticks already drew under both old and new code.

## Acceptance notes

No findings withheld.

- The one adjacent thing that looked like a reproducible defect, found while
  building the real-browser harness: recharts' tick-size measurement span
  inherits `document.body`'s font-family rather than the chart's own, so a
  host page whose body font differs from the chart's silently mis-measures
  tick overlap. Separate, unassigned, out-of-scope: filed as #9430.
- An observation, not filed: the horizontal-bar family's category axis (the
  YAxis in the `isHorizontal` branch) formats its ticks with `xTickFormatter`
  directly rather than `xAxisTickFormatter`, so it never gets the ≤5-bucket
  rotated-label ellipsis objectui#7247 gave the vertical-bar path. Possibly
  deliberate — that axis already grows its own width from the longest label
  instead of rotating — so this is recorded as an asymmetry, not asserted as
  a bug.


os-tesla:

---

PR #9433, opened 2026-09-13
fix(plugin-*,components): one record-overlay shell honouring all four `navigation.mode` values

### LINKED ISSUE #9299: [decision] overlay `navigation.mode`: three renderers collapse four declared values into the drawer — and the measurement says portability is NOT the blocker, so this is a product question
**Decision needed.** Filed by the `domain:ui` PM seat (`os-tesla`) at R31, as objectui#9113's triage comment requires: *"If the answer requires changing what the spec's enum means for these view types, that is a contract change and it goes to the decision box — a `needs-user-decision` card with the measurement attached. ⛔ A seat may not take it."*

The measurement round is **complete** (PR objectui#9296, report on objectui#9113). The delivering seat **declined to recommend**, correctly — the fence forbade a seat from choosing. This card carries the measurement, prices the routes, and adds a PM recommendation the maintainer can accept or discard.

## The question in one line

The spec publishes four overlay `navigation.mode` values. **`ObjectGantt`, `ObjectKanban` and `ObjectCalendar` collapse all four into the drawer**; `ObjectGrid` and `ObjectTree` honour them through `NavigationOverlay`. So the same authored metadata means different things depending on which view type reads it, with no diagnostic. **What should the four values mean on these surfaces?**

## ⭐ The measurement changes the shape of the question

**Question 1 — is the `Sheet` the container or the component? — is answered by execution: it is the CONTAINER.**

`RecordDetailDrawer`'s payload (`InlineEditProvider` → `DetailView`, plus `InlineEditSaveBar`) **renders in the modal, split and popover shells**, behind a lit drawer control. ⇒ **portability is not the constraint.** The question is a **product** one — what `split` and `popover` should *mean* on a gantt or a kanban — not a technical-feasibility one.

⚠️ **And one sub-premise of the original fence is corrected by measurement.** The fence listed four capabilities as blockers: record fetching, inline field save, lock handling, and a full-page href. **`RecordDetailDrawer` does not fetch the record.** I verified this myself rather than relaying it — `RecordDetailDrawerProps` declares `record: Record<string, any>` as a **required prop** and `objectSchema?` as an optional one (`packages/plugin-detail/src/RecordDetailDrawer.tsx:54-74`), with no fetch in the component. ⇒ of the four named blockers, **three travel inside the payload and the fourth is the caller's**. The "straight swap is not obviously available" premise is weaker than it looked.

## ⭐⭐ Two further narrowings, both verified independently at this seat

**1. `popover` is honoured NOWHERE — including on the two renderers this card calls correct.**

```
packages/components/src/custom/navigation-overlay.tsx   popoverTrigger declared, destructured, guarded, rendered  (5 hits)
packages/plugin-view/src/ObjectView.tsx:2228            a COMMENT: "Popover mode — uses NavigationOverlay Dialog fallback (no popoverTrigger)"
                                                        ⇒ ZERO renderers pass it
```

`NavigationOverlay` falls back to a compact `Dialog` when `popoverTrigger` is absent, and no renderer ever supplies one. ⇒ `popover` degrades to a Dialog on **five** surfaces, not three.

**2. On `ObjectTree`, an authored `split` renders NOTHING.** The split branch is `if (!isOpen || !mainContent) return null`, and `ObjectTree` never passes `mainContent` (`ObjectGrid` does, via a dedicated early return).

⇒ **the honest net picture: no renderer honours all four modes, and the two axes are anti-correlated.** The pair that honours `mode` renders the *poorer* payload; the three that collapse `mode` render the *rich* one. **Nobody gets both today.**

## The three routes, priced

| | route | what it costs |
|:--|:--|:--|
| **(a)** | **Honour the modes in all five renderers** | A payload/shell split at three call sites **plus an edit to `plugin-detail`** (`RecordDetailDrawer` brings its own `Sheet`). Chrome de-duplication that **moves the default drawer presentation on three published surfaces**. Reconciling two independent drag-resize implementations with different storage keys (`ov:KEY` vs `objectui.drawerWidth.OBJECT`) and different floors (**360 px vs 480 px**), so **existing users' persisted widths shift**. A structural change per renderer for `split`, which needs `mainContent`. A `popoverTrigger` that must be the clicked bar / card / event. And it must repair both narrowings above. ⚠️ **Blocked on the very product question this card asks.** |
| **(b)** | **Narrow the declared enum for these view types** (ADR-0049 enforce-or-remove) | An `@objectstack/spec` change ⇒ a contract change by definition. Authoring-time rejection, migration for metadata already carrying these values, docs. Needs a shape that **does not exist yet** — a per-view-type mode vocabulary is a new kind of thing for this enum. ⚠️ **The measurement argues against it**: the payload is portable, so this narrows the contract to fit an implementation that **could** have met it. |
| **(c)** | **Keep the drawer, make the narrowing loud** | Much the cheapest — one shared helper, one call in each of three renderers, no spec change, no payload extraction, no chrome or resize reconciliation, no change to the default path. ⚠️ **But what it costs is a RULING, not work**: *refuse or warn?* A dev-only console warning changes nothing for a production author; a visible refusal is a **user-visible behaviour change on a published surface**, because today's drawer would stop appearing for anyone who authored `modal`. "Fall back to drawer but warn" is exactly today's code plus a log line. ⛔ And it leaves both narrowings unfixed — they are (a)-shaped. |

## ⭐ PM recommendation — **split the answer by mode**, because the modes are not alike

⚠️ Offered as a recommendation only; the decision is the maintainer's, and the three routes above stand unaltered if this is discarded.

The three values are lumped together by the card, but the measurement separates them:

- **`modal` has no product question at all.** It means "the same payload, centred, modal" — there is nothing to decide, and the measurement shows the payload already renders there. ⇒ **honour it** (route (a), restricted to one mode). Still not free: it needs the payload/shell split and touches `plugin-detail`. But it avoids **every** cost that is `split`- or `popover`- or resize-specific.
- **`split` and `popover` have no agreed meaning on a gantt, a kanban or a calendar** — what is the "main content" beside a split record panel on a kanban board? what element is the popover anchored to? ⇒ **route (c) for these two**, loud, until someone decides what they mean.
- ⭐ And **(c) applied to `popover` is close to free-standing value on its own**, because popover is honoured on **no** surface — so making it loud is honest everywhere rather than a special case for three renderers.

⇒ this gets the unambiguous mode working, stops the silent lie on the two ambiguous ones, and ⛔ does not spend a spec change on a limitation the implementation could have met.

**On refuse-vs-warn, if (c) is taken:** recommend **warn in development, fall back in production** — a visible refusal removes a working drawer from authors who did nothing wrong, and the drawer is a *reasonable* answer to `modal`, just not the declared one.

## Where the evidence is

- **PR objectui#9296** — the measurement, 5 test files, +992/−0, **no product source touched**. `Test Files 4 passed / Tests 30 passed`.
- **objectui#9113** — the card, its binding triage comment, and the seat's full report.
- ⭐ The instrument caught itself once and it is worth noting: the question-1 probe's **lit control failed on the first run** (a network escape to `localhost:3000`, plus an over-loose assertion), which was read as **NOT MEASURED** rather than as a negative answer, repaired, and re-run green.

## Blocked

objectui#9113 cannot be retired without this ruling. Both narrowings above are inputs to it rather than separable repairs, so ⛔ neither was filed as its own card.

⚠️ **Dedup gap, named:** REST `/search/issues` is proxy-refused in this session, so dedup ran over **open** issues only (450 read, objectui#9113 itself as the lit control). A closed duplicate would not have been seen.

Refs: objectui#9113 · PR objectui#9296 · objectui#7334 · objectui#7210 (route B, unruled, ⛔ out of scope) · objectui#7742 · ADR-0049

---

os-tesla:
Fixes #9299

Implements the director-seat ruling of decision batch #128 item 1 (2026-09-13): one shared overlay shell, so the four declared overlay `navigation.mode` values mean the same thing on every list-type renderer.

`RecordDetailDrawer`'s payload — `InlineEditProvider` wrapping `DetailView` plus `InlineEditSaveBar`, and the objectSchema-to-typed-field derivation that feeds them — is extracted into a new shell-free `RecordDetailPanel` and mounted through `NavigationOverlay` by `ObjectGrid`, `ObjectTree`, `ObjectGantt`, `ObjectKanban` and `ObjectCalendar`. `RecordDetailDrawer` keeps its published props and is now a thin wrapper over the same shell in `drawer` mode, which is what makes item 4 literally true: there is one drag-resize implementation left, not two. No spec change (`Clause-2: no`, declared by the ruling itself).

## What each ruled item became

| Item | Where |
|---|---|
| 1 — one shared overlay shell | `packages/plugin-detail/src/RecordDetailPanel.tsx` (new), `RecordDetailDrawer.tsx` (delegates), five renderers |
| 2 — `split` gets each renderer's own view as `mainContent` | every renderer's `renderRecordOverlay(mainContent)` |
| 3 — `popover` anchored to the clicked element | `popoverAnchorRef` + `useOverlayAnchor` in `navigation-overlay.tsx` |
| 4 — one resize store, migrated not reset | `useDrawerResize`'s `legacyStorageKey` read |
| 5 — no spec change, no dev-only warning | nothing in `packages/types` or `@objectstack/spec` moved |

## The `popover` mechanism, and why it is not `popoverTrigger`

The ruling names `popoverTrigger` as the vehicle for "anchored to the clicked element". Measured: that prop cannot carry it. It is rendered through Radix's `PopoverTrigger asChild`, which MOUNTS its child inside the overlay's own subtree — so handing it the clicked row would draw a **second copy of that row** inside the popover rather than point at the first. The element the popover must anchor to is one the renderer already drew somewhere else entirely.

So the ruled BEHAVIOUR is delivered through the shape Radix publishes for exactly this case — a virtual anchor (`Popper.Anchor virtualRef`) behind a new `popoverAnchorRef` prop. `popoverTrigger` is untouched and still works for a host that wants the overlay to render its own trigger. The `Dialog` fallback now requires BOTH to be absent, and none of the five reaches it.

A vehicle that provably cannot carry the ruled behaviour is a measurement, not a licence to change the behaviour — and the behaviour is unchanged.

## Two things the card did not know, both measured here

**1. `ObjectGrid`'s `split` was blank too.** The card calls `ObjectGrid` one of the two renderers that get overlay modes right, and says the `split` blank is an `ObjectTree` problem because the grid passes `mainContent` "via a dedicated early return". The measurement checked that the prop is PASSED, never that the closed state renders. It does not: the grid entered its split path on the authored mode alone, while the shell's split branch opens `if (!isOpen || !mainContent) return null` — so a grid authored `split` rendered nothing at all, and nothing could ever be clicked to open it. Same defect class as the `ObjectTree` blank, on the renderer the card calls correct. Repaired here (item 2 names no exception), and pinned by the mount step of every `split` case in the five new shell-mode files.

**2. The shared payload needs DECLARED fields, so grid and tree keep their own reading when there are none.** `RecordDetailPanel` renders the object's declared fields; `ObjectGrid` and `ObjectTree` render the same values by INFERRING from them — a date-shaped string on a `*_date` key, a currency-ish number on an `amount` key — which is what objectui#4541 (locale-aware date fallback), objectui#8491 (the localized `Empty` placeholder) and objectui#8920 (the `format` hint) put there. With the payload ungated, `recordDetailDateLocale` read back `close_date2024-03-15` in a zh session: three landed behaviours regressing for any grid over inline `value` rows or a DataSource with no `getObjectSchema`.

So those two renderers take the shared payload wherever the object declares its fields — every console surface — and keep their inference reading where there is nothing to declare.

⭐ **This is a payload-CONTENT deviation only, and it is pinned as such rather than asserted.** On the no-schema path the overlay is still `NavigationOverlay`, it still receives all four authored modes as four distinct boundary pairs, `split` still gets `mainContent` and `popover` still gets the clicked element — driven, not read off the call graph, in the closing `describe` of `ObjectGrid.overlayShellModes-9299.test.tsx` (with a lit control proving that mount really does take the fallback branch) and again in `ObjectTree.overlayShellModes-9299.test.tsx`, where a schema-less tree is the common case rather than an edge.

## Consequences on published surfaces, stated rather than discovered

- **Drawer chrome converges** for gantt / kanban / calendar on the shared shell's header (breadcrumb-style title, close, optional expand) in place of the sr-only `SheetHeader`. This is the "chrome de-duplication" route (a) priced in. The shell also adopts `RecordDetailDrawer`'s deliberate `onOpenAutoFocus` suppression, so the grid and tree drawers stop flashing a focus ring on open.
- **Default width is unchanged at every viewport.** The shell expresses an authored width as a floor, `max(WIDTH, min(60vw, 880px))`, and for the ruled default `min(960px, 60vw)` that reduces to `min(960px, 60vw)` at every viewport — the three `navWidthDefault` pins assert the reduced expression and carry the arithmetic. objectui#6584 holds.
- **An authored width below `min(60vw, 880px)` now widens to it** on gantt / kanban / calendar. That is the shell's long-standing policy on grid and tree; converging on one shell means converging on one width policy. Pinned explicitly (`an authored navigation.width still reaches the shell unchanged`).
- **Mobile**: below the `sm` breakpoint the overlay occupies the full viewport instead of the authored width, which is the shell's documented mobile behaviour.
- **Grid and tree overlays now issue the record's `POST /api/v1/security/explain` read** (two per open) because `DetailView` asks whether the record may be updated and deleted. That call is `DetailView`'s, ungated by `inlineEdit`, and pre-existing on the three renderers that already rendered this payload; it is noted, not changed here.
- **Richer record body on grid and tree** wherever fields are declared: typed widgets, declared labels, honoured `hidden`, the object's own field order.

## Item 4 — the migration, proved from an existing width

`packages/components/src/__tests__/navigation-overlay-9299.test.tsx` seeds `objectui.drawerWidth.contacts = 640` with the new key ABSENT, then reads back that 640px is what the overlay renders, that `ov:drawer-width:contacts` now holds it, and that the retired key is gone — a migration proof rather than a new-key proof. Plus: a stale-legacy-loses case, a no-persisted-width control, a below-floor case, and a pin on both key spellings. The surviving floor is the LOWER of the two (360 vs 480), so no width either implementation could have written is discarded.

## Reverse verification (one-time, not left in the tree)

Both new shell mechanisms were ablated at HEAD with an `EXIT INT TERM` trap, the mutation proved on disk by a grep count in both directions, and the restore proved by blob hash plus an empty `git diff HEAD`:

| Ablation | Mutation landed | Result |
|---|---|---|
| delete the legacy-key migration read | `legacyRaw` 3 hits to 0, marker injected | 1 failed / 7 passed — `starts from an EXISTING persisted width` red: `expected 'max(min(960px, 60vw), min(60vw, 880px))' to be '640px'`, i.e. exactly the silent reset the ruling forbids |
| widen the popover fallback guard to ignore `popoverAnchorRef` | guard 1 hit to 0, marker injected | 2 failed / 6 passed — both anchor cases red |

## Test files: replaced, not kept

The four objectui#9113 measurement files pin the branch this card deleted (`RecordDetailDrawer:undefined` four times over, and on `ObjectTree` a `split` that renders nothing and a `popover` with no anchor). They are replaced by `*.overlayShellModes-9299.test.tsx` on all five renderers, reusing the same two instruments so the two rounds are comparable — the hook spy and the shell recorder used to DISAGREE, and the disagreement was the defect; they agree now. Assertions stay on the mode VALUE because `sheet.tsx` and `dialog.tsx` both render the same Radix role with no `data-slot`, so a DOM assertion is blind to this defect by construction.

Three `navWidthDefault` files move their reading from the panel's inline `max-width` to the `--ov-w` custom property the shell publishes; reading `style.maxWidth` there would return an empty string in both worlds and pin nothing.

A shared `installRecordSecurityExplainDouble` lands in the private `@object-ui/test-support` so the seven files whose renders now reach `DetailView` serve that read from a recording double instead of a real socket — it records every URL and fails on anything that is not the explain route, so an escape elsewhere reds where it happens.

## Verification

All runs at `69fe1e7` unless noted; test suites through the shared verify lock.

| Run | Result |
|---|---|
| `pnpm lint` (whole repo, 47 packages) | exit 0 — 47/47 tasks, 0 errors |
| `turbo run type-check` (7 changed packages + test-support) | exit 0 — 22/22 tasks |
| `vitest run packages/plugin-grid/ packages/plugin-tree/` | 146 files / 1288 tests passed |
| `vitest run packages/plugin-gantt/` | 69 files / 538 tests passed |
| `vitest run packages/plugin-kanban/ packages/plugin-calendar/` | 91 files / 609 tests passed |
| `vitest run packages/plugin-detail/` | 175 files / 1625 tests passed |
| `vitest run packages/components/ packages/test-support/` | 277 files / 2692 tests passed |
| gates | `check:control-bytes`, `check:vi-mock-specifiers`, `check:vi-mock-inherit`, `check:vi-mock-override-shape`, `check:test-path-roots`, `check:phantom-deps`, `check:unused-deps`, `check:self-import`, `check:esm-specifiers`, `check:dist-completeness`, `check:side-effects-array`, `check:published-dist`, `check:entry-guard`, `check:lint-rule-coverage`, `check:i18n-keys`, `check:readme-exports`, `check:doc-snippets`, `check:doc-examples`, `check:doc-fences`, `check:doc-types`, `check:unreferenced-sources`, `check:changeset-claims`, changeset presence + no-major — all exit 0 |
| NOT MEASURED locally | `check:eager-closure` and `check:sdui-registration-pins` both exit 2 (PRECONDITION NOT MET — they weigh `apps/console/dist`, which needs a console build). Neither is a failure reading; CI's `Build & E2E` builds it. |

The grid/tree, detail and components suite numbers were taken at `4d9bb7e`; the only commits since touch a gantt test file (deleted), one components test file (re-run green at HEAD, 8/8) and two doc files, none of which those suites read.

`Bundle Analysis` is red on `origin/main`'s own tip and on every PR from this lane; it is outside the required nine and is not this PR's.

---

Generated by Claude Code, session `session_011QreXiyMEqKLN4U5daMPVa`.

---


os-tesla:

---

PR #9437, opened 2026-09-13
fix(plugin-detail): an interpolated `titleFormat` H1 hides no field row (objectui#8351, ruled option B)

### LINKED ISSUE #8351: ruling needed(components/core): PageHeaderRenderer ranks `titleFormat` ABOVE the declared name pointer — the opposite of `getRecordDisplayName`'s documented ADR-0079 precedence
Filed by the `domain:ui` PM seat (`session_01YBWFb5YgMU5dw8p2VKj16S`) on behalf of the objectui#8175 dev, who found it while making `record:details`' dedupe ladder agree with the header chain and correctly refused to fold it in — **closing it changes what the H1 shows**, which is a product ruling, not a repair. ⛔ Not claimed, ⛔ not graded.

## What

Two implementations of the same ADR-0079 concept order two rungs differently:

- **`@object-ui/core`'s `getRecordDisplayName`** documents `titleFormat` at **step 3**, i.e. **BELOW** the declared pointer (`nameField` / `displayNameField`).
- **`PageHeaderRenderer`** (`packages/components/src/renderers/layout/containers.tsx`) ranks the interpolated `titleFormat` template **ABOVE** the declared pointer. This is not incidental — it is pinned green today, in `packages/components/src/__tests__/page-header-title.test.tsx` as `titleFormat still outranks nameField`.

So on an object that declares both, the two answer differently, and the currently-green pin is the one asserting the divergent order.

## Why it surfaced now

objectui#8175 (PR #8349) taught `record:details`' dedupe ladder to lead with the resolver instead of a six-entry literal walk, so that the grid hides the row the H1 is showing. That works everywhere the two chains agree. On a `titleFormat` object the ladder now hides **the resolver's field** where it used to hide **the literal walk's** — and *neither* is what the H1 displays, because the H1 is showing an interpolated template that names no single field at all.

⚠️ That is not a regression introduced by objectui#8175: the ladder was wrong on these objects before and is wrong differently now. It is the same defect made visible.

## The related second case, recorded so it is not lost

`page:header`'s own `schema.title`: when an author sets it, the H1 is an arbitrary interpolated string and the dedupe still hides a row no heading is showing. `plugin-detail` cannot see the header schema at all, so nothing in that package can answer it.

**Common shape:** both are rungs that **name no field**. A dedupe keyed on "which single field is the H1" is structurally unable to answer them.

## Options — this is the ruling being asked for

- **A — leave both.** Current behaviour is no worse than before objectui#8175, and a dedupe keyed on one field genuinely cannot answer a rung that names none. Cost: on `titleFormat` objects a grid row still vanishes for no visible reason.
- **B — suppress the dedupe when `objectSchema.titleFormat` is present and fully interpolates.** Hide nothing, because the H1 is not any one field's value. Cheap and local to `plugin-detail`; does **not** touch `schema.title`, which stays invisible to that package.
- **C — close the ordering divergence at its source in `@object-ui/components`,** so the header stops ranking the legacy template above the ADR-0079 pointer. This is the contract-first answer and the only one that makes the two implementations one. ⚠️ It **changes what the H1 shows** on existing records and **contradicts a currently-green pin**, so it is squarely a maintainer ruling.

**Reporting dev's recommendation, and the PM's:** **A** now (nothing is blocked, and PR #8349 is a strict improvement without it), with **B** as its own card if record pages carrying a `titleFormat` turn out to be common. **C** is the real ADR-0079 question and needs the maintainer, because it moves user-visible output and retires a green pin.

## Related

objectui#8175 / PR #8349 (where it was measured) · objectui#7287 / PR #7585 · objectui#7586 · objectui#6843 · objectui#7021 · ADR-0079

## Dedup

`search_issues` over open issues for `titleFormat` / `nameField` / `getRecordDisplayName` / `ADR-0079` — 7 hits, all read. Nearest: objectui#6843 (the record-key probe's stale "step 3b" name) and objectui#7021 (the record-title key in three spellings) — both about the same machinery, neither about the precedence between these two rungs. objectui#8175 is the card this was split out of.

os-tesla:
Fixes #8351

Implements the ruled **option B** (director seat, decision batch #82, 2026-09-08, maintainer 「同意」): in `plugin-detail`, an interpolated `titleFormat` H1 names no single field, so the dedupe ladder hides no row.

---

## ZONE 2 measurements — both assumptions checked before any edit

**1. Did `plugin-detail` move under objectui#9299? For this card's surface, no.** objectui#9299 landed as `7098eed` (this branch's base). Its `plugin-detail` diff extracted `RecordDetailDrawer`'s payload into `RecordDetailPanel` and mounted it through `NavigationOverlay` — the record **overlay shell**. The dedupe ladder was not in that diff: `git log -- packages/plugin-detail/src/renderers/record-details.tsx` stops at `345e24a`, two cards earlier. And the extraction did not relocate the H1 — `RecordDetailPanel.tsx` has zero occurrences of `h1`, `title`, `DisplayName`, `page:header` or `record:details`. The ladder is where the card says it is, in `packages/plugin-detail/src/renderers/record-details.tsx`, and the H1 is still drawn a package away by `@object-ui/components`' `PageHeaderRenderer`.

**2. What "fully interpolates" is measured against — established from the code, and it is NOT a bare presence test.**

The chain already owns both instruments, so no new predicate was written:

- `formatTitleTemplate` (`packages/core/src/utils/record-title.ts`) renders the template. It is what `getRecordDisplayName` step 3 calls **and** what this package's own `DetailView.resolveDisplayTitle` step 2 calls, so all three agree about what a `titleFormat` produces on a given record.
- `recordDisplayValueAt` — the emptiness authority objectui#8350 made this ladder import rather than re-spell — then answers the only question a dedupe has: **is that rendered string some candidate's value?**

Three outcomes, and the two that are not the headline case are what keep the rule honest:

| rendered title | H1 is… | ladder |
|---|---|---|
| a composite no candidate's value equals | not any one field's value | **hides nothing** — the ruled case |
| empty (no placeholder resolved) | the declared pointer; the header already walked past this rung | existing value-keyed walk, **unchanged** |
| collapsed onto one field's value | exactly that field's value | **still hides that row** |

⚠️ **A presence-only reading was measured and rejected.** Suppressing whenever a `titleFormat` merely resolves prints `Contract No: HT-2026-003` directly beneath an H1 reading `HT-2026-003`, on every record where a blank placeholder was dropped with its orphan separator — which is precisely the duplication Phase P.0 of this ladder exists to remove. It would also make the lit control this PR was asked for unconstructible: the suggested route's *"a `titleFormat` that does not fully interpolate, where the ladder must still hide"* only exists under a reading where partial interpolation still hides.

⚠️ **The match is a SCAN of the candidates, not a peek at the first one with a value.** With `titleFormat: '{name}'` over `nameField: 'contract_no'`, the first resolving candidate is `contract_no` while the H1 is `name`'s value — stopping early hides the wrong row **and** leaves the real duplicate, the objectui#8175 shape one rung along. Pinned.

---

## Red / green legs

The pin is lit, measured twice, in both directions:

- **Before the implementation existed** (new test file only, on `7098eed`): `Test Files 1 failed`, `Tests 2 failed | 5 passed (7)`. The two reds are `HALF 1 — KEEPS the declared nameField row` and `hides the row the TEMPLATE names…`.
- **After** (`425cd85`): `Test Files 1 passed (1)`, `Tests 7 passed (7)`.
- **Ablation from the committed state**, with on-disk proof and a `trap`-guarded restore: the rung's read site was neutralised (`const interpolatedTitle = ''`), anchor occurrences `1 → 0`, injected marker count `1`, on-disk hash differs from the `HEAD` blob — then `Tests 2 failed | 5 passed (7)`, the same two cases. Restore verified two ways, not by an exit code: `git diff HEAD` empty **and** `git hash-object` equal to the `HEAD` blob. No `dist/` is on the path — the pin imports `../record-details` relatively, from source in the same package — so no rebuild leg was owed.

The remaining five cases include three deliberate controls that pass on **both** sides (an object with no `titleFormat` dedupes exactly as before; a template resolving to nothing leaves the ladder alone; a single-field `{name}` template still dedupes) — without them, a change that simply disabled the H1 dedupe outright would pass everything.

---

## Gates run locally

All at `425cd85`; the working tree is byte-identical to it (`git diff HEAD` empty).

- `pnpm exec vitest run packages/plugin-detail/` — **176 files, 1632 tests, all passed** (under the shared `os-verify-lock.sh`).
- `pnpm --filter @object-ui/plugin-detail run type-check` — **exit 0** (`tsc --noEmit && tsc -p tsconfig.test.json`, so the new pin is type-checked too). ⚠️ This needed `pnpm --workspace-concurrency=2 --filter '@object-ui/plugin-detail^...' build` first: without the dependency `dist/*.d.ts` it reports `TS2307 Cannot find module` for every workspace import — a missing prerequisite, not a red gate.
- `pnpm exec eslint packages/plugin-detail --format json` — **233 files, 0 errors**, 1013 warnings (all pre-existing `no-explicit-any` / `react-refresh`, the same shape the neighbouring pins carry). **Declared narrowing** against the repo-wide `Lint` job: the population is eslint's own file selection for that directory, not a glob of mine; the count is read from `--format json`; and `eslint.config.js` enables no type-aware linting (no `parserOptions.project`, no `projectService`), so no rule crosses a file boundary and this diff cannot move the verdict on any file it does not itself contain. The repo-wide run is CI's.
- `node scripts/check-changeset-presence.mjs` — exit 0, `1 source file(s) of 1 released package(s) changed, and this change declares 1 changeset(s)`.
- `node scripts/check-changeset-no-major.mjs` — exit 0.
- `check:control-bytes` · `check:test-path-roots` · `check:changeset-claims` · `check:unreferenced-sources` · `check:new-line-citations` · `check:comment-mask-corpus` · `check:vi-mock-specifiers` — all exit 0. Plus a direct control-byte scan of the three changed files (`grep -naP` over the C0 range): no hits.
- `node scripts/check-governed-queue-guard.mjs --test` — `NOT GOVERNED` for all three paths.

**NOT MEASURED:** `check:sdui-registration-pins` — `❌ No console build to weigh at apps/console/dist/assets`. A missing prerequisite (it weighs an `apps/console` build this worktree has never produced), not a verdict on this diff. Left to CI.

**Not run locally, by design:** the repo-wide `Lint` / `Type Check` / the four `Test` shards / `Build & E2E` / `Build Docs`. Those are the farm's; CI convergence is the PM's read, not a local one.

---

## Acceptance notes

Out of scope, observed while measuring, **noted and not filed**:

- `@object-ui/components`' `PageHeaderRenderer` and `@object-ui/core`'s `formatTitleTemplate` are **two different interpolators** for the same `titleFormat`. The header's local `interpolate` matches `{[a-zA-Z0-9_.]+}` and then refuses the result if a `{` survived; `formatTitleTemplate` matches `{[^{}]+}`, normalises handlebars `{{x}}` first, and strips orphan separators. So a token such as a name containing a space is a value to one and a literal to the other, and the two halves can disagree about whether the template rung resolved at all. **Successor: objectui#9436**, whose whole subject is that code region and which must weigh both interpolators whichever way it is ruled — this is not a card of its own, it is context that card needs.
- `DetailView.resolveDisplayTitle` is a **third** ranking of the same rungs, inside this package (`primaryField` → `titleFormat` → unified resolver → `schema.title`). It reads `schema.primaryField` off the **view** schema, which is legitimate and explicitly protected by objectui#7586, so this is an observation about how many orderings exist, not a defect. **Successor: objectui#9436** — the same ruling decides whether three orderings collapse to one.

Neither is filed separately: both are the same ordering question that card asks, and a card whose answer is "see objectui#9436" is noise in the queue.

---

## Option C — filed, not folded in

The ruling placed an obligation on whoever took this card. **objectui#9436** is now open, carrying `needs-user-decision` / `domain:ui` / `priority:p2`. It states plainly that the change moves what the H1 shows on existing records, and it names the currently-green pin it would retire: `packages/components/src/__tests__/page-header-title.test.tsx`, case **`titleFormat still outranks nameField`**.

⛔ Nothing in this PR touches `@object-ui/components`, that pin, or the ordering. This ladder deliberately does not depend on which rung wins the header: it compares the **rendered title** against the candidates' values, which answers the dedupe under either order — so objectui#9436 can be ruled either way without reopening this.

⛔ Also untouched, excluded by the ruling by name: `page:header`'s own `schema.title`. It is a key on the header schema, which `record:details` never receives, so nothing in this package can answer it.

Dedup before filing: `/search/issues` is 403 in this container, so REST list endpoints plus a local grep over all 429 open issues, all 25 `needs-user-decision` issues and PRs (`state=all`), this card's cross-reference timeline, and 1902 closed issues — 38 cards matched the six keywords, all read, none asking the precedence question. Two control words fired on objectui#8351 in the same pass (`titleFormat still outranks nameField` → 1, `PageHeaderRenderer ranks` → 1), so the zero is a reading rather than a dead channel.

---

## Files changed

- `packages/plugin-detail/src/renderers/record-details.tsx` — the rung, plus the ladder docblock, which had recorded both these divergences as *"deliberately NOT repaired here"*; one of the two now is, and the comment says which and why the other cannot be.
- `packages/plugin-detail/src/renderers/__tests__/record-details.titleFormatNoDedupe-8351.test.tsx` — new pin, 7 cases.
- `.changeset/8351-titleformat-hides-no-row.md` — `patch` on `@object-ui/plugin-detail`; user-visible, so a real bump rather than an empty declaration.

---


os-tesla:

---

PR #9442, opened 2026-09-13
fix(components): split the letter-to-digit boundary in the icon tokeniser

### LINKED ISSUE #9414: finding(components): the icon seam's tokeniser never splits the letter-to-digit boundary, so 147 canonical PascalCase lucide names resolve nowhere in the platform
Filed by the `os-dev` seat working objectui#7472 (session `session_01FhBNJcLRZLe8M87VcUgpKr`), as an out-of-scope finding measured while wiring `packages/cli`'s generated layout to the seam. Unlabelled: routing and priority belong to triage, and this seat does not produce a `domain:` label.

Not fixed in objectui#7472's PR, deliberately — that card's fence is explicit that the seam's contract is not to be re-opened, and widening the tokeniser changes behaviour for every censused resolver at once, not just the one call site that card moves.

## The finding

`packages/components/src/lib/lazy-icon.tsx`'s `toKebabIconName` is the platform's single tokeniser since objectui#5935. It splits two boundaries:

```
.replace(/([a-z0-9])([A-Z])/g, '$1-$2')      // lower-or-digit  ->  Upper
.replace(/([A-Z]+)([A-Z][a-z])/g, '$1-$2')   // acronym run     ->  Word
```

Neither splits **letter to digit**. So `Building2` tokenises to `building2`, and lucide's canonical key is `building-2`. The name is rejected and `getLazyIcon` degrades to the `Database` fallback.

`Building2` is not an obscure spelling: it is the PascalCase component `lucide-react` exports, the spelling an author copies out of lucide's own site, and the spelling that worked in a generated CLI app until objectui#7472.

## Measurement

Measured on `origin/main` at `9ba7e9c3a2`, against `lucide-react@1.35.0` as resolved from `packages/components` in this workspace. Populations named in words:

| population | count |
|:--|--:|
| `lucide-react` namespace export keys (every spelling the library exports) | 6,110 |
| of those, component-valued | 6,109 |
| canonical kebab icon names (`iconNames` from `lucide-react/dynamic.mjs`) — what the seam accepts | 2,039 |
| component keys that DO survive `toKebabIconName` | 1,882 |
| component keys that do NOT | 4,227 |

Splitting the 4,227 by shape:

| shape | count | verdict |
|:--|--:|:--|
| `*Icon` suffix aliases (`HouseIcon`) | 2,037 | correctly rejected — alias, not a canonical key |
| `Lucide*` prefix aliases (`LucideHouse`) | 2,036 | correctly rejected — same |
| neither alias shape | 154 | **the finding** |

Of those 154, **147 are recovered by adding a letter-to-digit split and nothing else**: `ArrowDown01`, `Axis3D`, `BarChart2`, `BarChart3`, `BarChart4`, `BookUp2`, `Building2`, `Calendar1`, `CalendarCheck2`, `CheckCircle2`, `CircleSlash2`, `Clock1`, and 135 more.

The remaining 7 are `Grid2x2`, `Grid2x2Check`, `Grid2x2Plus`, `Grid2x2X`, `Grid3x2`, `Grid3x3` — where the digit-letter-digit run needs more than one rule — plus `useLucideContext`, which is a hook and correctly rejected.

Reproduce (run from `packages/components` so the specifiers resolve):

```
node --input-type=module -e "
import * as L from 'lucide-react';
import { iconNames } from 'lucide-react/dynamic.mjs';
const k = (n) => n.includes('-') ? n.toLowerCase()
  : n.replace(/([a-z0-9])([A-Z])/g,'\$1-\$2').replace(/([A-Z]+)([A-Z][a-z])/g,'\$1-\$2').toLowerCase();
const kd = (n) => n.includes('-') ? n.toLowerCase()
  : n.replace(/([a-z0-9])([A-Z])/g,'\$1-\$2').replace(/([A-Z]+)([A-Z][a-z])/g,'\$1-\$2').replace(/([A-Za-z])([0-9])/g,'\$1-\$2').toLowerCase();
const v = new Set(iconNames);
const keys = Object.keys(L).filter((x) => x !== 'icons');
const lost = keys.filter((x) => !v.has(k(x)));
const other = lost.filter((x) => !/Icon\$/.test(x) && !/^Lucide/.test(x));
console.log('namespace keys', keys.length, '| canonical names', v.size);
console.log('non-alias keys the tokeniser drops', other.length);
console.log('recovered by a digit-boundary split', other.filter((x) => v.has(kd(x))).length);
console.log('CONTROL, must be true:', v.has(k('House')), v.has(kd('Home')));
"
```

## Why no gate catches it

`scripts/check-lucide-icon-record-names.mjs` judges **authored/declared names that reach a resolver** against the live `icons` keys. Its reading on `9ba7e9c3a2` is

```
OK  lucide icon names: 184 authored/declared names reaching 1 record-reading resolver are live `icons` keys
```

Green, and correctly so: nothing in the tree currently authors a digit-suffixed name. The gate is a census of what IS authored, not of what the tokeniser CAN accept, so this gap is latent rather than live. It becomes live the first time an author copies `Building2` off lucide's site.

## Severity, stated honestly

No user-visible failure today — the same shape as objectui#7472 itself. What an author gets is the `Database` fallback rather than an error, so the symptom is a wrong glyph, not a crash or a blank.

## Not in this card

- ⛔ Widening `DECLARED_RECORD_READERS` or the gate's discovery predicate. Unrelated, and objectui#7472 already ruled that direction out for its own site.
- The 7 `Grid2x2`-shaped names. A single letter-to-digit rule does not fix them and they may not be worth a rule of their own; whoever takes this should decide that deliberately rather than inherit it.
- Whether the fix belongs in the tokeniser or in an explicit alias table. Both reach the 147; a tokeniser rule is smaller, an alias table is auditable. That is a design call for the taker.

Refs: objectui#5935 (the seam and its ruling) · objectui#7472 (where this was measured) · objectui#7830 (the adjacent, closed finding about identifiers imported as components that are not `icons` keys).

Generated by Claude Code, session `session_01FhBNJcLRZLe8M87VcUgpKr`.

os-tesla:
Fixes #9414

Session: `session_011QreXiyMEqKLN4U5daMPVa` — in prose because a footer link does not
survive an edit of this description.

## What was wrong

`packages/components/src/lib/lazy-icon.tsx`'s `toKebabIconName` split two boundaries —
`lower-or-digit -> Upper` and `acronym-run -> Word` — and never `letter -> digit`. So
`Building2`, the PascalCase component `lucide-react` exports and the spelling lucide's own
site shows an author, tokenised to `building2` while the canonical key is `building-2`. The
name matched nothing and `getLazyIcon` degraded it to the `Database` glyph with **no error,
no warning and no log** — the author saw *an* icon with no signal that it was not theirs.

## Re-measured at this branch's own base — the card's numbers were NOT inherited

Base `63d9ca6`, `lucide-react@1.35.0` as resolved from `packages/components`, the card's
reproducer run from that directory so the specifiers resolve:

```
namespace keys 6110 | canonical names 2039
non-alias keys the tokeniser drops 155
recovered by a digit-boundary split 147
CONTROL, must be true: true true
```

The `CONTROL` line is alive in that same run, so this is a reading rather than a void one.
**PR #8941 (lucide-react 1.43.0) had not landed at this base**, so the population is still
the one the card measured against.

Two differences from the card, both reported rather than smoothed over:

- **155, not 154.** The extra member is `default`, the namespace's own default-export key.
  It is not an icon and stays correctly rejected. The card's residual list of seven was
  otherwise exact.
- **The residual after the fix is two, not seven or eight.** The rule below reaches 153 of
  the 155, leaving `default` and `useLucideContext` — a default export and a hook. Neither
  is an icon, and both are still rejected.

The stronger reading, which is what the pin actually asserts: **every one of the 2,039
canonical icon names is now reachable from its own exported PascalCase spelling**, up from
1,882. ⛔ That number is not written into any source file — the test re-derives it from the
installed lucide on every run (AGENTS.md #9).

## The fix — one rule, and deliberately not an unconditional one

A third `.replace(...)` on the same chain: **match a letter immediately followed by a
digit and insert a hyphen between them, under a negative lookbehind that refuses the match
when the letter is itself preceded by a digit.**

⚠ Spelled in prose rather than pasted, and the reason is on-topic: the literal regex
opens with a question mark, a less-than sign and an exclamation mark, and GitHub's body
sanitizer eats that sequence as an HTML comment opener — **including inside a code
fence**. The first revision of this description was stored with the `!` silently removed,
turning the regex into a different one. The exact text lives in
`packages/components/src/lib/lazy-icon.tsx`.

The transform is the inverse of lucide's own `toPascalCase`, which builds each exported
component name by dropping the hyphens of a canonical name and upper-casing what followed.
The job is to put a boundary back exactly where `toPascalCase` removed one — and the
negative lookbehind is what stops it going further. When the letter is itself preceded by a
digit, it sits *inside* a segment rather than starting one: `Grid2x2` is the Pascal form of
`grid-2x2`, and an unconditional rule yields `grid-2x-2`, which is not a name.

## Design call 1 — a tokeniser rule, not an alias table

The card framed this as *smaller vs. auditable*, and the measurement inverts that framing.
A table would need 153 hand-listed rows that state a fact derived once and never again —
exactly the shape AGENTS.md #9 rules against — and would rot at every lucide bump with
nothing to notice. The rule is re-derived from its input on every call, and its acceptance
test enumerates the **whole** canonical population from the installed lucide, so it is
auditable by re-derivation rather than by reading a list. A table would also be a second
de-facto vocabulary standing beside the rule, which is the "N dialects" outcome
commandment #0.1 exists to refuse.

## Design call 2 — the `Grid2x2` family is IN

The card and triage both fenced this as the taker's call, so: **in**, and reached by the
*same single rule* rather than a second one.

The second rule the card anticipated would be a `digit -> letter` split. Measured, it
recovers five of the six by landing on lucide's **alias** spelling `grid-2-x-2`, and still
misses `Grid3x2` — for which lucide ships no `grid-3-x-2` alias at all. That leaves an
asymmetric hole that is more confusing than a clean stated fence. The lookbehind
refinement of rule 3 lands on lucide's **primary** `grid-2x2` instead, and reaches all six.
⇒ including them costs one lookbehind assertion, not a second rule, so leaving them out
would have been a fence with nothing behind it.

## The no-loss leg — ⛔ a count that only goes up is not a reading

Triage asked for proof that the 1,882 keys which already survived still survive
byte-identically. Measured over the whole population and pinned as two cases that run on
every CI run:

- every `lucide-react` export key the pre-#9414 tokeniser resolved produces the
  **byte-identical** output under the new one — the set of keys whose output moved is empty;
- every canonical kebab spelling is still returned untouched.

## Clause-②: no

`Contract-text:` the seam's published `.d.ts` — `packages/components` publishes
`files: ["dist", "README.md", "CHANGELOG.md", "LICENSE"]`, and `tsc` emits these doc
comments verbatim into `dist/lib/lazy-icon.d.ts`, where they have shipped since before this
card:

> Whether `name` (kebab-case or PascalCase) resolves to a real Lucide icon.

> Resolve a Lucide icon by name (kebab-case or PascalCase). ... Falls back to the `Database`
> icon when no `name` is provided or when the requested name is not a valid Lucide icon

and the published notifications guide, linked from that same published README:

> Every surface ... resolves `icon` through the same rule: a declared Lucide name
> (kebab-case or PascalCase) replaces the severity icon; anything else falls back to it.

> A name Lucide doesn't have costs the author their override and nothing more

⇒ the declared accept set is **real Lucide icon names, spelled kebab-case or PascalCase**.
`Building2` is precisely that: the PascalCase spelling `lucide-react` exports for the live
canonical key `building-2`. Not one of the names this change admits sits outside the
declared set, and the reject rule — a name Lucide does not have still degrades rather than
throwing — is untouched. So the accept set does not move; the implementation stops
under-delivering against it. That is the case the rules exempt from Clause ② in as many
words, which is why the claim's conservative `yes` is overturned here rather than carried.

⛔ The mechanical half is neutral, not supporting, and is reported as such:
`check-widening-tells.mjs --declaration no` exits 0 on this diff while printing
`5 NOT MEASURED ... no declared surface covers it`. An exit 0 that examined nothing is not
evidence, so the declaration rests on the contract text above.

## The pin, before and after

The gate is correctly green on the defect: `check:icon-record-names` censuses what IS
authored, and nothing in this tree authors a digit-suffixed name. So the test pins the
**capability** instead, and it was proved to fail before the fix — mutation proved on disk,
restore proved by state.

Mutation: rule 3 deleted from `lazy-icon.tsx`, i.e. the pre-#9414 tokeniser body restored.
Proved on disk by anchor count and by blob hash, ⛔ not by an editor's exit code:

```
MUTATION ON DISK: anchor 1 -> 0 ; blob e25f2199eefd3db3586df9d49bdee456b07acfbc -> a5c3ee15fd6cd13f46bcff3bc058da8f07ef2f41
```

With the fix reverted:

```
Test Files  3 failed (3)
     Tests  18 failed | 20 passed (38)
AssertionError: expected [ 'arrow-down-0-1', …(156) ] to deeply equal []
AssertionError: expected { kebab: 'building2', live: true } to deeply equal { kebab: 'building-2', live: true }
AssertionError: expected { kebab: 'grid2x2', live: true } to deeply equal { kebab: 'grid-2x2', live: true }
```

157 canonical names unreachable, which is exactly `2039 - 1882`. The two no-loss cases pass
under the mutation, which is the correct shape — they assert that nothing moved.

Restore proved by state, ⛔ not by re-running the test:

```
git diff HEAD: EMPTY
working blob : e25f2199eefd3db3586df9d49bdee456b07acfbc
HEAD blob    : e25f2199eefd3db3586df9d49bdee456b07acfbc
```

No build was involved in either leg and none was needed: the root `vitest.config.mts` maps
`@object-ui/components` to `packages/components/src`, so the mutated source is the subject
under test directly rather than a stale `dist`.

## Two existing pins recorded the defect as intended behaviour, and both move

- `lazy-icon-generated-app-contract-7472.test.ts` listed `Building2` among lucide's **alias**
  spellings that stopped resolving, beside `HouseIcon` and `LucideHouse`. It never was one.
  The two alias shapes are correctly described and are untouched here; `Building2` is
  canonical, and only this tokeniser could not reach it. The file now asserts the opposite,
  with the correction stated in place.
- `record-alert.severityIcons.test.ts` armed a tripwire reading "Should lucide ever add
  `check-circle2` to the dynamic surface, this line goes red — deliberately, because that
  would be worth a look." It fired, for a cause the note did not anticipate: lucide added
  nothing, and the seam learned the boundary. `CheckCircle2` moves from DARK to LIT; the
  DARK half moves to `LucideCheckCircle2`, a prefix alias that must stay unreachable, so the
  control still stands on the same footing.

## Interaction with PR #8941, stated rather than left to be discovered

The worked-example cases assert that each expected canonical name is a **live** member of
lucide's dynamic surface, so a lucide release that retires one reds here by design instead
of leaving the assertion restating a dead name. #8941 moves lucide to 1.43.0. If it lands
first, this branch should be merged with `origin/main` and the suite re-run; a red in those
cases is the pin doing its job and names the retired spelling directly. The universal leg
needs no such care — it is derived from whatever lucide is installed.

## What ran

| leg | result |
|:--|:--|
| new pin `lazy-icon-digit-boundary-9414.test.ts` | 20 passed |
| `packages/components/` + `packages/plugin-detail/` full suites | 4,266 tests — 2 failed on the first pass, both the stale pins above; green after they moved |
| every test file touching the seam or authoring a digit-bearing name (28 files, derived by grep) | 28 files, 309 tests passed |
| `turbo run type-check` for both packages | 14 tasks successful |
| `turbo run lint` for both packages | 3 tasks successful, 0 errors |
| `check:icon-record-names` | `OK ... 184 authored/declared names reaching 1 record-reading resolver are live` — unchanged, and ⛔ the gate was NOT widened |
| `check:control-bytes` · `check:changeset-presence` · `changeset:check` · `check:changeset-claims` · `check:changeset-overwrite` · `check:new-line-citations` · `check:test-path-roots` · `check:comment-mask-corpus` | all green |
| `check:readme-exports` | NOT MEASURED — it needs every package built and reports `the population COLLAPSED, this run proves nothing` otherwise. No export surface moved in this diff, so it is outside what this change owes; CI runs it. |
| repo-wide sweeps (`pnpm test`, `pnpm lint`, full `type-check`) | CI's run, not duplicated here |

## Acceptance notes

- The pending changeset for objectui#7472 lists "digit-suffixed spellings such as
  `Building2`" among lucide's alias forms that stopped resolving. The two alias shapes it
  names are correct; `Building2` never belonged with them. ⛔ That changeset was NOT edited —
  it belongs to another PR and `check-changeset-overwrite` exists for exactly that reason.
  The correction is carried in this PR's own changeset, which lands in the same CHANGELOG
  section, so a reader meets both together. Noted, not filed.
- `packages/components/src/renderers/action/__tests__/resolve-icon-seam.test.ts` pins
  `resolveIcon('building_2')` against the `icons` record — a different seam on the same
  subject, touched by PR #9399 and not by this one. It is unaffected and passes here.
  Noted, not filed.
- The `icons`-record resolver (`resolve-icon.ts`) has its own tokeniser and is out of this
  card's fence. Whether the two seams should agree is a live question that this change does
  not answer and does not make worse. Noted, not filed.
- ⛔ `DECLARED_RECORD_READERS` and the gate's discovery predicate were not widened, per the
  fence objectui#7472 set and this card inherited.

---


os-tesla:
## `Bundle Analysis` — the gate and the reason, recorded on the PR as the gate itself asks

Posted by the `domain:ui` PM seat. This is the record the enqueue rule requires **on the pull request** (⛔ not on the card) before a PR is queued carrying this red.

**The gate**: `Bundle Analysis` (`.github/workflows/performance-budget.yml`). It is a **blocking** check when it runs — `scripts/dependabot-merge-gate.mjs` classifies it under `OPTIONAL_CONTEXTS`, where "optional" is about the trigger-level **path filter**, ⛔ not about the check being advisory: *present ⇒ must be `success`; absent ⇒ not waited for*.

**The reason it is red here**, quoted from the gate's own output rather than paraphrased:

```
❌ chunk `ui-components`  387.9 KB measured / 389.6 KB ceiling
   (headroom 1.7 KB = 0.02x the 89.0 KB regression, under the 0.10x floor
   and held open by its declared 4289-byte allowance)
chunk `ui-components` was ALREADY declared exhausted before this run, and has
now lost a further 0.01x of a regression against the pinned figure.
```

and the gate's own instruction, verbatim:

> ⚠️ **READ THIS BEFORE AUDITING YOUR OWN DIFF.** This row's headroom is a standing debt that predates this change, and it moves under traffic that has nothing to do with the chunk — measured at three gzipped bytes across five unrelated merges. So this verdict is **NOT** an accusation that your diff spent the bytes…
>
> ⛔ There is therefore **nothing here for this pull request to "fix"**, and the two edits that would turn this green are both forbidden: ⛔ never raise the ceiling, and ⛔ never raise the allowance. Paying the row down is the open decision on the chunk, ⛔ not a task for whichever change the queue happened to weigh — **take it there, and say on this pull request that you did.**

⇒ Taken there: the debt is tracked at objectui#9251 / objectui#9204. ⛔ No duplicate filed, and ⛔ nothing in this PR attempts to turn the row green.

**The three conditions for carrying it red into the queue**, each verified structurally this round:

1. **Red by design on a pushed branch** ✓ — the gate says so in its own output above: nothing here for this PR to fix, and both green-making edits forbidden.
2. **Does not run on `merge_group`** ✓ — the workflow's `on:` carries only `push` and `pull_request`; the sole `merge_group` mentions in the file are in a comment explaining it cannot hold the queue.
3. **This comment.**

⚠️ Correction of this seat's own earlier record, so the bad sentence does not propagate: this seat previously wrote that the gate is *"red on `origin/main`'s own tip"*. That is **withdrawn** — `main`'s tip often carries no `Bundle Analysis` run at all, because the `push:` trigger is path-filtered, and a bare zero there reads as a pass while meaning nothing of the kind. The durable statement is **red on 7 of the last 7 `main` commits that actually ran it**. The entry-chunk half of the gate is green throughout (144.5 KB against a 350 KB line); only the `ui-components` closure row is red.


---

PR #9460, opened 2026-09-14
fix(plugins): row/card click props declare the modifier payload they are invoked with

os-tesla:
Part of objectui#9357 — the consumer half. The hook half landed as objectui#9360.

⚠️ Deliberately `Part of`, not a closing keyword: two sites that share the name
and the shape are left alone here with reasons, and one of them is a judgement
the in-seat review should confirm before this card is retired. See
**What is deliberately left** below.

## What this is

objectui#9360 made `UseNavigationOverlayOptions.onRowClick` declare both
arguments `handleClick` hands it — the record, and the optional modifier payload
a host needs for Cmd/Ctrl/middle-click. It deliberately stopped at the hook.

This is the rest: the component props that pass a host-supplied handler straight
into that option declared only the record, so the payload stayed invisible on the
one line a host actually reads.

**13 declarations across 9 packages**, each read individually for whether it is
on the forwarding path:

| declaration | the hop that puts it on the path |
|---|---|
| `ObjectGridComponentProps.onRowClick` | fed to `useNavigationOverlay` as its `onRowClick` |
| `ObjectKanbanComponentProps.onRowClick` | the `onRowClick` arm of `externalClick`, fed to the hook |
| `KanbanRendererProps.schema.onCardClick` | handed to `KanbanImpl`, whose `SortableCard` invokes it with the DOM event |
| `ObjectCalendar.onRowClick` | fed to the hook |
| `ObjectGantt.onRowClick` | fed to the hook |
| `ObjectMap.onRowClick` | fed to the hook |
| `ObjectTree.onRowClick` | fed to the hook |
| `ObjectTimeline.onRowClick` + `.onItemClick` | the two arms of `onRowClick ?? onItemClick` |
| `ListView.onRowClick` | fed to the hook |
| `ObjectGallery.onRowClick` + `.onCardClick` | the two arms of `props.onRowClick ?? props.onCardClick` |
| `RelatedList.onRowClick` | placed on the `object-gallery` schema it builds, which the renderer spreads into `ObjectGallery`'s props |

⭐ `ObjectKanbanComponentProps` is the clearest case in the tree: its two arms of
one `??` sat twelve lines apart, `onCardClick` already declaring both parameters
since objectui#9341 and `onRowClick` still declaring one.

## ⛔ The census was a population, not a defect count

The handover measured 24 one-parameter occurrences. Eleven are deliberately NOT
touched, each for a reason that had to be read off the tree rather than inferred
from the name:

| site | why it is out |
|---|---|
| `VirtualGrid.onRowClick` | second parameter is `index: number` — a different contract |
| `ManageViewsDialog.onRowClick` | invoked as `onRowClick?.(view.id)` — a view id, not a record callback |
| `DataTableSchema.onRowClick` | `data-table` invokes `schema.onRowClick(row)` with ONE argument |
| `ObjectDataTableSchema.onRowClick` | forwards into that same `data-table` channel |
| `plugin-view/ObjectView.onRowClick` (two of them) | its own `handleRowClick` truncates to `onRowClick(record)` |
| `RecordPickerGridSlotProps.onRowClick` | provider-side; the dialog hands its own handler OUT and terminates the chain in selection logic |
| 1 README example, 4 test fixtures | a one-parameter handler is still valid, and two are source-text fixtures pinning other cards' shapes |

For the `data-table` family and `plugin-view/ObjectView`, widening the
declaration **without** changing the call would have declared a payload that
never arrives — a contract lie in the opposite direction, and the thing
`AGENTS.md` #0.1 exists to stop.

## What is deliberately left

Those last two are the same **separate** defect: a hop that *receives* the
payload and drops it. `packages/components/.../data-table.tsx` has the DOM event
in scope at both row-click sites and calls `schema.onRowClick(row)`; the same is
true of `plugin-view/ObjectView`'s `handleRowClick`. Consequence: Cmd/Ctrl-click
through the table body reaches no host handler, whatever the declarations say.

That is a behaviour change, not a declaration change, so it is reported rather
than folded in here. **The reading that settles whether it belonged to this
card**: if the card's scope is *declarations that understate an actual call*,
these are out (their calls really are one-argument and the declarations are
honest). If the scope is *every hop between the hook and a host*, they are in,
and the fix is a forward at each hop, not a widening.

## Source compatibility — measured, not assumed

Both directions hold, so **no caller of these props is refused**:

- a one-parameter handler still satisfies the widened declaration;
- a handler written against the widened one already satisfied the narrow one (its
  minimum argument count is still 1).

And unlike objectui#9360, **there is no refused class here**. The hook names
`HandleClickModifiers` and therefore refuses a second parameter annotated
narrower — React's mouse event being the shape that hits in practice. These faces
spell it `any`, so that annotation keeps compiling. That is the whole reason for
the spelling: the population most likely to be passing a second argument already
is precisely the one that discovered the payload from the implementation, where
it is a React mouse event. `app-shell` spells its three host call sites
`(record: any, event?: any)` for exactly that reason, and they are the only
in-repo hosts that pass one.

`any` is also what `ObjectKanbanSchema.onCardClick` already carries for this same
payload (objectui#9341), what `BaseSchema`'s own `onClick` / `onChange` /
`onSubmit` use, and what is forced on the published twins in `@object-ui/types`,
which may not name a type living in a package that depends on them.

**Runtime behaviour is unchanged** — nothing is newly called, and no call site
newly passes an argument it did not pass before.

## The instrument, and how it was self-tested

`packages/react/src/hooks/__tests__/navigationOverlayConsumers.onRowClickArity-9357.test.ts`.

An assignability pin would assert nothing here — the two spellings satisfy each
other in both directions, so `extends` is green on the broken tree and on the
repaired one alike. The pin reads the **parameter list** two ways: a compile-time
exact-identity read of `Parameters` on the two spellings (no plugin package is
imported — `@object-ui/react` is what the plugins depend ON), and a **bytes** read
of every declaration off disk with a depth-aware arity counter.

⭐ Two instrument lessons from the handover were applied and are pinned in the
file:

1. **A control must vary only the claim.** The handover's first pathspec returned
   zero files and its control returned zero too, because the control reused the
   same pathspec. Mechanism, now measured: a **glob** pathspec must match the
   WHOLE path, so `packages` slash star slash `src` matches a directory and no
   file. Lit controls, same tree, one variable each:

   | pathspec | term | files |
   |---|---|---|
   | `packages` | onRowClick | **75** |
   | `packages` slash star slash `src` | onRowClick | 0 |
   | `packages` | useNavigationOverlay (known present) | **63** |
   | `packages` slash star slash `src` | useNavigationOverlay (known present) | **0** ← the pathspec is the broken part, not the term |
   | `packages` slash star slash `src` slash star star | useNavigationOverlay | **54** |

   The same discipline is in the pin itself: `RelatedList` reaches the hook
   through a schema, not by calling it, so each IN site carries its **own**
   forwarding-hop regex instead of one blanket pattern that would have hidden it.

2. **Self-test the counter before pointing it at the tree.** A naive comma split
   reads a single parameter whose type carries a comma as **two**, which would
   have scored seven one-parameter sites "already fixed". The counter here is
   depth-aware and is self-tested on nine legs — one-, two- and zero-parameter,
   nested generics, an object-literal parameter, a callback-typed parameter, and
   three non-declarations — **and the naive counter's wrong answer is pinned
   beside it**, so the two can never be read as equivalent. Pointed at the tree
   it independently reproduced the handover's census exactly: 24 files, 27
   occurrences, 24 one-parameter and 3 two-parameter.

## Ablation — prediction written before the run

Predicted in writing at commit `4bff3a7`, before either mutation was applied:

| mutation | predicted | observed |
|---|---|---|
| narrow `ObjectGrid.onRowClick` back to its pre-fix spelling | **turn red**, exactly 1 leg (the ObjectGrid IN leg), 21 passed / 1 failed; compile-time half unmoved | exit 1 — `expected [ 'record: any' ] to have a length of 2 but got 1`, **21 passed / 1 failed** |
| widen the out-of-scope `ManageViewsDialog.onRowClick` | **turn red**, exactly 1 leg (its OUT control leg), 21 passed / 1 failed | exit 1 — `expected [ 'id: string', 'event?: any' ] to have a length of 1 but got 2`, **21 passed / 1 failed** |

The second mutation is the one that matters for this card: the first proves only
that the IN legs fire, and says nothing about the OUT legs — which carry the
whole risk of treating a population as a defect count. A control that never fires
is not a control.

Both mutations were proven on disk before the reading (anchor counts in both
directions, plus a changed blob hash), and both restores proven **by state** —
`git checkout HEAD --` on an absolute path, then blob hash equal to the HEAD blob
hash AND `git diff HEAD` empty. The runner carried a `trap` on EXIT/INT/TERM.

## Test scope

⛔ No `--shard` anywhere: it is not reproducible between CI and this container.
Every reading below is a whole-tree or whole-package run.

**Tests — the affected packages' FULL trees, not just the edited files.** Repo-root
invocation, paths relative to the repo root, nothing after a bare double-dash
(the two forms AGENTS.md records as silently false-green):

```
pnpm exec vitest run packages/plugin-grid/ packages/plugin-kanban/ \
  packages/plugin-calendar/ packages/plugin-gantt/ packages/plugin-map/ \
  packages/plugin-timeline/ packages/plugin-tree/ packages/plugin-list/ \
  packages/plugin-detail/ packages/react/

  Test Files  695 passed (695)
       Tests  6523 passed (6523)          exit 0
```

Files verified **directly** by name, re-run after the last commit:
`packages/react/src/hooks/__tests__/navigationOverlayConsumers.onRowClickArity-9357.test.ts`,
`packages/plugin-list/src/__tests__/ListView.propsResolution.test.ts`,
`packages/react/src/hooks/__tests__/useNavigationOverlay.onRowClickArity-9357.test.tsx`
— 3 files / 34 tests, exit 0.

**Type-check — 44 of 44 tasks successful, exit 0**, over 15 packages: the nine
this PR edits, plus `@object-ui/react`, and then `app-shell`, `plugin-view`,
`components`, `fields` and `types` — i.e. every in-repo consumer of the moved
faces. `app-shell` is the one that matters for compatibility: it holds the only
three in-repo host handlers that pass a second argument.

⚠️ Two readings from the FIRST type-check run, recorded rather than smoothed over:

- It was genuinely **red**, on `ListView.propsResolution.test.ts(52,3) TS2344` —
  an existing objectui#4528 pin that reads this exact prop through an
  exact-identity `Equal`. That is the pin working, and it is the sharpest
  evidence in this PR that the change is real: it reds on a widening that
  `extends` waves through in both directions. It is also a reading vitest cannot
  produce — those assertions are erased at runtime, and the 695 test files above
  were green while it was red. Repaired in the second commit.
- The same run printed `ELIFECYCLE Command failed.` for `plugin-detail` with **no
  error line**: turbo aborted it when its sibling failed. That is **NOT
  MEASURED**, not red and not green, so the re-run adds `--continue` and
  `plugin-detail` is measured on its own.

**Lint — 11 of 11 tasks successful, exit 0** (`turbo run lint`, the 10 affected
packages): 216 problems, **0 errors**, 216 warnings. `lint.yml` records that
`--max-warnings` is deliberately not set, so warnings are not a failure here; the
new `event?: any` lines are among them, under `no-explicit-any`, which this repo
configures at `warn`.

**Declared narrowing, with its three pieces.** (i) Population read from turbo's
own task graph, not guessed: **47** lint tasks exist in this workspace; this
branch ran **10**; the other **37** are CI's. (ii) Every package this diff
touches is in the ran set (checked as a set equality, not by eye). (iii)
Invariance: `eslint.config.js` names no `projectService`, `project:` or
`parserOptions`, so linting here is not type-aware and each package's `eslint .`
reads only its own files — this diff cannot move a verdict in a package it does
not touch. The repo-wide run is CI's, never one this branch owes.

**Gates**, all exit 0: `check:control-bytes`, `check:handler-key-reads`,
`check:changeset-claims`, `check:test-path-roots`, `check:new-line-citations`,
`check:doc-fences`, `check:changeset-presence`, and post-build
`check:dist-completeness`, `check:doc-snippets`, `check:doc-examples`,
`check:phantom-deps`.

⚠️ `check:readme-exports` first exited 1 with "its type entry is not on disk —
run `pnpm build` first". That is a **prerequisite not met**, which reads NOT
MEASURED — ⛔ not a red gate and ⛔ not a green one. After `pnpm build` (43 of 43
tasks successful) it exits 0 and its own census confirms the earlier reading was
the prerequisite and nothing else: **0 unbuilt** where it had said 8, **0
unjudgeable** where it had said 1, and 58 keys compared where it had managed 28.

A control-byte scan was also run over the changed files directly, outside the
gate: zero hits.

## Labels

`needs:contract-review` attached to both carriers (this PR and objectui#9357):
these are published component-prop faces across nine packages.

---

Session (in prose and as a code span, because a markdown-link footer does not
reliably survive a later `PATCH`): `https://claude.ai/code/session_011QreXiyMEqKLN4U5daMPVa`

---


os-tesla:
## Contract review

Head reviewed: `2c937d371ca8ad071dcd823d590aa32dcd3275f8`
Seat: `domain:ui` (`os-tesla`) · **default judgment tier** — the contract-review tier and its fuse bind only the spec and skills seats.

```
Implemented-by: claude/issue-9357-onrowclick-modifier-arity
```

### ① Derived judgments

**13 declarations widened across 9 packages**, each from `(record) => void` to `(record, event?: any) => void`. Verified at source: every widened line in the diff carries `event?: any`, and **every named OUT site is absent from the diff** (`VirtualGrid`, `ManageViewsDialog`, `data-table`, `plugin-view` — the file list has no hunk for any of them).

**Source-compatible for callers**, which is the judgement that matters on a published face: an existing one-parameter handler still satisfies a two-parameter optional signature. Nothing is newly called and no call site newly passes an argument — the runtime is untouched. ⇒ the accept set **widens** (a caller may now supply a handler that reads the payload) and nothing narrows.

⭐ **The spelling is the sharpest judgement in the report.** `event?: any`, ⛔ not `HandleClickModifiers`, because `any` **refuses nothing** whereas the named interface would refuse a second parameter annotated `React.MouseEvent` — which is exactly what a host who discovered the payload from the implementation would have written. Seconded by two independent constraints: `any` matches `ObjectKanbanSchema.onCardClick` (objectui#9341) and `BaseSchema`'s own `onClick`/`onChange`/`onSubmit`, and it is **forced** on the published twins in `packages/types`, which may not name a type living in a package that depends on them. A narrower spelling would have been a new refusal shipped under the banner of a widening.

**The IN/OUT criterion is behavioural, ⛔ not nominal**: *IN iff the value flowing through the declaration is actually invoked with the payload.* That is why `RelatedList` is IN despite never calling the hook — it reaches `ObjectGallery` through the `object-gallery` schema it builds, which `SchemaRenderer` spreads into props — and why the data-table family is OUT despite copying the spelling byte-for-byte.

### ② Semver

Widening optional callback parameters on published faces is **additive** and source-compatible. No API removed, none retyped in a narrowing direction. `check-changeset-presence` and `check:changeset-claims` both exit 0 on the branch. Consistent.

### ③ Boundary flags

- ⭐ **The first `Type Check` run was genuinely RED** and is reported as such: `ListView.propsResolution.test.ts(52,3): error TS2344` — the existing objectui#4528 pin, which reads this prop through an **exact-identity `Equal`** and therefore separates the two spellings where `extends` could not. Repaired in `2c937d3`, which also adds a **two-parameter probe beside the existing one-parameter probe**, so source compatibility is asserted where the compiler checks it rather than claimed in prose. ⚠️ And the dev names the trap: *"that red is invisible to vitest — those assertions are erased at runtime, and the 695 files above were green while it was red."*
- ⭐ **`turbo` aborted `plugin-detail`'s type-check when a sibling failed, printing `ELIFECYCLE` with no error line** — read as **NOT MEASURED**, ⛔ neither red nor green, and the re-run added `--continue` so it was measured on its own. That distinction is the whole discipline in one move.
- `check:readme-exports` first exited 1 with a prerequisite message; recorded **NOT MEASURED**, ⛔ not as a red gate, then re-run green after the build — and its own census proves the first reading was the prerequisite (0 unbuilt where it had said 8, 58 keys compared where it had managed 28).
- Lint narrowing declared with all three pieces, including **set equality** against turbo's own task graph (not eyeballed) and an invariance argument from `eslint.config.js` naming no `projectService`/`project`.

### ⭐ RULING on the open question — **A**

`plugin-view/ObjectView` (×2) and the data-table family are **OUT** of this card.

They copied the understated spelling byte-for-byte, but their own calls really are one-argument, so **their declarations are currently honest** — they are hops that *receive* the payload and *drop* it. Widening them without repairing the calls would declare a payload that never arrives: the same contract lie this card exists to remove, pointing the other way. And repairing the calls makes Cmd/Ctrl-click newly reach host handlers — a **behaviour** change this card did not ask for and did not price.

⇒ objectui#9357 is *"declarations that understate an actual call"*. The dropped-payload defect is a different card, correctly filed **bare** as objectui#9462 with the failing probe named and `ObjectGallery` (which does forward) as the lit control. ⭐ The ambiguity is written into that card, so if this ruling is ever reversed the cost is re-pointing #9462, ⛔ not redoing work.

### Ablation

Two mutations, predicted in writing first. ⭐ **Mutation 2 is the one that earns the review**: it widens the deliberately-**out**-of-scope `ManageViewsDialog`, and the dev states why — *"mutation 1 alone would have proven only that the IN legs fire and said nothing about the OUT legs, which carry this card's whole risk."* Both landed exactly as predicted (`21 passed / 1 failed` each, with the quoted assertion text). Mutations proved on disk in both directions; restores proved by state.

The instrument was **self-tested on 9 known legs before the tree**, and the naive comma split's **wrong** answer is pinned beside it so the two can never be read as equivalent. Pointed at the tree it independently reproduced the handover census exactly (24 files, 27 occurrences, 24/3).

⭐ And it closed the loop on the lesson this card's handover taught: the zero-file pathspec is **explained**, not merely avoided — a git glob pathspec must match the *whole* path, so `packages/*/src` matches a directory and never a file — with a four-row lit-control table isolating the pathspec as the broken component rather than the search term.

### Verdict

**PASS.**

Carriers stripped in the same stroke as this record; landing follows on a **full green** read, ⛔ not the required subset alone.


---

PR #9465, opened 2026-09-14
refactor(app-shell): bind or rename the six spec-symbol mirrors, shrinking the ledger 11 to 5 (objectui#7265 app-shell slice)

os-tesla:
Part of objectui#7265 — the `@object-ui/app-shell` slice. ⚠️ That card stays open: it is one ledger worked in slices by design, and this takes one package's worth of it. Slice 1 (`@object-ui/core`, 2 entries) landed as objectui#9127.

## First: the ledger, re-derived rather than inherited

The card's body says 14. The seat that landed slice 1 said 11 and said plainly it could not re-measure. Neither number was taken on faith — `--ledger` is the script's own authoritative reading and it was run here, in a worktree with dependencies installed, exit code captured by redirect before any pipe:

```
node scripts/check-spec-symbol-derivation.mjs --ledger     # exit 0
```

**Before this PR: 11 names in 5 packages.** The gate's own banner agrees, which is the second reading of the same population: `11 untriaged collisions in 5 packages`.

| package | before | after |
|---|---|---|
| `@object-ui/app-shell` | 6 | **0 — group gone** |
| `@object-ui/types` | 2 | 2 |
| `@object-ui/components` | 1 | 1 |
| `@object-ui/data-objectstack` | 1 | 1 |
| `@object-ui/plugin-detail` | 1 | 1 |
| **total** | **11** | **5** |

`@object-ui/plugin-tree` / `TreeConfig` is genuinely gone — the card's 14 is stale by one entry beyond slice 1's two, exactly as slice 1 reported.

⚠️ Rule 2's `CLAIM_DEBT` came back **byte-identical** from `--claim-ledger`. Slice 1 was FORCED to regenerate it too, because the const it deleted carried a `Mirrors …` comment; none of the seven declarations here carried a spec-alignment claim, so that coupling did not repeat. Worth recording, since slice 1 told the next slice to expect it.

Both blocks were regenerated mechanically. Neither was hand-edited.

## Six names, seven sites, four routes — ⛔ not one mechanical pass

`ObjectLike` had two sites, and they did not get the same answer, which is the whole argument for reading each entry rather than sweeping them.

| entry | site | disposition | why |
|---|---|---|---|
| `RemoteTable` | `datasource/DatasourceResourcePage.tsx` | **burned down — imported** | The same wire row `GET …/remote-tables` returns. The sibling client one directory over (`external/api.ts`) already re-exports the spec's type, with the drift that motivated that call written down in place. The copy here had `columnCount` optional where the contract requires it. |
| `AdminScope` | `metadata-admin/PermissionAdvancedFacets.tsx` | **burned down — derived** | `Partial<>` of the spec's authoring shape. Same six keys, same order; the editor materializes the facet from `asObject(draft.adminScope)`, so a half-filled draft is the normal mid-edit state and the widening is confined to requiredness. |
| `FlowRuntimeState` | `studio-design/StudioDesignSurface.tsx` | **burned down — derived** | `Partial<>` of the spec's report. It is the same `getFlowRuntimeStates` contract, read as UNVALIDATED wire JSON from a possibly-older backend — the effect already narrows all three by hand. The local copy was three keys behind the contract. |
| `ObjectLike` (site 1) | `hooks/useTrackRouteAsRecent.ts` | **burned down — derived** | `Pick` of the spec's minimal object document, down to the two members the hook reads. |
| `ObjectLike` (site 2) | `utils/deriveRelatedLists.ts` | **renamed** to `MergedObjectLike` | Carries `list`, which `MetadataProvider.mergeViewsIntoObjects` merges on after authoring. The spec's shape is closed and has no member for it, and `name` is optional here because this also describes an object still loading. A different layer, so the spec's name goes back. |
| `AppLike` | `utils/appRoute.ts` | **renamed** to `AppRouteLike` | The spec's `AppLike` is what `translateApp` consumes. These helpers read `_packageId` (the ADR-0048 route key) first, which that shape does not declare — it would arrive through its index signature as `any`, taking the `typeof packageId !== 'string'` guard with it. And `resolveHostAppSegment`'s documented step 4 exists FOR an app with neither `_packageId` nor `name`, so a required `name` would assert away the branch it was written to serve. |
| `FlowEdge` | `metadata-admin/inspectors/FlowEdgeInspector.tsx` | **renamed** to the existing dialect `FlowDesignerEdge` | ⭐ Not a spec question at all once read: this was a THIRD hand copy of this package's own `FlowDesignerEdge`, which two sibling files already take from `flow-canvas-layout` and where the divergence from the spec is already reasoned. No new name was invented. |

⇒ **4 bound to the spec, 3 renamed, 0 moved to `ALLOW`, 0 left in place.** No `ALLOW` entry was added, so the declared-dialect count is unchanged at 17 — a reasoned waiver would have been the wrong tool for all seven sites.

Both renames get the two-way ratchet the package already runs in `spec-symbol-parity.test.ts`: the spec must not own the new name, and must still own the old one, so neither workaround can outlive its reason.

## Evidence

**Test scope.** The WHOLE `packages/app-shell` tree, ⛔ not just the edited files, run from the repo root through the shared verify lock, which printed its own verdict line:

```
os-verify-lock: VERDICT command-exit 0 · held the lock 821s
Test Files  701 passed (701)
     Tests  6860 passed | 1 skipped (6861)
```

Plus the scripts-side suite the edited gate owes (found with `git grep -l`, not recalled): 5 files, 178 tests, exit 0. `type-check` exit 0 for both projects; `--listFiles` confirms all seven edited files are in the compiled set, with a negative control that returns 0.

**Ablation — prediction written BEFORE running, both legs.**

| leg | varies | predicted | observed |
|---|---|---|---|
| A (IN) | restore the `RemoteTable` mirror at its site, ledger left as this PR leaves it | RED **by site**, ⛔ not a stale-ledger message | gate exit 1, `RemoteTable` reported as an untriaged collision at `DatasourceResourcePage.tsx:138`; new pin failed with `expected [ 'RemoteTable' ] to deeply equal []`, 1 failed / 18 passed |
| B (OUT) | delete ONE **OUT** row (`SortDirection`, `@object-ui/components`) from the block, ⛔ zero source files touched | RED from a DIFFERENT instrument, naming a package this slice never edits; my app-shell pins stay GREEN | gate exit 1 naming `SortDirection` at `packages/components/src/renderers/complex/data-table.tsx:65`; app-shell pins 19 passed |
| C (control) | nothing — same gate, unmutated tree | exit 0, else leg B's red proves nothing | exit 0 |

Leg B is the one that matters: a sibling card this shift showed that mutating only the IN legs proves nothing about the OUT legs. Every mutation was proven on disk by anchor count AND blob hash before any result was read, and every restore was proven **by state** — `git diff HEAD` empty and the blob hash back to HEAD's — ⛔ never by an exit code. Each script carried `trap … EXIT INT TERM` with absolute repo-root paths.

**The compile-time pins are compiled.** This file's own header records that its `Assert` lines were once literally commentary. Falsifying the new published-face pin reds `tsc -p tsconfig.test.json` with `TS2344: Type 'false' does not satisfy the constraint 'true'` at the exact line, so they can fail.

**Gates.** `check:spec-symbols` `check:control-bytes` `check:new-line-citations` `check:changeset-claims` `check:unreferenced-sources` `check:test-path-roots` `check:comment-mask-corpus` `check:installed-pin-claims` `check:phantom-deps` `check:unused-deps` — all exit 0. `check-changeset-presence` and `check-changeset-no-major` exit 0. `check-governed-queue-guard --test` on all 11 paths: NOT GOVERNED.

`check:spec-floors` exits 1 with 3 findings, all `no-artifact` for `plugin-gantt` / `plugin-timeline` / `plugin-tree` — a whole-workspace-build PRECONDITION this run did not meet, ⛔ NOT MEASURED rather than red. `@object-ui/app-shell` is judged and clean in the same run (18 of 39 packages inspected, 309 pairs judged). The floor question was also answered directly: all four newly imported symbols are present in `@objectstack/spec@17.0.0` — the declared floor — checked against the registry tarball with a negative control name that returned 0 hits.

`pnpm lint` is a whole-farm run and belongs to CI; the narrowing here is declared with its three readings. Population: 4965 of 4967 lintable tracked files in scope, read from eslint's own resolved config via `isPathIgnored`. Narrowed run: 10 files counted from `--format json`, exit 0, 0 errors / 35 warnings — and the SAME command over the same paths at the branch point gives 0 errors / 35 warnings, so this diff adds no finding. Invariance: this repo does not enable type-aware linting (`tseslint.configs.recommended`, no `parserOptions.project`, no `projectService`), so the diff cannot move any untouched file's verdict.

## Clause-② — declared `yes` on the claim, and the diff was measured

The measurement, taken by building `packages/app-shell` at this branch's head and again with the seven source paths reverted to the branch point:

- **all 461 emitted `.js` files are byte-identical**, `diff -rq` exit 0. The same command reported 3 differences on the `.d.ts` pass in the same session, so it is an instrument that can report one;
- **no name enters or leaves the published export set**;
- 3 `.d.ts` files move BYTES. Two are identical once comments are stripped and the renamed identifier is substituted — and neither identifier was ever exported, so no consumer can name either;
- the third (`useTrackRouteAsRecent`) reaches consumers through the exported `UseTrackRouteAsRecentOptions`, and is pinned at compile time to the exact shape the hand-written interface declared;
- no new spec subpath enters the published declarations — `@objectstack/spec/system` was already imported there, and the package is a direct `dependencies` entry.

⇒ On the structural reading the honest answer is that this diff moves **no published face**. ⚠️ The label is attached anyway, on both carriers, because the declaration limb the live carrier check reads is the CARD'S CLAIM COMMENT, which says `Clause-②: yes` — a seat re-declaring that unilaterally would create exactly the carrier half-state the patrol exists to catch. The measurement above is offered so a reviewer can clear it quickly, ⛔ not as a reason to skip the review.

## Acceptance notes

- **filed as objectui#9464** — Studio's Delegated Admin Scope editor can write an `adminScope` the spec rejects: every control there patches key-by-key, so flipping a switch before naming a business unit writes `{ includeSubtree: true }`, which `AdminScopeSchema` refuses (`businessUnit` is its one required key). There is no client-side parse before Save, and the section's own summary badge reads `0` in that state, pointing the author away from the cause. Behaviour is IDENTICAL before and after this diff — binding the type made the divergence legible, it did not create it. Dedup ran one targeted semantic search with a lit control in the same minute: control 2 hits including a closed card, subject 0.
- *noted, not filed*: `FlowNodeInspector.tsx` and `FlowPreview.tsx` still write `type FlowEdge = FlowDesignerEdge`, keeping a spec-owned name as a local alias. The gate skips these deliberately (`isPureAlias`), so they are not ledger entries and not a defect — but the reader hazard the ledger exists for is the same one. Carrier: none; recorded here rather than filed because the gate's narrowing is a decision, not an oversight.
- *noted, not filed*: this card's BODY census is now stale by six more names. The block is the readable source, not the body. Carrier: the next objectui#7265 slice, which reads the ledger.

## Notes for the next slice

The five that remain are ⛔ not one class, and the two hardest are still there: `normalizeFilterOperator` is a second normalizer over the spec's (a behaviour question, not a mirror deletion) and the `@object-ui/types` pair sits on `domain:spec`, which triage flagged for hand-off. What this slice adds to slice 1's method: where the concept really is the spec's but the local type must stay wider, a type OPERATOR (`Partial` / `Pick`) over the spec's symbol keeps the key set bound while stating the widening in one place — the local declaration stops being able to drift, which a hand copy with the same members cannot promise.

Session: `https://claude.ai/code/session_011QreXiyMEqKLN4U5daMPVa`

---


os-tesla:
## Contract review

Head reviewed: `e0e3c674bc8df6a2a18e2e2e7f27f07d2865804c`
Seat: `domain:ui` (`os-tesla`) · **default judgment tier**.

```
Implemented-by: claude/issue-7265-app-shell-spec-symbol-mirrors
```

### ① Derived judgments

⭐ **The declaration was MEASURED structurally, not argued.** `packages/app-shell` was built at head and again with the seven source paths reverted to the branch point: **all 461 emitted `.js` files byte-identical** (`diff -rq` exit 0) — and the instrument is **lit**, because the same command reported 3 differences on the `.d.ts` pass in the same session. No name enters or leaves the published export set; 3 `.d.ts` move bytes, two identical once comments are stripped and the renamed non-exported identifier substituted, the third pinned at compile time to the exact shape it replaced.

Corroborated independently at this seat: `packages/app-shell/src/index.ts` is **untouched** in the diff. ⇒ **no published face moves.**

⭐ **And the dev did NOT re-declare to `no` on that finding — which is the sharpest piece of protocol reasoning this shift.** Its reason, verbatim in substance: the declaration limb the live carrier check reads is **the card's claim comment**, which says `Clause-②: yes`; a dev re-declaring that unilaterally would manufacture exactly the carrier half-state the patrol exists to catch. So it attached `needs:contract-review` to both carriers, said why, and let the seat rule. `--pair 9465` ⇒ **exit 0**, both carriers agree.

⇒ **This seat's ruling: the `yes` stands as the declared state and the review discharges it.** The structural measurement is recorded here as the *finding*, ⛔ not as a retroactive re-declaration — a declaration is only as current as the commit it was made against, and the right place to correct one is a review record like this, not a unilateral edit.

**Six names, seven sites, four routes — deliberately not one pass**, which is what the dispatch asked for and what a blanket "import from spec" would have destroyed:
- **Bound to the spec (4)**: `RemoteTable` imported outright (the sibling client in that same package had already made that call); `AdminScope` and `FlowRuntimeState` derived as `Partial<>`, the widening confined to requiredness; `ObjectLike` `Pick<>`ed to the two members it reads.
- **Renamed (3)** — a different concept wearing the spec's name: `AppLike` → `AppRouteLike` (reads `_packageId`, the ADR-0048 route key the spec's shape does not declare — it would arrive as `any` through that shape's index signature and **take the `typeof` guard with it**); `ObjectLike` → `MergedObjectLike` (carries `list`, which `MetadataProvider` merges on and the spec's closed shape cannot hold); `FlowEdge` → the package's **own existing** `FlowDesignerEdge`, which on reading turned out not to be a spec question at all but a third hand-copy of a local type two sibling files already import.
- **ALLOW: zero. Left in place: zero.** No reasoned waiver went stale.

Verified here: the ledger diff removes the whole `@object-ui/app-shell` DEBT group with exactly those six names, and the one remaining `"@object-ui/app-shell"` occurrence is a **pre-existing** entry in the declared-dialects map — it appears in no added line.

### ② Semver

**No package bump at all** — the changeset declares *"Internal refactor only — releases nothing."* Correct, and now independently supported: the emitted `.js` is byte-identical and the export set does not move, so there is nothing for a consumer to observe.

### ③ Boundary flags

- ⭐ **The ledger was re-derived, not inherited** — the dispatch's first task. `--ledger` run with dependencies installed, exit captured **by redirect before any pipe**. Real population **11 names in 5 packages**, corroborated by the gate's own banner. ⇒ the card's **14 is stale** and the predecessor's **11 was right** — asserted from the instrument rather than taken on either's word. After: **5 in 4 packages**.
- ⭐ **The root cause of that staleness was fixed, not just its instances.** The ledger comment's "All 11 are real mirrors" sentence is replaced with a **pointer to the instrument**: *"How many are left is deliberately NOT written here — `--ledger` regenerates the block and the run banner prints the count, so a figure copied into this comment could only be a second, staler answer (AGENTS.md #9)."* That is the rule whose breach produced the card's stale 14.
- ⭐ **The compile-time pins were proven to be compiled.** The file's own header records that its `Assert` lines were once *literally commentary*. So the pin was falsified on disk (blob change proven, anchor 1→0) and `tsc -p tsconfig.test.json` reddened with `TS2344` at the exact line; restored and proven **by state**.
- **`check:spec-floors` exits 1 with 3 findings — reported NOT MEASURED**, correctly: all three are `no-artifact` for packages this slice never touches, a whole-workspace-build precondition this run did not meet. ⚠️ And the floor question was **answered directly rather than left at an absence**: all four newly imported symbols are present in `@objectstack/spec@17.0.0`, the **declared floor**, fetched from the registry tarball, with a negative control returning 0.
- The lint narrowing carries all three readings, including a population read from **eslint's own resolved config via `isPathIgnored`** and a before/after comparison showing the diff adds no finding.
- Trailer conflict declared rather than resolved silently; the dev followed the dispatch contract's model-free form. That conflict is the standing objectui#9441.

### Ablation — three legs, and the third is why I trust the second

Predicted in writing before running. **Leg A (IN)**: restore the `RemoteTable` mirror ⇒ predicted red **by site**, explicitly *not* a stale-ledger message; observed the gate naming it at `DatasourceResourcePage.tsx:138` with `1 failed / 18 passed`. **Leg B (OUT)**: touch **zero source files**, delete one **OUT** row (`SortDirection`, `@object-ui/components`) ⇒ predicted red from a *different instrument naming a package this slice never edits*, app-shell pins green; observed exactly that. ⭐ **Leg C (control for B)**: same gate, unmutated tree, exit 0 — *so leg B's red is caused by the varied OUT row and not by "any edit reds the gate."*

That third leg is the difference between a mutation that reddens and a mutation that reddens **for the reason claimed**.

### Verdict

**PASS.** Carriers stripped in this same stroke; landing follows on a **full green** read.


---

